PR #261, opened 2020-05-12
replace carry with indexedarray

ianna:
Address issue #204 

Following a discussion with @jpivarski to finish what @Ellipse0934 has started in https://github.com/scikit-hep/awkward-1.0/pull/214

PR #314, opened 2020-06-30
Implement axis_wrap_if_negative

### LINKED ISSUE #163: Remember to 'implement axis_wrap_if_negative'
A lot of operations are using it, but right now it raises a "not implemented" runtime error if the `axis` really is negative. Fixing this would solve it for all of the operations.

ianna:
Start working on issue #163


jpivarski:
Thanks for taking this up! I just finished updating your #261 and will be merging it now.

jpivarski (changes requested):
It seems like the output of `axis_wrap_if_negative` is off by one. The tests don't test it because `argmax` is a reducer; reducers and sorting don't use `axis_wrap_if_negative`. Unfortunately, all of the methods that use `axis_wrap_if_negative` don't have NumPy equivalents (to motivate how it ought to work), but we still have a pretty good guide on how it ought to work. (Try negative `axis` in [np.size](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.size.html), for instance.)

Below, this three-dimensional array should allow `axis` to be `0`, `1`, or `2`, and these should be equivalent to `-3`, `-2`, `-1` (in that order). It's always the case that `-1` is the deepest allowed `axis`.

```python
>>> a = awkward1.Array(numpy.arange(3*5*2).reshape(3, 5, 2))
>>> awkward1.type(a)
3 * 5 * 2 * int64
>>> awkward1.num(a, axis=0)
3
>>> awkward1.num(a, axis=1)
<Array [5, 5, 5] type='3 * int64'>
>>> awkward1.num(a, axis=2)
<Array [[2, 2, 2, 2, 2], ... [2, 2, 2, 2, 2]] type='3 * 5 * int64'>

>>> awkward1.num(a, axis=3)     # this should fail because axis=3 is out of range
  File "/home/pivarski/irishep/awkward-1.0/awkward1/operations/structure.py", line 199, in num
    out = layout.num(axis=axis)
ValueError: 'axis' out of range for 'num'

>>> awkward1.num(a, axis=-1)    # therefore, -1 should be the deepest, equivalent to axis=2
<Array [5, 5, 5] type='3 * int64'>
>>> awkward1.num(a, axis=-2)   # axis=-2 should be equivalent to axis=1
3
>>> awkward1.num(a, axis=-3)   # axis=-3 should be equivalent to axis=0
3
```

The same should be true of ListType nodes (the above are RegularType). This makes the same off-by-one mistake, which leads to a different behavior in the `axis=-3` case, but that's probably unimportant.

```python
>>> a = awkward1.Array(numpy.arange(3*5*2).reshape(3, 5, 2).tolist())
>>> awkward1.type(a)
3 * var * var * int64
>>> awkward1.num(a, axis=0)
3
>>> awkward1.num(a, axis=1)
<Array [5, 5, 5] type='3 * int64'>
>>> awkward1.num(a, axis=2)
<Array [[2, 2, 2, 2, 2], ... [2, 2, 2, 2, 2]] type='3 * var * int64'>
>>> awkward1.num(a, axis=3)
  File "/home/pivarski/irishep/awkward-1.0/awkward1/operations/structure.py", line 199, in num
    out = layout.num(axis=axis)
ValueError: 'axis' out of range for 'num'
>>> awkward1.num(a, axis=-1)
<Array [5, 5, 5] type='3 * int64'>
>>> awkward1.num(a, axis=-2)
3
>>> awkward1.num(a, axis=-3)
  File "/home/pivarski/irishep/awkward-1.0/awkward1/highlevel.py", line 1145, in __repr__
    limit_value, layout, self._behavior
  File "/home/pivarski/irishep/awkward-1.0/awkward1/_util.py", line 1257, in minimally_touching_string
    lft = next(leftgen)
  File "/home/pivarski/irishep/awkward-1.0/awkward1/_util.py", line 1244, in forever
    for token in iterable:
  File "/home/pivarski/irishep/awkward-1.0/awkward1/_util.py", line 1150, in forward
    for token in forward(x[i], sp):
  File "/home/pivarski/irishep/awkward-1.0/awkward1/_util.py", line 1150, in forward
    for token in forward(x[i], sp):
ValueError: in ListOffsetArray64 attempting to get 0, offsets[i] != offsets[i + 1] and offsets[i + 1] > len(content)
```

Since these have been passed through to Python, we can see the off-by-one directlly:

```python
>>> a.layout.axis_wrap_if_negative(0)
0
>>> a.layout.axis_wrap_if_negative(1)
1
>>> a.layout.axis_wrap_if_negative(2)
2
>>> a.layout.axis_wrap_if_negative(-1)   # should be 2
1
>>> a.layout.axis_wrap_if_negative(-2)   # should be 1
0
>>> a.layout.axis_wrap_if_negative(-3)   # should be 0
-1
```

In NumPy, negative axis is a syntactic convenience, allowing you to not know or not care how many dimensions an array has but get the deepest axis or the second-deepest axis anyway. (Also, there's more of a symmetry between the first axis and the last axis, so it's just helping people count from the other end if they want to.)

But in Awkward, negative axis allows you to do things you wouldn't otherwise be able to do. Consider this array:

```python
>>> a = awkward1.Array([
...     {"x": [1], "y": [[], [1]]},
...     {"x": [1, 2], "y": [[], [1], [1, 2]]},
...     {"x": [1, 2, 3], "y": [[], [1], [1, 2], [1, 2, 3]]}])
```

Starting from the shallow end (positive `axis`), we can address `axis` values up to `1` because this is how deep `"x"` goes (`"y"` goes up to `2`).

```python
>>> awkward1.num(a, axis=0).tolist()
{'x': 3, 'y': 3}
>>> awkward1.num(a, axis=1).tolist()
[{'x': 1, 'y': 2}, {'x': 2, 'y': 3}, {'x': 3, 'y': 4}]
>>> awkward1.num(a, axis=2).tolist()
  File "/home/pivarski/irishep/awkward-1.0/awkward1/operations/structure.py", line 199, in num
    out = layout.num(axis=axis)
ValueError: 'axis' out of range for 'num'
```

But suppose we want the `num` of the deepest of both `"x"` and `"y"`? We should be able to express that with negative `axis`:

```python
>>> awkward1.num(a, axis=-1).tolist()
[4]+  Done                    emacs src/libawkward/array/ListOffsetArray.cpp
Segmentation fault (core dumped)
```

What I'd expect from this is

```python
[{'x': 1, 'y': [0, 1]}, {'x': 2, 'y': [0, 1, 2]}, {'x': 3, 'y': [0, 1, 2, 3]}]
```

To do that, the `axis_wrap_if_negative` method must refrain from converting negative `axis` to positive `axis` if there's a branch, such as the branch between `"x"` (depth 2) and `"y"` (depth 3). Only once the recursion has descended past the branch can it switch the negative `axis` into a positive `axis`.

That also means that cases like this have limited valid negative `axis` values. For instance, `axis=-2` in the above should be an error because it happens above the branch point, so it doesn't have an unambiguous meaning.

jpivarski:
I'm going to merge with master for you, then give you back the branch.

jpivarski:
It should be up to date now. (I'm not making any more changes.)

tests/test_0163_negative-axis-wrap.py
```diff
@@ -0,0 +1,94 @@
...
+    array = awkward1.Array([
+            {"x": [1], "y": [[], [1]]},
+            {"x": [1, 2], "y": [[], [1], [1, 2]]},
+            {"x": [1, 2, 3], "y": [[], [1], [1, 2], [1, 2, 3]]}])
+
+    # FIXME: this should throw
+    a = awkward1.num(array, axis=-2)
+    # <RecordArray>
+    # <field index="0" key="x">
+    #     <NumpyArray format="l" shape="" data="0x 03000000 00000000" at="0x7fd818c06c60"/>
```
ianna: @jpivarski - it looks like `to_list` cannot handle a `NumpyArray` with no shape.

ianna (commented):
@jpivarski - I left the `axis_out_of_range` test in intentionally. Should it throw a `ValueError`?

tests/test_0163_negative-axis-wrap.py
jpivarski: A NumpyArray with `len(shape) == 0` should never get to the Python level. `Content` subclasses sent to Python should pass through this `box` function ([here](https://github.com/scikit-hep/awkward-1.0/blob/5f1a068fec0d00936bb353318911855788a655ec/src/python/content.cpp#L14-L130)) that turns a scalar NumpyArray (one with `len(shape) == 0`) into a Python number. The same is true of `None` (the subclass of `Content` that should turn into a Python `None`). If this is what you're seeing, the error is upstream.

All of the functions that return `Content` subclasses in src/python/content.cpp are _supposed to_ pass their output through `box`, but there isn't anything in place to ensure this is the case.

ianna: Ah! Thanks, @jpivarski! More debugging then :-)

ianna:
@jpivarski - The tests pass as expected. I've also added more tests for the negative axis. Please, review when you have time. Thanks!

jpivarski (approved):
This is great! I just checked it out and tried it; everything is perfect, including error messages.

If you have no additional changes to make, I'll merge it.

ianna:
Thanks, @jpivarski !

PR #317, opened 2020-07-04
Remove redundant cpu kernels from operations.h

reikdas:


jpivarski (approved):
I was going to say, "Don't you need to remove their implementations as well?" But, as it turns out, they weren't even implemented! They're definitely hold-overs from an incomplete removal.

PR #319, opened 2020-07-05
CI test to check if documentation is generated correctly

### LINKED ISSUE #318: CI test to check if documentation is generated correctly
It would be nice to have a CI test that checks if the generated readthedocs documentation breaks.

reikdas:
Fixes #318

### DIFF
--- a/.ci/azure-doctest-awkward.yml
+++ b/.ci/azure-doctest-awkward.yml
@@ -45,9 +45,11 @@ jobs:
         displayName: "Install Pip"
 
       - script: |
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Great! This at least verifies that the documentation can be built without errors, though it doesn't check that the documentation looks right. (I can't think of a way to do that, and this is already a good step.)

I'll be merging it if you're done (which I gather from the fact that you've labeled it "not a draft").

reikdas:
I am done with this PR :)

Issue #315, opened 2020-07-01
Integer fancy indexing with nulls behavior


nsmith-:
In the spirit of expanding [integer fancy indexing](https://numpy.org/doc/stable/reference/arrays.indexing.html#integer-array-indexing) to masked arrays, let's review how we got here. We'll use
```
import numpy as np
import pyarrow as pa
import awkward1 as ak
```

For basic numpy arrays,
```
a = np.arange(20).reshape(4, 5)
b = np.array([[0,0], [1,1], [2,2], [3,3]])
c = np.array([[0,1], [2,3], [4,4], [0,0]])
```
we have that
```
>>> a[b, c]
 array([[ 0,  1],
       [ 7,  8],
       [14, 14],
       [15, 15]])
```
i.e. the output has the same shape as `b` and `c` always, and we have to index each dimension of `a` individually.
In awkward, for rectangular arrays, this still holds true:
```
a2 = ak.Array(a)
b2 = ak.Array(b)
c2 = ak.Array(c)
```
gives
```
>>> a2[b2, c2]
 <Array [[0, 1], [7, 8], [14, 14], [15, 15]] type='4 * 2 * int64'>
```

However, awkward does depart from numy as soon as a ListArray is involved. For example, even what appears to be the same array as `c`:
```
cj = ak.Array([[0,1], [2,3], [4,4], [0,0]])
```
we can use `cj` by itself to achieve what previously used both `b` and `c`:
```
>>> ak.type(cj)
 4 * var * int64
>>> a2[cj]
 <Array [[0, 1], [7, 8], [14, 14], [15, 15]] type='4 * var * int64'>
```
Notice `var` in the type of `cj`, which causes the indexing operation to be interpreted differently. We've accepted this as a departure from numpy since the ListArray is a completely different beast and under "our control". Yet, the "shape" of the output continues to match `cj`, i.e. it has the same number of elements in each dimension.

In pyarrow, masked arrays inside indexing operations similarly preserve the "shape" of the indexer, where here we mean the length of the output array (as pyarrow only supports 1D arrays).
```
d = pa.array([0, 1, 2, None, 4, 5])
e = pa.array([0, 2, None, 3])
```
we have
```
>>> d.take(e)
 <pyarrow.lib.Int64Array object at 0x120e1b980>
[
  0,
  2,
  null,
  null
]
```
and awkward faithfully reproduces this:
```
d2 = ak.Array([0, 1, 2, None, 4, 5])
[...]

PR #322, opened 2020-07-07
Revised getitem operation for masked jagged indexers

nsmith-:
First pass, some new kernels will need to be made.


src/libawkward/Content.cpp
```diff
@@ -1275,6 +1306,11 @@ namespace awkward {
                                   "with NumPy-style advanced indexing");
     }
 
+    // would length ever not be 1 when the content is jagged?
+    if ( length() == 1 && dynamic_cast<SliceJagged64*>(missing.content().get())) {
+      return getitem_next_missing_jagged(missing, tail, advanced, shallow_copy());
+    }
```
jpivarski: I think the length might be != 1 if some of your tests were one level of nesting deeper. So if

```python
    a = awkward1.Array([[0, 1, 2], None, [5, 6]])
    c = awkward1.Array([[False, True, False], [False, True], [True, False]])
    assert awkward1.to_list(a[c]) == [[1], None, [5]]
```

were

```python
    a = awkward1.Array([[[0, 1, 2], None], [[5, 6]]])
    c = awkward1.Array([[[False, True, False], [False, True]], [[True, False]]])
    assert awkward1.to_list(a[c]) == [[[1], None], [[5]]]
```

you _might_ find that the length is not 1.

src/libawkward/array/ByteMaskedArray.cpp
```diff
@@ -1297,9 +1297,29 @@ namespace awkward {
...
+        // as if they were nulls. We would have to hack a bit on nextcarry and outindex though
+        if ( from >= 0 ) {
+          carrystarts.setitem_at_nowrap(k, slicestarts.getitem_at_nowrap(i));
+          carrystops.setitem_at_nowrap(k, slicestops.getitem_at_nowrap(i));
+          k++;
+        }
+      }
+      if ( k != carrystarts.length() ) {
+        throw std::runtime_error("failed to carry starts and stops properly");
+      }
```
jpivarski: You write kernels the same way I do: by first doing it inline, then splitting them out once everything works. The highlighted section is the part that would become a kernel, and you'd replace `setitem_at_nowrap` and `getitem_at_nowrap` with normal square brackets. You should also change the name of `from` since @reikdas auto-converts these into Python for documentation ([here](https://awkward-array.readthedocs.io/en/latest/_auto/kernels.html)) and that's a Python keyword.

You've caught on to the fact that I use `std::runtime_error` (`RuntimeError`) for errors that are our fault and `std::invalid_argument` (`ValueError`) for errors that are the user's fault. You can't raise that exception in a kernel, though, the only way to get errors out (`return success()` and `return failure(...)`) get mapped to `std::invalid_argument`.

src/libawkward/array/IndexedArray.cpp
```diff
@@ -2281,9 +2281,29 @@ namespace awkward {
...
+        // as if they were nulls. We would have to hack a bit on nextcarry and outindex though
+        if ( from >= 0 ) {
+          carrystarts.setitem_at_nowrap(k, slicestarts.getitem_at_nowrap(i));
+          carrystops.setitem_at_nowrap(k, slicestops.getitem_at_nowrap(i));
+          k++;
+        }
+      }
+      if ( k != carrystarts.length() ) {
+        throw std::runtime_error("failed to carry starts and stops properly");
+      }
```
jpivarski: Another kernel.

src/libawkward/array/ByteMaskedArray.cpp
jpivarski: These kernels would go into getitem.h/getitem.cpp, by the way.

src/libawkward/Content.cpp
jpivarski: On the other hand, maybe not. I seemed to think so:

https://github.com/scikit-hep/awkward-1.0/blob/bb205f3ff9f5db5c131ff53ec85cd3ebac62a1e8/src/libawkward/Content.cpp#L1226-L1227

src/libawkward/Content.cpp
```diff
@@ -1275,6 +1306,11 @@ namespace awkward {
...
+
     ContentPtr tmp = check_missing_jagged(shallow_copy(), missing);
```
jpivarski: Is `check_missing_jagged` relevant anymore? If these if statements (checking all option-type layout nodes) now have zero coverage in the test suite, then they're dead code and can be removed (maybe pulling the only non-option-type test out):

https://github.com/scikit-hep/awkward-1.0/blob/bb205f3ff9f5db5c131ff53ec85cd3ebac62a1e8/src/libawkward/Content.cpp#L1226-L1257

src/libawkward/Content.cpp
```diff
@@ -1189,6 +1189,37 @@ namespace awkward {
...
+  const ContentPtr getitem_next_missing_jagged(const SliceMissing64& missing,
+                                               const Slice& tail,
+                                               const Index64& advanced,
+                                               const ContentPtr& that) {
+    const SliceJagged64* jagged = dynamic_cast<SliceJagged64*>(missing.content().get());
+    const Index64 index = missing.index();
+    ContentPtr content = that.get()->getitem_at_nowrap(0);
+    if ( content.get()->length() != index.length() ) {
+      throw std::invalid_argument("Contents length does not match length of masked jagged slice");
+    }
```
jpivarski: The content length is allowed to be longer than the length that its parent can see, and hence longer than the slice, so I think the error condition should be `content.get()->length() < index.length() `.

src/libawkward/Content.cpp
```diff
@@ -1189,6 +1189,37 @@ namespace awkward {
...
+      starts.setitem_at_nowrap(i, start);
+      if ( index.getitem_at_nowrap(i) < 0 ) {
+        outputmask.setitem_at_nowrap(i, -1);
+        stops.setitem_at_nowrap(i, start);
+      } else {
+        outputmask.setitem_at_nowrap(i, i);
+        int64_t stop = jagged->offsets().getitem_at_nowrap(++k);
+        stops.setitem_at_nowrap(i, stop);
+      }
+    }
```
jpivarski: Another kernel! (I noticed this one out of order.)

src/libawkward/array/ByteMaskedArray.cpp
jpivarski: Note that when an `Index` (such as `outindex`, `slicestarts`, and `slicestops`) are passed into a kernel, you have to pass the raw pointer (`.ptr().get()`) and its offset (`.offset()`). Both pieces of information are needed to get at logical index `i` of an array, since one `Index` may be sharing a buffer with another `Index`, starting at a different offset: logical index `i` is at `offset + i`. It's a downside of forcing everything through a C interface. If I had been more organized at the beginning, I might have defined some C structs to carry both pieces of information over, though it would still be non-standard syntax to call a function that adds the offset in every get/put.

Except for one special case, _output_ `Indexes` can be assumed to have offset = 0 (because you just allocated it before entering the kernel; of course the offset is 0!). The special case involves concatenation.

src/libawkward/array/IndexedArray.cpp
```diff
@@ -2303,6 +2323,7 @@ namespace awkward {
       util::handle_error(err, classname(), identities_.get());
 
       ContentPtr next = content_.get()->carry(nextcarry);
+      // shouldn't we carry starts and stops here as well?
       return next.get()->getitem_next_jagged(slicestarts,
                                              slicestops,
                                              slicecontent,
```
jpivarski: It doesn't look to me like anything special is needed here. The non-option-type case is very different from the option-type case, and the feature you're adding only affects option-type cases.

From just staring at it, though, I can't tell if what's here is right. If this section has coverage, then it's very likely right.

jpivarski (commented):
Thanks a lot! I just have a lot of little comments.

src/libawkward/array/IndexedArray.cpp
nsmith-: This is in fact the same exact kernel as https://github.com/scikit-hep/awkward-1.0/pull/322#discussion_r450971475

jpivarski: Okay, that's how you counted a total of 2.

src/libawkward/Content.cpp
nsmith-: I think this new behavior does supersede that whole section, but will check.

src/libawkward/array/ByteMaskedArray.cpp
nsmith-: Why not pass the shifted pointer into the function instead?

src/libawkward/Content.cpp
nsmith-: Here the parent is the "dummy" length-1 RegularArray wrapper. I think the content should exactly match the length of the mask, otherwise we allow things like:
```
a = awkward1.Array([[0, 1, 2], [3, 4], [5, 6], [7]])
b = awkward1.Array([[0], [0], [0], None, [0], [0]])
a[b]
```
which I don't see a good reason to allow

src/libawkward/array/IndexedArray.cpp
nsmith-: It has coverage and I don't think there's any issue. I realized what I was doing with the starts and stops isn't really _carry_ but more like reduce

src/libawkward/Content.cpp
```diff
@@ -1189,81 +1189,48 @@ namespace awkward {
...
+      throw std::invalid_argument(
+        std::string("cannot fit masked jagged slice with length ")
+        + std::to_string(index.length()) + std::string(" into ")
+        + that.get()->classname() + std::string(" of size ") + std::to_string(content.get()->length()));
     }
-    return that;
+    Index64 outputmask(index.length());
+    Index64 starts(index.length());
+    Index64 stops(index.length());
+    Content_getitem_next_missing_jagged_getmaskstartstop(index.ptr().get() + index.offset(),
```
nsmith-: Here's my example of incrementing the pointer before passing to C function. We could provide a convenience function in `Index` type to do that in one call as well. Any reason not to go this route and pass the offset into the kernel?

src/libawkward/Content.cpp
```diff
@@ -1189,81 +1189,48 @@ namespace awkward {
...
+                                                         jagged->offsets().ptr().get() + jagged->offsets().offset(),
+                                                         outputmask.ptr().get(),
+                                                         starts.ptr().get(),
+                                                         stops.ptr().get(),
+                                                         index.length());
+    ContentPtr tmp = content.get()->getitem_next_jagged(starts, stops, jagged->content(), tail);
+    IndexedOptionArray64 out(Identities::none(), util::Parameters(), outputmask, tmp);
+    return std::make_shared<RegularArray>(Identities::none(), util::Parameters(), out.simplify_optiontype(), index.length());
```
nsmith-: I cribbed this line from `getitem_next_regular_missing` but I can't say I understand why I have to re-wrap the array with a RegularArray, and why it has length `index.length()` rather than the length `1` that it came in with.

src/libawkward/Content.cpp
nsmith-: I'm leaving a canary runtime_error when length is not 1 in for now. Maybe further testing will find it.

src/libawkward/Content.cpp
nsmith-: None of the unit tests reached that code after this change, so I deleted it. I think it makes sense too.

src/libawkward/array/ByteMaskedArray.cpp
jpivarski: That... is a great idea. Maybe a future refactoring will do that. It could be a method on `Index` to give the actual beginning of the data; it has to reference-count the base pointer, but there can be special methods to give something convenient to the C functions...

src/libawkward/Content.cpp
jpivarski: Contents are allowed to be longer than the nodes that contain them (i.e. have unreachable elements) because of how they're sometimes created: making several arrays is not an atomic operation, but if appends are performed from the innermost to the outermost, allowing long contents to be valid means that the whole array is valid for each of the atomic steps.

An example is if ArrayBuilder is filling some interior content with `begin_list` or `begin_record` that hasn't yet been paired with an `end_list` or `end_record`. During that time, the array as a whole is valid; it just doesn't have the partial list or record, though you might see some of the partial contents if you investigate its `content` during this time.

As for slices, I think the logic carries over. I think.

src/libawkward/Content.cpp
jpivarski: Exactly: I'm catching up to you. That would be a good refactoring, but I think this should be done in a single sweep, one that I should probably do someday. However, that should be done all at once when no one else is working on the kernels (@reikdas and @trickarcher).

For right now, just make them fit the established pattern. The improvement should happen all at once.

src/libawkward/Content.cpp
jpivarski: I'm not close enough to the details at the moment to give advice. Is `index.length()` equal to `1` in all your examples? It might be equal all the time and the use of `index.length()` is just a matter of being formal.

src/libawkward/Content.cpp
nsmith-: Are you saying I should put the offset into the kernel signature for this PR as well? Or can I leave it as-is?

jpivarski: For the purposes of factorization, just add the offset as a separate function argument to fit the pattern. Then when you or I am converting them all in bulk (in a new PR), the deviation from the pattern won't be confusing.

I just noticed that you did that by simply adding to the pointer. I thought that would be a type error—I thought C++ strictly separated pointers and integers. I can see how it knows about the `sizeof` when adding—it's some overloaded operator. Does it only work in some version of the standard?

When you or I update these, I think it should be in a named method (`rawdata`?), so that it's easier to do future sweeps through the code. (Waking through by incremental text search is reliable if semantic constructs always have the same textual pattern. In the case of this sweep, the text to search is `pter().get()` because that occurs in all instances where we'll want to change and few others.)

nsmith-: Ok. Regarding pointer + integer, it's C standard: http://c0x.coding-guidelines.com/6.5.6.html 1167
Although there is some nuance to whether the pointer actually points to an array.

src/libawkward/Content.cpp
nsmith-: I changed it as you suggested. Hopefully it doesn't surprise us later

src/libawkward/Content.cpp
nsmith-: In fact index.length() is long, as long as the slice. It seems to be used also in https://github.com/scikit-hep/awkward-1.0/blob/master/src/libawkward/array/ListArray.cpp#L1644-L1646 so I think it is consistent with other uses. I guess a calling function unwraps it.

nsmith-:
I think the functionality is there now. I think the boilerplate for kernels is pretty out of hand, would be nice to find a way to reduce it. Remaining items:

- How do I auto-format to 80 characters?
- Where do I put some documentation of this behavior?
- Why is the doctest failing

jpivarski:
> * How do I auto-format to 80 characters?

We don't have a tool for that; I did it by hand. If it's roughly or mostly 80 characters, that is fine.

> * Where do I put some documentation of this behavior?

`__getitem__` behavior is documented here: https://github.com/scikit-hep/awkward-1.0/blob/cfa3a04d679529175e742fb5015047bd499ea1a6/src/awkward1/highlevel.py#L507-L843

> * Why is the doctest failing

Merge upstream master; it was fixed today (PR #323).

jpivarski:
I'm going to merge your branch with master, which should fix the doctest.

After that, it's done, right? You're not planning any more changes?

jpivarski:
We're going to ignore the doctest error; I'll ask Reik to look into that after all the merges are done.

Assuming all the other tests pass, should I merge this? I'm dealing with the compiler warnings on another PR.

nsmith-:
Ok, now that its formatted and docs are added, I think its ready

nsmith-:
closes #322

nsmith-:
Closes #315

PR #323, opened 2020-07-08
Fixed Python generation

reikdas:
There was an error in the parser if the return type of a function was present in the name of the function.

Should fix the doc generation CI error in @trickarcher 's branch.


jpivarski (approved):
Looks good. Only replace the first instance.

(Though I wonder if you have some brittle parsing here, but that's a more general question...)

PR #327, opened 2020-07-09
Lookahead Python code generation for assignment with ArrayRef node having incr operator

reikdas:
Extending the C parser to handle more cases.

Should fix the failing CI documentation generation tests.


jpivarski (approved):
Thanks! It looks like the doctest passed, but I want to wait for all the other tests to finish before merging.

PR #330, opened 2020-07-12
Attach docstrings to newly created highlevel arrays

nsmith-:
There's no way to do this from the mixin since there is no initialize hook.
A possible alternative, which might also be useful for other things, would be to rewrite `Array.__init__` using `__new__` and allowing the mixin to override initialization.


jpivarski (approved):
This is good. I like this!

As soon as the tests pass, I'll merge it.

PR #337, opened 2020-07-14
Try to fully resolve the NumPy format string issues.

### LINKED ISSUE #313: Make NumpyArray's format checks insensitive to '=' (or '<' if little-endian, '>' if big-endian) at the beginning of the string
For minimal modification (and a tiny performance advantage), this can be done in the NumpyArray constructor. (Also check the NumpyForm and conversions from JSON into NumpyForms.) That way, the unnecessary character can be removed once in one place, without having to fix every instance in NumpyArray where `format_` is used.

If NumPy arrays are created as big-endian and then are converted with `astype` to little-endian, they have this extra character in their formats. Uproot4 is using a work-around so that a new version of Awkward with this fix is not needed right away, but it's important to fix anyway.

This would be an easy fix, though it's entirely in C++, not Python.

### LINKED ISSUE #333: Layout indexes don't support slices
Sometimes, when building arrays at a low level, it is useful to manipulate `Index64` and similar from python.
The usual getitem method does not accept python slices, however:
```python
import awkward1
a = awkward1.Array([[1, 2, 3], [4, 5], [6]])
a.layout.offsets[:-1]
```
gives:
```python
<ipython-input-30-53dc196bd178> in <module>
      2 
      3 a = awkward1.Array([[1, 2, 3], [4, 5], [6]])
----> 4 a.layout.offsets[:-1]

TypeError: __getitem__(): incompatible function arguments. The following argument types are supported:
    1. (self: awkward1._ext.Index64, arg0: int) -> int
    2. (self: awkward1._ext.Index64, arg0: int, arg1: int) -> awkward1._ext.Index64

Invoked with: <Index64 i="[0 3 5 6]" offset="0" length="4" at="0x7f869cad1a00"/>, slice(None, -1, None)
```

Amusingly, one can do `a.layout.offsets.__getitem__(0, -1)`

### LINKED ISSUE #335: Numpy dtype of Index64
It seems the dtype of numpy arrays converted from awkward Index64 have a dtype roundtrip issue. Consider:
```python
import awkward1
import numpy

a = awkward1.layout.Index64([0, 1, 2, 3])
b = numpy.asarray(a)
```
`b.dtype.char` is `q`, while the platform I am on expects `l`. Numpy handles this gracefully:
```python
>>> numpy.arange(4).dtype.char
'l'
>>> b + numpy.arange(4)
array([0, 2, 4, 6], dtype=int64)
```
however awkward1 does not:
```python
>>> b + awkward1.Array([0, 1, 2, 3])
  File "/Users/ncsmith/src/awkward-1.0/awkward1/highlevel.py", line 1169, in __repr__
    typestr = repr(str(awkward1._util.highlevel_type(layout, self._behavior, True)))
  File "/Users/ncsmith/src/awkward-1.0/awkward1/_util.py", line 1110, in highlevel_type
    return awkward1.types.ArrayType(layout.type(typestrs(behavior)), len(layout))
ValueError: Numpy format "q" cannot be expressed as a PrimitiveType
```

It is sufficient to recast (despite the dtype not actually changing other than `char`):
```python
>>> b.astype('i8') + awkward1.Array([0, 1, 2, 3])
<Array [0, 2, 4, 6] type='4 * int64'>
```

jpivarski:
- [X] `util::dtype` enum for all the recognized primitives.
  - [X] functions for converting between `(format, itemsize)` and `util::dtype`.
  - [x] `util::dtype npdtype` as a third type parameter in NumpyArray and NumpyForm, part of the constructor.
  - [x] Replace uses of `format` with uses of `npdtype` without changing behavior.
  - [x] Prefer `npdtype` when converting NumpyArray to and from np.ndarray.
  - [x] Use the same mechanism for `IndexOf<T>`.
  - [ ] ~~Extend NumpyArray implementations to all `util::dtype` values (except `NOT_PRIMITIVE`).~~
  - [ ] ~~Extend reducers as well.~~
  - [x] Revisit the linked issues to see if they've been fixed:
    - [x] #313
    - [x] #333
    - [x] #335
  - [x] NumpyForm JSON string representations should cover all of the `util::dtype` values (except `NOT_PRIMITIVE`).
  - [x] We need a function that turns NumPy dtypes into NumpyForms. The JSON trick is dumb.


jpivarski:
Initial survey: https://github.com/pybind/pybind11/issues/1908#issuecomment-658358767

PR #340, opened 2020-07-16
Convert 64-bit unsigned 32-bit Awkward arrays into 32-bit Arrow arrays if their indexes are small enough.

### LINKED ISSUE #312: ak.to_arrow should convert to string instead of large_string
If you load the bikeroutes from [here](https://github.com/jpivarski/2020-07-06-scipy2020/blob/c951aa50bbfa879f011c9f2c343a3d0168be28e0/bikeroutes-execution.ipynb) and try to convert to arrow, it works:
```
arr2 = ak.to_arrow(bikeroutes['features']). # cannot convert the top-level Record, so lose some attributes
t = pa.Table.from_arrays([arr2], names=['features'])
```
but this cannot be written to parquet because the schema is as follows
```
features: struct<type: large_string, properties: struct<STREET: large_string, TYPE: large_string, BIKEROUTE: large_string, F_STREET: large_string, T_STREET: dictionary<values=large_string, indices=int64, ordered=0>>, geometry: struct<type: large_string, coordinates: large_list<item: large_list<item: large_list<item: double>>>>>
  child 0, type: large_string
  child 1, properties: struct<STREET: large_string, TYPE: large_string, BIKEROUTE: large_string, F_STREET: large_string, T_STREET: dictionary<values=large_string, indices=int64, ordered=0>>
      child 0, STREET: large_string
      child 1, TYPE: large_string
      child 2, BIKEROUTE: large_string
      child 3, F_STREET: large_string
      child 4, T_STREET: dictionary<values=large_string, indices=int64, ordered=0>
  child 2, geometry: struct<type: large_string, coordinates: large_list<item: large_list<item: large_list<item: double>>>>
      child 0, type: large_string
      child 1, coordinates: large_list<item: large_list<item: large_list<item: double>>>
          child 0, item: large_list<item: large_list<item: double>>
              child 0, item: large_list<item: double>
                  child 0, item: double
```
Arrow doesn't know how to write the "large strings", but none of the input actually has any strings long enough to need that, would be 2**31 bytes, 2GB. Of course, arrow *should* be able to cope, but it does not.

jpivarski:
Fixes #312, as preparation for adding Parquet input and output.

### DIFF
--- a/src/awkward1/operations/convert.py
+++ b/src/awkward1/operations/convert.py
@@ -1326,6 +1326,16 @@ def to_arrow(array):
             layout,
             (awkward1.layout.ListOffsetArray64, awkward1.layout.ListOffsetArrayU32),
         ):
+            offsets = numpy.asarray(layout.offsets)
+
+            if len(offsets) == 0 or numpy.max(offsets) <= numpy.iinfo(numpy.int32).max:
+                small_layout = awkward1.layout.ListOffsetArray32(
+                    awkward1.layout.Index32(offsets.astype(numpy.int32)),
+                    layout.content,
+                    parameters=layout.parameters,
+                )
+                return recurse(small_layout, mask=mask)
+
             offsets = numpy.asarray(layout.offsets, dtype=numpy.int64)
 
             if layout.parameter("__array__") == "bytestring":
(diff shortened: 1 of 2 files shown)


martindurant:
Thank you! Does this cover largebytes and largestrings too?

jpivarski:
Yes; in Awkward, strings are lists with special parameters, so converting all lists that can be 32-bit into 32-bit gets the strings as well. The original unit tests don't seem to check the types of any Arrow strings, though.

I'm having second thoughts about converting all lists, since we might like to pass lists with 64-bit offsets directly to Arrow without a copy if possible. But since Arrow started with everything being 32-bits, the safest thing to do is to use 32-bit wherever possible and consider the avoidance of a down-conversion a future performance optimization.

I was about to start the Parquet reader/writer, depending on how much time I have this afternoon. #312 seemed to be a blocker.

martindurant:
>  we might like to pass lists with 64-bit offsets directly to Arrow

(but not for parquet, so I would defer this, if possible)

PR #343, opened 2020-07-16
Convert Arrow <--> Parquet, and hence Awkward <--> Parquet.

### LINKED ISSUE #172: Copy the OAMap Parquet samples and make awkward.arrow.fromparquet tests
The samples can be found here: https://github.com/diana-hep/oamap/tree/master/tests/samples

They cover a lot of cases.

### LINKED ISSUE #303: Does awkward want to promote itself yet as a general nested processor?
Prompted by [this stackoverflow question](https://stackoverflow.com/questions/62410128/how-should-i-process-nested-data-structures-e-g-json-xml-parquet-with-dask), to which my internal answer was "that's what awkward is for". Is there yet an example of how you might do the circuit of parquet file with complicated schema, to numba-jit function and fast processing.

jpivarski:


jpivarski:
@martindurant As expected, Parquet support was rather easy—I just had to copy the old code. Awkward PartitionedArrays translate into row groups, top-level RecordArrays translate into record batches, and any other kind of array is presented as a record batch with one field whose name is the empty string. (This convention is also used when reading back: if there's only one field and its name is the empty string, we read back into a non-empty array.)

I also tested it on some ancient samples I made for OAMap, but even after all these years, most of the data structures are not supported by pyarrow:

https://github.com/scikit-hep/awkward-1.0/blob/97b165e53666d13a88da247ea18d577bfdf85761/tests/test_0341-parquet-reader-writer.py#L105-L246

To mitigate this, I added an `explode_records` option to `ak.to_parquet`. The transformation isn't lossy, but the result would have to be read back into [ak.zip](https://awkward-array.readthedocs.io/en/latest/_auto/ak.zip.html). Not ideal.

But hey, we have both eager and lazy reading now!

PR #345, opened 2020-07-18
Prepare the Python Layer for the CUDA Kernels, and add Docker Images for CI

trickarcher:


jpivarski:
I have this checked out and I've been tinkering with it: I started `nvidia-smi -l 1` so that I see

```
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      3004      G   /usr/lib/xorg/Xorg                           174MiB |
|    0      3157      G   /usr/bin/gnome-shell                          84MiB |
|    0      4505      G   ...AAAAAAAAAAAACAAAAAAAAAA= --shared-files   109MiB |
|    0      8330      G   /opt/zoom/zoom                                18MiB |
+-----------------------------------------------------------------------------+
```

Then when I

```python
>>> import awkward1 as ak
>>> import numpy as np
>>> array = ak.Array(np.arange(1000000).reshape(-1, 1))
>>> array2 = ak.copy_to(array, "cuda")
>>> array2.layout
<NumpyArray format="l" shape="1000000 1" data="0x ..." at="0x7f72e2000000">
    <Lib name="cuda" device_number="0" device_name="GeForce GTX 1060"/>
</NumpyArray>
```

the monitor bumps up to

```
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      2935      C   python                                        61MiB |
|    0      3004      G   /usr/lib/xorg/Xorg                           174MiB |
|    0      3157      G   /usr/bin/gnome-shell                          84MiB |
|    0      4505      G   ...AAAAAAAAAAAACAAAAAAAAAA= --shared-files   109MiB |
|    0      8330      G   /opt/zoom/zoom                                18MiB |
+-----------------------------------------------------------------------------+
```

Good! It's allocated 61 MiB on the GPU.

Next, I expected the `ak.num` operation to entirely take place on the GPU, but it does not:

```python
>>> three = ak.num(array2, axis=1)
>>> three.layout
<NumpyArray format="l" shape="1000000" data="1 1 ... 1 1" at="0x7f730365e010"/>
```

I was expecting `three` to be a GPU array as well, which would have bumped the memory usage up to 122 MiB, but it stayed at 61 MiB.

```
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0      2935      C   python                                        61MiB |
|    0      3004      G   /usr/lib/xorg/Xorg                           174MiB |
|    0      3157      G   /usr/bin/gnome-shell                          84MiB |
|    0      4505      G   ...AAAAAAAAAAAACAAAAAAAAAA= --shared-files   109MiB |
|    0      8330      G   /opt/zoom/zoom                                18MiB |
+-----------------------------------------------------------------------------+
```

(so what I'm looking at isn't even a copy of a GPU-resident array; the only copy is in main memory).

Did you intend for it to work that way, or is it an error in the implementation? When I said that we'd need GPU arrays to stay on the GPU, I meant all of their derivatives as well.

That might not need to be in this PR, but we'll need it to work that way eventually.

(Meanwhile, I'm still working on this Docker image; I don't use Docker much but I'm figuring it out.)

jpivarski:
## What the Docker image should include

The _base_ from which to launch a test. This must not include recent code changes or any compiled artifacts, but it can include the git repository as it was at some date so that launching a test starts with `git pull`, rather than `git clone`.

I think the system Python that comes with Ubuntu 18.04 is too old. I know that Python 3.6 is one of our test versions, but something related to the system Python is sufficiently out of date that pip isn't compiling the cuda-kernels. It's generally a bad idea to use system Pythons anyway—we'll have troubles down the road if we follow that route. Therefore, I think we should use conda to install not just the Python, but the compilers and everything (to be sure we can update them as needed).

## What the test should execute

It should `git pull` to a specified branch (it will have to be possible to pass a script option into the `docker run`), pip-install the cuda-kernels, compile the main Awkward codebase any way we like (therefore, localbuild.py), and then run the `tests_cuda`.

## Working Dockerfile.build

In a completely clean recursive git-clone of awkward-1.0 (master branch), I put the following files:

```
########################## Dockerfile.build

FROM nvidia/cuda:10.2-devel-ubuntu18.04

WORKDIR /awkward-1.0

ENV PATH="/root/miniconda3/bin:${PATH}"
ARG PATH="/root/miniconda3/bin:${PATH}"

RUN apt-get update && apt-get install -y emacs-nox wget

RUN wget \
    https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh \
    && mkdir /root/.conda \
    && bash Miniconda3-latest-Linux-x86_64.sh -b \
    && rm -f Miniconda3-latest-Linux-x86_64.sh 

RUN conda config --add channels conda-forge

RUN conda update conda -y

RUN conda install pip git gcc_linux-64 gxx_linux-64 cmake numpy cupy pytest -y

RUN git config --global pull.rebase false

COPY . .
```

and

```bash
########################## do-tests.sh

#!/bin/bash

set -e

conda --version
python --version

git checkout $1
git pull
git status

cp -a include src VERSION_INFO cuda-kernels

cd cuda-kernels
pip install .
cd ..

python -c 'import awkward1_cuda_kernels; print(awkward1_cuda_kernels.shared_library_path)'

python localbuild.py --pytest tests

python -m pytest -vvrs tests_cuda
```

and build it with

```bash
docker build -f Dockerfile.build -t awkward1-cuda-tests:1.0-cuda10.2 .
```

So the Dockerfile is versioned with `1.0` (our version) and `-cuda10.2` (the CUDA image it's base on), and it contains

   * a text editor for debugging (I include `emacs-nox`; I don't know if you want vi or something)
   * wget, which is needed to get Miniconda
   * an updated Miniconda distribution based on conda-forge
   * git, compilers (gcc), CMake from conda
   * Python 3.7 (with all dev files for compiling)
   * Python packages we'll need, including NumPy, CuPy, and pytest (pybind11 is included in one of our submodules)
   * a clean git checkout of scikit-hep/awkward-1.0 (master branch)
   * the `do-tests.sh` file

When we run

```bash
docker run awkward1-cuda-tests:1.0-cuda10.2 sh do-tests.sh trickarcher/python_cuda_interface
```

the `do-tests.sh` file:

   1. checks out the branch we want to test and does a `git pull`
   2. copies the `include`, `src`, `VERSION_INFO` files into `cuda-kernels` because pip is running in a different directory than we thought it might (IMPORTANT: the copy must happen repeatedly for each test, so that these files do not become stale—it's very important that this is not "baked into" the Docker image)
   3. does a pip install of awkward1-cuda-kernels
   4. compiles and tests the main Awkward codebase
   5. tests the `tests_cuda`.

## How this should be arranged in the git repo

Dockerfile.build for the different versions of CUDA should be in the `dev` directory (possibly a subdirectory) because it's something that a developer would do rarely. Also, it's not going to be done once for some value of

```bash
modinfo nvidia --field version
```

because the version that we have on the machine that builds the Docker images is not necessarily the version that we have on the cloud instance. I'll be investigating AWS soon, to see what it has (and how to run Docker images on it).

The `do-tests.sh` should probably have a name related to CUDA, such as `run-cuda-tests.sh`, and it may be inside `.ci` because it's something that runs tests. It should probably have executable permissions so that we don't have to run it from `sh`, but I was able to test it above without that.

The `tests_cuda` directory should be `tests-cuda` and should include an empty `__init__.py` file (required by old Pythons, now just a convention). The `pytest` command should be run from within `python` and have `-vvrs` options (totally verbose, listing skipped tests). That will make it easier to debug.

**Now I need to find out how to run Docker images on AWS, so that we'll have a procedure in place to run this.**

trickarcher:
@jpivarski  Thanks for looking into my commits. That `ak.num` was a serious bug. So first things first, `num` wasn't implemented for `RegularArray`(`NumpyArray` calls the `RegularArray` `num` kernel) so the program didn't crash because we are using Unified Memory. After I wrote the kernel for it, I noticed that there was a constant `bus error` happening which caused the program to crash. I fixed it because someone posted the same issue as a `gist` on github, by adding `cudaDeviceSynchronize()` at the end of the `cuda launch` call. This behaviour is very inconsistent in the sense that it would sometimes occur for the other `num`  kernel I had written(`ListArray`). I would like to investigate this further, but an obvious guess would be that something dangerous was happening on the lower levels of `CUDA` . As a note to self, always add `cudaDeviceSynchronize()` , after a kernel launch.

jpivarski (approved):
It works! I used the AWS instance to verify that, and also to try it out interactively and see that the arrays stay on the GPU, even after deriving something like `ak.num`.

**Awesome!**

The Docker images on the AWS instance are a bit of a mess (we'll run out of disk space if you don't purge soon). The latest images contain a checked-out copy of this branch with local changes, making it impossible to run the push-button testing script because it wants to `git stash`. The Docker image should be a clean copy of master onto which the latest changes are pulled ephemerally (i.e. after it runs, it reverts to the clean state and has to pull and compile again). That way, we don't need to make new Docker images as the code changes—GitHub changes and the Docker image pulls from GitHub each time it starts. In fact, the Docker image doesn't even need to have the repo on it at all—it could start with a `git clone` instead of a `git checkout` and `pull`. Considering that it has to recompile, adding the extra 10 seconds for `git clone` wouldn't matter much.

Some of that is in the GitHub repo; I'd like to get that testing procedure streamlined in a subsequent PR, removing such files as `.dockerignore` (we wouldn't have to ignore temporary files if we never build images using any temporary files). But I want to merge this, and the testing infrastructure can be cleaned up later.

The main thing is: we have GPU arrays! Yay!

Issue #328, opened 2020-07-10
Specifying dtype (for converting nested lists to ragged arrays or directly with ArrayBuilder)


ChristianMichelsen:
Hi,

First and foremost thanks for a really useful tool! 

I am having trouble specifying the dtype for awkward arrays. For simple arrays it works well, e.g:
```python
import numpy as np
import awkward1 as ak

x = np.arange(5, dtype=np.int8)

ak.Array(x)
ak.from_numpy(x)
```
Here both methods preserve the dtype: `<Array [0, 1, 2, 3, 4] type='5 * int8'>` (compared to e.g. `ak.from_iter(x)`: `<Array [0, 1, 2, 3, 4] type='5 * int64'>`. )

My problem is that I have to create nested lists (or, preferably, ragged arrays) of specific dtype (to reduce memory). 
Right now I make the nested lists in numba:

```python

N = 5
nested_list = initialize_nested_lists(2, dtype=np.int8) 
for i in range(N):
    rand_int = np.random.randint(2)
    nested_list[rand_int].append(i)
```
which gives the following output:
```python
ListType[ListType[int8]]([[1, 2, 3], [0, 4]])
```

How do I convert such a nested list to an awkward array while still keeping the correct dtype? 

I have tried a few different things:

### 1.
In:
```python 
ak.Array(nested_list)
```
Out:
```python 
<Array [[1, 2, 3], [0, 4]] type='2 * var * int64'>
```

### 2.
In:
```python 
x = ak.Array(np.array(nested_list[0], dtype=np.int8))
y = ak.Array(np.array(nested_list[1], dtype=np.int8))
ak.concatenate((x, y))
```
Out:
```python 
<Array [[1, 2, 3], [0, 4]] type='2 * var * int64'>
```

### 3.
In:
```python 
b = ak.ArrayBuilder()
for x in nested_list:
    b.append(np.array(x, dtype=np.int8))
```
Out:
```python 
<Array [[1, 2, 3], [0, 4]] type='2 * var * int64'>
```

As I see it, the ArrayBuilder would be the way to go, since I then don't have to create the nested list in Numba and then convert it to an awkward-array, however, looking at the [documentation](https://awkward-array.readthedocs.io/en/latest/_auto/ak.ArrayBuilder.html) of it I fail to see how one specify the dtype (other than int/float/bool/etc.).

[...]

PR #346, opened 2020-07-21
Operation to change the number type

ianna:
Addresses issue https://github.com/scikit-hep/awkward-1.0/issues/173


ianna:
@jpivarski - there is a temporary solution to silence the warnings in `NumpyArray::merge` where a compiler asks to add missing switch cases.

jpivarski:
Unhanded enumerations might have been intentional because they might be guaranteed by previous code. (`merge` makes several passes over the LHS and RHS types). However, I'm surprised that there are ~14 unhanded cases—I think there might be about 14 total cases. Something's odd about that message. (The enumeration is defined in include/awkward/util.h.)

If the enumeration is guaranteed by previous code (most likely `dtype::NOT_PRIMITIVE`), then the right exception for something that shouldn't happen is `std::runtime_error`. We use `std::invalid_argument` as feedback to the user that the user did something wrong, but if the default case happens in this switch, then it's our fault.

Incidentally, this enum is what has become of all those `format` checks. We were getting errors about corner cases, so it's now centralized and carried around as an enum.

ianna:
@jpivarski - here are some thoughts. Please, comment.

`awkward1.can_cast(fromtype, totype)` to lookup casting rules:
```python
assert awkward1.can_cast(dtype=numpy.float64, dtype=numpy.float32) == True or False  # allow double rounding?
assert awkward1.can_cast(dtype=numpy.float32, dtype=numpy.float64) == True
```

`awkward1.astype(array, totype)` is a method of an array. It takes the values of the array and produces a new array with the desire `dtype`:
```python
content_float64 = awkward1.layout.NumpyArray(numpy.array([1.1, 2.2, 3.3, 4.4, 5.5], dtype=numpy.float64))
array_float64 = awkward1.layout.UnmaskedArray(content_float64)
assert awkward1.to_list(array_float64) == [1.1, 2.2, 3.3, 4.4, 5.5]
assert str(awkward1.type(content_float64)) == "float64"
assert str(awkward1.type(awkward1.Array(content_float64))) == "5 * float64"
assert str(awkward1.type(array_float64)) == "?float64"
assert str(awkward1.type(awkward1.Array(array_float64))) == "5 * ?float64"

content_float32 = awkward1.astype(content_float64, dtype=numpy.float32)
array_float32 = awkward1.layout.UnmaskedArray(content_float32)
assert awkward1.to_list(array_float32) == [1.1, 2.2, 3.3, 4.4, 5.5]
assert str(awkward1.type(content_float32)) == "float32"
assert str(awkward1.type(awkward1.Array(content_float32))) == "5 * float32"
assert str(awkward1.type(array_float32)) == "?float32"
assert str(awkward1.type(awkward1.Array(array_float32))) == "5 * ?float32"

content_int8 = awkward1.astype(content_float32, dtype=numpy.int8)
array_int8 = awkward1.layout.UnmaskedArray(content_int8)
assert awkward1.to_list(array_int8) == [1, 2, 3, 4, 6]
assert str(awkward1.type(content_int8)) == "int8"
assert str(awkward1.type(awkward1.Array(content_int8))) == "5 * int8"
assert str(awkward1.type(array_int8)) == "?int8"
assert str(awkward1.type(awkward1.Array(array_int8))) == "5 * ?int8"
```

ianna:
> Unhanded enumerations might have been intentional because they might be guaranteed by previous code. (`merge` makes several passes over the LHS and RHS types). However, I'm surprised that there are ~14 unhanded cases—I think there might be about 14 total cases. Something's odd about that message. (The enumeration is defined in include/awkward/util.h.)
> 
> If the enumeration is guaranteed by previous code (most likely `dtype::NOT_PRIMITIVE`), then the right exception for something that shouldn't happen is `std::runtime_error`. We use `std::invalid_argument` as feedback to the user that the user did something wrong, but if the default case happens in this switch, then it's our fault.
> 
> Incidentally, this enum is what has become of all those `format` checks. We were getting errors about corner cases, so it's now centralized and carried around as an enum.

Thanks, @jpivarski ! I've corrected the error messages.

jpivarski:
> `awkward1.can_cast(fromtype, totype)` to lookup casting rules:
> 
> ```python
> assert awkward1.can_cast(dtype=numpy.float64, dtype=numpy.float32) == True or False  # allow double rounding?
> assert awkward1.can_cast(dtype=numpy.float32, dtype=numpy.float64) == True
> ```

I don't think a `can_cast` function will be necessary because the numeric conversions desired here would be truncating conversions. Someone may have float64 that they want to convert to int8. An example of a user needing a feature like this is issue #328: it's easy to make arrays containing int64, but not int8—having an `ak.to_numeric_type(array, dtype)` would allow them to create an array with int64 and convert it to int8 afterward. (The solution I gave them was for building the array with int8 in the first place, which involves more low-level routines.)

One common use would be after applying the NumPy ufunc `np.trunc`, you get floating point numbers without any decimals. Users would very likely want to then convert those floating point numbers into actual integers, perhaps to use the array as an integer-array slice.

So the function would be used in lossy contexts often.

I don't think it should take an `axis` parameter: a function like this can only apply to the leaves of a structure. The simplest thing is probably to make it apply to all leaves (NumpyArrays), with the exception of NumpyArrays that have parameter `__array__` == `"char"` or `"byte"`, since we wouldn't want strings to turn into floating point numbers or anything like that.

jpivarski:
As for a name, I think users wouldn't get anything involving the word "leaf." Maybe `ak.to_numeric_type` (there are a lot of `to_` functions in `operations/convert.py`), but it would help if the name specified that this is only changing the types of the numbers, not all other types. It should pass through records, lists, and option-type nodes without affecting them (i.e. nullable floats can be turned into nullable ints).

Maybe `ak.numbers_to_type(...)`?

This would be a good application of the new enums; the NumPy dtype would be examined in the pybind11 layer, an enum selected, and then that enum can pass down through the tree.

This:

https://github.com/scikit-hep/awkward-1.0/blob/1755384ab381074eaa595e4ef984f651e966e7eb/src/python/forms.cpp#L22-L42

is an example of converting a dtype into `kind` (a `char`) and `itemsize` (an `int`), which can then be converted into an enum with something like

https://github.com/scikit-hep/awkward-1.0/blob/1755384ab381074eaa595e4ef984f651e966e7eb/src/libawkward/Content.cpp#L33-L211

This numeric type converter shouldn't do anything with `inner_shape`. NumPy dtypes can have `shape`, but that should be an error if passed to the numeric type converter. (Without handling it explicitly, the `kind` of a dtype with `shape` is `"V"`, which is not a numeric `kind`.)

ianna:
> As for a name, I think users wouldn't get anything involving the word "leaf." Maybe `ak.to_numeric_type` (there are a lot of `to_` functions in `operations/convert.py`), but it would help if the name specified that this is only changing the types of the numbers, not all other types. It should pass through records, lists, and option-type nodes without affecting them (i.e. nullable floats can be turned into nullable ints).
> 
> Maybe `ak.numbers_to_type(...)`?
> 
> This would be a good application of the new enums; the NumPy dtype would be examined in the pybind11 layer, an enum selected, and then that enum can pass down through the tree.
> 
> This:
> 
> https://github.com/scikit-hep/awkward-1.0/blob/1755384ab381074eaa595e4ef984f651e966e7eb/src/python/forms.cpp#L22-L42
> 
> is an example of converting a dtype into `kind` (a `char`) and `itemsize` (an `int`), which can then be converted into an enum with something like
> 
> https://github.com/scikit-hep/awkward-1.0/blob/1755384ab381074eaa595e4ef984f651e966e7eb/src/libawkward/Content.cpp#L33-L211
> 
> This numeric type converter shouldn't do anything with `inner_shape`. NumPy dtypes can have `shape`, but that should be an error if passed to the numeric type converter. (Without handling it explicitly, the `kind` of a dtype with `shape` is `"V"`, which is not a numeric `kind`.)

I like `ak.to_numeric_type`

jpivarski:
> I like `ak.to_numeric_type`

I'm having second thoughts, wanting to avoid the impression that this changes the array's type, whatever it happens to be, into something numeric, when it actually changes the numeric types to other numeric types. Maybe only the word order needs to change:

```
ak.numeric_to_type(array, dtype)
```

or

```
ak.numbers_to_type(array, dtype)
```

? ("Numbers" would probably be easier to remember than "numeric".)

ianna:
@jpivarski - please, let me know if this is `awkward1.to_list` expected behaviour? Thanks

The array of `float32`:
```python
<NumpyArray format="f" shape="5" data="1.1 2.2 3.3 4.4 5.5" at="0x7f9d9446c6d0"/>

```
and its `awkward1.to_list`:
```python
>       assert awkward1.to_list(array_float32) == [1.1, 2.2, 3.3, 4.4, 5.5]
E       assert [1.100000023841858,\n 2.200000047683716,\n 3.299999952316284,\n 4.400000095367432,\n 5.5] == [1.1, 2.2, 3.3, 4.4, 5.5]
E         At index 0 diff: 1.100000023841858 != 1.1
E         Full diff:
E           [
E         -  1.100000023841858,
E         -  2.200000047683716,
E         -  3.299999952316284,
E         -  4.400000095367432,
E         +  1.1,
E         +  2.2,
E         +  3.3,
E         +  4.4,
E            5.5,
E           ]

```

jpivarski:
It's because the float32 is losing precision. The Python `float` type is the IEEE 64-bit binary floating-point: i.e. Python's `float` is C's `double`. When numbers are written in the source code, they're parsed as doubles and compared as doubles, but if the values went through a 32-bit floating point format (in NumPy or Awkward), then there will be round-off error.

pytest has a "nearly equal to" comparison, but I think it makes more sense here to just  put a literal `1.100000023841858, 2.200000047683716, 3.299999952316284, 4.400000095367432` in the test code (or make the test points factors of 2, like `0.5`, `0.25`).

ianna:
@jpivarski - I'm not sure what should I do about these failing doctests. This PR introduces macros. Could this be a problem?
```bash
writing _auto/ak._connect._numba.builder.lower_integer.rstTraceback (most recent call last):
  File "/home/vsts/work/1/s/docs-sphinx/../dev/genpython.py", line 710, in <module>
    ast = pycparser.c_parser.CParser().parse(pfile)
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pycparser/c_parser.py", line 149, in parse
    return self.cparser.parse(
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pycparser/ply/yacc.py", line 331, in parse
    return self.parseopt_notrack(input, lexer, debug, tracking, tokenfunc)
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pycparser/ply/yacc.py", line 1199, in parseopt_notrack
    tok = call_errorfunc(self.errorfunc, errtoken, self)
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pycparser/ply/yacc.py", line 193, in call_errorfunc
    r = errorfunc(token)
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pycparser/c_parser.py", line 1858, in p_error
    self._parse_error(
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/pycparser/plyparser.py", line 67, in _parse_error
    raise ParseError("%s: %s" % (coord, msg))
pycparser.plyparser.ParseError: :1168:28: before: #

Configuration error:
There is a programmable error in your configuration file:

  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/sphinx/config.py", line 319, in eval_config_file
    execfile_(filename, namespace)
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/site-packages/sphinx/util/pycompat.py", line 88, in execfile_
    exec(code, _globals)
  File "/home/vsts/work/1/s/docs-sphinx/conf.py", line 71, in <module>
    subprocess.check_call(
  File "/opt/hostedtoolcache/Python/3.8.5/x64/lib/python3.8/subprocess.py", line 364, in check_call
subprocess.CalledProcessError: Command '['/opt/hostedtoolcache/Python/3.8.5/x64/bin/python', '/home/vsts/work/1/s/docs-sphinx/../dev/genpython.py', '/home/vsts/work/1/s/docs-sphinx/../src/cpu-kernels/identities.cpp', '/home/vsts/work/1/s/docs-sphinx/../src/cpu-kernels/operations.cpp', '/home/vsts/work/1/s/docs-sphinx/../src/cpu-kernels/reducers.cpp', '/home/vsts/work/1/s/docs-sphinx/../src/cpu-kernels/getitem.cpp', '/home/vsts/work/1/s/docs-sphinx/../src/cpu-kernels/sorting.cpp']' returned non-zero exit status 1.
```

jpivarski:
@reikdas Do you know that's happening here?

reikdas:
The C++ -> Python transpiler cannot handle macros yet. I'll try and fix it.

jpivarski:
@ianna and @reikdas I see what's happening here: the NumpyArray fill functions are very regular—I wrote them with an Emacs macro and Yana is writing them with a C macro, which is more DRY from a source code point of view, but Reik's parser needs to be able to parse this source code.

Extending Reik's parser to handle macros is likely too difficult, at least too difficult on the timescale we have for the rest of the summer. On the other hand, if we can invoke C's preparser manually to expand those macros into source code that Reik's parser can handle, that could be a good solution.

It's a little ironic that we have templated the function, then need to make a concrete instance of each specialization, so we use a C macro, but a C macro is a bit like a templated function anyway. In principle, we could have implemented the whole thing in a macro, rather than using the macro to instantiate the template! But we shouldn't do that because then Reik's parser wouldn't be able to tell that these are all specializations of a single function (that's important information, which he gets from the fact that they come from the same template).

The two courses of action that I see as possibilities are

   1. find a way to invoke the C preparser manually, to expand the macro into named functions that call template instantiations, for Reik's parser to pick up;
   2. expand out the WET code in the source file, as it had been before.

Option (2) requires almost no work; I'm sure Yana made the macro because, in usual circumstances, WET code is bad. In this circumstance, we have specific reasons why it's easier for the code to be redundantly written out in the source file like this.

If option (1) is easy—Yana, if you know what command needs to be invoked to run the C preparser with the output being a temporary source file for Reik's parser—then that could be a solution to both issues. On the other-other hand, having C macros expand template instantiations might be adding complexity just because it's more indirection—sometimes a flat, repetitive file is the most maintainable.

reikdas:
> Extending Reik's parser to handle macros is likely too difficult, at least too difficult on the timescale we have for the rest of the summer.

I was actually wondering how to handle this and asked you on Slack :p 

> find a way to invoke the C preparser manually, to expand the macro into named functions that call template instantiations, for Reik's parser to pick up;

I think this is possible - [pycparser even mentions it](https://github.com/eliben/pycparser#interaction-with-the-c-preprocessor). But since we are actually handling C++ code and not C, I am not sure if this will be difficult.

jpivarski:
> I think this is possible - [pycparser even mentions it](https://github.com/eliben/pycparser#interaction-with-the-c-preprocessor). But since we are actually handling C++ code and not C, I am not sure if this will be difficult.

That's right: you need to parse the templates before handing it to pycparser (which would then run the macros). Because of the existence of non-C constructs, we can't have the macros handled before Reik's parser.

In that case, @ianna, it would be better to remove this Don't-Repeat-Yourself macro in favor of Write-Every-Time. The fact that these functions are more like data than code is intentional: it's to help automatic conversion to CUDA. These NumpyArray-filling functions, in particular, are embarrassingly parallel and we especially want Reik's parser to pick them up and convert them. The fact that we're seeing the error in documentation-building is just because that's the _first_ thing Reik's parser is used for.

If you know how to use the C++ preparser to expand the macro instances, then it should be easy to copy-paste its output into the source code. I generated them with Emacs macros, which were a little more in the workflow of editing...

ianna:
> If you know how to use the C++ preparser to expand the macro instances, then it should be easy to copy-paste its output into the source code. I generated them with Emacs macros, which were a little more in the workflow of editing...

```bash
clang file.cc -E
``` 
The same for any other compiler: `gcc` or `g++`

In fact, there is no problem to expand them. I’ll do it tomorrow morning.

ianna:
@jpivarski and @reikdas - C++ preprocessor expands a macro in one line: https://github.com/scikit-hep/awkward-1.0/pull/346/commits/4ae8e44ca87c8262dd013c959c0d76b739a9d649
So, I've added a clang-format configuration and reformatted the code close to an original style. Now the kernel script fails:
```bash
Starting: Generate Kernel specification
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.164.2
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
Script contents:
python dev/generate-kernelspec.py
========================== Starting Command Output ===========================
/bin/bash --noprofile --norc /Users/runner/work/_temp/31384ca2-88ca-4b7b-b51c-8dbeff0cad96.sh
  File "dev/generate-kernelspec.py", line 1073, in <module>
    pfile, tokens = preprocess(filename)
  File "dev/generate-kernelspec.py", line 331, in preprocess
    parans.pop()
IndexError: pop from empty list

##[error]Bash exited with code '1'.
Finishing: Generate Kernel specification
```

reikdas:
I am not exactly sure about the reason for the issue yet, but it is possibly because I made some assumption about the file structure in my C++ -> C translator in `dev/generate-kernelspec.py` which is not always true in C/C++. I'll take a look and try to fix it. Sorry for the inconvenience. 

One other thing I should point out is that in the latest commit to `master`, we introduced a new coding style requirement for the cpu-kernel include files - https://github.com/scikit-hep/awkward-1.0/blob/master/CONTRIBUTING.md#contributing-to-files-in-includeawkwardcpu-kernels. It would mostly involve copy-pasting some doxygen style comment strings above each function definition (here is the complete set of roles - https://github.com/scikit-hep/awkward-1.0/blob/master/kernel-specification/samples.json). Not having this _will_ cause the specification generator to break, after the issue you currently encountered is fixed. I can add those in while fixing the C++ -> C translator if you want. 

An unrelated comment - I was looking at the file changes in this PR, and noticed that clang-format causes a huge diff. Maybe we should use something like [git clang-format](https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/git-clang-format) so that only the changes in the PR have formatting applied, to preserve git blame (and have a cleaner PR diff). The disadvantage of doing that would be that the file in question would not be uniformly formatted, so maybe there could be a different PR where we clang-format the entire codebase and require that all new code must be formatted according to our `clang-format` config file? What do you and @jpivarski think?

jpivarski:
At the very least, the clang-format change should be in its own PR. git blame is already shadowed by the PR I did that bound all lines by 80 characters—the lesson to learn from that is that formatting should be applied from day 1.

If the formatting is an improvement, then let's do it and be consistent from here on out. It might shadow real changes in git blame, which will be unfortunate, but if that has to happen anyway, it's better for it to happen earlier than later. (git blame is especially useful when a codebase has stabilized to mostly small, focused changes, which Awkward is getting pretty close to, with some exceptions.)

What formatting options are being requested in clang-format? This question should probably be answered in a PR dedicated to formatting, not here, but we might make the impact of the formatting on history lighter by picking only the most useful formatting options.

ianna:
@jpivarski - the clang-format rules are in `.clang-format` YAML file as a part of this PR. I added the following options:

* Should be used for C, C++.
* Based on Google style: https://google.github.io/styleguide/cppguide.html
* The column limit is 80
* The indentation used for all namespaces:
```c++
namespace out {
  int i;
  namespace in {
    int i;
  }
}
```
* Do not sort `#includes`
* The number of columns to use for indentation is 2
* The extra indent or outdent of access modifiers is -2, e.g. `public:`
* The penalty for each line break introduced inside a comment is 30
* The penalty for each character outside of the column limit: 100
* Horizontally align arguments after an open bracket
* Aligns trailing comments
```c++
int a;     // My comment a      vs.     int a; // My comment a
int b = 2; // comment  b                int b = 2; // comment about b
```
* AllowAllArgumentsOnNextLine: true
If a function call or braced initializer list doesn’t fit on a line, allow putting all arguments onto the next line:
```c++
true:
callFunction(
    a, b, c, d);

false:
callFunction(a,
             b,
             c,
             d);
```
* AllowAllParametersOfDeclarationOnNextLine: true
If the function declaration doesn’t fit on a line, allow putting all parameters of a function declaration onto the next line
```c++
true:
void myFunction(
    int a, int b, int c, int d, int e);

false:
void myFunction(int a,
                int b,
                int c,
                int d,
                int e);
```
* AllowShortBlocksOnASingleLine: false
Never merge blocks into a single line
* AllowShortIfStatementsOnASingleLine: Never
Never put short ifs on the same line
* AllowShortLambdasOnASingleLine: false
Never merge lambdas into a single line
* AllowShortLoopsOnASingleLine: false
Never put short loops on a single line
* AlwaysBreakAfterReturnType: All
Always break after the return type
* BinPackParameters: false
a function declaration’s or function definition’s parameters will either all be on the same line or will have one line each
* AlwaysBreakTemplateDeclarations: Yes
Always break after template declaration
* ReflowComments: false
Do not re-flow comments
* BinPackArguments: false
a function call’s arguments will either be all on the same line or will have one line each

Here is a link to full documentation:
https://clang.llvm.org/docs/ClangFormatStyleOptions.html

ianna:
> I am not exactly sure about the reason for the issue yet, but it is possibly because I made some assumption about the file structure in my C++ -> C translator in `dev/generate-kernelspec.py` which is not always true in C/C++. I'll take a look and try to fix it. Sorry for the inconvenience.
> 
No problem :-)
The work you've done is impressive! If it's formatting, I can always revert the last two commits and re-format only the auto-generated code. 
> One other thing I should point out is that in the latest commit to `master`, we introduced a new coding style requirement for the cpu-kernel include files - https://github.com/scikit-hep/awkward-1.0/blob/master/CONTRIBUTING.md#contributing-to-files-in-includeawkwardcpu-kernels. It would mostly involve copy-pasting some doxygen style comment strings above each function definition (here is the complete set of roles - https://github.com/scikit-hep/awkward-1.0/blob/master/kernel-specification/samples.json). Not having this _will_ cause the specification generator to break, after the issue you currently encountered is fixed. I can add those in while fixing the C++ -> C translator if you want.
> 
Thanks, I'll add these.
> An unrelated comment - I was looking at the file changes in this PR, and noticed that clang-format causes a huge diff. Maybe we should use something like [git clang-format](https://github.com/llvm/llvm-project/blob/master/clang/tools/clang-format/git-clang-format) so that only the changes in the PR have formatting applied, to preserve git blame (and have a cleaner PR diff). The disadvantage of doing that would be that the file in question would not be uniformly formatted, so maybe there could be a different PR where we clang-format the entire codebase and require that all new code must be formatted according to our `clang-format` config file? What do you and @jpivarski think?

Either is fine with me as long as I do not have to do it by hand :-)

reikdas:
#369 should fix the broken CI :)

ianna:
@jpivarski - I think, I'm done with this PR. Please, review it when you have time. Thanks.

jpivarski (approved):
This looks like it solves the problem, adding a low-level `numbers_to_type` (high-level would have to be added later) and a high-level `can_cast` implementing the NumPy function of the same name.

Some of the files have been whitespace-formatted, such as operations.cpp. That's going to make this hard to merge with #366, so I'm trying to figure out which order to do things in.

I think it might be easiest if I finish #366, merge it into master, stash operations.h and operations.cpp in this branch, merge in from master, and then re-introduce the new number-type converting operations, since they're formulaic and can be generated. I think that's what I'm going to do: I approve this PR but I'll have to merge it carefully.

Meanwhile, a few things that are disjoint from that:

1. I don't think the `numbers_to_type` method should do a `deep_copy` on the Identities. I don't think that's what most other methods do. That can be changed independently of the merging that I'll need to do (which mostly affects kernels, like operations.cpp).
2. `numbers_to_type` could have a high-level function in src/awkward1/operations/structure.py.
3. Most of these tests are for `can_cast`; should a few more of the node types be tested? They don't have to be hard-to-construct ones like IndexedArray; they can be easy-to-construct ones made with the `ak.Array` constructor.

I'll let you know when I'm going to try to merge #366 + master into this PR, taking care with operations.h and operations.cpp, in case you're working on any of the 3 above, so that we can coordinate. If, instead, you want to leave those things for a later PR and work on `concatenate` with `axis != 0`, that would be entirely disjoint and easy to merge.

ianna:
Thanks, @jpivarski ! I'll leave those things for a later PR and work on `concatenate`. Thanks.

PR #347, opened 2020-07-21
Remove redundant line from reducers kernel

reikdas:


jpivarski (approved):
Yeah, it looks like this `count` is never used. (It was probably a step in development.) Thanks!

PR #349, opened 2020-07-23
Fix Python SyntaxWarning

reikdas:


jpivarski (approved):
Yikes! That's right—we don't want to compare strings by reference.

ianna:
Thanks @reikdas !

PR #354, opened 2020-07-23
Mixin class decorators

### LINKED ISSUE #355: Decorators for making new mixins
This is a tracking issue for migrating https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/methods/mixin.py into awkward. The decorators allow a more concise declaration of mixin classes, e.g.:
```python
from awkward1 import mixin_class, mixin_class_method


@mixin_class(ak.behavior)
class Point:
    def distance(self, other):
        return np.sqrt((self.x - other.x) ** 2 + (self.y - other.y) ** 2)

    @mixin_class_method(np.equal, {"Point"})
    def point_equal(self, other):
        return np.logical_and(self.x == other.x, self.y == other.y)

    @mixin_class_method(np.abs)
    def point_abs(self):
        return np.sqrt(self.x ** 2 + self.y ** 2)
```
reproduces the functionality outlined in the `ak.behavior` introductory documents.

nsmith-:
Provide some convenience functions for making new behavior classes.

### DIFF
--- a/docs-sphinx/ak.behavior.rst
+++ b/docs-sphinx/ak.behavior.rst
@@ -315,7 +315,7 @@ Mixin decorators
 The pattern of adding additional properties and function overrides to records
 and arrays of records is quite common, and can be nicely described by the "mixin"
 idiom: a class with no constructor that is mixed with both the :doc:`_auto/ak.Array` and :doc:`_auto/ak.Record`
-class as to create new derived classes. The :doc:`_auto/ak.mixin_class` and :doc:`_auto/ak.mixin_class_method`
+class as to create new derived classes. The :doc:`_auto/ak.behaviors.mixins.mixin_class` and :doc:`_auto/ak.behaviors.mixins.mixin_class_method`
 python decorators assist with some of this boilerplate. Consider the ``Point`` class
 from above; we can implement all the functionality so far described as follows:
 
@@ -360,7 +360,7 @@ such a class as follows:
 .. code-block:: python
 
     @mixin_class(ak.behavior)
-    class WeightedPoint:
+    class WeightedPoint(Point):
         @mixin_class_method(np.equal, {"WeightedPoint"})
         def weighted_equal(self, other):
             return np.logical_and(self.point_equal(other), self.weight == other.weight)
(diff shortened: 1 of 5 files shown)


jpivarski (commented):
This is really nice! But can the src/awkward1/util.py file be renamed src/awkward1/behaviors/mixins.py? I'm in favor of importing it into the global namespace as `ak.mixin_class`, `ak.mixin_class_method`, etc. Everything related to behaviors should go in that submodule, though.

jpivarski:
Actually, there should probably also be tests for something like this. You have great examples in the documentation—just make them tests. Thanks!

jpivarski:
If this is related to a corresponding change in Coffea (you mentioned "migrating"), then I could deploy it with a new minor version number, to make the dependency easier to specify.

jpivarski (approved):
Great! I'd consider this done.

Are you planning any additional updates or can I merge it? Thanks!

docs-sphinx/ak.behavior.rst
```diff
@@ -309,6 +310,74 @@ Similarly for overriding ``abs``
...
+================
+The pattern of adding additional properties and function overrides to records
+and arrays of records is quite common, and can be nicely described by the "mixin"
+idiom: a class with no constructor that is mixed with both the :doc:`_auto/ak.Array` and :doc:`_auto/ak.Record`
+class as to create new derived classes. The :doc:`_auto/ak.behaviors.mixins.mixin_class` and :doc:`_auto/ak.behaviors.mixins.mixin_class_method`
+python decorators assist with some of this boilerplate. Consider the ``Point`` class
+from above; we can implement all the functionality so far described as follows:
+
+.. code-block:: python
+    @ak.mixin_class(ak.behavior)
```
nsmith-, line 323: I think the missing space here makes this not render in the sphinx docs

nsmith-: @jpivarski

jpivarski: That's true. (One of the quirks of ReST.) I didn't check the look of the documentation before merging it, but I can put in a correction. (They're updated with each push to master, not each new version deployed.)

jpivarski: That's true. (One of the quirks of ReST.) I didn't check the look of the documentation before merging it, but I can put in a correction. (They're updated with each push to master, not each new version deployed.)

jpivarski: Fixed in #360.

PR #356, opened 2020-07-24
Adapt to pyarrow/Arrow 1.0.

jpivarski:


jpivarski:
Arrow 1.0 does not implement any more of the Parquet types in [tests/test_0341-parquet-reader-writer.py::test_oamap_samples](https://github.com/scikit-hep/awkward-1.0/blob/816d203495c90db5a25cf7bbb64da72e2eb7b14c/tests/test_0341-parquet-reader-writer.py#L236-L246).

PR #357, opened 2020-07-25
Cleanup the Docker Residues like .test files and .dockerignore.

trickarcher:


trickarcher:
I have cleaned all the `none` tags from the AWS instance and I built the latest `awkward-cuda-test` images from this particular branch, which has all the residues from the other PR cleaned up. I'll build them once again(from master this time), once this PR gets merged.

.ci/launch-cuda-docker-ci.sh
```diff
@@ -1,10 +1,10 @@
...
-docker run -it --gpus all awkward1-cuda-tests:1.0-cuda10.2 sh .ci/run-cuda-tests.sh trickarcher/python_cuda_interface
+docker run -it --gpus all awkward1-cuda-tests:1.0-cuda10.2 sh .ci/run-cuda-tests.sh master
 
 docker build -f ../dev/docker/docker_cuda100/Dockerfile.build -t awkward1-cuda-tests:1.0-cuda10.0 ../.
-docker run -it --gpus all awkward1-cuda-tests:1.0-cuda10.0 sh .ci/run-cuda-tests.sh trickarcher/python_cuda_interface
+docker run -it --gpus all awkward1-cuda-tests:1.0-cuda10.0 sh .ci/run-cuda-tests.sh master
 
 docker build -f ../dev/docker/docker_cuda900/Dockerfile.build -t awkward1-cuda-tests:1.0-cuda9.0 ../.
-docker run -it --gpus all awkward1-cuda-tests:1.0-cuda9.0 sh .ci/run-cuda-tests.sh trickarcher/python_cuda_interface
+docker run -it --gpus all awkward1-cuda-tests:1.0-cuda9.0 sh .ci/run-cuda-tests.sh master
```
jpivarski: Shouldn't the branch to test be an argument to this script, like `$1`?

Also, how is it going to run all three if `docker run` is passed an "interactive" flag (`-i`)? It would be waiting for me to close the first image before it starts the second. It's true that we'd want to manually enter the image interactively to debug failed tests, but the normal mode is to just run them all and see that they were successful.

Can the output be written to a file? We'd like to be able to just start a run and go get a coffee, checking the results when we get back. An error shouldn't be hidden by a scrolling terminal. This script could be given a `set -e` to stop after the first error if `docker run` could be made to return an error if the script inside the image fails. That is, we'd like to bubble up any errors encountered so that we immediately see it when we get back with coffee (a low-tech version of CI's green or red indicator, but we don't need a GUI: an error message would suffice).

trickarcher: The `-i` flag is for applications that require interactivity like `emacs`. The `pytest` and `git pull` don't require any user input so the script passes normally. I'll be honest here, I passed the `-i` flag to get a colored  green PASSED checks while running `pytest`. But I'll remove it if that doesn't serve any purpose. 
I think I can dump all the pytest error log in a file using the flag `--log`. Is it necessary to include git logs too?

trickarcher: And these logs, will go under `~/awkward1-cuda-tests/logs`? Because we wouldn't want to clutter the repository itself.

jpivarski: It could even go to standard out—we only need enough formality for our own convenience. It should just be clear when something fails, so that we can interactively follow up.

Probably the most convenient place for a lot would be in the same directory where the script is invoked. Then if something goes wrong, it will be clear where to look for more information. It can contain all the detail—we can scroll (or more likely, page through with "less", which is also good for incremental searching).

If the `-i` flag doesn't prevent it from continuing through all three `docker run` commands, then it can be left in. The pytest colors don't always work for me—as soon as there's any red, they all become red, making it hard to pick out the ones that failed. Also, I don't think these colors would show up in the log (it might even make the log hard to read, if the VT color commands are shown as text).

trickarcher: Yup, I have made all the necessary changes. You can now log the output using `./launch-cuda-docker-ci.sh -l -b <branch-name>` and have `./launch-cuda-docker-ci.sh` for normal terminal output

jpivarski: Well, I wasn't able to run it. There are some wrong assumptions in the Docker image or script: it says things like

```
mkdir: cannot create directory ‘logs/2020-07-27_15-29’: No such file or directory
unable to prepare context: unable to evaluate symlinks in Dockerfile path: lstat /home/ubuntu/dev: no such file or directory
```

so evidently it's in the wrong directory. It will be important to get this into a push-button state, so that a set of instructions always does the right thing, since I'm giving up my GPU-enabled computer for a week and this will be the only way I have to run tests. I can log in with my other laptop, so this method will work once the issues of "Which working directory are we in?" get settled.

trickarcher: That's bad. Thanks for looking into this. I investigated the matter and I think it's due to relative paths. Did you launch the script using `./.ci/launch-cuda-docker-ci.sh`, just to confirm if that's the actual issue?

jpivarski: Yes, that's how I launched it, from within the ~/awkward1 directory. 

We're both logging into the same computer, so it's not a "my computer is configured differently from yours in some subtle way" issue. If you set up explicit instructions, starting from the first ssh into the AWS GPU instance, and they work for you (repeatedly; make sure that running it once doesn't put it into a state where it can't be run again), then it will work for me. Those instructions should go into the CONTRIBUTING.md, since it's something that all developers entrusted with a AWS GPU key would be doing regularly.

trickarcher: I have added the necessary documentation. Also, since I have included absolute paths now(Which should be fine since these scripts are only meant to be used with the AWS instance), we can launch the scripts from any directory and it will work.

jpivarski: It works! I updated CONTRIBUTING.md because I thought it wasn't clear (and I think we'll _always_ want to use the `-l` flag), so I did that myself. Otherwise, I think everything is done here and I can merge this PR. Is that right?

trickarcher: Yeah, you can merge it!

jpivarski: Great! I'll just wait for the tests to pass (even though they have to: CONTRIBUTING.md can't change the result) for formality.

PR #361, opened 2020-07-28
concatenate for a nonzero axis operation

ianna:
address Issue https://github.com/scikit-hep/awkward-1.0/issues/184


jpivarski:
Hi @ianna, any progress on this? (I'm asking because it's been 20 days; maybe you haven't pushed local changes.) Thanks!

nsmith-:
It would be cool to see this implemented, an axis=1 concatenate would be super slick for some analysis work

src/libawkward/array/ListArray.cpp
```diff
@@ -963,6 +968,9 @@ namespace awkward {
         ;
       }
       else {
+        // FIXME:???
+        return mergemany_as_union(others, 0, 0);
+
```
ianna: here is why:
```python
    def test_indexed_array_merge():
        one = awkward1.Array([[1, 2, 3], [None, 4], None, [None, 5]]).layout
        two = awkward1.Array([6, 7, 8]).layout
        three = awkward1.Array([[6.6], [7.7, 8.8]]).layout
    
>       assert awkward1.to_list(one.merge(two, 0)) == [[1, 2, 3], [None, 4], None, [None, 5], 6, 7, 8]
E       ValueError: cannot merge ListArray64 with NumpyArray
E       
E       (https://github.com/scikit-hep/awkward-1.0/blob/0.4.0/src/libawkward/array/ListArray.cpp#L976)

tests/test_0184-concatenate-operation.py:116: ValueError
```

jpivarski: It's true that a ListArray64 can't be merged with a NumpyArray. The `merge` and `mergemany` functions are low-level and can only be used after `mergeable` certifies that the two arrays can be merged. If `mergeable` returns false, then `merge_as_union` must be used instead. The fact that the methods have these unwritten rules is why they're low-level: in Python, one should use `ak.concatenate`, and even most of our internal use of these functions are mediated through a `simplify` operation. I don't see any reason why this concatenate with `axis > 0` feature couldn't be implemented entirely by making a UnionArray and calling `simplify` on it.

ianna: @jpivarski - thanks! done

ianna:
@jpivarski - yes, with `broadcast_and_apply` the arrays must be equal length. What's the best approach here? Padding them with `None`s and then dropping the `None`s after the `concatenate` operation works.
```
[[0.0, 1.1, 2.2], [], [3.3, 4.4], [5.5], [6.6, 7.7, 8.8, 9.9], None, None]
[[1.1, 2.2], [3.3], [], [4.4, 5.5], [6.6, 7.7, 8.8], [], [9.9]]
```

jpivarski:
Actually, users should not be allowed to concatenate arrays with `axis > 0` if their lengths differ. After all, if we want to concatenate

```
[[1, 2, 3], [], [4, 5]]   and   [["a", "b"], ["c"], ["d", "e", "f"]]
```

at `axis=1`, that means we want

```
[[1, 2, 3, "a", "b"], ["c"], [4, 5, "d", "e", "f"]]
```

This operation wouldn't make any sense if the two arrays we want to concatenate have different lengths.

Concatenation at `axis=0` is qualitatively different: the input arrays can have any lengths at all. But for all other `axis` values, the lengths have to be the same. For `axis > 1`, the lengths have to be the same at several depths, but `broadcast_and_apply` ensures that as well, either through error messages or broadcasting, if possible.

ianna:
```
tests/test_0397-arrays-as-constants-in-numba.py::test_refcount FAILED
```
https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=4560&view=logs&j=81547bd0-3aa9-547f-25e1-12509e21f92a&t=6f2c7081-a276-5c13-1bbf-f0639c0526de&l=509

ianna:
@jpivarski - please, have a look when you can. Thanks!

jpivarski:
> ```
> tests/test_0397-arrays-as-constants-in-numba.py::test_refcount FAILED
> ```
> 
> https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=4560&view=logs&j=81547bd0-3aa9-547f-25e1-12509e21f92a&t=6f2c7081-a276-5c13-1bbf-f0639c0526de&l=509

That one failure is disturbing, since it seems to have nothing to do with your changes. I'm rerunning it and running another test on master, though it's not necessarily a good sign if those pass. (Even worse than an inexplicable error is an intermittent inexplicable error.)

src/awkward1/operations/structure.py
```diff
@@ -744,20 +740,49 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
...
-        if batch[-1].mergeable(x, mergebool=mergebool):
-            batch.append(x)
-        else:
-            collapsed = batch[0].mergemany(batch[1:])
-            batch = [collapsed.merge_as_union(x)]
+    if axis < 0:
+        raise ValueError(
+            "the 'axis' of concatenate must be non-negative"
+            + awkward1._util.exception_suffix(__file__)
+        )
```
jpivarski: Is this necessary? Your implementation uses `axis_wrap_if_negative`.

src/awkward1/operations/structure.py
```diff
@@ -744,20 +740,49 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
...
+            )
+
+        def getfunction(inputs, depth):
+            if depth == axis:
+                arr = inputs
+                batch = [arr[0]]
+                for x in arr[1:]:
+                    collapsed = batch[0].mergemany_as_union(batch[1:], 1)
+                    batch = [collapsed.merge_as_union(x, 1)]
+                out = batch[0].mergemany_as_union(batch[1:], 1)
```
jpivarski: I don't understand this: why is `mergemany_as_union` and `merge_as_union` being called once per array? Why shouldn't this implementation just be the following?

```python
out = inputs[0].mergemany_as_union(inputs[1:], 1)
```

src/awkward1/operations/structure.py
```diff
@@ -744,20 +740,49 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
...
+                return lambda: (
+                    out,
+                )
```
jpivarski: Just

```python
return lambda: out
```

is fine. This looks like something that an auto-formatter like Black went nuts on.

kernel-specification.yml
```diff
@@ -2696,6 +2696,51 @@ kernels:
...
+          for i in range(length - 1):
+            if i < leftlen:
+              a =  fromleft[i + 1] - fromleft[i]
+            else:
+              a = 0
+            if i < rightlen:
+              b = fromright[i + 1] - fromright[i]
+            else:
+              b = 0
+            tooffsets[i + 1] = tooffsets[i] + a + b
```
jpivarski: Thanks for filling in the normative definitions! Just a note, though: we use 4-space indent in Python. (I used to use 2-space indent in Python, too, since it's a syntactical element, it needs to be more visible and obvious. Also, Black and various style guides recommend 4 spaces. In C++, 2 spaces is fine because we always use curly brackets, even for one-line bodies.)

kernel-specification.yml
```diff
@@ -4689,6 +4734,54 @@ kernels:
...
+                  totags[j] = tag
+                  j = j + 1
+              if tag == 0:
+                if iright < rightlen:
+                  tag = 1
+                ileft = ileft + 1
+              else:
+                if ileft < leftlen:
+                  tag = 0
+                iright = iright + 1
```
jpivarski: Same thing: 4-space indent. (Your first indentation is 6 spaces.)

src/awkward1/operations/structure.py
jpivarski: Oh! Is it because you check

```python
if depth == axis
```

in `getfunction`? Maybe you could expose the `axis_wrap_if_negative` to Python, since this needs to be checked while descending, not when you've already reached the target axis. That could enable the feature for `axis < 0`.

jpivarski (approved):
This is a good suite of tests, particularly the fact that it agrees well with multidimensional NumPy arrays. So it appears to be fully working!

I wonder, though, about the `getfunction` implementation in Python. It seems to be calling `mergemany_as_union` and `merge_as_union` a lot more times than I'd think are necessary. Doesn't each `mergemany_as_union` merge all arrays in `inputs`? Why does that function need to be called for each array in `inputs`? This calculation ought to be _O(n)_ in the number of arrays _n_, but it looks like it's _O(n²)_.

There might also be low-hanging fruit in getting this to work for `axis < 0`; see inline comment.

src/awkward1/operations/structure.py
ianna: done

src/awkward1/operations/structure.py
ianna: I think, it expects a tuple as an output:
```python
awkward1/operations/structure.py:778: in concatenate
    out = awkward1._util.broadcast_and_apply(arrays, getfunction, behavior)[0]
awkward1/_util.py:975: in broadcast_and_apply
...
>               assert isinstance(outcontent, tuple)
E               AssertionError

awkward1/_util.py:746: AssertionError

```

kernel-specification.yml
ianna: done

kernel-specification.yml
ianna: done

jpivarski:
> > ```
> > tests/test_0397-arrays-as-constants-in-numba.py::test_refcount FAILED
> > ```
> > 
> > 
> > https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=4560&view=logs&j=81547bd0-3aa9-547f-25e1-12509e21f92a&t=6f2c7081-a276-5c13-1bbf-f0639c0526de&l=509
> 
> That one failure is disturbing, since it seems to have nothing to do with your changes. I'm rerunning it and running another test on master, though it's not necessarily a good sign if those pass. (Even worse than an inexplicable error is an intermittent inexplicable error.)

A run on the master branch produced no errors: https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=4573&view=results while retrying this branch reproduced the error: https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=4571&view=logs&j=f72a8481-bd91-5d06-6025-0e453ddcac08&t=0b1b15ad-cf24-5193-0114-2bb1bf897da3, which makes me feel a little better because it's not intermittent. But it also means that somewhere in this PR, there's something that affects reference counting in Numba on Windows. (!?!) I'll pore through the diffs to see if anything looks suspicious.

Edit: I didn't see anything in your branch that could have caused that. The specific error is that Numba takes references to Python objects when they're used in a closure. Normal Python functions like

```python
def f2():
    array, array
    return 3.14
```

don't take a reference to `array` because the closure does a fresh lookup in the surrounding scope (i.e. if you were to change `array` after having defined this function, the effect of the function would change). Numba functions, like

```python
@numba.jit
def f2():
    array, array
    return 3.14
```

take references to `array` (2 of them) for each compilation of the function (i.e. whenever you call the function with different arguments, triggering a compilation). In this case, `array` is truly a constant: changing it in the surrounding scope doesn't change the function.

The test verifies that Numba is taking these references, increasing the Python reference count for `array`, but not dropping them when `f2` is removed. Strangely, the test shows that dropping `f1` removes the reference but dropping `f2` does not remove its two references. Even more strangely, this is only happening in Numba in Python 3.8 in Windows 32-bit.

src/awkward1/operations/structure.py
jpivarski: You're right! It should be

```python
return lambda: (out,)
```

Python's tuple syntax can be subtle (because trailing commas are ignored in argument lists). But anyway, it doesn't need to be three lines.

ianna: done

jpivarski:
Is this branch up-to-date with master? Maybe something old is causing the Windows 32-bit Numba error.

I've run tests on master twice with no such error, but it seems like it appears in every test on this branch. Even if its probability is 1/2, the always-on-this-branch and never-on-master is somewhat improbable.

jpivarski:
#495 does some cross-checks on your kernel-specifications.yml, and that has been merged to master. If you update again, you'll get this cross-checker to see if your newly added kernels match specifications.

src/awkward1/operations/structure.py
jpivarski: I lost a huge comment here...

Looking at the original, I can see what it was doing:

```python
    batch = [contents[0]]
    for x in contents[1:]:
        if batch[-1].mergeable(x, mergebool=mergebool):
            batch.append(x)
        else:
            collapsed = batch[0].mergemany(batch[1:])
            batch = [collapsed.merge_as_union(x)]

    out = batch[0].mergemany(batch[1:])
```

It was accumulating batches of mutually `mergeable` arrays into a Python list (which does not scale with the size of the data) and then either calling `mergemany` on them all once at the end or once as soon as a non-mergeable array was encountered. Thus, the algorithm was _O(n)_ in the total length of the array: nothing was double-merged.

I haven't checked your implementation to see if it has the same characteristics. We'd want it to be able to batch mutually `mergeable` arrays and only `mergemany` them once (but at a deeper `axis`) and we'd want it to not attempt to `mergemany` on non-mergeable arrays. Keep in mind that the `mergemany` method raises RuntimeErrors (i.e. our internal errors) if you try to merge anything that can't be merged. If we don't compare explicitly verify that two arrays are `mergeable` before trying to merge them, it's a bug.

In the tests, can you verify that all combinations of the following arrays containing numbers and records (at `axis=1`) concatenate correctly?

```python
numbers = [
    awkward1.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]),
    awkward1.Array([[10, 20], [30], [], [40, 50, 60]]),
    awkward1.Array(numpy.array([[101, 102, 103], [201, 202, 203]])),
]
records = [
    awkward1.Array([[{"x": 0.0, "y": []}, {"x": 1.1, "y": [1]}], [], [{"x": 2.2, "y": [1, 2]}]]),
    awkward1.Array([[{"x": 3.3, "y": [1, 2, 3]}], [{"x": 4.4, "y": [2, 1]}, {"x": 5.5, "y": [1]}]]),
]
```

That is, lists like `numbers + records`, `numbers[:1] + records + numbers[1:]`, etc.? The two constraints that must be upheld are

   1. It must call `mergemany` exactly once on a contiguous run of same-type arrays (all numbers or all records).
   2. It must not call `mergemany` on different-type arrays.

You've also implemented a `mergemany_as_union`, which would be nice to call only once, rather than for every pair. Note that a UnionArray is `mergeable` with anything, since the merge would only make a UnionArray with more unique tags.

ianna:
@jpivarski - the branch is up-to-date. Judging by the tests - it looks like the error was introduced in the https://github.com/scikit-hep/awkward-1.0/pull/361/commits/74fb429799996cbe123b9715aec1550ce892af13 commit. I've reverted the changes to both `src/libawkward/Content.cpp` and `src/python/content.cpp`. It's unlikely that changes in the `tests/test_0184-concatenate-operation.py` caused it, so it must be due to `src/awkward1/operations/structure.py`.

ianna:
@jpivarski - The `tests/test_0397-arrays-as-constants-in-numba.py` is SKIPPED on the following architectures due to it's minimum requirement of Python version, I guess:

- Windows py27-32bit
- Windows py27-64bit
- Windows py35-64bit
- Windows py39-32bit
- Windows py39-64bit
- MacOS py27
- MacOS py35
- MacOS py39
- Linux py27-np13
- Linux py27-np16
- Linux py35-np13
- Linux py39-np*

I have installed numba locally and when I run on a recent clean master, all the tests including the above one run fine:
```
% python localbuild.py --pytest tests 
...
```

```python
tests/test_0397-arrays-as-constants-in-numba.py::test_refcount PASSED                                                                                                                                                                                [ 90%]
tests/test_0397-arrays-as-constants-in-numba.py::test_Array PASSED                                                                                                                                                                                   [ 91%]
tests/test_0397-arrays-as-constants-in-numba.py::test_Record PASSED                                                                                                                                                                                  [ 91%]
tests/test_0397-arrays-as-constants-in-numba.py::test_ArrayBuilder PASSED                                                                                                                                                                            [ 91%]

```

However, if I try to rerun the test, all four fail. Am I missing something?

```python
% python localbuild.py --pytest tests/test_0397-arrays-as-constants-in-numba.py
cmake --build localbuild -- -j12
[ 21%] Built target awkward-objects
[ 90%] Built target awkward-cpu-kernels-objects
[ 91%] Built target awkward-cpu-kernels-static
[ 91%] Built target awkward-cpu-kernels
[ 91%] Built target awkward-static
[ 92%] Built target awkward
[ 93%] Built target test0030
[ 94%] Built target test0019
[ 95%] Built target test0016
[ 96%] Built target test0074
[100%] Built target _ext
python -m pytest -vv -rs tests/test_0397-arrays-as-constants-in-numba.py
=================================================================================================================== test session starts ====================================================================================================================
platform darwin -- Python 3.8.1, pytest-5.3.5, py-1.8.1, pluggy-0.13.1 -- /Users/yana/.pyenv/versions/3.8.1/bin/python
cachedir: .pytest_cache
rootdir: /Users/yana/Projects/PR184-rebase-with-numba/awkward-1.0, inifile: setup.cfg
collected 4 items                                                                                                                                                                                                                                          

tests/test_0397-arrays-as-constants-in-numba.py::test_refcount FAILED                                                                                                                                                                                [ 25%]
tests/test_0397-arrays-as-constants-in-numba.py::test_Array FAILED                                                                                                                                                                                   [ 50%]
tests/test_0397-arrays-as-constants-in-numba.py::test_Record FAILED                                                                                                                                                                                  [ 75%]
tests/test_0397-arrays-as-constants-in-numba.py::test_ArrayBuilder FAILED                                                                                                                                                                            [100%]

========================================================================================================================= FAILURES =========================================================================================================================
______________________________________________________________________________________________________________________ test_refcount _______________________________________________________________________________________________________________________

    def test_refcount():
        array = awkward1.Array([1, 2, 3])
    
        @numba.njit
        def f1():
            array
            return 3.14
    
        @numba.njit
        def f2():
            array, array
            return 3.14
    
        assert sys.getrefcount(array) == 2
>       f1()

tests/test_0397-arrays-as-constants-in-numba.py:29: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:415: in _compile_for_args
    error_rewrite(e, 'typing')
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:358: in error_rewrite
    reraise(type(e), e, None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tp = <class 'numba.core.errors.TypingError'>
value = TypingError('Failed in nopython mode pipeline (step: nopython frontend)\nUntyped global name \'array\': cannot determi...ray\'>\n\nFile "tests/test_0397-arrays-as-constants-in-numba.py", line 20:\n    def f1():\n        array\n        ^\n')
tb = None

    def reraise(tp, value, tb=None):
        if value is None:
            value = tp()
        if value.__traceback__ is not tb:
>           raise value.with_traceback(tb)
E           numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E           Untyped global name 'array': cannot determine Numba type of <class 'awkward1.highlevel.Array'>
E           
E           File "tests/test_0397-arrays-as-constants-in-numba.py", line 20:
E               def f1():
E                   array
E                   ^

../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/utils.py:80: TypingError
________________________________________________________________________________________________________________________ test_Array ________________________________________________________________________________________________________________________

    def test_Array():
        array = awkward1.Array([1, 2, 3])
    
        @numba.njit
        def f1():
            array
            return 3.14
    
>       f1()

tests/test_0397-arrays-as-constants-in-numba.py:48: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:415: in _compile_for_args
    error_rewrite(e, 'typing')
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:358: in error_rewrite
    reraise(type(e), e, None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tp = <class 'numba.core.errors.TypingError'>
value = TypingError('Failed in nopython mode pipeline (step: nopython frontend)\nUntyped global name \'array\': cannot determi...ray\'>\n\nFile "tests/test_0397-arrays-as-constants-in-numba.py", line 45:\n    def f1():\n        array\n        ^\n')
tb = None

    def reraise(tp, value, tb=None):
        if value is None:
            value = tp()
        if value.__traceback__ is not tb:
>           raise value.with_traceback(tb)
E           numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E           Untyped global name 'array': cannot determine Numba type of <class 'awkward1.highlevel.Array'>
E           
E           File "tests/test_0397-arrays-as-constants-in-numba.py", line 45:
E               def f1():
E                   array
E                   ^

../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/utils.py:80: TypingError
_______________________________________________________________________________________________________________________ test_Record ________________________________________________________________________________________________________________________

    def test_Record():
        record = awkward1.Record({"x": 1, "y": [1, 2, 3]})
    
        @numba.njit
        def f1():
            return record.y[1]
    
>       assert f1() == 2

tests/test_0397-arrays-as-constants-in-numba.py:92: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:415: in _compile_for_args
    error_rewrite(e, 'typing')
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:358: in error_rewrite
    reraise(type(e), e, None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tp = <class 'numba.core.errors.TypingError'>
value = TypingError('Failed in nopython mode pipeline (step: nopython frontend)\nUntyped global name \'record\': cannot determ...e "tests/test_0397-arrays-as-constants-in-numba.py", line 90:\n    def f1():\n        return record.y[1]\n        ^\n')
tb = None

    def reraise(tp, value, tb=None):
        if value is None:
            value = tp()
        if value.__traceback__ is not tb:
>           raise value.with_traceback(tb)
E           numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E           Untyped global name 'record': cannot determine Numba type of <class 'awkward1.highlevel.Record'>
E           
E           File "tests/test_0397-arrays-as-constants-in-numba.py", line 90:
E               def f1():
E                   return record.y[1]
E                   ^

../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/utils.py:80: TypingError
____________________________________________________________________________________________________________________ test_ArrayBuilder _____________________________________________________________________________________________________________________

    def test_ArrayBuilder():
        builder = awkward1.ArrayBuilder()
        assert sys.getrefcount(builder._layout) == 3
    
        @numba.njit
        def f():
            builder.append(1)
            builder.append(2)
            builder.append(3)
            return builder, builder
    
        @numba.njit
        def g():
            builder.append(1)
            builder.append(2)
            builder.append(3)
    
>       b, c = f()

tests/test_0397-arrays-as-constants-in-numba.py:112: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:415: in _compile_for_args
    error_rewrite(e, 'typing')
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/dispatcher.py:358: in error_rewrite
    reraise(type(e), e, None)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tp = <class 'numba.core.errors.TypingError'>
value = TypingError('Failed in nopython mode pipeline (step: nopython frontend)\nUntyped global name \'builder\': cannot deter...le "tests/test_0397-arrays-as-constants-in-numba.py", line 101:\n    def f():\n        builder.append(1)\n        ^\n')
tb = None

    def reraise(tp, value, tb=None):
        if value is None:
            value = tp()
        if value.__traceback__ is not tb:
>           raise value.with_traceback(tb)
E           numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E           Untyped global name 'builder': cannot determine Numba type of <class 'awkward1.highlevel.ArrayBuilder'>
E           
E           File "tests/test_0397-arrays-as-constants-in-numba.py", line 101:
E               def f():
E                   builder.append(1)
E                   ^

../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numba/core/utils.py:80: TypingError
==================================================================================================================== 4 failed in 1.17s =====================================================================================================================
  File "localbuild.py", line 131, in <module>
    check_call(["python", "-m", "pytest", "-vv", "-rs", args.pytest], env=env)
  File "localbuild.py", line 56, in check_call
    return subprocess.check_call(args, env=env)
  File "/Users/yana/.pyenv/versions/3.8.1/lib/python3.8/subprocess.py", line 364, in check_call
subprocess.CalledProcessError: Command '['python', '-m', 'pytest', '-vv', '-rs', 'tests/test_0397-arrays-as-constants-in-numba.py']' returned non-zero exit status 1.

```

jpivarski:
You're right that the restrictions on testing with Numba are based on Python version and platform constraints. Numba doesn't support Python 2 anymore (our minimal Numba version doesn't), and it doesn't support Python 3.9 _yet_. Supporting a new version of Python is difficult for Numba because it directly uses the Python bytecode, which changes from one minor version to the next.

For these particular errors, it looks like Awkward's Numba extensions have not been registered. This happens through Python's "entry points" mechanism, which might only be enabled with a full package installation. That is, you'd need to install Awkward with pip or conda (it can be `pip install .` in your `awkward-1.0` repo directory) just to get the entry point registered, then the version in the `localbuild` directory would override it if you invoke Python from `awkward-1.0`, as you have been.

If that makes the difference, I'll make a note of it in the CONTRIBUTING.md.

tests/test_0184-concatenate-operation.py
```diff
@@ -0,0 +1,179 @@
...
+    assert awkward1.to_list(numpy.concatenate([np2, np1], 1)) == awkward1.to_list(awkward1.concatenate([ak2, ak1], 1))
+    assert awkward1.to_list(numpy.concatenate([np1, np2], 2)) == awkward1.to_list(awkward1.concatenate([ak1, ak2], 2))
+    assert awkward1.to_list(numpy.concatenate([np2, np1], 2)) == awkward1.to_list(awkward1.concatenate([ak2, ak1], 2))
+
+def test_numbers_and_records_concatenate():
+
+    numbers = [
+        awkward1.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5], []]),
+        awkward1.Array([[10, 20], [30], [], [40, 50, 60]]),
+        awkward1.Array(numpy.array([[101, 102, 103], [201, 202, 203], [], []])),
```
ianna: @jpivarski - `broadcast_and_apply` doesn't like different lengths. What is a correct approach here?
```python
        numbers = [
            awkward1.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]),
            awkward1.Array([[10, 20], [30], [], [40, 50, 60]]),
            awkward1.Array(numpy.array([[101, 102, 103], [201, 202, 203]])),
        ]
        records = [
            awkward1.Array([[{"x": 0.0, "y": []}, {"x": 1.1, "y": [1]}], [], [{"x": 2.2, "y": [1, 2]}]]),
            awkward1.Array([[{"x": 3.3, "y": [1, 2, 3]}], [{"x": 4.4, "y": [2, 1]}, {"x": 5.5, "y": [1]}]]),
        ]
    
>       assert awkward1.to_list(awkward1.concatenate(numbers, 1)) == [

tests/test_0184-concatenate-operation.py:166: 

ValueError: cannot broadcast RegularArray of size 3 with RegularArray of size 4
E                           
E                           (https://github.com/scikit-hep/awkward-1.0/blob/0.4.4/src/awkward1/_util.py#L754)
```

ianna (commented):
@jpivarski - please, have a look at the example results. Is it what you'd expect?

tests/test_0184-concatenate-operation.py
jpivarski: So, `broadcast_and_apply` _should_ be requiring all input arrays to have equal lengths at all levels _before_ you apply your customization (through `getfunction`). So if `axis=10`, all of the input arrays would have to have the same lengths for 10 levels of recursion, and then any lengths they have deeper than that do not matter. For instance, with `axis=1`, all input arrays must have the same length (4 in this example). If `axis=2`, then they must all have the same outer array length as well as all nested list lengths, but then they can have whatever they want deeper than that.

These arrays can only do `axis=0` and `axis=1`. Looking closely, yes, I goofed up all their lengths. The first one has length 3, the second has length 4, and so you get the error message. Here's a better set:

```python
        numbers = [
            awkward1.Array([
                [1.1, 2.2, 3.3],
                [],
                [4.4, 5.5],
                [6.6, 7.7, 8.8, 9.9],
            ]),
            awkward1.Array([
                [10, 20],
                [30],
                [],
                [40, 50, 60],
            ]),
            awkward1.Array(numpy.array([
                [101, 102, 103],
                [201, 202, 203],
                [301, 302, 303],
                [401, 402, 403],
            ])),
        ]
        records = [
            awkward1.Array([
                [{"x": 0.0, "y": []}, {"x": 1.1, "y": [1]}],
                [],
                [{"x": 2.2, "y": [1, 2]}],
                [{"x": 3.3, "y": [1, 2, 3]}, {"x": 4.4, "y": [1, 2, 3, 4]}],
            ]),
            awkward1.Array([
                [{"x": 0.0, "y": []}],
                [{"x": 1.1, "y": [1]}, {"x": 2.2, "y": [1, 2]}],
                [],
                [{"x": 3.3, "y": [1, 2, 3]}],
            ]),
        ]
```

They all have length 4. The lengths within each list differ wildly, but `broadcast_and_apply` should only do its length check for the first level when `axis=1`.

Here's what you should get from `ak.concatenate([numbers[0], numbers[1]], axis=1)`:

```python
[
    [1.1, 2.2, 3.3, 10, 20],
    [30],
    [4.4, 5.5],
    [6.6, 7.7, 8.8, 9.9, 40, 50, 60],
]
```

And here's what you should get from `ak.concatenate([numbers[0], records[0]], axis=1)`:

```python
[
    [1.1, 2.2, 3.3, {"x": 0.0, "y": []}, {"x": 1.1, "y": [1]}],
    [],
    [4.4, 5.5, {"x": 2.2, "y": [1, 2]}],
    [6.6, 7.7, 8.8, 9.9, {"x": 3.3, "y": [1, 2, 3]}, {"x": 4.4, "y": [1, 2, 3, 4]}],
]
```

If you want an example that tests `broadcast_and_apply`'s length checking at more than one level, there's this:

```python
arrays = [
    ak.Array([[[0.0, 1.1, 2.2], []], [[3.3, 4.4]], [[5.5], [6.6, 7.7, 8.8, 9.9]]]),
    ak.Array([[[10, 20], [30]], [[40]], [[50, 60, 70], [80, 90]]]),
]
# nothing is required to have the same length
assert ak.concatenate(arrays, axis=0).tolist() == [
    [[0.0, 1.1, 2.2], []],
    [[3.3, 4.4]],
    [[5.5], [6.6, 7.7, 8.8, 9.9]],
    [[10, 20], [30]],
    [[40]],
    [[50, 60, 70], [80, 90]],
]
# the outermost arrays are required to have the same length, but nothing deeper than that
assert ak.concatenate(arrays, axis=1).tolist() == [
    [[0.0, 1.1, 2.2], [], [10, 20], [30]],
    [[3.3, 4.4], [40]],
    [[5.5], [6.6, 7.7, 8.8, 9.9], [50, 60, 70], [80, 90]],
]
# the outermost arrays and the first level are required to have the same length, but nothing deeper
assert ak.concatenate(arrays, axis=2).tolist() == [
    [[0.0, 1.1, 2.2, 10, 20], [30]],
    [[3.3, 4.4, 40]],
    [[5.5, 50, 60, 70], [6.6, 7.7, 8.8, 9.9, 80, 90]],
]
```

You should be able to concatenate `[arrays[0][:, :-1], arrays[1]]` at `axis=1`, but not at `axis=2`.

ianna: ok, the tests pass, but there is a warning:
```python
tests/test_0184-concatenate-operation.py::test_list_offset_array_concatenate PASSED                                                                                                                                                                  [ 10%]
tests/test_0184-concatenate-operation.py::test_list_array_concatenate PASSED                                                                                                                                                                         [ 20%]
tests/test_0184-concatenate-operation.py::test_records_concatenate PASSED                                                                                                                                                                            [ 30%]
tests/test_0184-concatenate-operation.py::test_indexed_array_concatenate PASSED                                                                                                                                                                      [ 40%]
tests/test_0184-concatenate-operation.py::test_bytemasked_concatenate PASSED                                                                                                                                                                         [ 50%]
tests/test_0184-concatenate-operation.py::test_listoffsetarray_concatenate PASSED                                                                                                                                                                    [ 60%]
tests/test_0184-concatenate-operation.py::test_numpyarray_concatenate_axis0 PASSED                                                                                                                                                                   [ 70%]
tests/test_0184-concatenate-operation.py::test_numpyarray_concatenate PASSED                                                                                                                                                                         [ 80%]
tests/test_0184-concatenate-operation.py::test_numbers_and_records_concatenate PASSED                                                                                                                                                                [ 90%]
tests/test_0184-concatenate-operation.py::test_jims_test PASSED                                                                                                                                                                                      [100%]

===================================================================================================================== warnings summary =====================================================================================================================
tests/test_0184-concatenate-operation.py::test_numbers_and_records_concatenate
  /Users/yana/Projects/PR184-rebase/awkward-1.0/awkward1/_util.py:64: DeprecationWarning: In version 1.0.0 (target date: 2020-12-01), this will be an error.
  (Set ak.deprecations_as_errors = True to get a stack trace now.)
  
  ValueError: cannot broadcast: <class 'awkward1._ext.RecordArray'>, <class 'awkward1._ext.RecordArray'>
  
  (https://github.com/scikit-hep/awkward-1.0/blob/0.4.4/src/awkward1/_util.py#L884)
    warnings.warn(message, DeprecationWarning)

-- Docs: https://docs.pytest.org/en/latest/warnings.html
============================================================================================================== 10 passed, 1 warning in 0.77s ===============================================================================================================

```

edit:
hmm... here is a future error. 

```python
>       assert ak.concatenate([numbers[0], records[0]], 1) == [
            [1.1, 2.2, 3.3, {"x": 0.0, "y": []}, {"x": 1.1, "y": [1]}],
            [],
            [4.4, 5.5, {"x": 2.2, "y": [1, 2]}],
            [6.6, 7.7, 8.8, 9.9, {"x": 3.3, "y": [1, 2, 3]}, {"x": 4.4, "y": [1, 2, 3, 4]}],
        ]

tests/test_0184-concatenate-operation.py:193: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.1/lib/python3.8/site-packages/numpy/lib/mixins.py:25: in func
    return ufunc(self, other)
awkward1/highlevel.py:1369: in __array_ufunc__
awkward1/_connect/_numpy.py:173: in array_ufunc
    out = awkward1._util.broadcast_and_apply(
awkward1/_util.py:996: in broadcast_and_apply
awkward1/_util.py:759: in apply
awkward1/_util.py:847: in apply
awkward1/_util.py:659: in apply
    outcontents.append(apply(nextinputs, depth))
awkward1/_util.py:886: in apply
    deprecate(exception, "1.0.0", "2020-12-01")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

exception = ValueError("cannot broadcast: <class 'awkward1._ext.RecordArray'>, <class 'awkward1._ext.RecordArray'>\n\n(https://github.com/scikit-hep/awkward-1.0/blob/0.4.4/src/awkward1/_util.py#L884)"), version = '1.0.0', date = '2020-12-01'

    def deprecate(exception, version, date=None):
        if awkward1.deprecations_as_errors:
>           raise exception
E           ValueError: cannot broadcast: <class 'awkward1._ext.RecordArray'>, <class 'awkward1._ext.RecordArray'>
E           
E           (https://github.com/scikit-hep/awkward-1.0/blob/0.4.4/src/awkward1/_util.py#L884)

awkward1/_util.py:54: ValueError

```

I've also removed all kernels I wrote for this PR to make sure this code was not causing the dereference error. The PR is pretty minimal now. It relies on the `offsets_and_flattened` and `broadcast_and_apply`. I can try to move this PR tests after the 0397 test, but apart from that I'm running out of the ideas on how to catch the dereference bug. Could it be that the garbage collector is simply delayed?

src/awkward1/operations/structure.py
ianna: done

ianna:
Ha. All checks passed ;-)

jpivarski:
The warning is a good thing—I've deprecated a feature and wanted to make sure that users will see that message, so I added it to the tests. It's also a reminder to remove the feature in version 1.0.

I was going to suggest bisecting to find what's causing the Windows bug, but you beat me to it.

It would be bad if the Windows bug only happens because the garbage collector runs—delaying the garbage collector is not a fix, it only indicates that the bug is hidden in other systems. But I've already checked for this bug by explicitly invoking the garbage collector (on Linux). I don't think we need to worry about that possibility.

ianna:
> The warning is a good thing—I've deprecated a feature and wanted to make sure that users will see that message, so I added it to the tests. It's also a reminder to remove the feature in version 1.0.
> 
> I was going to suggest bisecting to find what's causing the Windows bug, but you beat me to it.
> 
> It would be bad if the Windows bug only happens because the garbage collector runs—delaying the garbage collector is not a fix, it only indicates that the bug is hidden in other systems. But I've already checked for this bug by explicitly invoking the garbage collector (on Linux). I don't think we need to worry about that possibility.

The `numbers` and `records` are not `mergeable`. They can be merged as a union in `axis=0` whilst in `axis>0` `broadcast_and_apply` will not accept this union due to its different length from the other arrays.

jpivarski:
The `numbers` and `records` are intentionally not mergeable (except "as union"). But the second set, which has the same lengths, should be concatenateable with `axis > 0` ("as union").

tests/test_0184-concatenate-operation.py
ianna: fixed

ianna:
@reikdas - could you, please, have a look with a fresh eye? Why and where does the PR defines `tests-cpu-kernels/test_cpuawkward_RegularArray_merge_offsets64.py`? I think, I've deleted it a while back. Thanks!

reikdas:
> @reikdas - could you, please, have a look with a fresh eye? Why and where does the PR defines `tests-cpu-kernels/test_cpuawkward_RegularArray_merge_offsets64.py`? I think, I've deleted it a while back. Thanks!

I see that there are some commits after this comment and that tests are passing. Is the issue fixed?

ianna:
> > @reikdas - could you, please, have a look with a fresh eye? Why and where does the PR defines `tests-cpu-kernels/test_cpuawkward_RegularArray_merge_offsets64.py`? I think, I've deleted it a while back. Thanks!
> 
> I see that there are some commits after this comment and that tests are passing. Is the issue fixed?

No, I just squashed the last three commits. I've no idea why it all works now. It was failing at a test that did not have any kernel code defined and it still fails in my local area:
```python
========================================================================================================================= FAILURES =========================================================================================================================
______________________________________________________________________________________________________ test_cpuawkward_RegularArray_merge_offsets64_1 ______________________________________________________________________________________________________

    def test_cpuawkward_RegularArray_merge_offsets64_1():
        tooffsets = [123, 123, 123]
        tooffsets = (ctypes.c_int64*len(tooffsets))(*tooffsets)
        tolength = 3
        length = 3
        size = 3
        otherlength = 3
        othersize = 3
>       funcC = getattr(lib, 'awkward_RegularArray_merge_offsets64')

tests-cpu-kernels/test_cpuawkward_RegularArray_merge_offsets64.py:13: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
../../../.pyenv/versions/3.8.1/lib/python3.8/ctypes/__init__.py:386: in __getattr__
    func = self.__getitem__(name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <CDLL '/Users/yana/Projects/PR184-rebase/awkward-1.0/awkward1/libawkward-cpu-kernels.dylib', handle 7f93a8021720 at 0x1110188b0>, name_or_ordinal = 'awkward_RegularArray_merge_offsets64'

    def __getitem__(self, name_or_ordinal):
>       func = self._FuncPtr((name_or_ordinal, self))
E       AttributeError: dlsym(0x7f93a8021720, awkward_RegularArray_merge_offsets64): symbol not found

../../../.pyenv/versions/3.8.1/lib/python3.8/ctypes/__init__.py:391: AttributeError
```
I think, I will start a new branch.

ianna:
move to https://github.com/scikit-hep/awkward-1.0/pull/539

reikdas:
I checked out your branch locally, but I I wasn't able to reproduce the error. Maybe there were some auto generated files that were not removed? 
(Even if that is not the culprit, @jpivarski maybe we should include a script in `dev/` to clean up all auto generated files, apart from the ones that are checked into git like the kernel headers, so that one can start from a clean slate?)

jpivarski:
> (Even if that is not the culprit, @jpivarski maybe we should include a script in `dev/` to clean up all auto generated files, apart from the ones that are checked into git like the kernel headers, so that one can start from a clean slate?)

That's a good idea, though these generated files are not in the `tests` directory, which is where the failing test was.

PR #362, opened 2020-07-28
Add from_cupy to Index class

trickarcher:


jpivarski (changes requested):
I see that this PR only covers Index.from_cupy, but that's okay—it's more focused (and easier to review) that way. I reviewed the source files manually and tested them on the GPU instance.

The source files look good to me and the tests pass. But there's something seriously odd about the fact that wrapping a little 10-element array 9 times takes 30 seconds to run. I say "wrapping" because the time required shouldn't even scale with the size of the array: it's already on the GPU, you're just getting another copy of the pointer.

This speed is an issue for two reasons: the first is that in the course of normal data analysis, a user would be wrapping hundreds to thousands of these, not just 9, and if the time scales like it does here, GPU analysis would be next to impossible. But the other reason is that it's a sign that it's not doing what you think it's doing. Does the `test_cupy_refcount` test take 30 seconds on your laptop, too? (It might just be the AWS test-stand configuration that's wrong, not the code: that's a possibility, but one that needs to be worked out.)

Try figuring out exactly where the time is being spent. That's easiest on your own laptop, but possible through the GPU instance. This is not like tracking down a performance issue, in which you need to carefully consider how many times a given subroutine is called, each contributing microseconds to the whole. This is a whopping 30 seconds (or 3 seconds, 9 times) that very likely happens at some single step in the process. You don't need to time it, you can count it off manually. (I got the initial figure of 30 seconds by counting "one Mississippi, two Mississippi, three Mississippi," then used a stopwatch. [Galileo used his pulse.](https://sites.google.com/site/abriefhistoryoftimelessness/basic-timelessness/galileo-and-time) :)

Another point that should be tested before declaring victory is to verify that you can use this `Index` in a meaningful way on the GPU. Perhaps use `getitem_at` to look at the values or, ideally, the "num" kernel that you've already ported. ("num" applied to a ListOffsetArray's `offsets` doesn't inspect the ListOffsetArray's `content`, so it would be possible to set up such a test before getting NumpyArrays from CuPy.) Maybe the pointer you got points to nothing.

(Minor nit-pick: the tests in `tests-cuda` should be labeled with PR number, just like the ones in `tests`. That makes it easier to find tests relevant for particular code changes. The correlation isn't 100%, but it's better than nothing.)

trickarcher:
So I called the function `test_cupy_refcount` and used the linux command `time python tests-cuda/test_cupy_refcount.py` and here's the result. `python tests-cuda/test_cupy_refcount.py  0.53s user 0.24s system 98% cpu 0.776 total`. That 30s seems a bit too much, I think it might be because the CUDA syncs with the GPU when we first run any cuda program but still 30s is a very big number.
As far as testing the `Index.from_cupy()` is concerned. I forgot to add the `len` and `getitem` tests but you can test it out on the Python prompt and do stuff like `array[3]` or `len(array)` and it works.

trickarcher:
Also, since we are doing some interpreter level things with `pybind11`, I think it has to do that Initialization call 9 times in a row instead of just one time and that might(I am not sure at all about this) be the reason why that 30s is even happening. But as I am wrting this I realized that there's only one place where `cupy` really allocates an array and all the other awkward arrays are just derived from it. So as a hypothesis maybe that time is `Time taken in the CuPy import` which include the `CudaDeviceInitialization` that might happen when we import `CuPy`. Anyways, I'll have to research both on `pybind11` and `CuPy` to see if this is the actual issue. I hope I can resume work from Tuesday and I'll start looking into this issue.

jpivarski:
If you mean the `py::module::import("cupy")`, Python modules can be imported arbitrarily many times without a penalty. The "initialize once; all subsequent times are a pass-through" is built into Python's `import` (which can make it difficult if you actually _do_ want to re-initialize, but that's not our case here).

You can diagnose it by putting print statements between every step—one of them will be a big one.

trickarcher:
Okay, maybe the bottleneck is somewhere else. I'll try figuring it out.

jpivarski:
I'm confused about which PR is which. If this is the one that was fixed earlier today with a clean environment, then let's get it up to date (5 files conflict) and I'll merge it. If this is not supposed to be merged, then let's close it. Thanks!

jpivarski (approved):

jpivarski:
I've lost track of what changes I requested, so I've marked them as approved. But what's the status of this PR, and how does it fit in with #372?

trickarcher:
Can I close this PR, this was intended to be merged a lot earlier, and since then I have moved on to the `Numpy` and `Identites` integration, and that has a lot of different changes within the C++ layer. I'll try to resolve that `to_list` segmentation fault and get the PR to a completion state as soon as possible.

jpivarski:
You can close this.

Does its content exist elsewhere? The title of this PR says that it's introducing the `from_cupy` method to Index—does that happen in a later PR? (I want to make sure each feature gets implemented exactly once!)

trickarcher:
Yes, the Index integration with CuPy exists in #372 PR. So that PR can be treated as one CuPy Integration PR, except the `high level` `to_cupy` and `from_cupy`, because that has some issues for which you might want to look at Slack, where I detailed them out.

PR #365, opened 2020-08-04
Do not iterate over lists while comparing in pytest

reikdas:


jpivarski:
This looks like it does the right thing and the tests pass. I'll take your word for it that the it's modifying the tests-kernels that get generated and they are running in the tests that pass. Give me a thumbs-up on that and I'll merge it.

PR #369, opened 2020-08-04
Improve C++ to C generator

reikdas:
Should fix the errors that @ianna has been seeing in the CI for her PR.


jpivarski (approved):
I can't tell if it's correct from reading it, but if the tests pass, I guess it worked. (Or maybe we need both these tests and @ianna's to pass—but anyway, if these pass, I'll merge it so that it can be merged into @ianna's.

PR #372, opened 2020-08-06
Add to and from_cupy for Numpy Array and Identities

trickarcher:


src/python/content.cpp
```diff
@@ -1710,6 +1710,82 @@ make_NumpyArray(const py::handle& m, const std::string& name) {
...
+              (py::module::import("cupy").attr("ndarray")(
+                pybind11::make_tuple(py::cast<ssize_t>(cuda_arr->length())),
+                ak::util::dtype_to_format(cuda_arr->dtype()),
+                cupy_memoryptr,
+                pybind11::make_tuple(py::cast<ssize_t>(cuda_arr->itemsize()))));
+          }
+          else {
+            throw std::invalid_argument("specify 'cpu' or 'cuda'");
+          }
+      })
```
jpivarski: You've used the name `copy_to` for a different purpose: converting Awkward-CPU objects into Awkward-GPU objects or vice-versa. This function is "`to_numpy`" or "`to_cupy`". Rather than depending on a `ptr_lib` string (the way that `copy_to` does, because it's more symmetric), this should probably be two functions and raise an exception if you're trying to convert an Awkward-CPU object into a CuPy object or an Awkward-GPU object into a NumPy object.

src/python/content.cpp
```diff
@@ -1710,6 +1710,82 @@ make_NumpyArray(const py::handle& m, const std::string& name) {
...
+            py::dtype(array.attr("dtype")).itemsize(),
+            ak::util::dtype_to_format(cupy_dtype),
+            cupy_dtype,
+            ak::kernel::lib::cuda);
+        }
+        else {
+          throw std::invalid_argument(name + std::string(
+            ".from_cupy() can only accept CuPy Arrays!"));
+        }
+    }, py::arg("array"),
```
jpivarski: Some platforms are having trouble compiling this. Don't make the function signature different from the other functions that return layout nodes. The others have a formal return type of `py::object` (the Python object corresponding to a NumpyArray, for instance) and use the `box` helper function to convert a `std::shared_ptr<Content>` into the Python object.

On platforms where this even compiles, it could be a source of segfaults because this NumpyArray is being allocated on the stack and pybind11 doesn't know that it's supposed to take a `shallow_copy` of it to get a copy on the heap. Following the pattern set by all the other functions that return layout nodes, you should

```c++
return box(std::make_shared<ak::NumpyArray>(... arguments ...));
```

to allocate the NumpyArray directly on the heap (no `shallow_copy` needed), get a reference-counted pointer to it, and then `box` it as a Python object. (In principle, pybind11 can do the boxing by itself, but some functions that can return layout nodes can also return scalars, like `None` or numbers. The `box` function does the conversion manually to handle the special cases, and for uniformity, we use it everywhere in this file.)

src/libawkward/array/NumpyArray.cpp
```diff
@@ -4820,6 +4820,157 @@ namespace awkward {
...
+        break;
+      case util::dtype::int8:
+        ptr = kernel::ptr_alloc<int8_t>(ptr_lib, (int64_t)length);
+        err = kernel::copy_to<int8_t>(
+          ptr_lib,
+          ptr_lib_,
+          reinterpret_cast<int8_t*>(ptr.get()),
+          reinterpret_cast<int8_t*>(data()),
+          length);
+        break;
```
jpivarski: Maybe I should have brought this up earlier, but why does `copy_to` depend on data type? Without casting, the NumpyArray carries a `void*` (actually, `std::shared_ptr<void>`). To copy data from main memory to GPU or back, the CUDA copy-to-device or copy-to-host functions might take a `char*`, but that's one `reinterpret_cast`. You shouldn't need to specialize by dtype just to copy data (you have the `strides` and `shape` to determine the number of bytes to copy). That could save a lot of unnecessary functions.

jpivarski (commented):
It's my understanding that this is a work in progress—I've put some comments inline.

src/libawkward/array/NumpyArray.cpp
trickarcher: Yes, this makes sense. I could pass the `ptr` and the number of bytes to copy from the head of the ptr (`length() * itemsize()`) and that would be suffice for the `cudaMallocManaged`

include/awkward/Index.h
```diff
@@ -90,7 +90,7 @@ namespace awkward {
     IndexOf<T>(const std::shared_ptr<T>& ptr,
                int64_t offset,
                int64_t length,
-               kernel::lib ptr_lib = kernel::lib::cpu);
```
jpivarski, line 93: It would be fair to say that your troubles with segfaults were mostly due to this line and the equivalent in NumpyArray.h (line 175): letting the default `ptr_lib` label be `kernel::lib::cpu` led to a lot of cases in which a buffer was really on the GPU but the object said it was on the CPU. This happened especially because of `shallow_copy`; these C++ objects are lightweight "handles" to the arrays, so they're shallowly copied all the time, and `shallow_copy` was taking `ptr_lib = kernel::lib::cpu` as a default. The cases that _didn't_ segfault were due to considerable luck that the Index/NumpyArray/Identities weren't copied.

This is why I don't like default parameters (in C++ or in Python) for internal functions. For our purposes, all C++ functions are internal functions, so that's why so little of the C++ codebase has default parameters. Default parameters are great for user convenience, so the highlevel interface has a lot of default parameters, mostly opt-in things with a default of None.

This is similar to using fully qualified names in the code (e.g. `awkward1.operations.convert.from_cupy`) but not in the user interface (e.g. `ak.from_cupy`). For internal functions, convenience is bad and explicitness is good. For users, especially data analysts working interactively in Jupyter, the opposite is true.

include/awkward/array/NumpyArray.h
```diff
@@ -172,7 +172,7 @@ namespace awkward {
                ssize_t itemsize,
                const std::string format,
                util::dtype dtype,
-               const kernel::lib ptr_lib = kernel::lib::cpu);
+               const kernel::lib ptr_lib);
```
jpivarski, line 175: See comment on Index.h line 93.

src/awkward1/_util.py
```diff
@@ -1145,6 +1150,65 @@ def recursively_apply(layout, getfunction, args=(), depth=1, keep_parameters=Tru
         )
 
 
+def recursive_walk(layout, apply, args=(), depth=1, materialize=False):
```
jpivarski, line 1153: This is for determining the overall `ptr_lib` from a tree structure of nested array nodes. I split the recursive part and the `ptr_lib`-determining part, with the recursion here in `_util.py` as a utility function. That way, the logic of how to walk down a tree of nested array nodes is in one place, and if we ever add a new node type, it can be updated in one place.

To do that, the action of finding `ptr_lib` has to be defined in a function, with that function being passed in here. The `braodcast_and_apply` and `apply_recursively` work the same way. This is an example of the value of functional programming. There are old articles about how LISP seemed "magical" to C programmers back in the 80's—this kind of factorization over arbitrary functions is very useful! (Though if it's overused, it leads to Javascript's callback hell.)

src/awkward1/highlevel.py
```diff
@@ -29,6 +29,14 @@
 _dir_pattern = re.compile(r"^[a-zA-Z_]\w*$")
 
 
+def _suffix(array):
+    out = awkward1.operations.convert.kernels(array)
+    if out == "cpu":
+        return ""
+    else:
+        return ":" + out
```
jpivarski, line 37: The new highlevel interface is

```python
>>> import awkward1 as ak
>>> import cupy as cp
>>> array = ak.Array(cp.arange(10))
>>> array
<Array:cuda [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] type='10 * int64'>    # <-- note the :cuda
>>> ak.kernels(array)
'cuda'
>>> array2 = ak.to_kernels(array, "cpu")
>>> array2
<Array [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] type='10 * int64'>
>>> ak.kernels(array2)
'cpu'
```

src/awkward1/__init__.py
```diff
@@ -63,7 +63,6 @@
 from awkward1.operations.describe import *
 from awkward1.operations.structure import *
 from awkward1.operations.reducers import *
-from awkward1.operations.transfers import *
```
jpivarski, line 66: I moved the `copy_to` function into convert.py and renamed it `to_kernels` to fit the pattern of the other "from" and "to" functions.

src/awkward1/nplike.py
```diff
@@ -269,6 +303,25 @@ def ma(self):
...
+            awkward1.layout.Content,
+            awkward1.layout.Record,
+        )):
+            out = awkward1.operations.convert.to_cupy(array)
+            if dtype is not None and out.dtype != dtype:
+                return self._module.asarray(out, dtype=dtype)
+            else:
+                return out
+        else:
+            return self._module.asarray(array, dtype=dtype)
```
jpivarski, line 323: This is what I was talking about on Friday. CuPy's `asarray` expects the input to be in main memory, but we want it to be like `numpy.asarray` except expect the input to be on the GPU (to get a lot of already-written code to work with a substitution of libraries). The NumPy and CuPy libraries aren't symmetric in that particular way, so we have these classes to act as indirection between what the libraries actually do and what we need them to do. We need `asarray` to be `to_cupy` for GPU-bound Awkward arrays, so I just overloaded that behavior here. There might be a few more subtleties, but we can fix them as they come up because we have this indirection layer.

src/python/content.cpp
```diff
@@ -28,14 +28,70 @@ box(const std::shared_ptr<ak::Content>& content) {
...
     if (raw->isscalar()) {
-      return py::array(py::buffer_info(
-        raw->data(),
-        raw->itemsize(),
-        raw->format(),
-        raw->ndim(),
-        raw->shape(),
-        raw->strides()
-      )).attr("item")();
+      switch (raw->dtype()) {
```
jpivarski, line 31: Despite what I said on Friday, I left the switch over dtype here, largely because we need the types to match. Hiding this dependence on dtype would be a good thing to do in principle, since the changes would be more consolidated if we added a new one, but since we'd have to deal with types identified at runtime anyway, this split-by-type would still be here. Since the general rule isn't helping, we break the rule.

src/python/content.cpp
```diff
@@ -1676,6 +1734,52 @@ make_ListOffsetArrayOf(const py::handle& m, const std::string& name);
 
 ////////// NumpyArray
 
+const ak::NumpyArray
+NumpyArray_from_cupy(const std::string& name,
+                     const py::object& array,
+                     const py::object& identities,
+                     const py::object& parameters) {
```
jpivarski, line 1741: This was split out so that it can be used in the NumpyArray constructor as well, and so a lot of places in the Python code don't have to be changed to "if CuPy, call `NumpyArray.from_cupy`, else call the constructor." This is a case of hiding complexity so that more code can be naive.

jpivarski (approved):
Thank you for all of your work this summer! This is a good PR, though it had a lot of loose ends because it had a huge scope.

I started by fixing the issues you were seeing earlier, but then added a lot of code cleanups (adherence to conventions) and the high-level interface. That meant some renaming; it's all described below.

The inline comments aren't things to change—I've already changed them—but they're explanations if you're looking over the changes, so that you understand why I changed them. The key thing, though, is the first one: your segfaults can be traced back to the use of a default parameter for `ptr_lib`; the choice of `kernel::lib::cpu` shouldn't be assumed because it's very common for NumpyArrays and Indexes to be built from preexisting NumpyArrays and Indexes, some of which have `kernel::lib::cuda`. This is especially true of `shallow_copy`, which is used frequently and just gives us a new handle to the same array. The first segfault I encountered was when it was simply doing a `shallow_copy` for formal reasons and then ended up calling a CUDA array `kernel::lib::cpu`.

I'm going to merge this as soon as the tests pass.

PR #377, opened 2020-08-07
Rename 'astype' to 'numbers_to_type', for use as a high-level function. Also removed 'can_cast', since it's exactly the same as the NumPy version(users should use NumPy).

### LINKED ISSUE #376: Don't forget to add a high-level ak.numbers_to_type function.
A high-level wrapper for the function @ianna added of the same name. It should go in src/awkward1/operations/structure.py.

jpivarski:


jpivarski:
The function was named `ak.astype` in #346, but now it's `ak.numbers_to_type` and it's documented.

PR #378, opened 2020-08-07
Update the what-is-awkward to align with (and include) the video.

jpivarski:


jpivarski:
Only the doctest needs to pass here (and I'm not sure it's testing the JupyterBook part). The real test will be to see if Netlify can run it.

PR #380, opened 2020-08-10
Example of Identities_to_Identities64 integration with the codebase.

trickarcher:
@reikdas I have shown an example as to how you could go about integrating functions into the codebase. @jpivarski This can be merged if you want. But if you'd like to merge this, can you give me an example for writing tests for `Identities`(I'd like to have a integration test in place although since it's just casting it should pass)

### DIFF
--- a/src/cuda-kernels/identities.cu
+++ b/src/cuda-kernels/identities.cu
@@ -1,0 +1,43 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/master/LICENSE
+
+#include "awkward/cuda-kernels/cuda_identities.h"
+#include <stdio.h>
+
+__global__
+void cuda_Identities32_toIdentities64(
+  int64_t* toptr,
+  const int32_t* fromptr,
+  int64_t length) {
+  int64_t block_id = blockIdx.x + blockIdx.y * gridDim.x + gridDim.x * gridDim.y * blockIdx.z;
+  int64_t thread_id = block_id * blockDim.x + threadIdx.x;
+  if(thread_id < length) {
+    toptr[thread_id] = (int64_t)(fromptr[thread_id]);
+  }
+}
+
+ERROR awkward_cuda_Identities32_to_Identities64(
+  int64_t* toptr,
+  const int32_t* fromptr,
+  int64_t length,
+  int64_t width) {
+
+  dim3 blocks_per_grid;
+  dim3 threads_per_block;
+
+  if (length > 1024) {
+    blocks_per_grid = dim3(ceil((length) / 1024.0), 1, 1);
+    threads_per_block = dim3(1024, 1, 1);
+  } else {
+    blocks_per_grid = dim3(1, 1, 1);
+    threads_per_block = dim3(length, 1, 1);
+  }
+
+  cuda_Identities32_toIdentities64<<<blocks_per_grid, threads_per_block>>>(
+    toptr,
+    fromptr,
+    length * width);
+
+  cudaDeviceSynchronize();
+
+  return success();
+}
--- a/include/awkward/cuda-kernels/cuda_identities.h
+++ b/include/awkward/cuda-kernels/cuda_identities.h
@@ -1,0 +1,17 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/master/LICENSE
+
+
+#ifndef AWKWARD_IDENTITIES_H
+#define AWKWARD_IDENTITIES_H
+
+#include "awkward/common.h"
+
+extern "C" {
+  ERROR awkward_cuda_Identities32_to_Identities64(
+  int64_t* toptr,
+  const int32_t* fromptr,
+  int64_t length,
+  int64_t width);
+};
+
+#endif //AWKWARD_IDENTITIES_H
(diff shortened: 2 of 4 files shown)


src/cuda-kernels/identities.cu
```diff
@@ -0,0 +1,43 @@
...
+ERROR awkward_cuda_Identities32_to_Identities64(
+  int64_t* toptr,
+  const int32_t* fromptr,
+  int64_t length,
+  int64_t width) {
+
+  dim3 blocks_per_grid;
+  dim3 threads_per_block;
+
+  if (length > 1024) {
```
reikdas, line 27: Should this be `(length * width) > 1024` instead? (same for the other references to `length` when assigning `blocks_per_grid` and `threads_per_block`)

src/cuda-kernels/identities.cu
```diff
@@ -0,0 +1,43 @@
...
+    threads_per_block = dim3(1024, 1, 1);
+  } else {
+    blocks_per_grid = dim3(1, 1, 1);
+    threads_per_block = dim3(length, 1, 1);
+  }
+
+  cuda_Identities32_toIdentities64<<<blocks_per_grid, threads_per_block>>>(
+    toptr,
+    fromptr,
+    length * width);
```
reikdas, line 38: I would prefer to pass `length` and `width` as separate arguments since it is easier to reason about in the code generation.

src/cuda-kernels/identities.cu
trickarcher: Yes, you're right. It should be `length * width`. I'll put this entire thing in a utility function which could be named, `fetch_num_blocks` and `fetch_num_threads` and call that.

src/cuda-kernels/identities.cu
trickarcher: Okay, I did this to reduce computation within the kernel, since global memory is slow. Doing this will not break things, but if `nvcc` doesn't optimise this there could be a slow down. But we can catch all these cases after the generation, right?

jpivarski: When you say, "did this," what "this" are you referring to? What are you expecting to be slow and how did you try to fix it (or not)?

In general, we can put off performance tweaks for later, though we should design around a structure that can be performant.

trickarcher: So, by "did this" I was referring to how I passed `length * width` as one parameter to do the multiplication on the host side, rather than on the device side(multiple times for each thread).

jpivarski: Yes, we should absolutely not worry about whether a multiplication is done on the host side or device side, and whether it's done once per thread or once for all. Addition and multiplication are the two fastest operations a processor can do and they're usually vastly overshadowed by time spent fetching memory. On a GPU, the memory-fetching is less significant because global memory is essentially like a CPU's cache—it's close to the processor—but still, it's unlikely that the other work these kernels need to do (especially the parallel prefix sum ones) will be close to the speed of multiplication.

Also, it's not "structural," we can easily change it in the future, in response to profiling tests. So it's not something that we'll worry about right now.

src/cuda-kernels/identities.cu
jpivarski: It doesn't look like that's happened yet, so I won't merge right away. But for generality of utility functions, note that the three cases all do different things:

   * NumpyArray has a raw buffer of bytes; how it should be divided up into threads and blocks depends on if and how those bytes are being interpreted. Reducers interpret the bytes as numbers, but basic copy operations like `carry` only need to know byte ranges. NumpyArrays might not be contiguous: `array[:, :-1]` is an example of an operation that takes a contiguous `array` to a non-contiguous one. For some operations, dealing with non-contiguousness was too hard to think about so we run `NumpyArray::contiguous` to get it into that state before calling the kernel, but others deal with the non-contiguousness via `strides` and `shape`.
   * Index always has a compile-time integer type, is always one-dimensional, and is always contiguous. We use Indexes a lot more than NumpyArray (for example, `ak.num` doesn't need to see the underlying NumpyArray: the number of items is described by a list node's `offsets`, `starts`/`stops`, or `size`). Since we use them everywhere and don't want to overcomplicate the code, Index is forced to be contiguous.
   * Identities always has a compile-time integer type, is always two-dimensional, and is always contiguous. We also don't use them much.

Even though one of these (Identities) is two-dimensional, and some NumpyArray data are n-dimensional (if `len(shape) > 1`), I don't think we ever need to use CUDA's two or three-dimensional blocks because we never make use of n-dimensional "neighbor-ness" in any kernels. I think the two and three-dimensional blocks exist for people who are doing image analysis or PDE solving, in which memory for cell `(i, j)` needs to be close to `(i + 1, j)` and `(i, j + 1)` because they're taking differences or computing convolutions across those dimensions. Some of our (non-embarrassingly parallel) kernels need to relate cell `i` with `i + 1`, but never anything that resembles a two or three-dimensional grid.

Just in case you're thinking that we might need to do that—we don't.

jpivarski:
Is this PR intended as an example (from @trickarcher to @reikdas) or as something to be merged in? It's getting a little stale as src/libawkward/kernel.cpp has changed. If it's an example, let's convert it into a draft and eventually close it after Reik is done, so that I'm not asking myself whether I need to merge it over and over. If it's meant to be merged in, then it will need to be brought up to date and I'll review it.

reikdas:
(I hope I am not stepping on any toes) I guess we could either merge it in after @trickarcher finishes making the possible changes discussed inline, or we could convert this PR to a draft since I will be attempting to generate this in my PR anyway and then close it when my PR has been merged, like @jpivarski suggested.

jpivarski:
This is still a good example for @reikdas to follow, but I'm closing it because we have no intention of merging it.

PR #382, opened 2020-08-10
Fixed broken kernel page in sphinx docs and silenced a sphinx-build warning

reikdas:
Somehow the kernel documentation at https://awkward-array.readthedocs.io/en/latest/_auto/kernels.html broke with the latest commits to master - this fixes it.

And this also silences a sphinx-build warning.

### DIFF
--- a/dev/generate-kerneldocs.py
+++ b/dev/generate-kerneldocs.py
@@ -21,7 +21,7 @@ A second implementation, ``libawkward-cuda-kernels.so``, is provided as a separa
 
     <img src="../_static/awkward-1-0-layers.svg" style="max-width: 500px; margin-left: auto; margin-right: auto;">
 
-The functions are implemented in C with templates for integer specializations (cpu-kernels) and as CUDA (cuda-kernels), but the function signatures and normative definitions are expressed below using a subset of the Python language. These normative definitions are used as a stable and easy-to-read standard that both implementations must reproduce in tests, regardless of how they are optimized.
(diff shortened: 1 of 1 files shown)


jpivarski (commented):
I didn't know that I broke the documentation with this update. (I didn't see doctest failing.)

But just removing the section on Pandas compatibility isn't the right way to go: what's broken about this section? The ``:doc:`ak.to_pandas` `` should be okay because [ak.to_pandas](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_pandas.html) exists in the documentation. The external link looks okay to me, too.

How can we fix this text without removing it? (And what's wrong with it, specifically?)

reikdas:
>I didn't know that I broke the documentation with this update. (I didn't see doctest failing.)

doctest passes because the documentation is still being created. It just doesn't look right (also I have no idea why it changed with the recent commits to `master`). This is what I am trying to fix - 

image

>But just removing the section on Pandas compatibility isn't the right way to go: what's broken about this section? The :doc:`ak.to_pandas` should be okay because ak.to_pandas exists in the documentation. The external link looks okay to me, too.

I get this warning when building documentation locally - 

image

This warning doesn't have anything to do with fixing the awkward kernel documentation and I just added this in the same PR because it seemed like something that might have been removed elsewhere but not here. I'll remove the changes in `index.rst` and try and fix all sphinx-build warnings in another PR (there are some others apart from the one about `ak.to_pandas`).

jpivarski:
ak.pandas.rst is new, but it does exist. (That's the link I sent.)

That part that you circled is wrong and should be fixed. I think your first correction (adding a carriage return) would do that.

It looks like this PR now includes only the carriage return—so I'm happy with it. (I just didn't want my text removed.)

Do you want me to merge it?

reikdas:
>Do you want me to merge it?

Yes, please.

PR #384, opened 2020-08-10
Make from_arrayset even more lazy

nsmith-:
This wraps all RecordArray child nodes with VirtualArray.

### DIFF
--- a/src/awkward1/operations/convert.py
+++ b/src/awkward1/operations/convert.py
@@ -2707,6 +2707,9 @@ def _form_to_layout(
     prefix,
     sep,
     partition_first,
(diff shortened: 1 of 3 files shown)


nsmith-:
This is ready

jpivarski:
I looked this over carefully, and I think it's good. The Form → JSON → Python → modify in place → JSON → Form sequence wouldn't be necessary if we had a good traverser for Forms, as we do Content layouts (`_ak._util.recursively_apply`), but it's fine.

So, all looks good, the tests would catch errors if there were any, and they pass, so I'm merging it. Thanks!

PR #390, opened 2020-08-14
VirtualArray has correct __record__ parameter

nsmith-:
Sliced virtual arrays could have a form if the original virtual array does, but this requires duplicating the entire getitem logic in parallel for Form classes. This is a big task, which can be put off. However, we need at least to be able to find `purelist_parameter("__record__")` to properly type the highlevel virtual arrays, so this implements just enough Form traversal to always be able to extract the `__record__` type.

In parallel (probably should have been a separate PR) numba needs a Form to type sliced VirtualArrays. Since they are materialized on entry into the numba function, we can cache the Form in ArrayGenerator when it was not previously available.

### DIFF
--- a/tests/test_0390-virtual-forms.py
+++ b/tests/test_0390-virtual-forms.py
@@ -3,74 +3,74 @@
 from __future__ import absolute_import
 from collections import defaultdict
 
+import pytest
 import awkward1
-import numba
 
 
-def test_virtual_slicearray_numba():
+def test_virtual_record():
     materialize_count = defaultdict(int)
 
     def gen(x, name):
         materialize_count[name] += 1
         return x
 
-    array = awkward1.Array([1, 2, 3, 4, 5])
-    virtual = awkward1.virtual(lambda: gen(array, "array"), length=len(array), form=array.layout.form)
-    slicearray = awkward1.Array([True, False, False, True, False])
-    slicedvirtual = virtual[slicearray]
-
-    @numba.njit
-    def dostuff(array):
-        x = 0
-        for item in array:
-            x += item
-        return x
-
-    assert slicedvirtual.layout.form.form is None
-    assert materialize_count["array"] == 0
-    assert dostuff(slicedvirtual) == 5
-    assert materialize_count["array"] == 2
-    materialize_count.clear()
-
-    x = awkward1.Array([1, 2, 3, 4, 5])
-    y = x * 10.
+    x1 = awkward1.Array([1, 2, 3, 4, 5])
+    x2 = awkward1.Array([1, 2, 3, 4, 5])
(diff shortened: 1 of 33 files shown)


nsmith-:
I don't understand the windows errors, is it using a different c++ standard?

jpivarski:
The first error is

https://github.com/scikit-hep/awkward-1.0/blob/e4186d058e55f0b2bd5747eb1546e694e4258d21/src/libawkward/array/VirtualArray.cpp#L545

I would say that a `not` operator doesn't exist in C++, but maybe it does in some recent standard. (I don't know how the non-Windows builds passed.) Anyway, the standard is C++11. Negation is `!`.

jpivarski (approved):
I read all the code changes. It looks like this solves all the problems we talked about on Slack—right? If you're slicing (creating a SliceGenerator), you put the `__record__` parameter on the new VirtualArray, so it will get it from there instead of trying to descend the Form that SliceGenerator doesn't have... That sounds like a great solution to the problem!

The tests pass—I assume you're done and I should merge this? (There aren't any new tests for the new feature, but you'll be testing that in the context of your project.)

nsmith-:
It did, but what do you know, I found another case where the lack of a proper form for purely field slices in SliceGenerator is causing usability issues: passing a subarray to a numba function.
Given that this implementation is most of the way to a proper implementation of virtual `getitem_field`, maybe we should just finish it.

jpivarski:
Okay, I'll leave this open. It won't be hard to merge with any upcoming changes.

tests/test_0390-virtual-forms.py
```diff
@@ -0,0 +1,108 @@
...
+    assert dostuff(virtual[["x"]]) == 15
+    assert dict(materialize_count) == {"x": 1, "array": 2}
+    materialize_count.clear()
+
+    assert dostuff(virtual[::2]) == 9
+    assert dict(materialize_count) == {"x": 1, "array": 2}
+    materialize_count.clear()
+
+    assert dostuff(virtual[:3]) == 6
+    assert dict(materialize_count) == {"x": 1, "array": 1}
```
nsmith-, line 102: I can't say I understand all these materialization passes and why they differ, but since a cache would obviate any concerns about them I'm inclined to leave it be. Someday an update might change these numbers and make these tests fail unnecessarily, but maybe it will be good to draw attention anyway.

jpivarski: I agree.

It could be that the number of materializations is multiplied by Numba's type-checking passes. Numba runs our "typer" functions a surprising number of times. At least 3 for each type (part of which is checking const and non-const types...).

But anyway, that's why we have a cache.

nsmith-:
I think this is done

PR #397, opened 2020-08-18
Awkward data should be registered as Numba constants so that they can be passed as closures to Numba-compiled functions.

### LINKED ISSUE #175: Bytestrings in Numba
Encoded strings materialize as a "unicode" type in Numba, but bytestrings are uninterpreted. This is an oversight: bytestrings should get the same kind of treatment as encoded strings.

### LINKED ISSUE #291: Awkward-Numba's ArrayView needs to support enumerate
This fails for `events` as an ak.Array:

```python
@nb.njit
def f(events):
    for index, event in enumerate(events):
        pass
```

but it doesn't fail for NumPy arrays. That's because Numba's lowered enumerate doesn't accept any iterable (or ak.Arrays aren't recognized as such because I haven't registered them somewhere? They do have `iter` overloaded). Presumably, the right way to fix this is to specialize an implementation for `enumerate`, just as an implementation for `len` has been specialized.

For completeness, the error message is

```
TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Invalid use of Function(<class 'enumerate'>) with argument(s) of type(s): (awkward1.ArrayView(awkward1.RecordArrayType((awkward1.NumpyArrayType(array(int32, 1d, A), none, {}), awkward1.NumpyArrayType(array(uint32, 1d, A), none, {}), awkward1.NumpyArrayType(array(uint64, 1d, A), none, {}), awkward1.RecordArrayType((awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(float32, 1d, A), none, {})), (('x', 'y', 'z')), none, {}), awkward1.ListArrayType(array(int64, 1d, C), awkward1.RecordArrayType((awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(int32, 1d, A), none, {}), awkward1.NumpyArrayType(array(float32, 1d, A), none, {}), awkward1.NumpyArrayType(array(bool, 1d, A), none, {})), (('pt', 'eta', 'phi', 'mass', 'charge', 'pfRelIso04_all', 'tightId')), none, {}), none, {})), (('run', 'luminosityBlock', 'event', 'PV', 'muons')), none, {}), None, ()))
 * parameterized
In definition 0:
    All templates rejected with literals.
In definition 1:
    All templates rejected without literals.
This error is usually caused by passing an argument of a type that is unsupported by the named function.
[1] During: resolving callee type: Function(<class 'enumerate'>)
[2] During: typing of call at <ipython-input-158-85980f709330> (3)
```

but the key thing is that the `enumerate` function doesn't recognize `ArrayView` as an argument. (`ArrayView` is qualified by a complex content type in this case.)

I'm labeling this as a "good first issue" because it's a straightforward, self-contained project within Awkward-Numba lowering. However, lowering Numba function is itself not for the faint of heart—to clarify, it's a good first issue for developers with background in or interest in learning Numba internals.

### LINKED ISSUE #341: Numba "cannot lower constant of type awkward1.ArrayBuilderType(None)": can anything be done?
I would want it to take the ArrayBuilder as part of a closure, not a constant. Maybe ask about this on the Numba Gitter.

jpivarski:


jpivarski:
Not "as closures," but "in closures."

jpivarski:
The `Sequence` base class was removed from `ak.Array` and `ak.ArrayBuilder` in cae8c26e50c252c48bc5342fa0e7365ab0e994af because we don't want `__contains__`, `index`, and `count` (see [abstract containers documentation](https://docs.python.org/3/library/collections.abc.html#collections-abstract-base-classes)). But (consulting the same documentation), we do want `Iterable` and `Sized`. The next commit adds them to the high-level `ak.Array`, `ak.ArrayBuilder` and to Numba-lowered `ak.Array`. (`ak.ArrayBuilder` can't be read in a Numba-lowered function.)

PR #398, opened 2020-08-18
Truncate generated outparams only to required length

reikdas:


jpivarski (approved):
Great! I checkout out this branch, ran everything (no errors), and spot-checked a few of them. The output arrays look like they're the right lengths now.

Let me know before the tests finish if you _don't_ want me to merge this as-is. Thanks!

Issue #400, opened 2020-08-19
Writing a non-minimal categorical/dictionary to Parquet corrupts it


jpivarski:
Moving @drahnreb's comment to a new issue, because it's unrelated to Parquet file partitioning. Original text below:

I found a rather weird behaviour that seems to be linked to partitioned parquets and my suggested "workaround".
> ```python
> # current workaround with pq directly
> table = pq.read_table(parquet_dir)
> ak.from_arrow(table)
> ```

As this is not a supported feature, I did not want to open a bug report but figured it's worth sharing as it took me a while to figure out that the above seems to cause the following issue.
Previously loaded partitioned data and slicing it (on a low level) will result in a `ValueError: in IndexedArray32 attempting to get xxx, index[i] >= len(content)` when saving to parquet and reloading.

Working Example:
A bit more extensive that explains the principles of what I am trying to achieve, as I realized the combination of certain aspects matter esp. details about the containing dtypes and conversions.
I couldn't figure out what is happening. I first suspected wrong advanced slicing (#370) or the type conversion awk-parquet-arrow as described [here](https://awkward-array.org/how-to-convert-arrow.html) or (#393).
It was difficult to reproduce the results - so it might not be really a minimal example. Sorry for that!

```python
# let's create some fake data that will be saved as parquet
df = pd.DataFrame([{"id": n, "label": np.random.choice(["OK", "NOK"], 1)[0], "year": np.random.choice([2019, 2020], 1), "arr": np.random.rand(np.random.randint(0,100))} for n in range(10000)])
# cast label
df.label = df.label.astype("category")
df.year = df.year.astype("uint32")
# index
df.set_index(['id'], inplace=True)
# save as partioned parquet
df.to_parquet('test.parquet',
              partition_cols=['year'], # this seems to be crucial
              version='2.0',
              data_page_version='2.0')
# let's get started. reload
table = pq.read_table('t.parquet')
# load as awkward array
original = ak.from_arrow(table)
```
```python
>>> original.layout
<RecordArray>
    <field index="0" key="label">
        <IndexedArray64>
            <index><Index64 i="[0 0 0 1 0 ... 19 19 19 18 18]" offset="0" length="20026" at="0x55d32435ff10"/></index>
            <content><ListArray64>
                <parameters>
                    <param key="__array__">"string"</param>
                </parameters>
[...]

PR #403, opened 2020-08-19
Introduce a 'categorical' type (behavioral, just as 'string' is) that is the only thing that passes to Arrow as DictionaryArray.

jpivarski:
As discussed in #400.

Don't forget:

  - [x] Better string representation of categorical types.
  - [x] Zip without an explicit `depth_limit` should not break up strings.
  - [x] Arrays from Arrow without a bitmask must be mapped to `UnmaskedArray` to maintain type.
  - [x] Apply this to Arrow conversion.

### DIFF
--- a/src/awkward1/behaviors/categorical.py
+++ b/src/awkward1/behaviors/categorical.py
@@ -62,25 +62,22 @@ def _hashable(obj):
         return obj
 
 
-_all_indexedtypes = (
-    awkward1.layout.IndexedOptionArray32,
-    awkward1.layout.IndexedOptionArray64,
-) + awkward1._util.indexedtypes
-
-
 def _categorical_equal(one, two):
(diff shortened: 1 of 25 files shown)


src/awkward1/behaviors/categorical.py
```diff
@@ -110,11 +110,109 @@ def _categorical_equal(one, two):
...
+        ),
+        getfunction,
+    )
+
+    if output[0] is None:
+        return None
+    elif highlevel:
+        return awkward1._util.wrap(output[0], awkward1._util.behaviorof(array))
+    else:
+        return out
```
drahnreb, line 168:
```suggestion
        return output[0]
```

jpivarski: Ouch! I added this directly to master because it is absolutely right.

PR #405, opened 2020-08-20
Introduce ak.from_pandas

drahnreb:


jpivarski:
You're quicker than me: instead of opening an issue, this PR can serve as the place to discuss. (Better to have only one place, anyway.)

> I'm interested in writing `ak.from_pandas`, kindly guide me through.

Cool! The `ak.to_pandas` function is here:

https://github.com/scikit-hep/awkward-1.0/blob/83232704d7cd5770804f3ece75559cd3de59eaf3/src/awkward1/operations/convert.py#L3467-L3685

though some of the issues are different: Awkward → Pandas requires, in general, multiple DataFrames to be lossless, and so there's a lot of logic there to determine which DataFrame each column should go in to make a minimal set of DataFrames. For Pandas → Awkward, only one DataFrame is ever involved. Converting non-trivial indexes to Awkward structures is more straightforward than the reverse.

The best way to start would be to open a feature-request issue and propose some translation rules with examples: "If given a DataFrame like X, we should make an Awkward Array like Y," which we can talk over before implementation. I have some ideas to start:

   * Converting Pandas column types to Awkward types is a completely independent problem from handling indexes.
   * If the DataFrame has a RangeIndex, the columns can simply be wrapped by a [RecordArray](https://awkward-array.readthedocs.io/en/latest/ak.layout.RecordArray.html).
   * If the DataFrame has an Int64Index, the above should be wrapped in an [IndexedOptionArray](https://awkward-array.readthedocs.io/en/latest/ak.layout.IndexedOptionArray.html) to arrange the order such that `ak_array[i] == df.loc[i]`, and any missing values can be None (`-1` in the `index`).
   * If the DataFrame has a MultiIndex, the above can be wrapped in layers of [ListOffsetArray](https://awkward-array.readthedocs.io/en/latest/ak.layout.ListOffsetArray.html), one for each MultiIndex level.
   * String-valued and datetime-valued indexes would simply have to become columns, since Awkward arrays can't be indexed by anything other than integers.

The complexity of this project is progressive; if you find you need to stop after RangeIndex or something, that would already be valuable and completing the index types would be a matter of layering on top of what you've done.

Thanks!

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward-1.0/issues/400#issuecomment-677829921_

jpivarski:
Let me know when you get a chance to think about this. It's okay if it has to wait.

Thanks!

drahnreb:
Sorry, I haven‘t gotten to it yet. I only looked into your ideas and compiled a few rough examples.
I‘ll update here when I have a more complete list of proposals as suggested. I hope within the next few days.

jpivarski:
Hi @drahnreb! It has been a while since you started this PR, and pretty soon, I'll be applying a name change to the whole Awkward Array codebase: all instances of "`awkward1`" will become "`awkward`" ([see details](https://github.com/scikit-hep/awkward-1.0/wiki/Name-change-procedure)). Since there are no changes here, perhaps you want to close this now and reopen a new PR after the name change? That name change is scheduled for Nov 30/Dec 1.

jpivarski:
@drahnreb, feel free to reopen this, but merge to master as a first step because a lot has changed! Or equivalently, open a new PR.

drahnreb:
> @drahnreb, feel free to reopen this, but merge to master as a first step because a lot has changed! Or equivalently, open a new PR.

Sorry for the late reply. Will re-/open a new PR after your transition and with concrete translation rules and examples.

drahnreb:
I haven't found enough time for this.
If anyone stumbles over it and really needs to do this in-mem op, this one-liner might get you started:

```python
import pandas as pd
import awkward as ak
import pyarrow as pa

ak.from_pandas = lambda df, columns=None, preserve_index=False: ak.from_arrow(pa.Table.from_pandas(df, preserve_index=preserve_index, columns=columns))

a_df = pd.DataFrame([{'one':1, 'two':2}])
awk_arr = ak.from_pandas(a_df)
```

This mostly suits my rare pandas to awkward transitions. It is one of my few other awkward "hacky extensions" (kind of a cheat sheet)...

PR #406, opened 2020-08-21
Handle some cases where C for loop was translated to Python while loop

reikdas:
Also fixes single kernel spec printing.

This should make the Python -> CUDA generation a little easier.


jpivarski:
This seems like a safe change and I'd accept it, but all the MacOS builds are failing? That doesn't sound like it could be related. I'm going to launch a new run of tests on master to see if this is a general problem.

jpivarski:
MacOS failed because skbiuld (Scikit-Build) is not installed. We don't require it/don't use it. It was a thought while we were setting things up, but it didn't work for us, so we didn't use it. Does this seem familiar to you? (Is it in this branch and I just didn't see it or something?)

reikdas:
I didn't do anything with `skbuild` in this branch. If tests on master are passing, but not here, it is weird..

jpivarski:
No, the tests on master didn't pass. In fact, I manually ran two tests on master of exactly the same commit (the one that says, "Fix typo identified..."): yesterday at 10:22am and today at 7:19am (my time). The second one failed in MacOS. Something external has changed.

jpivarski:
It fails when installing the pip package for [cmake](https://pypi.org/project/cmake/), which expects [scikit-build](https://pypi.org/project/scikit-build/) to be present, and it isn't (on MacOS, at least). I think the cmake pip maintainers may be missing a dependency in their requirements list.

I'll try adding it explicitly and then post a bug report if that turns out to be it.

jpivarski:
It might be related to this:

https://github.com/scikit-build/cmake-python-distributions/issues/103

but that was months ago. Whatever happened to us happened in the last 21 hours.

jpivarski (approved):

jpivarski:
Nope. That didn't do it.

jpivarski:
@henryiii helped me find it on Slack: the latest cmake package is broken (no wheel for MacOS Python 3). For now, at least, I put in a requirement excluding cmake 3.18.2 (we get 3.18.0 instead) and the builds are passing. When this branch passes, I'll merge it. Meanwhile, I'll keep an eye on cmake and remove our kludge when they're ready.

jpivarski:
Bah! I forgot to pull the change that fixed this in from master. One moment.

henryiii:
`PIP_ONLY_BINARY="cmake"` will fix this properly.

henryiii:
(Thanks to @YannickJadoul who originally pointed me at that tip)

jpivarski:
We should move any further discussion of the cmake issues to #407, where I'm trying the `PIP_ONLY_BINARY` method in c21c55a6701c9c04c0936d72da18acfe5a8533ff .

jpivarski:
This PR (about documentation) is going to pass all of its tests soon and I'll be merging it.

jpivarski:
(Incidentally, the environment variable method in #407 worked. I'll be merging that, too, so we'll have a proper way of dealing with the incomplete cmake deployment.)

PR #412, opened 2020-08-22
Add const info to kernel spec

reikdas:


jpivarski (approved):

jpivarski:
I had to work with generate-kernels.py and generate-tests.py last night when I added some kernels (missing cases in argmin/argmax). I hadn't noticed before how deeply nested the code is. The rule of breaking things up into smaller functions is especially true in Python because of the intention-is-nesting. It can be very hard to know what depth you're in if it gets to be more than 5.

reikdas:
Actually breaking up some of these into smaller functions shouldn't be too hard. I'll add that to my TODO list (for a later PR).

PR #414, opened 2020-08-24
Restructure test locations

reikdas:
tests-cuda-kernels will be a new directory in the future


jpivarski (approved):
It's fine to move this to dev, though the tests aren't passing yet.

reikdas:
I had edited the CI for just macOS and not Linux and Windows, causing the tests to fail. All the tests pass now.

PR #373, opened 2020-08-06
Generate CUDA kernels from kernel specification

reikdas:
The code generator will improve with each new kernel that it can handle; and this PR will likely be a rather long one.

PR #421, opened 2020-08-28
Support complex numbers

### LINKED ISSUE #392: Complex numbers (in ak.from_iter and elsewhere)
Probably low priority, as ``awkward1.from_numpy`` does.

```python
>>> import awkward1 as ak
>>> import numpy as np

>>> ak.from_iter([0j])

~/venv/awkward/lib/python3.6/site-packages/awkward1/operations/convert.py in from_iter(iterable, highlevel, behavior, allow_record, initial, resize)
    359     out = awkward1.layout.ArrayBuilder(initial=initial, resize=resize)
    360     for x in iterable:
--> 361         out.fromiter(x)
    362     layout = out.snapshot()
    363     if highlevel:

ValueError: cannot convert 0j (type complex) to an array element

>>> ak.from_numpy(np.asarray([0j]))
 <Array [0j] type='1 * complex128'>
```

sjperkins:
Closes #392

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,11 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+
+[packages]
+
+[requires]
+python_version = "3.8"
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -1,0 +1,20 @@
+{
+    "_meta": {
+        "hash": {
+            "sha256": "7f7606f08e0544d8d012ef4d097dabdd6df6843a28793eb6551245d4b2db4242"
+        },
+        "pipfile-spec": 6,
+        "requires": {
+            "python_version": "3.8"
+        },
+        "sources": [
+            {
+                "name": "pypi",
+                "url": "https://pypi.org/simple",
+                "verify_ssl": true
+            }
+        ]
+    },
+    "default": {},
+    "develop": {}
+}
--- a/include/awkward/array/NumpyArray.h
+++ b/include/awkward/array/NumpyArray.h
@@ -744,6 +744,12 @@ namespace awkward {
   void
     tojson_real(ToJson& builder, bool include_beginendlist) const;
 
+  /// @brief Internal function to fill JSON with complex floating-point values.
+  template <typename T>
+  void
+    tojson_complex(ToJson& builder, bool include_beginendlist) const;
+
+
   /// @brief Internal function to fill JSON with string values.
   void
     tojson_string(ToJson& builder, bool include_beginendlist) const;
--- a/include/awkward/common.h
+++ b/include/awkward/common.h
@@ -47,6 +47,7 @@
 
 #include <iostream>
 #include <algorithm>
+#include <complex>
 #include <map>
 #include <vector>
 #include <mutex>
--- a/include/awkward/kernels/operations.h
+++ b/include/awkward/kernels/operations.h
@@ -836,36 +836,6 @@ extern "C" {
   /// @param fromptr inparam role: NumpyArray-ptr
   /// @param length inparam
   EXPORT_SYMBOL struct Error
-  awkward_NumpyArray_fill_tofloat64_fromcomplex64(
-    double * toptr,
-    int64_t tooffset,
-    const std::complex<float> * fromptr,
-    int64_t length);
-  /// @param toptr outparam
-  /// @param tooffset inparam role: IndexedArray-index-offset
-  /// @param fromptr inparam role: NumpyArray-ptr
-  /// @param length inparam
-  EXPORT_SYMBOL struct Error
-  awkward_NumpyArray_fill_tofloat64_fromcomplex128(
-    double * toptr,
-    int64_t tooffset,
-    const std::complex<double> * fromptr,
-    int64_t length);
-  /// @param toptr outparam
-  /// @param tooffset inparam role: IndexedArray-index-offset
-  /// @param fromptr inparam role: NumpyArray-ptr
-  /// @param length inparam
-  EXPORT_SYMBOL struct Error
-  awkward_NumpyArray_fill_tocomplex64_fromcomplex128(
-    std::complex<float> * toptr,
-    int64_t tooffset,
-    const std::complex<double> * fromptr,
-    int64_t length);
-  /// @param toptr outparam
-  /// @param tooffset inparam role: IndexedArray-index-offset
-  /// @param fromptr inparam role: NumpyArray-ptr
-  /// @param length inparam
-  EXPORT_SYMBOL struct Error
     awkward_NumpyArray_fill_tobool_fromint16(
       bool* toptr,
       int64_t tooffset,
@@ -2163,20 +2133,20 @@ extern "C" {
   /// @param fromptr inparam role: NumpyArray-ptr
   /// @param length inparam
   EXPORT_SYMBOL struct Error
-  awkward_NumpyArray_fill_tocomplex64_fromcomplex128(
-    std::complex<float> * toptr,
+  awkward_NumpyArray_fill_tofloat64_fromcomplex128(
+    double * toptr,
     int64_t tooffset,
-    const std::complex<double>* fromptr,
+    const std::complex<double> * fromptr,
     int64_t length);
   /// @param toptr outparam
   /// @param tooffset inparam role: IndexedArray-index-offset
   /// @param fromptr inparam role: NumpyArray-ptr
   /// @param length inparam
   EXPORT_SYMBOL struct Error
-  awkward_NumpyArray_fill_tocomplex128_fromcomplex64(
-    std::complex<double> * toptr,
+  awkward_NumpyArray_fill_tocomplex64_fromcomplex128(
+    std::complex<float> * toptr,
     int64_t tooffset,
-    const std::complex<float>* fromptr,
+    const std::complex<double>* fromptr,
     int64_t length);
   /// @param toptr outparam
   /// @param tooffset inparam role: IndexedArray-index-offset
--- a/include/awkward/kernels/operations.h
+++ b/include/awkward/kernels/operations.h
@@ -826,6 +826,26 @@ extern "C" {
   /// @param fromptr inparam role: NumpyArray-ptr
   /// @param length inparam
   EXPORT_SYMBOL struct Error
+  awkward_NumpyArray_fill_tofloat64_fromcomplex128(
+    double * toptr,
+    int64_t tooffset,
+    const std::complex<double> * fromptr,
+    int64_t length);
+  /// @param toptr outparam
+  /// @param tooffset inparam role: IndexedArray-index-offset
+  /// @param fromptr inparam role: NumpyArray-ptr
+  /// @param length inparam
+  EXPORT_SYMBOL struct Error
+  awkward_NumpyArray_fill_tocomplex64_fromcomplex128(
+    std::complex<float> * toptr,
+    int64_t tooffset,
+    const std::complex<double> * fromptr,
+    int64_t length);
+  /// @param toptr outparam
+  /// @param tooffset inparam role: IndexedArray-index-offset
+  /// @param fromptr inparam role: NumpyArray-ptr
+  /// @param length inparam
+  EXPORT_SYMBOL struct Error
   awkward_NumpyArray_fill_tocomplex128_fromint8(
     std::complex<double> * toptr,
     int64_t tooffset,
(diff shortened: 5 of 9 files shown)


jpivarski:
Wow, this is great! I just went over all the files, and everything is in the right place. (It's a lot of boilerplate, I know, but you're not missing anything.) I think you didn't mean to check in the "Pipfile" and its lock, though.

Just to keep track, the two main features that are missing are reducers (some don't make sense for complex, such as min and max) and a Builder. I think when we talked earlier, I didn't mention that the ArrayBuilder should become aware of complex numbers—I was reminded of it by your test. The append method is capable of adding types to an ArrayBuilder that the ArrayBuilder doesn't know about (float32 in the previous example), but something as distinct as complex numbers should have its own method, so you can pass in Python complex objects, rather than having to make Awkward arrays and pass them in. (That would enable complex numbers anywhere in a list of lists passed in through `ak.from_iter`.) Unless you want to try it, I can add a Complex128Builder after this PR.

jpivarski:
I was just looking for a standard way to encode complex numbers in JSON, and there don't seem to be any protocols out there. This might be the closest: math.js makes objects with two keys: `"re"` and `"im"`.

https://mathjs.org/docs/datatypes/complex_numbers.html

sjperkins:
Thanks for taking a look. Just note that this isn't ready (I've marked the PR as a draft).

The Pipfile should be removed from the final version, I actually committed it because I wanted to do a quick `git clean -fxd`. Guess I could add it to .gitignore.

> (It's a lot of boilerplate, I know, but you're not missing anything).

I was wondering if much of this could be stamped out with macros? I didn't bite the bullet.

Also I need to cast to double and float in the C layer as you mentioned at the end of your comment in https://github.com/scikit-hep/awkward-1.0/issues/392#issuecomment-674110892.


> Just to keep track, the two main features that are missing are reducers (some don't make sense for complex, such as min and max) and a Builder. 

OK will have a look. 

> I think when we talked earlier, I didn't mention that the ArrayBuilder should become aware of complex numbers—I was reminded of it by your test. The append method is capable of adding types to an ArrayBuilder that the ArrayBuilder doesn't know about (float32 in the previous example), but something as distinct as complex numbers should have its own method, so you can pass in Python complex objects, rather than having to make Awkward arrays and pass them in. (That would enable complex numbers anywhere in a list of lists passed in through ak.from_iter.) Unless you want to try it, I can add a Complex128Builder after this PR.

Thanks for the explanation. As I understand it, this would be something like the existing Float64Builder class that the ArrayBuilder would then use when calling it's snapshot method?

Let me see how i go and if it's too much I'll ask for your assistance.

jpivarski:
> > (It's a lot of boilerplate, I know, but you're not missing anything).
> 
> I was wondering if much of this could be stamped out with macros?

Some of it probably will, either C preprocessor macros or the code-generation (Python scripts generating kernels from YAML) that @reikdas is developing.

> Thanks for the explanation. As I understand it, this would be something like the existing Float64Builder class that the ArrayBuilder would then use when calling it's snapshot method?
> 
> Let me see how i go and if it's too much I'll ask for your assistance.

It would be like Float64Builder. The tricky thing about adding a new Builder class and fill method—`Builder::complex`, in analogy with `Builder::real`, etc.—is that an appropriate action has to be assigned to every Builder class for the new method. Each Builder class has a "natural" method: for example, `Float64Builder::real` simply appends to its internal list, but the rest require structures to change. Most "non-natural" methods replace the current node with a UnionBuilder with itself nested in the UnionBuilder, and then the UnionBuilder adds new types to the union if they don't match any existing ones. However, Float64Builder has a special rule for generalizing integers unlike any of the others: it replaces the integers with floating-point numbers, then fills the next real number. The `Builder::complex` methods should have a similar behavior: they should convert integers and floating-point numbers into complex numbers, then continue filling, instead of replacing the current node with a union for these types.

Handling Complex128Builder should probably be a separate pull request, whether you do it or I do it. Actually, it doesn't even need for this PR to be finished—Builders are sufficiently independent of Contents that a Complex128Builder could be fully implemented and tested without NumpyArrays being complex-aware. (`ak.to_list` wouldn't work in the tests, but the correctness of the output could be tested by deconstructing the layout.) The offer stands: if you want to concentrate on the NumpyArray part, I can add a Complex128Builder in a separate PR, which you could use for tests when you get to that point. But if you want to try it, go ahead, but put it in a separate PR to avoid scope creep.

(Also, do you want to be added to the developers list so that you can work on this PR as a scikit-hep/awkward-1.0 branch, rather than in your own fork of the repo?)

sjperkins:
Thanks, I'll take you up on the Complex128Builder!

I tried to add a complex number overload for `awkward_NumpyArray_fill` but it's not picking it up. 

https://github.com/scikit-hep/awkward-1.0/pull/421/commits/a5483153dd7ed3e3ae94e3d05a2677e554d3cbf8#diff-b9071bed05afe0f918e34768fe605eefR1165-R1173

```bash
$ python localbuild.py 
cmake --build localbuild -- -j8
[  1%] Building CXX object CMakeFiles/awkward-cpu-kernels-objects.dir/src/cpu-kernels/operations.cpp.o
[ 64%] Built target awkward-objects
/home/sperkins/work/ska/code/awkward-1.0/src/cpu-kernels/operations.cpp: In instantiation of ‘Error awkward_NumpyArray_fill(TO*, int64_t, const FROM*, int64_t) [with FROM = std::complex<double>; TO = double; int64_t = long int]’:
/home/sperkins/work/ska/code/awkward-1.0/src/cpu-kernels/operations.cpp:2082:40:   required from here
/home/sperkins/work/ska/code/awkward-1.0/src/cpu-kernels/operations.cpp:1157:27: error: invalid cast from type ‘const std::complex<double>’ to type ‘double’
     toptr[tooffset + i] = (TO)fromptr[i];
                           ^~~~~~~~~~~~
CMakeFiles/awkward-cpu-kernels-objects.dir/build.make:134: recipe for target 'CMakeFiles/awkward-cpu-kernels-objects.dir/src/cpu-kernels/operations.cpp.o' failed
make[2]: *** [CMakeFiles/awkward-cpu-kernels-objects.dir/src/cpu-kernels/operations.cpp.o] Error 1
CMakeFiles/Makefile2:864: recipe for target 'CMakeFiles/awkward-cpu-kernels-objects.dir/all' failed
make[1]: *** [CMakeFiles/awkward-cpu-kernels-objects.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
  File "localbuild.py", line 86, in <module>
    check_call(["cmake", "--build", "localbuild", "--", "-j" + args.j])
  File "localbuild.py", line 55, in check_call
    return subprocess.check_call(args, env=env)
  File "/usr/lib/python3.8/subprocess.py", line 364, in check_call
subprocess.CalledProcessError: Command '['cmake', '--build', 'localbuild', '--', '-j8']' returned non-zero exit status 2.
```

This is to support the NumPy complex -> real casting where the imaginary part is discarded.

Any ideas? The following seems to work as I expect:

```cpp
#include <complex>
#include <iostream>


template <typename FROM, typename TO>
void convert(TO * to, const FROM * from)
{
    std::cout << "General Case " << *from << std::endl;
}

template <typename FROM, typename TO>
void convert(TO * to, const std::complex<FROM> * from)
{
    std::cout << "COMPLEX " << from[0].real() << " " << from[0].imag() << std::endl;
}


int main()
{
    float a = 2.0;
    float b = 3.0;
    std::complex<double> c = {1, 2};
    std::string bob = "bob";

    convert<float, float>(&a, &b);
    convert<std::complex<double>, float>(&a, &c);
    convert<std::string, float>(&a, &bob);

    return 0;
}
```

sjperkins:
> (Also, do you want to be added to the developers list so that you can work on this PR as a scikit-hep/awkward-1.0 branch, rather than in your own fork of the repo?)

Yes, this would be helpful.

sjperkins:
> I tried to add a complex number overload for awkward_NumpyArray_fill but it's not picking it up.

I think the technical term here is [partial ordering of overloaded function templates](https://en.cppreference.com/w/cpp/language/function_template).

jpivarski:
The cast to a template parameter that works for all other types doesn't seem to be working for `std::complex<double>`.

```
/home/sperkins/work/ska/code/awkward-1.0/src/cpu-kernels/operations.cpp:1157:27: error: invalid cast from type ‘const std::complex<double>’ to type ‘double’
     toptr[tooffset + i] = (TO)fromptr[i];
```

It could be that `std::complex<double>`, being an explicitly C++ type, requires [static_cast](https://stackoverflow.com/a/37497093/1623645). I've been regretting my use of C type casts—I use C++ `dynamic_cast` and `reinterpret_cast` when I want to downcast a virtual class or reinterpret the bits of an array, respectively, but I used C type casts everywhere that I wanted to convert a numeric type. I should have been using `static_cast` (and have no C type casts). I plan to sweep through the codebase at some point, replacing C type casts with `static_cast` (since I know that I _never_ used C type casts to mean anything other than `static_cast`), but it will be difficult because the C cast syntax is not searchable.

So this `(TO)fromptr[i]` really ought to be

```c++
static_cast<TO>(fromptr[i])
```

anyway. Try that.

This brings up another point I need to make now because the compiler isn't enforcing it: the [kernel functions need to have a pure C interface](https://github.com/scikit-hep/awkward-1.0/blob/master/CONTRIBUTING.md#project-organization)—C++ is not allowed. Specifically, that means that `std::complex<double>` can't be in the argument list. The argument lists could either call these types `void*` or `double*` and then do a `reinterpret_cast` on both sides of the C interface. That's how we ended up with all of this boilerplate: we want to use some C++ features, such as templating many versions of a function for different numeric types, but the interface has to be pure C so that it can be swapped for cuda-kernels and so that non-C++ libraries can be built on the same suite of kernels. So we ended up implementing the kernel functions with templates and adding the template abstraction back in with kernel-dispatch.h and .cpp. In the case of complex numbers, we'll also need to reinterpret-cast the `std::complex<double>` type on the buffer on both sides of the C interface.

I hope that all C++ compilers use a binary representation of a `std::complex<double>` array of length `N` that is the same as a `double` array of length `2*N` with the real and imaginary parts of each number interleaved. Not only is that important for passing arrays between kernels and the C++ codebase (which can be compiled by different compilers; awkward1-cuda-kernels is a separately distributed library), but it's necessary for interpreting complex NumPy buffers:

```python
>>> array = numpy.array([3.14 + 1.1j, 2.71 + 2.2j])
>>> array.view(numpy.float64)   # reinterpret_cast
array([3.14, 1.1 , 2.71, 2.2 ])
```

jpivarski:
Oh, and @all-contributors please add @sjperkins  for code

sjperkins:
Note that I deleted the scikit-hep/awkward-1.0:support-complex-numbers branch -- I mistakenly created it instead of pushing up to my fork.

I think it makes more sense to keep this PR on my fork as we have the conversation here, but I'll push branches to the main repo in future PRs.

jpivarski:
Okay—makes sense!

reikdas:
I can see CI is failing because the kernel specification generator is unable to handle some of the kernels you added. @sjperkins This is not an issue on your end and I'll try to fix this.

sjperkins:
Thanks for the headsup @reikdas. I haven't yet merged master into this branch, so I wouldn't suspect breakage changes from #373.

I haven't had a good look at the internals of #373 but it looks like it auto-generates CUDA functions. Is there some overlap with generating the boilerplate for this PR? If I could leverage #373 I'd prefer to manually adding functions is error-prone and cases are likely to be missed.

reikdas:
There is no overlap of this and #373. The problem is that the C++ parser in `dev/generate-kernelspec.py` needs to be extended to handle `std::complex<double>`. 
If the failing CI is a blocker for you to complete this PR, you could temporarily comment out these blocks - https://github.com/scikit-hep/awkward-1.0/blob/master/.ci/azure-buildtest-awkward.yml#L92-L103 (for all 3 of Windows, MacOS and Linux CI) and remove the [azure-doctest-awkward.yml](https://github.com/scikit-hep/awkward-1.0/blob/master/.ci/azure-doctest-awkward.yml) file. Once you are done with the PR, please re-enable them so we can fix the C++ parser and make sure CI passes before merging.

sjperkins:
Ah thanks for explaining this part of the build process @reikdas. I had to search a bit for link to the azure pipeline logs. I've been testing with `python localbuild.py`.

Note that I'm working on this when I have the time so this may not all happen in a quick burst.

jpivarski:
Hi @sjperkins! It has been a while since you started this PR, and pretty soon, I'll be applying a name change to the whole Awkward Array codebase: all instances of "`awkward1`" will become "`awkward`" ([see details](https://github.com/scikit-hep/awkward-1.0/wiki/Name-change-procedure)). It might be easier to close this now and reopen a new PR to reintroduce this work after the name change? That name change is scheduled for Nov 30/Dec 1.

sjperkins:
> Hi @sjperkins! It has been a while since you started this PR, and pretty soon, I'll be applying a name change to the whole Awkward Array codebase: all instances of "`awkward1`" will become "`awkward`" ([see details](https://github.com/scikit-hep/awkward-1.0/wiki/Name-change-procedure)). It might be easier to close this now and reopen a new PR to reintroduce this work after the name change? That name change is scheduled for Nov 30/Dec 1.

Hi @jpivarski. Thanks for the headsup. I think that's probably a good idea, since I'd started on some Python code to autogenerate the kernels in any case.

Issue #402, opened 2020-08-19
Make Form::getitem_field correctly mirror Content::getitem_field


jpivarski:
I closed PR #390 as-is because it solves @nsmith-'s problem, but `Form::getitem_field` is defined in a non-intuitive way.

https://github.com/scikit-hep/awkward-1.0/blob/6a3e6bed7fb5312d3190f2299abca71299b1a86c/include/awkward/Content.h#L226-L234

This issue is to clean that up and make sure that @nsmith-'s tests still work. Maybe we should add `Form::getitem_fields` as well, since it's nearly the same thing and would be expected to have the same materializing/non-materializing behavior.

This is good first issue for someone interested in how these nodes fit together and is willing to delve into C++.

PR #422, opened 2020-08-28
Forward purelist_parameter "__doc__" in lazy slices

nsmith-:
In addition to `__record__` since both are queried in every highlevel array instantiation.
This is a follow-up to #390

### DIFF
--- a/src/libawkward/array/VirtualArray.cpp
+++ b/src/libawkward/array/VirtualArray.cpp
@@ -569,6 +569,12 @@ namespace awkward {
       if (record != std::string("null")) {
         params["__record__"] = record;
       }
+      std::string doc =
+          form(false).get()->getitem_field(key)->purelist_parameter(
+              "__doc__");
+      if (doc != std::string("null")) {
+        params["__doc__"] = doc;
+      }
     }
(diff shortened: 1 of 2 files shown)


src/libawkward/array/VirtualArray.cpp
```diff
@@ -615,13 +621,8 @@ namespace awkward {
     ArrayGeneratorPtr generator = std::make_shared<SliceGenerator>(
                  form, carry.length(), shallow_copy(), slice);
     ArrayCachePtr cache(nullptr);
-    util::Parameters params(parameters_);
```
nsmith-, line 618: There is one case where the parameters were copied from the parent VirtualArray. I decided it was a non-uniformity and dropped it. We could instead always copy the parameters if that makes more sense.

jpivarski (approved):
I'm good with this. You know better than I do which parameters should be forwarded in virtual arrays because you have experience with real cases that exhibit good or bad features.

The tests just passed; thumbs up here and I'll merge it.

nsmith-:
Of course in the long run it will be better to have #402 but for now it is 👍

PR #428, opened 2020-09-02
Fix a bug in repr due to weak cache refs

nsmith-:
Because I descoped cache chaining. Reasoning: as far as I can see, either:
  a) the highlevel array has a cache property, in which case no need to chain
  just for repr; or
  b) the array does not have a cache which means either no virtual nodes are
  in it or someone has very consciously said “materialize every time”, in which
  case why should we provide a cache for repr

### DIFF
--- a/src/awkward1/highlevel.py
+++ b/src/awkward1/highlevel.py
@@ -406,16 +406,11 @@ class Array(
             return self._array._str(limit_value=limit_value)
 
         def _repr(self, limit_value=40, limit_total=85):
-            import awkward1.operations.structure
-
             suffix = _suffix(self)
             limit_value -= len(suffix)
 
-            layout = awkward1.operations.structure.with_cache(
-                self._layout, {}, chain="last", highlevel=False
-            )
             value = awkward1._util.minimally_touching_string(
-                limit_value, layout, self._behavior
+                limit_value, self._layout, self._behavior
             )
 
             try:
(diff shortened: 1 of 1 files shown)


src/awkward1/highlevel.py
```diff
@@ -406,16 +406,11 @@ def _str(self, limit_value=85):
...
-
             suffix = _suffix(self)
             limit_value -= len(suffix)
 
-            layout = awkward1.operations.structure.with_cache(
-                self._layout, {}, chain="last", highlevel=False
-            )
             value = awkward1._util.minimally_touching_string(
-                limit_value, layout, self._behavior
+                limit_value, self._layout, self._behavior
```
nsmith-: p.s. I think this `Mask` object has some bugs: it doesn't have `self._layout`, but its parent array would, right?

jpivarski: That's entirely likely. I think most of the examples with mask immediately call `__getitem__` in it, so bugs in its `__repr__` might not have been caught.

jpivarski:
Is this done? I haven't looked carefully into `awkward1.highlevel.Array.Mask._repr`, but if you need me to, I can.

There are three similar cases:

   * `awkward1.highlevel.ArrayBuilder.List._repr`
   * `awkward1.highlevel.ArrayBuilder.Record._repr`
   * `awkward1.highlevel.ArrayBuilder.Tuple._repr`

These four nested classes are all intermediate objects (`Mask` for supplying `array.mask[whatever]` and the `ArrayBuilder` ones for `with builder.list():`, etc.).

nsmith-:
Ok its done now. I patched up Mask because I think I understand what it should display, e.g.:
```
In [1]: events.mask
Out[1]: <NanoEventsArray.mask [<event 1:3:2021>, ... ] type='100000 * event'>
```

jpivarski:
Thanks! That looks great.

Now I remember why `__repr__` wanted to chain caches: so that data read by printing it to the screen doesn't have to be reread when using it in calculations. The double-read seemed wasteful. But if the `ak.Array` has a persistent cache and the `ak.Array` continues to exist, then the problem goes away. Your solution of ignoring chaining is a good one.

PR #429, opened 2020-09-02
Generate more CUDA kernels

reikdas:


reikdas:
@jpivarski We are able to generate CUDA for a lot more kernels. The ones that are left need to be handled in a different way. Maybe we should merge this?

src/cpu-kernels/reducers.cpp
```diff
@@ -1359,8 +1359,8 @@ ERROR awkward_reduce_argmin_bool_64(
   const int64_t* parents,
   int64_t lenparents,
   int64_t outlength) {
-  for (int64_t k = 0;  k < outlength;  k++) {
-    toptr[k] = -1;
+  for (int64_t i = 0; i < outlength;  i++) {
+    toptr[i] = -1;
   }
   for (int64_t i = 0;  i < lenparents;  i++) {
     int64_t parent = parents[i];
```
jpivarski: Is there a reason for changing the dummy variable in this kernel? The fact that `k` is distinct from `i` is useful for keeping track of the fact that I'm iterating over different things, though they are in different scopes and a compiler doesn't care.

jpivarski (approved):
Looks good to me. I only have a question about the change in dummy variables in the reducer kernels.

Also, I didn't actually test this in the AWS instance; I'd like positive confirmation that you did that because GitHub doesn't tell me (and we want to be as formal with that as we are with the CPU-bound CI).

reikdas:
>Also, I didn't actually test this in the AWS instance; I'd like positive confirmation that you did that because GitHub doesn't tell me (and we want to be as formal with that as we are with the CPU-bound CI).

Thanks for reminding me. I thought that kernel having trouble with different Python versions was fixed, but apparently not. I'll move it out of the list of kernels that can be generated for now.

src/cpu-kernels/reducers.cpp
reikdas: The Specification -> CUDA compiler assumes that `i` is the outermost for-loop, `j` is the one inside that and in the future, `k` is the one inside both. Should I generalise it? - It is probably not so difficult to do so.

jpivarski: It would be much better to keep that general. The names of the dummy variables are given to you by the example code; there's no reason to change them.

reikdas: This has been fixed now.

reikdas:
>I thought that kernel having trouble with different Python versions was fixed, but apparently not. I'll move it out of the list of kernels that can be generated for now.

There was a more fundamental issue - it is fixed now.

Everything works on the AWS instance.

Issue #320, opened 2020-07-06
The MacOS test has an intermittent error in ak.argsort


jpivarski:
It is usually cured by rerunning it, but an intermittent error is an error: something's wrong. The error manifests itself as a `ValueError` evaluating (I think!) one of the reducer tests. Next time, I should paste the error output here.

ianna:
@jpivarski - if this is it, it has happened on MacOS Python 3.8 as well. 
```python
>       assert awkward1.to_list(array.argsort(0, True, False)) == [
            [    3,    1,    2 ],
            [    0,    0,    0 ],
            [    1, None,    1 ],
            [    2, None, None ],
            [ None, None, None ]]

tests/test_0074-argsort-and-sort.py:122: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/Users/runner/hostedtoolcache/Python/2.7.18/x64/lib/python2.7/site-packages/awkward1/operations/convert.py:403: in to_list
    return [to_list(x) for x in array]
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

array = <IndexedOptionArray64>
    <index><Index64 i="[3 -1 9]" offset="9" length="3" ...ata="3 0 1 2 1 0 2 0 1" at="0x7fa0ecf3f2f0"/></content>
</IndexedOptionArray64>

>           return [to_list(x) for x in array]
E           ValueError: in IndexedOptionArray64 attempting to get 2, index[i] >= len(content)


>       assert awkward1.to_list(array5.sort(0, True, False)) == [
            [-4.4, -5.5, -6.6],
            [2.2, 1.1, 3.3],
            [4.4, 5.5, None],
            [5.5, None, None],
            [None, None, None]]

tests/test_0074-argsort-and-sort.py:177: 
[...]

PR #431, opened 2020-09-03
Allow ArrayCache(None) construction

nsmith-:
An omission from the previous implementation of weakrefs.
Apparently I had no cases of creating arrays that didn't have some sort of cache.


nsmith-:
All but one test passed, and it seems to be in an unrelated section of code.

jpivarski:
Issue #320 has happened again!

PR #435, opened 2020-09-05
intermittent error bugfix

### LINKED ISSUE #320: The MacOS test has an intermittent error in ak.argsort

ianna:
Addressing issue #320

### DIFF
--- a/include/awkward/Index.h
+++ b/include/awkward/Index.h
@@ -222,7 +222,7 @@ namespace awkward {
     /// @brief See #ptr_lib
     const kernel::lib ptr_lib_;
     /// @brief See #offset.
-    const int64_t offset_ = 0;
+    const int64_t offset_;
     /// @brief See #length.
     const int64_t length_;
   };
--- a/include/awkward/Index.h
+++ b/include/awkward/Index.h
@@ -75,10 +75,8 @@ namespace awkward {
 #ifdef AWKWARD_INDEX_NO_EXTERN_TEMPLATE
   LIBAWKWARD_EXPORT_SYMBOL
 #endif
-  IndexOf final : public Index {
+  IndexOf: public Index {
   public:
-    IndexOf<T>() = delete;
-
     /// @brief Creates an IndexOf from a full set of parameters.
     ///
     /// @param ptr Reference-counted pointer to the integer array buffer.
@@ -89,13 +87,13 @@ namespace awkward {
     /// buffer.
     /// @param length Number of elements in the array.
     /// @param Choose the Kernel Library for this array, default:= kernel::lib::cpu
-    explicit IndexOf<T>(const std::shared_ptr<T>& ptr,
-                        int64_t offset,
-                        int64_t length,
-                        kernel::lib ptr_lib);
+    IndexOf<T>(const std::shared_ptr<T>& ptr,
+               int64_t offset,
+               int64_t length,
+               kernel::lib ptr_lib);
 
     /// @brief Allocates a new integer array buffer with a given #length.
-    explicit IndexOf<T>(int64_t length, kernel::lib ptr_lib = kernel::lib::cpu);
+    IndexOf<T>(int64_t length, kernel::lib ptr_lib = kernel::lib::cpu);
 
     /// @brief Reference-counted pointer to the integer array buffer.
     const std::shared_ptr<T>
--- a/include/awkward/Index.h
+++ b/include/awkward/Index.h
@@ -75,8 +75,10 @@ namespace awkward {
 #ifdef AWKWARD_INDEX_NO_EXTERN_TEMPLATE
   LIBAWKWARD_EXPORT_SYMBOL
 #endif
-  IndexOf: public Index {
+  IndexOf final : public Index {
   public:
+    IndexOf<T>() = delete;
+
     /// @brief Creates an IndexOf from a full set of parameters.
     ///
     /// @param ptr Reference-counted pointer to the integer array buffer.
@@ -87,13 +89,13 @@ namespace awkward {
     /// buffer.
     /// @param length Number of elements in the array.
     /// @param Choose the Kernel Library for this array, default:= kernel::lib::cpu
-    IndexOf<T>(const std::shared_ptr<T>& ptr,
-               int64_t offset,
-               int64_t length,
-               kernel::lib ptr_lib);
+    explicit IndexOf<T>(const std::shared_ptr<T>& ptr,
+                        int64_t offset,
+                        int64_t length,
+                        kernel::lib ptr_lib);
 
     /// @brief Allocates a new integer array buffer with a given #length.
-    IndexOf<T>(int64_t length, kernel::lib ptr_lib = kernel::lib::cpu);
+    explicit IndexOf<T>(int64_t length, kernel::lib ptr_lib = kernel::lib::cpu);
 
     /// @brief Reference-counted pointer to the integer array buffer.
     const std::shared_ptr<T>
@@ -220,7 +222,7 @@ namespace awkward {
     /// @brief See #ptr_lib
     const kernel::lib ptr_lib_;
     /// @brief See #offset.
-    const int64_t offset_;
+    const int64_t offset_ = 0;
     /// @brief See #length.
     const int64_t length_;
   };
--- a/include/awkward/array/RawArray.h
+++ b/include/awkward/array/RawArray.h
@@ -40,13 +40,13 @@ namespace awkward {
   /// @class RawForm
   ///
   /// @brief Form describing RawArray.
-  class LIBAWKWARD_EXPORT_SYMBOL RawForm final : public Form {
+  class LIBAWKWARD_EXPORT_SYMBOL RawForm: public Form {
   public:
     /// @brief Creates a RawForm. See RawArray for documentation.
-    explicit RawForm(bool has_identities,
-                     const util::Parameters& parameters,
-                     const FormKey& form_key,
-                     const std::string& T)
+    RawForm(bool has_identities,
+            const util::Parameters& parameters,
+            const FormKey& form_key,
+            const std::string& T)
         : Form(has_identities, parameters, form_key)
         , T_(T) { }
 
@@ -188,7 +188,7 @@ namespace awkward {
   ///  - 8-bit, 16-bit, 32-bit, and 64-bit signed and unsigned integers
   ///  - 8-bit booleans
   template <typename T>
-  class LIBAWKWARD_EXPORT_SYMBOL RawArrayOf final : public Content {
+  class LIBAWKWARD_EXPORT_SYMBOL RawArrayOf: public Content {
   public:
     /// @brief Creates a RawArray from a full set of parameters.
     ///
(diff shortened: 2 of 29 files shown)


ianna:
The `42adfe5` build fail with:
```log
error C3861: 'assert': identifier not found
```
that is fixed in the next commit.

ianna:
@jpivarski - Unfortunately, I could not reproduce the error, not even once. So, trying to guess from the previous reports.
 
a) it is cured by re-running and b) the error message shows a non zero `offset` in an `IndexedArray` that is most likely the case of the error

The changes in this PR make sure that the `offset` is always initialised. The constructors are explicit, the explicit copy constructors are used between C++ and Python (by pybind11). Well, it could be an overkill and I should trust the compiler to do implicit copies correctly...

Please, have a look when you have time. Thanks!

include/awkward/Index.h
```diff
@@ -220,7 +222,7 @@ namespace awkward {
     /// @brief See #ptr_lib
     const kernel::lib ptr_lib_;
     /// @brief See #offset.
-    const int64_t offset_;
+    const int64_t offset_ = 0;
```
jpivarski: Is this supposed to be related to the bug? If there's any way for `offset_` to be uninitialized in an `IndexOf<T>` instance, then that itself is a bug that needs to be addressed at its source (assuming `0` is a bad idea: it must always be given when the instance is constructed).

include/awkward/array/RawArray.h
```diff
@@ -1292,9 +1292,9 @@ namespace awkward {
   private:
     const kernel::lib ptr_lib_;
     const std::shared_ptr<T> ptr_;
-    const int64_t offset_;
-    const int64_t length_;
-    const int64_t itemsize_;
+    const int64_t offset_ = 0;
+    const int64_t length_ = 0;
+    const int64_t itemsize_ = 0;
```
jpivarski: Same question as above on `IndexOf<T>`: if there's any way for `offset_`, `length_`, or `itemsize_` to be uninitialized in a `RawArrayOf<T>` instance, then that itself is a bug. We don't want to hide it by giving these fields default values.

include/awkward/array/RecordArray.h
```diff
@@ -441,7 +447,7 @@ namespace awkward {
     /// @brief See #recordlookup.
     const util::RecordLookupPtr recordlookup_;
     /// @brief See #length.
-    int64_t length_;
+    int64_t length_ = 0;
```
jpivarski: Same as above: `length_` needs to always be set by a constructor, never assumed to be `0` if that somehow gets missed.

include/awkward/array/RegularArray.h
```diff
@@ -81,7 +87,7 @@ namespace awkward {
 
   private:
     const FormPtr content_;
-    int64_t size_;
+    int64_t size_ = 0;
```
jpivarski: Same comment as above.

include/awkward/array/RegularArray.h
```diff
@@ -366,7 +372,7 @@ namespace awkward {
 
   private:
     const ContentPtr content_;
-    int64_t size_;
+    int64_t size_ = 0;
```
jpivarski: Same comment as above.

src/libawkward/Index.cpp
```diff
@@ -84,7 +84,6 @@ namespace awkward {
   IndexOf<T>::IndexOf(int64_t length, kernel::lib ptr_lib)
     : ptr_(kernel::malloc<T>(ptr_lib, length * (int64_t)sizeof(T)))
     , ptr_lib_(ptr_lib)
-    , offset_(0)
```
jpivarski: Why was this removed? This particular constructor should set `offset_` to `0` (because we're initializing the array right here; we know it has an offset of `0`). The others should set it to a given value.

jpivarski (commented):
The errors were quite rare—they only happened three or four times since the first report on July 6 (#320), which means hundreds of CI runs. Perhaps they could be induced on a Mac by re-running that test in a tight loop, but then again, maybe that sets up the memory in such a way that new allocations hit a lucky spot of memory 100% of the time.

I created a branch (not a PR!) to try to make the bug more likely:

https://github.com/scikit-hep/awkward-1.0/compare/master...jpivarski/intermittent_error_bugfix-DEBUG

It is relative to master, so the bug should still be there. (I didn't want to take the chance that your PR fixed it, because then we'd be chasing nothing!) This assigns `-999` to all elements of all Indexes and integers that are to be filled before filling them in `argsort_next` with the idea that some value might be left unfilled and we're looking at uninitialized memory. That's probably not it because doing so didn't cause an error on Linux, and a value of `-999` ought to cause some problems. (I also tried switching them to positive `999`, in case it's something that's just compared with zero, like in IndexedOptionArray. That didn't trigger it either, on Linux.)

If you check out this branch and run it on MacOS, does it make the bug show itself? (What about changing all the `-999` to `999`? Adding more or bigger `little_arrays` in `test_IndexedOffsetArray`? In Python 2 vs 3?) If none of that brings out the bug, then perhaps we should just apply your possible fix and keep an eye on it for the next few months.

---------------------------------------

There are a lot of changes here unrelated to the bug; they're C++ features from clang-tidy. Some of them explicitly define or delete special functions (copy/assignment constructors, etc.), some of them add `explicit` and `final` keywords, which I'm not familiar with, though I've just read up on them. (I'm not sure why we're setting `final` on Forms; the class hierarchies are very flat, most concrete classes directly inherit from a single abstract class. Unintended subclassing is not something we're likely to do.) From the documentation, these seem to be exclusively C++11 features, which is required, though I don't know if CI would catch a C++14 feature, if one were accidentally used.

I've written inline comments about initializing class members to `0`. I don't know if that's related to the clang-tidy or if that was intended to fix the bug. Either way, I'd rather not do that because it gives the impression that constructors don't have to assign them explicitly. Everything here should be 100% RAII, with all class members set by every constructor. If there's a way to enforce it with a C++11 feature, I'd be up for that.

ianna:
@jpivarski - Thanks! Using your branch I managed to reproduce and fix the bug.

Last thing to do: a newly introduced parameter in https://github.com/scikit-hep/awkward-1.0/pull/435/commits/0619e197cf10b7912ac9d740f349ff9f576b4e81 needs to be documented to fix the error:
```shell
Starting: Generate Kernel specification
==============================================================================
Task         : Command line
Description  : Run a command line script using Bash on Linux and macOS and cmd.exe on Windows
Version      : 2.164.2
Author       : Microsoft Corporation
Help         : https://docs.microsoft.com/azure/devops/pipelines/tasks/utility/command-line
==============================================================================
Generating script.
Script contents:
python dev/generate-kernelspec.py
========================== Starting Command Output ===========================
/bin/bash --noprofile --norc /Users/runner/work/_temp/9db7d3b6-b487-4f50-97f2-37657c0ebe8f.sh
  File "dev/generate-kernelspec.py", line 855, in <module>
    paramchecks = get_paramcheck(funcroles, tokens, funcargs)
  File "dev/generate-kernelspec.py", line 821, in get_paramcheck
    funcs[funcname][arg] = funcroles[keyfunc][arg]["check"]
KeyError: Token(PARAMNAME, 'nextparentslength')

##[error]Bash exited with code '1'.
Finishing: Generate Kernel specification
```

jpivarski:
> @jpivarski - Thanks! Using your branch I managed to reproduce and fix the bug.

Great! I'm glad that helped.

> Last thing to do: a newly introduced parameter in [0619e19](https://github.com/scikit-hep/awkward-1.0/commit/0619e197cf10b7912ac9d740f349ff9f576b4e81) needs to be documented to fix the error:

I think you did the right thing in 3d6ee018d120fd69ca9dde41d4c276b10009a23f, namely adding `@param` in the header file for the new parameter, but I'll ping @reikdas in case the generated kernels fail again. So far, they're looking good.

Note to self: I should add these extra steps to the `localbuild.py` script so that it's easier to do them locally. I usually

```bash
cat .ci/azure-buildtest-awkward.yml
```

and find the command lines to copy-paste to test them locally, but this is not sustainable.

--------------------

@ianna If the assignment of class members to zero is no longer necessary, I'd like to remove those changes. If there's a way to make the compiler raise a warning or an error for any constructor that doesn't set all of the class members explicitly, then let's turn on that compiler check instead. As I said in my review comments above, leaving these members uninitialized is bad, but arbitrarily setting them to zero is almost as bad.

If you think our diligence in manually checking that all constructors set all class members is not good enough and there's no automated way to check for it, then I suppose the next best thing would be to make the default values `-999` to try to ensure a runtime error. The problem with that is that the runtime error would likely be a segfault, and segfaults are hard to debug.

ianna:
@jpivarski - hopefully it fixes the bug without introducing unrelated changes :-)

Perhaps, it would be useful if the script `dev/generate-kernelspec.py` produces an error such as:

"Failed to generate documentation for parameter` 'nextparentslength'.` Have you forgotten to document it in the file `'include/awkward/kernels/sorting.h'`?"

I think, there is enough information in the script to print these details.

src/libawkward/array/IndexedArray.cpp
```diff
@@ -2215,14 +2226,15 @@ namespace awkward {
                                               stable,
                                               keepdims);
 
-    Index64 nextoutindex(index_.length());
+    Index64 nextoutindex(parents_length);
```
jpivarski, line 2229: I take it this was the significant change, right?

src/cpu-kernels/sorting.cpp
```diff
@@ -575,14 +575,14 @@ ERROR awkward_IndexedArray_local_preparenext_64(
...
+    const int64_t parentslength,
+    const int64_t* nextparents,
+    const int64_t nextlen) {
   int64_t j = 0;
   for (int64_t i = 0;  i < parentslength;  i++) {
     int64_t parent = parents[i];
     int64_t start = starts[parent];
-    int64_t nextparent = nextparents[j];
-    if (parent == nextparent) {
+    if (j < nextlen  &&  parent == nextparents[j]) {
```
jpivarski, line 585: Oh, and this must be significant, too, right?

jpivarski (approved):
Great! When the tests pass one last time, I'll merge it.

As I understand it (two comments below), the significant changes were

1. adding a `nextlen` parameter to the kernel
2. initializing `nextoutindex` with `parents_length`, rather than `index_length`.

Other than that, there was a lot of tidying and fixing of warnings. In one, a warning was fixed by casting an integer type, but I changed it from `unsigned int` to `rj::SizeType` because that's the specific type named by the RapidJSON API. At the same time, I went through and made sure we always cast to specific integer type sizes because that's a major source of bugs on Windows (which supports both 32-bit and 64-bit). That's the reason for the last CI run.

Thanks!!!

src/libawkward/array/IndexedArray.cpp
ianna: it looks like the length of parents and index are equal in all tested cases, but I changed it to be more explicit because the iteration is on parents

src/cpu-kernels/sorting.cpp
ianna: That was it! Sometimes it could happen that `nextparent` at `nextlen` is equal to `parent`. For example, the changes in the `argsort` function from your branch triggered this problem in the `sort` function in a last test:
```
parents:
0, 0, 0, 0, 0, 1, 1, 1, 1, 1, 2, 2, 2, 2, 2, 
next parents:
0, 0, 0, 0, 1, 1, 2, 2, 2, 2, 0, 3, 5, 6, 1, 
starts:
0, 0, 0, 0, 0, 5, 5, 5, 5, 5, 10, 10, 10, 10, 10, 
tocarry:
0, 1, 2, 3, -1, 4, 5, -1, -1, -1, 6, 7, 8, 9, -1, 
```

PR #436, opened 2020-09-06
Working on the Loop Dependent Kernels.

trickarcher:


trickarcher:
@jpivarski Now, that I have the automated test generation framework running for the hand written kernels, I think it is a good checkpoint to see if this is going fine, since after this a lot of those `loop-dependent-variable-kernels` and some others follow exactly the same pattern. To run the tests, you'll have to `cd studies/cuda-kernels` and then do a `./test-cuda-kernels.sh`. A review on the three kernels I have written would be nice!. And you can merge this if you want, everything is in the studies directory.

trickarcher:
@jpivarski the hand written kernels are now being tested by the automated test generator. To launch it, you'll have to `cd studies/cuda-kernels` and `./test-cuda-kernels.sh`. It would be great if you could review the three kernels I have written, since if these are satisfactory, the rest of the 26 `loop-dependent-variable-kernels` are more or less of a similar nature. Also, you can merge this if you want, everything is in the studies directory.

(I did comment a few hours back, but I can't see that comment now. Hence, I am making a new comment. Please ignore this if you've read the previous comment.)

jpivarski:
This looks good! I paused on the `cudaMalloc` for a moment because the kernels are not supposed to allocate memory, but then I realized that these have to because the intermediate scan has to go somewhere. In C++, this would be dangerous because an exception between the allocation and the deallocation would be a memory leak, but the C interface forbids us from raising exceptions from these functions, anyway, so it's okay. If there's ever an error in one of the first steps, there should be a way to skip the later steps, deallocate, and return it, but these kernels might not have errors to raise in the early steps. (How could a prefix scan ever fail?)

I'd be happy to accept this as-is. We'll integrate it into the main line of code eventually (though the basic plumbing should be established with the auto-generated kernels, first).

PR #440, opened 2020-09-09
Fix cuda shared object retrieval

reikdas:


jpivarski (approved):
This looks good!

If `awkward1_cuda_kernels` can't be imported, you'll get a `ModuleNotFoundError` when `pkg_resources` calls `__import__` on it. Especially considering that this is a developer-level tool, that's a good enough error message.

PR #448, opened 2020-09-10
Set up interface between Uproot and Awkward so that Awkward can be used to optimize object-reading.

jpivarski:
Corresponds to scikit-hep/uproot4#96.


jpivarski:
Many of these commits, starting with "Differences in parameters are checked at the 'mergeable' level..", are actually from PR #449, which I merged into this for testing. It all goes to the same master branch, anyway.

PR #449, opened 2020-09-10
Upgrade Content::merge from a single 'other' argument to a std::vector of 'others'.

jpivarski:
@ianna This might impact your work on PR #361, but the fact that only two arrays can be merged at a time is now a bottleneck (see https://github.com/scikit-hep/uproot4/pull/96#issuecomment-690622622).

Specifically, this interface:

https://github.com/scikit-hep/awkward-1.0/blob/911ba0b56429ed9ec19c42993c6a5c95a3447ff8/include/awkward/Content.h#L654-L657

is going to become:

```c++
 virtual const ContentPtr 
   merge(const ContentPtrVec& others) const = 0; 
```

(where `ContentPtrVec` is `std::vector<std::shared_ptr<Content>>`). I'll be updating this method and everything that calls it, which includes this Python:

https://github.com/scikit-hep/awkward-1.0/blob/911ba0b56429ed9ec19c42993c6a5c95a3447ff8/src/awkward1/operations/structure.py#L747-L754

If you're accessing `merge` by calling [UnionArray::simplify_uniontype](https://github.com/scikit-hep/awkward-1.0/blob/911ba0b56429ed9ec19c42993c6a5c95a3447ff8/src/libawkward/array/UnionArray.cpp#L471), then the direct interface changes may be hidden from you. On the other hand, you might be arranging loops to set things up in a pairwise way; with this fix, they could be made more efficient by loading all at once. If you're okay with that refactoring, then you could put off addressing this until the end, as a (huge) performance improvement.

The Content::mergeable interface, however,

https://github.com/scikit-hep/awkward-1.0/blob/911ba0b56429ed9ec19c42993c6a5c95a3447ff8/include/awkward/Content.h#L642-L652

will _not_ be changing. This only checks for type-compatibility and doesn't stand to benefit in performance from taking a list as an argument.


ianna:
@jpivarski - thanks for heads-up! It makes total sense.

src/libawkward/Content.cpp
```diff
@@ -891,6 +891,56 @@ namespace awkward {
...
   }
 
+  const ContentPtr
+  Content::reverse_merge(const ContentPtr& other) const {
+    throw std::runtime_error(
+      std::string("undefined operation: ") + classname() + ("::reverse_merge")
+      + FILENAME(__LINE__));
+  }
+
+  const std::pair<ContentPtrVec, ContentPtrVec>
```
ianna, line 894: @jpivarski - I often need to look it up what is `first` and what is `second` in a pair :-) 

 I wonder if using a `struct` here would be more descriptive. For example:
```c++
struct ContentMerger {
  ContentPtrVec head;
  ContentPtrVec tail;
}
```
line 940 in `src/libawkward/array/ListArray.cpp`  would become:
```c++
auto head_tail = merging_strategy(others);
ContentPtrVec head = head_tail.head;
ContentPtrVec tail = head_tail.tail;
```

jpivarski: Thanks for the suggestion! In this case, however, we're just splitting a sequence into two subsequences while maintaining order; the first is the first elements and the last is the last elements. Also, it's an internal intermediary, not a user interface (where I worry about these things more).

There are cases like `branch_depth` that may be more difficult to remember because there is no intrinsic ordering to "does this array branch? (bool)?" and "what is its depth, if it doesn't? (int)". However, the fact that they're different types lets the compiler check misinterpretations.

jpivarski:
One last thing to do: the Python ak.concatenate function should call `mergemany`, not `merge`.

PR #450, opened 2020-09-13
Explicitly cast list to CuPy array in cumsum test

reikdas:
cupy.cumsum does not cast Python list and NumPy array to CuPy array from 8.0.0rc1.
https://github.com/cupy/cupy/issues/3984


jpivarski (approved):
Easy peasy.  `:)`

PR #451, opened 2020-09-13
Change kernel specification format

reikdas:


jpivarski (approved):
Being able to say `arg["type"]` rather than `list(arg.values())[0]` is a definite improvement!

(Before, nothing ensured that the dict had only one entry, and if they come in a random order because it's Python 3.5 or below, it would have been randomly wrong.)

PR #453, opened 2020-09-17
Add more kernel test cases

reikdas:
Black seems to be inconsistent with how ti wants to format `from parser_utils import PYGEN_BLACKLIST, SUCCESS_TEST_BLACKLIST, TEST_BLACKLIST`

### DIFF
--- a/dev/generate-tests.py
+++ b/dev/generate-tests.py
@@ -8,8 +8,7 @@ from itertools import product
 
 import yaml
 from numpy import uint8
-from parser_utils import (PYGEN_BLACKLIST, SUCCESS_TEST_BLACKLIST,
-                          TEST_BLACKLIST)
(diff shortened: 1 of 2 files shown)


reikdas:
@trickarcher Maybe this PR also solves the issue about the missing tests for the kernels you mentioned earlier on slack.

trickarcher:
Yes, this fixes that issue. Thanks!

jpivarski (approved):
Great! Since this passes CI, I'll merge it _and then_ run it on my computer to see how many tests we now have. But I think it's safe to crystallize the Python definitions in specification.yml now.

Now that Uproot 4 documentation is out (https://uproot4.readthedocs.io), I plan to focus on reviewing the set of roles and samples more carefully. I can start on that without the new framework, but I'll want to wait for the "specification as source" framework before integrating it.

PR #454, opened 2020-09-17
Add some more Loop Dependent Kernels

trickarcher:


studies/cuda-kernels/loop-dependent-variable-kernels/manual_awkward_IndexedArray_getitem_nextcarry_outindex.cu
```diff
@@ -0,0 +1,149 @@
...
+    atomicMin(error_i, thread_id);
+  }
+  else if (j < 0) {
+    toindex[thread_id] = -1;
+  }
+  else {
+    tocarry[prefixedsum_mask[thread_id] - 1] = j;
+    toindex[thread_id] = (C)(prefixedsum_mask[thread_id] - 1);
+  }
+}
```
trickarcher: Hi @jpivarski , this is one of the kernels returning deterministic error messages(`error_i`). Do you have any suggestions/comments or improvements that I can do here?

trickarcher: For context `atomicCAS` acquires `error_flag`, checks if it is equal to the second argument(`i.e` 0 here) and if yes, it assigns it the value of `1`. `atomicMin` acquires `error_i` and if `thread_id` is less than `error_i`, it updates the value of `error_i` to `thread_id`.

jpivarski: This looks good. It should be possible to use `unsigned long long` and `uint64_t` interchangeably. Although the size of `long long` is machine-dependent, it will compile correctly on 64-bit machines and raise compilation errors on 32-bit machines, which is just as well, because we weren't expecting to compile this on such machines (it's better for errors to be explicit than implicit). Types in C are synonyms—it should be possible to use them interchangeably. If this doesn't work, it may be because `long long` is actually 128-bit, but I doubt it.

It's worth keeping in mind that the `atomicCAS` and the `atomicMin` are separate atomic operations: another thread could slip between them. But looking at the logic of this, that would be okay.

You might want to abstract this out into a `__device__` function that updates a `struct Error`. It will be more difficult getting the error string in: the error condition is indicated by swapping a `char *` from `nullptr` to a statically defined string. Kernels that can report more than one error string could be tricky. Maybe you need an `error_flag` for each string?

The two atomic operations could be simplified into one by initializing `error_i` to one greater than the maximum possible `thread_id` (which is always known before launching a kernel: it's the length of the array you're parallelizing over) and only doing the `atomicMin`, no `atomicCAS`. Then if the `error_i` still has its initial value after the kernel is done, there were no errors. Simplifying this to one step could make dealing with multiple error strings less complicated. Maybe then you would just have an `error_i` for each string, and checking for an error at the end would be to find the string with the minimum number (after the kernel is done), and it's no error if it's in its initialized value.

trickarcher: About the `uint64_t`, the CUDA API uses `unsigned long long` explicitly, and even if I swap in `uint64_t` it doesn't work. We could do a `typedef` in the `common.h` header like `typedef unsigned long long uint64_t` after a `static_assert` confirming that it's a 64bit type. 

Abstracting the entire thing to a `__device__` function, is something that I thought about, but I don't understand how we can implement it without atomics. We don't have a pure atomic write operation in CUDA. We can handle everything in one error variable `error_i` and then as you suggested, `if (error_i != lenindex + 1) { // return failure(args) }`, after transferring just one variable `error_i` from `device` to `host`.

jpivarski:
Hi @trickarcher! Since all of your work is in the "studies" directory, I don't think you'll be affected by the name change that's coming early next week: all instances of "`awkward1`" will become "`awkward`" ([see details](https://github.com/scikit-hep/awkward-1.0/wiki/Name-change-procedure)). I've been warning all the other PR developers about it, but I don't think your PR will have any issues.

jpivarski (approved):

cuda-build.sh
```diff
@@ -38,7 +38,7 @@ echo "cuda_version ='"$CUDA_VERSION"'" >> build/awkward_cuda_kernels/__init__.py
 echo "docker_image ='docker.io/nvidia/cuda:"$DOCKER_IMAGE_TAG"'" >> build/awkward_cuda_kernels/__init__.py
 
 export DOCKER_ARGS="-v`pwd`:/home -w/home docker.io/nvidia/cuda:"$DOCKER_IMAGE_TAG
-export BUILD_SHARED_LIBRARY="nvcc -std=c++11 -Xcompiler -fPIC -Xcompiler -DVERSION_INFO="$AWKWARD_VERSION" -Iinclude src/cuda-kernels/*.cu --shared -o build/awkward_cuda_kernels/libawkward-cuda-kernels.so"
+export BUILD_SHARED_LIBRARY="nvcc -arch=sm_35 -std=c++11 -Xcompiler -fPIC -Xcompiler -DVERSION_INFO="$AWKWARD_VERSION" -Iinclude src/cuda-kernels/*.cu --shared -o build/awkward_cuda_kernels/libawkward-cuda-kernels.so"
```
reikdas, line 41: Should we revert this change?

jpivarski: Hi! We were just on the moved Zoom call, talking about changes. This one had to be manually resolved. Why shouldn't this flag be included?

reikdas: Sorry I couldn't join. I am still in another meeting.

> This one had to be manually resolved. Why shouldn't this flag be included?

I thought we decided in an earlier meeting some months ago that we didn't want to explicitly state the `-arch` flag since we didn't get any direct benefit from specifying a lower bound of the target architecture.

src/cuda-kernels/manual_awkward_ListArray_num.cu
```diff
@@ -91,12 +80,10 @@ awkward_ListArray64_num_64(
...
-    fromstarts,
-    fromstops);
+  cuda_ListArray_num<int64_t, int64_t>
+      <<<blocks_per_grid, threads_per_block>>>(tonum, fromstarts, fromstops);
 
   cudaDeviceSynchronize();
 
   return success();
-}
+}
```
reikdas, line 89: Can we end the file with an empty line?

cuda-build.sh
jpivarski: From follow up on Slack:

> I just mentioned it since I think we talked about not including it in some earlier meeting. I don't think there is any downside to keeping it since sm_35 has been supported by NVIDIA cards for quite a few years now.

> Okay. That might become an explicit requirement. I vaguely remember that conversation—I'm still unsure whether sm_35 is already implied by the CUDA >= 9 requirement.

src/cuda-kernels/manual_awkward_ListArray_num.cu
jpivarski: It's generally good to end all files with an empty line—I hadn't noticed this one. However, I don't want to restart the tests for it, which can take a total of 20 minutes, because I'm waiting for it to do the master → main thing. We can open new PRs to fix things like this immediately afterward.

PR #455, opened 2020-09-17
Check kernel specification into git

reikdas:
No longer generate it.


reikdas:
@jpivarski I wanted to point you to this, even while it is still a draft, in case there were any changes to be made in spec.yml.

jpivarski (changes requested):
This is looking great! As I said I'd be nit-picky at this point, I have a few requests:

   * Put `description: null` immediately before `definition`. That would make it easier to deal with manually. (If it wasn't for the other requests, I'd do that one manually, since an Emacs macro would make short work of it. But you might need to regenerate this spec.yml again, so it's not time to edit spec.yml directly yet.)

   * Put an `extra-tests: []` immediately after `definition` both as a placeholder for non-role-generated tests and to help the eye see where the indentation ends, now that `description: null` is before the `definition`.

   * The argument lists are now very long. Let's reduce that by inlining their attributes:

Instead of

```yaml
- name: same
  type: List[bool]
  direction: out
- name: bytemask
  type: Const[List[int8_t]]
  direction: in
  role: ByteMaskedArray-mask
- name: missingindex
  type: Const[List[int64_t]]
  direction: in
  role: IndexedArray-index
- name: length
  type: int64_t
  direction: in
  role: default
```

change to

```yaml
- {name: same, type: "List[bool]", dir: out}
- {name: bytemask, type: "Const[List[int8_t]]", dir: in, role: ByteMaskedArray-mask}
- {name: missingindex, type: "Const[List[int64_t]]", dir: in, role: IndexedArray-index}
- {name: length, type: "int64_t", dir: in, role: default}
```

(I also shortened `direction` to `dir` because now horizontal screen space matters.)

   * I think the distinction between a kernel with just `args` and a kernel with a `specializations` is going to be cumbersome in the future (maybe already). To homogenize things, you should make every kernel take a `specializations`, though some of them would have only one specialization and its name might be equal to the kernel name.

   * The kernels with a specialization for every numeric type, such as `awkwrad_NumpyArray_fill`, are not sorted by type, though they'll be much easier to maintain if they are. This is the canonical order for numerical types:

https://github.com/scikit-hep/awkward-1.0/blob/6f40d2aa1c7d1ac674172623fb5e080a31f0dbc7/include/awkward/util.h#L28-L45

   * What had originally been a separate samples.json should be part of this monolithic YAML. The top-level of this document should be a mapping with a place for front-matter (`description: "Specification for all Awkward Array kernel functions."`). The list that this spec.yml currently represents should be under a key named `kernels`, and there should be another key named `tests` that has samples.json converted into YAML. With the files combined like that, there would be no reason to have "kernel-specification" as  a directory. Instead, it could be a single file at the top level of the repo named "kernel-specification.yml".

Thanks!

reikdas:
>What had originally been a separate samples.json should be part of this monolithic YAML. The top-level of this document should be a mapping with a place for front-matter (description: "Specification for all Awkward Array kernel functions."). The list that this spec.yml currently represents should be under a key named kernels, and there should be another key named tests that has samples.json converted into YAML. With the files combined like that, there would be no reason to have "kernel-specification" as a directory. Instead, it could be a single file at the top level of the repo named "kernel-specification.yml".

Can I do this in a different PR?

jpivarski:
> > What had originally been a separate samples.json should be part of this monolithic YAML. The top-level of this document should be a mapping with a place for front-matter (description: "Specification for all Awkward Array kernel functions."). The list that this spec.yml currently represents should be under a key named kernels, and there should be another key named tests that has samples.json converted into YAML. With the files combined like that, there would be no reason to have "kernel-specification" as a directory. Instead, it could be a single file at the top level of the repo named "kernel-specification.yml".
> 
> Can I do this in a different PR?

When this list is put inside of a structure with front-matter, it will add two spaces to every line.

Maybe do this for now:

```yaml
kernels:
  - name: first_kernel
    ...
  - name: second_kernel
    ...
```

so that the indentation is right, and we can add front matter before the `kernels:` line later. At least this way, most of the text will be final.

jpivarski:
Or we could have thousands of line changes. It's not the first time it's happened and it doesn't break git. We want to minimize it, but if it's hard to do so, it's okay.

reikdas:
>The kernels with a specialization for every numeric type, such as awkwrad_NumpyArray_fill, are not sorted by type, though they'll be much easier to maintain if they are.

For specializations that are named something like `from<some_type>to<some_type>` (eg - https://awkward-array.readthedocs.io/en/latest/_auto/kernels.html#awkward_NumpyArray_fill_toint16_fromuint32), we order by the first type in the name?

reikdas:
>Or we could have thousands of line changes. It's not the first time it's happened and it doesn't break git. We want to minimize it, but if it's hard to do so, it's okay.

I didn't realise it would involve a lot of changes. I'll just implement all of the fixes you suggested and once those are in place, I'll work on this one :D So we won't need different PRs.

jpivarski:
> > The kernels with a specialization for every numeric type, such as awkwrad_NumpyArray_fill, are not sorted by type, though they'll be much easier to maintain if they are.
> 
> For specializations that are named something like `from<some_type>to<some_type>` (eg - https://awkward-array.readthedocs.io/en/latest/_auto/kernels.html#awkward_NumpyArray_fill_toint16_fromuint32), we order by the first type in the name?

Lexicographically (as it is now in the header file, by the way). First sort by first template parameter, then sort by second. You can define a `sorted` key function by returning the two as a tuple and Python knows how to lexicographically sort tuples.

reikdas:
@jpivarski Can you take a look? If this is okay, I'll merge the samples.json into this (and delete dev/generate-kernelspec.py)

jpivarski (approved):
The only thing that I see is that awkward_NumpyArray_fill is not sorted. I checked a few others with specializations and this wasn't true for them, either. Fixing this after the PR would mean some big diff blocks, but not enormous. It could be done in a subsequent PR if that helps you finish.

Once the samples are appended at the end, the name can change to kernel-specification.yml and the directory can go away. That would be easier to manage.

Other than that, it looks great! The documentation, tests, and CUDA generation can all start from the specification in this form?

reikdas:
>The only thing that I see is that awkward_NumpyArray_fill is not sorted. I checked a few others with specializations and this wasn't true for them, either.

The sorting should be fixed now.

>The documentation, tests, and CUDA generation can all start from the specification in this form?

Yes.

If you give me the go ahead, I'll add in the samples.json to the yml file ;)

jpivarski:
> The sorting should be fixed now.

It is!

> If you give me the go ahead, I'll add in the samples.json to the yml file ;)

Yes, please!

reikdas:
@jpivarski The PR should be complete.

jpivarski:
I don't remember what we said: is this done now? Can I merge it?

reikdas:
I don't think we reached a definite conclusion about this. Maybe I should change extra-tests to manual-tests and add the automatic-tests field before we merge this in?

jpivarski:
Oh, that's right. Okay, I'll wait for you to do that, then give me an explicit go-ahead.

reikdas:
Should be done now.

jpivarski:
Great! This will make it into 0.3.0.

PR #456, opened 2020-09-17
Add numpy boolean type to ak describe

nsmith-:
Also added a more clear error message when an unknown array type is encountered, e.g. when someday someone might pass a complex-valued numpy array to `ak.type`


jpivarski (approved):
That day might be soon: #421.

PR #458, opened 2020-09-18
Change the Windows build following @chrisburr's suggestion

jpivarski:
on https://github.com/conda-forge/staged-recipes/pull/12684/files#diff-4d5ada45354b5bb597c5bb127a4ae50a

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,14 +67,15 @@ class CMakeBuild(setuptools.command.build_ext.build_ext):
                 "-DCMAKE_LIBRARY_OUTPUT_DIRECTORY_{0}={1}".format(cfg.upper(), extdir),
                 "-DCMAKE_WINDOWS_EXPORT_ALL_SYMBOLS=TRUE",
             ]
-            if sys.maxsize > 2**32:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -69,13 +69,9 @@ def build_extension(self, ext):
             ]
             if sys.maxsize > 2**32:
```
chrisburr: I think this will be enough to make both visual studio and NMake happy:
```suggestion
            if sys.maxsize > 2**32 and os.environ.get("CMAKE_GENERATOR") != "NMake Makefiles":
```

chrisburr: (in addition to the other changes you have here that is)

chrisburr (commented):
This has worked both here and on conda-forge 👍 Thanks for the help! 

(though you might want to drop the two commented out lines)

jpivarski:
I had the same thought—I'm also dropping commented out old code for the scikit-build based build, which is now old enough that swapping in these commented out lines probably wouldn't work.

I'm glad this works for the Conda build now!

PR #460, opened 2020-09-18
Remove Awkward-as-a-Pandas-column feature, as discussed in #350.

### LINKED ISSUE #350: Should Awkward Arrays be usable as Pandas columns?
This was one of the design goals described in the [original motivations document](https://docs.google.com/document/d/1lj8ARTKV1_hqGTh0W_f01S6SsmpzZAXz9qqqWnEB3j4/edit?usp=sharing), but it has required some non-intuitive sorcery to implement and it's not clear to me that it's a valuable feature. To be clear, we're talking about

```python
>>> import awkward1 as ak
>>> import pandas as pd
>>> pd.DataFrame({"awkward": ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]])})
           awkward
0  [1.1, 2.2, 3.3]
1               []
2       [4.4, 5.5]
```

and not 

```python
>>> ak.pandas.df(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
                values
entry subentry        
0     0            1.1
      1            2.2
      2            3.3
2     0            4.4
      1            5.5
```

The explicit conversion into a MultiIndex DataFrame with [ak.pandas.df](https://awkward-array.readthedocs.io/en/latest/ak.pandas.df.html) has no issues: the implementation is straightforward and I know how I would use it—there are plenty of Pandas functions for dealing with MultiIndex. For example,

```python
>>> df = ak.pandas.df(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
>>> df.unstack()
         values          
subentry      0    1    2
entry                    
0           1.1  2.2  3.3
2           4.4  5.5  NaN
```

But for the Awkward-in-Pandas, the only things I know of that can be used directly are ufuncs:

```python
>>> pd.DataFrame({"awkward": ak.Array([[1, 2, 3], [], [4, 5]])}) + 100
           awkward
0  [101, 102, 103]
1               []
2       [104, 105]
```

but not all ufuncs, for some Pandas reason:

```python
>>> np.sqrt(pd.DataFrame({"awkward": ak.Array([[1, 2, 3], [], [4, 5]])}))
  File "/home/pivarski/irishep/awkward-1.0/awkward1/highlevel.py", line 996, in __getattr__
    raise AttributeError("no field named {0}".format(repr(where)))
AttributeError: no field named 'sqrt'

TypeError: loop of ufunc does not support argument 0 of type Array which has no callable sqrt method
```

Presumably, we could narrow in on that reason and get it to work, but there are a _lot_ of Pandas functions to test. The fundamental problem is that Awkward objects are "black boxes" to Pandas. Sure, we can put them _in_ a DataFrame, but what's Pandas going to do with them once they're there?

There are other downsides to making Awkward Arrays subclasses of `pandas.core.arrays.base.ExtensionArray` (so that they can be columns). For one thing, it implies that we have to `import pandas` at startup, which can cost up to a second on slow machines or might try to import a broken installation of Pandas even if the user isn't planning on using Pandas. (If Pandas is not installed, we can change the class hierarchy, but that means `ak.Array` behaves differently, depending on whether you've installed Pandas, even if you're not using it.)

To avoid the above, the current implementation only makes `ak.Array` inherit from `pandas.core.arrays.base.ExtensionArray` if you try to use it in Pandas, which can be detected by a call to `dtype`. But for consistency, that's even worse, since the inheritance of `ak.Array` now changes at runtime, depending on whether you've ever tried to use an Awkward Array in a DataFrame. This came up in a difference in behavior (reported on Slack) that I couldn't reproduce at first because my test didn't invoke Pandas. Namely, the `pandas.core.arrays.base.ExtensionArray` defines some methods, and these methods exist or don't exist on `ak.Array` unless they're overshadowed by my own implementations. At the very least, I should overshadow all the non-underscored ones so that their existence is not history-dependent, but it fills up the `ak.Array` namespace with names I don't necessarily want.

   * `to_numpy`: This would be fine; it would call [ak.to_numpy](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_numpy.html), though the other methods don't have an underscore, such as `tolist` (for consistency with NumPy).
   * `dtype`: Already tricky, since Pandas requires a new one, `AwkwardDType`, and Dask requires `np.dtype("O")`.
   * `shape`: Pandas needs this to be one-dimensional, which is misleading for an Awkward Array. Preferably, Awkward Arrays would have _no_ `shape` at all; the combined `dtype` and `shape` can only be fully captured by [ak.type](https://awkward-array.readthedocs.io/en/latest/_auto/ak.type.html).
   * `ndim`: Much like `shape`, it's misleading for this to always be `1`.
   * `nbytes`: This is fine, and other libraries expect such a property, too.
   * `astype`: This was the surprise that triggered this issue: I didn't think Awkward Arrays had an `astype`, since it's not clear what it should mean. For changing numeric types, there's an open PR #346, but it's a new function since it doesn't change the whole type of the array, it descends to the leaves where the numbers are.
   * `isna`: This can go to [ak.is_none](https://awkward-array.readthedocs.io/en/latest/_auto/ak.is_none.html), though "na" is not how we refer to missing data.
   * `argsort`: This can go to [ak.argsort](https://awkward-array.readthedocs.io/en/latest/_auto/ak.argsort.html).
   * `fillna`: This can go to [ak.fill_none](https://awkward-array.readthedocs.io/en/latest/_auto/ak.fill_none.html), but see the note on `isna` above.
   * `dropna`: We don't have an `ak.drop_none`, but such a thing wouldn't be too hard to write.
   * `shift`: This one _only_ makes sense for rectangular tables. (See [the definition](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.shift.html).)
   * `unique`: We don't have an `ak.unique` and there could be some subtitles there. We don't have a definition for record equality, for example, and string equality is already handled through a [behavioral extension](https://awkward-array.readthedocs.io/en/latest/ak.behavior.html).
   * `searchsorted`: Only makes sense if the data are actually sorted. Should there be an `axis=1` version of this for variable-length lists? Usually, physics events are unsorted but the particles (`axis=1`) are sorted by `pT`.
   * `factorize`: This is a non-intuitive name, but it could be good to have an Awkward function that turns arrays into an [IndexedArray](https://awkward-array.readthedocs.io/en/latest/ak.layout.IndexedArray.html) of unique values. But for complex objects like records, this brings up the same issues as `unique` (above).
   * `repeat`: We don't have an `ak.repeat`, but that might be useful in some contexts. I usually find [np.repeat](https://numpy.org/doc/stable/reference/generated/numpy.repeat.html) and [np.tile](https://numpy.org/doc/stable/reference/generated/numpy.tile.html) to be a pair that have to be used together, usually to make a Cartesian product (and we already have [ak.cartesian](https://awkward-array.readthedocs.io/en/latest/_auto/ak.cartesian.html)).
   * `take`: This seems unnecessary to me, since we already have `__getitem__` with integer arrays.
   * `copy`: I don't know if we have a high-level "copy" function, but we have the low-level ones to link it up.
   * `view`: This wouldn't make much sense for an Awkward Array. It's not a simple buffer.
   * `ravel`: Maybe the equivalent of this is [ak.flatten](https://awkward-array.readthedocs.io/en/latest/_auto/ak.flatten.html)? Flattening variable-length arrays, particularly ones that include records, is a different kind of thing from flattening rectilinear data.

Given these mismatches, I'm strongly considering removing the Awkward-in-Pandas feature before Awkward1 actually becomes 1.0. The explicit conversion functions, [ak.pandas.df](https://awkward-array.readthedocs.io/en/latest/ak.pandas.df.html) and [ak.pandas.dfs](https://awkward-array.readthedocs.io/en/latest/ak.pandas.dfs.html), would be kept.

But I might be wrong—there might be some fantastic use-case for Awkward-in-Pandas that I don't know about. This question is an informal vote on the feature. You might have been sent here by an error message, where the feature is provisionally removed with a way to opt-in. If you find it useful to include Awkward Arrays inside of Pandas DataFrames (distinct from the [ak.pandas.df](https://awkward-array.readthedocs.io/en/latest/ak.pandas.df.html) conversion), then say so here, describing the use-case. You can opt-in now by calling [ak.pandas.register()](https://awkward-array.readthedocs.io/en/latest/ak.pandas.register.html), but if I don't hear from people saying that they really use it, the feature will be removed and you won't be able to use it past 1.0.

So let me know!

jpivarski:


nsmith-:
This PR removed `Array.columns`, is this intended? How do we enumerate the items of a highlevel record array in its absence?

jpivarski:
That was intended. Wasn't there a deprecation message? The proper way is

```python
ak.keys(array)
```

jpivarski:
I also forgot that `array.tojson` should be dropped in favor of `ak.to_json(array)`. The only method that stays is `array.tolist` (no space) because it's so useful, NumPy has it, and it's not going to conflict with a domain-specific method.

Is a non-method for `keys` too cumbersome? Would `array.keys()` conflict with any physics cases? `ak.Array` isn't going to be a `Mapping`, and I don't want to give that impression. `array.columns` was one of the unintended side-effects of supporting Pandas, which confused at least one person in a tutorial.

nsmith-:
On master, 
```python
>>> import awkward1
>>> a = awkward1.Array([{"x": 1, "y": 2}])
>>> a.columns
  File "/Users/ncsmith/src/awkward-1.0/awkward1/highlevel.py", line 1058, in __getattr__
    raise AttributeError(
AttributeError: no field named 'columns'

(https://github.com/scikit-hep/awkward-1.0/blob/0.3.0/src/awkward1/highlevel.py#L1060)
```
We've been using `a.columns` in coffea documentation for a while, and it was available in ak0..
Why drop this and not `['Mask', 'behavior', 'cache', 'layout', 'mask', 'nbytes', 'numba_type', 'slot0', 'slot1', 'slot2', 'slot3', 'slot4', 'slot5', 'slot6', 'slot7', 'slot8', 'slot9', 'tojson', 'tolist']` ? It seems to me the well-intentioned separation between awkward operations and behaviors is becoming overzealous.

jpivarski:
`columns` was never intended or desired. The high-level name for them everywhere is `keys`. (Although internally, I do slip into `fields`.)

`tojson` was a generalization of `tolist`, but having `tolist` as a method is far more useful.

`nbytes` is because caches use this to determine how much of the cache an object is using (not just Uproot's).

The `slot*` form a pattern. It's not really taking up 10 names from the namespace, but 1-ish.

`layout` and `behavior` were intended from the beginning, as intentional gaps in the namespace.

`numba_type` was a later addition, but unlikely to hit anyone's domain-specific data because it has a product name in it.

`mask` is specifically for the `array[cut]` → `array.mask[cut]` generalization, so it has to be a property. `Mask` is the same thing, capitalized, so the two of these are 1-ish.

`cache` was originally `metadata` so that any more such things would be hidden inside that one spot in the namespace. But you know how that turned out.

That's all of them. They all had reasons. If it's convenient to want to add `.keys()` or `.fields` as an afterthought, after typing the `array`, then I'm willing to add it in the same spirit as `array.tolist()`. You've pointed out how I'm not 100% purist about using this namespace; I'm willing to make this a method/property, but not named "columns." We don't use the word "columns" anywhere else to refer to this concept, and it is good to reign in the number of different names for the same thing.

nsmith-:
I guess now or never to change. I scanned code I have and it looks like switching `x.columns` to `ak.keys(x)` is palatable. The fact that awkward0 arrays were near enough to pandas (or Spark) dataframes with respect to this property was, I guess, an unintended feature that we've gotten used to.

PR #463, opened 2020-09-18
Generalize NumPy ufunc behavioral matching.

jpivarski:


jpivarski:
@nsmith- I'm going to keep this open until you have a chance to test it (as discussed on Slack).

We want to make this generalization, but I want you to test it on your case to be sure we're not generalizing wrong.

nsmith-:
I can confirm this works for my use case.

PR #465, opened 2020-09-19
Attempt to make check for -A x64 flag more robust

chrisburr:
As the version of the compilers used will change from time to time I think this might be a slightly more robust check to use.


jpivarski:
This certainly works in the three cases we know about: our CI and the two cases encountered in the conda-forge deployment. Since they of the strings differed in only having "Win64" at the end, perhaps this is a suffix. But "NMake Makefiles" is completely different, and might be hiding whether the bit size is fixed or not.

PR #468, opened 2020-09-21
Generate kernel header files from specification

reikdas:


jpivarski (approved):
Looks good; it can be merged as soon as it passes tests.

PR #473, opened 2020-09-25
Properly broadcast over empty ListArray

nsmith-:


nsmith-:
It looks like for ListOffsetArray, the canonical way to describe an empty array is `offsets = [0]` ? In which case we don't need to change anything for that.

src/awkward1/_util.py
```diff
@@ -821,8 +821,11 @@ def apply(inputs, depth):
                         awkward1.layout.ListArray64,
                     )):
                         starts, stops = x.starts, x.stops
-                        lencontent = nplike.max(stops)
-                        nextinputs.append(x.content[:lencontent])
+                        if len(stops) == 0:
+                            nextinputs.append(x.content[:0])
+                        else:
+                            lencontent = nplike.max(stops)
+                            nextinputs.append(x.content[:lencontent])
```
jpivarski: Oh, of course `nplike.max(stops)` fails if `stops` is empty.

However, the `stops` is allowed to be longer than the `starts`, and it's the `starts` that defines the length of the ListArray. ([Documented here](https://awkward-array.readthedocs.io/en/latest/ak.layout.ListArray.html).) So a ListArray could have length zero and still get into the else-branch of that if-statement. That could be fixed once and for all by setting

```python
stops = stops[:len(starts)]
```

right after defining `starts` and `stops`.

I can't say that would necessarily be wrong, though: you might get a `lencontent` that is larger than it needs to be...

Oh, wait! That is wrong: we're not supposed to run ufuncs on more array elements than a user can see (to avoid confusion about a ufunc failing on data that's not "there"). I tried to set up an example that would demonstrate this, but apparently it's being properly sliced somewhere else, so maybe the extra slicing doesn't have to be done here. (It doesn't hurt, though.)

```python
>>> import awkward1 as ak
>>> import numpy as np
>>> import numbers
>>> starts = ak.layout.Index64(np.array([0], dtype=np.int64))
>>> stops = ak.layout.Index64(np.array([3, 3, 5], dtype=np.int64))
>>> content = ak.layout.NumpyArray(np.array([1.1, 2.2, 3.3, 4.4, 5.5]))
>>> array = ak.Array(ak.layout.ListArray64(starts, stops, content))
>>> array
<Array [[1.1, 2.2, 3.3]] type='1 * var * float64'>
>>> def testy(x, y):
...     print("testy", x, y)
...     return ak.layout.NumpyArray(np.asarray(x.layout) + np.asarray(y.layout))
... 
>>> ak.behavior[np.add, numbers.Number, numbers.Number] = testy
>>> array + array
testy [1.1, 2.2, 3.3] [1.1, 2.2, 3.3]
<Array [[2.2, 4.4, 6.6]] type='1 * var * float64'>
```

I was thinking that this might try to add `[1.1, 2.2, 3.3, 4.4, 5.5]` instead of the properly sliced `[1.1, 2.2, 3.3]`. If not here, I don't know where that slicing is happening, but anyway, it's a good thing.

When testing this with a length-0 `starts`, I run into an error, which is similar to but (I think) not related to the one you fixed here:

```python
>>> starts = ak.layout.Index64(np.array([], dtype=np.int64))
>>> array = ak.Array(ak.layout.ListArray64(starts, stops, content))
>>> array
<Array [] type='0 * var * float64'>
>>> array + array
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/numpy/lib/mixins.py", line 21, in func
    return ufunc(self, other)
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward1/highlevel.py", line 1380, in __array_ufunc__
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward1/_connect/_numpy.py", line 107, in array_ufunc
    out = awkward1._util.broadcast_and_apply(inputs, getfunction, behavior)
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward1/_util.py", line 972, in broadcast_and_apply
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward1/_util.py", line 745, in apply
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward1/_util.py", line 752, in apply
    elif not all_same_offsets(nplike, inputs):
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward1/_util.py", line 521, in all_same_offsets
    offsets[-1] != stops[-1]
IndexError: index -1 is out of bounds for axis 0 with size 0
```

That's in

https://github.com/scikit-hep/awkward-1.0/blob/4ba42ac43707555d5847a50deb200b6f157f78bd/src/awkward1/_util.py#L519-L523

which should be

```python
                elif (
                    not nplike.array_equal(offsets[:-1], starts) or
                    (len(stops) !=0 and offsets[-1] != stops[-1])
                ):
                    return False
```

jpivarski:
> It looks like for ListOffsetArray, the canonical way to describe an empty array is `offsets = [0]` ? In which case we don't need to change anything for that.

For ListOffsetArray, `[n]` for any number `n` represents an empty array. [Documentation](https://awkward-array.readthedocs.io/en/latest/ak.layout.ListOffsetArray.html):

> The offsets must have at least length 1 (corresponding to an empty array), **but it need not start with 0** or include all of the content. Just as ak.layout.RegularArray can have unreachable content if it is not an integer multiple of size, a ListOffsetArray can have unreachable content before the first list and after the last list.

An `offsets` of, say, `[33]`, would be turned into a `starts` of `[]` and a `stops` of `[]` because they drop the last and first elements, respectively.

src/awkward1/_util.py
nsmith-: Sounds good, more checks are better. Have a test suggestion?

jpivarski: The example (above) that raises an exception shouldn't raise an exception. `array + array` should return an array with length zero.

jpivarski:
Is this good/finished? I wouldn't want this to sit unmerged because I forgot about it.

I see comments from me (answering a question), but not requests for changes.

nsmith-:
I think its done yes

PR #477, opened 2020-09-29
CPU kernel source files will have 1 file per kernel

reikdas:


dev/generate-cuda.py
```diff
@@ -26,10 +26,7 @@
     "awkward_content_reduce_zeroparents_64",
     "awkward_ListOffsetArray_reduce_global_startstop_64",
     "awkward_IndexedArray_reduce_next_fix_offsets_64",
-    "awkward_Index8_to_Index64",
-    "awkward_IndexU8_to_Index64",
-    "awkward_Index32_to_Index64",
-    "awkward_IndexU32_to_Index64",
+    "awkward_Index_to_Index64",
     "awkward_carry_arange",
```
jpivarski, line 30: This is an improvement. I think those separate kernels were defined before I got into the cadence of making such things specializations of a single kernel.

jpivarski (approved):
How does it feel to change 10,000 lines in one commit?

This is a major change that should definitely be isolated (no other features should go into this PR).

I'm going to test it on my computer as well as CI as a sanity check.

include/awkward/kernel-utils.h
```diff
@@ -99,4 +99,44 @@ extern "C" {
...
 
+template <typename T>
+void awkward_ListArray_combinations_step(
+  T** tocarry,
+  int64_t* toindex,
+  int64_t* fromindex,
+  int64_t j,
+  int64_t stop,
+  int64_t n,
+  bool replacement) {
```
jpivarski: Wait—we should avoid putting code in a .h file unless absolutely necessary. (There are some cases that are absolutely necessary.) I suppose this is here because it's templated and gets instantiated by awkward_ListArray_combinations and awkward_RegularArray_combinations (in different files, now). But, as a utility (not part of the public API to be guaranteed in future versions), it's never used without `T = int64_t`, so maybe just define one function as a utility.

reikdas:
> How does it feel to change 10,000 lines in one commit?

Haha, I actually had several arbitrary commits before squashing them into a single one.

jpivarski (changes requested):

jpivarski:
While we're moving everything around, the allocators.h should probably go into kernel-utils.h, too, because these are two files of "not regular kernels" that can get lost in the hundreds-long list. Putting all the weirdos in one place makes them easier to find.

src/cpu-kernels/awkward_RegularArray_combinations.cpp
```diff
@@ -0,0 +1,51 @@
...
+      tocarry,
+      toindex,
+      fromindex,
+      0,
+      stop,
+      n,
+      replacement);
+  }
+  return success();
+}
```
jpivarski: This function is templated by `C` and `C` is never used, not even once. It gets specialized to `int32_t` only, but that's kind of a throw-away. Are there other cases like this?

reikdas: I don't think so, but I never explicitly looked for it, so I am not a 100% sure.

jpivarski: * awkward_ListArray_getitem_next_range_counts defines `T` but doesn't use it.
   * awkward_RegularArray_combinations defines `C` but doesn't use it.

That's it! They've become a lot easier to scan.

These two things should be fixed; I can see this as being part of the refactor.

reikdas:
@jpivarski The changes you suggested should be fixed :D

jpivarski (approved):

reikdas:
@jpivarski I am pushing another commit.. I think I didn't move the allocators into kernel-utils in the proper way. Sorry!

jpivarski:
This looks good! Let me know when you're ready to merge.

I noticed that

```bash
python -m pytest -vv -rs tests-cuda
```

is now broken, but I think that can/should be a separate PR. (It's not part of the standard CI tests and nobody is using it yet. However, let's get it fixed soon!)

reikdas:
> This looks good! Let me know when you're ready to merge.

I think it can be merged now.

> I noticed that
> 
> ```shell
> python -m pytest -vv -rs tests-cuda
> ```
> 
> is now broken, but I think that can/should be a separate PR. (It's not part of the standard CI tests and nobody is using it yet. However, let's get it fixed soon!)

Yikes, I missed that. Thanks for pointing it out. I'll create a new PR to fix it.

PR #478, opened 2020-09-30
Fix broken CUDA tests

reikdas:


jpivarski (approved):
Works (locally) for me!

The standard way to test this should be with the AWS instance. Did you do that?

reikdas:
Just checked - works on the AWS instance.

PR #485, opened 2020-10-07
Update to pybind11 2.5.0

### LINKED ISSUE #483: In dependent project, ak.layout fails to convert to ak::Content. Update pybind11!
I've been working on some fastjet bindings with an awkward1 interface based on pybind11 (I'm aware of pyjet and the SWIG bindings, but for various reasons, this was a better approach for me. I'm not linking them here because they're early and ugly and not really ready for anyone else, but the repo can be found on my profile if it's somehow helpful). I have an early version working on macOS, but when I moved to linux (ubuntu 18.04 for what it's worth, but I'm working in a virtualenv), it throws a TypeError whenever I try to pass an array layout to c++. For a trivial function which just returns what it receives, on linux I get

```python
Python 3.6.9 (default, Jul 17 2020, 12:50:27)
[GCC 8.4.0] on linux
>>> import awkward1 as ak
>>> import pybind11_awkward as pba
>>> pba.awkward_test(ak.Array([1,2,3]).layout)
TypeError: awkward_test(): incompatible function arguments. The following argument types are supported:
    1. (arr: awkward::Content) -> awkward::Content

Invoked with: <NumpyArray format="l" shape="3" data="1 2 3" at="0x000001be97b0"/>
```

On macOS, I get:

```python
Python 3.8.5 (default, Jul 21 2020, 10:48:26)
[Clang 11.0.3 (clang-1103.0.32.62)] on darwin
>>> import awkward1 as ak
>>> import pybind11_awkward as pba
>>> pba.awkward_test(ak.Array([1,2,3]).layout)
In function
<NumpyArray format="l" shape="3" data="1 2 3" at="0x7ff17b283a00"/>
```

Other functions seem to work fine - it's apparently something about passing awkward arrays. I'm using awkward 0.3.1.

I'm a bit stuck on how to proceed. I recognize there's a huge phase space where I could have gone wrong, so I made a quick repository that reproduces the issue, and I tried to keep as much consistent as possible. It's available [here](https://github.com/raymondEhlers/pybind11-awkward-array-test). You can install it with `poetry install`, and then run `test_simple.py`. (In principle, it should work with pip using PEP 517, but my build script isn't quite right, sorry). I'm not 100% sure that this is actually an awkward1 issue (there are still some differences in the setups), but it seems like a reasonable place to start.

As always, help or suggestions are greatly appreciated! Thanks!

raymondEhlers:
Fixes #483


jpivarski:
That does it! This commit hash corresponds to an official release of pybind11, which is the only constraint I'd apply, other than the tests passing. So I'll take it. `:)`

jpivarski:
Hi,

I know this was 4 years ago and a minor update (upgrading pybind11, which has been upgraded many times since then), but I want to be sure to acknowledge everyone who made some contribution. Thanks!

@all-contributors please add @raymondEhlers for maintenance

PR #488, opened 2020-10-16
Add conda install instructions

lgray:
Advertise conda availability in README.

No code changes.


jpivarski:
Good point, though the command won't work without `-c conda-forge` or

```bash
conda config --add channels conda-forge
```

for users who haven't already set up conda-forge as one of their channels. (It's not one of the default ones.)

lgray:
Ah, duh, let me fix that. :-)

PR #482, opened 2020-10-06
Add Python 3.9 to tests and deployment.

jpivarski:

PR #495, opened 2020-10-21
Add a developer tool to check if kernel specification file is sorted

reikdas:


jpivarski:
I want to hold off on merging this until after #482. That involved updating the pybind11 dependency, and I got a little confused with updating submodules.

(Nice to see that this is moving forward, though!)

reikdas:
@jpivarski Ping

jpivarski:
Thanks for the reminder. I'll start by merging with master.

jpivarski:
First failure: no module "yaml" in Python 3.9. If installing a package like yaml is a problem because Python 3.9 is so new, we could restrict this test to only a few or only one task. After all, if the kernels aren't sorted, that can be checked by any one task: it's not a version-dependent statement.

By the way, are all of these still requirements?

```bash
% head -n 1000 requirements*.txt
==> requirements-dev.txt <==
PyYAML
pycparser
lark-parser
numba>=0.50.0;python_version>="3.6"
pandas>=0.24.0;python_version>="3.6"
numexpr;python_version>="3.6"
autograd;python_version>="3.6"
pyarrow>=1.0;python_version>="3.6" and sys_platform != "win32"

==> requirements-test.txt <==
pytest>=3.9

==> requirements.txt <==
numpy>=1.13.1
```

Maybe pycparser and lark-parser are not needed if we're not routinely parsing C code anymore? That is, everything comes from the kernel-specification now, right? That only needs a Python parser, which is built in.

The yaml module you're looking for might be PyYAML—the Python 3.9 task avoids installing requirements-dev.txt (and does fewer tests as a result). Some of those things aren't available for Python 3.9 yet. If all of the failures are 3.9, that's an indication that the new test should be conditioned on "not Python 3.9," which is the case for most tests right now.

jpivarski (approved):
Great!

PR #508, opened 2020-11-02
Fixes #499 by removing gaps from ListArray::content.

### LINKED ISSUE #499: Bug in take operation on nested arrays
I found a rather complex array layout that cannot for some reason be sliced by a boolean mask. Both are attached in the pickle file [take_bug.pkl.gz](https://github.com/scikit-hep/awkward-1.0/files/5431943/take_bug.pkl.gz) as I didn't want to recreate the whole layout with a more canned example, although it could probably be done.
To reproduce:
```python
import gzip
import pickle
import awkward1 as ak

with gzip.open("take_bug.pkl.gz", "rb") as fin:
    a, b = pickle.load(fin)

# works fine
ak.Array(ak.to_list(a))[b]

# raises
a[b]
```
There is definitely something peculiar about the layout that is causing issue since re-creating the array with a simplified layout allows the take to work.

jpivarski:


jpivarski:
Note to self: It's a coincidence that the two MacOS failures are the oldest Python versions; it's just the Azure/NodeJS bug again. When everything else finishes, restart failed jobs to get these two done.

PR #524, opened 2020-11-09
argsort and sort for indexed option arrays bug fix 

ianna:
address https://github.com/scikit-hep/awkward-1.0/issues/523

### DIFF
--- a/src/libawkward/array/IndexedArray.cpp
+++ b/src/libawkward/array/IndexedArray.cpp
@@ -2133,6 +2133,13 @@ namespace awkward {
     util::handle_error(err2, classname(), identities_.get());
 
     ContentPtr next = content_.get()->carry(nextcarry, false);
+
+    bool inject_nones = false;
+    std::pair<bool, int64_t> branchdepth = branch_depth();
+    if (numnull > 0  &&  !branchdepth.first  &&  negaxis != branchdepth.second) {
+      keepdims = false;
+      inject_nones = true;
+    }
     ContentPtr out = next.get()->sort_next(negaxis,
                                            starts,
                                            nextparents,
@@ -2158,7 +2165,13 @@ namespace awkward {
             nextoutindex,
             out);
 
-    std::pair<bool, int64_t> branchdepth = branch_depth();
+    if(inject_nones) {
+      out = std::make_shared<RegularArray>(Identities::none(),
+                                           util::Parameters(),
+                                           out,
+                                           parents_length);
+    }
+
     if (!branchdepth.first  &&  negaxis == branchdepth.second) {
       return out;
     }
@@ -2201,7 +2214,8 @@ namespace awkward {
       else {
         throw std::runtime_error(
           std::string("sort_next with unbranching depth > negaxis is only "
-                      "expected to return RegularArray or ListOffsetArray64; "
+                      "expected to return RegularArray or ListOffsetArray64 or "
+                      "IndexedArrayOf<int64_t, ISOPTION>; "
                       "instead, it returned ") + out.get()->classname()
           + FILENAME(__LINE__));
       }
@@ -2246,6 +2260,13 @@ namespace awkward {
     util::handle_error(err2, classname(), identities_.get());
 
     ContentPtr next = content_.get()->carry(nextcarry, false);
+
+    bool inject_nones = false;
+    std::pair<bool, int64_t> branchdepth = branch_depth();
+    if (numnull > 0  &&  !branchdepth.first  &&  negaxis != branchdepth.second) {
+      keepdims = false;
+      inject_nones = true;
+    }
     ContentPtr out = next.get()->argsort_next(negaxis,
                                               starts,
                                               nextparents,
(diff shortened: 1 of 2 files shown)


ianna:
@jpivarski - exactly the same issue with `sort_next `

jpivarski (changes requested):
My minimally reproducing example no longer raises a RuntimeError, but I'm not 100% sure that the answer is what we want it to be.

```python
>>> ak.argsort(ak.Array([[3, 2, 1], [], None, [4, 5]]), axis=1)
<Array [[2, 1, 0], [], [0, 1]] type='3 * var * int64'>
```

Should the None disappear entirely? NumPy doesn't provide much guidance here because there's no equivalent of Nones in a NumPy array. However, when we apply a reducer to the same thing, we've decided to keep the None in place:

```python
>>> ak.sum(ak.Array([[3, 2, 1], [], None, [4, 5]]), axis=1)
<Array [6, 0, None, 9] type='4 * ?int64'>
```

Sorting doesn't turn lists into scalars, but I think it's close enough that it should follow the same pattern of keeping the None in place.

As a related question, if the IndexedOptionArray/IndexedArray also reorders/duplicates elements, does `ak.argsort` do the right thing?

```python
>>> content = ak.Array([[3, 2, 1], [], [4, 5]])
>>> index = ak.layout.Index64(np.array([2, 1, 1, -1, 0]))
>>> array = ak.Array(ak.layout.IndexedOptionArray64(index, content.layout))
>>> array
<Array [[4, 5], [], [], None, [3, 2, 1]] type='5 * option[var * int64]'>
>>> ak.argsort(array, axis=1)
<Array [[0, 1], [], [], [2, 1, 0]] type='4 * var * int64'>
```

It does: the length-2 list that was originally in the right order is first in the output (`[0, 1]`), then the two empty lists, then the length-3 list that has to be reversed (`[2, 1, 0]`). But the None value is gone.

I think `ak.argsort` (and `ak.sort`) should keep the None values where they were. There is a [nextcarry_outindex](https://github.com/scikit-hep/awkward-1.0/blob/c3b9442644004784d81fbdd3cde577a528ada93e/src/libawkward/array/IndexedArray.cpp#L2480-L2503) utility function for helping with this: the sort functions seem to be applying the "nextcarry" (to remove None values and keep lists in the right order before sorting their contents) but not the "outindex" (to put the Nones back in).

Looking at the sort implementations, it looks like you're calculating "nextcarry" and "outindex" in its own kernel; instead of refactoring to use the above-mentioned utility function, maybe the implementation of that kernel needs to be tweaked to include the `-1` values in the "outindex" representing Nones? If there's a better way to do it that I'm overlooking, that would be good.

tests/test_0074-argsort-and-sort.py
```diff
@@ -7,8 +7,22 @@
...
+#     assert awkward1.to_list(array.argsort(1, True, False)) == [[5, 4, 3, 2, 1, 0], [1, 0]]
+#     with open('arg_sort_boom.pcl', 'rb') as f:
+#         data = pickle.load(f)
+#
+#         so = awkward1.argsort(data, axis=-1)
+#         assert awkward1.to_list(so) == [[5, 4, 3, 2, 1, 0], [1, 0]]
+#         assert awkward1.to_list(awkward1.sort(data, axis=-1)) == data[so]
+
+def test_keep_None_in_place_test():
+    assert awkward1.to_list(awkward1.argsort(awkward1.Array([[3, 2, 1], [], None, [4, 5]]), axis=1)) == [[2, 1, 0], [], None, [0, 1]]
```
jpivarski, line 13: This looks good! Does it also work for `sort`?

src/libawkward/array/IndexedArray.cpp
```diff
@@ -2133,6 +2133,13 @@ namespace awkward {
     util::handle_error(err2, classname(), identities_.get());
 
     ContentPtr next = content_.get()->carry(nextcarry, false);
+
+    bool inject_nones = false;
+    std::pair<bool, int64_t> branchdepth = branch_depth();
+    if (numnull > 0  &&  !branchdepth.first  &&  negaxis != branchdepth.second) {
+      keepdims = false;
```
jpivarski, line 2140: `keepdims` is needed for reducers, which would ordinarily remove a dimension (`keepdims` keeps the dimension from being removed, but makes its length 1). I haven't understood the role of `keepdims` for sorting, which always keeps the dimensions the same.

If it's not needed but harmless, we don't have to get rid of it right away, but should probably make a note of it.

jpivarski (approved):

tests/test_0074-argsort-and-sort.py
ianna: yes, I'm cleaning up the tests and have added one for sort.

src/libawkward/array/IndexedArray.cpp
ianna: Yes, agree.

ianna:
@jpivarski - I think, I'm done with it. Please, review it when you have time. Thanks!

jpivarski:
> I think, I'm done with it.

I agree! I'll just wait for the tests before merging.

PR #525, opened 2020-11-09
Fix #402: Form::getitem_field must return the Form of what Content::getitem_field would return.

### LINKED ISSUE #402: Make Form::getitem_field correctly mirror Content::getitem_field

jpivarski:


jpivarski:
@nsmith- Make sure this doesn't break NanoEvents: it affects VirtualArrays.

(Remember how `Form::getitem_field` was returning the content of the field, rather than the Form of what `Content::getitem_field` would return? This fixes that. It also means that SliceGenerators from field-slices can have Forms now, because now they're correct. But if you have any downstream code that depends on the old kluge, that will need to get updated!)

nsmith-:
One concern would be that if SliceGenerator now has a form, slices may act slightly _more_ lazy than before, which may have undesired implications if they get nested a bunch and the form metadata builds up to be a performance/memory hog.

jpivarski:
Field-slices probably want to be lazy, though, right? Because a field-slice on an array is always an _O(1)_ operation, but getting the array might be expensive.

PR #536, opened 2020-11-13
Enforce minimum Arrow version 2.0.0 and fix issues due to ARROW-9556.

### LINKED ISSUE #531: Complain about minimal pyarrow version
From a report on Gitter, pyarrow 0.17.1 can fail with a segfault (illegal hardware instruction!) when used with Awkward Array, but in general it won't work with anything less than pyarrow 1.0 (possibly 2.0: have to check). The `import pyarrow` line should check the pyarrow version and raise an exception if the minimum isn't met (like we're already doing with Numba).

jpivarski:


jpivarski:
https://issues.apache.org/jira/browse/ARROW-9556 has been fixed since August 20, 2020, so it should be in 2.0.0. Therefore, I need to get those two tests up to date now.

Issue #184, opened 2020-03-20
Concatenate for axis != 0


jpivarski:
This was an oversight: it's really needed but I closed the "concatenate" issue (#76) once the function existed, though there's a NotImplementedError in it:

https://github.com/scikit-hep/awkward-1.0/blob/afaa40f3993819d7beac903f701a4a1fb7f69897/src/awkward1/operations/structure.py#L151-L154

This operation is known to be useful from Awkward0; it was even an external contribution (https://github.com/scikit-hep/awkward-array/pull/80).

In Awkward1, it can probably be implemented as a broadcast in Python + specialized C++ method on ListArray, ListOffsetArray, and RegularArray (nothing else). That would allow all possible `axis` values.

jpivarski:
I'm calling it a bug because it's not optional, though I suppose that's stretching the terminology.

Ellipse0934:
I want to work on this. I think a specialized C++ method which concatenates two arrays at a time with Python calling it repeatedly for all input arrays should do it.

Writing it in Python as a recursive function:
```python
def concat(arrays, axis=0, mergebool=True, highlevel=True):
    if axis != 0:
        res = [subarr for subarr in arrays[0]]
        for arr in arrays[1:]:
            for i in range(len(arrays[0])):
                res[i] = concat([res[i], arr[i]], axis=axis-1, mergebool=mergebool, highlevel=highlevel)
        return ak.Array(res)
    else:
        return ak.concatenate(arrays, mergebool=mergebool, highlevel=highlevel)
```
```python
>>> x = ak.Array([[[1, 2, 3], [4, 5]], [[6, 7]]])
>>> y = ak.Array([[[], [8, 9]], [[10, 11]]])
>>> x, y
(<Array [[[1, 2, 3], [4, 5]], [[6, 7]]] type='2 * var * var * int64'>,
 <Array [[[], [8, 9]], [[10, 11]]] type='2 * var * var * int64'>)

>>> concat([x, y], axis = 0).tolist()
[[[1, 2, 3], [4, 5]], [[6, 7]], [[], [8, 9]], [[10, 11]]]

>>> concat([x, y], axis = 1).tolist()
[[[1, 2, 3], [4, 5], [], [8, 9]], [[6, 7], [10, 11]]]

>>> concat([x, y], axis = 2).tolist()
[[[1, 2, 3], [4, 5, 8, 9]], [[6, 7, 10, 11]]]
```
[...]

PR #539, opened 2020-11-15
concatenate for a nonzero axis operation

ianna:
address Issue #184

TODO:

- [x] introduce negative axis
- [x] fix deprecated warning

code has been moved from https://github.com/scikit-hep/awkward-1.0/pull/361

### DIFF
--- a/src/awkward1/operations/structure.py
+++ b/src/awkward1/operations/structure.py
@@ -989,8 +989,9 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
             "need at least one array to concatenate"
             + awkward1._util.exception_suffix(__file__)
         )
+    posaxis = contents[0].axis_wrap_if_negative(axis)
 
-    if axis == 0:
+    if posaxis == 0:
         batch = [contents[0]]
         for x in contents[1:]:
             if batch[-1].mergeable(x, mergebool=mergebool):
@@ -1011,7 +1012,7 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
                 batch = [collapsed.merge_as_union(x, 1)]
 
         def getfunction(inputs, depth):
-            if depth == axis:
+            if depth == posaxis:
                 out = inputs[0].mergemany_as_union(inputs[1:], 1)
                 return lambda: (out,)
             else:
--- a/src/awkward1/operations/structure.py
+++ b/src/awkward1/operations/structure.py
@@ -1003,14 +1003,6 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
         out = batch[0].mergemany(batch[1:])
 
     else:
-        batch = [contents[0]]
-        for x in contents[1:]:
-            if batch[-1].mergeable(x, mergebool=mergebool):
-                batch.append(x)
-            else:
-                collapsed = batch[0].mergemany_as_union(batch[1:], 1)
-                batch = [collapsed.merge_as_union(x, 1)]
-
         def getfunction(inputs, depth):
             if depth == posaxis:
                 out = inputs[0].mergemany_as_union(inputs[1:], 1)
@@ -1018,7 +1010,7 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
             else:
                 return None
 
-        out = awkward1._util.broadcast_and_apply(batch, getfunction, behavior=awkward1._util.behaviorof(*arrays))[0]
+        out = awkward1._util.broadcast_and_apply(contents, getfunction, behavior=awkward1._util.behaviorof(*arrays))[0]
 
     if isinstance(out, awkward1._util.uniontypes):
         out = out.simplify(mergebool=mergebool)
--- a/src/awkward1/operations/structure.py
+++ b/src/awkward1/operations/structure.py
@@ -1003,7 +1003,6 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
         out = batch[0].mergemany(batch[1:])
 
     else:
-        allow_records=True
         for x in contents[1:]:
             if x.axis_wrap_if_negative(axis) != posaxis:
                 raise ValueError(
--- a/src/awkward1/operations/structure.py
+++ b/src/awkward1/operations/structure.py
@@ -978,7 +978,7 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
     element for element, and similarly for deeper levels.
     """
     contents = [
-        awkward1.operations.convert.to_layout(x, allow_record=False) for x in arrays
+        awkward1.operations.convert.to_layout(x, allow_record=False if axis == 0 else True) for x in arrays
     ]
     contents = [
         x.toContent() if isinstance(x, awkward1.partition.PartitionedArray) else x
@@ -989,8 +989,8 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
             "need at least one array to concatenate"
             + awkward1._util.exception_suffix(__file__)
         )
-    posaxis = contents[0].axis_wrap_if_negative(axis)
 
+    posaxis = contents[0].axis_wrap_if_negative(axis)
     if posaxis == 0:
         batch = [contents[0]]
         for x in contents[1:]:
@@ -1003,6 +1003,7 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
         out = batch[0].mergemany(batch[1:])
 
     else:
+        allow_records=True
         for x in contents[1:]:
             if x.axis_wrap_if_negative(axis) != posaxis:
                 raise ValueError(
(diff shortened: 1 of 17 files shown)


ianna:
@jpivarski - I need some guidance on negative axis.

ianna:
@jpivarski - I need some guidance on negative axis. So far if the `axis` is negative, the first array is used to convert it to positive axis. That assumes that all arrays have the same depth. This will not work if we want to concatenate arrays of different depths at axis -1. I guess, I'd need to do the axis wrap and check for each array in a `getfunction`.

The deprecated warning/future error comes from the following:
https://github.com/scikit-hep/awkward-1.0/blob/4ecae71066c10d8a3ea5a0efc80bade0394f360e/src/awkward1/_connect/_numpy.py#L214
where `broadcast_and_apply` does not allow records:
```python
    out = awkward1._util.broadcast_and_apply(
        inputs, getfunction, behavior, allow_records=False
    )
```
on the other hand, merging records as union should be allowed. I wonder if the `allow_records` flag could be percolated from here:
```python
out = awkward1._util.broadcast_and_apply(contents, getfunction, behavior=awkward1._util.behaviorof(*arrays),allow_records=True)[0]
```

jpivarski:
> So far if the `axis` is negative, the first array is used to convert it to positive axis. That assumes that all arrays have the same depth. This will not work if we want to concatenate arrays of different depths at axis -1.

That's an important point that I hadn't thought of. Because of that, it could be acceptable to require `axis` to be nonnegative for all functions that take more than one array, but your solution of just refusing them if they have different depths shrink-wraps around the problem more naturally.

jpivarski:
> where `broadcast_and_apply` does not allow records:

The new `allow_records` parameter of `broadcast_and_apply` determines whether the recursion will go through records. Should `concatenate` recurse _through_ records? That would mean

```python
ak.concatenate([ak.Array([{"x": [1], "y": [1.1]}]), ak.Array([{"x": [2, 3], "y": [2.2]}])], axis=-1)
```

would return

```python
ak.Array([{"x": [1, 2, 3], "y": [1.1, 2.2]}])
```

because `broadcast_and_apply` can match up field names. The new parameter was added to prevent this for NumPy ufuncs (e.g. `np.sqrt`, `np.cos`, `np.exp`, ...) because records with special meanings like Lorentz vectors overload these ufuncs (especially operations like addition and subtraction), and the default implementation is often wrong (such as elementwise addition when the Lorentz vector is polar). Automatically broadcasting _ufuncs_ through records is error-prone because someone might think that the ufunc has been overloaded when it's actually the (wrong) default behavior.

That doesn't apply to concatenation. Since concatenation can't be overloaded, I think it would be fine to allow concatenation to recurse through records. Negative `axis` is defined through `wrap_if_negative`, which descends through records—the reason it needs to be called at each depth is because it's inconclusive until it passes through any records whose fields have different depths. By using `wrap_if_negative`, you're already hinting that it's possible to recurse through records.

So I think you can set `allow_records=True`. It shouldn't require any work other than setting that value.

jpivarski (approved):
This is looking very good. I agree with the decision to start a new PR.

jpivarski:
`allow_records=True` may be all you need to fix the deprecation warning. When `allow_records=False`, such as in ufuncs, it doesn't raise an error but a warning because it's a new behavior.

Are you still getting errors when you run `localbuild.py --pytest tests` with Numba? When you last posted them, they (all three of them) were not the same kind of error as was in the Windows job (which was one error with a different message). I think your errors can be fixed by installing Awkward Array with pip in addition to running `localbuild.py`.

ianna:
> `allow_records=True` may be all you need to fix the deprecation warning. When `allow_records=False`, such as in ufuncs, it doesn't raise an error but a warning because it's a new behavior.
> 
> Are you still getting errors when you run `localbuild.py --pytest tests` with Numba? When you last posted them, they (all three of them) were not the same kind of error as was in the Windows job (which was one error with a different message). I think your errors can be fixed by installing Awkward Array with pip in addition to running `localbuild.py`.

Thanks, @jpivarski ! No, I don't have any errors except for the future one I've just pushed. I cannot figure out how to pass `allow_records=True` to `apply`...

src/awkward1/operations/structure.py
```diff
@@ -996,15 +990,37 @@ def concatenate(arrays, axis=0, mergebool=True, highlevel=True):
...
+
+        def getfunction(inputs, depth):
+            if depth == posaxis:
+                out = inputs[0].mergemany_as_union(inputs[1:], 1)
+                return lambda: (out,)
+            else:
+                return None
+
+        out = awkward1._util.broadcast_and_apply(contents, getfunction,
+            behavior=awkward1._util.behaviorof(*arrays), allow_records=True)[0]
```
jpivarski, line 1021: This is the call to `broadcast_and_apply` that can take `allow_records=True` as an optional argument. Oh, you've already got it here...

ianna: Here is where it fails:
```python
>       assert ak.concatenate([numbers[0], records[0]], 1) == [
            [1.1, 2.2, 3.3, {"x": 0.0, "y": []}, {"x": 1.1, "y": [1]}],
            [],
            [4.4, 5.5, {"x": 2.2, "y": [1, 2]}],
            [6.6, 7.7, 8.8, 9.9, {"x": 3.3, "y": [1, 2, 3]}, {"x": 4.4, "y": [1, 2, 3, 4]}],
        ]

tests\test_0184-concatenate-operation.py:194: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\numpy\lib\mixins.py:25: in func
    return ufunc(self, other)
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\highlevel.py:1369: in __array_ufunc__
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\_connect\_numpy.py:215: in array_ufunc
    inputs, getfunction, behavior, allow_records=False
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\_util.py:996: in broadcast_and_apply
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\_util.py:759: in apply
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\_util.py:847: in apply
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\_util.py:659: in apply
    outcontents.append(apply(nextinputs, depth))
C:\hostedtoolcache\windows\Python\2.7.18\x86\lib\site-packages\awkward1\_util.py:886: in apply
    deprecate(exception, "1.0.0", "2020-12-01")
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
```
changing `allow_records `to `True` in line  `_numpy.py:215` causes this failure:
```python

    def test_this_should_raise_a_warning():
        awkward1.deprecations_as_errors = True
    
        one = awkward1.Array([{"x": 1}, {"x": 2}, {"x": 3}])
        two = awkward1.Array([{"x": 1.1}, {"x": 2.2}, {"x": 3.3}])
    
        with pytest.raises(ValueError):
>           one + two
E           Failed: DID NOT RAISE <class 'ValueError'>

tests/test_0457-remove-broadcasting-over-fields.py:20: Failed
```

jpivarski: Don't set `allow_records=True` in `_numpy.py:215` (the `array_ufunc` method). It was for that exact case that `allow_records` was introduced, to set it to `False`.

You are correctly setting `allow_records=True` in `structure.py:1021`.

The reason it's a parameter is so that functions like `concatenate` can set it to True while `array_ufunc` can set it to False. NumPy's ufuncs are not allowed to descend through RecordArrays unless a customization is defined, so that's why it was set to False in `array_ufunc`, and that's what the failing test was supposed to catch (and it did catch it!). But I think it makes sense for `concatenate` to descend through RecordArrays—the same motivation isn't there. (You can't overload `concatenate` the way you can overload ufuncs.)

jpivarski:
I fixed the deprecation warning; it was because

```python
ak.Array([1, 2, 3]) == ak.Array([1, 2, 3])
```

evaluates to

```python
ak.Array([True, True, True])
```

not

```python
True
```

and `==` is the NumPy ufunc we will not be allowing to descend through records in version 1.0.0. The deprecation warning was telling us something real.

This has another bad consequence:

```python
assert ak.Array([1, 2, 3]) == ak.Array([3, 2, 1])
```

_would not_ raise an assertion. The immediate fix was to add "`awkward1.to_list(·)`" around your arrays, since Python lists do evaluate to True or False with the `==` operator. This business of returning an array of True/False is (very convenient!) NumPy behavior that Awkward Array adheres to. Since assertions (or any checks) with `==` are now dangerous, NumPy raises an assertion whenever an array is used where a language feature (`if`, `while`, ...) wants to determine the scalar truth of an expression. Therefore, Awkward Array should, too, so I added that.

And then—wouldn't you know it?—this uncovered a few more bugs.

I put in fixes for all of these things. I think this PR is now finished, but I'll let you be the judge of that. I'd like to include it in version 0.4.4 tomorrow morning, but we can talk about that on our call.

src/awkward1/operations/structure.py
jpivarski: I can see that you didn't actually change `_numpy.py:215` in the branch, so I'm resolving this thread.

ianna:
Wow! Thanks, @jpivarski ! Yes, I think I'm done with this PR.

PR #542, opened 2020-11-17
Pickle-able mixin classes from decorator

nsmith-:
By attaching the corresponding *Array and *Record classes to the module from which the class comes. May be fragile (but no more so than pickle itself)


jpivarski:
That looks like a good idea, and I don't see how it could hurt. Is this complete? (Can I merge?)

nsmith-:
I'm done with it yes

nsmith-:
As a note, an unexpected but easily remedied consequence of this was that sphinx crashed on trying to load the documentation for `XArray` or `XRecord`. We solved this by simply adding an `__all__ = ["X", ...]` declaration in our modules where the mixin decorator was used.

PR #548, opened 2020-11-17
concatenate bug-fix

ianna:
address issue https://github.com/scikit-hep/awkward-1.0/issues/545

### DIFF
--- a/tests/test_0184-concatenate-operation.py
+++ b/tests/test_0184-concatenate-operation.py
@@ -10,12 +10,16 @@ import numpy
 import awkward1
 
 def test_concatenate_number():
-    assert awkward1.to_list(awkward1.concatenate([awkward1.Array([[1, 2, 3], [], [4, 5]]), 999], axis=1)) == [[1, 2, 3, 999], [999], [4, 5, 999]]
+    assert awkward1.to_list(awkward1.concatenate([
+        awkward1.Array([[1, 2, 3], [], [4, 5]]), 999], axis=1)) == [
+            [1, 2, 3, 999], [999], [4, 5, 999]]
     assert str(awkward1.type(awkward1.concatenate([
         awkward1.Array([[1, 2, 3], [], [4, 5]]),
         awkward1.Array([[123], [223], [323]])], axis=1))) == "3 * var * int64"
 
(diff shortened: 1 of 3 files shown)


tests/test_0184-concatenate-operation.py
```diff
@@ -9,6 +9,26 @@
...
+    assert str(awkward1.type(awkward1.concatenate([
+        awkward1.Array([[1, 2, 3], [], [4, 5]]),
+        awkward1.Array([[123], [223], [323]])], axis=1))) == "3 * var * int64"
+    assert awkward1.to_list(awkward1.concatenate([
+        awkward1.Array([[1, 2, 3], [], [4, 5]]),
+        awkward1.Array([[123], [223], [323]])], axis=1)) == [
+                [1, 2, 3, 123], [223], [4, 5, 323]]
+
+    one = awkward1.Array([[1, 2, 3], [], [4, 5]])
+    two = awkward1.Array([123, 223, 323])
```
ianna, line 28: @jpivarski - I shall admit it looks strange to me that both this and the above test case should give the same result, but I cannot argue with a user request :-)

ianna (commented):
@jpivarski - I think all issues are addressed. Please, have a look. Thanks.

tests/test_0184-concatenate-operation.py
jpivarski: It's broadcasting; part of this general facility:

```python
>>> ak.Array([10, 20, 30]) + ak.Array([[1, 2, 3], [], [4, 5]])
<Array [[11, 12, 13], [], [34, 35]] type='3 * var * int64'>
```

It applies across the board for all functions of more than one array, so it would be a little strange if concatenation was an exception. Also, using `broadcast_and_apply` will cause axes above the relevant one to be broadcasted, so it's doubly strange for the relevant axis to not be broadcasted. This is just a consistency thing (and likely convenient, since physicists often have arrays of particle parameters and arrays of event parameters, which are at different levels of depth).

jpivarski (approved):
This looks good, thanks!

I tested it with some additional examples and fixed up a few more corner cases (e.g. asking for the depth from an `input`, which might be a non-`Content`).

It didn't handle a broadcasting-through-records case, but I think we can shelve that until someone actually asks for it.

PR #556, opened 2020-11-22
Add script to clean generated files

reikdas:


jpivarski:
Great, thanks! I don't see any issues, and I'll want to keep the PRs clear during the "`awkward1`" → "`awkward`" name change:

https://github.com/scikit-hep/awkward-1.0/wiki/Name-change-procedure

jpivarski (approved):

PR #565, opened 2020-12-01
Tool to check if kernel is implemented in all places

reikdas:
and reword flag from --check-sorted to --check-spec-sorted

### DIFF
--- a/.ci/azure-buildtest-awkward.yml
+++ b/.ci/azure-buildtest-awkward.yml
@@ -112,7 +112,7 @@ jobs:
         condition: "and(succeeded(), ne(variables['python.version'], '3.9'))"
 
       - script: |
-          python dev/kernel-diagnostics.py --check-sorted
+          python dev/kernel-diagnostics.py --check-spec-sorted
         displayName: "Check if kernel specification is sorted"
         condition: "and(succeeded(), ne(variables['python.version'], '3.9'))"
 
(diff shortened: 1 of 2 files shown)


jpivarski (commented):
As I understand it, this gives a warning (not an error) in case a kernel is not implemented, right? Not all of the CUDA kernels are implemented. Keep in mind that it's easy to miss warnings in the CI—that which does not break the build is easy to ignore. I often find compiler warnings that my Linux GCC ignores but Visual Studio or Mac Cling show and wonder how long they've been in the CI logs before I had noticed. But it would certainly be good to have this tool that tells us on-demand which kernels still need to be written.

With the name change, I've also re-started running black, flake8, and flake8-print on the Python code (including dev). Black is not enforced, but flake8 and flake8-print are in the CI now (nothing _changes_ the code, but it can make tests fail). There's probably a mode in which black can be run in "fail if changed" mode; if so, I'll add that to Awkward and Uproot. I don't like code formatters that change things for you, but they're fine as gatekeepers.

reikdas:
> As I understand it, this gives a warning (not an error) in case a kernel is not implemented, right?

It prints out the names of the kernels that have not been implemented (either in the specification, as a cpu kernel, or a gpu kernel) instead of raising an exception. We consider a kernel to "exist" if it's name is present in the specification. When this script says that a kernel is not implemented in the specification, it means that the Python implementation is missing. 


I havn't added this to CI. Should I? (The other part of this script which checks if the specification is sorted is a part of the CI, since it raises an exception if the specification file isn't sorted)

jpivarski:
Oh, I see. Yes: it makes sense to include the sorted-check in CI, but not the list of missing kernels, since we want to ensure that the kernels are sorted, but we can't ensure that there are no missing kernels until they've all been written. (So that's a check that should someday be in CI, but not yet.)

jpivarski:
By the way, is it done?

reikdas:
Yup :)

PR #571, opened 2020-12-04
validity checking based on known parameters

ianna:
addressing issue https://github.com/scikit-hep/awkward-1.0/issues/404


ianna:
@jpivarski - should a user-defined parameters be allowed? For example, the following tests would raise a `ValueError`:
```python
   def test_dress():
        class Dummy(ak.highlevel.Array):
            def __repr__(self):
                return "<Dummy {0}>".format(str(self))
    
        ns = {"Dummy": Dummy}
    
        x = ak.layout.NumpyArray(np.array([1.1, 2.2, 3.3, 4.4, 5.5]))
        x.setparameter("__array__", "Dummy")
>       a = ak.Array(x, behavior=ns, check_valid=True)

tests/test_0032-replace-dressedtype.py:72: 
```
and
```python
    def test_dress():
        class Dummy(ak.highlevel.Array):
            def __repr__(self):
                return "<Dummy {0}>".format(str(self))
    
        ns = {"Dummy": Dummy}
    
        x = ak.layout.NumpyArray(np.array([1.1, 2.2, 3.3, 4.4, 5.5]))
        x.setparameter("__array__", "Dummy")
>       a = ak.Array(x, behavior=ns, check_valid=True)

tests/test_0039-no-hanging-types.py:62: 
```
```python
>               raise ValueError(out)
E               ValueError: at layout (NumpyArray): __array__ can not be Dummy
E               
E               (https://github.com/scikit-hep/awkward-1.0/blob/1.0.0rc3/src/libawkward/array/NumpyArray.cpp#L1346)
```

jpivarski:
User-defined types are definitely allowed—that's how we stake a claim for new functionality before it is implemented. In the space of all possible parameter names, only a few (`__array__`, `__record__`, and `__typestr__`) have been given meaning so far, but more will likely be added. For `__array__`, a few values have been given special meaning: `"string"`, `"bytestring"`, `"char"`, `"byte"`, `"categorical"`, but more will likely be added. Uproot is consistently presenting `std::map` objects as RecordArrays with `__record__` = `"sorted_map"` (after sorting the key-value pairs) with the expectation that someday, we'll define behaviors that allow people to use these maps to lookup values by sets of keys (probably using NumPy's `np.searchsorted`). @nsmith- has been using `__doc__` as a way to carry documentation from NanoAOD TBranch titles into Awkward Arrays. The [vector](https://github.com/scikit-hep/vector) project will be adding meanings for `__record__` = `"LorentzVector"`, `"Vector2D"`, and `"Vector3D"`.

So the rules you're defining should be of the form "if `__array__` = `"string"`, then require such-and-such," but not "`__array__` must be `null` or such-and-such."

ianna:
@jpivarski - as discussed, I'm done with this PR for the time being :-)

Note, the following ASIS implementation is work in progress. The `quick sort` algorithm implemented in `src/cpu-kernels/awkward_NumpyArray_subrange_equal.cpp` does not use the C++ standard library, but it's still not suitable for the GPU. A previously implemented lexicographical ranges comparison is not done yet.

src/libawkward/array/EmptyArray.cpp
```diff
@@ -17,6 +17,12 @@
 
 #include "awkward/array/EmptyArray.h"
 
+namespace {
+  std::vector<std::string> valid_parameters = { "\"char\"", "\"byte\"",
+    "\"string\"", "\"bytestring\"" };
+  std::vector<std::string> non_valid_parameters = { "\"categorical\"" };
+}
+
```
jpivarski: What is this? Is this supposed to be in EmptyArray.cpp?

src/libawkward/array/EmptyArray.cpp
```diff
@@ -377,6 +383,25 @@ namespace awkward {
...
+      [&](const std::string& i){
+        return (parameter_equals("__array__", i)) ? true : false;
+      });
+      if (result) {
+        return (std::string("at ") + path + std::string(" (") + classname()
+                + std::string("): __array__ can not be ")
+                + util::parameter_asstring(parameters_, "__array__")
+                + FILENAME(__LINE__));
+      }
+    }
```
jpivarski: Personally, I would just write a `for` loop, or even a single `if` statement with explicit checks. The complexity cost of introducing lambda functions and `std::any_of` for something this simple doesn't seem worth it.

If we want to allow for growth in the set of allowed and not-allowed parameters, we can just centralize this check using an internal function call, maybe a method on `Content`. That would also be more DRY than listing the set of allowed/not-allowed parameters in every node's .cpp file. When we add a new recognized parameter with rules about which nodes it can be used on, do we really want to add it to the not-allowed list of every .cpp file except one? That would be a big maintenance burden.

src/libawkward/array/IndexedArray.cpp
```diff
@@ -28,6 +28,11 @@
 #define AWKWARD_INDEXEDARRAY_NO_EXTERN_TEMPLATE
 #include "awkward/array/IndexedArray.h"
 
+namespace {
+  std::vector<std::string> valid_parameters = { "\"categorical\"" };
+  std::vector<std::string> non_valid_parameters = {};
+}
+
```
jpivarski: I see: these are in every .cpp file.

src/libawkward/array/NumpyArray.cpp
```diff
@@ -2891,10 +2922,11 @@ namespace awkward {
     else {
       std::shared_ptr<Content> out;
       std::shared_ptr<void> ptr;
+      int64_t ptr_length = length();
 
       switch (dtype_) {
       case util::dtype::boolean:
-        ptr = array_sort<bool>(reinterpret_cast<bool*>(data()),
+        std::tie(ptr, ptr_length) = array_sort<bool>(reinterpret_cast<bool*>(data()),
```
jpivarski, line 2942: As far as I've been able to find, this `std::tie` unpacking (to the left of an assignment) is allowed in C++11, so it's fine.

There are a few places where I've used `std::pair` to return two values from a function, such as `minmax_depth`, `branch_depth`, `offsets_and_flatten`, and `nextcarry_outindex`. Those would be easier to read as a `std::tie`, but I didn't know C++ had this feature! Maybe some of these can adopt the `std::tie` technique: `nextcarry_outindex` is only ever used internally, so that's a good candidate. The others are mapped as Python return values by pybind11, so it depends on whether pybind11 recognizes the `std::tie` in the same way that it does `std::pair`. (It probably does, but I don't know for certain.)

jpivarski (approved):
Looks good! As I said earlier, we can leave the uniqueness-checking in its incomplete state. I won't add a high-level interface for it (`ak.unique`) just yet.

I'm going to add a little to this PR to move the parameters checking out of each .cpp file, which would be a maintenance burden. Maybe this is what you meant about having validity checks all in one file, rather than per-file—if so, I didn't realize what you meant. The validity checks that are specific to ListArray should be in ListArray.cpp, etc., but we're going to want to add rules for each new parameter value in one place.

I'll make those changes and then I'll merge this PR.

src/libawkward/array/NumpyArray.cpp
ianna: I was thinking that `std::tie` may be needed to return more than two parameters for a numpy-like `unique` operation. If it's just two, perhaps, it would make sense to return a `ContentPtr` here.

src/cpu-kernels/awkward_NumpyArray_subrange_equal.cpp
```diff
@@ -0,0 +1,222 @@
...
+{
+  T t = *a;
+  *a = *b;
+  *b = t;
+}
+
+template <typename T>
+int partition (T* arr, int64_t low, int64_t high)
+{
+  int64_t pivot = arr[high];
```
ianna, line 18:
```diff
diff --git a/src/cpu-kernels/awkward_NumpyArray_subrange_equal.cpp b/src/cpu-kernels/awkward_NumpyArray_subrange_equal.cpp
index 78e6825..28ebf13 100644
--- a/src/cpu-kernels/awkward_NumpyArray_subrange_equal.cpp
+++ b/src/cpu-kernels/awkward_NumpyArray_subrange_equal.cpp
@@ -15,7 +15,7 @@ void swap(T* a, T* b)
 template <typename T>
 int partition (T* arr, int64_t low, int64_t high)
 {
-  int64_t pivot = arr[high];
+  T pivot = arr[high];
   int64_t i = (low - 1);
 
   for (int64_t j = low; j <= high - 1; j++) {
```

ianna (commented):
@jpivarski - looking at the code with a fresh eye...

PR #583, opened 2020-12-09
Implement unflatten function

### LINKED ISSUE #555: ak.unflatten: an operation like JaggedArray.fromcounts
It's useful to have the opposite of `ak.flatten`:

```python
>>> original = ak.Array([[0, 1, 2], [], [3, 4], [5], [6, 7, 8, 9]])
>>> counts = ak.num(original)
>>> array = ak.flatten(original)
>>> counts
<Array [3, 0, 2, 1, 4] type='5 * int64'>
>>> array
<Array [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] type='10 * int64'>
>>> ak.unflatten(array, counts)
<Array [[0, 1, 2], [], ... [5], [6, 7, 8, 9]] type='5 * var * int64'>
```

If the `counts` is a scalar integer, this would make a RegularArray; if it's a one-dimensional array (i.e. `ak.to_numpy(counts)` succeeds with the result being one-dimensional) of integers whose integers add up to less than the length of `array`, then it would make a ListOffsetArray. Such a function wouldn't take an `axis` and would probably live in `structure.py`, right after `flatten`.

jrueb:
This implements an `unflatten` function for awkward as described in #555.
As I really needed this function and the issue was marked as "good first issue" I thought I would give it a try.
Would you like to have a unit test for the function? I'm not sure as this is my first pull request, but if needed I can write one.

I noticed one problem with `cumsum` while writing this. However, this might need to go into a separate issue. Consider
```
arr = ak.Array(cupy.array([1,2,3]))
ak.nplike.of(arr.layout).cumsum(arr.layout)
```
will raise a `TypeError`. However it will work fine if you change `cupy` to `numpy`.

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1465,12 +1465,16 @@ def unflatten(array, counts, highlevel=True):
     Args:
         array: Data to create an array with an additional level from.
         counts (int or array): Number of elements the new level should have.
+            This is either an int or a one dimensional array of ints.
+            In the former case the additional level will be regularly sized,
+            similar to if counts were an array full with this value.
         highlevel (bool): If True, return an #ak.Array; otherwise, return
             a low-level #ak.layout.Content subclass.
 
     Returns an array with an additional level of nesting. This is roughly the
     inverse of #ak.flatten, where `counts` were obtained by #ak.num (both with
     `axis=1`).
+
     For example,
 
         >>> original = ak.Array([[0, 1, 2], [], [3, 4], [5], [6, 7, 8, 9]])
@@ -1487,26 +1491,39 @@ def unflatten(array, counts, highlevel=True):
         array, allow_record=False, allow_other=False
     )
     if isinstance(counts, int):
+        if counts < 0 or counts > len(layout):
+            raise ValueError(
+                "too large counts for array or negative counts"
+                + ak._util.exception_suffix(__file__)
+            )
         out = ak.layout.RegularArray(layout, counts)
     else:
         counts_layout = ak.operations.convert.to_layout(
             counts, allow_record=False, allow_other=False
         )
-        if isinstance(counts_layout, (
-            ak.layout.IndexedOptionArray64, ak.layout.IndexedOptionArray32
-        )):
-            mask_layout = counts_layout
-            counts_layout = mask_layout.content
+        if counts_layout.purelist_depth != 1:
+            raise ValueError(
+                "multi-dimensional counts"
+                + ak._util.exception_suffix(__file__)
+            )
+        if isinstance(counts_layout, ak._util.optiontypes):
+            mask = is_none(counts_layout, highlevel=False)
+            counts_layout = fill_none(counts_layout, 0, highlevel=False)
         else:
-            mask_layout = None
+            mask = None
         nplike = ak.nplike.of(counts_layout)
+        if nplike.asarray(counts_layout).dtype != "int":
+            raise ValueError(
+                "non-integer counts" + ak._util.exception_suffix(__file__)
+            )
         offsets = nplike.empty(len(counts_layout) + 1, np.int64)
         offsets[0] = 0
         nplike.cumsum(counts_layout, out=offsets[1:])
         offsets = ak.layout.Index64(offsets)
         out = ak.layout.ListOffsetArray64(offsets, layout)
-        if mask_layout is not None:
-            out = mask_layout.__class__(mask_layout.index, out)
+        if mask is not None:
+            index = ak.layout.Index8(nplike.asarray(mask).astype(np.int8))
+            out = ak.layout.ByteMaskedArray(index, out, valid_when=False)
 
     if highlevel:
         return ak._util.wrap(out, ak._util.behaviorof(array))
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1513,13 +1513,13 @@ def unflatten(array, counts, highlevel=True):
         else:
             mask = False
         nplike = ak.nplike.of(counts_layout)
-        if nplike.asarray(counts_layout).dtype != "int":
+        if counts.dtype != "int":
             raise ValueError(
                 "non-integer counts" + ak._util.exception_suffix(__file__)
             )
-        offsets = nplike.empty(len(counts_layout) + 1, np.int64)
+        offsets = nplike.empty(len(counts) + 1, np.int64)
         offsets[0] = 0
-        nplike.cumsum(counts_layout, out=offsets[1:])
+        nplike.cumsum(counts, out=offsets[1:])
         offsets = ak.layout.Index64(offsets)
         out = ak.layout.ListOffsetArray64(offsets, layout)
         if mask is not False:
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1465,9 +1465,8 @@ def unflatten(array, counts, highlevel=True):
     Args:
         array: Data to create an array with an additional level from.
         counts (int or array): Number of elements the new level should have.
-            This is either an int or a one dimensional array of ints.
-            In the former case the additional level will be regularly sized,
-            similar to if counts were an array full with this value.
+            If an integer, the new level will be regularly sized; otherwise,
+            it will consist of variable-length lists with the given lengths.
         highlevel (bool): If True, return an #ak.Array; otherwise, return
             a low-level #ak.layout.Content subclass.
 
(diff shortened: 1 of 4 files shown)


src/awkward/operations/structure.py
```diff
@@ -1460,6 +1460,60 @@ def apply(layout):
...
+        >>> array
+        <Array [0, 1, 2, 3, 4, 5, 6, 7, 8, 9] type='10 * int64'>
+        >>> ak.unflatten(array, counts)
+        <Array [[0, 1, 2], [], ... [5], [6, 7, 8, 9]] type='5 * var * int64'>
+    """
+    layout = ak.operations.convert.to_layout(
+        array, allow_record=False, allow_other=False
+    )
+    if isinstance(counts, int):
+        out = ak.layout.RegularArray(layout, counts)
```
jpivarski: Good—just be sure to raise an exception if `counts <= 0`. The layout constructors don't do many sanity checks, so user-supplied inputs need to be sanitized to avoid harder-to-debug errors downstream.

src/awkward/operations/structure.py
```diff
@@ -1460,6 +1460,60 @@ def apply(layout):
...
+        counts_layout = ak.operations.convert.to_layout(
+            counts, allow_record=False, allow_other=False
+        )
+        if isinstance(counts_layout, (
+            ak.layout.IndexedOptionArray64, ak.layout.IndexedOptionArray32
+        )):
+            mask_layout = counts_layout
+            counts_layout = mask_layout.content
+        else:
+            mask_layout = None
```
jpivarski: If `counts` is an array, it needs to be a one-dimensional array of numbers, so

```python
if not type(counts).__module__.startswith("cupy."):
    counts = ak.nplike.numpy.asarray(counts)
```

will take any array-like into a NumPy or CuPy array. If `counts` is an Awkward Array, it will be (zero-copy) converted _if_ it is one-dimensional and raise an exception if it is not. This will handle any node types, not just IndexedOptionArray (and I think this handling here is not sufficiently general).

Or wait—do you have this because you want to allow and pass through missing values? That is, if there are any None values in your `counts` array, you want these to show up as None in the output? In that case, you can do

```python
if not type(counts).__module__.startswith("cupy."):
    counts = ak.operations.convert.to_numpy(counts, allow_missing=True)
    mask = ak.nplike.numpy.ma.getmask(counts)
    counts = ak.nplike.numpy.ma.filled(counts, 0)
else:
    mask = False
```

The first line in the `if` block returns a NumPy array or NumPy masked array, the second gets a boolean array of missing values or a single False if it's not a masked array, and the third line ensures that any masked values are filled with zero.

After creating the `offsets` and `out` as a ListOffsetArray64, you can then replace the `if mask_layout is not None` block with

```python
if mask is not False:
    out = ak.layout.ByteMaskedArray(ak.layout.Index8(mask.view(np.int8)), out, valid_when=False)
```

IndexedOptionArray is not the only way to make (or get) option-type data. Using `to_numpy` with `allow_missing=True` handles all of the cases, giving you a masked array, and since you have a masked array, the easiest way to use it is to make a ByteMaskedArray.

src/awkward/operations/structure.py
```diff
@@ -1460,6 +1460,60 @@ def apply(layout):
...
+    Args:
+        array: Data to create an array with an additional level from.
+        counts (int or array): Number of elements the new level should have.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.layout.Content subclass.
+
+    Returns an array with an additional level of nesting. This is roughly the
+    inverse of #ak.flatten, where `counts` were obtained by #ak.num (both with
+    `axis=1`).
+    For example,
```
jpivarski, line 1477: Thanks for including documentation with examples! For the "For example" to be a new paragraph, it just needs a line break separating it from the above.

Maybe the `counts` argument documentation should say that the array should be one-dimensional and may include None values (if you go that route: see my comment below).

src/awkward/operations/structure.py
```diff
@@ -1460,6 +1460,60 @@ def apply(layout):
         return out
 
 
+def unflatten(array, counts, highlevel=True):
```
jpivarski, line 1463: We can leave the problem of how to add an `axis` argument for later, but since it should go before the `highlevel=True` (that's usually last, with the exception of very low-level arguments), it's a good idea to add it now, to ensure backward compatibility:

```python
def unflatten(array, counts, axis=0, highlevel=True):
```

(Just in case someone sets `highlevel` by position, we don't want their code to break when `axis` is added later.)

Then in the body of the function, raise a NotImplementedError if `axis` is not `0`.

src/awkward/operations/structure.py
jpivarski: You might also want to ensure that the `counts` array contains only integers:

```python
if not issubclass(counts.dtype.type, np.integer):
    raise ValueError(...)
```

See other examples of exception-raising in the same file for the pattern: there's a utility to ensure that all exceptions include links to this spot in GitHub.

jpivarski (commented):
Thanks for doing this! I have only small (explicit) corrections.

One question: do you want to allow missing values (None) in `counts` and pass them through to the output? If so, that's most of what the corrections do: they handle that in more generality. (IndexedOptionArray isn't the only node class that provides option-type, there's also ByteMaskedArray, BitMaskedArray, and UnmaskedArray. You don't have to explicitly check for all of these cases; the `to_numpy` function with `allow_missing=True` does that. See the comments.)

If you don't actually want this feature, if you want to require `counts` to only contain integers, then it's simpler (and I describe that, too).

In the corresponding issue, we talked about adding an `axis` parameter, but this PR doesn't have to address that. We can add a placeholder and have it raise NotImplementedError for all but the default value.

jrueb:
Thanks for this long comment. I'm learning a lot. Will add the changes later.

For the missing values: Yes, I intended to pass them to the output. This is to undo `ak.flatten` that was called on a masked array.

I did not include an `axis` parameter, because of the statement "Such a function wouldn't take an axis" in #555. Was this parameter discussed somewhere else? Anyhow, I like the idea of having an `axis` parameter in parallel to the ones of `ak.num` and `ak.flatten`. Ultimately I think it would be nice if `unflatten` could take `axis=None` just as `flatten` does. However for that to be possible probably also `num` would need to learn to accept `None`.

jpivarski:
> I did not include an `axis` parameter, because of the statement "Such a function wouldn't take an axis" in #555. Was this parameter discussed somewhere else?

Nope, I just forgot! No need to include `axis`, and I'll resolve the conversations about anything to do with `axis`.

src/awkward/operations/structure.py
jpivarski: No axis! See https://github.com/scikit-hep/awkward-1.0/pull/583#issuecomment-741939466.

src/awkward/operations/structure.py
jrueb: Is there a reason for raising an exception for `counts == 0`? I could imagine users actually wanting to unflatten with `counts=0`, because they have routines that do `flatten` followed by `unflatten` and the input just ended up being in the form of `"0 * 0 * dtype"`.

src/awkward/operations/structure.py
jrueb: This is very numpy explicit. I'm also not sure about the reason behind explicitly excluding `"cupy."` as module name.
In the additional changes I made now I tried to be more general so that it should also work with cupy.

jpivarski: The reason for

```python
if not type(counts).__module__.startswith("cupy."):
```

is to let only CuPy and NumPy pass through that step. If `counts` begins life as a Python sequence, Awkward Array, or anything with an `__array__` method, it will get through this block as a CuPy, NumPy, or possibly NumPy masked array (second version, which you want).

We don't yet have an idiom for this, the way we have `to_layout` to turn whatever you have into an Awkward layout node; before supporting TensorFlow, PyTorch tensors, and JAX, we'll need something like `to_nplike`, but that can be corrected later. (The fact that this advertises itself as a hardcoded check for `"cupy."` will make it stick out in a sweep—this will be one of the easy ones to find.)

src/awkward/operations/structure.py
jpivarski: Well, RegularArray can't have a `size` of `0`: this is part of the specification in the [RegularArray documentation](https://awkward-array.readthedocs.io/en/latest/ak.layout.RegularArray.html) (see the `__init__` of the Python pseudocode). It's also prohibited by RegularArray's `is_valid` function:

```python
>>> import awkward as ak
>>> import numpy as np
>>> ak.Array(ak.layout.RegularArray(ak.layout.NumpyArray(np.array([1, 2, 3])), 0), check_valid=True)
ValueError: at layout (RegularArray): size < 1
```

This means that functions using RegularArrays freely assume that `size` is not `0`, so the consequences are unpredictable: it doesn't fail right away (though this validity constraint _could_, since `size` is just a scalar and checking it is _O(1)_...) but some operation on it might fail.

In looking for alternative representations that give you this desired behavior for `counts == 0`, I realized that we will have to support RegularArrays with `size == 0`. They can be made fairly naturally:

```python
>>> array1 = ak.Array([1, 2, 3, 4, 5])
>>> array2 = ak.Array([[1, 2, 3], [], [4, 5]])
>>> array1[np.newaxis][0:0, 0:0]
<Array [] type='0 * 0 * int64'>
>>> array2[np.newaxis][0:0, 0:0]
<Array [] type='0 * 0 * var * int64'>
```

but the second of these is a RegularArray with `size == 0`:

```python
>>> array2[np.newaxis][0:0, 0:0].layout
<RegularArray size="0">
    <content><ListArray64>
        <starts><Index64 i="[]" offset="0" length="0" at="0x000000000000"/></starts>
        <stops><Index64 i="[]" offset="0" length="0" at="0x000000000000"/></stops>
        <content><NumpyArray format="l" shape="5" data="1 2 3 4 5" at="0x5576b495c4b0"/></content>
    </ListArray64></content>
</RegularArray>
```

I'll open an issue to find all the places where this was assumed and relax the constraint. Go ahead and let `counts == 0`; the consequences will be dealt with in that other issue and PR.

src/awkward/operations/structure.py
jrueb: What if `counts` is an array like this
```python
ak.layout.NumpyArray(cupy.array([1,2,3]))
```
Is it proper to call `to_numpy` on this? I'm currently getting a segmentation fault when I try to do this. Do we want to transfer counts to CPU in this case?

jpivarski: This was an oversight that's getting fixed in #591. What `to_numpy` is supposed to do is convert it to a NumPy array (i.e. bring it from the GPU to main memory).

The handling of GPU arrays isn't done: there are `// DERIVE` comments throughout the C++ codebase indicating spots where we're supposed to identify whether the buffers are all main memory (`kernel::lib::cpu`), all on GPUs (`kernel::lib::cuda`), or mixed (raising an exception). The Python side was mostly made NumPy/CuPy generic, but you found a missed case. I should do a more thorough sweep, like the one I'm doing for RegularArray `size == 0` in #590. We also need to get more meaningful tests for the kernels (the test cases in kernel-specification.yml haven't been checked for exhaustiveness) before implementing all the kernels in CUDA.

That's why I'm not too concerned about hard-coding the name `"cuda."` at this time: the sweep that will homogenize all of these things and make them ready for TensorFlow, PyTorch tensors, and JAX in addition to just CuPy is in the future. (That is, we're not regressing an already correct thing.)

jrueb:
Marked it a moment too early as "ready for review" but now everything should be ready for a review.

jpivarski:
I noticed that it didn't have tests, so I added them, and that gave me more things to check. Some of these appeared in the `ak.unflatten` definition, but a few others were in nplike and `to_layout`. Unless you have any complaints, I'll merge this when the tests pass.

Thanks!

jpivarski:
I know this was 4 years ago, but thanks for doing this and I'm sorry that I forgot to add you to the all-contributors list!

@all-contributors please add @jrueb for code

Issue #585, opened 2020-12-09
ak.flatten with axis = 0


Superharz:
Allow ```ak.flatten(array, axis= 0)```. As of ```awkward 0.4.5``` I get:
```
ValueError: axis=0 not allowed for flatten
```
This would make it possible to pass already flat arrays (it would just return the array as is was) and you wouldn't have to worry about the depth of ```array```.

jpivarski:
I couldn't exactly reproduce this, even going back to 0.4.5. If you pass `axis=0` to `ak.flatten`, the case is handled in Python code and you don't get this error message that comes from C++. Tinkering with it, I did find an unhandled corner-case: if you pass the negative axis that is equivalent to `axis=0` (such as `axis=-2` for a singly jagged array or `axis=-1` for a flat array), it missed the check for `axis == 0` and went into C++. As far as I can see, that's the only way to encounter this error message.

(The C++ code was written under the assumption that `axis == 0` would never be allowed, but it made sense to add that case, though its behavior is entirely different from any other `axis` value. Since `axis == 1` flattens a possibly jagged two-dimensional array into a one-dimensional array, `axis == 0` does not change the dimension of _any_ array. However, all flattening operations remove None values, so `axis == 0` only removes None from the top level. This is the only behavior that's consistent with the pattern set by all other `axis` values.)

The only way to express "flatten if not already flat" without exceptions is `axis=None`:

```python
>>> ak.flatten(ak.Array([[1, 2, 3], [], [4, 5]]), axis=None)
<Array [1, 2, 3, 4, 5] type='5 * int64'>
>>> ak.flatten(ak.Array([1, 2, 3, 4, 5]), axis=None)
```

Be sure it's what you want, though. Unlike a single-axis flatten, it flattens all levels of depth:

```python
>>> ak.flatten(ak.Array([[[1, 2], [], [3]], [], [[4, 5]]]), axis=None)
<Array [1, 2, 3, 4, 5] type='5 * int64'>
```

and even flattens records:

```python
>>> ak.flatten(ak.Array([{"x": 1.1, "y": [1]}, {"x": 2.2, "y": [1, 2]}]), axis=None)
<Array [1.1, 2.2, 1, 1, 2] type='5 * float64'>
```

PR #586, opened 2020-12-09
Fix corner-case revealed by issue #585, but distinct from that issue.

jpivarski:


Superharz:
Nice! It actually was that issue #585 .

PR #605, opened 2020-12-16
fix: avoid hardcoded threads and macOS target

henryiii:
This is broken on conda-forge because it's running out of memory on Travis - you should always be able to build single threaded if you need to!

And I've also avoided overriding the macOS deployment target - most build systems (cibuildwheel, conda-forge, etc) set this and you should avoid overriding that.

See https://github.com/pybind/cmake_example/blob/master/setup.py - this should be similar to that.

This is stopping this PR: https://github.com/conda-forge/awkward-feedstock/pull/58 (though the broken build is not the one that was being added in that PR). I'll add it as a patch as soon as it's merged. _Please be sure to Squash and Merge, not "regular merge", to make the patch clean and easy to apply._


jpivarski:
Thank you for this fix and for working on the conda-forge!

My default is to "squash and merge", and I'll do that now.

PR #610, opened 2020-12-19
Revise concatenate with axis != 0

jpivarski:


tests/test_0184-concatenate-operation.py
```diff
@@ -37,7 +37,7 @@ def test_concatenate_number():
     ) == [[1, 2, 3, 123], [223], [4, 5, 323]]
 
     one = ak.Array([[1, 2, 3], [], [4, 5]])
-    two = ak.Array([123, 223, 323])
+    two = ak.Array([[123], [223], [323]])
```
jpivarski, line 40: I revised this definition because it didn't agree with real broadcasting (now that we're getting the broadcasting from `broadcast_and_apply`). This also affects tests in the last function.

PR #620, opened 2020-12-24
Prototype Forth VM for filling Awkward Arrays from Uproot.

jpivarski:


jpivarski:
@ianna This is a conceptual start on the array-filling interface from Uproot to Awkward Array that we talked about last Tuesday.

As a reminder, the motivation is to try to keep Uproot and Awkward Array decoupled. I've made mistakes in the past where Uproot has used Awkward implementation details, which then broke when the wrong versions of Uproot and Awkward Array were used together. Although the new Uproot 4 uses only Awkward's public API, next year's project, in which Uproot offloads ROOT TBaskets to Awkward Array for filling, gets dangerously close to too much coupling. The prototype (scikit-hep/uproot4#96) puts explicit knowledge of how TTrees are serialized in Awkward Array. (In the following, `0x40000000` and 2-byte headers are ROOT details, which should stay in Uproot, not Awkward Array.)

https://github.com/scikit-hep/awkward-1.0/blob/257b24602c343e8420c30499ade40bb7bce1c429/src/libawkward/io/uproot.cpp#L64-L91

That sort of thing should be exclusively contained within Uproot.

So the idea I've been thinking about for decoupling them is to have Uproot create a sequence of ROOT-deserialization instructions that Awkward Array can follow in compiled code. "Instructions," in a general sense, is "code"—Uproot has to send Awkward Array a program. Since the motivation for this is performance (currently, Uproot is generating these instructions as Python code, which it executes using Python's `exec`), it wouldn't help much for these "instructions" to be something that has to be interpreted (after all, Python is interpreted by a C program, the Python interpreter). Ideally, we might like to just-in-time compile these instructions, but Awkward Array doesn't come with a built-in compiler.

That's why I've been thinking about this idea of using Forth. Forth is an extremely simple language—it has _less_ syntax than LISP!—and it's a half-step above assembly language, a half-step below a conventional language like C or Fortran. The following is a great introduction:

https://skilldrick.github.io/easyforth/

Instead of having proper functions with arguments, it has a single stack and "words" that manipulate that stack. It's an untyped language in the sense that the stack has only one type: integers (typically 16-bit, but we'll be using 64-bit integers). Beyond that, we get to decide what the builtin words are and how they should be wired up.

Here's how I'm thinking it should work: Uproot determines what kind of [Form](https://awkward-array.readthedocs.io/en/latest/ak.forms.Form.html) it needs to fill and how to traverse the ROOT TBaskets to fill them, in the form of a Forth program. Uproot sends the raw TBaskets and the Form to Awkward Array. Awkward Array makes [GrowableBuffers](https://awkward-array.readthedocs.io/en/latest/_static/classawkward_1_1GrowableBuffer.html) for each node in the Form, compiles the Forth program, and runs it. Our version of the Forth language includes builtin words for reading from the TBasket data, possibly putting these values on the stack (as 64-bit integers), or putting them into the GrowableBuffers (as whatever types they need to be). Awkward Array then assembles the array from the GrowableBuffers and the Form and sends it back to Uproot.

Here's why Forth seems like a good option:

   * Whereas declarative or functional languages endeavor to be side-effect free, Forth operates entirely through side-effects. Each word pulls a value off of the TBasket (updating its seek position in-place), pushes or pops values on the stack (for scratch-work), or pushes a value onto one of the GrowableBuffers. Everything we _want_ to do is a side-effect, because our main interest here is I/O. Forth seems like a better fit to this task than a declarative/functional language.
   * Forth is simple enough to minimize overheads when evaluating it. Documentation about Forth calls it an "interpreted" language, but these words can be mapped 1:1 onto integer instructions for a simple (stack-based) virtual machine. I would call that "compiled," in the same sense as Java, but for an extremely simple machine like NumExpr. The hot loop is a double `while` loop over a `dictionary`, which can be implemented as a `std::vector<std::vector<int>>`. Here's what it looks like in Python:

https://github.com/scikit-hep/awkward-1.0/blob/16c08f74c4f42262bee04cdfec24ad86d3f64c38/studies/awkward-forth/virtual-machine.py#L157-L190

The `if ... else if` can become a `switch` statement, which a C++ compiler knows how to optimize. The `which` and `where` are stacks of integers, which also have optimized data structures in the C++ STL. Even with this minimal runtime environment, we can compile and run programs that only use builtin words:

```python
>>> vm.do("3 2 +", verbose=True)
instruction          | stack before instruction
---------------------+-------------------------
3                    | <- top
2                    | 3 <- top
+                    | 3 2 <- top
                     | 5 <- top
```

Programs that define and use a function:

```python
>>> vm.do(": foo 3 2 + ; foo", verbose=True)
instruction          | stack before instruction
---------------------+-------------------------
foo                  | 
  3                  | <- top
  2                  | 3 <- top
  +                  | 3 2 <- top
                     | 5 <- top
```

Nested function definitions:

```python
>>> vm.do(": foo : bar 1 + ; 3 2 + ; foo bar", verbose=True)
instruction          | stack before instruction
---------------------+-------------------------
foo                  | 
  3                  | <- top
  2                  | 3 <- top
  +                  | 3 2 <- top
bar                  | 
  1                  | 5 <- top
  +                  | 5 1 <- top
                     | 6 <- top
```

And recursion (I haven't defined an `if` builtin word yet, so there's no way to stop the recursion gracefully):

```python
>>> vm.do(": foo + foo ; 1 2 3 4 5 foo", verbose=True)
instruction          | stack before instruction
---------------------+-------------------------
1                    | <- top
2                    | 1 <- top
3                    | 1 2 <- top
4                    | 1 2 3 <- top
5                    | 1 2 3 4 <- top
foo                  | 
  +                  | 1 2 3 4 5 <- top
  foo                | 
    +                | 1 2 3 9 <- top
    foo              | 
      +              | 1 2 12 <- top
      foo            | 
        +            | 1 14 <- top
        foo          | 
          +          | 15 <- top
  File "<string>", line 160, in do
  File "<string>", line 112, in run
  File "<string>", line 20, in pop
ValueError: stack underflow
```

The rest is adding builtin words to make this language usable for ROOT I/O. I'm going to develop this prototype by testing it with complex data from Uproot, to make sure that it has enough builtin words to parse data like that. When it's clear that it will be capable of all the use-cases Uproot requires, we can think about ways to optimize it in C++.

jpivarski:
Interesting ROOT files to try parsing with Forth:

```python
>>> import uproot
>>> import skhep_testdata

>>> # A good place to start.
>>> vector_of_vectors = uproot.open(skhep_testdata.data_path("uproot-vectorVectorDouble.root"))["t/x"]
>>> vector_of_vectors.array().tolist()
[[], [[], []], [[10.0], [], [10.0, 20.0]], [[20.0, -21.0, -22.0]], [[200.0], [-201.0], [202.0]]]

>>> # Everything in this tree is an interesting data structure.
>>> stl_containers = uproot.open(skhep_testdata.data_path("uproot-stl_containers.root"))["tree"]

>>> # A nested record with non-trivial stuff in it that isn't overwhelmingly large.
>>> f = uproot.open(skhep_testdata.data_path("uproot-issue46.root"))
>>> f["tree/evt"].array().type
100 * Event["ArrayBool": 10 * bool, "Bool": bool, "N": int64, "SliceBool": var * bool, "StlVecBool": var * bool]

# Like the stl_containers, but a bit larger.
>>> f = uproot.open(skhep_testdata.data_path("uproot-issue243-new.root"))
>>> f["triggerList/triggerMap.first"].array().type
10252 * var * string
>>> f["triggerList/triggerMap.second"].array().type
10252 * var * float64

>>> # None of the following have been interpreted correctly by the "interpret" function.
>>> # Same with uproot-issue434.root, though it looks like Uproot 3 managed it?
>>> f = uproot.open(skhep_testdata.data_path("uproot-issue124.root"))
>>> f["KM3NET_TIMESLICE"].keys()
['KM3NET_TIMESLICE',
 'KM3NET_TIMESLICE/KM3NETDAQ::JDAQPreamble',
 'KM3NET_TIMESLICE/KM3NETDAQ::JDAQTimesliceHeader',
 'KM3NET_TIMESLICE/vector<KM3NETDAQ::JDAQSuperFrame>']
```

For some whopping huge examples (performance scaling), there's the `std::vector^N<float>` from the CHEP 2019 studies. I think I have a copy. Otherwise, they need to be regenerated from the code in [this directory](https://github.com/scikit-hep/awkward-1.0/tree/main/studies/chep-2019).

jpivarski:
Note to self: this is how to get the raw data for the Forth interpreter to work on.

```python
>>> vector_of_vectors.basket(0)
<TBasket 0 of 'x' at 0x7843507860d0>
>>> vector_of_vectors.basket(0).data
array([ 64,   0,   0,   6,   0,   9,   0,   0,   0,   0,  64,   0,   0,
        14,   0,   9,   0,   0,   0,   2,   0,   0,   0,   0,   0,   0,
         0,   0,  64,   0,   0,  42,   0,   9,   0,   0,   0,   3,   0,
         0,   0,   1,  64,  36,   0,   0,   0,   0,   0,   0,   0,   0,
         0,   0,   0,   0,   0,   2,  64,  36,   0,   0,   0,   0,   0,
         0,  64,  52,   0,   0,   0,   0,   0,   0,  64,   0,   0,  34,
         0,   9,   0,   0,   0,   1,   0,   0,   0,   3,  64,  52,   0,
         0,   0,   0,   0,   0, 192,  53,   0,   0,   0,   0,   0,   0,
       192,  54,   0,   0,   0,   0,   0,   0,  64,   0,   0,  42,   0,
         9,   0,   0,   0,   3,   0,   0,   0,   1,  64, 105,   0,   0,
         0,   0,   0,   0,   0,   0,   0,   1, 192, 105,  32,   0,   0,
         0,   0,   0,   0,   0,   0,   1,  64, 105,  64,   0,   0,   0,
         0,   0], dtype=uint8)
>>> vector_of_vectors.basket(0).byte_offsets
array([  0,  10,  28,  74, 112, 158], dtype=int32)
```

jpivarski:
The Python prototype is done. The C++ prototype will be a new PR. (I'm merging this one to be sure that the `ak.from_buffers` correction gets into the main codebase.)

nsmith-:
Interesting.. I assume on the uproot side of this the form keys don't depend on the partition, right? If so, and if we find they are invariant for ~files with identical branches~ TTrees with identical metadata, then this satisfies the needs for establishing a base form for NanoEvents and then I don't have to worry about maintaining https://github.com/CoffeaTeam/coffea/blob/master/coffea/nanoevents/mapping/uproot.py#L34 for all types of root files (and can drop the `!skip` and `!content` DSL commands as well)

PR #638, opened 2021-01-05
Prototype Forth virtual machine in C++ (close to what will be added to Awkward Array).

jpivarski:


jpivarski:
Also worth noting: half of C++'s time is spent dereferencing the `std::vectors` of `std::shared_ptrs` with a virtual method call into the `ForthOutputBuffer` (whose specialized type would not be known at compile time). There isn't any one of these that is much more significant than the others. They can't be turned off because it will _always_ be the case that the number of types of input/outputs will be unknown at compile-time, but considering that the input/output management takes half of the time in a metric where Forth is 3× worse, a heavy calculation not dominated by input/output might be 5× worse.

PR #639, opened 2021-01-05
add quick sort and argsort without recursion

ianna:


jpivarski:
This looks great—removing the recursion in quicksort by adding a stack. But does a finite `kMaxLevels` mean that there are arrays that would be too big to sort? That would be an unwelcome surprise. Is it possible to compute an upper limit maximum number of levels given the length of a list? I would guess that such an upper limit would be logarithmic in that length, and therefore not too large. If it takes _O(n log(n))_ space to sort _n_ elements, I think that would be acceptable...

ianna:
> This looks great—removing the recursion in quicksort by adding a stack. But does a finite `kMaxLevels` mean that there are arrays that would be too big to sort? That would be an unwelcome surprise. Is it possible to compute an upper limit maximum number of levels given the length of a list? I would guess that such an upper limit would be logarithmic in that length, and therefore not too large. If it takes _O(n log(n))_ space to sort _n_ elements, I think that would be acceptable...

I think, the algorithm ensures that an array of N entries can handle a set of 2<sup>N</sup> entries: the larger range is pushed into the array and the loop iterates on the smaller range. It still has quadratic time complexity on sorted arrays, but at least it would sort arrays of all possible sizes.

jpivarski:
> It still has quadratic time complexity on sorted arrays, but at least it would sort arrays of all possible sizes.

I just read up on the algorithm. If the worst case of _O(n^2)_ is rare, as Wikipedia claims, then that's okay. If this case happens when the arrays are already sorted, then that's not good because running a sorting algorithm "just to be sure it's sorted" is a common thing to do. But if the already-sorted case is that _O(n^2)_ worst case, then it's easily fixed: the algorithm could start with a quick _O(n)_ check to determine whether it's sorted, then follow with quicksort if it's not. Most unsorted lists wouldn't even take a full _O(n)_ to see that they're not sorted.

And just to be sure: the actual iteration over data in the arrays is done in cpu-kernels, not libawkward, right?

ianna:
@jpivarski - I think, the PR is more or less done. Both `sort` and `argsort` support `ascending` and `descending` order. `Stable` is not supported. I wonder if a stable `argsort` is something to work on next. Perhaps, I'll spend some time on it over the week-end.

Next, I'd like to run some profiling tests and to compare this implementation to an old std-based version. Also, I'm trying to get (an easy) access to a GPU to test the kernels. While there is data dependency between iterations within the `quick_sort` and `quick_argsort` functions, sorting the ranges are independent loop iterations.

jpivarski (commented):
I just have some questions: is quicksort what you get if you call `sort` and `argsort` with `stable=False` now?

What is the role of `kMaxLevels`? Is there a maximum size array or maximum size list in an array that can be sorted? What happens if one attempts to sort an arbitrarily large array? The CPU version of this function, at least, must work for any dataset that can fit in memory (tens of GB, with the understanding that some working space will be required, and the time will scale as _O(n log n)_).

ianna:
> I just have some questions: is quicksort what you get if you call `sort` and `argsort` with `stable=False` now?
> 
`Introsort` begins with `quicksort` and if the recursion depth goes more than a particular limit it switches to `Heapsort` to avoid Quicksort’s worse case `O(N2)` time complexity. It also uses insertion sort when the number of elements to sort is quite less.

So first it creates a partition. Three cases arises from here.

* If the partition size is such that there is a possibility to exceed the maximum depth limit then the `Introsort` switches to `Heapsort`. The maximum depth limit is defined as `2*log(N)`.
* If the partition size is too small then `Quicksort` decays to `Insertion Sort`. This cutoff is defined as 16 (due to research). So if the partition size is less than 16 then it will do insertion sort.
* If the partition size is under the limit and not too small (i.e- between 16 and 2*log(N)), then it performs a simple `quicksort`.

> What is the role of `kMaxLevels`? Is there a maximum size array or maximum size list in an array that can be sorted? What happens if one attempts to sort an arbitrarily large array? The CPU version of this function, at least, must work for any dataset that can fit in memory (tens of GB, with the understanding that some working space will be required, and the time will scale as _O(n log n)_).

The algorithm pushes the larger range into the array and the loop ingrates on a smaller range: `kMaxLevels` defines when to stop. BTW, `kMaxLevels` set to 48 allows sorting an array of about 280TB

jpivarski:
> BTW, `kMaxLevels` set to 48 allows sorting an array of about 280TB

I see! Then it is logarithmic. I agree that 280 TB is technically a limit but practically not a limit, especially since these arrays have to be in RAM. We also use integers without worrying about overflow. (And UnionArrays are limited to 128 distinct tags.) I have no qualms about accepting that kind of artificial limitation.

The rules that you specified for switching between different sorting algorithms, are those what your implementation does or what the C++ standard library does?

ianna:
> > BTW, `kMaxLevels` set to 48 allows sorting an array of about 280TB
> 
> I see! Then it is logarithmic. I agree that 280 TB is technically a limit but practically not a limit, especially since these arrays have to be in RAM. We also use integers without worrying about overflow. (And UnionArrays are limited to 128 distinct tags.) I have no qualms about accepting that kind of artificial limitation.
> 
> The rules that you specified for switching between different sorting algorithms, are those what your implementation does or what the C++ standard library does?

That’s what std does  😀

jpivarski:
I can merge it now if you think it's ready. It looks ready to me. I would then follow up with making quicksort the default from the high-level interface, and I think you wanted to do a bit more work as another PR, right?

ianna:
> I can merge it now if you think it's ready. It looks ready to me. I would then follow up with making quicksort the default from the high-level interface, and I think you wanted to do a bit more work as another PR, right?

One question I have is about sorting an array copy rather than an array itself. This can be addressed later, I think.

jpivarski:
Sorry, I missed your response earlier; I hope it's not too late. Do you mean that the `Content::sort` operation sorts the array in-place? None of the Awkward operations are supposed to be in-place: they should all leave the input unchanged.

ianna:
@jpivarski - I'm trying to convince myself that this new implementation should be used instead of an std-based one. Because, usually, I'd go with the std whenever possible...
I started with a best case scenario sorting a random one dimensional array. Here is what I get.

This PR `sort` implementation:
```python
In [5]: content = ak.layout.NumpyArray(np.random.rand(100000))                                                                                                                                                                                              

In [6]: content                                                                                                                                                                                                                                             
Out[6]: <NumpyArray format="d" shape="100000" data="0.289661 0.559196 0.306584 0.27053 0.030667 ... 0.595118 0.326157 0.993009 0.695484 0.356429" at="0x7f8301160000"/>

In [7]: %%timeit 
   ...: rand_sorted = content.sort(0, True, True) 
   ...:  
   ...:                                                                                                                                                                                                                                                     
18.8 ms ± 157 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```
vs previously implemented using an `std::sort`:
```python
In [3]: content = ak.layout.NumpyArray(np.random.rand(100000))                                                                                                                                                                                              

In [4]: %%timeit 
   ...: rand_sorted = content.sort(0, True, True) 
   ...:  
   ...:                                                                                                                                                                                                                                                     
22.7 ms ± 231 µs per loop (mean ± std. dev. of 7 runs, 10 loops each)
```
vs Numpy `sort` implementation (is this the price we pay for not sorting in place?):
```python
In [8]: nparr = np.random.rand(100000)                                                                                                                                                                                                                      

In [9]: %%timeit 
    ...: nrand_sorted = np.sort(nparr) 
    ...:  
    ...:                                                                                                                                                                                                                                                    
6.13 ms ± 45.4 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)
```

This PR sort on a larger array:
```python
In [12]: content2 = ak.layout.NumpyArray(np.random.rand(1000000))                                                                                                                                                                                           

In [13]: %%timeit 
    ...: rand_sorted2 = content2.sort(0, True, True) 
    ...:  
    ...:                                                                                                                                                                                                                                                    
221 ms ± 3.02 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
vs using an std::sort-based implementation:
```python
In [5]: content2 = ak.layout.NumpyArray(np.random.rand(1000000))                                                                                                                                                                                            

In [6]: %%timeit 
   ...: rand_sorted2 = content2.sort(0, True, True) 
   ...:  
   ...:                                                                                                                                                                                                                                                     
326 ms ± 20.2 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
``` 
finally this PR:
```python
In [14]: content3 = ak.layout.NumpyArray(np.random.rand(10000000))                                                                                                                                                                                          

In [15]: %%timeit 
    ...: rand_sorted3 = content3.sort(0, True, True) 
    ...:  
    ...:                                                                                                                                                                                                                                                    
2.53 s ± 15.4 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```
vs previos std-based:
```python
In [7]: content3 = ak.layout.NumpyArray(np.random.rand(10000000))                                                                                                                                                                                           

In [8]: %%timeit 
   ...: rand_sorted3 = content3.sort(0, True, True) 
   ...:  
   ...:                                                                                                                                                                                                                                                     
5.52 s ± 103 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

jpivarski:
Before I merge it, I need to know that it doesn't change the inputs (that it's an immutable operation) and that you're happy with it. I think it would be a little difficult to remove a commit, so let's be sure before putting it in.

On the performance tests: true ones in which the standard deviation is ten times larger than the mean are inconclusive. They'll need larger samples or repeated runs off the same samples so that caches get filled and the variances settle down. That said, I can't imagine that our own implementation would beat one from the standard library in the common single-big-list case.

The motivation for implementing our own sorting library was to control it, to ensure that what happens on the GPU is the same. However, for sorting, the specification is clear: it has to put the elements in the right order. It's understood that we'll use different algorithms on the CPU and GPU to get the same outcome. The only user-visible difference could be in how some-value elements are treated in two unstable `argsorts`, but we can argue that the correct order of these are undefined, if we do have such a difference.

One place where performance might favor your hand-written quicksort is in sorting of many small lists. That's a much more likely physics use-case than the single-large-list that sorting is usually optimized for. If you make a large ListOffsetArray full of many small lists (3, 5, 10, 20 elements), you _might_ find that your hand-written quicksort beats the standard library one because this tests a different aspect of the algorithm: how quickly it can set up and tear down as it moves from one list to the next. Surely, the standard library algorithms weren't designed for that, but that's what physicists need (sorting sets of particles in each independent event).

Finally, though, we don't necessarily want to get rid of the standard library version in favor of the hand-written quicksort: we want to provide the same options as NumPy: [np.sort](https://numpy.org/doc/stable/reference/generated/numpy.sort.html). There, a pure quicksort is the default, but they have two other algorithms ("stable" is a synonym) and one of them might correspond with the standard library algorithm. It's that true? (We talked about this, but I don't remember the outcome.) In that case, we'd want both, and I'd introduce the `kind` argument to the high-level function to choose between them.

ianna:
> Before I merge it, I need to know that it doesn't change the inputs (that it's an immutable operation) and that you're happy with it. I think it would be a little difficult to remove a commit, so let's be sure before putting it in.
> 
It does not change the input. I have another commit to push removing the comment and unifying `awkward_sort` and `awkward_quick_sort`. They both use the same algorithm, no need for code duplication. Or would you rather keep an std-based `awkward_sort` and have an `awkward_quick_sort` as an option?.

> On the performance tests: true ones in which the standard deviation is ten times larger than the mean are inconclusive. They'll need larger samples or repeated runs off the same samples so that caches get filled and the variances settle down. That said, I can't imagine that our own implementation would beat one from the standard library in the common single-big-list case.
> 
agree.

> The motivation for implementing our own sorting library was to control it, to ensure that what happens on the GPU is the same. However, for sorting, the specification is clear: it has to put the elements in the right order. It's understood that we'll use different algorithms on the CPU and GPU to get the same outcome. The only user-visible difference could be in how some-value elements are treated in two unstable `argsorts`, but we can argue that the correct order of these are undefined, if we do have such a difference.
> 
yes, unfortunately, there is a difference.

> One place where performance might favor your hand-written quicksort is in sorting of many small lists. That's a much more likely physics use-case than the single-large-list that sorting is usually optimized for. If you make a large ListOffsetArray full of many small lists (3, 5, 10, 20 elements), you _might_ find that your hand-written quicksort beats the standard library one because this tests a different aspect of the algorithm: how quickly it can set up and tear down as it moves from one list to the next. Surely, the standard library algorithms weren't designed for that, but that's what physicists need (sorting sets of particles in each independent event).
> 
yes. Also, this is where we can parallelize, I think 
> Finally, though, we don't necessarily want to get rid of the standard library version in favor of the hand-written quicksort: we want to provide the same options as NumPy: [np.sort](https://numpy.org/doc/stable/reference/generated/numpy.sort.html). There, a pure quicksort is the default, but they have two other algorithms ("stable" is a synonym) and one of them might correspond with the standard library algorithm. It's that true? (We talked about this, but I don't remember the outcome.) In that case, we'd want both, and I'd introduce the `kind` argument to the high-level function to choose between them.

Ok, I think, it's best to have both for the moment. I'll update the PR.

ianna:
@jpivarski - I'm done with this PR. Could you, please, review it? Thanks.

jpivarski (approved):
Great! (We talked on Zoom, and I gave it another quick scan.)

PR #642, opened 2021-01-06
Materialize virtual arrays in to_arrow

nsmith-:


jpivarski:
Was this simply missing? If so, thanks for noticing this!

Whereas `ak.to_buffers` might have an experts-only `persist_virtual` option, `ak.to_arrow` can't (no virtual Arrow).

And, of course, let me know if you're done with this branch and I'll merge it after tests pass.

nsmith-:
I'm done

PR #645, opened 2021-01-12
This PR adds the from_cuda_array_interface, to form CUDA arrays in a more general fashion.

trickarcher:


tests-cuda/test_0645-to-jax.py
```diff
@@ -0,0 +1,41 @@
...
+# def test_from_jax_2():
+#     content0 = ak.Array(np.array([1, 2, 3], dtype=np.int64)).layout
+#     content1 = ak.layout.NumpyArray(np.array([1, 2, 3, 4, 5], dtype=np.int32))
+#     tags = ak.layout.Index8(np.array([1, 1, 0, 0, 1, 0, 1, 1], dtype=np.int8))
+#     index = ak.layout.Index32(np.array([0, 1, 0, 1, 2, 2, 4, 3], dtype=np.int32))
+#     unionarray = ak.layout.UnionArray8_32(tags, index, [content0, content1])
+
+#     unionarray_cuda = unionarray.copy_to("cuda")
+
+#     jax_array = ak.to_jax(unionarray_cuda)
```
jpivarski, line 41: The right thing to do here would be `pytest.skip`, which is visible at the end of a test (as yellow, rather than red for a failure). That serves as a reminder that there's something that needs to be fixed in the future.

I won't hold up this PR merge for it, though.

include/awkward/kernel-dispatch.h
```diff
@@ -130,6 +130,12 @@ namespace awkward {
         void operator()(T const *ptr) { }
     };
 
+    /// @brief Returns the number of the device associated with the pointer
+    const int32_t
+    lib_device_num(
+      kernel::lib ptr_lib,
+      void* ptr);
+
```
jpivarski, line 138: Almost all of the scalar integers in Awkward Array are `int64_t`. (NumpyArray has `ssize_t` to match what comes out of pybind11, but now I think that was a mistake: it would have been cleaner to convert them once in the pybind11 interface than everywhere in the interfaces between NumpyArray and everything else.)

Perhaps the device number returned by CUDA is 32-bit, but is there any reason not to immediately convert it to 64-bit so that it can be passed around as 64-bit without compiler warnings? (Some compiler warnings are raised by just one or two of the compilers in CI: gcc on Linux, cling on MacOS, and msvc on Windows.)

requirements-dev.txt
```diff
@@ -5,3 +5,5 @@ pandas>=0.24.0;python_version>="3.6"
 numexpr;python_version>="3.6"
 autograd;python_version>="3.6"
 pyarrow>=2.0.0;python_version>="3.6" and sys_platform != "win32"
+jax>=0.2.7;python_version>="3.6" and sys_platform != "win32"
+jaxlib>=0.1.57;python_version>="3.6" and sys_platform != "win32"
```
jpivarski, line 9: This is perfectly fine: some libraries are not available on Windows.

jpivarski (approved):
This looks great! I'm going to merge it.

PR #648, opened 2021-01-13
Add a ForthMachine to the codebase, copying from 'studies'.

jpivarski:
- [x] Make stubs of all the necessary files.
- [x] Make stubs of all the necessary functions.
- [x] Pass through the public functions to Python, so that developments can be tested from Python.
- [x] Connect those inputs and outputs to Python.
- [x] Rebuild input/output framework, so that declaration-only programs can be run.
- [x] Develop parser and test output with `bytecodes` and `assembly_instructions`.
- [x] Set up stepped-running with a generator interface in Python.
- [x] Fill in runtime machine.
- [x] Ensure that we can obtain 5 ns per instruction again.
- [x] Set up example to parse TTree data from Uproot. (Test is gated by `pytest.importorskip("uproot")`.)

Full-scale ROOT parsing measurements are not part of this PR.

Nor is the Avro → Awkward example.


jpivarski:
I verified that we still have the 5 ns per instruction.

By contrast, Python's virtual machine on compiled bytecodes (similar in _kind_ to what we do) takes ~900 ns per instruction.

```python
>>> import dis
>>> def f():
...   x = 0
...   for i in range(100000):
...     x = 10 + x + 10 + x + 10 + x + 10 + x + 10 + x + 10
... 
>>> # There are 25 Python bytecode instructions in this loop, which gets repeated 100000 times.
>>> dis.dis(f)
  2           0 LOAD_CONST               1 (0)
              2 STORE_FAST               0 (x)

  3           4 LOAD_GLOBAL              0 (range)
              6 LOAD_CONST               2 (100000)
              8 CALL_FUNCTION            1
             10 GET_ITER
        >>   12 FOR_ITER                48 (to 62)
             14 STORE_FAST               1 (i)

  4          16 LOAD_CONST               3 (10)
             18 LOAD_FAST                0 (x)
             20 BINARY_ADD
             22 LOAD_CONST               3 (10)
             24 BINARY_ADD
             26 LOAD_FAST                0 (x)
             28 BINARY_ADD
             30 LOAD_CONST               3 (10)
             32 BINARY_ADD
             34 LOAD_FAST                0 (x)
             36 BINARY_ADD
             38 LOAD_CONST               3 (10)
             40 BINARY_ADD
             42 LOAD_FAST                0 (x)
             44 BINARY_ADD
             46 LOAD_CONST               3 (10)
             48 BINARY_ADD
             50 LOAD_FAST                0 (x)
             52 BINARY_ADD
             54 LOAD_CONST               3 (10)
             56 BINARY_ADD
             58 STORE_FAST               0 (x)
             60 JUMP_ABSOLUTE           12
        >>   62 LOAD_CONST               0 (None)
             64 RETURN_VALUE

>>> begintime = time.time(); f(); print((time.time() - begintime) * 1e9)
2324416875.8392334
>>> 2324416875.8392334 / (25*100000)
929.7667503356934

>>> begintime = time.time(); f(); print((time.time() - begintime) * 1e9)
2277071237.564087
>>> 2277071237.564087 / (25*100000)
910.8284950256348

>>> begintime = time.time(); f(); print((time.time() - begintime) * 1e9)
2262091398.2391357
>>> 2262091398.2391357 / (25*100000)
904.8365592956543
```

jpivarski:
For the future: the calculation in https://github.com/scikit-hep/awkward/pull/648#issuecomment-761296216 is wrong! Since `x` increases throughout the loop, it gets bigger than any int32 or int64, and most of the time measured is in handling bigint (arbitrary precision) arithmetic, rather than stepping through Python bytecodes.

A function like this:

```python
def f():
    x = 0
    for i in range(100000):
        x = 10 + x - 10 + x + 10 + x - 10 + x + 10 + x - 10
```

would walk through the same number of bytecodes (see the `dis.dis(f)`) but the final result is about 3 ns per instruction!

See the bottom of https://github.com/jpivarski-talks/2024-08-19-python-school-setting-stage/blob/main/snake-eats-its-tail.ipynb

This is still about 40× slower than the throughput of compiled instructions (at `-O0` for apples-to-apples), which is about the right order of magnitude. Compiled code can get even faster by taking advantage of vectorization and such, but this is the closest comparison that I think makes sense.

Anyway, no more saying, "Python takes 900 ns per instruction"! It takes about 3 ns per instruction. Incidentally, that's on par with AwkwardForth. The difference between Python and AwkwardForth must be due to something else. (Type checks?)

PR #651, opened 2021-01-13
Clarify cpp docs, same as #650 for python docs.

maxnoe:


jpivarski:
Oh: our changes crossed in the mail. I added a commit to PR #650 for this file (ddbcd74d66d66084f399b6983137a6aa05b4b844). I'll just use the one PR so that both files are fixed with the same commit in `main`. Thanks, though!

PR #652, opened 2021-01-13
complex numbers support

ianna:
address issue https://github.com/scikit-hep/awkward-1.0/issues/392

However, I'm not sure what the status of @sjperkins work is: https://github.com/scikit-hep/awkward-1.0/pull/421
@jpivarski - I'll be happy to close this PR if it's a duplicate.

The main reason why I'm submitting it as a draft is to ask a question about complex numbers representation in JSON. It has been suggested [here](https://github.com/scikit-hep/awkward-1.0/issues/392#issuecomment-674110892) to use `{"real: #, "imag": #}`. Is that what we want?
```python
def test_complex_numbers():
    array = ak.layout.NumpyArray(np.array([(1.+2.j), (3.+4.j)],dtype=np.complex_))
    assert ak.to_list(array) == [(1.+2.j), (3.+4.j)]
    assert (
        ak.to_json(array)
        == '[{"real":1.0,"imag":2.0},{"real":3.0,"imag":4.0}]'
    )
```


jpivarski:
I've already talked with @sjperkins (in GitHub or email? I can't remember), and he's not actively working on this: you won't collide. However, it's good to ping him here to be sure that the work isn't being duplicated.

sjperkins:
@jpivarski @ianna Yep, I'm not working on it at present. Note that the snippet in https://github.com/scikit-hep/awkward-1.0/issues/392#issuecomment-752348789 may be useful for stamping out C/C++ templates.

ianna:
Note to myself: it seems to me, the minimum amount that is necessary and sufficient to recreate a complex object is the real and imaginary part, plus the metadata:
```json
{
    "__complex__": true,
    "real": 42,
    "imag": 36
}
```
Here is an example in Python:
```python
>>> import json
>>> z = 3 + 8j
>>> type(z)
<class 'complex'>
>>> json.dumps(z)
  File "/Users/yana/anaconda3/lib/python3.8/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
  File "/Users/yana/anaconda3/lib/python3.8/json/encoder.py", line 199, in encode
    chunks = self.iterencode(o, _one_shot=True)
  File "/Users/yana/anaconda3/lib/python3.8/json/encoder.py", line 257, in iterencode
    return _iterencode(o, 0)
  File "/Users/yana/anaconda3/lib/python3.8/json/encoder.py", line 179, in default
    raise TypeError(f'Object of type {o.__class__.__name__} '
TypeError: Object of type complex is not JSON serializable
```
```python
>>> def encode_complex(z):
...     if isinstance(z, complex):
...         return (z.real, z.imag)
...     else:
...         type_name = z.__class__.__name__
...         raise TypeError(f"Object of type '{type_name}' is not JSON serializable")
... 
>>> json.dumps(9 + 5j, default=encode_complex)
'[9.0, 5.0]'
>>> json.dumps(z, default=encode_complex)
'[3.0, 8.0]'
>>> s = """
... {
...     "__complex__": true,
...     "real": 42,
...     "imag": 36
... }
... """
>>> def decode_complex(dct):
...     if "__complex__" in dct:
...         return complex(dct["real"], dct["imag"])
...     return dct
... 
>>> x = json.loads(s, object_hook=decode_complex)
>>> type(x)
<class 'complex'>
>>> x
(42+36j)
>>> 
```

jpivarski:
> Note to myself: it seems to me, the minimum amount that is necessary and sufficient to recreate a complex object is the real and imaginary part, plus the metadata:
> 
> ```json
> {
>     "__complex__": true,
>     "real": 42,
>     "imag": 36
> }
> ```

The reason I think complex numbers will be more straightforward than datatimes is that no new metadata will be needed for them. The NumpyArray class already passes data around as a `std::shared_ptr<void>`, `util::dtype` (enum), `format` (string), and `itemsize`. Arrays of complex numbers fit into this scheme just as easily as any other numbers. NumPy sets a convention that the real and imaginary parts are interleaved in the same buffer with the real part first.

```python
>>> import numpy as np
>>> array = np.array([1, 1j, 1 + 1j], np.complex64)
>>> array
array([1.+0.j, 0.+1.j, 1.+1.j], dtype=complex64)
>>> array.view(np.uint8)
array([  0,   0, 128,  63,   0,   0,   0,   0,   0,   0,   0,   0,   0,
         0, 128,  63,   0,   0, 128,  63,   0,   0, 128,  63], dtype=uint8)
>>> array.view(np.float32)
array([1., 0., 0., 1., 1., 1.], dtype=float32)
```

The two dtypes that are relevant here are `complex64` (two `float32` numbers) and `complex128` (two `float64` numbers) because of the built-in `float` and `double` in C++. Any others can be indefinitely postponed (like `float16` and `float128`, which only have stubs).

When such arrays come into Awkward Array, they go through Python's buffer interface, which assigns a `format` string. The string for `complex64` is `Zf` and the string for `complex128` is `Zd`.

```python
>>> memoryview(np.array([1, 1j, 1 + 1j], np.complex64)).format
'Zf'
>>> memoryview(np.array([1, 1j, 1 + 1j], np.complex128)).format
'Zd'
```

So, just in case you thought you would need to add attributes to classes or something, the project is simpler than that. All the places where complex numbers could be relevant are already in the code (commented out, I believe). They can be searched with a multifile grep.

ianna:
> > Note to myself: it seems to me, the minimum amount that is necessary and sufficient to recreate a complex object is the real and imaginary part, plus the metadata:
> > ```json
> > {
> >     "__complex__": true,
> >     "real": 42,
> >     "imag": 36
> > }
> > ```
> 
> The reason I think complex numbers will be more straightforward than datatimes is that no new metadata will be needed for them. The NumpyArray class already passes data around as a `std::shared_ptr<void>`, `util::dtype` (enum), `format` (string), and `itemsize`. Arrays of complex numbers fit into this scheme just as easily as any other numbers. NumPy sets a convention that the real and imaginary parts are interleaved in the same buffer with the real part first.
> 
> ```python
> >>> import numpy as np
> >>> array = np.array([1, 1j, 1 + 1j], np.complex64)
> >>> array
> array([1.+0.j, 0.+1.j, 1.+1.j], dtype=complex64)
> >>> array.view(np.uint8)
> array([  0,   0, 128,  63,   0,   0,   0,   0,   0,   0,   0,   0,   0,
>          0, 128,  63,   0,   0, 128,  63,   0,   0, 128,  63], dtype=uint8)
> >>> array.view(np.float32)
> array([1., 0., 0., 1., 1., 1.], dtype=float32)
> ```
> 
> The two dtypes that are relevant here are `complex64` (two `float32` numbers) and `complex128` (two `float64` numbers) because of the built-in `float` and `double` in C++. Any others can be indefinitely postponed (like `float16` and `float128`, which only have stubs).
> 
> When such arrays come into Awkward Array, they go through Python's buffer interface, which assigns a `format` string. The string for `complex64` is `Zf` and the string for `complex128` is `Zd`.
> 
> ```python
> >>> memoryview(np.array([1, 1j, 1 + 1j], np.complex64)).format
> 'Zf'
> >>> memoryview(np.array([1, 1j, 1 + 1j], np.complex128)).format
> 'Zd'
> ```
> 
> So, just in case you thought you would need to add attributes to classes or something, the project is simpler than that. All the places where complex numbers could be relevant are already in the code (commented out, I believe). They can be searched with a multifile grep.

Thanks, @jpivarski ! I've been thinking the metadata is needed to represent complex type in `to_` and `from_json` because the `json` module expects all custom types to be expressed as objects in the JSON standard.

jpivarski:
> Thanks, @jpivarski ! I've been thinking the metadata is needed to represent complex type in `to_` and `from_json` because the `json` module expects all custom types to be expressed as objects in the JSON standard.

Oh, that's right: the JSON methods would need to handle this type in a special way. Unfortunately, there's no standard way to encode complex numbers in JSON. If you were to make:

```python
{"real": 1, "imag": 0.5}
```

mean `1+0.5j`, then there could be cases where someone actually wanted a record with these two fields, and they would not want complex numbers. Even embedding some tag like `__complex__: true` in the _JSON_ (I'm not talking about Awkward properties) wouldn't guarantee that someone doesn't actually have and want data like that. The default parsing of JSON must be vanilla.

This situation is similar to the one where we wanted to encode floating point NaN and infinity as strings. The default parsing of a string `"infinity"` is the string `"infinity"`, but we let the user pass arguments to `from_json` that changed how this is parsed. We could do the same thing with complex numbers. Assuming that we'll always want complex numbers to go to or from a record, all that must be established is what those field names are. Unfortunately, not only is there not a standard, but everybody's solution differs: `"r"/"i"`, `"R"/"I"`, `"re"/"im"`, `"Re"/"Im"`, `"RE"/"IM"`, `"real"/"imag"`, `"Real"/"Imag"`, `"REAL"/"IMAG"`, `"real"/"imaginary"`, etc. Here's [one that hopes to become standardized](https://observablehq.com/@kgryte/stdlib-json-serialization-and-deserialization-of-exotic-d) that chose `"re"/"im"` with an extra `type` field.

A pretty good rule would be to take `complex_fields` as a 2-tuple of strings in [ak.from_json](https://awkward-array.readthedocs.io/en/latest/_auto/ak.from_json.html) and if any JSON object in the data have at least these two fields, interpret it as a complex number. By default, `complex_fields` would be None. By asking for "at least these two fields," we allow the records to have a tag. The same `complex_fields` argument should exist in [ak.to_json](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_json.html) as well.

But I would think of JSON reading/writing as one of the last things to worry about in adding complex types. The very first thing that would be needed is to pass these data from NumPy arrays into Awkward Arrays and back into NumPy arrays. Then there are reducers, which have to be handled carefully because a construct like `std::complex<float>` can only exist below and above the `extern "C"` layer. In between, it has to be a `void*` or maybe a `float*` and `double*`, depending. The name of the function determines that those numbers are to be reduced as complex numbers, rather than twice as many real numbers.

src/cpu-kernels/awkward_NumpyArray_fill_tocomplex.cpp
```diff
@@ -0,0 +1,194 @@
...
+
+#include "awkward/kernels.h"
+
+template <typename FROM, typename TO>
+ERROR
+awkward_NumpyArray_fill_tocomplex(TO* toptr,
+                                  int64_t tooffset,
+                                  const FROM* fromptr,
+                                  int64_t length) {
+  for (int64_t i = 0; i < 0.5*length; i++) {
```
ianna: TODO: this `0.5*` needs to be moved out

include/awkward/Reducer.h
```diff
@@ -679,6 +766,10 @@ namespace awkward {
   /// ReducerMin and ReducerMax are exceptions to the no-state rule: they are
   /// parameterized by `initial`, used as an initial value in the minimization
   /// or maximization.
+  ///
+  /// For the complex numbers array it
+  /// returns the minimumm absolute value (also known as norm, modulus,
+  /// or magnitude) of a complex number.
```
ianna, line 773: @sjperkins - as implemented in this PR, the reduces `min`, `max`, `argmin`, `argmax` will return either min/max absolute value or its index. You are welcome to comment or change the code to anything more suitable.

include/awkward/array/NumpyArray.h
```diff
@@ -793,27 +799,38 @@ namespace awkward {
...
+    template<typename T>
+    const std::shared_ptr<void> array_sort(const T* data,
+                                           int64_t length,
+                                           const Index64& starts,
+                                           const Index64& parents,
+                                           int64_t outlength,
+                                           bool ascending,
+                                           bool stable) const;
+    template<typename T>
+    const std::shared_ptr<void> array_unique(const T* data,
```
ianna, line 810: @jpivarski - prompted by static analyser, I separated `sort` from `is_unique` operation. The change is purely technical.

src/cpu-kernels/awkward_NumpyArray_fill_fromcomplex.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "awkward/kernels.h"
+
+template <typename FROM, typename TO>
+ERROR
+awkward_NumpyArray_fill_fromcomplex(TO* toptr,
+                                    int64_t tooffset,
+                                    const FROM* fromptr,
+                                    int64_t length) {
+  for (int64_t i = 0; i < length; i++) {
+    toptr[tooffset + i] = (TO)fromptr[i*2];
```
ianna: as implemented by @sjperkins in [here](https://github.com/scikit-hep/awkward-1.0/pull/421/files#diff-d42f70fefa3b239e75f4d450c832faed1ec03069fe9b55b8ee8b2fca898d9527R1197) - only real part is kept when converting complex to any other type.

src/libawkward/array/NumpyArray.cpp
```diff
@@ -1574,7 +1577,7 @@ namespace awkward {
             dtype_ == util::dtype::complex64  ||
             dtype_ == util::dtype::complex128  ||
             dtype_ == util::dtype::complex256  ||
-            dtype_ == util::dtype::boolean  ||
+            rawother->dtype() == util::dtype::boolean  ||
```
ianna, line 1650: reported by static analyser

src/libawkward/array/NumpyArray.cpp
```diff
@@ -2877,17 +3080,17 @@ namespace awkward {
                                                     ptr_lib_);
 
       if (mask) {
-        Index8 mask(outlength);
+        Index8 outmask(outlength);
         struct Error err = kernel::NumpyArray_reduce_mask_ByteMaskedArray_64(
           kernel::lib::cpu,   // DERIVE
-          mask.data(),
+          outmask.data(),
```
ianna, line 3156: reported by static analyser

src/libawkward/io/json.cpp
```diff
@@ -81,6 +83,16 @@ namespace awkward {
...
+    void complex(std::complex<double> x) {
+      beginrecord();
+      field("__complex__");
+      boolean(true);
+      field("real");
+      real(x.real());
+      field("imag");
+      real(x.imag());
+      endrecord();
+    }
```
ianna, line 95: TODO: as discussed complex number de/serialisation should be defined by a user

src/libawkward/io/json.cpp
```diff
@@ -195,6 +212,16 @@ namespace awkward {
...
     void real(double x) { writer_.Double(x); }
+    void complex(std::complex<double> x) {
+      beginrecord();
+      field("__complex__");
+      boolean(true);
+      field("real");
+      real(x.real());
+      field("imag");
+      real(x.imag());
+      endrecord();
```
ianna, line 235: TODO: as discussed complex number de/serialisation should be defined by a user

src/libawkward/io/json.cpp
```diff
@@ -313,6 +345,16 @@ namespace awkward {
...
+    }
```
ianna, line 381: TODO: as discussed complex number de/serialisation should be defined by a user

src/libawkward/io/json.cpp
```diff
@@ -426,6 +473,16 @@ namespace awkward {
...
     void real(double x) { writer_.Double(x); }
+    void complex(std::complex<double> x) {
+      beginrecord();
+      field("__complex__");
+      boolean(true);
+      field("real");
+      real(x.real());
+      field("imag");
+      real(x.imag());
+      endrecord();
```
ianna, line 520: TODO: as discussed complex number de/serialisation should be defined by a user

src/libawkward/io/json.cpp
```diff
@@ -586,8 +650,15 @@ namespace awkward {
...
 
     bool Double(double x) {
-      moved_ = true;
+      if (state_ == kExpectRealValue) {
+        complex_number_ = std::complex<double>(x);
+      }
+      else if (state_ == kExpectImaginaryValue) {
+        complex_number_ = {complex_number_.real(), x};
+        state_ = kExpectComplexEnd;
+      }
```
ianna: TODO: as discussed complex number de/serialisation should be defined by a user

src/libawkward/io/json.cpp
```diff
@@ -643,6 +715,14 @@ namespace awkward {
...
     Key(const char* str, rj::SizeType length, bool copy) {
       moved_ = true;
+      if (strcmp(str, "__complex__") == 0) {
+        state_ = kExpectComplexStart;
+      }
+      else if (strcmp(str, "real") == 0) {
+        state_ = kExpectRealValue;
+      } else  if (strcmp(str, "imag") == 0) {
+        state_ = kExpectImaginaryValue;
+      }
```
ianna: TODO: as discussed complex number de/serialisation should be defined by a user

tests/test_0437-stream-of-many-json-files.py
```diff
@@ -7,6 +7,18 @@
...
+    assert ak.to_list(array) == [(1.+2.j), (3.+4.j)]
+    assert (
+        ak.to_json(array)
+        == '[{"__complex__":true,"real":1.0,"imag":2.0},{"__complex__":true,"real":3.0,"imag":4.0}]'
+    )
+
+    str = '[{"__complex__":true,"real":1.0,"imag":2.0},{"__complex__":true,"real":3.0,"imag":4.0}]'
+    array2 = ak.from_json(str)
+    assert ak.to_list(array2) == [{'__complex__': None, 'real': 1.0, 'imag': 2.0}, {'__complex__': None, 'real': 3.0, 'imag': 4.0}]
+
```
ianna: TODO: as discussed complex number de/serialisation should be defined by a user

tests/test_0395-complex-type-arrays.py
```diff
@@ -0,0 +1,41 @@
...
+    assert str(ak.type(content_complex64)) == "complex64"
+    assert str(ak.type(ak.Array(content_complex64))) == "5 * complex64"
+    assert str(ak.type(array_complex64)) == "?complex64"
+    assert str(ak.type(ak.Array(array_complex64))) == "5 * ?complex64"
+
+    assert content_complex64.prod() == (10.828125+0j)
+    assert content_complex64.min() == (0.25+0j)
+    assert content_complex64.max() == (5.5+0j)
+    assert content_complex64.argmin() == 0
+    assert content_complex64.argmax() == 4
```
ianna, line 234: TODO: more tests are needed

tests/test_0395-complex-type-arrays.py
```diff
@@ -0,0 +1,41 @@
...
+    assert np.can_cast(np.float64, np.float32, "unsafe") == True
+    assert np.can_cast(np.float64, np.int8, "unsafe") == True
+    assert np.can_cast(np.float64, np.complex64, "unsafe") == True
+    assert np.can_cast(np.float64, np.complex128, "unsafe") == True
+    assert np.can_cast(np.complex64, np.float64, "unsafe") == True
+    assert np.can_cast(np.complex128, np.float64, "unsafe") == True
+
+    content_complex64 = ak.values_astype(content_float64, "complex64", highlevel=False)
+    array_complex64 = ak.layout.UnmaskedArray(content_complex64)
+    assert ak.to_list(array_complex64) == [0.25, 0.5, 3.5, 4.5, 5.5]
```
ianna: @jpivarski - should it be: 
```python
[(0.25+0j), (0.5+0j), (3.5+0j), (4.5+0j), (5.5+0j)]
```

ianna (commented):
@jpivarski and @sjperkins - comments and contributions are welcome :-)

Note to myself: 

- [x] cross-check kernel algorithms and the array lengths
- [x] add more extensive tests
- [x] refactor json
- [x] support complex numbers in `ArrayBuilder` to support an original use case:

```python
>>> import numpy as np
>>> import awkward as ak
>>> array = ak.from_iter([0j])
>>> array
<Array [0j] type='1 * complex128'>
>>> 
```

ianna:
oh, "Kernel specializations not sorted in specification" :-(

src/cpu-kernels/awkward_NumpyArray_fill_tocomplex.cpp
ianna: done

tests/test_0395-complex-type-arrays.py
ianna: ok, I think it's a feature :-)

include/awkward/Reducer.h
sjperkins: Thanks for pinging me @ianna! I guess that it would be most suitable if the behaviour matches numpy's for these functions?

ianna: Thanks @sjperkins ! `numpy` implements `min`, `max` based on `sort`. The sort order for complex numbers is lexicographic:
https://numpy.org/doc/stable/reference/generated/numpy.sort.html 
I'd better disable these reducers to avoid confusion.

@jpivarski - should this functionality be provided as the `absmin`, `absmax` reducers?

tests/test_0395-complex-type-arrays.py
jpivarski: `ak.to_list` should turn arrays with complex dtype into Python `complex` numbers.

Some of the rules in `ak.to_list` convert to NumPy and call `np.ndarray.tolist`, which would make Python `complex` numbers if the NumPy `dtype` is a complex type, so either the NumPy `dtype` is coming out wrong or it's going through one of the other rules:

https://github.com/scikit-hep/awkward-1.0/blob/c171b744ea5cb4a68708fa7bd8bad64743747d34/src/awkward/operations/convert.py#L744-L772

include/awkward/Reducer.h
jpivarski: Does NumPy really define an ordering on complex numbers?

If users want to sort complex numbers (or find their minimum/maximum) by absolute value or some other criteria, they can construct an intermediate. Calling Python's `abs` on the Awkward Array propagates `np.absolute` to every NumpyArray in the structure, so if the `dtype` is properly preserved as complex, `abs` would automatically give the modulus of all the complex numbers in an array. Then, users can argsort (or argmin/argmax) and apply it the result to the original array. We don't have to provide every option when they can be constructed.

ianna: Thanks, @jpivarski ! I'll remove `min`, `max`, `argmin`, `argmax` complex number reducers.

tests/test_0395-complex-type-arrays.py
ianna: done

tests/test_0437-stream-of-many-json-files.py
ianna: done

src/libawkward/io/json.cpp
ianna: done

src/libawkward/io/json.cpp
ianna: done

src/libawkward/io/json.cpp
ianna: done

src/libawkward/io/json.cpp
ianna: done

src/libawkward/io/json.cpp
ianna: done

src/libawkward/io/json.cpp
ianna: done

include/awkward/Reducer.h
ianna: > Does NumPy really define an ordering on complex numbers?
> 
strangely, yes, a lexicographic one:
```python
>>> import numpy as np
>>> zarray = np.array([[1+2j,1+3j],[5+6j,3+8j]])
>>> zarray
array([[1.+2.j, 1.+3.j],
       [5.+6.j, 3.+8.j]])
>>> zarray.max()
(5+6j)
>>> np.sort(zarray)
array([[1.+2.j, 1.+3.j],
       [3.+8.j, 5.+6.j]])

```

jpivarski:
Note (affects all open PRs, of which this is the only one): #645 added a new submodule, which might require special handling in merging. Since I'm not comfortable with advanced git, I just cloned a new copy of the repo with `--recursive` to get everything and now I'm merging it into my PR.

include/awkward/Reducer.h
jpivarski: In that case, I guess that's what we have to do for compatibility. Do you already have that lexicographic sorting implemented for complex numbers? If so, then that's what should be included.

If you don't already have it implemented, it would be fair to `throw std::runtime_error("sorting complex numbers not implemented yet")`.

include/awkward/array/NumpyArray.h
jpivarski: ok

jpivarski (commented):
I just changed the parameter names in `ak.to_json` and `ak.from_json` and ran black and flake8.

I'm going to do the rest of this review by testing a bunch of cases that I think should work (I already tested a few and found a few that didn't). It's pretty hard to see mistakes by scanning the code—which I've done and it looks fine apart from the parameter names I just fixed—but it seems easier to find issues by writing test examples, so that's how I'll do this review.

My next commit will break the build because it will include assertions that are supposed to work but fail. So far, everything I've found looks easy enough to fix.

jpivarski:
I think you need to include something in Builder.h for Windows to find "`complex` in namespace `std`".

jpivarski (approved):
This next build will fail because I added a suite of tests (tests/test_0652-tests-of-complex-numbers.py) that check all the corner cases I could think of. Once the tests pass again, it will be ready to merge. (That's why this review is "approve.")

ianna:
@jpivarski - here is a very simple solution :-)
```python
    array = ak.from_json(
        '[{"r": 1.1, "i": 1.0}, {"r": 2.2, "i": 2.0}]', complex_record_fields=("r", "i")
    )
    assert (array["r"] + array["i"]*1j).tolist() == [(1.1 + 1j), (2.2 + 2j)]
```
The rest is done.

tests/test_0652-tests-of-complex-numbers.py
```diff
@@ -0,0 +1,214 @@
...
+def test_from_json():
+    array = ak.from_json('[{"r": 1.1, "i": 1.0}, {"r": 2.2, "i": 2.0}]')
+    assert array.tolist() == [
+        {"r": 1.1, "i": 1.0},
+        {"r": 2.2, "i": 2.0},
+    ]
+    array = ak.from_json(
+        '[{"r": 1.1, "i": 1.0}, {"r": 2.2, "i": 2.0}]', complex_record_fields=("r", "i")
+    )
+    assert (array["r"] + array["i"]*1j).tolist() == [(1.1 + 1j), (2.2 + 2j)]
```
jpivarski: The logic that performs this substitution should be inside the `ak.from_json` Python function, though not inside the `Content::fromjson` C++ function. The test should be reverted to

```python
    array = ak.from_json(
￼        '[{"r": 1.1, "i": 1.0}, {"r": 2.2, "i": 2.0}]', complex_record_fields=("r", "i")
￼    )
￼    assert array.tolist() == [(1.1 + 1j), (2.2 + 2j)]
```

The `ak.from_json` function can be made to do this using the trick you described. To get it to work at any level, such as in this test:

```python
    array = ak.from_json(
￼        '[[{"r": 1.1, "i": 1.0}, {"r": 2.2, "i": 2.0}], [], [{"r": 3.3, "i": 3.0}]]', complex_record_fields=("r", "i")
￼    )
￼    assert array.tolist() == [[(1.1 + 1j), (2.2 + 2j)], [], [(3.3 + 3j)]]
```

by using the internal utility function [ak._util.recursively_apply](https://github.com/scikit-hep/awkward-1.0/blob/main/src/awkward/_util.py#L1068-L1076). Here is a sample usage:

https://github.com/scikit-hep/awkward-1.0/blob/3908bff17f4552b4dad3299cfcee66a1aef77735/src/awkward/operations/structure.py#L574-L590

In the above, the `getfunction` only returns non-None when it encounters a RecordArray at some depth. Yours would return non-None when it encounters a RecordArray with at least the fields specified by `complex_record_fields`. The above returns a zero-argument lambda that returns a modified RecordArray. Yours would return a NumpyArray of complex type.

There's an open question of what to do if you find a record with fields specified by `complex_record_fields` that are not one-dimensional NumpyArrays. Should the substitution just not apply or should it raise an error? I'm inclined to think it should just not apply.

Also, although the

```python
return lambda: recordnode[complex_record_fields[0]] + recordnode[complex_record_fields[1]]*1j
```

is slick, the following crosses all the "t"s and dots all the "i"s for alternate backends:

```python
real = recordnode[complex_record_fields[0]]
imag = recordnode[complex_record_fields[1]]
nplike = ak.nplike.of(real, imag)
real_array = nplike.asarray(real)
imag_array = nplike.asarray(imag)
float_dtype = ak.nplike.numpy.result_type(real_array.dtype, imag_array.dtype, np.dtype(np.float32))
out = nplike.empty(2*len(recordnode), dtype=float_dtype)
out[::2] = real_array
out[1::2] = imag_array
complex_dtype = {
    np.dtype(np.float32): np.complex64,
    np.dtype(np.float64): np.complex128
}[float_dtype]
return lambda: ak.layout.NumpyArray(out.view(complex_dtype))
```

I've been thinking about it and the slick solution is probably just as good as this long one. I was worried that the slick solution would over-promote the complex type, returning a `np.complex128` when a `np.complex64` is desired, but the ArrayBuilder always returns the largest floating-point type, so this isn't an issue.

I guess you should stick with the slick solution: multiplication and addition. It's certainly easier to read.

tests/test_0652-tests-of-complex-numbers.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # Somewhere in from_json, a handler that turns integer record fields into
+    # parts of a complex number is missing.
+    array = ak.from_json(
+        '[{"r": 1, "i": 1}, {"r": 2, "i": 2}]', complex_record_fields=("r", "i")
+    )
+    assert (array["r"] + array["i"]*1j).tolist() == [(1 + 1j), (2 + 2j)]
```
jpivarski: Same as above.

tests/test_0652-tests-of-complex-numbers.py
```diff
@@ -0,0 +1,214 @@
...
+
+    # # This should fail with some message like "complex number fields must be numbers,"
+    # # not "called 'end_record' without 'begin_record' at the same level before it."
+    # with pytest.raises(ValueError) as err:
+    #     array = ak.from_json(
+    #         '[{"r": [], "i": 1}, {"r": [1, 2], "i": 2}]',
+    #         complex_record_fields=("r", "i"),
+    #     )
+    #     assert array["r"].type == array["i"].type
+    # assert "Complex number fields must be numbers" in str(err)
```
jpivarski: This test can probably be turned on again, now that the error message is handled in Python.

And also, it closes what I called above an "open question" about how to handle fields that aren't one-dimensional NumpyArrays. When I wrote that comment, I thought it should be an error. Sure, it could be an error. I'll agree with myself.

tests/test_0652-tests-of-complex-numbers.py
```diff
@@ -0,0 +1,214 @@
...
+        '[{"i": 1.1}, {"i": 2.2}]', complex_record_fields=("r", "i")
+    ).tolist() == [{"i": 1.1}, {"i": 2.2}]
+
+    # In this one, the extra field should simply be ignored. A record with *at least*
+    # the two specified fields should be recognized as a complex number, so that
+    # the protocol can include a type marker, as some protocols do.
+    array = ak.from_json(
+        '[{"r": 1.1, "i": 1.0, "another": []}, {"r": 2.2, "i": 2.0, "another": [1, 2, 3]}]',
+        complex_record_fields=("r", "i"))
+    assert (array["r"] + array["i"]*1j).tolist() == [(1.1 + 1j), (2.2 + 2j)]
```
jpivarski: Same as above.

jpivarski (commented):
See inline.

tests/test_0652-tests-of-complex-numbers.py
ianna: Oh, I'm missing something needed for multiplication:
```python
TypeError: unsupported operand type(s) for *: 'awkward._ext.NumpyArray' and 'complex'
```

jpivarski: All of the operator overloads like `__mul__` are implemented on `ak.Array`, not the layout node classes. But this could be a better way to do it than re-wrapping those NumpyArrays as `ak.Arrays`:

```python
nplike = ak.nplike.of(recordnode)
real = nplike.asarray(recordnode[complex_record_fields[0]])
imag = nplike.asarray(recordnode[complex_record_fields[1]])
return lambda: ak.layout.NumpyArray(real + imag*1j)
```

This avoids a lot of unnecessary infrastructure that `ak.Array` has to go through to implement `__add__` and `__mul__` on arbitrary structures. Since we know that these are both 1-dimensional NumPy arrays with equal lengths, casting them as such is more direct. (Although they might be GPU arrays to be handled by CuPy, but that's why we use `nplike`.)

tests/test_0652-tests-of-complex-numbers.py
ianna: done

tests/test_0652-tests-of-complex-numbers.py
ianna: done

tests/test_0652-tests-of-complex-numbers.py
ianna: done

tests/test_0652-tests-of-complex-numbers.py
ianna: @jpivarski - thanks! It looks nicer. Done and dusted :-)

src/libawkward/builder/UnionBuilder.cpp
```diff
@@ -195,6 +196,49 @@ namespace awkward {
...
+      if (tofill.get() == nullptr) {
+        i = 0;
+        for (auto content : contents_) {
+          if (dynamic_cast<Int64Builder*>(content.get()) != nullptr) {
+            tofill = content;
+            break;
+          }
+          i++;
+        }
+        if (tofill.get() != nullptr) {
```
ianna, line 236: I think, I'm missing `Float64Builder` here

src/python/content.cpp
```diff
@@ -817,6 +826,10 @@ builder_fromiter(ak::ArrayBuilder& self, const py::handle& obj) {
   else if (py::isinstance<py::float_>(obj)) {
     self.real(obj.cast<double>());
   }
+  // FIXME: should be something like py::isinstance<py::complex_>(obj)
+  else if (obj.attr("__class__").attr("__name__").cast<std::string>() == "complex") {
```
ianna: This should be something like `py::isinstance<py::complex_>(obj)`, but `pybind11` didn't like `py::complex_`. It does know `std::complex`, though.
@jpivarski, @henryiii - any suggestions?

tests/test_0395-complex-type-arrays.py
ianna: Ha... I forgot about this. I'm debugging it... I'm going to push one more commit to pretty-print complex NumpyArrays.

The following example:
```python
content_float64 = ak.layout.NumpyArray(
  np.array([0.25, 0.5, 3.5, 4.5, 5.5], dtype=np.float64)
)
content_complex64 = ak.values_astype(content_float64, "complex64", highlevel=False)
array_complex64 = ak.layout.UnmaskedArray(content_complex64)
```
it creates an
```python
<UnmaskedArray>
    <content><NumpyArray format="Zf" shape="5" data="0x 0000803e 00000000 0000003f 00000000 ... 00009040 00000000 0000b040 00000000" at="0x7f9979c44cf0"/></content>
</UnmaskedArray>
```
that is interpreted by `ak.to_list` as an `ak.layout.Content` or `ak.partition.PartitionedArray`. Then `ak.to_list` is iterating over the `bool`, `str`, `bytes`, or `numbers.Number` receiving complex numbers:
```python
to_list 
(0.25+0j)
 is bool, str, bytes, or numbers.Number
to_list 
(0.5+0j)
 is bool, str, bytes, or numbers.Number
to_list 
(3.5+0j)
 is bool, str, bytes, or numbers.Number
to_list 
(4.5+0j)
 is bool, str, bytes, or numbers.Number
to_list 
(5.5+0j)
 is bool, str, bytes, or numbers.Number
```
this results in:
```python
assert ak.to_list(content_complex64) == [0.25, 0.5, 3.5, 4.5, 5.5]
```
while the following works correctly:
```python
assert ak.to_list(ak.nplike.of(array_complex64).asarray(array_complex64)) == [(0.25+0.j), (0.5 +0.j), (3.5 +0.j), (4.5 +0.j), (5.5 +0.j)]
```
because `ak.to_list` is called on  `np.ndarray`:
```diff
to_list 
[0.25+0.j 0.5 +0.j 3.5 +0.j 4.5 +0.j 5.5 +0.j]
np.ndarray
```
Given an example of 
```python
    content_complex128 = ak.values_astype(content_float128, "complex128", highlevel=False)
    array_complex128 = ak.layout.UnmaskedArray(content_complex128)
    assert ak.to_list(content_complex128) == [0.25, 0.5, 3.5, 4.5, 5.5]
```
where
```
to_list 
<NumpyArray format="Zd" shape="5" data="0x 00000000 0000d03f 00000000 00000000 ... 00000000 00001640 00000000 00000000" at="0x7fe6f1e68ca0"/>
ak.layout.NumpyArray
```
`ak.to_list` is fine with `ak.layout.NumpyArray`:
```diff
to_list 
<NumpyArray format="Zd" shape="9" data="0x 9a999999 9999f13f 9a999999 9999b93f ... cdcccccc cccc2340 00000000 00000000" at="0x7f9979c37290"/>
ak.layout.NumpyArray
```

ianna (commented):
@jpivarski - it looks like there is more work to do to complete this PR.

ianna:
@jpivarski - got it! it looks like `assert` treats `[(0.25+0j), (0.5+0j), (3.5+0j), (4.5+0j), (5.5+0j)]` as equal to `[0.25, 0.5, 3.5, 4.5, 5.5]`

Here is the real `ak.to_list` output:
```python
>>> import numpy as np
>>> import awkward as ak
>>> content_float64 = ak.layout.NumpyArray(
... np.array([0.25, 0.5, 3.5, 4.5, 5.5], dtype=np.float64)
... )
>>> content_float64
<NumpyArray format="d" shape="5" data="0.25 0.5 3.5 4.5 5.5" at="0x7f8ef843e820"/>
>>> content_complex64 = ak.values_astype(content_float64, "complex64", highlevel=False)
>>> content_complex64
<NumpyArray format="Zf" shape="5" data="0.25+0j 0.5+0j 3.5+0j 4.5+0j 5.5+0j" at="0x7f8f18404180"/>
>>> array_complex64 = ak.layout.UnmaskedArray(content_complex64)
>>> array_complex64
<UnmaskedArray>
    <content><NumpyArray format="Zf" shape="5" data="0.25+0j 0.5+0j 3.5+0j 4.5+0j 5.5+0j" at="0x7f8f18404180"/></content>
</UnmaskedArray>
>>> ak.to_list(content_complex64)
[(0.25+0j), (0.5+0j), (3.5+0j), (4.5+0j), (5.5+0j)]
>>> 
```

src/libawkward/builder/GrowableBuffer.cpp
```diff
@@ -142,4 +146,7 @@ namespace awkward {
   template class EXPORT_TEMPLATE_INST GrowableBuffer<uint64_t>;
   template class EXPORT_TEMPLATE_INST GrowableBuffer<float>;
   template class EXPORT_TEMPLATE_INST GrowableBuffer<double>;
+  template class EXPORT_TEMPLATE_INST GrowableBuffer<std::complex<float>>;
```
ianna, line 149: is it needed?

ianna (commented):
@jpivarski - I'm pretty happy with this PR. I'm not planning to work on it unless there are any requests. I left your comments in the tests. Please, feel free to remove them while reviewing it or leave it for me to do it later. Thanks!

src/python/content.cpp
henryiii:
```cpp
bool is_complex(py::object obj) {
    static py::object cmplx = py::module_::import("builtins").attr("complex");
    return py::isinstance(obj, cmplx);
}
```

Gives me:

```python
>>> import cppimport, setuptools, numpy
>>> cplx_check = cppimport.imp("cplx_check")
>>> cplx_check.is_complex(1)
False
>>> cplx_check.is_complex(1.2j)
True
>>> cplx_check.is_complex(numpy.complex128(1j))
True
```

You'll need to fix the above import to also support Python 2.

src/libawkward/array/NumpyArray.cpp
jpivarski: It was a good catch! This had been wrong and is now fixed.

src/libawkward/array/NumpyArray.cpp
jpivarski: Yes, that overshadowed a variable of a different type (`bool` vs `Index8`) in the parent scope, which isn't wrong, but it definitely isn't good style. Changing the name resolves that, so this is good.

tests/test_0395-complex-type-arrays.py
jpivarski: This looks okay. It is not necessary to have pretty-printing of all NumPy dtypes in `NumpyArray::tostring_part`. If a dtype is not one of a small set of types we need for debugging, it prints the hex of the data.

The important thing is that `ak.to_list` produces Python `complex` objects, and if I'm reading the above correctly, it does. If that's the case, then you can resolve this conversation.

src/python/content.cpp
jpivarski: In Python 2, it's

```c++
py::module_::import("__builtin__").attr("complex")
```

(singular, with double-underscores). It should be possible to try to import both modules—only one of them will exist.

jpivarski: You can try `"builtins"` first, then fall back to `"__builtin__"` by catching `py::import_error` (https://pybind11.readthedocs.io/en/stable/advanced/exceptions.html).

tests/test_0395-complex-type-arrays.py
ianna: Thanks! Yes, it does.

src/libawkward/builder/GrowableBuffer.cpp
jpivarski: Your Complex128Builder is undoubtedly using the `GrowableBuffer<std::complex<double>>`, but probably nothing is using `GrowableBuffer<std::complex<float>>`. But the same could be said of `GrowableBuffer<double>` and all of the integer types. Even if they're not being used right now, TypedArrayBuilder will need them (to build arrays with an explicit dtype), so we might as well leave them in.

jpivarski:
Remaining to-do:

   - [x] getting all the tests to work: I re-read the testing suite and it's correctly checking all the things I wanted it to check
   - [ ] https://github.com/scikit-hep/awkward-1.0/pull/652#pullrequestreview-577105992 is described as incomplete (the `builtins.complex` vs `__builtin__.complex` thing), but I don't see how the tests could pass if this were not complete. Is there anything left to do on this item?
   - ~~pretty-print NumpyArrays with complex numbers~~ There's no need to do this.

So it sounds like everything might be done. Let me know if that one item above is incomplete. If not, I'll merge this PR.

ianna:
> Remaining to-do:
> 
> * [x]  getting all the tests to work: I re-read the testing suite and it's correctly checking all the things I wanted it to check
> * [ ]  [#652 (review)](https://github.com/scikit-hep/awkward-1.0/pull/652#pullrequestreview-577105992) is described as incomplete (the `builtins.complex` vs `__builtin__.complex` thing), but I don't see how the tests could pass if this were not complete. Is there anything left to do on this item?
> * ~pretty-print NumpyArrays with complex numbers~ There's no need to do this.
> 
> So it sounds like everything might be done. Let me know if that one item above is incomplete. If not, I'll merge this PR.

Thanks, @jpivarski ! Yes, there is an open question if the following comparison though it works, could be done better `obj.attr("__class__").attr("__name__").cast<std::string>() == "complex"`
It looks from the builds that like both Python 2 and 3 are happy with it. If you think, it's ok to leave it asis, please, go ahead and merge it. Thanks!

src/python/content.cpp
jpivarski: I kept wavering on the necessity of this one: it works without it, but if a class happens to be named "`complex`" while not being the built-in Python `complex` type, then this would pass when it shouldn't. I just did the fix and will be pushing it soon.

jpivarski:
> Yes, there is an open question if the following comparison though it works, could be done better `obj.attr("__class__").attr("__name__").cast<std::string>() == "complex"`
> It looks from the builds that like both Python 2 and 3 are happy with it.

I kept wavering on the necessity of this one: it works without it, but if a class happens to be named "`complex`" while not being the built-in Python `complex` type, then this would pass when it shouldn't. I just did the fix and will be pushing it soon.

It looks like a significant merge will also be needed, so that might be another commit from me. Once those two commits go through and pass tests, I'll merge it.

Thanks!

src/python/content.cpp
henryiii: An if would be quite a lot cheaper than an exception, and a reader can instantly tell which one is Python 2 and which is Python 3. In fact, you can make this a compile-time `#if` and simply not do the check at runtime at all.

henryiii: `#if PY_VERSION_HEX < 0x03000000` or `#if PY_MAJOR_VERSION < 3`

jpivarski:
The merge to `main` turned out to be no big deal (something must have been bad in my local git, so I've refreshed it).

@ianna When the tests pass on this, I'll merge it, or you can do the honors!

src/python/content.cpp
jpivarski: That would work, too. I was about to say, "I'm not concerned about the performance in the Python <--> C++ interface," but this one happens to be in the `from_iter` loop, which can potentially be big. Okay, since I haven't merged the PR yet, I'll do it.

jpivarski: Also, a Python 2.7 test failed; apparently the ImportError wasn't getting caught. I just checked in a fix and I'm manually testing Python 2.7 as well.

jpivarski:
Both 3.8 and 2.7 worked in my local tests. Once it works in Azure, it can be merged.

ianna:
> The merge to `main` turned out to be no big deal (something must have been bad in my local git, so I've refreshed it).
> 
> @ianna When the tests pass on this, I'll merge it, or you can do the honors!

Thanks @jpivarski ! Please, go ahead and merge it. I'll be in a meeting for an hour or so :-) Thanks!

PR #654, opened 2021-01-18
Forth machine warnings cleanup

ianna:
@jpivarski - current commits clean up all warnings for my local setup:
```
Apple clang version 12.0.0 (clang-1200.0.32.28)
Target: x86_64-apple-darwin19.6.0
```


jpivarski:
Since this PR currently cleans the warnings and nothing else, could you rename it and merge it as-is? That way, I can work with the same warning-free starting point. Thanks!

ianna:
> Since this PR currently cleans the warnings and nothing else, could you rename it and merge it as-is? That way, I can work with the same warning-free starting point. Thanks!

Ok, I'll wait for the tests to pass

PR #658, opened 2021-01-19
run static analyser on Forth machine

ianna:
@jpivarski - these commits apply modifications suggested by `cppcheck` and fix the `ForthOutputBuffer` warnings. 

Excluding the `inconclusive` and `style` suggestions:
```
include/awkward/forth/ForthMachine.h:92:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_can_push' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      stack_can_push() const noexcept {
      ^
include/awkward/forth/ForthMachine.h:98:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_can_pop' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      stack_can_pop() const noexcept {
      ^
include/awkward/forth/ForthMachine.h:111:7: style:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_pop' can be const. [functionConst]
      stack_pop() noexcept {
      ^
include/awkward/forth/ForthMachine.h:373:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_cannot_push' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      stack_cannot_push() const noexcept {
      ^
include/awkward/forth/ForthMachine.h:379:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_cannot_pop' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      stack_cannot_pop() const noexcept {
      ^
include/awkward/forth/ForthMachine.h:385:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_cannot_pop2' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      stack_cannot_pop2() const noexcept {
      ^
include/awkward/forth/ForthMachine.h:391:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::stack_cannot_pop3' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      stack_cannot_pop3() const noexcept {
      ^
include/awkward/forth/ForthMachine.h:424:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::bytecodes_pointer_push' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      bytecodes_pointer_push(int64_t which) noexcept {
      ^
include/awkward/forth/ForthMachine.h:432:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::bytecodes_pointer_pop' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      bytecodes_pointer_pop() noexcept {
      ^
include/awkward/forth/ForthMachine.h:450:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::do_loop_push' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      do_loop_push(int64_t start, int64_t stop) noexcept {
      ^
include/awkward/forth/ForthMachine.h:459:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::do_steploop_push' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      do_steploop_push(int64_t start, int64_t stop) noexcept {
      ^
include/awkward/forth/ForthMachine.h:474:5: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::do_abs_recursion_depth' can be static (but you may consider moving to unnamed namespace). [functionStatic]
    do_abs_recursion_depth() const noexcept {
    ^
include/awkward/forth/ForthMachine.h:486:7: performance:inconclusive: Technically the member function 'awkward::ForthMachineOf::do_loop_is_step' can be static (but you may consider moving to unnamed namespace). [functionStatic]
      do_loop_is_step() const noexcept {
      ^
```
```
src/libawkward/forth/ForthMachine.cpp:988:33: style: Consider using std::find_if algorithm instead of a raw loop. [useStlAlgorithm]
        if (pair.first == name) {
                                ^
src/libawkward/forth/ForthMachine.cpp:1411:25: style: Consider using std::any_of algorithm instead of a raw loop. [useStlAlgorithm]
      if (name == word) {
                        ^
src/libawkward/forth/ForthMachine.cpp:1806:43: style: Consider using std::find_if algorithm instead of a raw loop. [useStlAlgorithm]
          if (pair.first == dtype_string) {
                                          ^
```

### DIFF
--- a/include/awkward/array/NumpyArray.h
+++ b/include/awkward/array/NumpyArray.h
@@ -174,7 +174,7 @@ namespace awkward {
                const std::vector<ssize_t>& strides,
                ssize_t byteoffset,
                ssize_t itemsize,
-               const std::string format,
+               const std::string& format,
                util::dtype dtype,
                const kernel::lib ptr_lib);
 
--- a/include/awkward/forth/ForthMachine.h
+++ b/include/awkward/forth/ForthMachine.h
@@ -56,7 +56,7 @@ namespace awkward {
       decompiled_at(int64_t bytecode_position, const std::string& indent="") const;
 
     /// @brief HERE
-    const std::vector<std::string>
+    const std::vector<std::string>&
       dictionary() const;
 
     /// @brief HERE
@@ -122,7 +122,7 @@ namespace awkward {
       variables() const;
 
     /// @brief HERE
-    const std::vector<std::string>
+    const std::vector<std::string>&
       variable_index() const;
 
     /// @brief HERE
@@ -146,7 +146,7 @@ namespace awkward {
       outputs() const;
 
     /// @brief HERE
-    const std::vector<std::string>
+    const std::vector<std::string>&
       output_index() const noexcept;
 
     /// @brief HERE
--- a/include/awkward/forth/ForthMachine.h
+++ b/include/awkward/forth/ForthMachine.h
@@ -56,7 +56,7 @@ namespace awkward {
       decompiled_at(int64_t bytecode_position, const std::string& indent="") const;
 
     /// @brief HERE
-    const std::vector<std::string>&
+    const std::vector<std::string>
       dictionary() const;
 
     /// @brief HERE
@@ -122,7 +122,7 @@ namespace awkward {
       variables() const;
 
     /// @brief HERE
-    const std::vector<std::string>&
+    const std::vector<std::string>
       variable_index() const;
 
     /// @brief HERE
@@ -146,7 +146,7 @@ namespace awkward {
       outputs() const;
 
     /// @brief HERE
-    const std::vector<std::string>&
+    const std::vector<std::string>
       output_index() const noexcept;
 
     /// @brief HERE
--- a/src/libawkward/forth/ForthMachine.cpp
+++ b/src/libawkward/forth/ForthMachine.cpp
@@ -646,7 +646,7 @@ namespace awkward {
   }
 
   template <typename T, typename I>
-  const std::vector<std::string>&
+  const std::vector<std::string>
   ForthMachineOf<T, I>::dictionary() const {
     return dictionary_names_;
   }
@@ -714,7 +714,7 @@ namespace awkward {
   }
 
   template <typename T, typename I>
-  const std::vector<std::string>&
+  const std::vector<std::string>
   ForthMachineOf<T, I>::variable_index() const {
     return variable_names_;
   }
@@ -783,7 +783,7 @@ namespace awkward {
   }
 
   template <typename T, typename I>
-  const std::vector<std::string>&
+  const std::vector<std::string>
   ForthMachineOf<T, I>::output_index() const noexcept {
     return output_names_;
   }
@@ -1505,8 +1505,9 @@ namespace awkward {
   }
 
   template <typename T, typename I>
-  void ForthMachineOf<T, I>::tokenize(std::vector<std::string>& tokenized,
-                                      std::vector<std::pair<int64_t, int64_t>>& linecol) {
+  void
+  ForthMachineOf<T, I>::tokenize(std::vector<std::string>& tokenized,
+                                 std::vector<std::pair<int64_t, int64_t>>& linecol) {
     IndexTypeOf<int64_t> start = 0;
     IndexTypeOf<int64_t> stop = 0;
     bool full = false;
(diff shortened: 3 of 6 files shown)


include/awkward/array/NumpyArray.h
```diff
@@ -174,7 +174,7 @@ namespace awkward {
                const std::vector<ssize_t>& strides,
                ssize_t byteoffset,
                ssize_t itemsize,
-               const std::string format,
+               const std::string& format,
```
jpivarski, line 177: Good catch, static analyzer! I've been trying to pass anything that isn't just a bool/number by reference.

include/awkward/forth/ForthMachine.h
```diff
@@ -56,7 +56,7 @@ namespace awkward {
       decompiled_at(int64_t bytecode_position, const std::string& indent="") const;
 
     /// @brief HERE
-    const std::vector<std::string>
+    const std::vector<std::string>&
       dictionary() const;
```
jpivarski: I didn't turn return values into references because (a) downstream users might modify my class's internal data (though maybe that could be prevented with `const std::vector<const std::string>`?) and (b) downstream users might own these data longer than my class exists. Case (b) can look like valid code but randomly segfault, and I don't even want to go there, especially for the limited value of passing a list of strings to Python more quickly (which Python will then use slowly). Rather than make this decision independently each time, I have a blanket policy of always returning copies. That's why there are methods like `variable_at(int index)` to get a variable's value quickly, rather than getting a copy of the whole vector just to pull out one value.

src/libawkward/forth/ForthMachine.cpp
```diff
@@ -302,7 +302,7 @@ namespace awkward {
       out << std::endl;
     }
     out << decompiled_segment(0);
-    return out.str();
+    return std::move(out.str());
```
jpivarski, line 305: This prevents copying a string while building the string that represents decompiled code. That's not an operation much in need of optimization, but I also don't see how this could become an ownership problem (it's okay if the returned string outlives the ForthMachine object), so I'll keep it.

src/libawkward/forth/ForthMachine.cpp
```diff
@@ -1282,8 +1282,8 @@ namespace awkward {
     else {
       int64_t which = current_which_[recursion_current_depth_ - 1];
       int64_t where = current_where_[recursion_current_depth_ - 1];
-      if (where < bytecodes_offsets_[which + 1] - bytecodes_offsets_[which]) {
-        return bytecodes_offsets_[which] + where;
+      if (where < bytecodes_offsets_[(IndexTypeOf<int64_t>)which + 1] - bytecodes_offsets_[(IndexTypeOf<int64_t>)which]) {
+        return bytecodes_offsets_[(IndexTypeOf<int64_t>)which] + where;
```
jpivarski, line 1286: I saw these from a previous PR. It seems like a long way to say "`size_t`". Is there a reason these shouldn't just be cast as `size_t`?

src/libawkward/forth/ForthMachine.cpp
```diff
@@ -1358,7 +1358,7 @@ namespace awkward {
       try {
         value = (int64_t)std::stoul(word.substr(2, word.size() - 2), nullptr, 16);
       }
-      catch (std::invalid_argument err) {
+      catch (std::invalid_argument& err) {
         return false;
       }
```
jpivarski, line 1363: I haven't been consistently catching references to exception objects, rather than copies. But there isn't a lot of `try`-`catch` logic in the codebase—this isn't something that needs to be made consistent just yet.

src/libawkward/forth/ForthOutputBuffer.cpp
```diff
@@ -527,7 +527,7 @@ namespace awkward {
                                         bool byteswap) noexcept {
     int64_t next = length_ + num_items;
     maybe_resize(next);
-    std::memcpy(&ptr_.get()[length_], values, sizeof(bool) * num_items);
+    std::memcpy(&ptr_.get()[length_], values, sizeof(bool) * (size_t)num_items);
```
jpivarski, line 530: Right, that's probably a warning...

jpivarski (approved):
Okay! I think I'll edit this PR to keep all of the changes except returning references to member data, for the reasons described inline.

Thanks!

ianna:
@jpivarski - Thanks! Shall I merge it?

jpivarski:
Yes! I was going to merge it as soon as tests pass, but I lost track of it among other things.

jpivarski:
Just to be clear, we always "squash and merge." I just pulled an update in main and it looks like 5 commits, whereas I was expecting 1 commit. I'll squash and merge this now.

jpivarski:
I see—I was too late. I don't plan to change any git history after the fact, but from now on, let's just always "squash and merge" so that every PR counts as one commit in the main branch. Thanks!

PR #661, opened 2021-01-20
AwkwardForth additions to make Avro and Parquet parsing possible.

jpivarski:


jpivarski:
This is the payoff: the reason why AwkwardForth was developed. Uproot will be able to read unsplit data as fast as ROOT does, without any compilation.

Uncompressed ROOT files → Awkward Arrays (warm cache)

ZLIB(1) ROOT files → Awkward Arrays (warm cache)

Pulling data from warm cache (i.e. already in RAM, using [vmtouch](https://hoytech.com/vmtouch/)) is limited only by the memory bus at about 1000 MB/sec—the ROOT test is slower because I'm not using ROOT's BulkIO, which I should do in a future version of this test. When the data are compressed, that becomes the bottleneck at about 100 MB/sec.

In past studies, the singly jagged case was about as fast as the non-jagged case because [that "jagged" was arrays](https://github.com/scikit-hep/uproot3#uproot), this "jagged" is `std::vector`. Skipping that header means a bit of NumPy work, which slows it down, but not as much as doubly jagged or triply jagged (non-NumPy code). These results settle a question I had in my mind: Uproot should use AwkwardForth even for the singly jagged case (so, not just `AsObjects`, but also `AsJagged`).

Here's what the jagged3 specialized Forth looks like:

```forth
input data
input byte_offsets
output offsets2 int32
output offsets1 int32
output offsets0 int32
output content float32

0 offsets2 <- stack
0 offsets1 <- stack
0 offsets0 <- stack

begin
  byte_offsets i-> stack
  6 + data seek
  data !i-> stack
  dup offsets2 +<- stack
  0 do
    data !i-> stack
    dup offsets1 +<- stack
    0 do
      data !i-> stack
      dup offsets0 +<- stack
      data #!f-> content
    loop
  loop
again
```

jpivarski:
Same thing for Avro. The [fastavro](https://fastavro.readthedocs.io/en/latest/) library is a precompiled Python extension, but it doesn't know the data type until runtime and it has to create Python objects, not Awkward Arrays. That's why Awkward Array can read Avro much faster. (A similar story would apply to Protobuf, etc.)

Uncompressed Avro files → Awkward Arrays (warm cache)

LZ4(1) Avro files → Awkward Arrays (warm cache)

Here's what the jagged3 specialized Forth looks like:

```forth
input stream
output offset0 int32
output offset1 int32
output offset2 int32
output content float32

0 offset0 <- stack
0 offset1 <- stack
0 offset2 <- stack

0 do
  stream zigzag-> stack
  dup offset0 +<- stack
  0 do
    stream zigzag-> stack
    dup offset1 +<- stack
    0 do
      stream zigzag-> stack
      dup offset2 +<- stack
      stream #f-> content
      stream b-> stack drop
    loop
    stream b-> stack drop
  loop
  stream b-> stack drop
loop
```

oshadura:
@jpivarski super cool! Out of curiosity, I think I am not familiar with what means here 'doubly' and 'triply' jagged?

jpivarski:
> @jpivarski super cool! Out of curiosity, I think I am not familiar with what means here 'doubly' and 'triply' jagged?

   * non-jagged: `float`
   * jagged: `std::vector<float>` (which has different performance in Uproot than `float[nValues]`, the "jagged" used by NanoAOD, because the `std::vector` has a header to skip; the [else clause in this code](https://github.com/scikit-hep/uproot4/blob/main/uproot/interpretation/jagged.py#L173-L201))
   * doubly jagged: `std::vector<std::vector<float>>`
   * triply jagged: `std::vector<std::vector<std::vector<float>>>`

jpivarski:
For completeness, here's the Parquet test.

Uncompressed Parquet files → Awkward Arrays (warm cache)

ZLIB(1) Parquet files → Awkward Arrays (warm cache)

Look at the code for unpacking repetition levels:

```forth
input stream
output replevels uint8

stream I-> stack
begin
  stream varint-> stack
  dup 1 and 0= if
    ( run-length encoding )
    stream {rle_format}-> replevels
    1 rshift 1-
    replevels dup
  else
    ( bit-packed )
    1 rshift 8 *
    stream #{bit_width}bit-> replevels
  then
  dup stream pos 4 - <=
until
```

and then for turning repetition levels into offsets:

```forth
input reps
output offsets0 int32 output offsets1 int32 output offsets2 int32
variable count0 variable count1 variable count2

begin
  reps b-> stack

  dup 3 = if
    1 count2 +!
  then
  dup 2 = if
    count2 @ offsets2 +<- stack
    1 count2 !
    1 count1 +!
  then
  dup 1 = if
    count2 @ offsets2 +<- stack
    1 count2 !
    count1 @ offsets1 +<- stack
    1 count1 !
    1 count0 +!
  then
  0 = if
    count2 @ offsets2 +<- stack
    1 count2 !
    count1 @ offsets1 +<- stack
    1 count1 !
    count0 @ offsets0 +<- stack
    1 count0 !
  then

  reps end
until

count2 @ offsets2 +<- stack
count1 @ offsets1 +<- stack
count0 @ offsets0 +<- stack
```

Very little about this needs to be specialized: it could as easily be a precompiled loop over N offset arrays. It doesn't benefit from specialization the way that ROOT (jagged > 1) and Avro do. Then we're left with the fact that the Forth VM is not as fast as compiled code, especially for a large number of instructions per iteration through the loop.

PR #664, opened 2021-01-24
Actually remember to release the GIL before doing some multithreading tests.

jpivarski:


jpivarski:
With this PR, the ForthMachine releases the Python GIL, so it scales. These were tested on a c5.18xlarge AWS instance (which has 72 cores). The ceiling might be the rate of memory access: my laptop maxes out at 2000 MB/sec.

AwkwardForth (future Uproot)

Of course, Uproot's original deserializer in Python doesn't scale at all.

Python (current Uproot)

jpivarski:
Ran it a few more times, averaged, and got a smoother curve.

AwkwardForth (future Uproot) (1)

jpivarski:
It looks like the (single-threaded) raw data copying rate is 2000 MiB/sec, so it's not surprising that we hit a wall at 5000 MiB/sec.

```python
>>> import numpy as np
>>> import time
>>> array = np.ones(10000*1024**2, np.uint8)
>>> begin = time.time(); array2 = np.copy(array); print(time.time() - begin)
4.9701831340789795
>>> del array2
>>> begin = time.time(); array2 = np.copy(array); print(time.time() - begin)
4.9057841300964355
>>> del array2
>>> begin = time.time(); array2 = np.copy(array); print(time.time() - begin)
4.90533709526062
>>> del array2
>>> begin = time.time(); array2 = np.copy(array); print(time.time() - begin)
4.904132127761841
```

jpivarski:
It has the same limiting read speed when multiple threads are reading the same array.

```python
>>> import numpy as np
>>> import time
>>> import concurrent.futures
>>> executor = concurrent.futures.ThreadPoolExecutor(2)
>>> begin = time.time(); tmp = list(executor.map(np.copy, [array] * 2)); print(time.time() - begin)
5.315511703491211
>>> del tmp
>>> begin = time.time(); tmp = list(executor.map(np.copy, [array] * 2)); print(time.time() - begin)
5.248187303543091
>>> del tmp
>>> executor = concurrent.futures.ThreadPoolExecutor(4)
>>> begin = time.time(); tmp = list(executor.map(np.copy, [array] * 4)); print(time.time() - begin)
5.308006763458252
>>> del tmp
>>> begin = time.time(); tmp = list(executor.map(np.copy, [array] * 4)); print(time.time() - begin)
5.11659049987793
>>> del tmp
>>> executor = concurrent.futures.ThreadPoolExecutor(8)
>>> begin = time.time(); tmp = list(executor.map(np.copy, [array] * 8)); print(time.time() - begin)
5.400676012039185
>>> del tmp
>>> begin = time.time(); tmp = list(executor.map(np.copy, [array] * 8)); print(time.time() - begin)
5.453733205795288
```

PR #665, opened 2021-01-25
Fix cleanup script

reikdas:
We shouldn't use an additional flag to delete `kernels.py` since that file is generated along with the tests.
Earlier, the script threw an error if `kernels.py` didn't exist.


jpivarski (approved):
LGTM!

PR #666, opened 2021-01-25
Remove unused variable from dev/generate-cuda.py

reikdas:


jpivarski (approved):
Same thing; this seems to be fine. Flake8 would have caught this, but I'm not sure if flake8 is running on the Python scripts in `dev`.

PR #667, opened 2021-01-25
Update CONTRIBUTING.md documentation

reikdas:
to reflect that `kernel-specification.yml` is now checked into git.


jpivarski (approved):
Also good. (This is a lot of PRs!)

PR #668, opened 2021-01-25
Write a type parser for TypedArrayBuilder

reikdas:


jpivarski:
This is good; doing it in `studies` for now.

jpivarski (commented):
I checked it out and tried some examples. Some of these didn't work, so I labeled them as `pytest.mark.skip`, but that makes them visible when you run

```bash
pytest -vv -rs parser.py
```

At least check to see how difficult it would be to add these. (I want to be sure that the structure of the parser will accommodate them.) Thanks!

reikdas:
@jpivarski All the tests you added pass now. Should we merge this into `studies/` now and then create a separate PR where we integrate it into the main codebase?

jpivarski:
> @jpivarski All the tests you added pass now. Should we merge this into `studies/` now and then create a separate PR where we integrate it into the main codebase?

Yes! You can merge it whenever you're ready.

PR #669, opened 2021-01-25
Avoid specifying target CUDA architecture

reikdas:
I think we should avoid specifying a target CUDA architecture unless there is a reason we need to.

I think CUDA >= 9 implies a lower bound of sm_30 - https://docs.nvidia.com/cuda/archive/9.0/cuda-compiler-driver-nvcc/index.html#ptxas-options so if we want to be verbose, maybe we should set it to `sm_30` instead?

### DIFF
--- a/cuda-build.sh
+++ b/cuda-build.sh
@@ -38,7 +38,7 @@ echo "cuda_version ='"$CUDA_VERSION"'" >> build/awkward_cuda_kernels/__init__.py
 echo "docker_image ='docker.io/nvidia/cuda:"$DOCKER_IMAGE_TAG"'" >> build/awkward_cuda_kernels/__init__.py
 
 export DOCKER_ARGS="-v`pwd`:/home -w/home docker.io/nvidia/cuda:"$DOCKER_IMAGE_TAG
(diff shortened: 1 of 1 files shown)


jpivarski (commented):
Since you've found a URL to the document that specifies the connection between CUDA 9 and `sm_30`, let's change the explicit bound to `sm_30` and include that URL in a comment to explain why. That way, this serves as technical documentation and an indication of why we've chosen these numbers. In the future, someone will probably want to bump it up to a newer CUDA and they'll learn from this that they should bump up the `sm_*` to a corresponding value.

reikdas:
Without the architecture specifically mentioned, we can just change the `CUDA_VERSION` variable in the `cuda-build.sh` script, to switch to a different CUDA version, because of this block - https://github.com/scikit-hep/awkward-1.0/blob/main/cuda-build.sh#L11-L30. The architecture would be the default architecture for that CUDA version.

If we specify the architecture and later change `CUDA_VERSION` to a version that is more than 9.0 and which doesn't support `sm_30`, I think there would probably be an error. The architecture would have to be changed as well. This might be unwanted behaviour since I think the architecture choice of `sm_30` or `sm_35` might still be a little arbitrary right now.

Should we still hardcode the `-arch sm_30` flag?

jpivarski:
Okay, fine: we'll just drop the `-arch sm_30` flag. I guess that means I'd accept this PR as-is. Give me a thumbs-up and I'll merge it (so that I don't do it while you're working on anything).

PR #670, opened 2021-01-25
fix warnings and run static analyser

ianna:
@jpivarski - please, consider the following commit that cleans up warning on: 
```
Apple clang version 12.0.0 (clang-1200.0.32.28)
Target: x86_64-apple-darwin19.6.0
Thread model: posix
```
and silences a `multiCondition` static analyser warning:
```
src/libawkward/forth/ForthOutputBuffer.cpp:127:13: style: Expression is always false because 'else if' condition matches previous condition at line 115. [multiCondition]
    else if (std::is_same<OUT, uint8_t>::value) {
            ^
src/libawkward/forth/ForthOutputBuffer.cpp:130:13: style: Expression is always false because 'else if' condition matches previous condition at line 118. [multiCondition]
    else if (std::is_same<OUT, uint16_t>::value) {
            ^
src/libawkward/forth/ForthOutputBuffer.cpp:133:13: style: Expression is always false because 'else if' condition matches previous condition at line 121. [multiCondition]
    else if (std::is_same<OUT, uint32_t>::value) {
            ^
src/libawkward/forth/ForthOutputBuffer.cpp:136:13: style: Expression is always false because 'else if' condition matches previous condition at line 124. [multiCondition]
    else if (std::is_same<OUT, uint64_t>::value) {
            ^
```


jpivarski:
I was just about to try to do that dependent project fix. Thanks for beating me to it!

jpivarski (approved):
All of these appear to be limiting the scope of include files, replacing include "density" with forwarded declarations, which would seem to always be a good thing, notwithstanding the error in the downstream dependent project.

If your fix to dependent-project makes the tests pass, then you can merge this PR. I approve!

ianna:
@jpivarski - FYI, by reducing the number of includes, especially in the headers, total build plus test time on my Mac went down from `1 min 40 sec` to `32 sec`.

jpivarski:
> @jpivarski - FYI, by reducing the number of includes, especially in the headers, total build plus test time on my Mac went down from `1 min 40 sec` to `32 sec`.

Wow—I'm looking forward to that! I thought most of the compilation time was spent in linking, not repeatedly compiling headers, but maybe.

PR #675, opened 2021-01-27
style: pre-commit

henryiii:
This is a basic pre-commit setup. It adds the following checks:

The "common checks":

- chore: add pre-commit
- style: run pre-commit

This is a set of checks for things like whitespace at the end of the lines.

- chore: add black
- style: run black

This adds Black.

I've also added a section in contributing.

- docs: pre-commit instructions

I can add ignore patterns if there are any files that need changes. Once ready, this PR should be rebased and merged (the "style" commits can be added to a `.git-blame-ignore-revs` eventually if we want)

@JPivarski, thoughts? Other committers?

I'd like to slowly introduce the checks from https://scikit-hep.org/developer/style , and also add style checking to CI in the future. But I think small steps are best, so just adding two things in this PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,9 +73,9 @@ python -m pytest -vv -rs tests-cuda
 
 ### Building documentation locally
 
-We use [Sphinx](https://pypi.org/project/Sphinx/) to generate documentation.  
+We use [Sphinx](https://pypi.org/project/Sphinx/) to generate documentation.
 
-You need some additional packages installed on your system to build the documentation - 
+You need some additional packages installed on your system to build the documentation -
 * [Doxygen](https://www.doxygen.nl/download.html)
 * [pycparser](https://pypi.org/project/pycparser/)
 * [black](https://pypi.org/project/black/)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,7 +36,7 @@ Most pull requests are merged with the "squash and merge" feature, so details ab
 
 It is unnecessary to manually edit (rebase) your commit history. If, however, you do want to save a pull request as multiple commits on `main`, ask me and we'll discuss. (The default branch is named `main`, not `master`.)
 
-### Building and teseting locally
+### Building and testing locally
 
 As described in [the README](https://github.com/scikit-hep/awkward-1.0#installation-for-developers), you can build locally using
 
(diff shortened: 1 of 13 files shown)


CONTRIBUTING.md
```diff
@@ -36,7 +36,7 @@ Most pull requests are merged with the "squash and merge" feature, so details ab
 
 It is unnecessary to manually edit (rebase) your commit history. If, however, you do want to save a pull request as multiple commits on `main`, ask me and we'll discuss. (The default branch is named `main`, not `master`.)
 
-### Building and teseting locally
+### Building and testing locally
```
henryiii, line 39: I added this one, the checks are not that smart. :)

studies/virtual_array_caching.py
```diff
@@ -2,15 +2,20 @@
...
+
+
 def generate():
     counter[0] += 1
-    dict_generate = {1: 'Ferrari', 2: 'Mercedes', 3: 'Red Bull', 4: 'McLaren'}
+    dict_generate = {1: "Ferrari", 2: "Mercedes", 3: "Red Bull", 4: "McLaren"}
     return dict_generate[counter[0]]
 
-class Cache():
-    def __init__(self, )
```
henryiii: This is the spot I had to correct the syntax to get the AST to pass.

jpivarski:
Could we exclude the `studies` and maybe also `docs-*` directories from these checks? The studies are completely free-form for figuring out the logic of a problem before committing it to mainline code. After having solved the logic, they're left for historical interest and maybe for figuring out a related problem. (I've used old studies in new studies from time to time.) The code in this directory is not necessarily even expected to work, which is why you found a syntax error.

Perhaps we could remove code from this directory, saying that it's all in history, but in practice, it can be hard to find the last commit that has something you're looking for. GitHub's search doesn't search all of the history, for instance. When I need to find an old study, I don't want it to be super-hard! (I won't necessarily know what it was named.)

Formatting the studies with Black is not only unnecessary, given what they are and what they're for, but it would add noise to this one PR, for instance, since this scratch-work is far from being properly formatted.

jpivarski:
You know, it's also putting newlines at the ends of SVG files. Those are made by Inkscape, and if edited by Inkscape, it will remove the newline again (ah, but the pre-commit would put it back...).

Inkscape is otherwise pretty good about minimizing diffs, in that it has an internally consistent style and maintains SVG element ids.

henryiii:
Sure, I can exclude svgs and those two directories.

CONTRIBUTING.md
jpivarski: Thanks!

jpivarski:
The test seem to be trailing whitespace in markdown, configuration files, and black in places where I hadn't thought of running it, like the dependent tests. I'm happy with the changes and with having them become automated.

Just not in the studies directory, and probably not for SVG files.

henryiii:
Reran with a global exclusion for `^docs.*` and `^studies.*` (regex), and removed the svg file type from the end-of-file fixer.

jpivarski:
Now that I'm looking at this on a computer and not a phone, I found one more: "tests/samples/test-two-arrays.json" was badly formatted on purpose, to test the JSON-reader's insensitivity to such issues. I see that `*.svg` have been excluded; `*.json` should be excluded as well, or else (or also) anything in the "tests/samples" directory, since these are all raw data and may have formatting issues on purpose.

henryiii:
As soon as this PR is merged, I'd like all contributors to try running pre-commit. It should be as simple as `brew install pre-commit` / `pipx install pre-commit` / `pip install pre-commit` (whatever you like to use for your main applications), and then just try `pre-commit run -a`. (or `pre-commit run --all-files`; I like descriptive names, but since I type this hundreds of times per day, short is better ;) )

We can set it up via CI, but ideally everyone will be able to run locally as well.

jpivarski:
I've installed pre-commit and will try

```bash
pre-commit run -a
```

on my next PR (when it's still small and doesn't have many changes).

jpivarski:
I scanned them again and I'm 100% happy with the changes. If you're not editing it locally, you can merge it or let me know that I can merge it.

henryiii:
:tada: Nope, not editing locally.

PR #677, opened 2021-01-27
docs: include GitHub button in GitHub dropdown

henryiii:
Also moves a few things to the standard locations per https://jupyterbook.org/customize/config.html

### DIFF
--- a/docs-src/_config.yml
+++ b/docs-src/_config.yml
@@ -11,6 +11,7 @@ launch_buttons:
 
 execute:
   allow_errors: false
+  execute_notebooks: auto  # one of ("auto", "force", "cache", "off")
 
 sphinx:
   config:
@@ -21,6 +22,7 @@ sphinx:
       repository_branch: main
       path_to_docs: docs-src
       use_edit_page_button: true
+      use_repository_button: true
       use_issues_button: true
       search_bar_text: Search...
       google_analytics_id: ""


henryiii:
My last local build might be missing the GitHub button (which was the point of this!), will check tomorrow.

henryiii:
Will revert to html_theme_options, opened an issue here:  https://github.com/executablebooks/jupyter-book/issues/1194

jpivarski:
A lot of these options used to not work outside of `sphinx_config`. JupyterBook has been changing too rapidly—I didn't like how the docs kept breaking because of upstream changes.

The configuration parameter that I want most is `show_navbar_depth: 2`, and I see that's still there. The second-tier topics in the left-bar were written to be visible at all times. They used to be grouped by section headers, but that feature was removed from JupyterBook, so my next-best approximation was to make them nested within fake pages whose content is a manually-maintained table of contents for that section, and then the default navbar depth was dropped to 1, making them hidden again. I'm afraid I might have adopted this too early!

When this is ready for review, I should check it out and look at the documentation it produces, since the visual layout is not something that can be caught by a test. (Note: even the "doctest-awkward (JupyterBook py38-np*)" is new. Until recently, we had no automated test to ensure that those notebook examples were correct! At least there's that.)

jpivarski:
> My last local build might be missing the GitHub button (which was the point of this!), will check tomorrow.

In some documentation, I've deliberately removed the "edit on GitHub" button because the sources were auto-generated and don't exist in the git repo. Not this one, though: every one of the pages in docs-src are hand-written (Jupytext, sometimes edited in Jupyter, sometimes manually). The "edit on GitHub" button would be a good addition, especially since any errors introduced into the code examples will be tested by the "doctest-awkward (JupyterBook)."

henryiii:
Edit on GitHub is already there, the "repository" button was the one missing.

Screen Shot 2021-01-27 at 11 29 05 AM

But I'd like to see:

Screen Shot 2021-01-27 at 11 29 24 AM

I'm testing locally, so I'll have to move a few things back until the linked Issue gets resolved in juptyer-book. The idea is to have no visual change expect for the addition of the missing menu item.

henryiii:
Eventually we might be able to refactor a bit, but this is simple and adds what I wanted to see. :/ Don't quite know why I was losing the button entirely with the version that put the values in the "standard" places, maybe it doesn't merge correctly, or maybe it was my error.

docs-src/_config.yml
```diff
@@ -11,6 +11,7 @@ launch_buttons:
 
 execute:
   allow_errors: false
+  execute_notebooks: auto  # one of ("auto", "force", "cache", "off")
```
henryiii, line 14: This is the default, but makes it easier to change it to "off" for testing locally.

jpivarski: I agree! `"cache"` is what I would usually use.

jpivarski:
I just checked it out, ran the documentation-generator, and visually inspected it. That new "repository" item in the drop-down menu correctly links to the main GitHub page. Also, the left-bar is still expanded, which is what I wanted to maintain.

PR #683, opened 2021-01-28
refactor: pulling static info into setup.cfg

henryiii:
Pulling out some static info into setup.cfg. Partially automated via a tool I'm working on, also contains the following changes:

* Find package refactored just a bit
* I've dropped the deprecated test_requires / test suite stuff.
* Fiddled with the URLs a bit
* Put the standard "Scikit-HEP" for the maintainers
* Fixed pins so Awkward supports building on macOS 11 (Intel) now

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,3 +1,64 @@
+[metadata]
+name = awkward
+url = https://github.com/scikit-hep/awkward-1.0
+project_urls =
+  Documentation = https://awkward-array.org
+  Source Code = https://github.com/scikit-hep/awkward-1.0
+  Bug Tracker = https://github.com/scikit-hep/awkward-1.0/issues
+  Discussions = https://github.com/scikit-hep/awkward-1.0/discussions
+  Chat = https://gitter.im/Scikit-HEP/awkward-array
+  Releases = https://github.com/scikit-hep/awkward-1.0/releases
+author = Jim Pivarski
+author_email = pivarski@princeton.edu
+maintainer = Scikit-HEP
+maintainer_email = scikit-hep-admins@googlegroups.com
+version = file: VERSION_INFO
+classifiers =
+    Development Status :: 5 - Production/Stable
+    Intended Audience :: Developers
+    Intended Audience :: Information Technology
+    Intended Audience :: Science/Research
+    License :: OSI Approved :: BSD License
+    Operating System :: MacOS :: MacOS X
+    Operating System :: Microsoft :: Windows
+    Operating System :: POSIX :: Linux
+    Operating System :: Unix
+    Programming Language :: Python
+    Programming Language :: Python :: 2.7
+    Programming Language :: Python :: 3.5
+    Programming Language :: Python :: 3.6
+    Programming Language :: Python :: 3.7
+    Programming Language :: Python :: 3.8
+    Programming Language :: Python :: 3.9
+    Topic :: Scientific/Engineering
+    Topic :: Scientific/Engineering :: Information Analysis
+    Topic :: Scientific/Engineering :: Mathematics
+    Topic :: Scientific/Engineering :: Physics
+    Topic :: Software Development
+    Topic :: Utilities
+license = BSD 3-clause
+description = Manipulate JSON-like data with NumPy-like idioms.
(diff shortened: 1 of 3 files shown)


setup.cfg
```diff
@@ -1,3 +1,64 @@
...
+  Documentation = https://awkward-array.org
+  Source Code = https://github.com/scikit-hep/awkward-1.0
+  Bug Tracker = https://github.com/scikit-hep/awkward-1.0/issues
+  Discussions = https://github.com/scikit-hep/awkward-1.0/discussions
+  Chat = https://gitter.im/Scikit-HEP/awkward-array
+  Releases = https://github.com/scikit-hep/awkward-1.0/releases
+author = Jim Pivarski
+author_email = pivarski@princeton.edu
+maintainer = Scikit-HEP
+maintainer_email = scikit-hep-admins@googlegroups.com
```
henryiii, line 14: @jpivarski Do you want to be listed as the maintainer, or would you like to put the standard Scikit-HEP snippet here? This is what the Scikit-HEP one looks like, can do either one.

jpivarski: References to my name as email and any copyright messages should transition toward Scikit-HEP. Since there's a standard email, let's go with it. Thanks!

henryiii:
I couldn't build this locally to check it; it turns out `wheel` is pinned in pyproject.toml to a version of wheel that does not support macOS 11. (It expects the version to have length 2, like 10.15, not 11).

I think we should unpin, probably just set a minimum (currently building in docker).

jpivarski:
> I couldn't build this locally to check it; it turns out `wheel` is pinned in pyproject.toml to a version of wheel that does not support macOS 11.

I don't think that there was a strong reason for pinning `wheel` to a particular version. It might have just been the version that was current last February.

henryiii:
`awkward-1.0.2.dist-info/entry_points.txt` looks okay, `awkward-1.0.2.dist-info/METADATA` seems reasonable , I'm going to say this is ready enough. If we are installing and testing the built wheels, then I'm completely happy. We will be soon enough if not. :)

setup.cfg
```diff
@@ -1,3 +1,64 @@
+[metadata]
+name = awkward
+url = https://github.com/scikit-hep/awkward-1.0
+project_urls =
+  Documentation = https://awkward-array.org
+  Source Code = https://github.com/scikit-hep/awkward-1.0
+  Bug Tracker = https://github.com/scikit-hep/awkward-1.0/issues
+  Discussions = https://github.com/scikit-hep/awkward-1.0/discussions
+  Chat = https://gitter.im/Scikit-HEP/awkward-array
+  Releases = https://github.com/scikit-hep/awkward-1.0/releases
```
henryiii, line 10: Other packages I was looking at often fill these out, it makes nice links on the sidebar.

setup.cfg
```diff
@@ -1,3 +1,64 @@
...
+packages = find:
+package_dir =
+    =src
+include_package_data = True
+zip_safe = False
+
+[options.packages.find]
+where = src
+exclude =
+    awkward_cuda_kernels
```
henryiii, line 55: You can include or exclude here.

setup.cfg
```diff
@@ -1,3 +1,64 @@
...
+
+[options.entry_points]
+numba_extensions =
+    init = awkward._connect._numba:register
```
henryiii, line 59: Just curious, how does numba handle name clashes here?

henryiii: I guess it's not a map, but more of an unordered list.

jpivarski: I wonder that myself. The space of type names is a global namespace. When two versions of Awkward are accidentally loaded (possible through an old `awkward1` and a new `awkward`), the attempt to register another type with the same name causes an exception on Numba-loading time. I've been thinking they're opening themselves to complaints about other people's bugs by including this entry-point. On the other hand, I think that might be true of any entry-point: if someone writes a bad package that uses your entry-point, it would fail when _your_ package loads, making it look like your bug.

setup.cfg
jpivarski: Okay!

jpivarski (approved):

PR #685, opened 2021-01-28
chore: flake8

henryiii:
Each fix is a commit, which the full error message from the check fixed.

This now includes the pre-commit flake8 check; it checks the last few items as well, #682 and #681. I've disabled quite a few checks, but hopefully over time we'll turn them back on; each disabled check has a note in the cfg.

Have all committers had a chance to play with pre-commit? I can turn on checking in CI once I'm pretty sure everyone is able to run locally, as I think it's a good skill to have before letting CI help with it. @ianna @reikdas @trickarcher 

Can be rebased and merged or squashed and merged, flake8 needs hand fixes, so not `style:` commits.

### DIFF
--- a/dev/generate-cuda.py
+++ b/dev/generate-cuda.py
@@ -81,10 +81,7 @@ def getthread_dim(pos):
     return code
 
 
-def traverse(node, args=None, forvars=None, declared=None):
-    args = args or {}
-    forvars = forvars or []
-    declared = declared or []
+def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
     if node.__class__.__name__ == "For":
         forvars.append(traverse(node.target, args, [], declared))
         if len(forvars) == 1:
--- a/dev/generate-cuda.py
+++ b/dev/generate-cuda.py
@@ -81,7 +81,10 @@ def getthread_dim(pos):
     return code
 
 
-def traverse(node, args={}, forvars=[], declared=[]):
+def traverse(node, args=None, forvars=None, declared=None):
+    args = args or {}
+    forvars = forvars or []
+    declared = declared or []
     if node.__class__.__name__ == "For":
         forvars.append(traverse(node.target, args, [], declared))
         if len(forvars) == 1:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,4 +24,4 @@ repos:
   rev: 3.8.4
   hooks:
   - id: flake8
-    additional_dependencies: [flake8-bugbear]
+    additional_dependencies: [flake8-bugbear, flake8-print]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,3 +19,9 @@ repos:
   rev: 20.8b1
   hooks:
   - id: black
+
+- repo: https://gitlab.com/pycqa/flake8
+  rev: 3.8.4
+  hooks:
+  - id: flake8
+    additional_dependencies: [flake8-bugbear]
--- a/dev/generate-tests.py
+++ b/dev/generate-tests.py
@@ -12,6 +12,8 @@ CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 
 
 class Argument(object):
+    __slots__ = ("name", "typename", "direction", "role")
+
     def __init__(self, name, typename, direction, role="default"):
         self.name = name
         self.typename = typename
--- a/localbuild.py
+++ b/localbuild.py
@@ -47,11 +47,11 @@ thisstate = {
 
 try:
     localbuild_time = os.stat("localbuild").st_mtime
-except:
+except Exception:
     localbuild_time = 0
 try:
     laststate = json.load(open("localbuild/laststate.json"))
-except:
+except Exception:
     laststate = None
 
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,12 @@
 norecursedirs = src pybind11 rapidjson dependent-project studies
 
 [flake8]
-ignore = E203, W503, E501, E266, N801, N802, N806
+# F401 can be fixed by adding __all__'s
+# E712 should be dropped, False/True should compared by "is" or checked as booleans, but not "=="
+# E711 should be dropped, always compare None with is, not ==.
+# F841 should be removed, use nothing or _ or _<name> for an unused variable
+# E402 can be improved in the future with isort
+# F403 should be removed, import * is evil
+# B007 should be removed, variables that are not used should start with _
+ignore = E203, W503, E501, E266, N801, N802, N806, F401, E711, E712, F841, E402, F403, B007
 max-complexity = 100
-exclude = .git, dev, tests, studies, pybind11, rapidjson, dlpack, dependent-project, docs-sphinx, setup.py, localbuild.py awkward/__init__.py awkward/layout.py awkward/types.py awkward/forms.py awkward/_io.py awkward/forth.py src/awkward/__init__.py src/awkward/layout.py src/awkward/types.py src/awkward/forms.py src/awkward/_io.py src/awkward/forth.py build
--- a/setup.cfg
+++ b/setup.cfg
@@ -2,6 +2,7 @@
 norecursedirs = src pybind11 rapidjson dependent-project studies
 
 [flake8]
+select = C,E,F,W,B,B9,T
 # F401 can be fixed by adding __all__'s
 # E712 should be dropped, False/True should compared by "is" or checked as booleans, but not "=="
 # E711 should be dropped, always compare None with is, not ==.
@@ -9,5 +10,12 @@ norecursedirs = src pybind11 rapidjson dependent-project studies
 # E402 can be improved in the future with isort
 # F403 should be removed, import * is evil
 # B007 should be removed, variables that are not used should start with _
-ignore = E203, W503, E501, E266, N801, N802, N806, F401, E711, E712, F841, E402, F403, B007
+ignore = E203, W503, E501, E266, N801, N802, N806, B950, F401, E711, E712, F841, E402, F403, B007
 max-complexity = 100
+exclude = studies, pybind11, rapidjson, dlpack, docs-*
+per-file-ignores =
+    tests/*: T
+    dependent-project/*: T
+    dev/*: T
+    setup.py: T
+    localbuild.py: T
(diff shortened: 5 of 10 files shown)


henryiii:
@jpivarski Do you care about being able to run flake8 manually, not through pre-commit? I generally keep black and mypy hand-runnable too, but usually don't bother to keep an ignore list for flake8 and just use pre-commit for it. It wasn't working all that well by hand.

jpivarski:
> @jpivarski Do you care about being able to run flake8 manually, not through pre-commit? I generally keep black and mypy hand-runnable, but usually don't bother to keep an ignore list for flake8 and just use pre-commit for it. It wasn't working all that well by hand.

I've been running black and flake8 manually. If the `pre-commit` command is equivalent, I suppose I won't need to. I was just compiling a list of directories to ignore...

jpivarski:
Should not be scanned for code quality:

   * dependency sub-modules: pybind11, rapidjson, dlpack
   * studies

Should not be prevented from having `print` statements (in addition to the above):

   * docs-*
   * dependent-project   (it's an example, not production code)

henryiii:
pre-commit can run everything, which frees you from running these by hand (and you can name a hook to run just that one - `pre-commit run -a flake8` will just run the flake8 check). Pre-commit is git aware, so there's no need to build ignore lists of things that are not in git.

Pre-commit is also smart; if you run `pre-commit run`, it only checks changed files, so it's much faster if you just want to check your staged changes.

So I could drop the flake8 jobs, and add a general pre-commit job? That will also cut our CI time if the flake8 job really does run on each workflow!

henryiii:
Pre-commit passes, it's only the manual flake8 jobs that are having problems, since they are not using the pre-commit ignores.

jpivarski:
Yes, I think that the pre-commit would replace the flake8 step in CI. It's not a significant change in CI time, since most of the time is in compilation.

The flake8-print check is an example of something that I wouldn't want to be automatically changed. Print statements in developer tools and _a very few_ places in the production code are intended. The few that this caught in tests were not intended and I'll remove them as part of this PR. (The problem there is that pytest hides stdout by default, so I forget to check for residual debugging code.)

henryiii:
Print is not bad in tests all the time - sometimes I'll print because the printout is easier to read than the pytest assert failure output - and it only shows up in tests. I didn't actually enable the print add-on in pre-commit (one benefit of pre-commit for flake8 - due to the entry-point based add ons, it depends on what you have installed - using pre-commit forces an exact, repeatable set of extensions).

I'll add the print check and tell it to be ignored in tests and anywhere else (and check to see what other flake8 extensions might be already included that I missed).

henryiii:
Flake8 _cannot_ auto-correct anything, it's a linter only.

dev/generate-cuda.py
```diff
@@ -81,7 +81,10 @@ def getthread_dim(pos):
     return code
 
 
-def traverse(node, args={}, forvars=[], declared=[]):
+def traverse(node, args=None, forvars=None, declared=None):
+    args = args or {}
+    forvars = forvars or []
+    declared = declared or []
```
henryiii: Was the code depending on the fact these were mutable defaults?

jpivarski: First, I'll ping the author, @reikdas, but I'll also look into it.

henryiii: I've attempted to make all passing explicit to see if that fixes it. Since the first call uses the defaults, all modifications after that change the default globally, so it's like passing these in when they are not listed there. If I get it working, @reikdas should still look it over.

jpivarski: Of the three, `forvars` and `declared` are mutated in place, which would have been a problem if `traverse` were called multiple times without somehow clearing those arguments. (I'm not really sure how you'd get access to the objects; maybe in `__code__` or something.)

The `args` variable could remain a dict argument, not that it matters all that much. But it's not safe for `forvars` and `declared` to be list arguments.

Good catch, whichever automated test this was. It's a highly recursive function—it could be difficult to walk to ensure that a caller never changes something in place.

jpivarski: I have a personal pet peeve against Python's "truthiness," which is used in

```python
forvars = forvars or []
```

I would say

```python
if forvars is None:
    forvars = []
```

You'll notice that I avoid putting an object in the predicate of an `if` statement unless it has `bool` type (as much as I'm aware, since I track types manually). In these expressions, the `or`-based assignment is clear enough, but in general it would be nice to always require explicit `· is not None`, `· != 0`, `len(·) != 0` checks.

henryiii: You get access by leaving the arguments out. Which is exactly what this did several times, for forvars and declared.

It's still bad for a function to modify it's mutable argument unless that is unmistakably part of the interface, but it's really bad if there's also a default. As a hard-and-fast rule: no mutable default arguments, ever (which is the contents of the check). 

```python
def f(x = []):
    x.append(1)
    print(x)

f() # -> [1]
f() # -> [1, 1]
```
This is what was happening when this was leaving off the final two arguments (which happened several times).

jpivarski: > I've attempted to make all passing explicit to see if that fixes it. Since the first call uses the defaults, all modifications after that change the default globally, so it's like passing these in when they are not listed there. If I get it working, @reikdas should still look it over.

Our messages crossed. In general, it's not a bad technique for the entry of a recursive call to create mutable lists and for these lists to be mutably appended through it, but your last commit was necessary to make that work, now that they're not global.

Functionally passing immutable objects down a recursive walk through a tree makes the changes apply to subtrees; non-functionally passing a mutable object and modifying it in place through the recurisve walk makes changes apply in the "reading order" through the tree. This is a walk over an AST of some code, where both cases can be relevant.

For instance, if you were walking over an AST of this code:

```python
def f(x):
    y = (x + 1)**2
    return y + 1
```

passing an immutable thing into the first Assign node, it wouldn't exist in the Return, but if you are in-place updating a list or dict, it's still available in the Return node. The immutable-passing is good for nested scopes, such as the scope of variables available in function `f`, but the mutable-passing is good for accumulating things, like which variables exist by the time you get to a particular node.

So it's not in principle bad that these variables are being changed in-place, if the latter feature is desired. Those variables do, however, have to be cleared or created anew at the beginning of recursion and always passed through all other steps of recursion. For things like this, I often separate the "entry point" of recursion with one function and the "continue walking" with another function. The "continue walking" function shouldn't have default arguments (to avoid errors like the one that Henry fixed), but the "entry point" often should, for a nice API.

henryiii: You have to be careful with Python truthiness, but there are specific places where it's safer than comparisons - for example, `==` is much more likely to throw than a truthiness check. I admit it's not as good as Ruby (where only False and Nil are "falsy"). In this case, the `or` trick kept me from having to repeat myself, though `forvars = [] if forvars is None else []` would make this completely explicit without providing too much extra boilerplate that a reader has to slog through. I'd be happy with either one. The `or` trick is seen in Python occasionally and is technically safe for any Optional collection, but it's more common in Ruby (due to the better true/false handling).

There's a PEP to add None aware operators, not sure if it ever will make it, but that would be nice.

jpivarski: > As a hard-and-fast rule: no mutable default arguments, ever (which is the contents of the check).

That is a good hard-and-fast rule. (Our messages crossed again.) You'd get that if the "entry point" is one function, with defaults, and the "continue walking" is another function. That "entry point" function probably doesn't even need `forvars` and `declared`, since they're probably just temporary variables for the recursion, not something that the interface needs.

This two-functions-for-recursion pattern comes up a lot. That was behind the motivation for separating the `ak.Array` interface from the `Content` tree nodes (separating the data structure into an "entry point" and "recursively composable" parts).

We probably won't hear from @reikdas until morning in India.  `:)`

henryiii: > In general, it's not a bad technique for the entry of a recursive call to create mutable lists and for these lists to be mutably appended through it, but your last commit was necessary to make that work, now that they're not global.

No, I only mean you should never depend on Python's default arguments for this. It's perfectly fine to create and pass a mutable structure (that's what I mean by "unless that is unmistakably part of the interface"). If it's clear that it's going to modify something in place, then that's fine. But mutable argument _defaults_ are very unexpected, and will cause hidden and very strange effects for the caller.

henryiii: Setting None as the default, and letting "None" start the chain is fine too. Though two functions make the interface tighter and explicit, which is better. :)

reikdas: It seems like there is a rather big bug in the code. I originally intended to pass the default values to `traverse` when not explicitly specified, i.e. `{}, [], []` but that is clearly not how it is supposed to work in Python. But doing that also does not work, so there is a logical error in my code somewhere that was being hidden. I _was_ mutating some arguments in place intentionally which I might need to refactor out to fix this. 

Sorry for the trouble. ~~I'll try and find a fix ASAP (perhaps I won't be able to provide it tonight, but I should be able to get a fix out by tomorrow).~~

Thanks for finding this! And also thanks for the above discussion - it is really relevant for my current project as well!

jpivarski: Hi @reikdas! It's not a giant rush, though this PR will have to wait for your fix.

It's funny that just after having this whole conversation, I ran into [the same issue on a StackOverflow answer](https://stackoverflow.com/a/59898173/1623645). I'll be a good citizen and edit that answer.

henryiii: I've put a `# noqa: B006` here so I can verify that this passes tests in its original form, and it's not something else causing problems (which is unlikely). If it passes, we could probably either get this in and then make a PR that removes the `noqa` comment and properly fixes this in the near future. Might need a `# TODO` or something like that to be extra sure it does not get missed? :)

It's a common mistake; it seems like it should work unless you realize that the function definition only happens once, including default arguments. And, if you don't mutate the arguments, it doesn't harm you - but as soon as you do, it's becomes very confusing.

jpivarski: This script is only executed when generating CUDA code, which is developer-initiated, not very frequent, and not permanent. The CUDA that it generated is going to be a "first draft" of the CUDA kernels as hand-written code.

So yes, just marking this `noqa` and moving on would also be acceptable, but we should remember this issue before the next time we generate CUDA code.

jpivarski:
Since @henryiii is done with this PR, I'll merge it, though @reikdas should remember that it has only be tagged as "don't complain," not necessary "is correct."

PR #688, opened 2021-01-29
Give lazy Parquet files information about their Forms.

jpivarski:


jpivarski:
All fields of the MillionSongDataset can be lazily read.

PR #690, opened 2021-02-01
array builder time profiler study

ianna:
@jpivarski - here is a small fix based on the time profiler report. The following test shows `5%` improvement :-)

Perhaps, it's not the most representative test. However, it looks like moving away from using a `shared_ptr` and replacing it with a `unique_ptr` could speed it up significantly. The `shared_ptr` destructor accounts for around `6%` in `ArrayBuilder::field_check()` and around `2%` in every other `ArrayBuilder` method: `integer()`, `real()`, `boolean()`, `beginrecord()`, `endrecord()`.

```c++
ak::ArrayBuilder myarray(ak::ArrayBuilderOptions(1024, 2.0));

  for (int64_t i = 0; i < 10000000; i++) {

    // populate builder with lists
    myarray.beginrecord();
    myarray.field_check("one");
    myarray.boolean(true);
    myarray.field_check("two");
    myarray.integer(1);
    myarray.field_check("three");
    myarray.real(1.1);
    myarray.endrecord();

    myarray.beginrecord();
    myarray.field_check("one");
    myarray.boolean(false);
    myarray.field_check("two");
    myarray.integer(2);
    myarray.field_check("three");
    myarray.real(2.2);
    myarray.endrecord();

    myarray.beginrecord();
    myarray.field_check("one");
    myarray.boolean(true);
    myarray.field_check("two");
    myarray.integer(3);
    myarray.field_check("three");
    myarray.real(3.3);
    myarray.endrecord();
  }
  // take a snapshot
  std::shared_ptr<ak::Content> array = myarray.snapshot();
```
Here is the report before the PR that showed that around 5% was spent in retrieving a key container size:
Screenshot 2021-02-01 at 15 59 49

Here is the result with this PR:
Screenshot 2021-02-01 at 16 29 11


jpivarski:
While the `ArrayBuilder` is building arrays, it does uniquely own the data, but then when `snapshot` is called, the passes a reference of its data to a new `Content`. In the initial design, both the `ArrayBuilder` and the `Content` own it from that point onward, so it has to be a `shared_ptr`. The `ArrayBuilder` can continue to append to the buffer while the `Content` uses its reference, though the `Content` assumes that it's only valid up to the `length` when the `snapshot` was taken. If the `ArrayBuilder` exceeds the allocation, it then makes a new buffer for itself and the `Content` ends up being the sole owner of the original. Two `snapshots` taken from the same `ArrayBuilder` might share the same buffer or they might not, depending on whether the `ArrayBuilder` had to grow the allocation in between `snapshots`. (That, by the way, is why we had to roll our own `GrowableBuffer` and couldn't just use `std::vector`. A `std::vector` doesn't provide its buffer as a `shared_ptr` and might delete it at any time.)

This policy could be changed for the sake of performance, but 5% or 6% doesn't sound like much to be interested in. Two other possible policies are:

   * `ArrayBuilder` has a `unique_ptr`, which it _copies_ into a `shared_ptr` in the `snapshot` operation.
   * `ArrayBuilder` has a `unique_ptr`, which it somehow upgrades into a `shared_ptr` when a `snapshot` is taken. Then both the `ArrayBuilder` and the `Content` both have a reference to the same `shared_ptr`. This might be unworkably complex because `ArrayBuilders` would have to work in two different modes: when it has `unique_ptr` and when it has `shared_ptr`.
   * `ArrayBuilder` has a `unique_ptr`, which it upgrades into a `shared_ptr` to give to a `Content` (the `Content` always must have `shared_ptr`), but then the `ArrayBuilder` ceases to exist or ceases to be usable. This new policy would disallow multiple snapshots. That's a loss of a feature, though the most common case is to snapshot only once.

If we know ahead of time that the best any of these new policies can give is a 5‒6% increase in speed, then I doubt it's worthwhile.

Your metric could be made a bit better with [field_fast](https://awkward-array.readthedocs.io/en/latest/_static/classawkward_1_1ArrayBuilder.html#a950563074bc8987b63196de7d9066041), rather than [field_check](https://awkward-array.readthedocs.io/en/latest/_static/classawkward_1_1ArrayBuilder.html#a81883f6d746612b422a32b7990b9450f). The "fast" method checks field name equivalence by comparing pointers (valid for `const char*` strings like these), whereas the "check" method does string equality checks each time. A `TypedArrayBuilder` can go one better by requiring fields to be provided in the same order each time—putting a boolean or a floating-point number into a field that's supposed to be an integer would always be an error, so assuming a constant order (which is almost always the case) becomes safer. [The logic that RecordBuilder goes through to verify that you're picking the right field is complex](https://github.com/scikit-hep/awkward-1.0/blob/1bfa9c60ba59d6e549011b287ebbcb49bebe5840/src/libawkward/builder/RecordBuilder.cpp#L408-L452); removing that in the typed case would surely be a win. (That logic starts by guessing that the order will be the same each time, so in the case when it is, it shouldn't suffer too much, but still.)

Also, unions can be improved in the `TypedArrayBuilder` case by requiring the user to provide a tag before filling the value, just as records and tuples require the user to provide a field name or number. That's not possible for the untyped `ArrayBuilder` because we don't even know _if_ data at a given level is going to be a union, but a `TypedArrayBuilder` would be initialized with the fact that the data should be a union (and it might not even be a union of different types, or it might be a union of mutual subtypes, like integer and floating-point, so this tag information would be absolutely required).

The API of `TypedArrayBuilder` will go a long way toward speeding things up, so chasing this 5‒6% probably isn't worth it. (We can consider `unique_ptr` and the one-snapshot rule for `TypedArrayBuilder`, though, now that we know it's the biggest bottleneck of `ArrayBuilder`.)

ianna:
@jpivarski - indeed, switching to `field_fast` gives around 4% improvement (see `3.7%` cost of `::compare`) :-)
Screenshot 2021-02-01 at 18 24 47

PR #697, opened 2021-02-02
check for an out of bounds index in a tuple builder

ianna:
@jpivarski - yes, it was observed both in C++ and in Python:
```python
>>> import awkward as ak
>>> a = ak.layout.ArrayBuilder()
>>> a.begintuple(2)
>>> a.index(2)
>>> a.boolean(True)
zsh: segmentation fault  python
```
now is fixed with "an out of bounds" message:
```python
>>> import awkward as ak
>>> a = ak.layout.ArrayBuilder()
>>> a.begintuple(2)
>>> a.index(2)
ValueError: 'index' 2 is out of bounds for a tuple with number of fields 2

(https://github.com/scikit-hep/awkward-1.0/blob/1.1.0rc2/src/libawkward/builder/TupleBuilder.cpp#L307)

```

### DIFF
--- a/src/libawkward/builder/TupleBuilder.cpp
+++ b/src/libawkward/builder/TupleBuilder.cpp
@@ -298,6 +298,14 @@ namespace awkward {
         std::string("called 'index' without 'begin_tuple' at the same level before it")
         + FILENAME(__LINE__));
     }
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Fantastic, thank you!

PR #711, opened 2021-02-08
test typed array builder prototype

ianna:
The array builder type is known at compile time. 
A given primitive type array builder accepts only its own type.
For example:
```c++
  auto options = ak::ArrayBuilderOptions(1024, 2.0);
  auto bool_builder = ak::BoolTypedBuilder(options, ak::GrowableBuffer<uint8_t>::empty(options));
  auto int_builder = ak::Int64TypedBuilder(options, ak::GrowableBuffer<int64_t>::empty(options));

  integer(bool_builder, 1);
  integer(int_builder, 1);

  boolean(bool_builder, true);
  boolean(int_builder, true);

  std::shared_ptr<ak::Content> bool_array = bool_builder.snapshot();
  std::shared_ptr<ak::Content> int_array = int_builder.snapshot();

  std::cout << bool_array.get()->tostring() << "\n";
  std::cout << int_array.get()->tostring() << "\n";

```
```log
BoolTypedBuilder does not have "integer" method
Int64TypedBuilder does not have "boolean" method
<NumpyArray format="?" shape="1" data="true" at="0x7fa9a9808800"/>
<NumpyArray format="l" shape="1" data="1" at="0x7fa9a9808c00"/>
```
or
```c++
  auto myarray = ak::TypedArrayBuilder<uint8_t, ak::BoolTypedBuilder>(ak::ArrayBuilderOptions(1024, 2.0));

  myarray.boolean(true);
  myarray.boolean(false);
  myarray.boolean(true);
  myarray.boolean(false);
  myarray.boolean(true);
  myarray.boolean(false);
  myarray.integer(1);
  myarray.real(1.1);
  myarray.null();
  myarray.boolean(false);

  std::cout << myarray.snapshot().get()->tostring() << "\n";

```

```log
BoolTypedBuilder does not have "integer" method
BoolTypedBuilder does not have "real" method
BoolTypedBuilder does not have "null" method
<NumpyArray format="?" shape="7" data="true false true false true false false" at="0x7fa9a980bc00"/>
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -105,6 +105,7 @@ addtest(test0016 tests/test_0016-finish-getitem-for-rawarray.cpp)
 addtest(test0019 tests/test_0019-use-json-library.cpp)
 addtest(test0030 tests/test_0030-recordarray-in-numba.cpp)
 addtest(test0074 tests/test_0074-argsort-and-sort-rawarray.cpp)
+addtest(test0707 tests/test_0710-typed-array-builder.cpp)
 
 # Third tier: Python modules.
 if (PYBUILD)
--- a/include/awkward/builder/TypedBuilder.h
+++ b/include/awkward/builder/TypedBuilder.h
@@ -1,0 +1,571 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+#ifndef AWKWARD_TypedBuilder_H_
+#define AWKWARD_TypedBuilder_H_
+
+#include <complex>
+#include <string>
+#include <vector>
+
+#include "awkward/common.h"
+#include "awkward/Content.h"
+#include "awkward/type/Type.h"
+#include "awkward/array/IndexedArray.h"
+#include "awkward/builder/ArrayBuilderOptions.h"
+#include "awkward/builder/GrowableBuffer.h"
+
+namespace awkward {
+
+  class Content;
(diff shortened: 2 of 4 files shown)


jpivarski:
The eventual TypedArrayBuilder will get its type at runtime, right? For it to be useful in Python, it still has to be compiled generically into a binary that we deliver through pip. At runtime and before filling, it can create a fixed number of output buffers and generate some Forth code specialized for its type, but that has to be entirely after the C++ compilation.

Maybe there were be reason someday for a third kind of ArrayBuilder, which takes templated types and compiles to a specialized machine at C++ compile time (raising compiler errors if it is used incorrectly, rather than slow exceptions), but not enough people are using the C++ interface yet to justify that now.

ianna:
@jpivarski - tested on `7 x 1 000 000 000` `BoolBuilder::boolean`

Initial change from a runtime type to determining properties of the type at compile time plus dropping the shared pointers
reduces the time from `2.62 min` to `58.85 s`:
Screenshot 2021-02-09 at 18 16 43
Screenshot 2021-02-09 at 18 17 27
Second step a `TypedGrowableBuffer` (also `unique_ptr` based) is reserving a large enough memory chunk (as a unique pointer `std::unique_ptr<T> ptr = kernel::unique_ptr_malloc<T>`), plus not checking if it needed a resize:
`GrowableBuffer::append` time went down from `48.60 s` to `7.54 s`:
Screenshot 2021-02-09 at 18 27 44

PR #720, opened 2021-02-09
fix: add missing files to the manifest, include a check

henryiii:
Following guidelines here: https://scikit-hep.org/developer/style#check-manifest

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,3 +25,9 @@ repos:
   hooks:
   - id: flake8
     additional_dependencies: [flake8-bugbear, flake8-print]
+
+- repo: https://github.com/mgedmin/check-manifest
+  rev: "0.46"
+  hooks:
+  - id: check-manifest
+    stages: [manual]
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,11 +1,19 @@
-include requirements*.txt VERSION_INFO CMakeLists.txt LICENSE README*.md CONTRIBUTING.md
+include requirements*.txt VERSION_INFO CMakeLists.txt LICENSE README*.md CONTRIBUTING.md runtime.txt
 recursive-include include *.h
-recursive-include src *.cpp
-recursive-include src/awkward *.py
+recursive-include src *.cpp *.py *.cu
 recursive-include tests *.cpp *.py samples/*
-include *.dll
+recursive-include tests-cpu-kernels *.py
+recursive-include tests-cuda-kernels *.py
+recursive-include tests-cuda *.py
+recursive-include tests-spec *.py
+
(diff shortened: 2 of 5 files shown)


henryiii:
1.1.0 had an incomplete SDist which broke Conda (and anyone not getting a wheel), this should fix it, and improve testing so it's less likely to happen again.

jpivarski:
That's a recent one! I just installed flake8-bugbear; I'll be using it in addition to the pre-commit.

I'm in favor of this PR: as soon as the tests pass, you can merge it.

henryiii:
Pre-commit runs the flake8 bugbear test. It handles everything for you, just run `pre-commit run -a`, and every check now and in the future will run. :) (though the check-manifest check I made manual because it's rather slow, you have to ask for it with `pre-commit run -a --hook-stage manual` (for all checks include manual ones), or just that check with `pre-commit run -a --hook-stage manual check-manifest` (all pre-commit checks can be run individually by id).

henryiii:
Force pushing that last commit again to see if I can get pre-commit.ci to trigger.

henryiii:
By the way, to see this bug without Conda, just run:


```bash
pip install awkward --no-binary awkward
```
For example, from `docker run -it --rm python:3.9 bash`. Or try a regular pip install from a non-manylinux Linux, like Alpine or ClearLinux. It will die at 95% with "fatal error: dlpack/dlpack.h: No such file or directory".

Will help with https://github.com/conda-forge/awkward-feedstock/pull/64 (once a fixed release is made).

henryiii:
I'll try to get this to fail if you do a regular build and install in the tests in the future. If I can't, we can add a test where we make an SDist then install it. But for now, this should fix the problem with Conda and source installs.

jpivarski:
> By the way, to see this bug without Conda, just run:
> 
> ```shell
> pip install awkward --no-binary awkward
> ```
> 
> For example, from `docker run -it --rm python:3.9 bash`. Or try a regular pip install from a non-manylinux Linux, like Alpine or ClearLinux. It will die at 95% with "fatal error: dlpack/dlpack.h: No such file or directory".
> 
> Will help with [conda-forge/awkward-feedstock#64](https://github.com/conda-forge/awkward-feedstock/pull/64) (once a fixed release is made).

The source distribution has been fixed several times in the past; the reason it's broken now is because dlpack was added. It's a recurrent issue.

henryiii:
This new check should help a lot. The tests I was mentioning should be another big help. I _thought_ the PEP 517 build was supposed to solve this, but for some reason it's not isolating. Will look into this when I work on the build system.

henryiii:
> It's a recurrent issue.

This is exactly why check-manifest is a recommendation in the Scikit-HEP developer docs. ;)

Issue #674, opened 2021-01-26
categorical twists with arrow (invalidation of valid DictionaryArray, wrong minimization of ChunkedArray, option ChunkedArray are non-categorical)


drahnreb:
Hi Jim, there are still problems with categoricals. This is a new issue but it is related to #400 and PR #403, which we discussed a few months ago.

When reading in partitioned parquets (e.g. by `year`, which produce categoricals, similar to 3.) I encountered several problems with categoricals debugging. Data doesn't seem to be corrupted but e.g. writing to parquet is affected.

I combined three bugs in one issue as they might be partly linked to each other and be addressed in a joint way. If you disagree please feel free to split the following:
1. Invalidation of valid categorical arrays
2. Minimization fails for `ChunkedArray` with categoricals using `ak.from_arrow()`
3. Option-typed `DictionaryArray` are treated as non-categorical awkward arrays

#### Minimal reproducible examples:
```python
>>> import awkward as ak
>>> import pyarrow as pa
>>> (ak.__version__, pa.__version__) # pyarrow 2.0 should fix ARROW-9801 and DictionaryArray indices look good
('1.0.2', '2.0.0')
```

**1.**
_Invalidation of valid categorical arrays_
```python
>>> original = ak.Array([2019, 2020, 2021, 2020, 2019])
>>> ak.is_categorical(original)
>>> categorical = ak.to_categorical(original)
>>> categorical.layout # looks fine
<IndexedArray64>
     <parameters>
         <param key="__array__">"categorical"</param>
     </parameters>
     <index><Index64 i="[0 1 2 1 0]" offset="0" length="5" at="0x56393ae35f60"/></index>
     <content><NumpyArray format="l" shape="3" data="2019 2020 2021" at="0x56393ab84550"/></content>
</IndexedArray64>
>>> ak.is_categorical(categorical) # is fine
True
>>>  ak.categories(categorical) # does fine
<Array [2019, 2020, 2021] type='3 * int64'>
>>> ak.is_valid(categorical) # but is invalid
False
>>> ak.is_valid(original)
True
```

**2.**
_Minimization fails for `ChunkedArray` with categoricals using ak.from_arrow()_
```python
>>> indices = pa.array([0, 1, 0, 1, 2, 0, 2])
>>> dictionary = pa.array([2019, 2020, 2021])
>>> dict_array = pa.DictionaryArray.from_arrays(indices, dictionary)
>>> # create a table with 3 chunkedarrays
[...]

PR #727, opened 2021-02-10
fix some issues of categorical arrays

drahnreb:
I tried my luck with a partial fix as part of #674.

- [x] 1. Validation of valid categorical arrays
- [x] 2. fix minimization for ChunkedArray with categoricals using ak.from_arrow()
- [x] 3. add parameter `__array__ = "categorical"` for option-typed categorical arrays
- [x] 4. add tests

Feedback very welcome.

You mentioned that it is purely on python level but I think the validation checks were C++ based. If I am mistaken please advise.

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -2860,9 +2860,12 @@ def _from_arrow(
                 arrays = [handle_arrow(batch) for batch in batches if len(batch) > 0]
                 return ak.operations.structure.concatenate(arrays, highlevel=False)
 
-        elif isinstance(obj, Iterable) and len(obj) > 0 and all(
-            isinstance(x, pyarrow.lib.RecordBatch) for x in obj
-        ) and any(len(x) > 0 for x in obj):
+        elif (
+            isinstance(obj, Iterable)
+            and len(obj) > 0
+            and all(isinstance(x, pyarrow.lib.RecordBatch) for x in obj)
+            and any(len(x) > 0 for x in obj)
+        ):
             chunks = []
             for batch in obj:
                 chunk = handle_arrow(batch)
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -2608,7 +2608,11 @@ def _from_arrow(
 
             if isinstance(index, ak.layout.BitMaskedArray):
                 return ak.layout.BitMaskedArray(
-                    index.mask, out, True, len(index), True,
+                    index.mask,
+                    out,
+                    True,
+                    len(index),
+                    True,
                     parameters={"__array__": "categorical"},
                 ).simplify()
             else:
@@ -2846,15 +2850,15 @@ def _from_arrow(
                 return ak.layout.RecordArray(child_array, obj.schema.names)
 
         elif isinstance(obj, pyarrow.lib.Table):
-            chunks = []
-            for batch in obj.to_batches():
-                chunk = handle_arrow(batch)
-                if len(chunk) > 0:
-                    chunks.append(chunk)
-            if len(chunks) == 1:
-                return chunks[0]
+            batches = []
+            for record_batch in obj.combine_chunks().to_batches():
+                batch = handle_arrow(record_batch)
+                if len(batch) > 0:
+                    batches.append(batch)
+            if len(batches) == 1:
+                return batches[0]
             else:
-                return ak.operations.structure.concatenate(chunks, highlevel=False)
+                return ak.operations.structure.concatenate(batches, highlevel=False)
 
         elif isinstance(obj, Iterable) and all(
             isinstance(x, pyarrow.lib.RecordBatch) for x in obj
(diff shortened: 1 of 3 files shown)


src/libawkward/Content.cpp
```diff
@@ -1827,22 +1827,32 @@ namespace awkward {
...
         return (std::string("at ") + path + std::string(" (") + classname()
                 + std::string("): __array__ = \"categorical\" only allowed for "
                               "IndexedArray and IndexedOptionArray"));
       }
+      if (!content.get()->is_unique()) {
+        return (std::string("at ") + path + std::string(" (") + classname()
+            + std::string("): __array__ = \"categorical\" requires contents "
+                            "to be unique"));
+      }
     }
```
jpivarski, line 1856: This is a proper correction. If an array is labeled `"categorical"`, then it is _only_ allowed to be an IndexedArray or an IndexedOptionArray, and it is the _contents_ that must be unique, not the array itself. It was wrong before, but I also don't see any tests complaining about the change, so it must not have been tested, either. Thanks for fixing that!

jpivarski (approved):

src/libawkward/Content.cpp
jpivarski: I see that you have other things to add to this PR; I won't merge it until you let me know that it's ready by un-drafting it. Then I'll check the next set of changes.

src/awkward/operations/convert.py
```diff
@@ -2846,15 +2850,15 @@ def handle_arrow(obj):
                 return ak.layout.RecordArray(child_array, obj.schema.names)
 
         elif isinstance(obj, pyarrow.lib.Table):
-            chunks = []
```
drahnreb, line 2848: I renamed this var to avoid confusion. `to_batches` creates RecordBatch obj that we concat.

src/awkward/operations/convert.py
```diff
@@ -2846,15 +2850,15 @@ def handle_arrow(obj):
...
-            for batch in obj.to_batches():
-                chunk = handle_arrow(batch)
-                if len(chunk) > 0:
-                    chunks.append(chunk)
-            if len(chunks) == 1:
-                return chunks[0]
+            batches = []
+            for record_batch in obj.combine_chunks().to_batches():
```
drahnreb: `.combine_chunks()` combines (possible) chunks of a table by creating a new one. Not sure if this might be resulting in memory issues.
Overall worth to think about treating Tables like RecordBatches and iterate over columns rather than converting to batch obj. After all, it is a columnar memory format. As such, each column could be type checked separately early on.

drahnreb (commented):
I fixed the ChunkedArray Problems. And added tests. Please check these changes.
I am not sure if the test coverage is sufficient.
Thanks.

src/awkward/operations/convert.py
jpivarski: I've checked out your code and I'm playing with it now. As I understand from [the documentation](https://arrow.apache.org/docs/python/generated/pyarrow.Table.html#pyarrow.Table.combine_chunks) and a few tests, `combine_chunks` ensures that you'll have zero or one RecordBatches, so Arrow does the concatenation and Awkward doesn't have to (line 2861).

If a table consists of exactly one RecordBatch (arrays are not ChunkedArrays):

```python
>>> single = pyarrow.Table.from_batches([
...     pyarrow.RecordBatch.from_arrays([ak.to_arrow(ak.Array(range(10)))], ["name"])
... ])
```

then calling `combine_chunks` on it does not allocate a new array. Arrow is smart enough to notice that there's nothing to concatenate, so it returns a view to the original.

```python
>>> np.frombuffer(single.to_batches()[0].column(0).buffers()[1], np.int64).ctypes.data
94720844398896
>>> np.frombuffer(single.combine_chunks().to_batches()[0].column(0).buffers()[1], np.int64).ctypes.data
94720844398896
```

This is not the case when there's multiple RecordBatches:

```python
>>> multi = pyarrow.Table.from_batches([
...     pyarrow.RecordBatch.from_arrays([ak.to_arrow(ak.Array(range(10)))], ["name"]),
...     pyarrow.RecordBatch.from_arrays([ak.to_arrow(ak.Array(range(10)))], ["name"]),
...     pyarrow.RecordBatch.from_arrays([ak.to_arrow(ak.Array(range(10)))], ["name"]),
... ])
>>> np.frombuffer(multi.to_batches()[0].column(0).buffers()[1], np.int64).ctypes.data
94720844573024
>>> np.frombuffer(multi.combine_chunks().to_batches()[0].column(0).buffers()[1], np.int64).ctypes.data
140673485410304
```

So I'm going to change the code to combine the chunks, check to see if the result has zero or one RecordBatches, and if it's one, just transform that without a concatenation.

jpivarski (approved):
I'll be making one change and then merging. Let me know if you have any objections. Thanks!

jpivarski:
> You mentioned that it is purely on python level but I think the validation checks were C++ based. If I am mistaken please advise.

You're right: sorry about that mistake.

drahnreb:
Looks good. Thanks

jpivarski:
@all-contributions please add @drahnreb for code and tests

jpivarski:
@all-contributors please add @drahnreb for code, tests

PR #737, opened 2021-02-12
Define the depth of an array of strings to be 1.

jpivarski:
So like this:

```python
>>> ak.Array(["one", "two", "three"])
<Array ['one', 'two', 'three'] type='3 * string'>
>>> ak.Array(["one", "two", "three"]).layout.purelist_depth
1
>>> ak.Array([["one", "two"], ["three"]]).layout.purelist_depth
2
>>> ak.Array([1, 2, 3]).layout.purelist_depth
1
>>> ak.Array([[1, 2], [3]]).layout.purelist_depth
2
```

The things that depend on it are exactly the things I'm working on now, so it's an issue that's just beginning to be relevant. Doing this later would be much harder.

@Dominic-Stafford, this addresses the issue you raised in #716.

### DIFF
--- a/src/awkward/behaviors/categorical.py
+++ b/src/awkward/behaviors/categorical.py
@@ -237,7 +237,6 @@ def to_categorical(array, highlevel=True):
     """
 
     def getfunction(layout):
-        p = layout.purelist_parameter("__array__")
         if layout.purelist_depth == 1:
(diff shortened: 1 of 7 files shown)


jpivarski:
(Note: the Uproot tests also pass. This won't be a surprise when it gets released.)

PR #762, opened 2021-02-21
More documentation, starting with "how to build".

jpivarski:


jpivarski:
This is just one tutorial, but anyway that's progress.

In addition, it has some bug-fixes that I want to get into main.

PR #767, opened 2021-02-23
Conceptual test of TypedArrayBuilder through AwkwardForth.

jpivarski:
@ianna, I found that I had to think through the logic of how the TypedArrayBuilder will interact with the ForthMachine to be able to write about it. For instance, from working through this I found out that TypedArrayBuilder will need a finite state machine in addition to using the ForthMachine as a bag of functions. Also, in a typed environment, `begin_list`, `begin_record`, etc. become superfluous and `field` or `index` (for setting the next record field or tuple index) can't be supported: it has to know which functions are coming next.

Wait a minute... maybe the finite state machine can be encoded in the ForthMachine after all...

### DIFF
--- a/studies/TypedArrayBuilder-forth-test.py
+++ b/studies/TypedArrayBuilder-forth-test.py
@@ -45,9 +45,9 @@ class TypedArrayBuilder:
             output part0-node4-data int64
 
             : node4-int64
-                ." node4-int64" cr
-                ( no pause because this is first after an implicit begin_list )
-                0 = if ( int64 )
+                ( no pause because this is first after a begin_list )
+
+                {int64} = if
                     0 data seek
                     data q-> part0-node4-data
                 else
@@ -56,11 +56,15 @@ class TypedArrayBuilder:
             ;
 
             : node3-list
-                ." node3-list" cr
+                pause
+                {begin_list} <> if
+                    halt
+                then
+
                 0   ( initialize the node3 list counter )
                 begin
                     pause
-                    dup 2 = if ( end_list )
+                    dup {end_list} = if
                         drop
                         part0-node3-offsets +<- stack ( absorbs node3 counter )
                         exit
@@ -72,9 +76,8 @@ class TypedArrayBuilder:
             ;
 
             : node2-float64
-                ." node2-float64" cr
-                ( no pause because this is first after an implicit begin_list )
-                1 = if ( float64 )
+                pause
+                {float64} = if
                     0 data seek
                     data d-> part0-node2-data
                 else
@@ -83,23 +86,31 @@ class TypedArrayBuilder:
             ;
 
             : node1-record
-                ." node1-record" cr
+                ( no pause because this is first after a begin_list )
+
+                {begin_record} <> if
+                    halt
+                then
+
                 node2-float64
                 node3-list
 
-                ( hmm... end_record is superfluous, too )
                 pause
-                3 <> if ( end_record )
+                {end_record} <> if
                     halt
                 then
             ;
 
             : node0-list
-                ." node0-list" cr
+                pause
+                {begin_list} <> if
+                    halt
+                then
+
                 0   ( initialize the node0 list counter )
                 begin
                     pause
-                    dup 2 = if ( end_list )
+                    dup {end_list} = if
                         drop
                         part0-node0-offsets +<- stack ( absorbs node0 counter )
                         exit
@@ -118,7 +129,7 @@ class TypedArrayBuilder:
                 node0-list
                 1+   ( increment the total counter )
             again
-        """)
+        """.format(int64=0, float64=1, begin_list=2, end_list=3, begin_record=4, end_record=5))
 
         self.data = np.empty(8, np.uint8)
         self.vm.run({"data": self.data})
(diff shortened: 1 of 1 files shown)


jpivarski:
@ianna I went through a few versions of this as you can see in the above commits. Now that we know the type in advance, `begin_record` and `end_record` are superfluous because we know we need to march from field x to field y, to field x to field y, etc.:

```python
# typedbuilder.begin_record()
typedbuilder.int64(1)
typedbuilder.float64(1.1)
# typedbuilder.end_record()

# typedbuilder.begin_record()
typedbuilder.int64(2)
typedbuilder.float64(2.2)
# typedbuilder.end_record()

# typedbuilder.begin_record()
typedbuilder.int64(3)
typedbuilder.float64(3.3)
# typedbuilder.end_record()
```

Also, `begin_list` is superfluous because when we reach a state of filling a list, we either insert a list item or end the list.

Oh! If we have a list of lists of lists, `begin_list` is necessary because we won't otherwise know which nested list we're closing! Let me go fix that.

jpivarski:
Okay, one last time: knowing types in advance makes all of the record-related commands irrelevant: `begin_record`, `end_record`, `field` (for picking out a field, we can't do them in arbitrary order easily in Forth, and removing the _capability_ also removes the _requirement_, with its associated speed-up), and `index` (same as `field`, but for tuples).

On the other hand, `begin_list` and `end_list` are required because we could have lists directly within lists (which didn't happen in my example), and we need to be able to distinguish `[[[]]]` from `[[]]`.

The latest version of the test has a justification for each `pause`, which will help to determine where to put them when generating the Forth code.

Union-types will take some thought. Option-types should be straightforward to add.

I think I can adapt your text into the paper now.

jpivarski:
This is a working example that I used to motivate what I wrote in the paper:

```python
import awkward.forth
import awkward as ak
import numpy as np

form = ak.forms.Form.fromjson("""
{
    "class": "ListOffsetArray32",
    "offsets": "i32",
    "content": {
        "class": "ListOffsetArray32",
        "offsets": "i32",
        "content": {
            "class": "ListOffsetArray32",
            "offsets": "i32",
            "content": {
                "class": "NumpyArray",
                "primitive": "float32",
                "form_key": "node3"
            },
            "form_key": "node2"
        },
        "form_key": "node1"
    },
    "form_key": "node0"
}
""")

class TypedArrayBuilder:
    def __init__(self, form):
        self.form = form

        def list_node(name, offsets, content_name):
            return """
: {name}
  {begin_list} <> if halt then
  0
  begin
    pause
    dup {end_list} = if
      drop
      {offsets} +<- stack
      exit
    else
      {content_name}
      1+
    then
  again
;
""".format(int64=0, float64=1, begin_list=2, end_list=3, **locals())

        self.vm = awkward.forth.ForthMachine32("""
input data
output offsets0 int32
output offsets1 int32
output offsets2 int32
output content float32

: node3
  {float64} = if
    0 data seek
    data d-> content
  else
    halt
  then
;

{node2} {node1} {node0}

0 offsets0 <- stack
0 offsets1 <- stack
0 offsets2 <- stack

0
begin
  pause
  node0
again
""".format(int64=0, float64=1, begin_list=2, end_list=3,
           node0=list_node("node0", "offsets0", "node1"),
           node1=list_node("node1", "offsets1", "node2"),
           node2=list_node("node2", "offsets2", "node3")))

        self.data = np.empty(8, np.uint8)
        self.vm.run({"data": self.data})

    def int64(self, x):
        self.data.view(np.int64)[0] = x
        self.vm.stack_push(0)
        self.vm.resume()

    def float64(self, x):
        self.data.view(np.float64)[0] = x
        self.vm.stack_push(1)
        self.vm.resume()

    def begin_list(self):
        self.vm.stack_push(2)
        self.vm.resume()

    def end_list(self):
        self.vm.stack_push(3)
        self.vm.resume()

    def snapshot(self):
        buffers = {
            "part0-node0-offsets": self.vm.outputs["offsets0"],
            "part0-node1-offsets": self.vm.outputs["offsets1"],
            "part0-node2-offsets": self.vm.outputs["offsets2"],
            "part0-node3-data": self.vm.outputs["content"],
        }
        return ak.from_buffers(self.form, self.vm.stack[0], buffers)

    def debug_step(self):
        print("stack: ", builder.vm.stack)
        for k, v in builder.vm.outputs.items():
            print(k + ":", np.asarray(v))
        print("array:", self.snapshot())
        print()

builder = TypedArrayBuilder(form)
builder.debug_step()

builder.begin_list()
builder.end_list()
builder.debug_step()

builder.begin_list()
builder.begin_list()
builder.end_list()
builder.end_list()
builder.debug_step()

builder.begin_list()
builder.begin_list()
builder.begin_list()
builder.end_list()
builder.end_list()
builder.end_list()
builder.debug_step()

builder.begin_list()
builder.begin_list()
builder.begin_list()
builder.float64(1.1)
builder.end_list()
builder.end_list()
builder.end_list()
builder.debug_step()

builder.begin_list()
builder.begin_list()
builder.begin_list()
builder.float64(1.1)
builder.float64(2.2)
builder.float64(3.3)
builder.end_list()
builder.end_list()
builder.end_list()
builder.debug_step()
```

PR #769, opened 2021-02-26
Typed Array Builder from Form

ianna:
@jpivarski - it should be moved to a new directory: `src/libawkward/typedbuilder`? Suggestions are welcome! Thanks!


ianna:
`TypedArrayBuilder` comparison with `ArrayBuilder` on records:
```python
In [1]: import awkward as ak                                                                                                                           

In [2]: form = ak.forms.RecordForm( 
   ...: {"one": ak.forms.NumpyForm([], 8, "d"), "two": ak.forms.NumpyForm([], 8, "d")} 
   ...: )                                                                                                                                              

In [3]: builder = ak.layout.TypedArrayBuilder(form)                                                                                                    

In [4]: %%timeit -n100000 -r1 
   ...: builder.float64(1.1) 
   ...: builder.float64(2.2) 
   ...:  
   ...:                                                                                                                                                
6.4 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 100000 loops each)

In [5]: array_builder = ak.ArrayBuilder()                                                                                                              

In [6]: %%timeit -n100000 -r1 
   ...: with array_builder.record(): 
   ...:     array_builder.field("one").real(1.1) 
   ...:     array_builder.field("two").real(2.2) 
   ...:                                                                                                                                                
15.1 µs ± 0 ns per loop (mean ± std. dev. of 1 run, 100000 loops each)

In [7]: array = array_builder.snapshot()                                                                                                               

In [8]: array2 = builder.snapshot()                                                                                                                    

In [9]: array                                                                                                                                          
Out[9]: <Array [{one: 1.1, two: 2.2}, ... two: 2.2}] type='100000 * {"one": float64, "tw...'>

In [10]: ak.Array(array2)                                                                                                                              
Out[10]: <Array [{one: 1.1, two: 2.2}, ... two: 2.2}] type='100000 * {"one": float64, "tw...'>


```

jpivarski:
More than 2× faster!

ianna:
> More than 2× faster!

just :-)

src/libawkward/typedbuilder/UnknownFormBuilder.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "awkward/typedbuilder/UnknownFormBuilder.h"
+#include "awkward/typedbuilder/TypedArrayBuilder.h"
+
+namespace awkward {
+
+  ///
+  UnknownFormBuilder::UnknownFormBuilder(const FormPtr& form)
+    : form_(form),
+      vm_empty_command_("( This does nothing. )\n"),
+      vm_error_("s\"Unknown Form Builder error\"") { }
```
jpivarski: The "unknown" type can only come from an EmptyArray/EmptyForm, right? If there is such a thing in the user-provided form, then any attempt to insert data at this level should be an error. For instance, if the form is `ListOffsetForm(ListOffsetForm(EmptyForm()))`, then it would be possible to make

```
[[], [[], [], []], [[], []], [], [[]]]
```

with an appropriate sequence of `begin_list`/`end_list`, but not

```
[[], [[], [], []], [[], []], [], [[1]]]
```

This is different from the use of "unknown" in ArrayBuilder. Every ArrayBuilder starts as unknown and any operation is valid on it because ArrayBuilder changes its type; TypedArrayBuilder doesn't change type, so no operation is valid on an unknown type.

(It wouldn't be particularly useful to use EmptyForms with TypedArrayBuilder; it's just here for completeness. The whole concept of unknown type, EmptyArray, and EmptyForm was added only for ArrayBuilder to have a starting point, and since it shouldn't be an error for some lists to never be filled, unknown type is not just an internal detail of ArrayBuilder.)

jpivarski (approved):
It looks great! Go ahead and merge it when you've finished the docstrings you're working on.

The high-level class can be a separate pull request.

src/libawkward/typedbuilder/UnknownFormBuilder.cpp
ianna: I've added this test and removed "unknown" array builder.

PR #773, opened 2021-03-01
Add type parser to main codebase

reikdas:


README.md
```diff
@@ -238,6 +238,8 @@ See the [prioritized list of open issues](https://github.com/scikit-hep/awkward-
 
 Support for this work was provided by NSF cooperative agreement OAC-1836650 (IRIS-HEP), grant OAC-1450377 (DIANA/HEP) and PHY-1520942 (US-CMS LHC Ops).
 
+We would also like to thank @erezsh and the developers of [lark](https://github.com/lark-parser/lark) for creating the [lark standalone parser generator](https://github.com/lark-parser/lark/blob/master/lark/tools/standalone.py), that we use in our type parser.
```
jpivarski: The authors capitalize "Lark": [their README](https://github.com/lark-parser/lark#readme), so we should do the same.

Let's use Erez Shinan's full name with GitHub handle in parentheses, maybe more like this:

> We also thank Erez Shinan (@erezsh) and the developers of the [Lark standalone parser](https://github.com/lark-parser/lark), which is used to parse type strings as type objects.

jpivarski (approved):
I've reviewed it and I think it's good! After changing the acknowledgement as I've suggested below, you can merge it. (If it triggers tests from the README change, which I think it _won't_, then please wait for the new tests to finish before merging.)

jpivarski:
It doesn't look like the at-sign automatically links in GitHub: https://github.com/scikit-hep/awkward-1.0/blob/978233d898ed5ffe8801441731199b0641a60e93/README.md

Since this is going through a whole test cycle anyway, perhaps you should explicitly link it to https://github.com/erezsh

jpivarski:
(If you don't catch this before the auto-merge, never mind. It's fine anyway.)

PR #775, opened 2021-03-03
This PR concentrates on covering all edge cases while doing a jax.jvp on Awkward Arrays.

trickarcher:


trickarcher:
Hi @jpivarski, I am facing a problem while indexing. In the test function if I do a `2 * array.y[0][1] or 2 * array.y[2][0]`, `JAX` gets the wrong values.I don't think `JAX` is the culprit here, since it always passes around the `aux_data` without changing it.

studies/DifferentiableArray-for-JAX.py
```diff
@@ -53,14 +54,21 @@ def layout(self, layout):
...
-        else:
-            return out
+        else: 
+            form, length, children = ak.to_buffers(self.layout)
+            child_buf_key = self.aux_data.datakeys.index(list(children.keys())[0])
+            if np.isscalar(out):  
+                if np.isscalar(where):
+                    return self.tracers[child_buf_key][where]
+                else:
+                    return NotImplementedError
```
jpivarski: What you want to do here is "if `out` is a scalar (non-`Content`),"

```python
if isinstance(where, (numbers.Integral, np.integer)) and isinstance(self.layout, ak.layout.NumpyArray):
    assert len(self.tracers) == 1
    return self.tracers[where]
else:
    raise NotImplementedError
```

That is, we can handle the "`where` is an integer" case when `out` is scalar and leave other cases for later. These other cases include

   * `where` is a string, which would mean that the `layout` has record type with field `where` being a primitive type
   * `where` is a tuple, which we can resolve by recursively running `__getitem__` on all tuple elements except the last (to get a non-scalar) and then feeding it the last
   * either of the above but there's an IndexedArray or option-type node wrapping the NumpyArray or RecordArray, which we simply have to unpack with `project()`.

But let's just handle this one "`where` is an integer and `layout` is a NumpyArray" case for now.

trickarcher: Oh okay I see. I was trying to get the correct tracer among the list of tracers in case of multiple buffers, when out is a scalar. I'll just comment out line `65 and 66` for now, Also, for Record Arrays the one test example, is giving wrong values for even when I am not trying to access the leaf values. So for the same test case. I get the value of the function using JAX as `[12, 12, 14.4]` instead of `[12, 14, 16]`. I printed out and tried figuring out the issue, and it's probably because of indexing. Would you know what might be the issue here? I wanted to bring up this issue now, because if it's really some indexing error, it won't be anything related to JAX, and we should fix that first before moving ahead.

jpivarski: I haven't looked closely at it. Print out object identifiers to find out what's getting called at every stage. It will be correct/what you want at each stage until you get to the wrong one. Don't treat it as a black box: bisect the bug.

trickarcher: Sure! I'll get on with it!

trickarcher: I found the offending line [here](https://github.com/scikit-hep/awkward-1.0/blob/f44e69ecc9b7d54bb5a65066b8a8875767390415/studies/DifferentiableArray-for-JAX.py#L61). It's happening there because new slices are coming in, without any knowledge of the previous slices and since self.tracers is always constant, in the last call where `indexes is {}` and that's the aux_data that is being passed into the special_unflatten function, the `VirtualArray` materializes the array based on the length. And that causes the wrong indexing. 

In the latest commit, I tried to keep a `list of where`, and use that list to index the original array in the final call of `special_unflatten`. This should work for all cases of element wise operations(It would be great if you can think of some cases that won't work), I am not very sure about reductions, though.

trickarcher: Also, a nice consequence of using the `where` list is that the scalars are no longer `DeviceArrays`, but integral values. As of now, indexing in the form of `[][][]` works perfectly. However, for all other cases it will raise a `NotImplementedError`. I have written a few more functions, which cover jvp on `RecordArrays` and `ListOffsetArrays` which work. I am currently writing a few more tests.

PR #779, opened 2021-03-08
Adapt getitem of DifferentiableArray for JAX

trickarcher:


trickarcher:
Hi @jpivarski , this code is ready for review.

jpivarski:
This is looking good! When you integrate it into the main codebase (next PR), put most of the code in an `ak._connect._jax` submodule that only gets called from `ak.highlevel`.

PR #781, opened 2021-03-09
AwkwardForth: add an s" core word to define strings

ianna:
User defined error messages can be passed to a `halt` command exception handling. For example:

```python
def test_user_defined_exception():
    vm32 = awkward.forth.ForthMachine32(
        """
variable x 0 x !
variable err
s" variable x reached its maximum 10"
: foo
    5 x +! x @
    dup 10 = if
        0 err ! err @ halt
    then
;
0
begin
    foo
    1+
again
"""
    )
    with pytest.raises(ValueError) as err:
        vm32.run()
    assert vm32.stack[-1] == 0
    assert vm32.string_at(vm32.stack[-1]) == "variable x reached its maximum 10"
```

### DIFF
--- a/src/libawkward/forth/ForthMachine.cpp
+++ b/src/libawkward/forth/ForthMachine.cpp
@@ -61,7 +61,7 @@ namespace awkward {
   #define CODE_LEN_OUTPUT 22
   #define CODE_REWIND 23
   // generic builtin instructions
-  #define CODE_STRING_ERROR 24
+  #define CODE_STRING 24
   #define CODE_PRINT_STRING 25
   #define CODE_PRINT 26
   #define CODE_PRINT_CR 27
@@ -570,7 +570,7 @@ namespace awkward {
           int64_t out_num = bytecodes_[(IndexTypeOf<int64_t>)bytecode_position + 1];
           return output_names_[(IndexTypeOf<int64_t>)out_num] + " rewind";
         }
-        case CODE_STRING_ERROR: {
+        case CODE_STRING: {
           int64_t string_num = bytecodes_[(IndexTypeOf<int64_t>)bytecode_position + 1];
           return "s\" " + strings_[(IndexTypeOf<int64_t>)string_num] + "\"";
         }
@@ -1270,13 +1270,9 @@ namespace awkward {
             "call 'begin' to 'step' again (note: check 'is_done')");
         }
         case util::ForthError::user_halt: {
-          std::string error_message("'user halt' in AwkwardForth runtime: user-defined error or stopping "
-          "condition");
-          if (stack_can_pop()) {
-            error_message = string_at(stack().back()).empty() ? error_message :
-              string_at(stack().back());
-          }
-          throw std::invalid_argument(error_message);
+          throw std::invalid_argument(
+            "'user halt' in AwkwardForth runtime: user-defined error or stopping "
+            "condition");
         }
         case util::ForthError::recursion_depth_exceeded: {
           throw std::invalid_argument(
@@ -1562,7 +1558,7 @@ namespace awkward {
         case CODE_WRITE_DUP:
         case CODE_LEN_OUTPUT:
         case CODE_REWIND:
-        case CODE_STRING_ERROR:
+        case CODE_STRING:
         case CODE_PRINT_STRING:
           return 2;
         default:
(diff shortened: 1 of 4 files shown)


src/libawkward/forth/ForthMachine.cpp
```diff
@@ -61,49 +61,50 @@ namespace awkward {
...
-  #define CODE_EQ0 56
-  #define CODE_INVERT 57
-  #define CODE_AND 58
-  #define CODE_OR 59
-  #define CODE_XOR 60
-  #define CODE_LSHIFT 61
-  #define CODE_RSHIFT 62
-  #define CODE_FALSE 63
-  #define CODE_TRUE 64
+  #define CODE_STRING_ERROR 24
```
jpivarski: Maybe just `CODE_STRING` because all that

```forth
s" Hello world!"
```

does is put an index of the static `"Hello world!"` string onto the stack. You'll be using it for communicating error messages from Forth to TypedArrayBuilder, but it could have other uses as well.

src/libawkward/forth/ForthMachine.cpp
```diff
@@ -1256,9 +1270,13 @@ namespace awkward {
...
-          throw std::invalid_argument(
-            "'user halt' in AwkwardForth runtime: user-defined error or stopping "
-            "condition");
+          std::string error_message("'user halt' in AwkwardForth runtime: user-defined error or stopping "
+          "condition");
+          if (stack_can_pop()) {
+            error_message = string_at(stack().back()).empty() ? error_message :
+              string_at(stack().back());
+          }
+          throw std::invalid_argument(error_message);
```
jpivarski: Let's not change `maybe_throw`. I think it would be better if the `util::ForthError::user_halt` exception does not require Forth programs to be written in such a way that they always set a string immediately before the `halt` keyword.

The above code assumes that `halt` is only ever used this way:

```forth
s" Error message" halt
```

and without the `s" Error message"`, it would be a C++ segfault (string not found in `strings_`) or a confusing behavior: returning a random string as an error message, if that string is defined anywhere else in the Forth code and there happened to be the right integer on the top of the stack.

Let's have the assumption that `s" Error message"` always precedes `halt` be an assumption that only TypedArrayBuilder makes and guarantees.

src/libawkward/forth/ForthMachine.cpp
```diff
@@ -3310,6 +3345,13 @@ namespace awkward {
               break;
             }
 
+            case CODE_STRING_ERROR: {
+              I string_num = bytecode_get();
+              bytecodes_pointer_where()++;
+              stack_push(string_num);
```
jpivarski, line 3352: Protect this by preceding `stack_push` with

```c++
              if (stack_cannot_push()) {
                current_error_ = util::ForthError::stack_overflow;
                return;
              }
```

jpivarski (changes requested):
I did a quick first check. Here is how the checklist now stands, I think:

   - [x] `#define CODE_STRING` (not `CODE_ERROR_STRING`, since this could be used in other ways.
   - [x] in `reserved_words_`
   - [x] handle in decompilation
   - [x] should return `2` in `bytecodes_per_instruction`
   - [x] in compilation (`parse`), although the `else if (word == "s\"")` block could possibly be merged with the `else if (word == ".\"")`, with one extra check for whether to `push_back` `CODE_STRING` or `CODE_PRINT_STRING`
   - [x] runtime implementation

Actually, it looks like you've done everything, though I have comments inline about relaxing assumptions that this is necessarily for error messages.

src/libawkward/forth/ForthMachine.cpp
ianna: done

src/libawkward/forth/ForthMachine.cpp
ianna: done

src/libawkward/forth/ForthMachine.cpp
ianna: done

ianna:
@jpivarski - thanks! I've addressed all the comments. I think, I'm done with this PR.
Thanks!

jpivarski (approved):
Great! Everything seems to be in place, and the tests demonstrate that it works.

PR #786, opened 2021-03-23
ci: cleanup and remove 3.9 restrictions

henryiii:
A bit experimental, now that numba is available this should work, perhaps?

Dropping flake8 since it's already tested in in a more controlled environment in pre-commit.

Do we need every combination of 32 bits and 64 bits on every supported Python for the tests on Windows? Couldn't a sampling be used?

### DIFF
--- a/.ci/azure-buildtest-awkward.yml
+++ b/.ci/azure-buildtest-awkward.yml
@@ -61,10 +61,6 @@ jobs:
           python.version: "3.7"
           python.architecture: "x64"
           numpy.version: "latest"
-        "py38-32bit":
-          python.version: "3.8"
-          python.architecture: "x86"
-          numpy.version: "latest"
         "py38-64bit":
           python.version: "3.8"
           python.architecture: "x64"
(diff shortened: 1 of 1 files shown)


jpivarski:
> Do we need every combination of 32 bits and 64 bits on every supported Python for the tests on Windows? Couldn't a sampling be used?

It has sometimes been just a sampling, but then we needed more tests for something, so the full matrix was put back in. It can be reduced again. The main thing that it catches are integer width mismatches in C++, which are sometimes just a warning and sometimes an error. (I haven't checked for warnings in a while.) It can also catch Numba errors, as some generated code works with `numba.intp` versus `numba.int64`. Neither of these have been in active development recently, so the matrix can be scaled back, especially as Python 3.9 is being added.

I've managed to get a 32-bit Linux Docker image working on my local machine to _diagnose_ 32-bit errors once they've been found by the Windows builds. It would be nice to integrate that someday, but not something I want to do right now.

henryiii:
I think it's a good idea to have a couple of 32 bit tests, just not for every Python version. :) Is this ready to merge, or would you like me to scale it back a bit in this PR?

henryiii:
Cleaned up one of them (32-bit 3.8), that should be fairly comprehesive but a little faster

jpivarski:
Somehow, GitHub hasn't gotten the message that Windows 32-bit Python 3.8 isn't included, but maybe it will after merging. I think it's fine to merge now. I'll do so if I don't hear back after a while.

henryiii:
It needs to be removed from the required checks. It thinks it will eventually show up.

henryiii:
I dropped it from the required checks.

jpivarski:
Great, thanks!

PR #793, opened 2021-03-28
This PR integrates JAX element wise differentiation into the main codebase

trickarcher:


tests/test_0793-jax-element-wise-ops.py
```diff
@@ -0,0 +1,32 @@
+import awkward as ak
+import numpy as np
+import jax
```
jpivarski:
```python
```

trickarcher:
Hi @jpivarski, the tests are queued right now, but this is ready for review!

src/awkward/_connect/_jax/__init__.py
```diff
@@ -0,0 +1,44 @@
...
+or
+
+    conda install jax jaxlib"""
+        )
+    else:
+        if not checked_version and distutils.version.LooseVersion(
+            jax.__version__
+        ) < distutils.version.LooseVersion("0.2.0"):
+            raise ImportError(
+                "Awkward Array can only work with jax 0.2.9 or later "
```
jpivarski: These numbers are not the same...

src/awkward/highlevel.py
```diff
@@ -364,6 +364,24 @@ def behavior(self, behavior):
                 "behavior must be None or a dict" + ak._util.exception_suffix(__file__)
             )
 
+    @classmethod
+    def set_jaxtracers(cls, layout, jaxtracers, isscalar=False):
```
jpivarski: This needs to be a hidden method, starting with an underscore. Also, the name "set_..." implies that it takes an existing thing and modifies it, but doesn't: it's an alternate constructor. Perhaps the name should be

```python
@classmethod
def _internal_for_jax(cls, layout, jaxtracers, isscalar=False)
```

to really emphasize that this makes something that regular users should not make.

src/awkward/highlevel.py
```diff
@@ -1024,18 +1045,23 @@ def __setitem__(self, where, what):
...
+                    "only fields may be assigned in-place (by field name)"
+                    + ak._util.exception_suffix(__file__)
+                )
+            array = ak.operations.structure.with_field(self.layout, what, where)
+            self._layout = array.layout
+            self._caches = ak._util.find_caches(self.layout)
+            self._numbaview = None
+        else:
+            raise ValueError(
+                "this operation cannot be performed in a JAX-compiled or JAX-differentiated function"
```
jpivarski: The error message could provide more context: "this operation cannot be performed" → "fields cannot be assigned to".

Not that there aren't other error messages that are just as vague...

src/awkward/nplike.py
```diff
@@ -126,6 +126,9 @@ def asarray(self, *args, **kwargs):
         # array[, dtype=]
         return self._module.asarray(*args, **kwargs)
 
+    def isscalar(self, *args, **kwargs):
+        return self._module.isscalar(*args, **kwargs)
```
jpivarski, line 130: Good: I verified that NumPy and CuPy both have `isscalar` and they seem to work the same way.

src/awkward/nplike.py
```diff
@@ -182,6 +185,9 @@ def searchsorted(self, *args, **kwargs):
         # haystack, needle, side="right"
         return self._module.searchsorted(*args, **kwargs)
 
+    def argsort(self, *args, **kwargs):
+        return self._module.argsort(*args, **kwargs)
```
jpivarski, line 189: Good: I verified that NumPy and CuPy both have `argsort` and they seem to work the same way.

src/awkward/nplike.py
```diff
@@ -192,6 +198,10 @@ def cumsum(self, *args, **kwargs):
         # arrays[, out=]
         return self._module.cumsum(*args, **kwargs)
 
+    def cumprod(self, *args, **kwargs):
+        # arrays[, out=]
+        return self._module.cumprod(*args, **kwargs)
```
jpivarski, line 203: Good: I verified that NumPy and CuPy both have `cumprod` and they seem to work the same way.

src/awkward/nplike.py
```diff
@@ -233,6 +243,10 @@ def broadcast_to(self, *args, **kwargs):
         # array, shape
         return self._module.broadcast_to(*args, **kwargs)
 
+    def append(self, *args, **kwargs):
+        # array, element
+        return self._module.append(*args, **kwargs)
```
jpivarski, line 248: Okay, CuPy doesn't have an `append`, so you'll need to overload the CupyLike class with an explicit implementation. (It will be easy; just create an array one bigger, copy the first into it, and set the last element.)

jpivarski (changes requested):
I've been following this along the way, so my comments aren't deep ones, just a few things to fix before I've signed off on it.

I'm also going to ask @lukasheinrich to review the _tests_ explicitly. We want to know if the tests show that it's working as a JAX user would want.

jpivarski:
Sorry, @lukasheinrich, I meant to ask you to be a reviewer, not an assignee of this PR, and just to check the tests (the last file in the diff).

src/awkward/nplike.py
trickarcher: I think CuPy does have an implementation of append [here](https://docs.cupy.dev/en/latest/reference/generated/cupy.append.html).

src/awkward/_connect/_jax/__init__.py
trickarcher: Yup, I didn't notice this. I pushed it back to the same version as our `requirements-dev` jax.

src/awkward/nplike.py
jpivarski: I don't see it, and I just upgraded from CuPy 8.5.0 to 8.6.0 (the latest, apparently):

```python
>>> import cupy
>>> cupy.__version__
'8.6.0'
>>> cupy.append
AttributeError: module 'cupy' has no attribute 'append'
```

jpivarski (approved):

lukasheinrich:
Hi @trickarcher  - apologies for the delay in the review!

I did some basic tests and a lot of them look very good and are functional. in my own testing I came across this issue:


```python
import awkward as ak
import numpy as np
import pytest

jax.config.update("jax_platform_name", "cpu")
ak.jax.register()


def func_numpyarray_1(x):
    return ak.Array(x[2:3] ** 2)


test_input = ak.Array([1.,2.,3.,4.])
v,f = jax.vjp(func_numpyarray_1,test_input)
print(v,type(v),repr(v))

(grads,) = f(ak.Array([1.0]))
print(grads)
 ```

gives

```
[0] <class 'awkward.highlevel.Array'> <Array [0] type='1 * float32'>
[0, 0, 0, 0]
```

while leaving out the wrapping `ak.Array` in the test function allows the gradients to flow.. is this something thats fixable within the scope of this PR? Note also that the value itself is zero which is unexpected

I also noticed that the value of `v` here is run-dependent, sometimes it's `0` somtimes `-0` sometimes `inf` or `-inf` .. is this accessing uninitialized memory?


for reference here the non-wrapping versino

```python
import awkward as ak
import numpy as np
import pytest

jax.config.update("jax_platform_name", "cpu")
ak.jax.register()


def func_numpyarray_1(x):
    return x[2:3] ** 2


test_input = ak.Array([1.,2.,3.,4.])
v,f = jax.vjp(func_numpyarray_1,test_input)
print(v,type(v),repr(v))

(grads,) = f(ak.Array([1.0]))
print(grads)
```

giving

```
[9] <class 'awkward.highlevel.Array'> <Array [9] type='1 * float32'>
[0, 0, 6, 0]
```

lukasheinrich:
here is another test case that seems to break (diffing through a `ak.Record`) but I'm not sure whethr that falls under the lilmitations we already identified

```
#works
def func_numpyarray_1(x):
    return x

arr = ak.Array([1.0,2.0])
print(jax.jvp(func_numpyarray_1,(arr,),(arr,)))


#does not work
def func_numpyarray_1(x):
    return x['x']

rec = ak.Record({'x': [1.0,2.0], 'y': [3.0]})
print(jax.jvp(func_numpyarray_1,(rec,),(rec,)))

#works
def func_numpyarray_1(x):
    return x['x']

arr = ak.Array([{'x': [1.0,2.0], 'y': [3.0]}])
print(jax.jvp(func_numpyarray_1,(arr,),(arr,)))
```

jpivarski:
We discussed it at our meeting and concluded that this PR is in a good state, provided that a high-level interface is added (as another PR) to wrap potential scalars (numbers, None, and `ak.Record`) into one-element arrays, run the JAX function, then unwrap them by taking the zeroth element. (These can be slices: `input[np.newaxis]` to add a dimension, `output[0]` to remove it.)

```python
ak.jax.jacobian
ak.jax.jacfwd
ak.jax.grad
ak.jax.hessian
```

@trickarcher had one more thing to add to the PR before declaring it done, though.

src/awkward/nplike.py
trickarcher: Hey, I completely forgot that we weren't putting a `CuPy` minimum version in `requirements-dev.txt` because Azure didn't have GPU's and the pip install would fail anyways. So this PR should be mergeable once the tests pass!

trickarcher:
@jpivarski Apparantly there was nothing to be done since CuPy doesn't run on CI builds. You can merge the PR once the tests pass. Thank You!

jpivarski:
@trickarcher Okay, thanks!

PR #803, opened 2021-04-07
argsort to return an index-type in the case that the input list array is empty

ianna:
addresses issue https://github.com/scikit-hep/awkward-1.0/issues/802

### DIFF
--- a/src/libawkward/array/NumpyArray.cpp
+++ b/src/libawkward/array/NumpyArray.cpp
@@ -3424,10 +3424,6 @@ namespace awkward {
                            bool ascending,
                            bool stable,
                            bool keepdims) const {
-    if (length() == 0 ) {
-      return shallow_copy();
-    }
-
     if (shape_.empty()) {
       throw std::runtime_error(
         std::string("attempting to argsort a scalar") + FILENAME(__LINE__));
--- a/.clang-format
+++ b/.clang-format
@@ -21,4 +21,3 @@ BinPackParameters: false
 AlwaysBreakTemplateDeclarations: Yes
 ReflowComments: false
 BinPackArguments: false
-BinPackParameters: false
--- a/tests/test_0803-argsort-fix-type.py
+++ b/tests/test_0803-argsort-fix-type.py
@@ -1,0 +1,27 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from __future__ import absolute_import
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import awkward as ak  # noqa: F401
+
+
+def test_argsort():
+    array = ak.Array([[1.1, 2.2], [3.3, 3.1]])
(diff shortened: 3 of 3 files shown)


src/libawkward/array/NumpyArray.cpp
```diff
@@ -3424,10 +3424,6 @@ namespace awkward {
                            bool ascending,
                            bool stable,
                            bool keepdims) const {
-    if (length() == 0 ) {
-      return shallow_copy();
-    }
-
```
ianna, line 3430: @jpivarski - if performance is a concern, this could be changed to return an `"l"` dtype NumpyArray such as:
```python
<NumpyArray format="l" shape="0" data="" at="0x000000000000"/>
```

ianna:
@henryiii - it's strange that the CI pre-commit disagrees with my local one:
```python
(base) yana@iannas-macbook-pro-2 awkward-1.0 % pre-commit run --all                  
Check for added large files..............................................Passed
Check for case conflicts.................................................Passed
Check for merge conflicts................................................Passed
Check for broken symlinks............................(no files to check)Skipped
Check Yaml...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Mixed line ending........................................................Passed
Fix requirements.txt.....................................................Passed
Trim Trailing Whitespace.................................................Passed
black....................................................................Passed
flake8...................................................................Passed
(base) yana@iannas-macbook-pro-2 awkward-1.0 % git status
On branch ianna/bugfix-issue-802
nothing to commit, working tree clean
```

src/libawkward/array/NumpyArray.cpp
jpivarski: It's not a concern. Thanks for fixing this!

jpivarski (approved):

jpivarski:
I've found out hard to ensure that local and remote pre-commits run the same set of checks. It has something to do with what you have installed. In this case, it's checking the .clang-format YAML, a test I've never seen before. Nevertheless, it's probably a real error on the file.

ianna:
> I've found out hard to ensure that local and remote pre-commits run the same set of checks. It has something to do with what you have installed. In this case, it's checking the .clang-format YAML, a test I've never seen before. Nevertheless, it's probably a real error on the file.

Indeed there was a duplicate `BinPackParameters`

src/libawkward/array/NumpyArray.cpp
ianna: Thanks for checking!

henryiii:
@ianna & @jpivarski, _most_ of the time, pre-commit pins quite well and it is very reproducible. However, that may not quite always be the case, as you've seen above; also this can change for clang-format (since the docker images are not pinned). I think this might be that `.clang-format` has started counting as a yaml file. Maybe your version of pre-commit itself is older? That's not "pinned" like the checks are.

henryiii:
Yes, here it is: https://github.com/pre-commit/identify/commit/9bb588c195843a3256a26a3503d7eebf15f6456d

You have an older pre-commit/identify than the CI has.

Issue #806, opened 2021-04-07
PartitionedArray class properties not correctly implemented 


drahnreb:
Descriptors for class `PartitionedArray` and `IrregularlyPartitionedArray` are not accessible via:

```python
>>> import awkward as ak
>>> ak.__version__
1.2.0
>>> array = ak.repartition([{"x": x, "y": x * 10} for x in range(10)], 2)
>>> array.layout
<IrregularlyPartitionedArray>
    <partition start="0" stop="2">
        <RecordArray length="2">
            <field index="0" key="x">
                <NumpyArray format="l" shape="2" data="0 1" at="0x00010f038200"/>
            </field>
            <field index="1" key="y">
                <NumpyArray format="l" shape="2" data="0 10" at="0x00010f03a200"/>
            </field>
        </RecordArray>
    </partition>
    <partition start="2" stop="4">
        <RecordArray length="2">
            <field index="0" key="x">
                <NumpyArray format="l" shape="2" data="2 3" at="0x00010f038200"/>
            </field>
            <field index="1" key="y">
                <NumpyArray format="l" shape="2" data="20 30" at="0x00010f03a200"/>
            </field>
        </RecordArray>
    </partition>
</IrregularlyPartitionedArray>

>>> ak.partitions(array)
[2,2]
>>> array.partitions

AttributeError: no field named 'partitions'
```
According to the docs, this should be a valid repartition? 

The [tests](https://github.com/scikit-hep/awkward-1.0/blob/main/tests/test_0056-partitioned-array.py) do not cover `ak.repartition` with partitioned array methods, but following one test leads to:
```python
>>> one = ak.from_iter([[1.1, 2.2, 3.3], [], [4.4, 5.5]], highlevel=False)
>>> two = ak.from_iter([[6.6], [], [], [], [7.7, 8.8, 9.9]], highlevel=False)
>>> array = ak.partition.IrregularlyPartitionedArray([one, two])
>>> array.layout

AttributeError: no field named 'layout'
```

According to the docs, I would expect this to work as
> [...] it should behave identically to a non-partitioned array [...]

jpivarski:
Both of these are problems with mixing high-level and low-level arrays.

[...]

Issue #490, opened 2020-10-18
removing none from awkward array 1


ramankhurana:
Hello 

I have an awkward array (1) which I obtained post-processing.

An array look like (with some of the elements as None):

>>> a=ak.Array([96., 99., 67., 13.,  3.,  None, 1.,  1.,None])

I want to remove the None elements from this array. I tried following but it does not give expected result: 
>>> b=a[a!=None]
>>> b
<Array [96, 99, 67, 13, ... None, 1, 1, None] type='9 * ?float64'>

 I could remove them using loop and worked, but I want to avoid it to save some computing time. Or writing a function and compiling using Numba is only option?

Thanks.

ramankhurana:
I just realised that is_none exist and work like charm, 

a[~ak.is_none(a)]

Superharz:
> 
> 
> I just realised that is_none exist and work like charm,
> 
> a[~ak.is_none(a)]

I agree with your solution. 
However, it doesn't seem to work for nested arrays, like:
```
a = [[1, None], [3]   , [] , [4, None, 6]]
a = ak.from_iter(a)
ak.is_none(a)
#<Array [False, False, False, False] type='4 * bool'>
``` 
It would be nice if an ```axis``` parameter could be added to ```ak.is_none()```

jpivarski:
I agree. That's #193.

ramankhurana:
I agree, i end up in a situation where I have the nested arrays. I would wait for the feature to be in the awkward array in future.

Superharz:
I haven't been aware of #193. Thanks for bringing it up.

jpivarski:
Also, someone contacted me on Slack about this—the natural-looking syntax

```python
a[a != None]
```

won't work because `np.equal` and `np.not_equal` have not been overridden to interpret missing value `==` missing value as True and missing value `==` anything else as False. (More likely, they'd be overridden to call `ak.is_none`.)

(If "None" were "NaN," then things would be even more complicated, semantically, because `NaN == NaN` is always _supposed to be_ False. By spelling it "None," we've committed to a user expectation that `None == None` is True, which is more palatable.)

[...]

Issue #487, opened 2020-10-13
ar != None unexpected behaviour


andrzejnovak:
Using `ar != None` results in an unexpected mask

Consider:
```
a = ak.Array([None, 2, None, 2, None])
b = np.array([None, 2, None, 2, None])
a, b
>>> (<Array [None, 2, None, 2, None] type='5 * ?int64'>,
 array([None, 2, None, 2, None], dtype=object))
```

```
a != None
>>> <Array [None, True, None, True, None] type='5 * ?bool'>

b != None
>>> array([False,  True, False,  True, False])
```

Expected behaviour:
```
a != None
>>><Array [False, True, False, True, False] type='5 * ?bool'>
```
such that 
```
a[a != None]
>>> <Array [2, 2] type='2 * ?int64'>,
```

jpivarski:
There's a whole discussion of this issue, starting at https://github.com/scikit-hep/awkward-1.0/issues/490#issuecomment-712250246

When I wrote it there, I thought I was narrowly saving it from oblivion because I only remembered hearing it on Slack. I could have just pointed to this issue. Anyway, the cross-reference will be useful in solving it.

jpivarski:
(I'm going through old issues, deciding what to do with them.)

I think that the `array == None` and `array != None` behavior _should not_ be changed. In the examples you presented, @andrzejnovak, the results look non-intuitive, but they're following a rule that applies to all mathematical functions, including `==` and `!=`, and I think it would be dangerous to have exceptions to that behavior. If we made

```python
ak.Array([1, 2, None, 3, None, 4]) == None
```

return

```python
ak.Array([False, False, True, False, True, False])
```

then someone else's use-case might break because they were assuming that `==` and `!=` would act like all other mathematical functions. Namely,

```python
ak.Array([1, 2, None, 3, None, 4]) + 10
```

returns

```python
ak.Array([11, 12, None, 13, None, 14])
```

That is, the scalar `10` broadcasts and the `None` values pass through any mathematical operation. When applied to `==` (same for `!=`), the expected result of

```python
ak.Array([1, 2, None, 3, None, 4]) == None
```

would be

```python
ak.Array([False, False, None, False, None, False])
```

because each `integer == None` is `False` and each missing value passes through.

@agoose77, do you concur? If you think there is something that we should do that would benefit all use-cases, reopen the issue. Thanks!

agoose77:
[...]

PR #808, opened 2021-04-08
Issue #805: fix an empty case when broadcasting UnionArrays

### LINKED ISSUE #805: ak.cartesian fails on empty lists of union types
Awkward version = 1.2.0

```python3
import awkward as ak

muon = ak.Array([[{"pt": 1.}],[]], with_name="muon")
electron = ak.Array([[],[{"pt": 1.}]], with_name="electron")

muon = muon[muon.pt > 5]
electron = electron[electron.pt > 5]

leptons = ak.concatenate([muon,electron], axis=1)
candidate = ak.firsts(leptons)

ak.cartesian([candidate, candidate])
```

fails with:
```
  File "make_example.py", line 12, in <module>
    prod = ak.cartesian([candidate,candidate])
  File "/opt/conda/lib/python3.8/site-packages/awkward/operations/structure.py", line 3089, in cartesian
    out = ak._util.broadcast_and_apply(
  File "/opt/conda/lib/python3.8/site-packages/awkward/_util.py", line 1037, in broadcast_and_apply
    out = apply(broadcast_pack(inputs, isscalar), 0, user)
  File "/opt/conda/lib/python3.8/site-packages/awkward/_util.py", line 799, in apply
    outcontent = apply(nextinputs, depth + 1, user)
  File "/opt/conda/lib/python3.8/site-packages/awkward/_util.py", line 752, in apply
    outcontent = apply(nextinputs, depth, user)
  File "/opt/conda/lib/python3.8/site-packages/awkward/_util.py", line 717, in apply
    for i in range(numoutputs)
UnboundLocalError: local variable 'numoutputs' referenced before assignment
```

and indeed numoutputs is not defined in the case there are no unique non-null entries.

ianna:
proposed error message for `ak.cartesian` reported by @lgray - issue https://github.com/scikit-hep/awkward-1.0/issues/805

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -708,9 +708,34 @@ def broadcast_and_apply(  # noqa: C901
                 numoutputs = len(outcontents[-1])
 
             if numoutputs is None:
-                return tuple(
-                    x[0:0] if isinstance(x, ak.layout.Content) else x for x in inputs
-                )
+
+                def copy_listtypes(contents):
+                    if all(
+                        isinstance(x, ak.layout.RegularArray) for x in contents
+                    ) and all(x.size == contents[0].size for x in contents):
+                        return ak.layout.RegularArray(
+                            copy_listtypes([x.content for x in contents]),
+                            contents[0].size,
+                            0,
+                        )
+                    elif all(isinstance(x, listtypes) for x in contents):
+                        return ak.layout.ListOffsetArray64(
+                            ak.layout.Index64(nplike.empty(0, dtype=np.int64)),
+                            copy_listtypes([x.content for x in contents]),
+                        )
+                    else:
+                        return ak.layout.EmptyArray()
+
+                nextinputs = []
+                for x in inputs:
+                    if isinstance(x, uniontypes):
+                        nextinputs.append(copy_listtypes(x.contents))
+                    elif isinstance(x, ak.layout.Content):
+                        nextinputs.append(x[0:0])
+                    else:
+                        nextinputs.append(x)
+                return apply(nextinputs, depth, user)
+
             else:
                 tags = ak.layout.Index8(tags)
                 index = ak.layout.Index64(index)
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -685,13 +685,13 @@ def broadcast_and_apply(  # noqa: C901
 
             tags = nplike.empty(length, dtype=np.int8)
             index = nplike.empty(length, dtype=np.int64)
+            numoutputs = None
             outcontents = []
             for tag, combo in enumerate(nplike.unique(combos)):
                 mask = combos == combo
                 tags[mask] = tag
                 index[mask] = nplike.arange(nplike.count_nonzero(mask))
                 nextinputs = []
-                numoutputs = None
                 i = 0
                 for x in inputs:
                     if isinstance(x, uniontypes):
@@ -707,15 +707,19 @@ def broadcast_and_apply(  # noqa: C901
                     assert numoutputs == len(outcontents[-1])
                 numoutputs = len(outcontents[-1])
 
-            tags = ak.layout.Index8(tags)
-            index = ak.layout.Index64(index)
-
-            return tuple(
-                ak.layout.UnionArray8_64(
-                    tags, index, [x[i] for x in outcontents]
-                ).simplify()
-                for i in range(numoutputs)
-            )
+            if numoutputs is None:
+                return tuple(
+                    x[0:0] if isinstance(x, ak.layout.Content) else x for x in inputs
+                )
+            else:
+                tags = ak.layout.Index8(tags)
+                index = ak.layout.Index64(index)
+                return tuple(
+                    ak.layout.UnionArray8_64(
+                        tags, index, [x[i] for x in outcontents]
+                    ).simplify()
+                    for i in range(numoutputs)
+                )
 
         elif any(isinstance(x, optiontypes) for x in inputs):
             mask = None
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2987,7 +2987,7 @@ def cartesian(
     else:
         # 'posaxis' > 0: check if given 'axis' does not exceed the array depth
         for x in new_arrays:
-            if x.minmax_depth[0] == x.minmax_depth[1] == posaxis:
+            if isinstance(x, ak.layout.EmptyArray):
                 raise ValueError(
                     "cartesian-product of empty arrays?\n"
                     "Please, set 'axis=0' for a cartesian-product of the arrays with "
@@ -2996,6 +2996,16 @@ def cartesian(
                     + "'"
                     + ak._util.exception_suffix(__file__)
                 )
+            elif isinstance(x, ak.layout.IndexedOptionArray64):
+                if x.minmax_depth[0] == x.minmax_depth[1] == posaxis:
+                    raise ValueError(
+                        "cartesian-product of empty arrays?\n"
+                        "Please, set 'axis=0' for a cartesian-product of the arrays with "
+                        "min and max depths that are equal to a given 'axis="
+                        + str(posaxis)
+                        + "'"
+                        + ak._util.exception_suffix(__file__)
+                    )
 
         def newaxis(layout, i):
             if i == 0:
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2985,27 +2985,6 @@ def cartesian(
         result = ak.partition.IrregularlyPartitionedArray(output)
 
     else:
-        # 'posaxis' > 0: check if given 'axis' does not exceed the array depth
-        for x in new_arrays:
-            if isinstance(x, ak.layout.EmptyArray):
-                raise ValueError(
-                    "cartesian-product of empty arrays?\n"
-                    "Please, set 'axis=0' for a cartesian-product of the arrays with "
-                    "min and max depths that are equal to a given 'axis="
-                    + str(posaxis)
-                    + "'"
-                    + ak._util.exception_suffix(__file__)
-                )
-            elif isinstance(x, ak.layout.IndexedOptionArray64):
-                if x.minmax_depth[0] == x.minmax_depth[1] == posaxis:
-                    raise ValueError(
-                        "cartesian-product of empty arrays?\n"
-                        "Please, set 'axis=0' for a cartesian-product of the arrays with "
-                        "min and max depths that are equal to a given 'axis="
-                        + str(posaxis)
-                        + "'"
-                        + ak._util.exception_suffix(__file__)
-                    )
 
         def newaxis(layout, i):
             if i == 0:
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2985,6 +2985,17 @@ def cartesian(
         result = ak.partition.IrregularlyPartitionedArray(output)
 
     else:
+        # 'posaxis' > 0: check if given 'axis' does not exceed the array depth
+        for x in new_arrays:
+            if x.minmax_depth[0] == x.minmax_depth[1] == posaxis:
+                raise ValueError(
+                    "cartesian-product of empty arrays?\n"
+                    "Please, set 'axis=0' for a cartesian-product of the arrays with "
+                    "min and max depths that are equal to a given 'axis="
+                    + str(posaxis)
+                    + "'"
+                    + ak._util.exception_suffix(__file__)
+                )
 
         def newaxis(layout, i):
             if i == 0:
--- a/tests/test_0806-empty-lists-cartesian-fix.py
+++ b/tests/test_0806-empty-lists-cartesian-fix.py
@@ -1,0 +1,39 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from __future__ import absolute_import
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import awkward as ak  # noqa: F401
+
+
+def test_empty_arrays_cartesian():
+    one = ak.Array([])
+    two = one = ak.Array([])
+
+    with pytest.raises(ValueError) as err:
+        ak.to_list(ak.cartesian([one, two]))
+
+    assert str(err.value).startswith("cartesian-product of empty arrays?")
+
+    ak.to_list(ak.concatenate([one, two], axis=0))
+
+
+def test_cartesian():
+    muon = ak.Array([[{"pt": 1.0}], []], with_name="muon")
+    electron = ak.Array([[], [{"pt": 1.0}]], with_name="electron")
+
+    muon = muon[muon.pt > 5]
+    electron = electron[electron.pt > 5]
+
+    leptons = ak.concatenate([muon, electron], axis=1)
+    candidate = ak.firsts(leptons)
+    assert ak.to_list(ak.Array(candidate)) == [None, None]
+
+    result = ak.cartesian([candidate, candidate], axis=0)
+    assert ak.to_list(result) == [
+        (None, None),
+        (None, None),
+        (None, None),
+        (None, None),
+    ]
--- a/tests/test_0806-empty-lists-cartesian-fix.py
+++ b/tests/test_0806-empty-lists-cartesian-fix.py
@@ -14,7 +14,7 @@ def test_empty_arrays_cartesian():
     with pytest.raises(ValueError) as err:
         ak.to_list(ak.cartesian([one, two]))
 
-    assert str(err.value).startswith("cartesian-product of empty arrays?")
+    assert isinstance(err.value, ValueError)
 
     ak.to_list(ak.concatenate([one, two], axis=0))
 
--- a/tests/test_0806-empty-lists-cartesian-fix.py
+++ b/tests/test_0806-empty-lists-cartesian-fix.py
@@ -37,3 +37,10 @@ def test_cartesian():
         (None, None),
         (None, None),
     ]
+
+    result = ak.cartesian([candidate, ak.Array([[1, 2, 3], []])], axis=1)
+    assert ak.to_list(result) == [None, None]
+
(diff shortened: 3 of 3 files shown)


lgray:
Let me check this with the full example of problematic data.

jpivarski:
I hadn't gotten this far in my email before writing https://github.com/scikit-hep/awkward-1.0/issues/805#issuecomment-815832985, but this solution seems too focused on the specific problem. It's checking for EmptyArray and IndexedOptionArray node types, when the programming error happened because `posaxis >= purelist_depth` shouldn't be allowed*, and therefore we (probably I) hadn't been considering what would happen in some specific case of that. Simple cases get stopped by the fact that they can't broadcast, and this more complex case missed a code block altogether (and hence, a variable didn't get defined).

If the wrong dimensionality is excluded up-front (before [line 2897](https://github.com/scikit-hep/awkward-1.0/blob/1492c4906df9f5bbf7b4f244de543188563015a4/src/awkward/operations/structure.py#L2897)), then the subsequent code can continue to be naive (and we get a clearer and more uniform error message for all such cases). I'd rather have a general fix that expresses a rule about what `axis` is allowed to be, given an array's dimension, than a specific check for layout node types that fixes this one case. The latter would be harder to understand and maintain in the future (and `ak.cartesian` is complicated enough!).

*I haven't thought deeply about where the cut-off should be. I _think_ `0 <= posaxis < purelist_depth` is the right boundary, since a one-dimensional array has `purelist_depth == 1` and only `axis=0` is allowed for that case. @ianna, we can keep the tests you wrote for this PR, but just change the code to express this general rule about the dimension of the input arrays and `axis`.

jpivarski:
@ianna, disregard the above: see https://github.com/scikit-hep/awkward-1.0/issues/805#issuecomment-815855292.

jpivarski:
I'm not sure whether I should post here or on issue #805, but since everyone involved is subscribed to both, I'll do it here.

First of all, disregard my request to exclude everything but `0 <= posaxis < purelist_depth` for `purelist_depth` of all inputs to `ak.cartesian` because that would exclude some good cases, like the one @lgray showed with a low-dimension array that gets broadcasted up to match the dimensionality of a high-dimension array. It _might_ be appropriate to require `0 <= posaxis < purelist_depth` for `purelist_depth` of _any_ (rather than _all_) inputs to `ak.cartesian`, but that would require some thought and doesn't address the real issue here.

Sorry for the confusion!

Anyway, the real problem is in broadcasting itself, as `ak.broadcast_arrays` raises the same error message. It could be fixed like this:

```diff
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -683,6 +683,7 @@ def broadcast_and_apply(  # noqa: C901
                 [(str(i), combos.dtype) for i in range(len(tagslist))]
             ).reshape(length)
 
+            numoutputs = None
             tags = nplike.empty(length, dtype=np.int8)
             index = nplike.empty(length, dtype=np.int64)
             outcontents = []
@@ -691,7 +692,6 @@ def broadcast_and_apply(  # noqa: C901
                 tags[mask] = tag
                 index[mask] = nplike.arange(nplike.count_nonzero(mask))
                 nextinputs = []
-                numoutputs = None
                 i = 0
                 for x in inputs:
                     if isinstance(x, uniontypes):
@@ -707,15 +707,20 @@ def broadcast_and_apply(  # noqa: C901
                     assert numoutputs == len(outcontents[-1])
                 numoutputs = len(outcontents[-1])
 
-            tags = ak.layout.Index8(tags)
-            index = ak.layout.Index64(index)
+            if numoutputs is None:
+                return tuple(
+                    x[0:0] if isinstance(x, ak.layout.Content) else x for x in inputs
+                )
 
-            return tuple(
-                ak.layout.UnionArray8_64(
-                    tags, index, [x[i] for x in outcontents]
-                ).simplify()
-                for i in range(numoutputs)
-            )
+            else:
+                tags = ak.layout.Index8(tags)
+                index = ak.layout.Index64(index)
+                return tuple(
+                    ak.layout.UnionArray8_64(
+                        tags, index, [x[i] for x in outcontents]
+                    ).simplify()
+                    for i in range(numoutputs)
+                )
 
         elif any(isinstance(x, optiontypes) for x in inputs):
             mask = None
```

Moving `numoutputs` out of the `for` loop makes its assertion in the loop stronger:

https://github.com/scikit-hep/awkward-1.0/blob/1492c4906df9f5bbf7b4f244de543188563015a4/src/awkward/_util.py#L706-L708

but previously, it had no strength at all—it was set to None immediately before checking for None, so that assertion happened in unreachable code, clearly not what was intended. So I think it's safe to say that I had intended `numoutputs = None` to be outside the loop (back when I was thinking deeply about broadcasting).

If `numoutputs` hasn't been set, then `outcontents` is an empty list. A UnionArray can't have zero contents, so the output has to be something other than a UnionArray. (Also, since the UnionArray that got created in the original code was immediately `.simplified`, it wasn't necessary for the output to have union-type.)

Also if `numoutputs` hasn't been set, then we know that `len(combos) == 0` and therefore `len(tagslist) == 0` and `length == 0`. So the output arrays are definitely going to be empty for this case (and wrapped as arrays of all None in a recursive step above this one). Instead of creating EmptyArrays, I sliced all input Contents to zero length so that they would retain their types:

```python
>>> ak.broadcast_arrays(candidate, ak.Array([[1, 2, 3], []]))
[
    <Array [None, None] type='2 * ?union[muon["pt": float64], electron["pt": float64]]'>,
    <Array [None, None] type='2 * option[var * int64]'>
]
```

After quite a few iterations on this comment, that seems to be the right solution.

@ianna, does this pass your tests (with _no alterations_ to `ak.cartesian`)?

jpivarski:
Actually, @ianna, since I have the files open, I can test it myself. It seems that everything is passing (after changing the test to not require a specific message). The next commit is the solution as described above.

@lgray, does it fix your problem?

lgray:
I'll check in a moment on the full data just to make sure. Thanks!

lgray:
@jpivarski 

Running the mre in #805 now yields:
```
[None, [({pt: 5}, {pt: 4})]]
  File "mre.py", line 21, in <module>
    mre()
  File "mre.py", line 16, in mre
    x = ak.cartesian([candidate, jet])
  File "/Users/lagray/miniconda3/envs/coffea-jetmet/lib/python3.7/site-packages/awkward/operations/structure.py", line 3092, in cartesian
    assert isinstance(out, tuple) and len(out) == 1
AssertionError
```

jpivarski:
Okay, that's because I checked `ak.broadcast_apply` and didn't check `ak.cartesian` directly (and @ianna's tests somehow missed this, too). The assertion is because `ak.broadcast_apply` expects as many arrays out as went in, `ak.cartesian` expects one array out, regardless of how many arrays went in, and my handling of the empty case has as many arrays coming out as in.

To get it right, I'll need to recurse down into the empty arrays, though that's just a formal thing to get the right number of empty arrays out as the operation requires (for _all_ operations that use `ak._util.broadcast_and_apply`, which are many).

However, that raises the question of what _should_ the output type be? I can do this, for instance:

```diff
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -708,9 +708,16 @@ def broadcast_and_apply(  # noqa: C901
                 numoutputs = len(outcontents[-1])
 
             if numoutputs is None:
-                return tuple(
-                    x[0:0] if isinstance(x, ak.layout.Content) else x for x in inputs
-                )
+                nextinputs = []
+                for x in inputs:
+                    if isinstance(x, uniontypes):
+                        nextinputs.append(x[0:0].project(0))   # first type in the union
+                    elif isinstance(x, ak.layout.Content):
+                        nextinputs.append(x[0:0])
+                    else:
+                        nextinputs.append(x)
+                return apply(nextinputs, depth, user)
+
             else:
                 tags = ak.layout.Index8(tags)
                 index = ak.layout.Index64(index)
```

in which case,

```python
>>> ak.cartesian([candidate, ak.Array([[1, 2, 3], []])])
<Array [None, None] type='2 * option[var * (muon["pt": float64], int64)]'>
```

It "unfairly" picks the first type in the union to set the type for the empty output by doing a `.project(0)` (we know that every union has at least one content).

If the `candidate` array contained non-missing electrons and muons, the output type would have electrons and muons; if the `candidate` array contained non-missing electrons (but the muons were all missing), the output type would have electrons only; etc. Since this `candidate` array has everything missing, maybe it ought to not have any of those types. Instead of projecting the UnionArray to its first content, we could replace it with an EmptyArray, but that would change the number of dimensions and can interfere with broadcasting further down the recursive chain (not in this case). Any dimensions that are shared by all members of the union should be copied:

```diff
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -708,9 +708,34 @@ def broadcast_and_apply(  # noqa: C901
                 numoutputs = len(outcontents[-1])
 
             if numoutputs is None:
-                return tuple(
-                    x[0:0] if isinstance(x, ak.layout.Content) else x for x in inputs
-                )
+                def copy_listtypes(contents):
+                    if (
+                        all(isinstance(x, ak.layout.RegularArray) for x in contents)
+                        and all(x.size == contents[0].size for x in contents)
+                    ):
+                        return ak.layout.RegularArray(
+                            copy_listtypes([x.content for x in contents]),
+                            contents[0].size,
+                            0,
+                        )
+                    elif all(isinstance(x, listtypes) for x in contents):
+                        return ak.layout.ListOffsetArray64(
+                            ak.layout.Index64(nplike.empty(0, dtype=np.int64)),
+                            copy_listtypes([x.content for x in contents]),
+                        )
+                    else:
+                        return ak.layout.EmptyArray()
+
+                nextinputs = []
+                for x in inputs:
+                    if isinstance(x, uniontypes):
+                        nextinputs.append(copy_listtypes(x.contents))
+                    elif isinstance(x, ak.layout.Content):
+                        nextinputs.append(x[0:0])
+                    else:
+                        nextinputs.append(x)
+                return apply(nextinputs, depth, user)
+
             else:
                 tags = ak.layout.Index8(tags)
                 index = ak.layout.Index64(index)
```

And now

```python
>>> ak.cartesian([candidate, ak.Array([[1, 2, 3], []])])
<Array [None, None] type='2 * option[var * (bool, int64)]'>
```

The "bool" comes from EmptyArray and might be another, unrelated issue.

```python
>>> ak.cartesian([ak.Array([None, None]), ak.Array([[1, 2, 3], []])])
<Array [None, None] type='2 * option[var * (bool, int64)]'>
```

For now, I'm going to focus on this issue, since it seems they can be solved indepenently.

I'm going to add some tests and commit this.

jpivarski:
About that "bool":

```python
>>> ak.broadcast_arrays(candidate, ak.Array([[1, 2, 3], []]))
[
    <Array [None, None] type='2 * option[var * bool]'>,
    <Array [None, None] type='2 * option[var * int64]'>,
]
>>> ak.broadcast_arrays(ak.Array([None, None]), ak.Array([[1, 2, 3], []]))
[
    <Array [None, None] type='2 * option[var * bool]'>,
    <Array [None, None] type='2 * option[var * int64]'>,
]
```

It doesn't have anything to do with `ak.cartesian`; the by-product of broadcasting an EmptyArray makes it boolean for some reason. Probably the right type is "unknown," but I'm not sure and I'm not going to try to solve that now.

lgray:
To me the output array should have the most covering type. Even in the case of selecting all electrons or all muons it should return union[electron, muon] as the underlying type, since as a user you have defined the input array as such and would not expect it to lose that type.

jpivarski:
> To me the output array should have to most covering type. Even in the case of selecting all electrons or all muons it should return union[electron, muon] as the underlying type, since as a user you have defined the input array as such and would not expect it to lose that type.

Most operations that can simplify UnionArrays do simplify them. For example, integers and booleans are considered different types, but since they both have a `~` operation (bitwise invert), it can be applied across a union of integers and booleans:

```python
>>> mixed = ak.Array([1, 2, False, True, True, 3])
>>> mixed
<Array [1, 2, False, True, True, 3] type='6 * union[int64, bool]'>
>>> ~mixed
<Array [-2, -3, True, False, False, -4] type='6 * union[int64, bool]'>
```

But when you focus down on a subset that happens to consist of only one type, an operation on it simplifies the UnionArray.

```python
>>> mixed[[0, 1, -1]]
<Array [1, 2, 3] type='3 * union[int64, bool]'>
>>> ~mixed[[0, 1, -1]]
<Array [-2, -3, -4] type='3 * int64'>           # not union[int64, bool]
>>> ~mixed[[2, 3, 4]]
<Array [True, False, False] type='3 * bool'>    # not union[int64, bool]
```

Arguably, the first step, in which we sliced the union to get only integers, should have already been simplified and maybe that's a bug. Exactly the same thing, in which a user sliced to get only non-missing data, has repeatedly been reported as a bug (#487, #490).

Here's another example where we'd really want the unions to be simplified but they aren't (and that, I think, is incorrect):

```python
>>> muon = ak.Array([[{"pt": 1.0}], []], with_name="muon")
>>> electron = ak.Array([[], [{"pt": 1.0}]], with_name="electron")
>>> leptons = ak.concatenate([muon, electron], axis=1)
>>> leptons
<Array [[{pt: 1}], [{pt: 1}]] type='2 * var * union[muon["pt": float64], electro...'>
>>> leptons.pt
<Array [[1], [1]] type='2 * var * union[float64, float64]'>
```

That ought to be `2 * var * float64`.

In a statically typed environment, the full type information would be carried through because different values could instantiate any of the possible types. Here, though, each typed array is also a set of values; we know how to constrict types. Constricting types does not reduce functionality—the restricted type can be used anywhere the general type could have been used—but the opposite is not necessarily true.

In the case of issue #806 specifically (and maybe I should have led with this), it wouldn't be possible to avoid constricting the types. On the next recursive level down, the union-ness needs to be removed or else this would be an infinite recursion. The whole issue comes up because

https://github.com/scikit-hep/awkward-1.0/blob/1492c4906df9f5bbf7b4f244de543188563015a4/src/awkward/_util.py#L681-L689

recurses down for all unique combinations of types that exist in the data: this is what would remove electron types if electrons never appeared in the data. It was an issue in your example because _no_ combinations appeared in the data, and we had to deal with that. Following the existing logic, which minimizes type explosion when broadcasting UnionArrays, the case of no combinations should be some EmptyArray thing.

However, I guess one could argue that we _shouldn't_ avoid type explosion: if you want to take a Cartesian product of arrays with types `union(A, B)` and `union(C, D, E)` and get type `union((A, C), (A, D), (A, E), (B, C), (B, D), (B, E))`, regardless of whether the output arrays contain any such combinations, then we would write the loop differently and it would never get a no-combination case. But this goes against the grain of the "simplify unions wherever possible" philosophy that is applied throughout the codebase (and not applied consistently enough, as the above examples show).

lgray:
So the answer I would give is somewhere in between, and perhaps not easily implementable.

For primitive types I think the simplifying is fine, largely you don't care where something that you've stripped down to being a float64, int, bool matters that much. You have factored out the bookkeeping or there was none to begin with.

However for something that is given a `__record__` name specifically, keeping the provenance of unions begins to make some sense because you'd like to be able to look at some object in a vacuum and tell what it is.
As a concrete example, there is a significant physics difference between some `leptons`, which is a union of electrons, muons, and taus, but consists only selected electrons somehow, variable being:
`?union[muon electron tau]`, and `?electron`

On a first reading of those types you'd think leptons was just a list of electrons in the latter case... And from the programming perspective it absolutely is, but from physics/analysis perspective it is very different. 

So really: type allowing explosion for user defined types and simplification for primitive types.

Perhaps this is just splitting hairs and we should tell users to be careful about bookkeeping?

lgray:
Anyway, with this fix both the mre and the full example execute without failure.

jpivarski:
This does it—I'm going to merge now.

Thanks @ianna for starting this! It got pretty deep into the broadcasting logic, though.

jpivarski:
> As a concrete example, there is a significant physics difference between some `leptons`, which is a union of electrons, muons, and taus, but consists only selected electrons somehow, variable being:
> `?union[muon electron tau]`, and `?electron`
> 
> On a first reading of those types you'd think leptons was just a list of electrons in the latter case... And from the programming perspective it absolutely is, but from physics/analysis perspective it is very different.

The granularity of a type system is a choice, designed with a particular purpose in mind. When I said that Femtocode and OAMap "started with a schema and fit values into that afterward" and Awkward Array "starts with the values and infers a type from that," I meant I was switching to this choice of putting values first, types second. The type that describes an array of values can be any type that correctly generalizes every value in the array. There's no maximum granularity other than "Any," but there's a minimum granularity, which is the type that correctly matches all the values in the array and nothing more. This is a "shrink-wrapped" type: if you take an array of "muons OR electrons" and remove all the electrons, the minimal type for the result is just "muons."

Awkward Arrays don't have a history of what has happened to them, so if you do some operation on "leptons" that removes all the electrons (and taus, I guess), then what you have really is "electrons." In a statically typed environment, in which you don't know what values you'll encounter, you _can't_ shrink-wrap it because you don't know if you'll encounter only electrons or not.

I guess what I'm saying is that Awkward Array types are dynamic types—you can't use them as type-safety for a program whose values are to be determined in the future. Maybe we should define some MyPy types for Awkward Arrays that would do that, but that's not what `array.type` is.

Also, I guess what I'm saying is that this was a design decision made in 2018 that is sufficiently fundamental that we won't be able to change it. As I've shown, there are cases that fall short of this "ideal" that I would consider bugs to fix, though from a type-safety point of view it's the opposite. Since every array does have a natural minimum type, it should be shrink-wrapped with `uniontype_simplify` to avoid UnionType nodes with empty content arrays.

Finally, this was the design decision that ended up being useful: the biggest problem with Femtocode and OAMap was improving the user experience—users had to carry around explicit Schemas and try (and fail) to fit data into them—it didn't encourage tinkering the way I wanted it to. As with all dynamically typed things, now we can tinker, but the problem is scaling it up and discovering corner-cases...

(Note the "without runtime errors" in the [description of Femtocode](https://github.com/diana-hep/femtocode#query-language-details). That would be possible with early, highly granular typing, but it comes at a cost of getting something half-working quickly. It's the same trade-off as in Rust: if it compiles, it will run forever. If it compiles... I've been in favor of very static type-checking and very dynamic at different times—I've never found a way to reconcile the trade-offs.)

Issue #811, opened 2021-04-08
Empty union types don't pass through ak.argsort correctly


lgray:
Using the following file as input, because constructing a trivial example has proven painful:
https://drive.google.com/file/d/1a1Zla1htgjgF_b74Y0TKgcA3PABSpTFf/view?usp=sharing

You will need a recent version of coffea installed to run this (just pip install coffea).
I am using awkward==1.3.0rc1.

The following script:
```python3
from coffea import nanoevents
import awkward as ak

nanoevents.NanoAODSchema.mixins["FatJetLS"] = "PtEtaPhiMLorentzVector"

x = nanoevents.NanoEventsFactory.from_root(
    './nano106Xv8_on_mini106X_2017_mc_NANO_py_NANO_7.root',
    entry_start=100560, 
    entry_stop=102266
)

events = x.events()

muons = events.Muon
electrons = events.Electron
jets = events.Jet
fatjets = events.FatJet
subjets = events.SubJet
fatjetsLS = events.FatJetLS
met = events.MET

goodmuon = (
    (muons.mediumId)
    & (muons.miniPFRelIso_all <= 0.2)
    & (muons.pt >= 27)
    & (abs(muons.eta) <= 2.4)
    & (abs(muons.dz) < 0.1)
    & (abs(muons.dxy) < 0.05)
    & (muons.sip3d < 4)
)
good_muons = muons[goodmuon]

# electrons
goodelectron = (
    (electrons.mvaFall17V2noIso_WP90)
    & (electrons.pt >= 30)
    & (abs(electrons.eta) <= 1.479)
    & (abs(electrons.dz) < 0.1)
    & (abs(electrons.dxy) < 0.05)
    & (electrons.sip3d < 4)
)
good_electrons = electrons[goodelectron]

leptons = ak.concatenate([muons, electrons], axis=1)
good_leptons = ak.concatenate([good_muons, good_electrons], axis=1)

idx1 = ak.argsort(leptons.pt, axis=1)
idx2 = ak.argsort(good_leptons.pt, axis=1)

leptons[idx1]

good_leptons[idx2]
```

fails with
```
Traceback (most recent call last):
  File "test_unions.py", line 52, in <module>
    good_leptons[idx2]
  File "/Users/lagray/miniconda3/envs/coffea-jetmet/lib/python3.7/site-packages/awkward/highlevel.py", line 974, in __getitem__
    return ak._util.wrap(self.layout[where], self._behavior)
[...]

PR #812, opened 2021-04-09
argsort bugfix for empty arrays

ianna:
address issue #811

### DIFF
--- a/src/libawkward/array/IndexedArray.cpp
+++ b/src/libawkward/array/IndexedArray.cpp
@@ -2504,7 +2504,9 @@ namespace awkward {
                                             bool stable,
                                             bool keepdims) const {
     if (length() == 0 ) {
-      return shallow_copy();
+      Index64 out(1);
+      out.setitem_at_nowrap(0, length());
+      return NumpyArray(out).getitem_nothing();
     }
 
     int64_t index_length = index_.length();
--- a/src/libawkward/array/IndexedArray.cpp
+++ b/src/libawkward/array/IndexedArray.cpp
@@ -2504,9 +2504,7 @@ namespace awkward {
                                             bool stable,
                                             bool keepdims) const {
     if (length() == 0 ) {
-      Index64 out(1);
-      out.setitem_at_nowrap(0, length());
-      return NumpyArray(out).getitem_nothing();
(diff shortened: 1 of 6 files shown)


lgray:
Cool - I'll try this out momentarily.

jpivarski (approved):
Yes, if the length is zero, `ak.argsort` should not be returning a copy of the input array, as this won't have integer type. The method you have of creating an empty NumpyArray should work, though this should too:

```python
return NumpyArray(Index64(0))
```

which would allocate nothing, rather than one element that isn't viewed.

But now it has me wondering if Indexes and NumpyArrays with no allocation (their array is a `nullptr`, pointing at the `0` address in memory) are correctly converted to NumPy arrays. It would be weird if there were any issues with that, as we've been using that technique for so long in so many places and never encountered any problems with it. I ought to do a test.

lgray:
Hmm hold on, still getting the same error despite rebuilding with this patch.

lgray:
OK - that was my fault, stale install. Looks good. Let me test one more thing.

ianna:
> Yes, if the length is zero, `ak.argsort` should not be returning a copy of the input array, as this won't have integer type. The method you have of creating an empty NumpyArray should work, though this should too:
> 
> ```python
> return NumpyArray(Index64(0))
> ```
> 
> which would allocate nothing, rather than one element that isn't viewed.
> 
> But now it has me wondering if Indexes and NumpyArrays with no allocation (their array is a `nullptr`, pointing at the `0` address in memory) are correctly converted to NumPy arrays. It would be weird if there were any issues with that, as we've been using that technique for so long in so many places and never encountered any problems with it. I ought to do a test.

yes, this works and is shorter:
```python
      return std::make_shared<NumpyArray>(Index64(0));
```

lgray:
OK - another test on a different file fails for a different reason now. :-)

I'll post another issue once I get to the bottom of that.

lgray:
Please go ahead, merge, and mint an rc so I can test this over the whole input fileset.

ianna:
> Please go ahead, merge, and mint an rc so I can test this over the whole input fileset.

Thanks! I'll do it as soon as the tests pass. Looking forward to seeing a new issue ;-)

PR #814, opened 2021-04-09
Added dtype argument to ones/zeros/full_like functions

### LINKED ISSUE #813: Can np.zeros_like(awkd_array) have a dtype argument?
Normally np.zeros_like can be given a dtype argument, like;
```python3
import numpy as np
cat = np.array([1, 2])
mask = np.zeros_like(bird, dtype=bool)
```
It's particularly useful when creating masks, but I've used it to ensure and array is an int array too.
When used on an awkward array this throws an error;
```python3
In [1]: import awkward

In [2]: import numpy as np

In [3]: dog = awkward.from_iter([2, 5])

In [4]: mask = np.zeros_like(dog, dtype=bool)
<ipython-input-4-8daf51a4c0a1> in <module>
----> 1 mask = np.zeros_like(dog, dtype=bool)

<__array_function__ internals> in zeros_like(*args, **kwargs)

~/Programs/anaconda3/envs/tree/lib/python3.7/site-packages/awkward/highlevel.py in __array_function__(self, func, types, args, kwargs)
   1396         See also #__array_ufunc__.
   1397         """
-> 1398         return ak._connect._numpy.array_function(func, types, args, kwargs)
   1399
   1400     @property

~/Programs/anaconda3/envs/tree/lib/python3.7/site-packages/awkward/_connect/_numpy.py in array_function(func, types, args, kwargs)
     26         return NotImplemented
     27     else:
---> 28         return function(*args, **kwargs)
     29
     30

TypeError: zeros_like() got an unexpected keyword argument 'dtype'
```
There are plenty of ways round this, but it would be neat if it did work.

HenryDayHall:
Is this the right branch to be making a pull request to? 
I still need to add the tests.

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1159,10 +1159,13 @@ def full_like(array, fill_value, highlevel=True, behavior=None, dtype=None):
     """
     if dtype is not None:
         # In the case of strings and byte strings,
-        # converting first avoids a ValueError.
+        # converting the fill avoids a ValueError.
         fill_value = dtype(fill_value)
         # Also, if the fill_value cannot be converted to the dtype
         # this should throw a clear, early, error.
+        if dtype is bool:
+            # then for bools, only 0 and 1 give correct string behavior
+            fill_value = int(fill_value)
 
     layout = ak.operations.convert.to_layout(
         array, allow_record=True, allow_other=False
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1048,7 +1048,7 @@ _ZEROS = object()
 
 
 @ak._connect._numpy.implements("zeros_like")
-def zeros_like(array, highlevel=True, behavior=None):
+def zeros_like(array, highlevel=True, behavior=None, dtype=None):
     """
     Args:
         array: Array to use as a model for a replacement that contains only `0`.
@@ -1056,6 +1056,7 @@ def zeros_like(array, highlevel=True, behavior=None):
             otherwise, return a low-level #ak.layout.Content subclass.
         behavior (None or dict): Custom #ak.behavior for the output array, if
             high-level.
+        dtype (None or type): Overrides the data type of the result.
 
     This is the equivalent of NumPy's `np.zeros_like` for Awkward Arrays.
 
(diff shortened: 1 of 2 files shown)


HenryDayHall:
Added the tests, not sure if my testing is idiomatic?

src/awkward/operations/structure.py
```diff
@@ -1234,6 +1249,9 @@ def getfunction(layout):
             return None
 
     out = ak._util.recursively_apply(layout, getfunction, pass_depth=False)
+    if dtype is not None:
+        out = strings_astype(out, dtype)
+        out = values_astype(out, dtype)
```
jpivarski, line 1254: Why is this doing both? This would chain the output through `strings_astype`, followed by `values_astype`. Is that really desirable? If `dtype=np.int64`, is `strings_astype` a no-op?

jpivarski (approved):
I have no complaints about the testing style. Just one question (below) about chaining through both `strings_astype` and `values_astype`.

src/awkward/operations/structure.py
HenryDayHall: If there is a string in the input array then  `out = strings_astype(out, dtype)` converts it (though for booleans to work properly the string needs to be '0' or '1', hence that thing on line 1068).
Where as `out = values_astype(out, dtype)` converts numeric things. 

So `strings_astype` is not a no-op, for example;
with line 1253 
`ak.ones_like([6, 'bird'], dtype=bool)` gives `<Array [True, True] type='2 * union[bool, bool]'>`
without line 1253 we get
`ak.ones_like([6, 'bird'], dtype=bool)` gives `<Array [True, '1'] type='2 * union[bool, string]'>`
I would have thought that either it should not touch the string at all or convert it fully, but I may well have misunderstood?

It could be compacted as;
`out = strings_astype(values_astype(out, dtype), dtype)`

jpivarski: That makes sense! I had forgotten that the first word ("strings" or "values") is what it's converting _from_. So this chain could be understood as "convert everything into the specified dtype."

Okay, I'm fine with it. The PR was easy to review, the tests passed, so if you're done, I'll merge it.

Thanks again!

HenryDayHall: Yes, all good to go.

PR #826, opened 2021-04-13
fixed string inequality comparison

HenryDayHall:
Really little fix.


jpivarski:
Thanks!

@all-contributors please add @HenryDayHall for code

PR #833, opened 2021-04-15
chore: cmake cleanup

henryiii:
Working on some CMake cleanups. Second comment is automated formatting, so look at the changes from the first comment only if you want to see what has changed.

- chore: cleanup cmake
- style: cmake format


CMakeLists.txt
```diff
@@ -1,133 +1,107 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
-cmake_minimum_required(VERSION 3.10...3.17)
+cmake_minimum_required(VERSION 3.14...3.18)
```
henryiii, line 3: Some of the existing settings ([`BUILD_RPATH_USE_ORIGIN`](https://cmake.org/cmake/help/latest/prop_tgt/BUILD_RPATH_USE_ORIGIN.html)) require CMake 3.14.

jpivarski:
Is there some sort of guarantee that this is the same CMake build? Some of these things are formatting (multi-line function calls, for instance), but some look substantial. Or if they were manual, intended edits, I'll defer to your expertise that they're the right things to do.

The tests pass, and that's good. The testing procedure currently doesn't test a deployment, though. It's possible for a change to successfully pass the build tests but then not create a manylinux audited wheel, for instance. I'm trying to think through all the things that _might_ go wrong, despite the success of the tests.

CMakeLists.txt
jpivarski: I think RPath stuff was added by @veprbl. I certainly have no objection to making the CMake minimal version correct, and had no reason to target CMake 3.10, unless that was a constraint imposed by manylinux. (I doubt it, as the first step in the Linux deployment is to get CMake from pip.)

henryiii:
The first commit are all manual, intentional edits. The second commit should be all automated formatting. I am somewhat hoping this is being tested; cibuildwheel would explicitly test the built wheel and make sure it also passes tests in a fresh environment; my old scripts do not do that.

PR #834, opened 2021-04-15
style: setup.cfg formatting

henryiii:
Leaving the pre-commit job commented out for now, waiting for the Flake8 commented out messages to be addressed (since it doesn't allow comments in setup.cfg).

Mostly follows a standard ordering and spacing, also can catch a few missing fields and classifiers. Checks for consistency, as well.


setup.cfg
```diff
@@ -58,19 +61,18 @@ exclude =
...
 
 [flake8]
-select = C,E,F,W,B,B9,T
 # F401 can be fixed by adding __all__'s
 # E712 should be dropped, False/True should compared by "is" or checked as booleans, but not "=="
 # E711 should be dropped, always compare None with is, not ==.
 # F841 should be removed, use nothing or _ or _<name> for an unused variable
 # E402 can be improved in the future with isort
 # F403 should be removed, import * is evil
 # B007 should be removed, variables that are not used should start with _
```
jpivarski: I'll investigate the reasons for these and perhaps lift some of them.

jpivarski: Oh wait—I have to fix these before work on setup.cfg can be completed? In that case, I'll do it now.

henryiii: No, you don't have to, just once they are cleared, we can enable automatic formatting here and forget about it. That also includes some nice consistency checks.

When you add `__all__`, I would also add `def __dir__(): return __all__`, as that really nicely cleans up tab completion in REPLs like IPython (Python 3.7+ only, but no harm on older Pythons).

jpivarski: Well, I already started and code-cleanup is a good thing, anyway.

I'll add a `__dir__` for each `__all__`.

jpivarski: Okay, all of these have been applied in #844 (including the `__dir__` functions), and it looks like it's going to pass tests so I'll enable auto-squash-and-merge on that.

henryiii: :tada: - I'll rebase #834 and enable automatic setup.cfg styling once this is in.

PR #835, opened 2021-04-15
support 'datetime64' and 'timedelta64' types

### LINKED ISSUE #367: Support for datetime[*] numpy dtype
Currently, a pandas DataFrame with a column of eg. datetime[ns] dtype is resulting in an obscure ValueError:
`ValueError: cannot include dtype 'M' in a buffer`

```python
values = {'time':  ['20190902093000','20190913093000','20190921200000']}
df = pd.DataFrame(values, columns = ['time'])
df['time'] = pd.to_datetime(df['time'], format='%Y%m%d%H%M%S')
df.dtypes
```

```bash
time       datetime64[ns]
dtype: object
```

```python
# current workaround
df['time'] = df.time.values.astype(int)
akarr = ak.Record({c: df[c].values for c in df.columns})
```

ianna:
issue https://github.com/scikit-hep/awkward-1.0/issues/367


tests/test_0835-datetime-type.py
```diff
@@ -0,0 +1,63 @@
...
+def test_date_time():
+
+    numpy_array = np.array(
+        ["2020-07-27T10:41:11", "2019-01-01", "2020-01-01"], "datetime64[s]"
+    )
+
+    array = ak.Array(numpy_array)
+    assert (
+        str(array.type)
+        == '3 * datetime64[parameters={"__array__": "datetime64", "__datetime64_data__": "<M8[s]", "__datetime64_unit__": "1s"}]'
```
ianna: the type is correct, I'm not sure if all these parameters are useful.

tests/test_0835-datetime-type.py
```diff
@@ -0,0 +1,63 @@
...
+    )
+    assert array.tolist() == [
+        1595846471,
+        1546300800,
+        1577836800,
```
ianna: these are the number of `datetime64` units: seconds in this case. The units and the step are stored in ` __datetime64_unit__` parameter.

ianna (commented):
@jpivarski - I'm not sure if I should box it differently or leave it as `uint64_t`

ianna:
It appears the test was failing due to a missing `"dtype"` attribute. The check if it is present is added in `content.cpp#2226`. 
```python
____________________________________________________________ test_sort_zero_length_arrays _____________________________________________________________

    def test_sort_zero_length_arrays():
        array = ak.layout.IndexedArray64(
>           ak.layout.Index64([]), ak.layout.NumpyArray([1, 2, 3])
        )
E       AttributeError: 'list' object has no attribute 'dtype'

tests/test_0074-argsort-and-sort.py:493: AttributeError

```

tests/test_0835-datetime-type.py
ianna: another approach is to list them as `np.datetime64`:
```python
assert array.tolist() == [
         np.datetime64("2020-07-27T10:41:11"),
         np.datetime64("2019-01-01T00:00:00"),
         np.datetime64("2020-01-01T00:00:00"),
     ]
```

jpivarski: It would be better if `tolist` makes a list of `np.datetime64` objects, but that would be automatic if `one_dimensional[0]` returns a `np.datetime64`. The `tolist` uses a recursive function defined in src/python/content.cpp that just calls `getitem_at` with appropriate `box`ing.

ianna: yes, that's what is done in the last commit, I think :-)

sterbini:
I would be very interested on the `datetime` support. Thanks!

ianna (commented):
there are two issues to fix: 

- [ ] precision loss: for example, when a `np.datetime64("2020-05")` is converted to seconds `np.datetime64("2020-05-01T20:56:24.000000")`
- [x] build `timedelta64` for a high-level `ak.ArrayBuilder`

ianna:
> * [ ]  precision loss: for example, when a `np.datetime64("2020-05")` is converted to seconds `np.datetime64("2020-05-01T20:56:24.000000")`

```python
>>> dt = np.datetime64("2020-05")
>>> dt.dtype
dtype('<M8[Y]')
>>> dt2 = np.datetime64("2020-05-01T20:56:24.000000")
>>> dt2
numpy.datetime64('2020-05-01T20:56:24.000000')
>>> dt2.astype(np.datetime64(1, 'M'))
numpy.datetime64('2020-05')
```

I think it could be left as is. @jpivarski ?

If not, possible solutions are to:
1. convert all incoming `np.datetime64` to `'s'` - **what if a user may want to keep given units?**
```python
>>> dt.astype(np.int64)
604
>>> dt2.astype(np.int64)
1588366584000000
>>> dt.astype(np.datetime64(1, 'us')).astype(np.int64)
1588291200000000
>>> dt.astype(np.datetime64(1, 's')).astype(np.int64)
1588291200
```
2. recalculate it later:
```python
>>> def leap_years_between(start,end):
...     if start < end:
...             return leap_years_before(end) - leap_years_before(start + 1)
...     else:
...             raise ValueError
... 
>>> def leap_years_before(year):
...     if year > 0:
...             year = year - 1
...             return (year / 4) - (year / 100) + (year / 400)
...     else:
...             raise ValueError
... 
```

jpivarski:
> I think it could be left as is. @jpivarski ?

I don't see anything wrong here. Maybe you should ask the potential users of datetimes?

src/awkward/_util.py
```diff
@@ -490,7 +490,32 @@ def completely_flatten(array):
...
+                [
+                    np.timedelta64(
+                        x,
+                        array.format[
+                            array.format.index("[") + 1 : array.format.index("]")
+                        ],
+                    )
+                    for x in array
+                ],
+            )
```
jpivarski: This is a Python iteration over the whole array, not vectorized: `for x in array` where `isinstance(array, ak.layout.NumpyArray)`. This would be a serious performance bomb.

I just checked in the src/python/content.cpp and there's no NumpyArray method that converts it into an `np.ndarray`, circumventing the buffer protocol, following `to_cupy` and `to_jax`. That's the function this would need to call: (1) check to see that the `array.format.startswith("M")` (or `"m"`) and (2) call the special method instead of just `asarray`.

src/awkward/operations/convert.py
```diff
@@ -958,7 +967,24 @@ def to_list(array):
...
+                for x in array
+            ]
+        elif array.format.startswith("m"):
+            return [
+                np.timedelta64(
+                    x,
+                    array.format[array.format.index("[") + 1 : array.format.index("]")],
+                )
+                for x in array
+            ]
```
jpivarski: Same thing again: this is not only iterating over the array in Python, but constructing each NumPy object from a type that is parsed for every entry in the array. It needs to be a zero-copy wrapping.

tests/test_0835-datetime-type.py
```diff
@@ -0,0 +1,408 @@
...
+        assert ak.to_numpy(array)[i] == numpy_array[i]
+
+
+def test_datetime64_ArrayBuilder():
+    builder = ak.layout.ArrayBuilder()
+    dt = np.datetime64("2020-03-27T10:41:12", "25us")
+    dt1 = np.datetime64("2020-03-27T10:41", "15s")
+    dt2 = np.datetime64("2020-05")
+    # FIXME: do we need to support this?
+    # builder.datetime64(dt.astype(np.int64), "datetime64[s]")
```
jpivarski: No, if someone passes an integer to `builder.datetime64`, they shouldn't expect it to work.

tests/test_0835-datetime-type.py
```diff
@@ -0,0 +1,408 @@
...
+    ]
+
+
+def test_highlevel_datetime64_ArrayBuilder():
+    builder = ak.ArrayBuilder()
+    dt = np.datetime64("2020-03-27T10:41:12", "25us")
+    dt1 = np.datetime64("2020-03-27T10:41", "15s")
+    dt2 = np.datetime64("2020-05")
+    # FIXME: do we need to support this?
+    # builder.datetime64(dt.astype(np.int64), "datetime64[s]")
```
jpivarski: No (see above).

jpivarski (changes requested):
I tried it out and ran into a few bugs, most likely minor, and added them to the testing suite as `test_more`.

I also made a modification to the repr format:

```
<Array [2021-06-03T10:00, ... 2021-06-03T13:00] type='4 * datetime64'>
```

and

```
<Array [60 minutes, 60 minutes, 60 minutes] type='3 * timedelta64'>
```

The string representation of datetime/timedelta types don't agree with [Datashape](https://datashape.readthedocs.io/en/latest/types.html#date-time-and-datetime), though. Is it possible to fix this? For one thing, the type strings should drop the "64", but is there a way to get timezone information in datetime (e.g. `datetime[tz='UTC']`)?

There is one significant problem, though: whenever we get data from a datetime/timedelta NumpyArray into an `np.ndarray`, it's through an expensive-looking loop in Python. This should be a zero-copy view. The NumpyArray class already has methods like `to_cupy` and `to_jax` that circumvent the buffer interface (and have to be called explicitly by the code layer above it): we need a similar one for `to_numpy` that would be called when the dtype is not something that can be expressed in a buffer.

Actually, we can short-cut this: I created a `view_int64` property on NumpyArray, which just drops the datetime/timedelta metadata so that you can send it through a buffer and add the metadata back in:

```python
>>> akarray
<Array [2021-06-03T10:00, 2021-06-03T11:00] type='2 * datetime64'>
>>> akarray.layout
<NumpyArray format="M8[m]" shape="2" data="0x 78ad9c01 00000000 b4ad9c01 00000000" at="0x562eaa6b1e10"/>
>>> akarray.layout.view_int64
<NumpyArray format="l" shape="2" data="27045240 27045300" at="0x562eaa6b1e10"/>
>>> np.asarray(akarray.layout.view_int64)
array([27045240, 27045300])
>>> np.asarray(akarray.layout.view_int64).view(akarray.layout.format)
array(['2021-06-03T10:00', '2021-06-03T11:00'], dtype='datetime64[m]')
```

Use this to fix the two Python loops over array data.

src/awkward/_util.py
ianna: done, thanks!

src/awkward/operations/convert.py
ianna: done, thanks!

tests/test_0835-datetime-type.py
ianna: ok, removed

tests/test_0835-datetime-type.py
ianna: ok, removed

src/awkward/operations/convert.py
```diff
@@ -232,6 +232,12 @@ def to_numpy(array, allow_missing=True):
             ]
         )
 
+    elif (
+        str(ak.operations.describe.type(array)) == "datetime64"
+        or str(ak.operations.describe.type(array)) == "timedelta64"
+    ):
+        return numpy.array([array[i] for i in range(len(array))])
```
jpivarski: `numpy.array([array[i] for i in range(len(array))])` in vectorized form is `numpy.arange(len(array))` (and you have to specify the dtype for consistency on Windows).

But why is `ak.to_numpy` returning `[0, 1, 2, 3, ...]` for datetimes/timedeltas?

jpivarski: Also, if you fix the string representation ("datetime" and "timedelta" are the Datashape words; they leave off the "64"), then the type-check will be broken: check the format for `startswith("m")`.

include/awkward/builder/Builder.h
```diff
@@ -78,6 +78,14 @@ namespace awkward {
...
       complex(std::complex<double> x) = 0;
 
+    /// @brief Adds a datetime64 value `x` to the accumulated data.
+    virtual const BuilderPtr
+      datetime64(int64_t x, const std::string& unit) = 0;
+
+    /// @brief Adds a timedelta64 value `x` to the accumulated data.
+    virtual const BuilderPtr
+      timedelta64(int64_t x, const std::string& unit) = 0;
+
```
jpivarski, line 88: For consistency, drop the "64" in these method names...

jpivarski (approved):

src/awkward/operations/convert.py
ianna: done

include/awkward/builder/Builder.h
ianna: done

src/awkward/operations/convert.py
```diff
@@ -970,10 +970,16 @@ def to_list(array):
     elif isinstance(array, ak.layout.NumpyArray):
         if array.format.upper().startswith("M"):
             return (
-                ak.nplike.of(array)
-                .asarray(array.view_int64)
-                .view(array.format)
-                .tolist()
+                [
```
ianna, line 973: @jpivarski - couldn't figure out why `tolist` returns integers for an array with `M` format...

src/python/content.cpp
```diff
@@ -2313,14 +2285,22 @@ make_NumpyArray(const py::handle& m, const std::string& name) {
...
-        else if (module.rfind("pandas.", 0) == 0) {
-          if (py::cast<bool>(anyarray.attr("_is_datelike_mixed_type"))) {
-            return NumpyArray_from_pandas_datetime(name, anyarray.attr("values"), identities, parameters);
-          }
-        }
 
         py::array array = anyarray.cast<py::array>();
 
+        if (!PyObject_CheckBuffer(anyarray.ptr())) {
+          // anyarray does not support buffer protocol
```
ianna, line 2292: @jpivarski - this catches a `pandas` case as well as a `builtin` case beneath:
```python
array = ak.layout.NumpyArray(
        [
            np.datetime64("2019-09-02T09:30:00"),
            np.datetime64("2019-09-13T09:30:00"),
            np.datetime64("2019-09-21T20:00:00"),
        ]
    )

```

ianna (commented):
@jpivarski - please, have a look when you have time. thanks!

ianna:
@jpivarski -here is an example where a pyarrow `date64` array is converted correctly to `NumpyArray`, but not to `Array` due to the units mismatch. I'm not sure how realistic it is.
```python
>>> parr = pa.array([datetime.datetime(2002, 1, 23), datetime.datetime(2019, 2, 20)], type=pa.date64())
>>> parr.type
DataType(date64[ms])
>>> narr = ak.layout.NumpyArray(parr)
>>> narr
<NumpyArray format="M8[D]" shape="2" data="Wed Jan 23 00:00:00 2002
 Wed Feb 20 00:00:00 2019
" at="0x000109071000"/>
>>> array = ak.Array(parr)
>>> array
<Array [1970-01-01T00:16:51.744000000, ... ] type='2 * datetime'>
```

src/awkward/operations/convert.py
drahnreb: This is somewhat expected behavior when providing `ns` resolution. 

```python
>>> np.datetime64('2019-09-02T09:30:00', 'ns').tolist()
1567416600000000000
>>> np.datetime64("2019-09-02T09:30:00.0000000").tolist()
1567416600000000000
```

It's a very [old and complex (if not weird)](https://stackoverflow.com/questions/13703720/converting-between-datetime-timestamp-and-datetime64/21916253#21916253) behavior.
Numpy considers datetime still [experimenting](https://numpy.org/doc/stable/reference/arrays.datetime.html), and this might change anytime, but I would prefer to mimic rather than sacrificing non-vectorized performance.

drahnreb:
> but is there a way to get timezone information in datetime (e.g. `datetime[tz='UTC']`)?

Numpy deprecated storing timezone information (at the end of [this paragraph](https://numpy.org/doc/stable/reference/arrays.datetime.html#basic-datetimes)). Overall, I would say a feature that is not really relevant. Though, `pyarrow` [does support timezones](https://arrow.apache.org/docs/python/generated/pyarrow.timestamp.html) and any persisted data will lose this information.

(/cc @jpivarski )

jpivarski:
@ianna To answer your question, this is not an intended use-case (and it works "by accident" because this pyarrow array happens to be flat and interpretable as a buffer):

```python
>>> parr = pa.array([datetime.datetime(2002, 1, 23), datetime.datetime(2019, 2, 20)], type=pa.date64())
>>> parr.type
DataType(date64[ms])
>>> narr = ak.layout.NumpyArray(parr)
```

The proper way to convert pyarrow data into Awkward Arrays is with `ak.from_arrow`.

The issues with pyarrow were because we rely on `to_pandas_dtype` to get a NumPy dtype (it's not well named) and this method is completely wrong for a lot of units. I'm going to file a bug report to the Arrow project.

The appearance of correct times in the low-level NumpyArray view was due to another bug—the two bugs cancelled—which was that the `scale` from datetime_util was an integer, but sometimes you need to scale down, not just up. I've promoted that quantity to `double` type (including the kernel that it's used in).

I think it's done! I noticed that we can't take `datetime` objects in the `ak.from_iter` constructor, but neither can NumPy (it makes a `dtype="O"` array). That might be nice to have, but it would be for another PR someday.

Congrats! This was a long time coming!

jpivarski:
Here's the Arrow issue: https://issues.apache.org/jira/browse/ARROW-13040

Until we start taking whichever version has this fix as a minimum (Arrow 5? 6?), we'll have to use special cases:

https://github.com/scikit-hep/awkward-1.0/blob/a5880b930b581bea95cee94c659764d540c72b5d/src/awkward/operations/convert.py#L2591-L2608

and

https://github.com/scikit-hep/awkward-1.0/blob/a5880b930b581bea95cee94c659764d540c72b5d/src/awkward/operations/convert.py#L2819-L2829

src/awkward/operations/convert.py
ianna: thanks! fixed in https://github.com/scikit-hep/awkward-1.0/pull/911

PR #840, opened 2021-04-16
Add an ak.drop_none function.

### LINKED ISSUE #832: Add a `drop_none()`
Originally mentioned in https://github.com/scikit-hep/awkward-1.0/issues/490#issuecomment-712250246. Mainly all I'm looking for is a nice shortcut for `array[~ak.is_none(array, axis=-1)]` (I think this would cover any of my own use cases), although the default behavior (`axis=None`?) for a `drop_none` function should probably be to do this on every axis. A particularly important case for this function is that functions like `np.histogram()` and `plt.hist()` do not handle masked arrays properly (https://github.com/numpy/numpy/issues/10019).

jpivarski:


jpivarski:
Nope, it's not going to be as easy as modifying `flatten`. There's too much that hasn't been thought through.

PR #844, opened 2021-04-19
Removed "ignore" from some Flake8 complaints

### LINKED ISSUE #843: Remove "ignore" from some Flake8 complaints
The full list is here: https://github.com/scikit-hep/awkward-1.0/pull/834/files#r616043382

jpivarski:


dev/generate-cuda.py
```diff
@@ -243,9 +243,9 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
     elif node.__class__.__name__ == "BitAnd":
         code = "&"
     elif node.__class__.__name__ == "Constant":
-        if node.value == True:
+        if node.value is True:
             code = "true"
-        elif node.value == False:
+        elif node.value is False:
```
henryiii, line 248: Excellent! This is much better. :heart:

```pycon
>>> True is 1
False
>>> True == 1
True
```

jpivarski: It was student-contributed code that got through because of the blanket-ignore in Flake8. Even if I'm replacing some cases with spot `# noqa` ignores, the general rule would still apply and would be good feedback for contributors.

PR #847, opened 2021-04-22
Bugfix: fix #846 - matrix multiplication with numpy array 

### LINKED ISSUE #846: `np.matmul` produces wrong result with NumPy RHS
The following multiplication produces a result with too few entries in the underlying contents, leading to errors when accessing items:
```python
transform = ak.Array(
    [
        [0, 1, 0],
        [1, 0, 0],
        [0, 0, 1],
    ]
)

vector = ak.Array(
    np.r_[4,5,6].reshape(3, 1)    
)

result = np.matmul(transform, vector)
result[0]
```

This does not happen if I create the `vector` Array by hand:
```python
vector = ak.Array(
    [
        [4],
        [5],
        [6]
    ]
)
result = np.matmul(transform, vector)
result[0]
```

This is also true of `transpose(vector) @ transform`

```python
>>> import awkward as ak
>>> ak.__version__
'1.2.2'
```

agoose77:
This PR tracks the fix for #846 (assuming that this is indeed a bug). For now, it just adds a test to demonstrate the issue.

### DIFF
--- a/src/awkward/_connect/_numpy.py
+++ b/src/awkward/_connect/_numpy.py
@@ -144,12 +144,12 @@ def array_ufunc(ufunc, method, inputs, kwargs):
         if custom is not None:
             return lambda: adjust(custom, inputs, kwargs)
 
-        inputs = [deregulate(x) for x in inputs]
-
         if ufunc is numpy.matmul:
             custom_matmul = getfunction_matmul(inputs)
             if custom_matmul is not None:
                 return custom_matmul
+
+        inputs = [deregulate(x) for x in inputs]
 
         if all(
             isinstance(x, ak.layout.NumpyArray)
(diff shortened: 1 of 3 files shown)


agoose77:
LGTM :+1:

jpivarski:
@all-contributors please add @agoose77 for test, code

agoose77:
Thanks for merging :tada:

jpivarski:
It only went so long because I hadn't noticed that it wasn't already merged. The fact that you had to fix a bug disabled the auto-merge (as it should, I think). So in my mind, it was merged—I was surprised to see it in the list of open PRs.

agoose77:
Oh wow, that's a good idea from a security perspective actually.

Issue #849, opened 2021-04-26
argsort of masked array has wrong layout


agoose77:
Reproducer on Binder: [![Binder](https://mybinder.org/badge_logo.svg)](https://mybinder.org/v2/gist/agoose77/0d2765f6d3b8e5b5b0cb75b28b7a95d0/HEAD?filepath=awkward-bug.ipynb)

Given an Awkward array and a mask:
```python 
label = ak.Array([[0, 1, 2, 3], [3, 3, 3, 2, 1]])
is_valid = label != 3
```
the result of `argsort(label.mask[is_valid])` is not a valid Array. 

It appears that the result does not maintain the mask, and so the `offsets` node in the layout is invalidated (it still refers to the masked structure).

```python
>>> label.mask[is_valid].layout
<ListOffsetArray64>
    <offsets><Index64 i="[0 4 9]" offset="0" length="3" at="0x0000026d8b10"/></offsets>
    <content><ByteMaskedArray valid_when="true">
        <mask><Index8 i="[1 1 1 0 0 0 0 1 1]" offset="0" length="9" at="0x0000026a7ad0"/></mask>
        <content><NumpyArray format="l" shape="9" data="0 1 2 3 3 3 3 2 1" at="0x0000026dab20"/></content>
    </ByteMaskedArray></content>
</ListOffsetArray64>
>>> ak.argsort(label.mask[is_valid]).layout
<ListOffsetArray64>
    <offsets><Index64 i="[0 4 9]" offset="0" length="3" at="0x0000026d8b10"/></offsets>
    <content><NumpyArray format="l" shape="5" data="0 1 2 1 0" at="0x0000028d0710"/></content>
</ListOffsetArray64>
```

I would expect the mask to persist after the call to `argsort` inasmuch as the resulting array has the same "shape" as the original array. 

### Awkward Array version
```python
>>> import awkward as ak
>>> ak.__version__
1.2.2
```

jpivarski:
@ianna, the ByteMaskedArray node is somehow being lost. I don't think this has anything in particular to do with using the `.mask` mechanism; it's probably an issue with ByteMaskedArray as opposed to IndexedOptionArray.

ianna:
@agoose77 - thanks for spotting the bug! https://github.com/scikit-hep/awkward-1.0/pull/850 should fix it.

ianna:
fixed via #850

PR #850, opened 2021-04-26
masked array sort and argsort bug fix 

ianna:
address issue #849

### DIFF
--- a/src/libawkward/array/ByteMaskedArray.cpp
+++ b/src/libawkward/array/ByteMaskedArray.cpp
@@ -1280,86 +1280,13 @@ namespace awkward {
                              bool ascending,
                              bool stable,
                              bool keepdims) const {
-    if (length() == 0 ) {
-      return shallow_copy();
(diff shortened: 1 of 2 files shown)


agoose77:
That is a _really_ nice delta! 

image

Thanks for fixing this so quickly.

jpivarski (approved):
@agoose77 

> That is a _really_ nice delta!
> 
> image

Well, this PR removes a custom implementation of argsort for ByteMaskedArrays and replaces it with converting the ByteMaskedArray into IndexedOptionArray, then using the known-to-work IndexedOptionArray's argsort. That's good—it definitely fixes the issue and it reduces the number of code paths—but a future improvement could increase the line count again by attempting to implement ByteMaskedArray's argsort correctly. Such a thing, if it happens at all, might be motivated by performance.

As it is, I'm completely in favor of merging this PR. Any time you want to, @ianna, you can merge it. Thanks!

PR #853, opened 2021-04-28
Performance optimization for Bit/ByteMaskedArray ufuncs.

### LINKED ISSUE #852: Very slow ufuncs, especially when masked
The ufunc in awkward seem to be too slow for what they are actually doing. Compared to the Numpy equivalent, there is a difference of a few magnitudes of order. I have studied this with `np.logical_not` on a simple one dimensional array. Other ufuncs appeared to have similar behavior.
awkward_performance

  <details>
    <summary>Python script to produce the plot</summary>
    

```python
#!/usr/bin/env python3

import numpy as np
import numpy.ma as ma
import awkward as ak
import matplotlib.pyplot as plt
from timeit import Timer

def timeit(func, number=100):
    return Timer(func).timeit(number=number)

testsizes = np.logspace(0, 6, 20).astype(int)

numpy_time = []
for size in testsizes:
    a = np.full(size, True)
    numpy_time.append(timeit(lambda: np.logical_not(a)))

numpymask_time = []
for size in testsizes:
    a = ma.masked_array(np.full(size, True), mask=[True] * size)
    numpymask_time.append(timeit(lambda: np.logical_not(a)))

awkward_time = []
for size in testsizes:
    a = ak.Array(np.full(size, True))
    awkward_time.append(timeit(lambda: np.logical_not(a)))

awkwardmask_time = []
for size in testsizes:
    a = ak.Array(np.full(size, True)).mask[[True] * size]
    awkwardmask_time.append(timeit(lambda: np.logical_not(a)))

plt.plot(testsizes, numpy_time, label="Numpy", color="C0", linestyle="--")
plt.plot(testsizes, numpymask_time, label="Numpy masked", color="C1", linestyle="--")
plt.plot(testsizes, awkward_time, label="Awkward", color="C0")
plt.plot(testsizes, awkwardmask_time, label="Awkward masked", color="C1")
plt.xscale("log")
plt.yscale("log")
plt.xlabel("Array length")
plt.ylabel("Time per 100 iterations in s")
plt.legend(title="Logical not with", loc="upper left")
plt.show()
```
  </details>

In https://github.com/scikit-hep/awkward-1.0/issues/442#issuecomment-689669226 the plot shows that the numpy and awkward execution times for large inputs becomes equal. However, I do not see this in my case. In above plot the blue lines might line up for input sizes far outside of what I am using, and the orange ones do not look like they will line up at all.

I think this is important because of this example timing I got from line_profiler using it on some of my code, `data` and `mask` are simple one dimensional awkward arrays with no further layout other than that they might use an OptionArray, times are in µs:
```
Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   130     34668    6706606.0    193.5      2.5              a = ak.is_none(data)
   131     34668   22385934.0    645.7      8.2              b = ~a
   132     34668  190116583.0   5483.9     69.9              mask = mask & b
```
and comepare this to
```
   130     35939   12825258.0    356.9     21.6              a = np.asarray(ak.is_none(data))
   131     35939     172479.0      4.8      0.3              b = ~a
   132     35939    4658937.0    129.6      7.8              c = np.asarray(mask)
   133     35938     108790.0      3.0      0.2              mask = c & b
```

This is all using awkward 1.2.2

jpivarski:


jpivarski:
The build fail, if it even reproduces itself, can't be related to the code change. I also [tested on main](https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=6343&view=results) and there were no issues. Somehow, the reference count is coming up wrong in Numba in Python 3.6 (Mac and Linux).

PR #860, opened 2021-05-10
corrected the class names in documentation

ianna:
@jpivarski - I've noticed the name mismatch when looking at the links you've posted on Slack.


jpivarski (approved):
That's right: this was an uncaught copy-paste error. Thanks for catching it!

jpivarski:
(Auto-merge is on, so this will be merged as soon as the tests pass. It will stop if any fall or if you commit something else before it's done.)

ianna (commented):
@jpivarski - it looks like the auto merge waits for the tests that will never run for this PR since the changes are only in documentation.

jpivarski:
Ah, I see. Nothing works as well as planned.

PR #867, opened 2021-05-17
Fix #865, only build forms for columns that are requested in lazy mode

### LINKED ISSUE #865: `ak.from_parquet` lazy loading does not consider `columns` arg early enough
When using `ak.from_parquet()` with `lazy=True` the function seems to take a different code path that ignores the `columns` argument in the early stages. Not sure if this is a bug or intended behavior because `ak.layout.VirtualArray` requires it (but then it seems to be undocumented - apart from datetime fields generally not supported)?

The following fails (a particularly useful example to exclude fields of yet (#835 🎉 ) unsupported datatypes):
```python
>>> import awkward as ak
>>> ak.__version__
'1.2.3'
>>> example_df_with_datetime_col = pd.DataFrame({
             'timestamp': [1621164368000000000],
             'datetime': [pd.to_datetime(1621164368000000000)]
})
>>> example_df_with_datetime_col.to_parquet('table_with_datetime_col.parquet')
>>> ak.from_parquet('table_with_datetime_col.parquet', columns=['timestamp'], lazy=True)

ValueError: cannot convert NumPy dtype with kind M into a NumpyForm
(https://github.com/scikit-hep/awkward-1.0/blob/1.2.3/src/libawkward/Content.cpp#L232)

# in contrast:
>>> ak.from_parquet('table_with_datetime_col.parquet', columns=['timestamp'])  # works as expected
>>> ak.from_parquet('table_with_datetime_col.parquet')  # fails as expected
ValueError: cannot include dtype 'M' in a buffer
```

drahnreb:
Possible fix for #865.
This passes my tests. Let me know if that is reasonable and if a test needs to be added.

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -3596,12 +3596,15 @@ def from_parquet(
 
     if columns is None:
         columns = all_columns
+    pa_fields = []
     for x in columns:
         if x not in all_columns:
             raise ValueError(
                 "column {0} not found in schema".format(repr(x))
                 + ak._util.exception_suffix(__file__)
             )
+        pa_fields.append(schema.field(x))
+    schema = pyarrow.schema(pa_fields)
 
     if len(row_groups) == 0:
         out = ak.layout.RecordArray(


agoose77:
@drahnreb thanks for contributing a fix! I'm not a maintainer, but I know that the general protocol is to add a test for any bug found. Is that something that you could do here?

jpivarski:
It looks like this takes a loop that was originally used to check for non-matching strings in the `columns` argument and adds to it a rewrite of `schema`, which is guaranteed to exist because one of the three paths above it,

   * `if isinstance(source, str) and os.path.isdir(source)`
   * `if not isinstance(source, str) and isinstance(source, Iterable)`
   * `if multimode is None`

has to trigger, and each one defines it. (At the moment, I don't see what would happen if the iterable in the middle one has zero files. It looks like `schema` would be None, which would certainly raise an error somewhere. That might be a case that needs an explicit error message, though.)

------------------

As for testing, it's a good idea, especially when it's easy. This case isn't easy, since you need to have a file that would trigger the bug, and that file would have to be created by or added to the testing suite (a frequent problem for testing Uproot). The test has to fail without the fix and succeed with it, which would be possible with a file containing datetimes until #835 is merged. Then it would succeed with or without the fix—a temporary test. The fact that @drahnreb's workflow is unblocked by this fix _is_ a temporary test. The value of having a similar test in the testing suite is to make sure that the bug never regresses, but it wouldn't have the power to make such a guarantee after #835 is merged.

Meanwhile, this is a heavily trafficked code path: every call to `ak.from_parquet` has to go through this point. It's testing in the sense of being sure there are no simple errors like misspelled variables.

So if you find a way to add a test (e.g. create a small Parquet file with a datetime column using pyarrow directly, with appropriate `pytest.importorskip`), then that would be great. But given the above, I'm not going to require it as a condition for merging the PR.

@drahnreb, let me know if you're done.

drahnreb:
> It looks like this takes a loop that was originally used to check for non-matching strings in the `columns` argument and adds to it a rewrite of `schema`, which is guaranteed to exist because one of the three paths above it,
> has to trigger, and each one defines it.

Sorry for hijacking it, but it looked so tempting with the given conditions and because of its proximity to the `columns is None` redefinition that has similar meaning (I hesitate to comment to keep your neat code style and it looked verbose enough).

> The test has to fail without the fix and succeed with it, which would be possible with a file containing datetimes until #835 is merged. Then it would succeed with or without the fix—a temporary test. The fact that @drahnreb's workflow is unblocked by this fix _is_ a temporary test. The value of having a similar test in the testing suite is to make sure that the bug never regresses, but it wouldn't have the power to make such a guarantee after #835 is merged.

For the exact same reason I did not include a test. I wanted to avoid any "intermediate" test as this PR is already in the pipeline and couldn't think of an appropriate alternative.

> So if you find a way to add a test (e.g. create a small Parquet file with a datetime column using pyarrow directly, with appropriate `pytest.importorskip`), then that would be great.

We could try purposefully corrupt one column of a written parquet and try to read the file partially (with/without the corrupt one). But this sounds complicated, is fragile in its setup (we would need to find a way to corrupt the column metadata so that the column chunk for _one_ row group is lost) and begs the question of what is being tested (recover a corrupt file vs. testing against this issue).

> @drahnreb, let me know if you're done.

I'm done, thanks for your feedback 👍🏼

jpivarski:
Thanks! Then I'll accept it as-is.

PR #868, opened 2021-05-17
Matrix multiplication of a non-array vector.

### LINKED ISSUE #881: Matrix multiplication of non-array object fails
:information_source:  This issue has already been reproduced, so this is just a placeholder for posterity.

Currently, this test fails:
```python
import numpy as np
import awkward as ak
import vector

def test():
    point = ak.Record({'x': 1, 'y': 2, 'z': 3}, with_name="Vector3D")    
    point @ vector.obj(x=1, y=0, z=2)
```

I think a non-`vector` equivalent would be this:
```python
import numpy as np
import awkward as ak
from collections import namedtuple

Point = namedtuple("Point", "x y z")


def point_dot(left, right):
    return left.x*right.x + left.y*right.y + left.z*right.z


def test():
    point = ak.Record({'x': 1, 'y': 2, 'z': 3}, with_name="Vector3D")    
    ak.behavior[np.matmul, "point", Point] = point_dot
    
    point @ Point(1, 0, 2)
```
I think what's happening is that Awkward expects the arguments to be `cast`-able to an Array, e.g. https://github.com/scikit-hep/awkward-1.0/blob/22501ba218646dc24dc515c4394eb22f126d340d/src/awkward/_connect/_numpy.py#L43-L53

I expected the test to succeed, because `vector` implements these behaviours: https://github.com/scikit-hep/vector/blob/8951b9bcfe3b6cc9fbbfc1a32fd84bdf50739682/src/vector/_backends/awkward_.py#L875-L902

jpivarski:
@agoose77 Please add a test for the error case you encountered, to see if this fix is sufficient. Thanks!

### DIFF
--- a/src/awkward/_connect/_numpy.py
+++ b/src/awkward/_connect/_numpy.py
@@ -294,6 +294,8 @@ def getfunction_matmul(inputs):
         ak._util.recursively_apply(
             x, (lambda _: _), pass_depth=False, numpy_to_regular=True
         )
+        if isinstance(x, (ak.layout.Content, ak.layout.Record))
+        else x
         for x in inputs
     ]
 
--- a/tests/test_0868-matrix-multiplication-of-vector.py
+++ b/tests/test_0868-matrix-multiplication-of-vector.py
@@ -1,0 +1,19 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from __future__ import absolute_import
+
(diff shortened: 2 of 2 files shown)


agoose77:
See #881 for the issue description :) If you agree that this is an expected use case (i.e. arguments to ufuncs are not array-castable) then I can push the test here. Otherwise, perhaps `vector` needs to provide a cast for `VectorObjectND`?

PR #874, opened 2021-05-19
Should strings from __getitem__ be Python str? (and bytes?)

### LINKED ISSUE #873: Have elements of an `Array` of strings be instances of `str`
In numpy, the elements of an `np.ndarray` of strings are `np.str_`, which is a subclass of `str`:

```python
>>> import numpy as np
>>> np.array(['asdf'])[0]
'asdf'
>>> type(np.array(['asdf'])[0])
<class 'numpy.str_'>
>>> isinstance(np.array(['asdf'])[0], str)
True
```

In awkward, the elements come up as `awkward.behaviors.string.CharBehavior`, which is not a `str`:

```python
>>> import awkward as ak
>>> ak.__version__
'1.2.3'
>>> ak.Array(['asdf'])[0]
'asdf'
>>> type(ak.Array(['asdf'])[0])
<class 'awkward.behaviors.string.CharBehavior'>
>>> isinstance(ak.Array(['asdf'])[0], str)
False
```

This makes it inconvenient to pass these values to functions that depend on `isinstance(input, str)` to detect strings in their arguments. Getting around this is straightforward (just wrap with `str()`), but is it possible to make the default behavior an instance of `str`?

I guess the same would go for `awkward.behaviors.string.ByteBehavior` and `bytes`, but that issue would never come up in any of my use cases.

jpivarski:


jpivarski:
See discussion on #873. Don't merge this PR until we're sure it's what we want to do!

jpivarski:
Nobody has spoken up on #873, but I've become convinced. NumPy returns a custom type,

```python
>>> type(numpy.array(["one", "two", "three"])[0])
<class 'numpy.str_'>
```

but this type behaves like a Python `str`:

```python
>>> numpy.array(["one", "two", "three"])[0] + numpy.array(["one", "two", "three"])[1]
'onetwo'
```

and _not_ like an array. In fact, `np.str_` is a direct subclass of `str` with a few number-like features:

https://github.com/numpy/numpy/blob/75f852edf94a7293e7982ad516bee314d7187c2d/numpy/__init__.pyi#L3418-L3432

Namely,

```python
>>> numpy.array(["one", "two", "three"])[0].real
'one'
>>> numpy.array(["one", "two", "three"])[0].imag
''
```

So sure, we'll return Python string objects from `__getitem__`.

PR #884, opened 2021-05-28
C++ refactoring: Index and Identities (Identifier)

ioanaif:


src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from __future__ import absolute_import
+
```
jpivarski, line 4: All the Python files start this way. (The "future import" is for Python 2.7 to have the same import semantics as Python 3.)

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
 import awkward as ak
 
+np = ak.nplike.NumpyMetadata.instance()
```
jpivarski, line 7: Although functions are split among `nplike` singletons, some basic types like `np.int8` and `np.dtype` are in a metadata singleton, which we call `np` everywhere (as opposed to the full "numpy", "cupy", or "nplike").

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
+
+
+class Index(object):
```
jpivarski, line 10: If classes aren't inheriting from anything else, they need to inherit from `object` to avoid "old-style classes" in Python 2.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
+    _dtype_to_form = {
+        np.dtype(np.int8): "i8",
+        np.dtype(np.uint8): "u8",
+        np.dtype(np.int32): "i32",
+        np.dtype(np.uint32): "u32",
+        np.dtype(np.int64): "i64",
+    }
```
jpivarski, line 17: The list of allowed dtypes and their conversion into forms are needed in two places, so I made a centralized list.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
 
-class Index:
     def __init__(self, data):
         self._nplike = ak.nplike.of(data)
+
         self._data = self._nplike.asarray(data, order="C")
+        if len(self._data.shape) != 1:
+            raise TypeError("an actual error string")
```
jpivarski: `data` must be one-dimensional.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
-    def __zeros__(length, nplike):
-        return nplike.zeros(length)
-
-    # the default for this is float64 so maybe type should also be mentioned
-    def __empty__(length, nplike):
-        return nplike.empty(length)
-
-    def __data__(self):
+        if self._T not in self._dtype_to_form:
+            raise TypeError("an actual error string")
```
jpivarski: Maybe I should have included "FIXMEs" in these strings, since we don't want to forget about them.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
+
+    @classmethod
+    def zeros(cls, length, nplike, dtype):
+        return nplike.zeros(length, dtype=dtype)
+
+    @classmethod
+    def empty(cls, length, nplike, dtype):
+        return nplike.empty(length, dtype=dtype)
```
jpivarski: Alternate constructors should be `@classmethods`, which take the class (`cls`) as the first argument (instead of `self` like a normal method or nothing like a `@staticmethod`).

You are right that a data type must be specified. I forgot about that!

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
+
+    @property
+    def data(self):
         return self._data
```
jpivarski: Data fields that were passed in as constructor arguments or are otherwise directly derived from them should be properties, which can be indicated with `@property`.

The double-underscores are only for the special Python methods (Python's protocols that we're adhering to).

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
+    @property
+    def nplike(self):
         return self._nplike
 
     def __len__(self):
-        return self._len
+        if self._data is None:
+            return 0
+        else:
+            return len(self._data)
```
jpivarski: In principle, this could have been in the constructor (as you had it) or generated on demand. When faced with a choice like that, prefer generating on demand. (The only exceptions would be if generating it is computationally intensive and should therefore be cached. That's rare.) The object's state (collection of attributes) is the most difficult thing to keep under control, so the more we can minimize its existence, the better.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
 
-    # array_str needs to be included in nplike
     def __repr__(self):
+        # FIXME
         return self._nplike.array_str(self._data)
```
jpivarski: Either [numpy.array_str](https://numpy.org/doc/stable/reference/generated/numpy.array_str.html) or [numpy.array2string](https://numpy.org/doc/stable/reference/generated/numpy.array2string.html). I'm not sure which would be better here. But there's also the rest of the XML to write, so I added a "FIXME."

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
 
-    def __form__(self):
-        type = str(self._T)
-        return type[0] + type.split("t")[1]
+    def form(self):
+        return self._dtype_to_form[self._data.dtype]
```
jpivarski: The constructor guarantees that this won't fail with a KeyError.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
 
     def __setitem__(self, index, value):
         self._data[index] = value
 
-    # the error that might occur from can be handled by the nplike, but if start stop are not in range it returns an empty array, is this behaviour constant with that it's expected?
-    def __getitem_range__(self, start, stop):
-        if start < stop and stop < self._len:
-            return self._data[start:stop]
-        else:
-            print("Illegal start:stop for this length")
```
jpivarski: `__getitem_range__` isn't a Python protocol the way that `__getitem__` is. (See [list of Python protocols](https://docs.python.org/3/reference/datamodel.html#special-method-names).)

Also, its functionality is covered by `__getitem__`, so an extra function is not needed (unlike C++).

Also, error messages are never print statements. This would be a ValueError (for consistency with the rest of Awkward Array) or an IndexError (for consistency with the rest of Python, which becomes possible now that we don't have to generate these in pybind11, which couldn't do IndexError for some reason).

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
-        type = str(self._T)
-        return type[0] + type.split("t")[1]
+    def form(self):
+        return self._dtype_to_form[self._data.dtype]
 
     def __getitem__(self, index):
         return self._data[index]
 
     def __setitem__(self, index, value):
         self._data[index] = value
```
jpivarski: After having written this, I realized that the convention throughout the rest of Awkward Array is to name the index "where" and the value for replacement "what", so

```python
    def __getitem__(self, where):
        return self._data[where]

    def __setitem__(self, where, what):
        self._data[where] = what
```

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
-    def __getitem_range__(self, start, stop):
-        if start < stop and stop < self._len:
-            return self._data[start:stop]
-        else:
-            print("Illegal start:stop for this length")
-
-    def __to64__(self):
-        self._data = self._data.astype("int64")
+    def to64(self):
+        return Index(self._data.astype(np.int64))
```
jpivarski: This should be returning an Index, not a raw array. (That's what the C++ does.)

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
 
-    def __iscontiguous__(self):
-        return self._data.strides[0] == self._data.itemsize
+    def iscontiguous(self):
+        if self._data is None:
+            return True
+        else:
+            return self._data.strides == (self._data.itemsize,)
```
jpivarski: Same here: should return an Index instance. Also handling the case of `self._data is None`. Also, note that comparisons with `None` use the Python `is` rather than `==`. (See [this if you're wondering](https://stackoverflow.com/q/26595/1623645).)

Also also, the expression

```python
self._data.strides == (self._data.itemsize,)
```

is more restrictive than the expression

```python
self._data.strides[0] == self._data.itemsize
```

since the latter allows `self._data.strides` to have more dimensions. We've constrained `self._data.shape` to one dimension in the constructor and presumably that restricts the `self._data.strides` as well, but we might as well be more careful.

src/awkward/_v2/index.py
```diff
@@ -1,79 +1,84 @@
...
-        np = ak.nplike.Numpy()
-        if str(self._nplike) == "Numpy":
-            self._data = self._nplike.asarray(self._data)
-            self._nplike = cp.ak.nplike.instance()
+        if self._data is None:
+            return Index(None)
+        else:
+            return Index(self._data.copy())
+
+    def __covert__(self, nplike):
```
jpivarski: I missed this one: it should be `convert_to`. The name `__convert__` is not a Python protocol.

tests/test_0884-index-and-identifier-refactoring.py
```diff
@@ -0,0 +1,12 @@
...
+from __future__ import absolute_import
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import awkward as ak  # noqa: F401
+
+
+def test():
+    # FIXME
+    pass
```
jpivarski: This is where you would put tests that compare the Python Index against the C++ Index8, IndexU8, Index32, IndexU32, Index64. If you have more than one test (very likely), the function names must be `test_XYZ` for some `XYZ`. (That's how pytest finds them.)

With this file in place,

```bash
python localbuild.py --pytest tests
```

would include this test and

```bash
python localbuild.py --pytest tests/test_0884-index-and-identifier-refactoring.py
```

would run only this test (faster).

jpivarski (commented):
I also forgot to add a `src/awkward/_v2/__init__.py` file, which would contain

```python

from __future__ import absolute_import
```

as universal boilerplate.

jpivarski:
Also worth knowing: if you run

```bash
pre-commit run -a
```

before pushing your commit, then you'll know if it will pass the pre-commit.ci test. That can be useful because pre-commit will change your code if it can (mostly to adhere to standard formatting rules). When pre-commit changes the files on GitHub, you have to `git pull` the changes locally, which can get messy if you've already started modifying them for the next step. (`git stash` before `git pull` is a good way to deal with that, if the changes aren't on the same lines.)

I think this is in the [CONTRIBUTING.md](https://github.com/scikit-hep/awkward-1.0/blob/main/CONTRIBUTING.md) document.

agoose77:
I can't comment the line number on mobile, but should `filedloc` be `fieldloc`? :)

jpivarski:
> I can't comment the line number on mobile, but should `filedloc` be `fieldloc`?

Yes. `:)`

src/awkward/_v2/identifier.py
```diff
@@ -0,0 +1,113 @@
...
+        self._fieldloc = fieldloc
+        if not isinstance(fieldloc, dict) or not all(
+            isinstance(k, int) and isinstance(v, str) for k, v in fieldloc.items()
+        ):
+            raise TypeError("Identifier fieldloc must be a dict of int -> str")
+        self._nplike = ak.nplike.of(data)
+
+        self._data = self._nplike.asarray(data, order="C")
+        if len(self._data.shape) != 2:
+            raise TypeError("Identifer data must be 2-dimensional")
```
agoose77, line 30:
```suggestion
            raise TypeError("Identifier data must be 2-dimensional")
```

jpivarski: Sorry! We missed this while we were talking (and not looking at updates here). I got your issue and PR, and have set it to auto-merge. Thanks!

PR #890, opened 2021-06-02
Feature: add `ak.ptp`

### LINKED ISSUE #889: Implement ptp
Is there a precedent for implementing functions that improve API parity with NumPy?

E.g. a naive `ptp` is something like two lines of code:
```python
@ak._connect._numpy.implements("ptp")
def ptp(arr, axis=None, keepdims=False, initial=None, mask_identity=True):
    """Return the range of values along a given axis."""
    return ak.max(
        arr, axis=axis, keepdims=keepdims, initial=initial, mask_identity=mask_identity
    ) - ak.min(
        arr, axis=axis, keepdims=keepdims, initial=initial, mask_identity=mask_identity
    )
```

But it's not an *essential* function. I'd vote in favour of implementing it (in https://github.com/scikit-hep/awkward-1.0/blob/main/src/awkward/operations/reducers.py) because `np.ptp` is more concise and quite widely used.

agoose77:
Fixes #889

### DIFF
--- a/src/awkward/operations/reducers.py
+++ b/src/awkward/operations/reducers.py
@@ -664,7 +664,7 @@ def ptp(arr, axis=None, keepdims=False, initial=0, mask_identity=True):
     because there are three lists, the first has a range of `3`, the second is
     empty, and the third has a range of `1`.
 
-    See #ak.ptp for a more complete description of nested list and missing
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Perfect, thanks!

PR #900, opened 2021-06-07
Bugfix: raise `ValueError` for incompatible union types in `ak.unzip` 

### LINKED ISSUE #898: `ak.unzip` does not fail for unions of heterogeneous records
Given two record arrays with different fields, e.g.

```python
>>> a = ak.Array([{"this": 100}])
>>> b = ak.Array([{"this": 90, "that": 100}])
```
the concatenated array has a union type:
```python3
>>> c = ak.concatenate((a, b))
>>> c.type
2 * union[{"this": int64}, {"this": int64, "that": int64}]
```

This union should fail to unzip, because the second array has field `that` whilst the first array does not. At present, this does not fail:
```python
>>> ak.unzip(c)
(<Array [100, 90] type='2 * union[int64, int64]'>,)
```

agoose77:
Fixes #898

### DIFF
--- a/tests/test_0898-unzip-heterogeneous-records.py
+++ b/tests/test_0898-unzip-heterogeneous-records.py
@@ -1,0 +1,16 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from __future__ import absolute_import
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import awkward as ak  # noqa: F401
+
+
+def test():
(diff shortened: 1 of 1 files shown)


agoose77:
I'm closing this for now, as I don't have time to work on it!

jpivarski:
Thanks! I appreciate having fewer open PRs, since I have to scan all of them when asking myself, "Is there anything wait for me to review?"

The issue is still there and the issue has a pointer to this closed PR, so whenever we take this up again in the future, the test you wrote will be something we can copy into the new PR for it.

PR #904, opened 2021-06-08
ArrayView expects contiguous NumpyArrays, so make sure they're contiguous.

### LINKED ISSUE #903: Numba retrieves wrong value for buffers with strides
This is a little tricky to reproduce late on a Friday. There is definitely a simpler example hiding in here somewhere
```python
@nb.njit
def get_11(record):
    return record[1].u


def test(order):
    # Create an array with either strided or contiguous columns
    table = np.zeros((2, 2), order=order)
    table[:] = np.arange(2  * 2).reshape(table.shape)

    record = ak.zip({"u": table[..., 0]})

    r1 = record[1].u
    r2 = get_11(record)

    assert r1 == r2, (r1, r2)


test(order="F")
test(order="C")
```

jpivarski:


agoose77:
@jpivarski I realise the complexity would increase, but wouldn't this be something that the user should enforce rather than the numba layer taking a potential contiguous copy? Unless ... do we already take a copy (of non-NumpyArray contents) in order to erase the indirection from non-structural layouts? (guesswork here)

jpivarski:
Copying NumpyArrays to make them contiguous is pretty common, perhaps more common that you think, and perhaps that's why it took so long to find this case (because most of the arrays that got this far _were_ contiguous). NumpyArray in C++ has a [contiguous](https://github.com/scikit-hep/awkward-1.0/blob/3f36053f0a2d40c6920aa27c65d9fb08e0677495/src/libawkward/array/NumpyArray.cpp#L3818-L3932) method to turn whatever we have into a contiguous array because nearly all operations except the ones that emulate NumPy either need that assumption are are greatly simplified by it.

It would be possible to iterate over non-contiguous arrays in Numba, but either the Numba type would have to carry the stride information (so that the iteration step size can be inserted into the compiled code as a compile-time constant) or the ArrayViewModel would have to be expanded to have this information at runtime. Carrying the stride data in the type would generate as-efficient compiled code, but it would mean that each array with a different stride would require a new compilation, and then some applications would end up spending a lot more time compiling. Carrying the stride data in the ArrayViewModel would slow down everything at runtime, not just the arrays that need non-trivial strides. The ArrayViewModel is like a "cursor" pointing to a part of the array tree; it's has the same memory layout for all types because it has to hold all information that hasn't been "burned into" the compilation. Most data in Numba are passed by value (to avoid a lot of memory management), meaning this cursor has to be copied in its entirety from one function to another; thus, we want it to be small. Including the strides in the runtime object would mean every cursor would be a bit bigger—they would need at least one more pointer, a pointer to all the stride values in the array-tree.

So it's not a clear-cut case of complexity vs optimization. Adding non-trivial strides in the Numba implementation would definitely increase the complexity, but it wouldn't always make it faster, and in quite a few cases would make it slower (including cases without non-trivial strides, in some possible implementations). But even if it were a clear-cut case of complexity vs optimization, I wouldn't want to implement it right away. In the case discussed here, 1D non-contiguous arrays are copied into contiguous arrays before entering Numba. It is also the case that nD NumpyArrays are copied into contiguous arrays before entering Numba, by virtue of turning n‒1 dimensions of the NumpyArray into RegularArrays, to avoid the extra complexity of dealing with multidimensional NumpyArray nodes in Numba. If anything, that case would have to be dealt with before the more-rare case of 1D non-contiguous arrays.

PR #907, opened 2021-06-10
Support for reading `pyarrow.lib.FixedSizeListType`

### LINKED ISSUE #906: Can't convert a pyarrow struct with a fixed-size list
Hi there,
Thanks for the library.

I'm trying to use `from_arrow` to convert a pyarrow struct containing fixed-size lists and it's currently failing. I used the following snippet to test:

```py
import pyarrow as pa
import awkward as ak
def test():
    struct = pa.struct([pa.field('x', pa.list_(pa.float64(), 2)), pa.field('y', pa.list_(pa.float64(), 2))])

    arr = pa.array([{'x': [1.0, 2.0], 'y': [3.0, 4.0]}], type=struct)
    assert ak.from_arrow(arr).tolist() == [{"x": [1.0, 2.0]}, {"y": [3.0, 4.0]}]
```

It fails at this assertion https://github.com/scikit-hep/awkward-1.0/blob/main/src/awkward/operations/convert.py#L2767 and I believe the reason is that pyarrow's type is `pyarrow.lib.FixedSizeListType` which is a subclass of `pyarrow.lib.DataType` thus matching it as a null data type.

agoose77:
Fixes #906

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -2655,6 +2655,18 @@ def _from_arrow(
             out = ak.layout.ListOffsetArray64(offsets, content)
             # No return yet!
 
+        elif isinstance(tpe, pyarrow.lib.FixedSizeListType):
+            assert tpe.num_buffers == 1
+            mask = buffers.pop(0)
+            list_size = tpe.list_size
+            offsets = ak.layout.Index32(numpy.array([0, list_size], dtype=np.int32))
+            content = popbuffers(array.values, tpe.value_type, buffers)
+            if not tpe.value_field.nullable:
+                content = content.content
+
+            out = ak.layout.ListOffsetArray32(offsets, content)
+            # No return yet!
+
         elif isinstance(tpe, pyarrow.lib.UnionType):
             if tpe.mode == "sparse":
                 assert tpe.num_buffers == 2
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -2658,12 +2658,11 @@ def _from_arrow(
         elif isinstance(tpe, pyarrow.lib.FixedSizeListType):
             assert tpe.num_buffers == 1
             mask = buffers.pop(0)
-            offsets = ak.layout.Index32(numpy.array([0, tpe.list_size], dtype=np.int32))
             content = popbuffers(array.values, tpe.value_type, buffers)
             if not tpe.value_field.nullable:
                 content = content.content
 
-            out = ak.layout.ListOffsetArray32(offsets, content)
+            out = ak.layout.RegularArray(content, tpe.list_size)
             # No return yet!
 
         elif isinstance(tpe, pyarrow.lib.UnionType):
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -2658,8 +2658,7 @@ def _from_arrow(
         elif isinstance(tpe, pyarrow.lib.FixedSizeListType):
             assert tpe.num_buffers == 1
             mask = buffers.pop(0)
-            list_size = tpe.list_size
-            offsets = ak.layout.Index32(numpy.array([0, list_size], dtype=np.int32))
+            offsets = ak.layout.Index32(numpy.array([0, tpe.list_size], dtype=np.int32))
             content = popbuffers(array.values, tpe.value_type, buffers)
             if not tpe.value_field.nullable:
                 content = content.content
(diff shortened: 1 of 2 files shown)


agoose77:
This will probably need better test coverage, but I'm not entirely familiar with pyarrow, so I welcome any inputs.

jpivarski:
Oh, so the implementation was there, but it was wrong? Simply untested? Then @VasilyFomin's example would serve as a test:

```python
pyarrow = pytest.importorskip("pyarrow")

tpe = pyarrow.struct([
    pyarrow.field('x', pyarrow.list_(pyarrow.float64(), 2)),
    pyarrow.field('y', pyarrow.list_(pyarrow.float64(), 2)),
])
arr = pyarrow.array([{'x': [1.0, 2.0], 'y': [3.0, 4.0]}], type=tpe)
assert ak.from_arrow(arr).tolist() == [{"x": [1.0, 2.0]}, {"y": [3.0, 4.0]}]
```

agoose77:
> Oh, so the implementation was there, but it was wrong? Simply untested? Then @VasilyFomin's example would serve as a test:
> 
> ```python
> pyarrow = pytest.importorskip("pyarrow")
> 
> tpe = pyarrow.struct([
>     pyarrow.field('x', pyarrow.list_(pyarrow.float64(), 2)),
>     pyarrow.field('y', pyarrow.list_(pyarrow.float64(), 2)),
> ])
> arr = pyarrow.array([{'x': [1.0, 2.0], 'y': [3.0, 4.0]}], type=tpe)
> assert ak.from_arrow(arr).tolist() == [{"x": [1.0, 2.0]}, {"y": [3.0, 4.0]}]
> ```

There was no switch case to handle the (new?) PyArrow type, or a test case, but the logic is very similar to the existing types. I added a test case that copies @VasilyFomin's example

jpivarski:
@agoose77 Thanks for all of this! I only want to add a test to ensure that the type is right (fixed-length, not `var`) and a `pytest.mark.skip` for old pyarrow versions, if there are any versions we support that don't have this feature (I need to check).

I've checked out your repo to make that one change, and then I'll do an auto-merge, but this would be much easier if you made pull requests off the main repo instead of your forked repo. (I wouldn't have to recompile it to test a Python change, for instance.) I just bumped up your permissions from "Write" to "Maintain" in case that was necessary. Thanks!

agoose77:
Thanks @jpivarski, I didn't consider that it would be easier for me to work on branches on the upstream, I'll do that from now on. There isn't much documentation/information about PyArrow, so this implementation is based upon my understanding + the test which passed first time. It's not a complex idea, a fixed length buffer though, so I am not too worried. Feel free to merge at will!

src/awkward/operations/convert.py
```diff
@@ -2658,12 +2658,11 @@ def popbuffers(array, tpe, buffers):
         elif isinstance(tpe, pyarrow.lib.FixedSizeListType):
             assert tpe.num_buffers == 1
             mask = buffers.pop(0)
-            offsets = ak.layout.Index32(numpy.array([0, tpe.list_size], dtype=np.int32))
             content = popbuffers(array.values, tpe.value_type, buffers)
             if not tpe.value_field.nullable:
                 content = content.content
 
-            out = ak.layout.ListOffsetArray32(offsets, content)
+            out = ak.layout.RegularArray(content, tpe.list_size)
```
jpivarski: The FixedSizeListType corresponds to an `ak.layout.RegularArray` for regular-length lists.

Also, these `offsets` only happened to be right for the length-1 example you tested: `tpe.list_size` is the length of all of the fixed-length lists; `offsets` of `np.array([0, N])` would mean the whole array consists of one list whose length is `N`. (That's why it worked when the test case had `len(out) == 1`.)

_If_ we wanted to turn the FixedSizeListType into an `ak.layout.ListOffsetArray64`, then the appropriate `offsets` would have been `np.arange(0, (len(array) + 1)*tpe.list_size, tpe.list_size)`.

jpivarski:
The FixedSizeListType has [existed since Arrow 0.16.0](https://github.com/apache/arrow/commit/1500d39aeccc5ed0137424fa806965103c50bc3a), and we require >= 2.0.0, so I don't need to put any test skips here.

src/awkward/operations/convert.py
agoose77: :facepalm: this is what I was worried about! I misunderstood this to be a fixed size 1D array. Thanks for the catch. Ah well, good practice.

Issue #910, opened 2021-06-11
`ak.unflatten` does not respect `IndexedArray` w.r.t `counts`


agoose77:
### Same-depth `IndexedArray`s
Given the following layout:
```python3
layout = ak.layout.IndexedArray64(
    ak.layout.Index64(np.array([3, 1, 0, 2])),
    ak.layout.ListOffsetArray64(
        ak.layout.Index64(np.array([0, 3, 6, 9, 12])),
        ak.layout.NumpyArray(np.array([0, 0, 0, 1, 1, 1, 2, 2, 3, 3, 3, 3])),
    ),
)
```
which has the list representations
```python
>>> display(
...     ak.to_list(layout),
...     ak.to_list(layout.content),
...     ak.to_list(layout.content.content),
... )
[[3, 3, 3], [1, 1, 1], [0, 0, 0], [2, 2, 3]]
[[0, 0, 0], [1, 1, 1], [2, 2, 3], [3, 3, 3]]
[0, 0, 0, 1, 1, 1, 2, 2, 3, 3, 3, 3]
```

the act of unflattening by the run lengths produces
```python
>>> ak.unflatten(
...     layout,
...     ak.flatten(ak.run_lengths(layout)),
...     axis=1
... ).tolist()
[[[3, 3], [3]], [[1, 1, 1]], [[0, 0, 0]], [[2, 2, 3]]]
```

If we apply `run_lengths` to the `ListOffsetArray` layout, then the result is "correct":
```python
>>> ak.unflatten(
...     layout,
...     ak.flatten(ak.run_lengths(layout.content)),
...     axis=1
... ).tolist()
[[[3, 3, 3]], [[1, 1, 1]], [[0, 0, 0]], [[2, 2], [3]]]
```
### Upper `IndexedArray`s
This is not just true for thee same-depth `index` layouts, but also,*any* layout at any depth above the current depth. Consider this layout:
```python3
layout = ak.layout.IndexedArray64(
    ak.layout.Index64([1, 0]),
    ak.layout.ListOffsetArray64(
        ak.layout.Index64([0, 2, 4]),
        ak.layout.ListOffsetArray64(
            ak.layout.Index64(([0, 3, 6, 9, 12])),
            ak.layout.NumpyArray(([0, 0, 0, 1, 1, 1, 2, 2, 3, 3, 3, 3])),
        ),
    ),
)
```
which has
```python
>>> display(
...     ak.to_list(layout),
...     ak.to_list(layout.content),
[...]

PR #912, opened 2021-06-11
Feature: add `ak.packed`

### LINKED ISSUE #746: We need an ak.packed function
Following the guide here https://awkward-array.org/how-to-convert-buffers.html it instructs to use `ak.to_buffers` in order to write HDF5 files. However, the output files can become unnecessary large very easily.
Please consider the following example
```python
import numpy as np
import awkward as ak
arr = ak.Array({"x": np.random.rand(1000)})
mask = [0, 2]
arr = arr[mask]
form, length, container = ak.to_buffers(arr)
```
`container`, which will get saved to the file, contains an array of 1000 numbers, even though we only want 2 of them. It doesn't have to be 1000, in fact this number can be much larger.
What I think would be very nice here is an option to have the container be restricted to only the data that is necessary. This could even be an additional function, condensing an awkward array so that it is compact in memory.
I know that flattening can have a similar effect, but it doesn't work on arrays with records. Surprisingly doing something like `ak.from_array(ak.to_arrow(arr))` has the desired effect on the array. However, this seems to be a very crude workaround.

agoose77:
This will simplify the length and order violating layouts such that external operations are compatible with per-layout transformations like `ak.unflatten`.

- [x] **EmptyArray:** unchanged
- [x] **NumpyArray:** converted to contiguous, if not already
- [x] **RegularArray:** truncate the `content` to `len(original) * size` with special handling (pass through) if `size == 0`
- [x] **ListArray:** convert `toListOffsetArray64(true)` (the `true` means starting at zero)
- [x] **ListOffsetArray:** convert `toListOffsetArray64(true)` (it's a pass-through if it's already true that `offsets[0] == 0`)
- [x] **RecordArray:** truncate all the `contents` to `len(original)`
- [x] **IndexedArray:** `project()` it
- [x] **ByteMaskedArray:** convert `toIndexedOptionArray` if the `content` does not have `PrimitiveType`. Doing so will naturally lead to the right kind of `index`. If not changing the type (because the `content` has `PrimitiveType`), at least truncate the `content` length to `len(original)`.
- [x] **IndexedOptionArray:** convert `toByteMaskedArray` if the `content` has `PrimitiveType`; otherwise, we want to project the `content` such that the non-negative `index` values become simple counting... the `index` should end up looking like `0, 1, 2, -1, 3, -1, -1, -1, 4, 5...`. That will take some thought.
- [x] **BitMaskedArray:** convert `toIndexedOptionArray` if the `content` does not have `PrimitiveType`. If not changing the type, at least truncate the `content` length to `len(original)`.
- [x] **UnmaskedArray:** unchanged (but recursively descend, of course)
- [x] **UnionArray:** simplify the `index` by `project`ing each of the `contents`
- [x] **VirtualArray:** materialize and recursively descend
- ~~**PartitionedArray** concatenate partitions?~~

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2088,7 +2088,7 @@ def unflatten(array, counts, axis=0, highlevel=True, behavior=None):
 
     else:
 
-        def getfunction(layout, depth, posaxis, apply):
+        def getfunction(layout, depth, posaxis):
             posaxis = layout.axis_wrap_if_negative(posaxis)
             if posaxis == depth and isinstance(layout, ak._util.listtypes):
                 # We are one *above* the level where we want to apply this.
@@ -2130,7 +2130,6 @@ def unflatten(array, counts, axis=0, highlevel=True, behavior=None):
                         getfunction,
                         pass_depth=True,
                         pass_user=True,
-                        pass_apply=True,
                         user=axis,
                     )
                 )
@@ -2141,7 +2140,6 @@ def unflatten(array, counts, axis=0, highlevel=True, behavior=None):
                 getfunction,
                 pass_depth=True,
                 pass_user=True,
-                pass_apply=True,
                 user=axis,
             )
 
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2205,24 +2205,7 @@ def packed(array, axis=None, highlevel=True):
                 ak.layout.ListArray64,
             ),
         ):
-            starts = nplike.asarray(layout.starts)
-            stops = nplike.asarray(layout.stops)
-            intervals = stops - starts
-            indices = nplike.repeat(
-                stops - nplike.cumsum(intervals), intervals
-            ) + nplike.arange(nplike.sum(intervals))
-            inner = ak.layout.IndexedArray64(ak.layout.Index64(indices), layout.content)
-
-            outer_offsets = nplike.asarray(layout.compact_offsets64(True))
-            outer_starts = outer_offsets[:-1]
-            outer_stops = outer_offsets[1:]
-            return ak.layout.ListArray64(
-                ak.layout.Index64(outer_starts),
-                ak.layout.Index64(outer_stops),
-                apply(inner, depth + 1, posaxis),
-                layout.identities,
-                layout.parameters,
-            )
+            return apply(layout.toListOffsetArray64(True), depth, posaxis)
 
         # ListOffsetArray performs resizing
         if isinstance(
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2157,28 +2157,25 @@ def unflatten(array, counts, axis=0, highlevel=True, behavior=None):
         return out
 
 
-def packed(array, axis=None, highlevel=True, behavior=None):
-    """
-    Args:
-        array: Array to simplify.
-        axis (int): The dimension at which this operation is applied. The
-            outermost dimension is `0`, followed by `1`, etc., and negative
-            values count backward from the innermost: `-1` is the innermost
-            dimension, `-2` is the next level up, etc.
-        highlevel (bool): If True, return an #ak.Array; otherwise, return
-            a low-level #ak.layout.Content subclass.
-        behavior (None or dict): Custom #ak.behavior for the output array, if
-            high-level.
-    """
+def _packed(array, axis=None, highlevel=True, behavior=None):
     layout = ak.operations.convert.to_layout(
         array, allow_record=False, allow_other=False
     )
     nplike = ak.nplike.of(layout)
 
     def truncate(layout, n):
-        return ak.layout.IndexedArray64(ak.layout.Index64(nplike.arange(n)), layout)
+        return ak.layout.IndexedArray64(
+            ak.layout.Index64(nplike.arange(n)), layout
+        ).project()
 
     def apply(layout, depth, posaxis):
+        if posaxis is not None:
+            # If a particular axis was given
+            posaxis = layout.axis_wrap_if_negative(posaxis)
+            # Do not proceed past that axis
+            if posaxis < depth - 1:
+                return layout
+
         if isinstance(layout, ak.layout.NumpyArray):
             return layout.contiguous()
 
@@ -2342,6 +2339,18 @@ def packed(array, axis=None, highlevel=True, behavior=None):
     if highlevel:
         return ak._util.wrap(out, ak._util.behaviorof(array, behavior=behavior))
     return out
+
+
+def packed(array, highlevel=True, behavior=None):
+    """
+    Args:
+        array: Array to simplify.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.layout.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+    """
+    return _packed(array, highlevel=highlevel, behavior=behavior)
 
 
 def local_index(array, axis=-1, highlevel=True, behavior=None):
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2190,12 +2190,14 @@ def packed(array, axis=None, highlevel=True):
                 return apply(layout.simplify(), depth, posaxis)
 
             index = nplike.asarray(layout.index)
-            n_options = nplike.sum(index < 0)
-            index[:n_options] = -1
-            index[n_options:] = nplike.arange(len(index) - n_options)
+            new_index = nplike.zeros_like(index)
+
+            is_none = index < 0
+            new_index[is_none] = -1
+            new_index[~is_none] = nplike.arange(len(new_index) - nplike.sum(is_none))
 
             return ak.layout.IndexedOptionArray64(
-                ak.layout.Index64(index), apply(layout.project(), depth, posaxis)
+                ak.layout.Index64(new_index), apply(layout.project(), depth, posaxis)
             )
 
         # Project indexed arrays
@@ -2247,7 +2249,9 @@ def packed(array, axis=None, highlevel=True):
             # Pack simplified layout
             tags = nplike.asarray(layout.tags)
             index = nplike.asarray(layout.index)
-            contents = [None] * len(layout.contents)
+
+            new_contents = [None] * len(layout.contents)
+            new_index = nplike.zeros_like(index)
 
             # Compact indices
             for i, content in enumerate(layout.contents):
@@ -2258,13 +2262,13 @@ def packed(array, axis=None, highlevel=True):
                 wrapped_content = ak.layout.IndexedArray64(
                     ak.layout.Index64(index[is_i]), content
                 )
-                contents[i] = apply(wrapped_content, depth, posaxis)
-                index[is_i] = nplike.arange(nplike.sum(is_i))
+                new_contents[i] = apply(wrapped_content, depth, posaxis)
+                new_index[is_i] = nplike.arange(nplike.sum(is_i))
 
             return ak.layout.UnionArray8_64(
                 ak.layout.Index8(tags),
-                ak.layout.Index64(index),
-                contents,
+                ak.layout.Index64(new_index),
+                new_contents,
                 layout.identities,
                 layout.parameters,
             )
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2180,11 +2180,11 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             return layout.contiguous()
 
         # EmptyArray is a no-op
-        if isinstance(layout, ak.layout.EmptyArray):
+        elif isinstance(layout, ak.layout.EmptyArray):
             return layout
 
         # Project indexed arrays
-        if isinstance(layout, ak._util.indexedoptiontypes):
+        elif isinstance(layout, ak._util.indexedoptiontypes):
             if isinstance(layout.content, ak._util.optiontypes):
                 return apply(layout.simplify(), depth, posaxis)
 
@@ -2200,11 +2200,11 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             )
 
         # Project indexed arrays
-        if isinstance(layout, ak._util.indexedtypes):
+        elif isinstance(layout, ak._util.indexedtypes):
             return apply(layout.project(), depth, posaxis)
 
         # ListArray performs both ordering and resizing
-        if isinstance(
+        elif isinstance(
             layout,
             (
                 ak.layout.ListArray32,
@@ -2215,7 +2215,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             return apply(layout.toListOffsetArray64(True), depth, posaxis)
 
         # ListOffsetArray performs resizing
-        if isinstance(
+        elif isinstance(
             layout,
             (
                 ak.layout.ListOffsetArray32,
@@ -2232,12 +2232,12 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             )
 
         # UnmaskedArray just wraps another array
-        if isinstance(layout, ak.layout.UnmaskedArray):
+        elif isinstance(layout, ak.layout.UnmaskedArray):
             return ak.layout.UnmaskedArray(apply(layout.content, depth, posaxis))
 
         # UnionArrays can be simplified
         # and their contents too
-        if isinstance(layout, ak._util.uniontypes):
+        elif isinstance(layout, ak._util.uniontypes):
             layout = layout.simplify()
 
             # If we managed to lose the drop type entirely
@@ -2267,7 +2267,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             )
 
         # RecordArray contents can be truncated
-        if isinstance(layout, ak.layout.RecordArray):
+        elif isinstance(layout, ak.layout.RecordArray):
             return ak.layout.RecordArray(
                 [
                     apply(truncate(c, len(layout)), depth, posaxis)
@@ -2280,7 +2280,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             )
 
         # RegularArrays can change length
-        if isinstance(layout, ak.layout.RegularArray):
+        elif isinstance(layout, ak.layout.RegularArray):
             if not len(layout):
                 return layout
 
@@ -2297,7 +2297,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             )
 
         # BitMaskedArrays can change length
-        if isinstance(layout, ak.layout.BitMaskedArray):
+        elif isinstance(layout, ak.layout.BitMaskedArray):
             layout = layout.simplify()
 
             if not isinstance(ak.type(layout.content), ak.types.PrimitiveType):
@@ -2314,7 +2314,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             )
 
         # ByteMaskedArrays can change length
-        if isinstance(layout, ak.layout.ByteMaskedArray):
+        elif isinstance(layout, ak.layout.ByteMaskedArray):
             layout = layout.simplify()
 
             if not isinstance(ak.type(layout.content), ak.types.PrimitiveType):
@@ -2328,7 +2328,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
                 layout.parameters,
             )
 
-        if isinstance(layout, ak.layout.VirtualArray):
+        elif isinstance(layout, ak.layout.VirtualArray):
             return apply(layout.array, depth, posaxis)
 
         # Finally, fall through to failure
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2332,7 +2332,12 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             return apply(layout.array, depth, posaxis)
 
         # Finally, fall through to failure
-        raise NotImplementedError
+        else:
+            raise AssertionError(
+                "unrecognized layout: "
+                + repr(layout)
+                + ak._util.exception_suffix(__file__)
+            )
 
     out = apply(layout, 1, axis)
 
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2227,7 +2227,6 @@ def packed(array, axis=None, highlevel=True, behavior=None):
             ),
         ):
             new_layout = layout.toListOffsetArray64(True)
-
             return ak.layout.ListOffsetArray64(
                 new_layout.offsets,
                 apply(new_layout.content, depth + 1, posaxis),
@@ -2294,7 +2293,7 @@ def packed(array, axis=None, highlevel=True, behavior=None):
             # multiple of the RegularArray size
             n, r = divmod(len(content), layout.size)
             if r:
-                content = truncate(content, r * layout.size)
+                content = truncate(content, n * layout.size)
 
             return ak.layout.RegularArray(
                 apply(content, depth + 1, posaxis), layout.size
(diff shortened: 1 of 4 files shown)


agoose77:
@jpivarski I'm going to be working on this but please feel free to make suggestions at any time!

agoose77:
I'm not sure how to handle `VirtualArray`. It could just pass through to avoid realising it, but equally, it would be confusing if you had to trigger a load in order for `simplify` to traverse into it. Perhaps using the same flag `materialize=True` as `recursive_walk` (but default to `True`) would be the right thing here?

jpivarski:
> Perhaps a flag `realise_virtual` (or something)

The word that we use for that is "materialize," though I think it would always be true for broadcasting or recursively applying. Operations should be expected to make a lazy array non-lazy. That's what this does:

https://github.com/scikit-hep/awkward-1.0/blob/94de4e5112ad3a2d5fc9c2ec0fc29e242543a0d6/src/awkward/_util.py#L1312-L1313

agoose77:
> > Perhaps a flag `realise_virtual` (or something)
> 
> The word that we use for that is "materialize," though I think it would always be true for broadcasting or recursively applying. Operations should be expected to make a lazy array non-lazy. That's what this does:
> 
> https://github.com/scikit-hep/awkward-1.0/blob/94de4e5112ad3a2d5fc9c2ec0fc29e242543a0d6/src/awkward/_util.py#L1312-L1313

Great. I sort of agree. I think maybe we don't add the flag for that reason, also because `recursively_apply` doesn't define the same arg.

jpivarski:
This `simplify` function looks very similar to one we've wanted for a while that would be called `ak.packed`: #746. I think you should make this into that: packing does simplify, but it also removes inaccessible elements.

In this comment, I'll write a list of rules that `ak.packed` should have. Since you'd be doing something on just about every node, it would probably make more sense to not try to shoehorn it into `ak._util.recursively_apply`. That function is for avoiding boilerplate when you only want to affect one node level; here, you'll be affecting them all.

   * **EmptyArray:** unchanged
   * **NumpyArray:** converted to contiguous, if not already
   * **RegularArray:** truncate the `content` to `len(original) * size` with special handling (pass through) if `size == 0`
   * **ListArray:** convert `toListOffsetArray64(true)` (the `true` means starting at zero)
   * **ListOffsetArray:** convert `toListOffsetArray64(true)` (it's a pass-through if it's already true that `offsets[0] == 0`)
   * **RecordArray:** truncate all the `contents` to `len(original)`
   * **IndexedArray:** `project()` it
   * **IndexedOptionArray:** convert `toByteMaskedArray` if the `content` has `PrimitiveType`; otherwise, we want to project the `content` such that the non-negative `index` values become simple counting... the `index` should end up looking like `0, 1, 2, -1, 3, -1, -1, -1, 4, 5...`. That will take some thought.
   * **ByteMaskedArray:** convert `toIndexedOptionArray` if the `content` does not have `PrimitiveType`. Doing so will naturally lead to the right kind of `index`. If not changing the type (because the `content` has `PrimitiveType`), at least truncate the `content` length to `len(original)`.
   * **BitMaskedArray:** convert `toIndexedOptionArray` if the `content` does not have `PrimitiveType`. If not changing the type, at least truncate the `content` length to `len(original)`.
   * **UnmaskedArray:** unchanged (but recursively descend, of course)
   * **UnionArray:** simplify the `index` by `project`ing each of the `contents`
   * **VirtualArray:** materialize and recursively descend

Before applying the above rules, option-type nodes (IndexedOptionArray, ByteMaskedArray, BitMaskedArray, UnmaskedArray) should be simplified and UnionArrays should be simplified. This prevents double-masking and unnecessary unions. In principle, we shouldn't be seeing these double-masked or unnecessary unions, but bugs happen and a function like this should be defensive against that.

All of the above rules are intended to make pickled and otherwise serialized data as small as they can be, without resorting to compression techniques (like run-length encoding, variable-length encoding) or packing bytes into bits (we'll leave them as bits if they're bits; otherwise no).

agoose77:
> This `simplify` function looks very similar to one we've wanted for a while that would be called `ak.packed`: #746. I think you should make this into that: packing does simplify, but it also removes inaccessible elements.

Sure, I had the same thoughts as I pushed the PR. Let's do it.

> In this comment, I'll write a list of rules that `ak.packed` should have. Since you'd be doing something on just about every node, it would probably make more sense to not try to shoehorn it into `ak._util.recursively_apply`. That function is for avoiding boilerplate when you only want to affect one node level; here, you'll be affecting them all.

Part of the logic for using `recursively_apply` is that I don't want to re-implement the type-aware tree logic that it already implements. What are you thoughts there?

jpivarski:
> Part of the logic for using `recursively_apply` is that I don't want to re-implement the type-aware tree logic that it already implements. What are you thoughts there?

The reasons one might want to do that are (a) to simplify/reduce boilerplate and/or (b) keep all "knowledge" of how to recurse in one place.

Some functions are made simpler with `recursively_apply` because they only do something specialized for one node type, all others "pass through" in a way that is very similar from one function to another. As long as we were talking about that (e.g. in `unflatten`), that's a good argument to use and possibly expand `recursively_apply`. But for this case, as you'll see when I finish the comment above, there will be something non-trivial to do for nearly every node type. Trying to shoehorn _this_ function into `recursively_apply` would only make it more complex, not less complex. It would be like trying to read some callback-heavy Javascript code. Sometimes, a straight-through, top-to-bottom Fortran style is more appropriate—whatever makes it easier to read, and that depends on context. So motivation (a) doesn't apply to this one.

For (b), that ship has sailed. The "knowledge" of what kinds of children each node type has is scattered throughout the codebase, mostly for the reason that centralizing it would make the code harder to read, and (you may have seen) some parts are already hard enough. Because of this lack of DRY centralization, we have lost the ability to easily add a new node type. That's something I accepted at the beginning because of how previous attempts went, and therefore put in an effort to get the set of node types right, knowing that it would be hard to add new ones in the future.

For examples of the ship sailing, look at convert.py.

jpivarski:
Ping (because GitHub doesn't email us about edited comments, only new comments). I've posted a list of rules in https://github.com/scikit-hep/awkward-1.0/pull/912#issuecomment-859636351.

agoose77:
> > Part of the logic for using `recursively_apply` is that I don't want to re-implement the type-aware tree logic that it already implements. What are you thoughts there?
> 
> The reasons one might want to do that are (a) to simplify/reduce boilerplate and/or (b) keep all "knowledge" of how to recurse in one place.

You're right. It's a combination of the two, but taking a step back and thinking about it - there is no saving on code here, it's just trampolining into the same function which implements most of the types. I'll move over to a recursion when I am back at it.

agoose77:
> Before applying the above rules, option-type nodes (IndexedOptionArray, ByteMaskedArray, BitMaskedArray, UnmaskedArray) should be simplified and UnionArrays should be simplified. This prevents double-masking and unnecessary unions. In principle, we shouldn't be seeing these double-masked or unnecessary unions, but bugs happen and a function like this should be defensive against that.

I'm not yet familiar with all of these types (though the docs make their usage clear). Are you suggesting that this should be a two-pass process, with the first pass operating upon option types only?

jpivarski:
> Are you suggesting that this should be a two-pass process, with the first pass operating upon option types only?

No, it can be a single pass, but when you encounter, say, an IndexedOptionArray, do

```python
elif isinstance(layout, ak._util.indexedoptiontypes):
    if isinstance(layout.content, ak._util.optiontypes):   # the bad case we want to protect against
        return recurse(layout.simplify())
    # apply the rules for IndexedOptionArray
```

This won't infinitely recurse because `layout.simplify()` will reduce the number of levels of option-types containing option-types by 1. Eventually, the option-type node won't contain an option-type node: it will eventually reach that state because it's possible to turn any option-type of option-type into just an option-type. But that also means that it's essential to guard this recursion with `isinstance(layout.content, ak._util.optiontypes)`.

For unions, it will be somewhat trickier because some unions can be simplified and others can't. There's a rule that I can help you with when you get to it: a UnionArray can be simplified when all of its `contents` are `mergeable` with each other. (Actually, you'd be able to work that out, I believe.)

Also note that we can `isinstance` groups of classes with similar properties using

https://github.com/scikit-hep/awkward-1.0/blob/94de4e5112ad3a2d5fc9c2ec0fc29e242543a0d6/src/awkward/_util.py#L66-L105

agoose77:
> For unions, it will be somewhat trickier because some unions can be simplified and others can't. There's a rule that I can help you with when you get to it: a UnionArray can be simplified when all of its `contents` are `mergeable` with each other. (Actually, you'd be able to work that out, I believe.)
> 
> Also note that we can `isinstance` groups of classes with similar properties using
> 
> https://github.com/scikit-hep/awkward-1.0/blob/94de4e5112ad3a2d5fc9c2ec0fc29e242543a0d6/src/awkward/_util.py#L66-L105

For unions, I think the behaviour that we want is:
1. to lift any immediate child unions into the parent union
1. to merge any mergable contents of the new top-level union
1. to compact the array contents and their associated indices. 
Does this match your expectations?

(1.) is already handled by `layout.simplify`, but doesn't behave as I would expect where the array types aren't compatible (e.g `float` vs `int`).  I wonder if this needs to be done manually instead.
(2.) this is also (I think) handled by `layout.simplify`, but doesn't behave well as discussed. Effectively, I think we want to find the partition of the set of contents with the smallest cardinality?

jpivarski:
> For unions, I think the behaviour that we want is:
> 
> 1. to lift any immediate child unions into the parent union
> 2. to merge any mergable contents of the new top-level union
> 3. to compact the array contents and their associated indices.
>    Does this match your expectations?

Yes, this is right. And UnionArray's `simplify` does this—the merging of ints and floats (and complex) is intentional: it's a choice to consider them subtypes of each other. (I can't conceive of a real-world advantage of a "union of ints and floats" over "just floats." There's a potential distinction in precision, but again, I can't conceive of a _real-world_ case where you'd want to mix them functionally without mixing them numerically.)

Just using UnionArray's `simplify` before going in and packing the contents is fine. What I thought would be tricky would be determining if `simplify` would change anything before calling it. If you always call `simplify` and recurse on the result of the simplification, it would be an infinite recursive loop. I was talking about how to write the guard.

The guard is easier for option-types, since option-type's `simplify` just ensures that two option-type nodes aren't double-stacked, nothing more. It's easy to write the guard for that.

agoose77:
Yes, I need to think about this a bit more. I'm not sure whether the
conversion is safe though. How do we handle converting uint64 to floats?
I'm afk right now but the thought occurred to me that the maxint is bigger
than the 2^53-1.

jpivarski:
> Yes, I need to think about this a bit more. I'm not sure whether the conversion is safe though. How do we handle converting uint64 to floats? I'm afk right now but the thought occurred to me that the maxint is bigger than the 2^53-1.

It's actually the same conversion as NumPy. (We let them decide the "right" way to merge numbers and then match it.)

agoose77:
Hmm. I can see that NumPy provide such a function, it just seems like a last-resort because of the lack of representation for large ints.

agoose77:
@jpivarski how would you feel about exposing `referentially_equal`, e.g. as `__eq__`? I think this would be useful, and would avoid us having to test for simplification. Looking at `simplify_uniontype`, it seems that all I need to test for is whether any contents are also union arrays, right? The merge ability is just a secondary optimisation for the contents of these nested union arrays, but from the top-level perspective, `simplify_uniontype` will *always* merge the nested union's contents with its own.

Furthermore, I don't think there a recursion risk here. If `simplify` returns a non UnionArray type, then we recurse on the result. Otherwise, we treat the simplification as a success (it doesn't matter if it actually did anything), and simplify the contents.

agoose77:
To implement the truncation step, I've added a private `truncate` helper that introduces a new `IndexedArray64` that wraps the existing content. Is this a bad idea? My thought process is that we then do not need to implement a new routine that truncates each type - Awkward already knows how to do this.

agoose77:
Why do we care whether the content has `PrimitiveType` in the masked types?

agoose77:
Now to push the tests...

agoose77:
I've made a mistake in the `IndexedOptionArray` case which I need to fix. I also need to make changes to stop mutating existing properties of layouts in-place.

agoose77:
OK, this is looking like it is nearly ready for a review. I have not yet implemented the axis / depth argument. I *don't* think we want to restrict this to only a single axis if `depth is not None`. Instead, I would probably want to define the upper bound (i.e. simplify to a given inclusive depth). The `depth = None` case can be used to simplify the entire structure. Do you think this is reasonable?

As you can imagine, the reason that I want to pass a depth parameter is that for routines like `unflatten`, we do not need any simplification below the axis of unflattening, so it would be unnecessary to invoke the copies etc that would be required to flatten the entire subtree.

agoose77:
Additionally, the tests are probably not quite finished - I don't thoroughly check that identities and parameters are preserved, or that the contents are always simplified too. Perhaps some input as to how much testing we want to do would be helpful.

agoose77:
When implementing the depth parameter (I settled on `to_axis` because it's up-to-and-including the axis), I observed that `UnionArray` raises an exception when asked to wrap `axis=-1` for the following layout
```python3
a = ak.layout.NumpyArray(np.arange(4))
b = ak.layout.RegularArray(ak.layout.NumpyArray(np.arange(12)), 3)
layout = ak.layout.UnionArray8_64(
    ak.layout.Index8([1, 1,  0, 0]), ak.layout.Index64([0,1, 0,1]), [a, b]
)
layout.axis_wrap_if_negative(-1)
```

I expect this to just return `-1`, because each array has its own concept of `axis=-1`. Do you have any thoughts here?

jpivarski:
I think `ak.packed` should always apply to all levels: no `axis` (depth) parameter. I can't imagine why anyone would want to partially pack an array: it's an "invisible" feature (like virtualness, partitioning, categoricals with IndexedArray, etc.). It would be hard to even know that it has been applied to one axis and not another—it's something one would want to apply to a whole array before `ak.to_buffers` or to reduce the indirection in an array and make it faster to access.

> * **PartitionedArray** concatenate partitions?

I think it would be more useful to leave the partitions as partitions. `ak.repartition` generalizes the process of concatenating them, so it would be good if `ak.packed` and `ak.repartition` can be used independently. Since `ak.packed` is important for storage, we don't want it to interfere with another storage-related property.

agoose77:
@jpivarski I was motivated to work on this by #910. In that case, we don't need to flatten every axis unless the user passes `axis=-1`, and indeed, one wouldn't want to flatten additional axes unless required to, because it introduces many copies of the data. For example, if I wanted to add structure to a wide RecordArray, `ak.unflatten` would then potentially require copying all of the branches (to project them).

jpivarski:
> `UnionArray` raises an exception when asked to wrap `axis=-1` for the following layout

Negative axes can't have a meaningful value in a union array with different depths. But this should be moot if you don't implement an `axis` argument.

agoose77:
> > `UnionArray` raises an exception when asked to wrap `axis=-1` for the following layout
> 
> Negative axes can't have a meaningful value in a union array with different depths. But this should be moot if you don't implement an `axis` argument.

I thought the `UnionArray` was supposed to return the negative axis value in such cases?

jpivarski:
#910 is about unflatten, an analysis function that takes an `axis` argument to know what dimension to turn into two dimensions. `ak.packed` is a whole different category of function: it produces an array that has the same high-level meaning as the input array with different low-level/internal structure. That's why I say that its changes are "invisible," like virtualness, partitioning, and categoricals. Whereas `ak.unflatten` needs an `axis` to make its high-level-visible change in the right place, an `axis` argument in `ak.packed` wouldn't be very useful. Which is good news because it's easier to not implement it!

agoose77:
@jpivarski we might be crossing wires here. I reported #910 because I noticed that there was a bug that entirely depended upon the (invisible) layout of the array, i.e. the `counts` argument used the layout from outside-in, whereas the `getfunction` only operates upon the internal content. This leads to incorrect structures for any arrays that have undergone some degree of manipulation.

The `ak.packed` function solves that problem by removing these problematic layouts, and ultimately moving them recursively down one level. As `ak.unflatten` needs something **like** `ak.packed`, it makes sense to me to implement an `axis` parameter (although we might make the user-default `axis=None`) so that `unflatten` can call `ak.packed` without changing the structure (or being slower) more than it needs to be.

jpivarski:
> I thought the `UnionArray` was supposed to return the negative axis value in such cases?

The definition is here:

https://github.com/scikit-hep/awkward-1.0/blob/98f4174041b1c9b72365da9ca8d371f8cdeb845f/src/libawkward/Content.cpp#L1686-L1711

It can only give a meaningful result if "min depth" is equal to "max depth." If they're not equal but it can keep recursing, then this function returns its input so that it can try again on the next recursive step. That's how it can work in a record array whose fields have different depths: the minimum depth and maximum depth would differ for all nodes above or at the RecordArray, but once a branch of the recursion enters one field, it gets a single depth value for that branch (and maybe some other recursive branch reaches another value). That allows `axis=-1` to mean a different deepest level for the `x` field of a record than the `y` field of a record, for example.

Union arrays, on the other hand, are a "sum type," not a "product type" (the type of an instance is `T1 or T2`, not `T1 and T2`), so everything in the union array has to count as the same depth or die trying. The exception gets raised by the `if (mindepth + axis == 0)`, which is the base case of this recursion if `mindepth == maxdepth` is never reached.

agoose77:
> Union arrays, on the other hand, are a "sum type," not a "product type" (the type of an instance is `T1 or T2`, not `T1 and T2`), so everything in the union array has to count as the same depth or die trying. The exception gets raised by the `if (mindepth + axis == 0)`, which is the base case of this recursion if `mindepth == maxdepth` is never reached.

Ah, I see. I was thinking about the product type rules being equivalent to the sum type. Thanks for clarifying that!

jpivarski:
> The `ak.packed` function solves that problem by removing these problematic layouts, and ultimately moving them recursively down one level. As `ak.unflatten` needs something **like** `ak.packed`, it makes sense to me to implement an `axis` parameter (although we might make the user-default `axis=None`) so that `unflatten` can call `ak.packed` without changing the structure (or being slower) more than it needs to be.

I see: you need it as a utility to solve the problem in `ak.unflatten`. In that case, the `axis` that it takes to solve the `ak.unflatten` problem doesn't need to be a part of the public API, which can simplify things by not having to handle negative `axis` values, for instance. If you define `_packed` as the general function that takes arguments you need for `ak.unflatten`, the user-visible `packed` function can expose only a subset of those arguments. Then the extra arguments in `_packed` don't need to follow the general patterns, like the ability to interpret negative `axis` (which can be hard).

This was done in convert.py; I think `_to_arrow` is distinguished from `to_arrow` so that `to_parquet` would have a backdoor to controlling the process more, in a way that we don't want to make available to users.

agoose77:
> I see: you need it as a utility to solve the problem in `ak.unflatten`. In that case, the `axis` that it takes to solve the `ak.unflatten` problem doesn't need to be a part of the public API, which can simplify things by not having to handle negative `axis` values, for instance. If you define `_packed` as the general function that takes arguments you need for `ak.unflatten`, the user-visible `packed` function can expose only a subset of those arguments. Then the extra arguments in `_packed` don't need to follow the general patterns, like the ability to interpret negative `axis` (which can be hard).
> 
> This was done in convert.py; I think `_to_arrow` is distinguished from `to_arrow` so that `to_parquet` would have a backdoor to controlling the process more, in a way that we don't want to make available to users.

Fab, I've made these changes, and I *think* it's ready for review. I modified the `truncate` helper to immediately project the `IndexedArray64` rather than recurse; otherwise, the `axis` parameter could prevent this from being performed.

agoose77:
Sorry for the git noise. I rebased to drop the first commit.

src/awkward/operations/structure.py
```diff
@@ -2157,6 +2157,202 @@ def getfunction(layout, depth, posaxis):
...
+        if isinstance(layout, ak.layout.RegularArray):
+            if not len(layout):
+                return layout
+
+            content = layout.content
+
+            # Truncate content if it is larger than a perfect
+            # multiple of the RegularArray size
+            n, r = divmod(len(content), layout.size)
+            if r:
```
jpivarski:
```python
if r != 0:
```

(I know that it's allowed in Python and black/flake8, but I want to avoid all weak typing, and Python's truthiness is the only part of Python that's weakly typed.)

src/awkward/operations/structure.py
```diff
@@ -2157,6 +2157,202 @@ def getfunction(layout, depth, posaxis):
...
+            # Do not proceed past that axis
+            if posaxis < depth - 1:
+                return layout
+
+        if isinstance(layout, ak.layout.NumpyArray):
+            return layout.contiguous()
+
+        # EmptyArray is a no-op
+        if isinstance(layout, ak.layout.EmptyArray):
+            return layout
```
jpivarski: All the `if` statements at this level should be an `elif` chain (i.e. avoid accidental fall-through if one handler is missing a `return` statement). It's just a more "functional" way of doing things, and a little safer. You'll notice that it's the same pattern in all the other places where we check all node types.

src/awkward/operations/structure.py
```diff
@@ -2157,6 +2157,202 @@ def getfunction(layout, depth, posaxis):
...
+                layout.valid_when,
+                layout.identities,
+                layout.parameters,
+            )
+
+        if isinstance(layout, ak.layout.VirtualArray):
+            return apply(layout.array, depth, posaxis)
+
+        # Finally, fall through to failure
+        raise NotImplementedError
```
jpivarski: Since the above should be an `elif` chain, this would be the `else` clause, and it should actually be an `AssertionError` (whose argument is the actual `layout` or `type(layout)`, for debugging) as in other such chains. Missing a case is considered a programming error: it's not an open-ended set—these are all the node types.

src/awkward/operations/structure.py
```diff
@@ -2157,6 +2157,202 @@ def getfunction(layout, depth, posaxis):
...
+def _packed(array, axis=None, highlevel=True, behavior=None):
+    layout = ak.operations.convert.to_layout(
+        array, allow_record=False, allow_other=False
+    )
+    nplike = ak.nplike.of(layout)
+
+    def truncate(layout, n):
+        return ak.layout.IndexedArray64(
+            ak.layout.Index64(nplike.arange(n)), layout
+        ).project()
```
jpivarski: `truncate` is used in quite a few places, and it would be a lot less expensive if it were implemented as

```python
def truncate(layout, n):
    return layout[:n]
```

That would be zero-copy. Should `ak.packed` copy everything it truncates?

   * If we're truncating a lot, then the original, much larger array could only be freed if we make a copy, and that would be an argument for copying on truncation. However, that's a rare case: they're usually shaving off a little bit, like the roundoff error in RegularArray.
   * In the primary use-case for this function, as a step right before serialization, the copy would be unnecessary because serialization is itself a copy (and it would only take the slice, not the underlying buffer).

So, generally, I think this function should not copy on truncation, returning `layout[:n]` is ideal. But if there's a use-case for copying on truncation, we could call `deep_copy` on all the `Index` attributes and `deep_copy` on `NumpyArray`. (That prevents it from recursing through the node structure, like `project`, but also it uses `memcpy` instead of creating an intermediate array of increasing numbers just to select all the elements in order.)

jpivarski (commented):
This is looking good! Just a few comments inline.

src/awkward/operations/structure.py
agoose77: Oh yikes! Good catch - I'd naively [assumed](https://xkcd.com/1339/) it would avoid copies for contiguous counts.

agoose77:
OK, since making the discussed changes, I added a missing pass-through case for `PartitionedArray`, and another for `Record`

jpivarski:
I just made `ak.packed` a preprocessing step for pickling, which resolves scikit-hep/awkward-0.x#246 and #701. Also, I think there's some alliterative possibilities there.

I added documentation (that I'm not 100% sure will render correctly; I'm just crossing my fingers and will look at awkward-array.org when it's done). I don't think you have any other changes to make, so I'm enabling auto-merge. If there's something else you want me to add, let me know right away and I can stop the process!

PR #914, opened 2021-06-11
C++ refactoring: Type and Form classes

jpivarski:
Types Checklist:

   - Type superclass
      - [x] `parameters` property
      - [x] `typestr` property
      - [x] private helper function(s) for "categorical" handling
   - UnknownType
      - [x] `__init__`
      - [x] `__str__`
      - [x] `__repr__`
   - NumpyType (renamed from PrimitiveType)
      - [x] `__init__`
      - [x] `primitive` property (renamed from `dtype`)
      - [x] `__str__` with special handling for datetime64/timedelta64, char, and byte
      - [x] `__repr__`
   - RegularType
      - [x] `__init__` (complain about negative size)
      - [x] `content` property (renamed from `type`)
      - [x] `size` property
      - [x] `__str__` with special handling for string and bytestring
      - [x] `__repr__`
   - ListType
      - [x] `__init__`
      - [x] `content` property (renamed from `type`)
      - [x] `__str__` with special handling for string and bytestring
      - [x] `__repr__`
   - RecordType
      - [x] `__init__`
      - [x] `contents` property (renamed from `types`)
      - [x] `recordlookup` property
      - [x] `__str__` with special handling for unnamed/named, `recordlookup is None`
      - [x] `__repr__`
   - OptionType
      - [x] `__init__`
      - [x] `content` property (renamed from `type`)
      - [x] `__str__` with special handling for nested RegularType/ListType
      - [x] `__repr__`
   - UnionType
      - [x] `__init__`
      - [x] `contents` property (renamed from `types`)
      - [x] `__str__`
      - [x] `__repr__`
   - ArrayType (not a subclass of Type)
      - [x] `__init__` (complain about negative length)
      - [x] `content` property (renamed from `type`)
      - [x] `length` property
      - [x] `__str__`
      - [x] `__repr__`

Forms checklist:

   - [x] `from_iter` function
   - [x] `from_json` just calls `from_iter(json.loads(data))`
   - [x] `numpyform.from_dtype` function
   - Form superclass
      - [x] `has_identities` property
      - [x] `parameters` property
      - [x] `form_key` property
      - [x] `__str__` method returns `json.dumps(self.tolist(verbose=False), indent="    ")`
      - [x] `tolist` method starts recursion into `_tolist_part` so that `toplevel=False` can be handled specially
      - [x] `to_list` is a synonym for `tolist` (historical)
      - [x] `to_json` just calls `json.dumps(self.tolist())`
   - EmptyForm
      - [x] `__init__`
      - [x] `__repr__`
      - [x] `_tolist_part`
   - NumpyForm
      - [x] `__init__`
      - [x] `primitive` property (replaces `itemsize`, `format`, and `dtype` from C++)
      - [x] `inner_shape`
      - [x] `__repr__`
      - [x] `_tolist_part` with special handling when `not verbose and not toplevel`
   - RegularForm
      - [x] `__init__`
      - [x] `content` property
      - [x] `size` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - ListForm
      - [x] `__init__`
      - [x] `starts` property
      - [x] `stops` property
      - [x] `content` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - ListOffsetForm
      - [x] `__init__`
      - [x] `offsets` property
      - [x] `content` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - RecordForm
      - [x] `__init__`
      - [x] `contents` property
      - [x] `recordlookup` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - IndexedForm
      - [x] `__init__`
      - [x] `index` property
      - [x] `content` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - IndexedOptionForm
      - [x] `__init__`
      - [x] `index` property
      - [x] `content` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - ByteMaskedForm
      - [x] `__init__`
      - [x] `mask` property
      - [x] `content` property
      - [x] `valid_when` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - BitMaskedForm
      - [x] `__init__`
      - [x] `mask` property
      - [x] `content` property
      - [x] `valid_when` property
      - [x] `lsb_order` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - UnmaskedForm
      - [x] `__init__`
      - [x] `content` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - UnionForm
      - [x] `__init__`
      - [x] `tags` property
      - [x] `index` property
      - [x] `contents` property
      - [x] `__repr__`
      - [x] `_tolist_part`
   - VirtualForm
      - [x] `__init__`
      - [x] `form` property
      - [x] `has_length` property
      - [x] `__repr__`
      - [x] `_tolist_part`


jpivarski:
@ioanaif I'm done setting up this PR. I think you won't have to write any tests at all, and the implementation will be considerably shorter than these tests.

Use the original C++ as a guide, with the translations described in the checklist. Those translations should be enforced by the tests. The dependency or is: UnknownType and NumpyType first, then the rest; EmptyForm, NumpyForm, and RegularForm first, then the rest.

If any test looks odd, like it breaks the pattern in a non-natural way, it might be a typo from me. Please ask me about any test you're unsure of!

jpivarski:
Windows can be excluded from tests that involve `float16`, `float128`, and `complex256`, since its NumPy doesn't have these types, and Python 2.7 and 3.5 can be excluded from tests that involve string representations of multiple parameters, since they don't have stable dict keys.

There are booleans for checking these things:

```python
>>> ak._util.win
False
>>> ak._util.py27
False
>>> ak._util.py35
False
```

src/awkward/_v2/forms/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+import awkward._v2.forms.bitmaskedform  # noqa: F401
+import awkward._v2.forms.unmaskedform  # noqa: F401
+import awkward._v2.forms.unionform  # noqa: F401
+import awkward._v2.forms.virtualform  # noqa: F401
+
+
+def from_iter(input):
+    has_identities = input["has_identities"] if "has_identities" in input else False
+    parameters = input["parameters"] if "parameters" in input else {}
+    form_key = input["form_key"] if "form_key" in input else None
```
jpivarski, line 25: These could be inline (below) because

```python
input.get("has_identities", False)
```

is the same thing as

```python
input["has_identities"] if "has_identities" in input else False
```

src/awkward/_v2/forms/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+            form_key=form_key,
+        )
+    if "VirtualArray" == input["class"]:
+        return awkward._v2.forms.virtualform.VirtualForm(
+            form=input["form"],
+            has_length=input["has_length"],
+            has_identities=has_identities,
+            parameters=parameters,
+            form_key=form_key,
+        )
```
jpivarski, line 151: If the above `if` statements are rearranged into an `if`-`elif`-`elif`-`else` chain, then the `else` case can raise a ValueError saying that `input["class"]` wasn't recognized.

src/awkward/_v2/forms/form.py
```diff
@@ -0,0 +1,63 @@
...
+            out["parameters"] = self._parameters
+            out["form_key"] = self._form_key
+        else:
+            if self._has_identities:
+                out["has_identities"] = self._has_identities
+            if len(self._parameters) > 0:
+                out["parameters"] = self._parameters
+            if self._form_key is not None:
+                out["form_key"] = self._form_key
+        return out
```
jpivarski: `toplevel` is not used in this function (and flake8 might be complaining about that).

The verbosity tests could have been implemented as

```python
            if verbose or self._has_identities:
                out["has_identities"] = self._has_identities
            if verbose or len(self._parameters) > 0:
                out["parameters"] = self._parameters
            if verbose or self._form_key is not None:
                out["form_key"] = self._form_key
```

with less duplication.

src/awkward/_v2/forms/numpyform.py
```diff
@@ -0,0 +1,147 @@
...
+if hasattr(np, "float16"):
+    _primitive_to_dtype["float16"] = np.dtype(np.float16)
+if hasattr(np, "float128"):
+    _primitive_to_dtype["float128"] = np.dtype(np.float128)
+if hasattr(np, "complex256"):
+    _primitive_to_dtype["complex256"] = np.dtype(np.complex256)
+
+_dtype_to_primitive = {}
+for primitive, dtype in _primitive_to_dtype.items():
+    _dtype_to_primitive[dtype] = primitive
```
jpivarski: This is also in types/numpytype.py; it should be centralized in one place (one or the other; maybe types/numpytype.py) so that the two don't have to be kept in sync.

src/awkward/_v2/forms/numpyform.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+def from_dtype(dtype, parameters={}, inner_shape=[]):
+    if "[" in str(dtype):
+        dtype = str(dtype).split("[")
+        # WIP
+        params = copy.deepcopy(parameters)
+        params["__unit__"] = dtype[1][:-1]
+        dtype = dtype[0]
```
jpivarski: I get that this is a work in progress, but the units are only going to exist in some `np.datetime64` and `np.timedelta64` (not all), so you can narrow to that for safety. (I don't know if other dtypes might use square brackets for other things.)

The defensive copy of `parameters` can be shallow, and dict has a handy `.copy()` method.

The `inner_shape` also comes from the dtype, or it can. (I don't remember whether I added examples of that to the tests.) That's if the dtype's `subdtype` is not `None`.

```python
>>> dt = np.dtype((np.int64, (1, 2, 3)))
>>> dt
dtype(('<i8', (1, 2, 3)))
>>> dt.shape
(1, 2, 3)
>>> dt.subdtype
(dtype('int64'), (1, 2, 3))
```

This function should probably check `_dtype_to_primitive` to make sure that the dtype is one of the allowed dtypes in the list (after breaking it down with `subdtype`, if need be). Here are some dtypes that we should not accept (with a good error message):

```python
>>> np.dtype("O")
dtype('O')
>>> np.dtype([("one", np.int64), ("two", np.float64)])
dtype([('one', '<i8'), ('two', '<f8')])
```

src/awkward/_v2/forms/numpyform.py
```diff
@@ -0,0 +1,147 @@
...
+def from_iter(input):
+    if isinstance(input, str):
+        return NumpyForm(primitive=input)
+    else:
+        primitive = input["primitive"]
+        inner_shape = input["inner_shape"] if "inner_shape" in input else []
+        has_identities = input["has_identities"] if "has_identities" in input else False
+        parameters = input["parameters"] if "parameters" in input else {}
+        form_key = input["form_key"] if "form_key" in input else None
+        return NumpyForm(primitive, inner_shape, has_identities, parameters, form_key)
```
jpivarski: This rule should be in the generic `from_iter`; maybe I used the wrong fully qualified path in the tests, but there should only be one `from_iter` function with all the rules.

jpivarski (commented):
Nice! Just minor comments.

src/awkward/_v2/forms/__init__.py
ioanaif: In the case of NumpyForm, the input could be str which will throw an AttributeError if the parameters are set with .get

jpivarski: The very first line in this function can check to see if it's in `_primitives_to_dtype` and take it on a different code path in that case. You could also have a check to see if it's a string that is not in `_primitives_to_dtype` (an error, but a different kind of error than not acting like a dict).

`if "has_identities" in input` could fail in a surprising way if `input` is a string like `"this is a strange has_identities string!"`

src/awkward/_v2/forms/numpyform.py
ioanaif: Unsupported dtypes are caught by __init__, I have added some tests to check if a TypeError is raised on unsupported types

jpivarski: This `from_dtype` function is not in the NumpyForm class; it's not necessarily called after any `__init__`.

PR #915, opened 2021-06-12
revert to datetime64 and timedelta64

ianna:
address https://github.com/scikit-hep/awkward-1.0/issues/913

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -233,8 +233,8 @@ def to_numpy(array, allow_missing=True):
         )
 
     elif (
-        str(ak.operations.describe.type(array)) == "datetime"
-        or str(ak.operations.describe.type(array)) == "timedelta"
+        str(ak.operations.describe.type(array)) == "datetime64"
+        or str(ak.operations.describe.type(array)) == "timedelta64"
     ):
         return array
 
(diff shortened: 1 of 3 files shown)


ianna:
@jpivarski - would you like me to revert the API as well? Thanks!

jpivarski (approved):
Great, thanks! I just wanted the "64" to be put back (and I didn't know how many places in the code to look for it) before this gets into frequent use and we _can't_ change it back.

jpivarski:
> @jpivarski - would you like me to revert the API as well? Thanks!

Actually, I don't know what you mean by this. I'd like it to be datetime64 and timedelta64 everywhere, as it is in NumPy.

ianna:
> > @jpivarski - would you like me to revert the API as well? Thanks!
> 
> Actually, I don't know what you mean by this. I'd like it to be datetime64 and timedelta64 everywhere, as it is in NumPy.

Shall I revert the function names such as:
https://github.com/scikit-hep/awkward-1.0/blob/98f4174041b1c9b72365da9ca8d371f8cdeb845f/src/libawkward/builder/ArrayBuilder.cpp#L102

jpivarski:
Actually, no, these should stay as they are because ArrayBuilder method names don't specify the number of bits in their types. (The names in this PR are to be consistent with `int8`, `float32`, etc., but none of the method names in ArrayBuilder look like that.)

PR #917, opened 2021-06-14
Feature: add axis parameter to `ak.fill_none`

agoose77:
## Change
This PR modifies the signature of `ak.fill_none`, by adding a new `axis` parameter. This parameter can either by an `int` or `None`. 

## Motivation
The existing behaviour of `ak.fill_none` is not easily predictable; the C++ layer ultimately operates upon the first layout(s) within the tree that are option types. Users sometimes encounter the need to remove *all* `None`s, which requires several calls to `ak.fill_none`. Framework authors might wish to ensure inputs are `None` free, this change now enables that more easily.

## Notes
Whilst the `type(axis) is int` cases are unambiguous as to their intended behaviour, the `axis=None` variant is up for debate.

Here, @jpivarski and I decided that `axis=None` should fill *all* `None`s. Because any choice of axis is now more restrictive than the existing behaviour, until the next major version the `fill_none` function will default to the old behaviour when no `axis` is set.


agoose77:
Sorry for the ping everyone - I was multitasking and managed to click everyone's names.

agoose77:
This breaks some existing tests, because this PR changes the behaviour of `ak.fill_none`. I believe it *should* break this behaviour because we *should* specify an axis. One possible addition might be to generalise to `axis=None`, which might be a safer default than `axis=1`.

agoose77:
@jpivarski how do you feel about the axis parameter? I think we *should* provide `axis=None`, and I would expect it to recursively `fill_none`. This will change the behaviour of the function. If we implement that behaviour, I'll probably want to merge the part of #912 that enables `getfunction` to re-enter.

jpivarski:
This is looking good!

As for the default, the most conservative thing to do is to make the default `axis` equivalent to the old behavior. People can start using explicit `axis` or not, but if they change nothing, nothing changes.

However, if the old default really is unexpected, we can change the behavior through the deprecation mechanism. This part of the README describes upcoming breaking changes: https://github.com/scikit-hep/awkward-1.0#roadmap We can have the new default `axis` take effect in version 1.4.0 on August 1, 2021 by advertising it here and putting a warning in the function:

```python
PLACEHOLDER = a singleton that isn't None

def fill_none(arguments, axis=PLACEHOLDER, more arguments):
    if axis is PLACEHOLDER:
        if ak.deprecations_as_errors:
            raise ValueError("ak.fill_none needs an explicit axis because the default will change in version 1.4.0")
        warnings.warn(
            """In version 1.4.0 (target date: August 1, 2021), the default axis for fill_none will be XXX
(Set ak.deprecations_as_errors = True to get a stack trace now.)""",
            FutureWarning,
        )
        axis=XXX
    normal code follows
```

This is similar to what `ak._util.deprecate` does, but this case is a little different because it's changing the default of an argument, not removing functionality.

https://github.com/scikit-hep/awkward-1.0/blob/98f4174041b1c9b72365da9ca8d371f8cdeb845f/src/awkward/_util.py#L49-L63

It's not obvious to me that the old default is surprising enough to require changing, but you're a better judge of that than I am. Anyway, if it needs to be changed, this is how we can do it, giving people a few months to prepare for it.

agoose77:
Thanks Jim.

> It's not obvious to me that the old default is surprising enough to require changing, but you're a better judge of that than I am. Anyway, if it needs to be changed, this is how we can do it, giving people a few months to prepare for it.
To be honest, I'm not *entirely* sure what the best option is. I feel slightly stronger in favour of deprecating the existing first-wins implementation because there are two possible scenarios:
1. The user has only a single axis of `None`s
2. The user has multiple axes of `None`s

In (1) the user can also provide an explicit axis. In (2) the user would need to repeatedly call `fill_none`. If we make `axis=None` recursive, then (2) becomes simpler.

However, I can see the benefit to the existing behaviour in the event that the user doesn't know where the `None`s are coming from. My gut feeling here is that you *should* know where None's are coming from; they're an important part of the schema. Moreover, if you notice that your data have `None`s, then you work out where they are coming from and fix them explicitly. I'll add the deprecation notice now; you can always nuke it in the code review :smile:

agoose77:
@jpivarski OK, I've taken a step back and realised that the current implementation won't facilitate an `axis` parameter, only a `start_depth` parameter; `IndexedArray` switches according to whether it is an option array cls or not, if it is, then it removes the option, otherwise it delegates to the content `fillna`. This is where the apparent first-wins behaviour comes from. 

This complicates things a little. I think that existing implementation in `main` is still undesirable long-term as it does not permit users to `fill_none` at a depth beyond an existing `None` e.g.
```python3
  ak.layout.Index64([0, 1, -1, 2]),
  ak.layout.ListOffsetArray64(
    ak.layout.Index64([0, 3, 6, 9, 12]),
      ak.layout.Index64([0, 1, 2, 3, 4, 5, -1, 7, 8, 9, 10, 11]),
      ak.layout.NumpyArray(np.arange(12))
    )
  )
)
```

So, what we want to do (I think) is circumvent  `fillna`'s recursion. I think the quickest solution here is to perform `fillna` only on the optiontypes, effectively, 
```python3
def getfunction(layout, depth, posaxis):
    # Check axis is correct
    ...

   # Invoke fillna
    if isinstance(layout, ak._utils.optiontypes):
        layout.fillna()
```

jpivarski (commented):
The old behavior was `axis=0`, which is a default some functions have (`ak.concatenate`, `ak.unflatten`, and `ak.is_none`, which is perhaps the most relevant for comparison with `ak.fill_none`).

I see some value in making the default `axis=None` since that's what people have sometimes expected (I remember some old issues about that), but I don't think the expectation is strong enough to warrant the pain of a deprecation cycle. If someone asks, "Why didn't it get the nested `Nones`?" we can just say, "Use `axis=None`" and that would probably be the end of the conversation. The documentation for this function should have examples with `axis=some int` and `axis=None` for users to hopefully notice.

So if you're asking me to make an executive decision, I'll do it: make the default `axis` of this function `0`, and then the old tests will work without modification.

Has the `axis=None` case not been implemented? I don't see it here or in the tests. Would that be hard to add (because `recursively_apply` has to be applied to the output of an `apply`)?

agoose77:
> The old behavior was `axis=0`, which is a default some functions have (`ak.concatenate`, `ak.unflatten`, and `ak.is_none`, which is perhaps the most relevant for comparison with `ak.fill_none`).
> 
> I see some value in making the default `axis=None` since that's what people have sometimes expected (I remember some old issues about that), but I don't think the expectation is strong enough to warrant the pain of a deprecation cycle. If someone asks, "Why didn't it get the nested `Nones`?" we can just say, "Use `axis=None`" and that would probably be the end of the conversation. The documentation for this function should have examples with `axis=some int` and `axis=None` for users to hopefully notice.
> 
> So if you're asking me to make an executive decision, I'll do it: make the default `axis` of this function `0`, and then the old tests will work without modification.
> 
> Has the `axis=None` case not been implemented? I don't see it here or in the tests. Would that be hard to add (because `recursively_apply` has to be applied to the output of an `apply`)?

Thanks for reading through that wall of text!

I'm really trying to say two things in one go:
1. The existing recursive behaviour cannot be restricted to single axes (e.g. `axis=i`)
2. The existing behaviour requires the user to know the underlying layout tree.
3. It would be nice to be able to remove *all* axes (e.g. `axis=None`)

At the moment, I don't think that the behaviour is `axis=0`, it's more like `axis > 0 && is_optional(axis)`. I propose that a better set of behaviours would be:

* Default `axis=SENTINEL` which removes the first optiontype it finds
* `axis=i` which operates only on that axis
* `axis=None` which removes *all* option types.

I'm not in favour of having a special sentinel value, so I don't know what the best API is here.

jpivarski:
If this would be simpler without using `fillna` at all (i.e. avoid the sentinel), then go for it. I think this function was written before `recursively_apply` existed.

The best API would be one that

   * takes `axis=#` to be the only list depth where None values are replaced with the replacement
   * takes `axis=None` to apply the replacement at all levels (using that new `pass_apply`)

I just tried a simple example and it _looked_ like the current behavior is `axis=0`. If it's not exactly that, then we could declare its deviation from `axis=0` to be a bug that we can change without a deprecation cycle. Whatever it was, if it wasn't equivalent to one of the two cases above, it wasn't intended. Then, since the old behavior was _closest to_ `axis=0`, we could make that the default and most code would continue to work as-is. (The code that stops working was depending on a bug; we can make an argument that _that_ code should be fixed.)

agoose77:
This PR (which just delays the call to `fillna` until the axis agrees with the depth) still demonstrates the first wins behaviour on this array:
```python
layout = ak.layout.RegularArray(
        ak.layout.Index64([0, 1, -1, 2, 3]),
        ak.layout.ListOffsetArray64(
            ak.layout.Index64([0, 3, 6, 9, 12]),
                ak.layout.Index64([0, 1, 2, 3, 4, 5, -1, 7, 8, 9, 10, 11]),
                ak.layout.NumpyArray(np.arange(12)),
            ),
        ),
    ),
    2,
)

```

```python
>>> ak.to_list(layout)
... [[[0, 1, 2], [3, 4, 5]], [None, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=0).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=1).tolist()
[[[0, 1, 2], [3, 4, 5]], [1, [None, 7, 8]]]
```

This behaviour can be prevented by checking that the type is an optiontype before calling `fillna` from Python. 

Of course, it is another question entirely as to whether we just remove fillna from the C++ layer entirely (or just ignore it for this PR) and re-implement it in Python. It sounds like that *might* align with the idea that we're lifting some of the C++ layer into Python now (apart from the kernels).

Lastly, longer-term I wonder how much of this multi-type aware code should be sitting as free functions vs member functions on types. This is a design decision I wouldn't feel comfortable tackling right now, but it is interesting!

agoose77:
OK, this PR now implements the described behaviour, with a default of `None`. The example above now behaves as follows:
```python
>>> ak.to_list(layout)
... [[[0, 1, 2], [3, 4, 5]], [None, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=0).tolist()
... [[[0, 1, 2], [3, 4, 5]], [None, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=1).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=None).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [1, 7, 8]]]
>>> ak.fill_none(layout, 1).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [1, 7, 8]]]
```


 I made this decision because after modifying the behaviour to *only* affect the specified axis, it was not clear whether `axis=0` would give the right result anymore. I had to change an existing use of `fill_none` to reflect the new default.

It might be that this needs to be in the next major release (if using SemVer) instead of minor, but I'll defer to you on that one :smile:

agoose77:
You can see from my above commits that I wavered between `axis=0` and `axis=None` defaults. In the end, I think we *do* need a deprecation notice (and have implemented one) - the new behaviour is more explicit, so it will miss existing cases. This also means that the old tests don't fail, so we can leave them in until the deprecation period expires!

New behaviour:
```python
>>> ak.to_list(layout)
... [[[0, 1, 2], [3, 4, 5]], [None, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=0).tolist()
... [[[0, 1, 2], [3, 4, 5]], [None, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=1).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [None, 7, 8]]]
>>> ak.fill_none(layout, 1, axis=None).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [1, 7, 8]]]
>>> ak.fill_none(layout, 1).tolist()
... [[[0, 1, 2], [3, 4, 5]], [1, [None, 7, 8]]]
```

agoose77:
Hopefully this is the *last* commit, which cleans things up by moving the deprecated implementation into a new private function.  Additionally, I removed the `PartitionedArray` case in this PR because in the new explicit axis code it's no longer required; it's handled by the `getfunction` logic automatically.


Thanks for **all** your help on this @jpivarski, I know it's been a bit of an intense PR.

jpivarski:
I was about to disagree with the introduction of a deprecation/breaking change, but now I see why. I had thought that the effective axis (in the main branch) is `axis=0`:

```python
>>> ak.fill_none(ak.Array([[None, 2], None, [4, None]]), 10)
<Array [[None, 2], 10, [4, None]] type='3 * union[var * ?int64, int64]'>
```

but if this were one level deeper, it still applies at the first level that has option-type.

```python
>>> ak.fill_none(ak.Array([[[None, 2], None, [4, None]]]), 10)
<Array [[[None, 2], 10, [4, None]]] type='1 * var * union[var * ?int64, int64]'>
```

That's not any particular `axis`, so you're right that some users of the old system would be broken by making the new default `axis=0` and others would not. There is no value we can make the new default `axis` that will not break _somebody's_ code. But the old way was wrong (before the general pattern of how these functions should work had been fully set), it needs to break.

We have a choice. The new default could be `axis=0` (like `ak.is_none`) or `axis=None` (I think this might be the usual expectation). What do you think? We'll be stuck with what we choose now. Do you want to open a discussion under Deprecations?

https://github.com/scikit-hep/awkward-1.0/discussions/categories/deprecations

agoose77:
The failing tests are caused by the new warning that's raised by this PR, with `deprecations_as_errors`. Should I modify the old tests that use `fill_none` to permit warnings?

jpivarski:
Since the behavior has changed, it is necessary to change the old tests. (Even though the tests are largely historical and labeled by PR or issue number, it's not uncommon to change them. Sometimes it's because the old tests were testing too much—more than we guaranteed, such as the exact string representation of a repr. But in a case like this, you _know_ you're changing the behavior, so it's a good sign that the old tests had enough coverage to fail!)

I just did a grep, and there are 10 cases of `ak.fill_none`, all in tests/test_0072-fillna-operation.py, tests/test_0590-allow-regulararray-size-zero.py, and tests/test_0634-fill_none-with-record.py. Probably setting `axis=None` would be sufficient for these; they probably didn't test multiple levels of missing values.

jpivarski:
Please don't remove `axis=None` as an option, considering that it works and produces valid arrays! Maybe it's less useful than other cases (because it makes union arrays), but it has positive usefulness (union arrays are good for some things).

agoose77:
> Please don't remove `axis=None` as an option, considering that it works and produces valid arrays! Maybe it's less useful than other cases (because it makes union arrays), but it has positive usefulness (union arrays are good for some things).

OK, I concede that I might have been a little *too* opinionated here. I'll revert the change :)

agoose77:
From #929:
> Our deprecation function and/or the custom one for ak.fill_none's axis argument should recommend this line of code instead of leaving it to the user to come up with that.
> Also, the warning for ak.fill_none's axis should recommend the new default, since that's the best way to silence the warning when they see it.

Yes, I have now removed the custom `deprecated` for `fill_none`'s old behaviour in favour of the existing decorator. I'll modify this in turn if we merge #929 first!

I will add a note about the new default now!

jpivarski:
@agoose77 Is this PR done? I'm trying to figure out what is waiting for me and what can be closed. The last message (27 days ago) references #929, which was merged.

I'll start by bring this branch up to date and seeing if the tests still work.

agoose77:
I *think* it's done. We have left it with axis=-1 as a future default, and
for now it incurs a deprecation warning when no axis is given, falling back
to axis=None

jpivarski:
It [relies on the API](https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=7108&view=logs&j=bb1c2637-64c6-57bd-9ea6-93823b2df951&t=8994636e-e094-5103-be26-64e72a4b5d0c&l=1856) of "recursively_apply", but that changed in another PR.

agoose77:
Good catch; I'd forgotten that we removed that flag. I'll fix this up tomorrow BST.

agoose77:
I've fixed the removal of `pass_apply` and am directly using the new transformer helper. I have a couple of questions about code-style — I don't normally use exhaustive `if` statements (which I recall you noting in a previous code review), so let me know if there's still some style elements that you'd like to see changed here.

agoose77:
@jpivarski thanks for the cleanups (in particular `_fill_none_deprecated`), I must have missed that when pushing.

jpivarski:
I just turned on warnings-as-errors (though it took a bit to figure out how to do that in pytest, now that we're doing it the standard way; notes to in the warning message have been updated).

See the [changes to Roadmap](https://github.com/scikit-hep/awkward-1.0/blob/c7f6e8227ceff6ca1a1c426c63e95ca25245ec9f/README.md#roadmap): this is how the deprecation is going to look. I think you were considering another backward-incompatible change; it would have to go here, too. The fact that all backward-incompatible changes have to be listed here puts a damper on how many there can be.

I've changed the once-every-two-month releases into monthly releases, and I don't plan to do many in-between releases anymore. Regular users are going to see the deprecation warning for the first time after August 1, and people who don't update more frequently than every two months (there are quite a few of them) won't see it at all—the new behavior could take them by surprise. However, there has to be some threshold frequency. (This is basically saying that people who don't update more frequently than every two months should expect surprising changes.)

PR #918, opened 2021-06-14
Feature: add `pass_apply` argument to util

agoose77:
This is a very simple PR to add `pass_apply` to `recursively_apply`. This permits clients of this function to re-enter, without re-implementing the same logic. This is not useful for #912 but is useful for #917


jpivarski (approved):
This is very simple and won't break any existing code. I'll enable auto-merge so that you can start using it as soon as the tests pass.

Issue #886, opened 2021-06-01
Use `pyarrow.dataset.parquet_dataset` in `ak.from_parquet`


martindurant:
Unfortunately, `pyarrow.parquet.ParquetFile` does not support remote filesystems.
You could use a construction like
```
with fsspec.open(url, 'rb', **storage_options) as f:
    q = pq.ParquetFile(f)
```
for the _metadata case or
```
pq.ParquetDataset(paths, filesystem=fs, **kwargs)
```
for the list-of-files case; this is what dask "legacy" does.

However, it seems that arrow will be using the new API only,  so
```
q = pyarrow.dataset.parquet_dataset("s3://pivarski-princeton/millionsongs/_metadata", filesystem=fs)  # note the full path to metadata
```
where the returned object has a `.schema` and fragments `list(q.get_fragments())` instead of the former pieces.

agoose77:
I've taken a look at this, and it's a bit more fiddly than I first anticipated. I'm keeping some notes here, in case anyone gets a chance to work on it before I do.

- The new `parquet_dataset` function ignores the order of the row groups ([due to `unordered_map`](https://github.com/apache/arrow/blob/133b1a904bf7fc1d24343c306a2279e27d4ebe6d/cpp/src/arrow/dataset/file_parquet.cc#L891-L915))
   - Need to use `pyarrow.parquet.ParquetDataset` *or* compose dataset manually by reading metadata first. I think the latter option is most future-proof
- Handling URIs *and* filepaths will be a logistical challenge. I propose instead to only handle paths, and allow user to specify the filesystem.
- fsspec expects POSIX paths, Arrow file systems convert to POSIX, and  currently we regularise paths according to the host OS. We need to be clear about when we have fs paths vs host paths.
- `dataset` is a very general API, and if we're not careful we'll have lots of tricky cases to handle. 

Awkward already does some boilerplate like finding the `_metadata` file for datasets, and in order to maintain the same interface we're going to need to do the same thing. 

So I think the way we need to tackle this is:
- Switch on `is_file_like(source)` to use new or old dataset
- Define default filesystem (`arrow.fs.LocalFileSystem`)
   - not sure if this should be parameter or in `options`
- Define four path-like cases for new API:
   -  `str` (single file)
   - `str` (directory)
   - `str` (directory with `_metadata`)
   - `Iterable[str]`
   - other (pass through to `arrow.dataset`)
- convert paths to POSIX (to avoid platform-specific pain)

[...]

PR #923, opened 2021-06-15
Bugfix: check for `file`-like objects in `from_parquet`

### LINKED ISSUE #921: `from_parquet` does not work with file-like objects
For example

```pycon
>>> import awkward as ak
>>> array = ak.Array([1, 2, 3])
>>> ak.to_parquet(array, "array.parquet")
>>> f = open("array.parquet", "rb")
>>> array2 = ak.from_parquet(f)
  File "/home/nikolai/.local/lib/python3.9/site-packages/awkward/operations/convert.py", line 3566, in from_parquet
    single_file = pyarrow.parquet.ParquetFile(filename)
  File "/home/nikolai/.local/lib/python3.9/site-packages/pyarrow/parquet.py", line 217, in __init__
    self.reader.open(source, use_memory_map=memory_map,
  File "pyarrow/_parquet.pyx", line 947, in pyarrow._parquet.ParquetReader.open
  File "pyarrow/io.pxi", line 1473, in pyarrow.lib.get_reader
  File "pyarrow/io.pxi", line 1466, in pyarrow.lib.get_native_file
TypeError: Cannot convert bytes to pyarrow.lib.NativeFile
```

presumably because file-like objects are instances of `Iterable`, so it tries to loop over them here:

https://github.com/scikit-hep/awkward-1.0/blob/0fe77918e1fa828e539ff287153150d028decfda/src/awkward/operations/convert.py#L3616-L3617

```python
>>> import awkward as ak
>>> ak.__version__
'1.3.0'
```

agoose77:
This doesn't fully address the larger changes that we probably need to make w.r.t #886, but does fix #921.

Perhaps this is a good stop-gap fix until we address the larger changes.

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -3613,7 +3613,11 @@ def from_parquet(
             relative_to = source
             source = sorted(glob.glob(source + "/**/*.parquet", recursive=True))
 
-    if not isinstance(source, str) and isinstance(source, Iterable):
+    if (
+        not isinstance(source, str)
+        and isinstance(source, Iterable)
+        and not hasattr(source, "read")
+    ):
         source = [_regularize_path(x) for x in source]
         if relative_to is None:
             relative_to = os.path.commonpath(source)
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This fixes an important oversight. (I didn't realize that file-like objects could be iterable.) Thanks!

If you're all done (thumbs up/checkmark), I'll merge it.

PR #926, opened 2021-06-15
start on sorting documentation

ianna:
@jpivarski - I'd like to add the following image to the description. Shall I add it to 
`docs-img/diagrams/axis.jpg`?

```
% ls -al docs-img/diagrams/axis.jpg
-rw-r--r--@ 1 yana  staff  31328 Jun 15 17:35 docs-img/diagrams/axis.jpg
```
axis


jpivarski:
If you like that diagram, I'd quickly trace over it in SVG before adding it. (It would take less than 5 minutes, but SVG and its derivatives, PDF and PNG, render better.)

The diagram is hard to interpret in the context of variable length dimensions, though. Maybe it could be modified to be more like this:



but with less information?

jpivarski:
Oh, and I think we need to avoid the word "column" for an array axis. "Column" is often used for record fields, which is a very different thing from an axis at any level.

For example, all axes of an array are converted into a Pandas index (its "rows"), but it's the record fields that become Pandas columns:

```python
>>> ak.to_pandas(
...     ak.Array([
...         [{"x": 1, "y": 1.1}, {"x": 2, "y": 2.2}, {"x": 3, "y": 3.3}],
...         [],
...         [{"x": 4, "y": 4.4}, {"x": 5, "y": 5.5}]
...     ])
... )
                x    y
entry subentry        
0     0         1  1.1
      1         2  2.2
      2         3  3.3
2     0         4  4.4
      1         5  5.5
```

ianna:
> If you like that diagram, I'd quickly trace over it in SVG before adding it. (It would take less than 5 minutes, but SVG and its derivatives, PDF and PNG, render better.)
> 
> The diagram is hard to interpret in the context of variable length dimensions, though. Maybe it could be modified to be more like this:
> 
> 
> 
> but with less information?

Yes, I'll update the snapshot. BTW, the axis are off by 1 :-)
```python
>>> array = ak.Array([[2, 3, 5], [], [None, 7], [11]])
>>> array
<Array [[2, 3, 5], [], [None, 7], [11]] type='4 * var * ?int64'>
>>> ak.prod(array)
2310
>>> ak.prod(array, axis=0)
<Array [22, 21, 5] type='3 * int64'>
>>> ak.prod(array, axis=1)
<Array [30, 1, 7, 11] type='4 * int64'>
>>> ak.prod(array, axis=2)
  File "/Users/yana/Projects/PR925/awkward-1.0/awkward/operations/reducers.py", line 395, in prod
    layout.prod(axis=axis, mask=mask_identity, keepdims=keepdims), behavior
ValueError: axis=2 exceeds the depth of the nested list structure (which is 2)
```

ianna:
Screenshot 2021-06-16 at 11 48 15
is this one better?

jpivarski:
Yes, that new diagram is much clearer, and thanks for the correction to the old one. I'll go fix it now (6a1c16f3378deeda234890c0c20c0ff89b1a70f1).

docs-src/how-to-restructure-sort.md
```diff
@@ -14,10 +14,73 @@ kernelspec:
...
+array = ak.Array([[None, None, 3, 2], [1]])
+ak.sort(array)
+```
+
+`None` values are dropped when sorting a one dimentinal option type array:
+
+```{code-cell} ipython3
+import awkward as ak
+array = ak.Array([None, None, 1, -1, 30])
+ak.sort(array)
```
ianna, line 71: @jpivarski - I wonder if dropping the `None`s is a bug:
```python
>>> import awkward as ak
>>> array = ak.Array([None, None, 1, -1, 30])
>>> array
<Array [None, None, 1, -1, 30] type='5 * ?int64'>
>>> ak.sort(array)
<Array [-1, 1, 30] type='3 * int64'>
```
because they are preserved in a higher dimensional array:
```python
>>> array = ak.Array([[None, None, 3, 2], [1]])
>>> ak.sort(array)
<Array [[2, 3, None, None], [1]] type='2 * var * ?int64'>

```

ianna (commented):
NumPy doesn't sort arrays with `NoneType`, based on `nan`:
```python
>>> narr =np.array([np.nan, np.nan, 1, -1, 30])
>>> np.sort(narr)
array([-1.,  1., 30., nan, nan])
```

docs-src/how-to-restructure-sort.md
jpivarski: NaN is not the same thing as None (which Pandas is finding, and that's why they're trying to deprecate all the methods that looked for NaN in favor of option-types). However, in

```python
>>> narr =np.array([np.nan, np.nan, 1, -1, 30])
>>> np.sort(narr)
array([-1.,  1., 30., nan, nan])
```

it looks like it's being sorted the same way. If NaN is insensitive to the choice of `ascending=True` vs `ascending=False`, that's to be expected because all comparisons with NaN result in false:

```python
>>> a = np.nan
>>> b = np.nan
```

would return false from `a < b` and also from `b < a`. So if ascending puts Nones at one end of the list and descending puts Nones at the other, but the same isn't true for NaN, I'd say that's okay. NaN is not None.

For the other case,

```python
>>> import awkward as ak
>>> array = ak.Array([None, None, 1, -1, 30])
>>> array
<Array [None, None, 1, -1, 30] type='5 * ?int64'>
>>> ak.sort(array)
<Array [-1, 1, 30] type='3 * int64'>
```

yes, that is a bug. It's probably a case of the implementation assuming that you're not at top-level, but in this case, you're at top-level. Perhaps it could be resolved by adding a case to check for top-level and do the right thing, or it could follow some of the other methods in wrapping the input array with a length-1 RegularArray, adjusting the `axis` (positive or negative), applying the operation, and then unwrapping the outer RegularArray when done. Wrapping/unwrapping a RegularArray is an _O(1)_ operation, and then the inner logic only ever has to deal with the "I'm nested inside of another node" case.

jpivarski:
I'm syncing this with `main` so that you get "modernized" JupyterBook pages to fill.

For writing these pages, note that you can start up a `jupyter lab` and right-click on the `.md` file you want to edit to "open as notebook." That lets you write the Jupytext (`.md`) as though it were a notebook, but the only things that gets saved to the file are the input cells (input code and markdown). The outputs are re-evaluated in testing and in production, which acts as a check to ensure they're up to date. (Also, we like Jupytext sources because git diffs are readable, unlike ipynb files.)

For local testing, you can do `jupyter-book build docs-src`.

ianna:
@jpivarski - given that this `array` is the same type as `another_array` -- `9 * {"x": float64, "y": var * int64}`:
```python
>>> array = ak.Array(
...         [
...             {"x": 0.0, "y": []},
...             {"x": 11.11, "y": [1]},
...             {"x": 2.2, "y": [0, 2]},
...             {"x": 33.33, "y": [3, -1, 10]},
...             {"x": 4.4, "y": [-2, 0, 4, 14]},
...             {"x": 5.5, "y": [25, 15, 5]},
...             {"x": 6.6, "y": [26, 6]},
...             {"x": 77.77, "y": [77]},
...             {"x": 8.8, "y": []},
...         ]
...     )
>>>
```

```python
another_array = ak.Array({"x": [0.0, 11.11, 2.2, 33.33, 4.4, 5.5, 6.6, 77.77, 8.8], "y": [[], [1], [0, 2], [3, -1, 10], [-2, 0, 4, 14], [25, 15, 5], [26, 6], [77], []]})
```
shouldn't sorted array be also the same type `9 * {"x": float64, "y": var * int64}`, such as?
```python
assert ak.to_list(ak.sort(array)) == [
        {'x': 0.0, 'y': []},
        {'x': 2.2, 'y': [1]},
        {'x': 4.4, 'y': [0, 2]},
        {'x': 5.5, 'y': [-1, 3, 10]},
        {'x': 6.6, 'y': [-2, 0, 4, 14]},
        {'x': 8.8, 'y': [5, 15, 25]},
        {'x': 11.11, 'y': [6, 26]},
        {'x': 33.33, 'y': [77]},
        {'x': 77.77, 'y': []}
    ]
```

Currently sorting of a `RecodArray` returns a `{"x": 9 * float64, "y": 9 * var * int64}` type with the following layout that seems to me unnecessarily complicated:
```python
<Record at="0">
    <RecordArray length="1">
        <field index="0" key="x">
            <RegularArray size="9">
                <content><NumpyArray format="d" shape="9" data="0 2.2 4.4 5.5 6.6 8.8 11.11 33.33 77.77" at="0x7f995cd86eb0"/></content>
            </RegularArray>
        </field>
        <field index="1" key="y">
            <RegularArray size="9">
                <content><ListOffsetArray64>
                    <offsets><Index64 i="[0 0 1 3 6 10 13 15 16 16]" offset="0" length="10" at="0x7f995d008200"/></offsets>
                    <content><NumpyArray format="l" shape="16" data="1 0 2 -1 3 ... 15 25 6 26 77" at="0x7f995cd85a10"/></content>
            </RegularArray>
        </field>
    </RecordArray>
</Record>
```
I'd like to change it to a simpler `9 * {"x": float64, "y": var * int64}` type:
```python
<RecordArray length="9">
    <field index="0" key="x">
        <NumpyArray format="d" shape="9" data="0 2.2 4.4 5.5 6.6 8.8 11.11 33.33 77.77" at="0x7f845bdb70b0"/>
    </field>
    <field index="1" key="y">
        <ListOffsetArray64>
            <offsets><Index64 i="[0 0 1 3 6 10 13 15 16 16]" offset="0" length="10" at="0x7f846c861600"/></offsets>
            <content><NumpyArray format="l" shape="16" data="1 0 2 -1 3 ... 15 25 6 26 77" at="0x7f845bdbb2b0"/></content>
        </ListOffsetArray64>
    </field>
</RecordArray>
```
Accessing the fields gives the same result as before:
```python
>>> sorted_array.x
<Array [0, 2.2, 4.4, 5.5, ... 11.1, 33.3, 77.8] type='9 * float64'>
>>> sorted_array.y
<Array [[], [1], [0, 2], ... [6, 26], [77], []] type='9 * var * int64'>
```

jpivarski:
Sorting only applies to numbers: `ak.sort` does not sort the records themselves. No comparison operators (`==`, `<=`, `<`, etc.) have been defined for records (or lists, for that matter), so we can't say what it means for an array of records to be in order.

Therefore, we can't sort `5 * {x: int64, y: float64}` (a simpler case) without sorting the _x_ values separately from the _y_ values. If the return type is `5 * {x: int64, y: float64}`, users would be very surprised to learn that the records are "broken." That is,

```
[
    {"x": 5, "y": 1.1},
    {"x": 4, "y": 2.2},
    {"x": 3, "y": 3.3},
    {"x": 2, "y": 4.4},
    {"x": 1, "y": 5.5},
]
```

would become

```
[
    {"x": 1, "y": 1.1},
    {"x": 2, "y": 2.2},
    {"x": 3, "y": 3.3},
    {"x": 4, "y": 4.4},
    {"x": 5, "y": 5.5},
]
```

The output records aren't (can't be) the same records as the input records. At least if the output has type `{x: 5 * int64, y: 5 * float64}`, that's an indication that the result is not what you think it might be.

Alternatively, we could _refuse_ to sort data with records in it if even this output is too hard to understand. I don't think we need to do this, though.

ianna:
closed in favour of https://github.com/scikit-hep/awkward-1.0/pull/948

PR #929, opened 2021-06-15
Refactor: remove `ak.deprecations_as_errors`

agoose77:
This behaviour can be replicated with `warnings.simplefilter("error", category)`
Additionally remove usage in tests where no deprecated behaviour exists.

From `grep`ing the code, there are currently no clients of this global variable besides that of `deprecate`.

This will be useful for making the tests pass in #917, but is not directly related to it.

### DIFF
--- a/src/awkward/__init__.py
+++ b/src/awkward/__init__.py
@@ -14,8 +14,6 @@ if distutils.version.LooseVersion(numpy.__version__) < distutils.version.LooseVe
 ):
     raise ImportError("Numpy 1.13.1 or later required")
 
-deprecations_as_errors = False
-
 # NumPy-like alternatives
 import awkward.nplike
 
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -46,12 +46,18 @@ def exception_suffix(filename):
     )
 
 
-def deprecate(message, version, date=None, category=FutureWarning):
+# Enable warnings for the Awkward package
+warnings.filterwarnings("default", module="awkward.*")
+
+
+def deprecate(message, version, date=None, category=DeprecationWarning):
     if date is None:
         date = ""
     else:
         date = " (target date: " + date + ")"
     warning = """In version {0}{1}, this will be an error.
+
+To raise these warnings as errors, run `warnings.filterwarnings("error", module="awkward.*")`.
 
 {2}""".format(
         version, date, message
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -46,21 +46,17 @@ def exception_suffix(filename):
     )
 
 
-def deprecate(exception, version, date=None):
-    if ak.deprecations_as_errors:
-        raise exception
+def deprecate(message, version, date=None, category=FutureWarning):
+    if date is None:
+        date = ""
     else:
-        if date is None:
-            date = ""
-        else:
-            date = " (target date: " + date + ")"
-        message = """In version {0}{1}, this will be an error.
-(Set ak.deprecations_as_errors = True to get a stack trace now.)
+        date = " (target date: " + date + ")"
+    warning = """In version {0}{1}, this will be an error.
 
-{2}: {3}""".format(
-            version, date, type(exception).__name__, str(exception)
-        )
-        warnings.warn(message, FutureWarning)
+{2}""".format(
+        version, date, message
+    )
+    warnings.warn(warning, category)
 
 
 virtualtypes = (ak.layout.VirtualArray,)
--- a/tests/test_0184-concatenate-operation.py
+++ b/tests/test_0184-concatenate-operation.py
@@ -298,8 +298,6 @@ def test_numpyarray_concatenate():
 
 
 def test_numbers_and_records_concatenate():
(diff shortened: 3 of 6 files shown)


jpivarski:
This is good. I like the use of a standard library call, rather than a global variable in Awkward.

Our deprecation function and/or the custom one for `ak.fill_none`'s `axis` argument should recommend this line of code instead of leaving it to the user to come up with that.

Also, the warning for `ak.fill_none`'s `axis` should recommend the new default, since that's the best way to silence the warning when they see it.

agoose77:
@jpivarski are you happy for this to merge? I'll reply to your other points on #917

jpivarski:
The recommendations for `ak.fill_none`'s `axis` can go on #917, but `ak._util.deprecate` is missing a message to the user about calling `warnings.simplefilter("error", FutureWarning)`. (Maybe we need a subclass of `FutureWarning` called `AwkwardDeprecationWarning`, so that such a filter doesn't eliminate too much?)

It's true that the `ak._util.deprecate` isn't currently in use, but just to avoid forgetting to add such a message later.

agoose77:
> AwkwardDeprecationWarning

OK, I'll add a note about raising warnings as `Exceptions` using `simplefilter`. It's a really unfortunate choice of name because `simplefilter` doesn't really filter - it changes the handling for (a category of) warnings.

I'm not sure about adding a custom warning type: there are already some useful ones:
```python
In [1]: DeprecationWarning.__mro__
Out[1]: (DeprecationWarning, Warning, Exception, BaseException, object)

In [2]: FutureWarning.__mro__
Out[2]: (FutureWarning, Warning, Exception, BaseException, object)
````

If we use the built-in ones, the only potential problem is if the user makes `DeprecationWarning` raise an Error, and some of our code (or dependencies) implements such a warning. I'd recommend your guidance on this. Worst case is, we make an ugly diamond inheritance:
```python
In [6]: class AwkwardDeprecationWarning(DeprecationWarning, AwkwardWarning):
   ...:     pass
   ...:
   ...:

In [7]: AwkwardDeprecationWarning.__mro__
Out[7]:
(__main__.AwkwardDeprecationWarning,
 DeprecationWarning,
 __main__.AwkwardWarning,
 Warning,
 Exception,
 BaseException,
 object)
```

jpivarski:
Python's DeprecationWarning is hidden from ordinary users: it gets reported in testing, but not when a library is used interactively. That's why we have to use or inherit from FutureWarning.

The only problem I see with diamond inheritance is being able to guess the method resolution order, but exception classes rarely have any custom methods. So that inheritance diagram looks fine to me. However,

> If we use the built-in ones, the only potential problem is if the user makes `DeprecationWarning` raise an Error, and some of our code (or dependencies) implements such a warning.

Wouldn't that be desirable? If a user makes `DeprecationWarning` raise an error, then they want our deprecations to raise errors, too.

Or—change the name in all of the above to FutureWarning. I guess for clarity, ours should be named AwkwardFutureWarning, too (so it's easy to guess that it inherits from FutureWarning).

agoose77:
> Python's DeprecationWarning is hidden from ordinary users: it gets reported in testing, but not when a library is used interactively. That's why we have to use or inherit from FutureWarning.

We can have Awkward register a default filter for the warning level for its own package to solve this:
> module is a string containing a regular expression that the module name must match. The expression is compiled to be case-sensitive.

That way we can use the appropriate warning type

> The only problem I see with diamond inheritance is being able to guess the method resolution order, but exception classes rarely have any custom methods. So that inheritance diagram looks fine to me. However,
> 
> > If we use the built-in ones, the only potential problem is if the user makes `DeprecationWarning` raise an Error, and some of our code (or dependencies) implements such a warning.
> 
> Wouldn't that be desirable? If a user makes `DeprecationWarning` raise an error, then they want our deprecations to raise errors, too.

I was thinking if someone that Awkward uses for something internal emits a warning, that would then break our code that uses it. However, I hadn't seen the `module` filter as above- this enables us to scope the `raise` functionality to just Awkward. So, actually, we can just use the built-in warning types!

> Or—change the name in all of the above to FutureWarning. I guess for clarity, ours should be named AwkwardFutureWarning, too (so it's easy to guess that it inherits from FutureWarning).

jpivarski:
> However, I hadn't seen the `module` filter as above- this enables us to scope the `raise` functionality to just Awkward. So, actually, we can just use the built-in warning types!

Oh, okay, then let's use that.

agoose77:
OK, this should do it.

jpivarski (approved):
Great! I suppose the "default" filter means that they'll see the DeprecationWarning, then?

agoose77:
> Great! I suppose the "default" filter means that they'll see the DeprecationWarning, then?

Another example of an obvious name ;) Yes, by default Python has the following filters
```
default::DeprecationWarning:__main__
ignore::DeprecationWarning
ignore::PendingDeprecationWarning
ignore::ImportWarning
ignore::ResourceWarning
```
which ignores `DeprecationWarning`s unless they're raised in `__main__` (`default` enables them for `__main__`).

agoose77:
Thanks!

Issue #932, opened 2021-06-16
`from_parquet` with `lazy` raises `ArrowInvalid` 


agoose77:
This is a weird bug that I can only trigger for certain forms. This form fails, but if you remove any of the inner-most fields, or remove an outer set of list brackets, it no longer fails.

```python
import io

import awkward as ak

array = ak.Array(
    [[{"addr": {"aget": 3,  "cobo": 0}}]]
)

f = io.BytesIO()
ak.to_parquet(array, f)
f.seek(0)

restored = ak.from_parquet(f, lazy=True)
ak.materialized(restored)
```

agoose77:
~~OK, this is a fiddly one. I think what's happening is that `read_row_group` is being given the wrong column name; the column name seems to be `".list.item..."` whereas Arrow uses `""`~~

agoose77:
This is being tracked in https://issues.apache.org/jira/browse/ARROW-13151

jpivarski:
This was fixed in some version of Arrow. By now (Arrow 6.0.0), it's no longer a problem.

PR #935, opened 2021-06-17
Separate `from_parquet` into different routines

agoose77:
This PR refactors the `from_parquet` implementation into several `Reader` classes. These objects are used by the reduced-scope `_ParquetGenerator` class to build a lazy array, or by the greedy (up-front) construction of an Array when `lazy=False`.

Some initial work is done to prepare for `_ArrowDatasetReader`.

This PR should *not* add any new functionality or (hopefully) break anything.

I plan in another PR to then remove all of the dataset readers, add an `_ArrowDatasetReader`, and leave only the `_ParquetFileReader` to handle file-like source arguments.

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -418,6 +418,13 @@ def wrap(content, behavior):
         return content
 
 
+def maybe_wrap(content, behavior, highlevel):
+    if highlevel:
+        return ak._util.wrap(content, behavior)
+    else:
+        return content
+
+
 def extra(args, kwargs, defaults):
     out = []
     for i in range(len(defaults)):
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -3467,34 +3467,34 @@ class _ParquetFileReader(_ParquetReader):
 
         if options is None:
             options = {}
-        self.file = pyarrow.parquet.ParquetFile(source, **options)
+        self._file = pyarrow.parquet.ParquetFile(source, **options)
 
         if row_groups is None:
-            row_groups = range(self.file.num_row_groups)
+            row_groups = range(self._file.num_row_groups)
 
-        schema = self.file.schema_arrow
+        schema = self._file.schema_arrow
         if columns is None:
             columns = schema.names
-
-        self.use_threads = use_threads
-
         schema = _partial_schema_from_columns(schema, columns)
+
+        self._use_threads = use_threads
+
         super(_ParquetFileReader, self).__init__(
             schema, row_groups, columns, partition_columns=[]
         )
 
     @property
     def row_group_metadata(self):
-        return [self.file.metadata.row_group(i) for i in self.row_groups]
+        return [self._file.metadata.row_group(i) for i in self.row_groups]
 
     def read(self, row_group, column_name):
-        return self.file.read_row_group(
-            row_group, [column_name], use_threads=self.use_threads
+        return self._file.read_row_group(
+            row_group, [column_name], use_threads=self._use_threads
         )
 
     def read_batches(self):
-        return self.file.read_row_groups(
-            self.row_groups, columns=self.columns, use_threads=self.use_threads
+        return self._file.read_row_groups(
+            self.row_groups, columns=self.columns, use_threads=self._use_threads
         )
 
 
@@ -3514,30 +3514,28 @@ class _ParquetDatasetReader(_ParquetReader):
         if options is None:
             options = {}
 
-        self.metadata_file = pyarrow.parquet.ParquetFile(metadata_filename, **options)
+        self._metadata_file = pyarrow.parquet.ParquetFile(metadata_filename, **options)
 
         if row_groups is None:
-            row_groups = range(self.metadata_file.num_row_groups)
+            row_groups = range(self._metadata_file.num_row_groups)
 
-        schema = self.metadata_file.schema_arrow
+        schema = self._metadata_file.schema_arrow
         if columns is None:
             columns = schema.names
-
         schema = _partial_schema_from_columns(schema, columns)
 
         if include_partition_columns:
             paths_and_counts = self._get_paths_and_counts(
-                self.metadata_file, row_groups
+                self._metadata_file, row_groups
             )
             partition_columns = _parquet_partitions_to_awkward(paths_and_counts)
         else:
             partition_columns = []
 
-        self.use_threads = use_threads
-        self.options = options
-        self.row_groups = row_groups
-        self.lookup = self._build_row_group_lookup(directory, self.metadata_file)
-        self.open_files = {}
+        self._use_threads = use_threads
+        self._lookup = self._build_row_group_lookup(directory, self._metadata_file)
+        self._options = options
+        self._open_files = {}
 
         super(_ParquetDatasetReader, self).__init__(
             schema, row_groups, columns, partition_columns
@@ -3587,18 +3585,18 @@ class _ParquetDatasetReader(_ParquetReader):
 
     @property
     def row_group_metadata(self):
-        return [self.metadata_file.metadata.row_group(i) for i in self.row_groups]
+        return [self._metadata_file.metadata.row_group(i) for i in self.row_groups]
 
     def read(self, row_group, column_name):
         import pyarrow.parquet
 
-        filename, local_row_group = self.lookup[row_group]
-        if filename not in self.open_files:
-            self.open_files[filename] = pyarrow.parquet.ParquetFile(
-                filename, **self.options
+        filename, local_row_group = self._lookup[row_group]
+        if filename not in self._open_files:
+            self._open_files[filename] = pyarrow.parquet.ParquetFile(
+                filename, **self._options
             )
-        return self.open_files[filename].read_row_group(
-            local_row_group, [column_name], use_threads=self.use_threads
+        return self._open_files[filename].read_row_group(
+            local_row_group, [column_name], use_threads=self._use_threads
         )
 
     def read_batches(self):
@@ -3606,12 +3604,12 @@ class _ParquetDatasetReader(_ParquetReader):
 
         batches = []
         for i in self.row_groups:
-            file_path, local_row_group = self.lookup[i]
-            local_file = pyarrow.parquet.ParquetFile(file_path, **self.options)
+            file_path, local_row_group = self._lookup[i]
+            local_file = pyarrow.parquet.ParquetFile(file_path, **self._options)
 
             batches.extend(
                 local_file.read_row_group(
-                    local_row_group, self.columns, use_threads=self.use_threads
+                    local_row_group, self.columns, use_threads=self._use_threads
                 ).to_batches()
             )
         return batches
@@ -3641,11 +3639,10 @@ class _ParquetDatasetOfFilesReader(_ParquetReader):
 
         if columns is None:
             columns = schema.names
-
         schema = _partial_schema_from_columns(schema, columns)
 
-        self.lookup = lookup
-        self.use_threads = use_threads
+        self._lookup = lookup
+        self._use_threads = use_threads
         super(_ParquetDatasetOfFilesReader, self).__init__(
             schema, row_groups, columns, partition_columns
         )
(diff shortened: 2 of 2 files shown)


martindurant:
Noting my issue #886 here

jpivarski:
This is a good point: the implementation might simplify a lot by using pyarrow's dataset function (which I hadn't known about).

agoose77:
In this PR I was planning to refactor the code and then take a step back and look at it. Perhaps it might be better to just rewrite it from scratch for the new API.

jpivarski:
I wouldn't be opposed, but take a good look at the old implementation—there's a lot going on there.

You definitely don't need to (or want to) rewrite the part that builds Forms with VirtualArrays and turns Parquet columns into Awkward Arrays. That's complicated because Parquet doesn't have lists with `starts`, `stops`, or `offsets`: it has "definition and repetition levels," which are a more bottom-up way of representing jaggedness. Unfortunately, this means that if you want to make a jagged array of records with virtual fields, you have to read one of the columns just to get the `offsets`. The bookkeeping involved in doing that is not something you want to get into, and it's orthogonal to the actual problem of managing datasets (multi-file data).

The code that handles datasets is too WET, as you've noticed, and one big function, rather than being broken down. That, along with @martindurant's suggestion if using pyarrow's function, would be definite improvements. But try to limit your scope to just that, or else this would quickly balloon.

agoose77:
> You definitely don't need to (or want to) rewrite the part that builds Forms with VirtualArrays and turns Parquet columns into Awkward Arrays. That's complicated because Parquet doesn't have lists with `starts`, `stops`, or `offsets`: it has "definition and repetition levels," which are a more bottom-up way of representing jaggedness. Unfortunately, this means that if you want to make a jagged array of records with virtual fields, you have to read one of the columns just to get the `offsets`. The bookkeeping involved in doing that is not something you want to get into, and it's orthogonal to the actual problem of managing datasets (multi-file data).

Yeah, "from scratch" is a slight exaggeration of my intentions. Having separated into routines, it's not as unruly as I first suspected, so I think we should be fine moving forward from here.

agoose77:
@jpivarski this is probably ready for some first-stage review if there's any going spare!

agoose77:
Ah drat, some tests were being skipped locally. I'll check those out.

src/awkward/_util.py
```diff
@@ -418,6 +418,13 @@ def wrap(content, behavior):
         return content
 
 
+def maybe_wrap(content, behavior, highlevel):
+    if highlevel:
+        return ak._util.wrap(content, behavior)
+    else:
+        return content
```
jpivarski, line 425: This could be used in a _lot_ of different places. Once this PR is merged, I can do a sweep to apply it consistently.

I think this kind of thing is on the borderline between DRY and WET, since it's such a small number of lines that it could be considered an "idiom," but the word "maybe_" in the function name makes it super-clear what this does (the tradeoff between indirection and repetition is not too bad on the indirection side: a reader of the code probably wouldn't have to look it up to see what "maybe_wrap" does).

src/awkward/operations/convert.py
```diff
@@ -2904,7 +2904,7 @@ def handle_arrow(obj):
                 return ak.operations.structure.concatenate(arrays, highlevel=False)
 
         elif (
-            isinstance(obj, Iterable)
+            isinstance(obj, Sequence)
```
jpivarski: Why not Iterable? Is it because Iterables are not guaranteed to have a length, and `len(obj)` fails with an exception?

If that's the case, then a more appropriately narrow guard would be `isinstance(obj, Iterable) and isinstance(obj, Sized)`. In the [Python collections ABCs](https://docs.python.org/3/library/collections.abc.html),

   * Iterable just means that it has `__iter__` (the `for x in obj` loops will work)
   * Sized just means that it has a `__len__` (the `len(obj) > 0` check will not raise an exception)
   * Sequence means that it has a `__len__` and a `__getitem__`. The `__getitem__` is not required by any other checks in the predicate or any steps in the body.

So just replace `isinstance(obj, Sequence)` with `(isinstance(obj, Iterable) and isinstance(obj, Sized))`; the `Sized` symbol can be imported the same way that `Sequence` is.

src/awkward/operations/convert.py
```diff
@@ -3367,23 +3292,387 @@ def __call__(self, row_group, unpack, length, form, lazy_cache, lazy_cache_key):
...
         masked = isinstance(form, ak.forms.ByteMaskedForm)
         if masked:
             form = form.content
-        table = self.read(row_group, column_name)
-        return _ParquetFile_arrow_to_awkward(table, struct_only, masked, unpack)
+        table = self._reader.read_row_group(row_group, [column_name])
+        return self._convert_arrow_to_awkward(table, struct_only, masked, unpack)
+
+
+class _ArrowReader(object):
```
jpivarski: As a superclass of three Parquet readers, I don't think the name "`_ArrowReader`" is appropriate. These are really all about Parquet, not Arrow. Arrow doesn't have concepts like "row groups," which is in most of the common method names. The name "`_ParquetReader`" isn't used by any of the subclasses, so that would be a good replacement.

src/awkward/operations/convert.py
```diff
@@ -3403,8 +3692,8 @@ def _parquet_partition_values(path):
 def _parquet_partitions_to_awkward(paths_and_counts):
     path, count = paths_and_counts[0]
     columns = [column for column, value in _parquet_partition_values(path)]
-    values = [[] for column in columns]
-    indexes = [[] for column in columns]
+    values = [[]] * len(columns)
+    indexes = [[]] * len(columns)
```
jpivarski: This raises loud warning bells: if you say `[[]] * len(columns)`, then you get a list of _references to the same empty list_, but if you say `[[] for column in columns]`, then you get a list of _distinct empty lists_. That matters if those empty lists are going to be filled, and it looks like that happens below.

If the original code relied upon the lists being independent—appending to one does not append the same item to all lists—which is very likely, then it would fail if it were tested for more than one column. It's quite possible, though, that the tests don't involve more than one column and the fact that this worked shows that we're missing such a test.

```python
>>> not_what_you_want = [[]] * 5
>>> not_what_you_want
[[], [], [], [], []]
>>> not_what_you_want[0].append("wee!")
>>> not_what_you_want
[['wee!'], ['wee!'], ['wee!'], ['wee!'], ['wee!']]
```

jpivarski: (As usual, mutation in-place is the source of most subtleties.)

agoose77:
OK, I've fixed the issues with the tests :relieved:

jpivarski (changes requested):
With such large chunks of code being moved, it's hard for me to tell by just reading it that the new code is equivalent to the old code, so mostly I'm relying on our tests to check it. Now would be a good time to do a spot-check with a coverage tool to see if any of the new functions correspond to untested cases. (Also, before this gets merged, I should test it on the Million Song Dataset (s3://pivarski-princeton/millionsongs/), which is not something we can put in the tests, but it's what I used to develop the partitioned interface in the first place.

There was one thing that looked like a bug, and I wrote an inline comment on it (`[[]] * len(columns)` vs `[[] for _ in columns]`), but I think you made an update and that part is gone now. I'll take another look after I press the green button on this review.

src/awkward/operations/convert.py
agoose77: Yes quite; I never use it in my own code for the same reasons (I also don't like it for the fact that `[]*n` isn't hugely intuitive even if it is a long term well-known feature. A senior moment I think!

src/awkward/operations/convert.py
agoose77: @jpivarski in the new dataset API, there is a `split_by_row_groups` implemented for non-parquet types IIRC, hence the name.

src/awkward/operations/convert.py
jpivarski: If this is a part of "I've fixed the issues with the tests" (your most recent force-push), then you can resolve this now.

I'm also glad that we have tests to catch it: this would have been a bad one.

src/awkward/operations/convert.py
jpivarski: Oh, so Arrow is aware of row groups now. Okay, but still, "Arrow" confuses me about the intention of this class. Since you chose the name by the fact that it's part of the Dataset interface, would "`_Dataset`" be a good name?

On the scale of things, this is the bunny, while `[[]] * len(columns)` was the tiger...

src/awkward/operations/convert.py
agoose77: I think we had the same thought; the two classes is more exact to our needs, but I also wondered whether sequence adequately covered our case. I'll change it!

src/awkward/operations/convert.py
agoose77: Haha perfect. I'm not one to bikeshed names - let's go with dataset

jpivarski:
I just tested `ak.from_parquet` on the Million Song Dataset with

   * directory containing `_common_metadata` and `_metadata`
   * directory without these files
   * a list of files
   * one file

and it works beautifully!

The only thing I'm explicitly asking for before merging is `(isinstance(obj, Iterable) and isinstance(obj, Sized))`. The name of the superclass is up to you; consider my feedback as you like.

martindurant:
Can you load directly from s3?

jpivarski:
> Can you load directly from s3?

I haven't tried that (because I don't know how it would be done in pyarrow by itself, if an S3 library has to be installed or what—I haven't read the documentation).

My test was on a local file.

martindurant:
You can maybe supply a path like "s3://bucket/path"; or else make a filesystem
```
import fsspec
fs = fsspec.filesystem("s3", anon=True)
```
and pass it as `fs=fs` in the calls to pyarrow. This would be a big deal!

martindurant:
(the latter invocation needs s3fs)

jpivarski:
That would be a great feature, and not too hard to implement via the route of passing `fs=fs` to pyarrow. (That is, not attempting to parse the URL and be aware of specific backends, but letting fsspec and pyarrow do that.)

This PR is a big change that supposed to be an interface-preserving refactor, so maybe in a new, much smaller PR?

agoose77:
@martindurant the plan is to remove all of the `_XXXReader` implementations apart from the file-like reader, and replace them with the new dataset API. That will be a tiny PR, and as @jpivarski, should be quite straightfoward :crossed_fingers:. I had your bug report in mind when writing this PR.

martindurant:
OK, keep me updated!

> apart from the file-like reader

You mean filesystem, right? Because reading parquet involves doing various folder listings and such.

agoose77:
> You mean filesystem, right? Because reading parquet involves doing various folder listings and such.

Sorry, didn't mean to edit your reply.


Currently we support passing in a file object (but not a file descriptor actually, which we perhaps should), which the new dataset API doesn't handle. For this reason, we'll need to keep a legacy API when we move to `arrow.dataset.dataset`

jpivarski:
I think it would be fine if passing a file-like object means you can't use the dataset API, but passing a string means you can. The main use-case that I can think of for passing a file-like object is `BytesIO`, which would necessarily have to be small data, hence a single "file."

jpivarski (approved):

agoose77:
@jpivarski when I started this PR, one of my intentions was to merge the code-paths between lazy and eager reading. I found a bug in the lazy reading that didn't happen for the eager (see #932). I think now that things are much cleaner, maybe we don't need to do that. I suppose in either case, that's for another PR!

agoose77:
OK, done :crossed_fingers:

martindurant:
If you want to use in-memory temporary "files", you might want to use MemoryFileSystem from fsspec
```
fs = fsspec.filesystem("memory")
```
which is fully compatible with other filesystems and, therefore, pyarrow. In the context of Dask, we never pass around actual open files.

agoose77:
Ah, I never checked the FS implementations in `fsspec`. That's very snazzy, I should probably consider supporting ffspec in my own work.

RE the file-object - it's just that we currently support it (and as a user, I sometimes have an open file that I want to use!)

jpivarski:
That makes a lot of sense for Dask, since an open file handle is a great example of a non-serializable, non-transferrable object. Trying to pass them around in a distributed environment would cause a lot of pain!

However, Awkward Array is not, by itself, a distributed processor; it's a library that could be used inside a distributed processor. There are enough people using it in a single-thread/single-machine setting that they'd be confused about a file-reading function not being able to take a file-like object as an argument. Python has a long history of passing file-like objects to functions like `json.load` as the primary API. I even got a request for adding this to Uproot: https://github.com/scikit-hep/uproot4/issues/98 There, the use-case was something I hadn't thought of: they had files in a tarball and wanted to use Python's `tarfile` to iterate over them.

I don't remember when `ak.from_parquet` acquired this ability to read file-like objects, but I wouldn't want to remove it. If it causes trouble for people using Awkward Array in a distributed setting—i.e. with Dask—then we'll explain that file-like objects are for small, non-distributed cases and they should be loading it by filename (and fsspec). That wouldn't be specific to Awkward Array, but would apply to anything in a distributed setting. That's also why the file-like object interface can miss out on this ability to read datasets (I don't see how the interface would make any sense for datasets, anyway).

(Uproot users of file-like objects don't get any of Uproot's parallel-reading features, which is a similar restriction.)

martindurant:
I agree in theory, and fastparquet supports file-like objects too; obviously, I've made something of a career in providing file-like objects to other libraries exactly for this reason. Indeed, I would have told your user to use fsspec's tar handling...

However, parquet is special in its directory-as-a-dataset approach, so at least that space needs good testing, both for local files and remote, with file-like objects (from fsspec or otherwise), *and* filesystem objects.

PR #937, opened 2021-06-18
Bugfix: truncate `ListOffsetArray` contents

### LINKED ISSUE #938: `ak.packed` doesn't truncate the contents of `ListOffsetArray`
```python3
For example, this layout
>>> content = ak.layout.NumpyArray(
...     np.array([1.1, 2.2, 3.3, 4.4, 5.5, 6.6, 7.7, 8.8, 9.9])
... )
>>> offsets = ak.layout.Index64(np.array([0, 3, 3, 5, 6]))
>>> layout = ak.layout.ListOffsetArray64(offsets, content)
```
Produces this when packed
```python3
>>> packed = ak.packed(layout, highlevel=False)
>>> packed.content
<NumpyArray format="d" shape="9" data="1.1 2.2 3.3 4.4 5.5 6.6 7.7 8.8 9.9" at="0x000001b74f90"/>
```
when it should be
```python3
>>> packed.content
<NumpyArray format="d" shape="6" data="1.1 2.2 3.3 4.4 5.5 6.6" at="0x0000017d2f90"/>
```

agoose77:
Currently, `ak.packed` doesn't truncate the contents of a `ListOffsetArray`, so buffers can be larger than necessary after calling the function.

Fixes #938

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2226,7 +2226,7 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             ),
         ):
             new_layout = layout.toListOffsetArray64(True)
-            new_length = new_layout.stops[-1]
+            new_length = new_layout.offsets[-1]
             return ak.layout.ListOffsetArray64(
                 new_layout.offsets,
                 apply(truncate(new_layout.content, new_length), depth + 1, posaxis),
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2226,9 +2226,10 @@ def _packed(array, axis=None, highlevel=True, behavior=None):
             ),
         ):
             new_layout = layout.toListOffsetArray64(True)
+            new_length = new_layout.stops[-1]
             return ak.layout.ListOffsetArray64(
                 new_layout.offsets,
-                apply(new_layout.content, depth + 1, posaxis),
+                apply(truncate(new_layout.content, new_length), depth + 1, posaxis),
                 new_layout.identities,
                 new_layout.parameters,
             )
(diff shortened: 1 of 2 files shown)


jpivarski:
This is good, thanks! But I think you didn't intend to check in src/awkward/operations/test-parquet.ipynb, right?

agoose77:
> This is good, thanks! But I think you didn't intend to check in src/awkward/operations/test-parquet.ipynb, right?

Oh drat :facepalm: . Working late :sleeping:

agoose77:
@jpivarski fixed!

src/awkward/operations/structure.py
```diff
@@ -2226,9 +2226,10 @@ def apply(layout, depth, posaxis):
             ),
         ):
             new_layout = layout.toListOffsetArray64(True)
+            new_length = new_layout.stops[-1]
```
jpivarski: Oh, I just noticed another thing: if the length of this ListOffsetArray is zero, the `stops` will be, too.  The `offsets` length would always be at least one, though, so this one would be okay with `stops` replaced by `offsets`.

jpivarski: (An empty ListOffsetArray's `offsets` has length 1, though the value of this `offsets[0]` is only guaranteed to be zero because the previous line ensured that the `offsets` starts with 0—the `True` in `toListOffsetArray`.)

jpivarski: I'm at my computer now (no longer emailing on my phone), so I'll get this one.

jpivarski: Done.

agoose77: > Oh, I just noticed another thing: if the length of this ListOffsetArray is zero, the `stops` will be, too. The `offsets` length would always be at least one, though, so this one would be okay with `stops` replaced by `offsets`.

Ah I didn't know that! Thanks for the catch. I'm a bit conscious of making these PRs when I'm not fully up to speed with every detail so i appreciate the input

Issue #941, opened 2021-06-18
`with_cache` should not accept `"new"` or bare-dict cache argument


agoose77:
In the case that `with_cache` is given the `"new"` literal or a bare-dict value for the cache, the returned array will ultimately be left with a broken cache. This is because in these cases a weak-referenceable proxy would is/need to be made which goes out of scope when `with_cache` returns.

agoose77:
Per #942 this is not correct!

PR #942, opened 2021-06-18
Bugfix: fix `with_cache` implementation

### LINKED ISSUE #940: `with_cache` raises `NameError` if cache is given and layout already has one
Traceback
```pytb
NameError                                 Traceback (most recent call last)
<ipython-input-55-18e9a13fc6a7> in <module>
----> 1 dataset = _from_hdf5_dataset("/tmp/dataset")

/mnt/app/texat/src/texat/awkward/io.ipynb in _from_hdf5_dataset(directory, lazy, h5py_options)
     82     partitions = sum([ak.partition.every(l) for l in layouts], start=[])
     83     layout = ak.partitioned(partitions, highlevel=False)
---> 84     cached = ak.with_cache(layout, cache={})
     85     return ak.Array(cached)
     86 

/opt/texat-venv/lib/python3.9/site-packages/awkward/operations/structure.py in with_cache(array, cache, highlevel, behavior)
   4193             return None
   4194 
-> 4195     out = ak._util.recursively_apply(
   4196         ak.operations.convert.to_layout(array), getfunction, pass_depth=False
   4197     )

/opt/texat-venv/lib/python3.9/site-packages/awkward/_util.py in recursively_apply(layout, getfunction, pass_depth, pass_user, pass_apply, user, keep_parameters, numpy_to_regular)
   1322             )
   1323 
-> 1324     return apply(layout, 1, user)
   1325 
   1326 

/opt/texat-venv/lib/python3.9/site-packages/awkward/_util.py in apply(layout, depth, user)
   1127         if isinstance(layout, ak.partition.PartitionedArray):
   1128             return ak.partition.IrregularlyPartitionedArray(
-> 1129                 [apply(x, depth, user) for x in layout.partitions]
   1130             )
   1131 

/opt/texat-venv/lib/python3.9/site-packages/awkward/_util.py in <listcomp>(.0)
   1127         if isinstance(layout, ak.partition.PartitionedArray):
   1128             return ak.partition.IrregularlyPartitionedArray(
-> 1129                 [apply(x, depth, user) for x in layout.partitions]
   1130             )
   1131 

/opt/texat-venv/lib/python3.9/site-packages/awkward/_util.py in apply(layout, depth, user)
   1120         custom = getfunction(layout, *args)
   1121         if callable(custom):
-> 1122             return custom()
   1123         else:
   1124             user = custom

/opt/texat-venv/lib/python3.9/site-packages/awkward/operations/structure.py in <lambda>()
   4185             return lambda: ak.layout.VirtualArray(
   4186                 layout.generator,
-> 4187                 newcache,
   4188                 layout.cache_key,
   4189                 layout.identities,

NameError: free variable 'newcache' referenced before assignment in enclosing scope
```

agoose77:
This PR fixes #940 and #941.

`with_cache` is one of two places (there's also something going on in the `convert` module w.r.t cache handling) the current cache handling is slightly broken. This PR simplifies the scope of the function by requiring the user to have created the cache prior to calling `with_cache`. Currently, the lifetime of the cache proxy that is created in the removed cases is not associated with the lifetime of the cache.

It would be possible to retain the existing API by storing a reference to the proxy on the cache object, but this will break the assumption that the cache is populated exclusively for virtual materialisation.

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -4125,7 +4125,7 @@ def materialized(array, highlevel=True, behavior=None):
         return out
 
 
-def with_cache(array, cache, highlevel=True, behavior=None):
+def with_cache(array, cache, behavior=None):
     """
     Args:
         array: Data to search for nested virtual arrays.
@@ -4135,8 +4135,6 @@ def with_cache(array, cache, highlevel=True, behavior=None):
             re-generated if `__getitem__` raises a `KeyError`. This mapping may
             evict elements according to any caching algorithm (LRU, LFR, RR,
             TTL, etc.). If "new", a new dict (keep-forever cache) is created.
-        highlevel (bool): If True, return an #ak.Array; otherwise, return
-            a low-level #ak.layout.Content subclass.
         behavior (None or dict): Custom #ak.behavior for the output array, if
             high-level.
 
@@ -4177,13 +4175,9 @@ def with_cache(array, cache, highlevel=True, behavior=None):
 
     def getfunction(layout):
         if isinstance(layout, ak.layout.VirtualArray):
-            if cache is None:
-                newcache = layout.cache
-            elif layout.cache is None:
-                newcache = cache
             return lambda: ak.layout.VirtualArray(
                 layout.generator,
-                newcache,
+                cache,
                 layout.cache_key,
                 layout.identities,
                 layout.parameters,
@@ -4194,10 +4188,7 @@ def with_cache(array, cache, highlevel=True, behavior=None):
     out = ak._util.recursively_apply(
         ak.operations.convert.to_layout(array), getfunction, pass_depth=False
     )
-    if highlevel:
-        return ak._util.wrap(out, ak._util.behaviorof(array, behavior=behavior))
-    else:
-        return out
+    return ak._util.wrap(out, ak._util.behaviorof(array, behavior=behavior))
 
 
 @ak._connect._numpy.implements("size")
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -4125,7 +4125,7 @@ def materialized(array, highlevel=True, behavior=None):
         return out
 
 
-def with_cache(array, cache, behavior=None):
+def with_cache(array, cache, highlevel=True, behavior=None):
     """
     Args:
         array: Data to search for nested virtual arrays.
@@ -4135,6 +4135,8 @@ def with_cache(array, cache, behavior=None):
             re-generated if `__getitem__` raises a `KeyError`. This mapping may
             evict elements according to any caching algorithm (LRU, LFR, RR,
             TTL, etc.). If "new", a new dict (keep-forever cache) is created.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.layout.Content subclass.
         behavior (None or dict): Custom #ak.behavior for the output array, if
             high-level.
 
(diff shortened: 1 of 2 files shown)


jpivarski:
Wait—the `cache="new"` and `cache={}` options are a necessary part of the `with_cache API. If there's something wrong with then, they need to be fixed, but they can't be removed.

The weak reference is a concession to the C++ layer that will be replaced with a regular reference when it's translated into Python. It prevents cyclic references through the C++ that Python's garbage collector is unable to identify. So for now, we have to live with the weak reference.

The way this is supposed to work is that any reference to the cache must be retained within the scope of the `with_cache` function, since the VirtualArray/ArrayCache isn't holding it. Before the function returns, it creates an `ak.Array` that _does_ hold the reference, strongly.

If you were using the function with `highlevel=False`, that mechanism would have broken. Is that what happened? It could be that we have to disable `highlevel=False` in this function—it's just not going to work, considering how the reference is held only by the `ak.Array` object. The `highlevel=False` will have to wait until the C++ layer is translated into Python.

With `highlevel=False`, `cache="new"` and `cache={}` will fall, but so would `cache=anything_constructed_on_the_spit()`, since its reference only exists in the argument list and will be fine after the `with_cache` function returns.

With `highlevel=False`, you can still use this for making low-level layouts. You have to pull the layout out of the unwanted `ak.Array`, but that `ak.Array` has to live long enough for the final `ak.Array` to pick up strong references to the caches.

agoose77:
> The way this is supposed to work is that any reference to the cache must be retained within the scope of the with_cache function, since the VirtualArray/ArrayCache isn't holding it. Before the function returns, it creates an ak.Array that does hold the reference, strongly.

Right
> If you were using the function with highlevel=False, that mechanism would have broken. Is that what happened? It could be that we have to disable highlevel=False in this function—it's just not going to work, considering how the reference is held only by the ak.Array object. The highlevel=False will have to wait until the C++ layer is translated into Python.
Yes, this is likely the culprit.

OK, then I'll repush this branch with the fix for #940.

agoose77:
@jpivarski OK, corrected my oversight. I've removed the `highlevel` argument, and I think*think* fixed the cache handling so that it matches the docstring. Could you give that a quick once over?

src/awkward/operations/structure.py
```diff
@@ -4177,13 +4175,9 @@ def with_cache(array, cache, highlevel=True, behavior=None):
...
     def getfunction(layout):
         if isinstance(layout, ak.layout.VirtualArray):
-            if cache is None:
-                newcache = layout.cache
-            elif layout.cache is None:
-                newcache = cache
             return lambda: ak.layout.VirtualArray(
                 layout.generator,
-                newcache,
+                cache,
```
jpivarski: This change should also be reverted. `newcache` only exists to switch between `cache` and `layout.cache`. In fact, flake8 should have complained that `newcache` was being defined without being used.

I _think_ that the only change this PR needs is to disable the possibility of `highlevel=False`.

jpivarski:
Just removing the `highlevel` parameter should do it, but eventually somebody's going to ask, "Why doesn't this function have a `highlevel` argument? Here, I'll add one."

To ensure that it doesn't just look like an oversight, let's leave the argument in (in other words, revert almost everything in this PR) and just add an error message if `highlevel is not True`. That makes the `highlevel` argument useless, but it does communicate the fact that it was removed on purpose, and the error message would say, "ak.with_cache cannot allow highlevel=False because the only strong reference to caches are held by ak.Array objects; VirtualArrays only hold weak references, which would go out of scope with this function. This will be fixed in Awkward 2.0, when VirtualArrays are reimplemented in Python and can safely hold strong references to caches. For now, use highlevel=True and extract the layout from the high-level array" (with good formatting so that that long explanation isn't on one long line).

This PR then just becomes a prevention of `highlevel=False` with an explanation of why it is temporarily disabled.

agoose77:
@jpivarski is be tempted to just put this information in a comment, but as it is planned to support this in future, we might as well keep the API stable. I'll make the changes at some point :)

src/awkward/operations/structure.py
jpivarski: I missed the fact that lines 4180 through 4183 are red (removed).

jpivarski:
I just took care of it. Thanks for opening this!

agoose77:
@jpivarski from the docstring, isn't the cache=None case supposed to unset the caches entirely?

jpivarski:
I think you might be right about that. I might have just committed a mistake. I didn't look carefully enough at the logic of it, I assumed that the old logic was right.

Is it now in disagreement with the documentation?

agoose77:
Again, could be wrong - long week. The existing implementation has a case
for which the existing cache is left in place. As fair as I can tell, the
cache should either be replaced or removed, with no pass through condition.

PR #944, opened 2021-06-18
test data generation

SantamRC:
@jpivarski check this out. 
In here, if we run `python dev/generate-strategies.py` it creates composite strategies by making use of contraints.json for particular kernel functions inside hypothesis-tests-spec folder. 
The plan is , after the strategies are generated, I would call those strategies and fetch the inputs they generate. For each of those input I would dynamically call the kernel and store the input and the output in a json file (tests-data/json). 
Then make a change in the current generate-tests.py to fetch those input and outputs and generate the test cases.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,10 @@ tests-cuda-kernels/test*
 tests-spec/kernels.py
 tests-spec/test*
 
+##Kernel tests using hypothesis
+hypothesis-tests-spec/strategies.py
+hypothesis-tests-spec/test*
+
 dependent-project/CMakeFiles
 dependent-project/CMakeCache.txt
 dependent-project/Makefile
--- a/constraints.json
+++ b/constraints.json
@@ -37,12 +37,12 @@
                     "var":"lsb_order",
                     "role":"Stores the lsb order",
                     "strategy_name":"booleans",
-                    "constraints": {                      
+                    "constraints": {
                     }
                 }
             ]
         }
-        
+
     ]
-    
+
 }
--- a/constraints.json
+++ b/constraints.json
@@ -1,0 +1,48 @@
+{
+    "kernels":[
+        {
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -14,6 +14,10 @@ tests-cuda-kernels/test*
 tests-spec/kernels.py
 tests-spec/test*
 
+##Kernel tests using hypothesis
+hypothesis-tests-spec/strategies.py
```
ianna, line 18: @SantamRC - is it something you wrote?

constraints.json
```diff
@@ -0,0 +1,48 @@
+{
+    "kernels":[
+        {
+            "name":"awkward_BitMaskedArray_to_ByteMaskedArray",
```
ianna, line 4: I guess it's a test name. Are you planning on writing ~2000+ constraints for each test?

constraints.json
```diff
@@ -0,0 +1,48 @@
...
+            "arguments":[
+                {
+                    "name":"BitMaskedArray-mask",
+                    "var":"mask",
```
ianna, line 8: I guess it's a variable name in this test. What if someone decides to name it differently?

constraints.json
```diff
@@ -0,0 +1,48 @@
...
+                    "role":"Stores the mask array",
+                    "strategy_name":"lists",
+                    "constraints": {
+                        "min_size":0,
+                        "max_size":6,
```
ianna, line 13: what does it define? A maximum value or a maximum length?

ianna (commented):
Good start. I couldn't run it due to missing "`strategies.py`" file.

jpivarski:
@SantamRC Is this PR still active, or is it superseded by #971?

SantamRC:
> @SantamRC Is this PR still active, or is it superseded by #971?

This PR is not active anymore

PR #946, opened 2021-06-22
sorting an indexed option array in `axis0` bug fix; argsort to account `None`s

### LINKED ISSUE #967: `argsort` produces invalid layout for `ListOffsetArray` with `offset[0] != 0`
Given the following layout
```python3
layout = ak.layout.ListOffsetArray64(
    ak.layout.Index64([1, 4]),
    ak.layout.NumpyArray(np.arange(64)),
)
```
calling `argsort` does not produce a valid layout. It appears that the `index` parameter is not projected to start at 0:
```python3
>>> ak.argsort(layout, highlevel=False)
<ListOffsetArray64>
    <offsets><Index64 i="[1 4]" offset="0" length="2" at="0x00000ba033d0"/></offsets>
    <content><NumpyArray format="l" shape="3" data="0 1 2" at="0x000003c7b190"/></content>
</ListOffsetArray64>

```

ianna:
* do not drop `None`s  when sorting a 1D `IndexedOptionArray`
* `argsort` to account `None`s by passing and post-processing `shifts`:

```python
>>> import awkward as ak
>>> array = ak.Array([None, None, 1, -1, 30])
>>> ak.sort(array)
<Array [-1, 1, 30, None, None] type='5 * ?int64'>
>>> ak.argsort(array)
<Array [3, 2, 4, None, None] type='5 * ?int64'>
>>> array[ak.argsort(array)]
<Array [-1, 1, 30, None, None] type='5 * ?int64'>
```
* unused `keepdim` has been dropped
* `argsort` documentation corrected

### DIFF
--- a/src/libawkward/array/ListOffsetArray.cpp
+++ b/src/libawkward/array/ListOffsetArray.cpp
@@ -1691,15 +1691,13 @@ namespace awkward {
                                   const Index64& parents,
                                   int64_t outlength,
                                   bool ascending,
-                                  bool stable,
-                                  bool keepdims) const {
+                                  bool stable) const {
     return toListOffsetArray64(true).get()->sort_next(negaxis,
                                                       starts,
                                                       parents,
                                                       outlength,
                                                       ascending,
-                                                      stable,
-                                                      keepdims);
+                                                      stable);
   }
 
   template <>
@@ -1709,8 +1707,7 @@ namespace awkward {
     const Index64& parents,
     int64_t outlength,
     bool ascending,
-    bool stable,
-    bool keepdims) const {
+    bool stable) const {
 
     if (length() == 0 ) {
       return shallow_copy();
@@ -1815,7 +1812,7 @@ namespace awkward {
 
       ContentPtr outcontent = nextcontent.get()->sort_next(
         negaxis - 1, nextstarts, nextparents, nextcontent.get()->length(),
-        ascending, stable, false);
+        ascending, stable);
 
       Index64 outcarry(nextlen);
       struct Error err5 = kernel::ListOffsetArray_local_preparenext_64(
@@ -1857,7 +1854,7 @@ namespace awkward {
 
       ContentPtr outcontent = trimmed.get()->sort_next(
         negaxis, util::make_starts(offsets_), nextparents,
-        offsets_.length() - 1, ascending, stable, false);
+        offsets_.length() - 1, ascending, stable);
 
       ContentPtr out = std::make_shared<ListOffsetArray64>(Identities::none(),
                                                            parameters_,
--- a/src/libawkward/array/ListOffsetArray.cpp
+++ b/src/libawkward/array/ListOffsetArray.cpp
@@ -1868,12 +1868,14 @@ namespace awkward {
   const ContentPtr
   ListOffsetArrayOf<T>::argsort_next(int64_t negaxis,
                                      const Index64& starts,
+                                     const Index64& shifts,
                                      const Index64& parents,
                                      int64_t outlength,
                                      bool ascending,
                                      bool stable) const {
     return toListOffsetArray64(true).get()->argsort_next(negaxis,
                                                          starts,
+                                                         shifts,
                                                          parents,
                                                          outlength,
                                                          ascending,
@@ -1884,6 +1886,7 @@ namespace awkward {
   const ContentPtr
   ListOffsetArrayOf<int64_t>::argsort_next(int64_t negaxis,
                                            const Index64& starts,
+                                           const Index64& shifts,
                                            const Index64& parents,
                                            int64_t outlength,
                                            bool ascending,
@@ -1988,11 +1991,37 @@ namespace awkward {
         nextlen);
       util::handle_error(err4, classname(), identities_.get());
 
+      bool make_shifts = true;
+
+      Index64 nextshifts(nextlen);
+      if (make_shifts) {
+        Index64 nummissing(maxcount);
+        Index64 missing(offsets_.getitem_at(offsets_.length() - 1));
+        struct Error err7 = kernel::ListOffsetArray_reduce_nonlocal_nextshifts_64(
+          kernel::lib::cpu,   // DERIVE
+          nummissing.data(),
+          missing.data(),
+          nextshifts.data(),
+          offsets_.data(),
+          offsets_.length() - 1,
+          starts.data(),
+          parents.data(),
+          maxcount,
+          nextlen,
+          nextcarry.data());
+        util::handle_error(err7, classname(), identities_.get());
+      }
+
       ContentPtr nextcontent = content_.get()->carry(nextcarry, false);
 
       ContentPtr outcontent = nextcontent.get()->argsort_next(
-        negaxis - 1, nextstarts, nextparents, maxnextparents + 1,
-        ascending, stable);
+        negaxis - 1,
+        nextstarts,
+        nextshifts,
+        nextparents,
+        maxnextparents + 1,
+        ascending,
+        stable);
 
       Index64 outcarry(nextlen);
       struct Error err5 = kernel::ListOffsetArray_local_preparenext_64(
@@ -2031,10 +2060,14 @@ namespace awkward {
 
       ContentPtr trimmed = content_.get()->getitem_range_nowrap(globalstart,
                                                                 globalstop);
-
       ContentPtr outcontent = trimmed.get()->argsort_next(
-        negaxis, util::make_starts(offsets_), nextparents,
-        offsets_.length() - 1, ascending, stable);
+        negaxis,
+        util::make_starts(offsets_),
+        shifts,
+        nextparents,
+        offsets_.length() - 1,
+        ascending,
+        stable);
 
       ContentPtr out = std::make_shared<ListOffsetArray64>(Identities::none(),
                                                            util::Parameters(),
--- a/src/libawkward/array/ListOffsetArray.cpp
+++ b/src/libawkward/array/ListOffsetArray.cpp
@@ -1824,12 +1824,7 @@ namespace awkward {
 
       outcontent = outcontent.get()->carry(outcarry, false);
 
-      Index64 outoffsets(offsets_.length());
-      // FIXME: move to kernel
-      for (int64_t i = 0; i < offsets_.length(); i++) {
-        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
-      }
-
+      Index64 outoffsets = compact_offsets64(true);
       ContentPtr out = std::make_shared<ListOffsetArray64>(Identities::none(),
                                                            parameters_,
                                                            outoffsets,
@@ -1862,12 +1857,7 @@ namespace awkward {
         negaxis, util::make_starts(offsets_), nextparents,
         offsets_.length() - 1, ascending, stable);
 
-      Index64 outoffsets(offsets_.length());
-      // FIXME: move to kernel
-      for (int64_t i = 0; i < offsets_.length(); i++) {
-        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
-      }
-
+      Index64 outoffsets = compact_offsets64(true);
       ContentPtr out = std::make_shared<ListOffsetArray64>(Identities::none(),
                                                            parameters_,
                                                            outoffsets,
@@ -2041,12 +2031,7 @@ namespace awkward {
 
       outcontent = outcontent.get()->carry(outcarry, false);
 
-      Index64 outoffsets(offsets_.length());
-      // FIXME: move to kernel
-      for (int64_t i = 0; i < offsets_.length(); i++) {
-        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
-      }
-
+      Index64 outoffsets = compact_offsets64(true);;
       ContentPtr out = std::make_shared<ListOffsetArray64>(Identities::none(),
                                                            util::Parameters(),
                                                            outoffsets,
@@ -2083,12 +2068,7 @@ namespace awkward {
         ascending,
         stable);
 
-      Index64 outoffsets(offsets_.length());
-      // FIXME: move to kernel
-      for (int64_t i = 0; i < offsets_.length(); i++) {
-        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
-      }
-
+      Index64 outoffsets = compact_offsets64(true);
       ContentPtr out = std::make_shared<ListOffsetArray64>(Identities::none(),
                                                            util::Parameters(),
                                                            outoffsets,
--- a/src/libawkward/array/ListOffsetArray.cpp
+++ b/src/libawkward/array/ListOffsetArray.cpp
@@ -1824,9 +1824,15 @@ namespace awkward {
 
       outcontent = outcontent.get()->carry(outcarry, false);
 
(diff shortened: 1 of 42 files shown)


ianna:
@jpivarski - should the following `argsort` feature be considered a bug?

Take an example where `argsort` is used to generate indices to get a sorted array:
```python
>>> import awkward as ak
>>> array = ak.Array([-5, 4, 1, -1, 30])
>>> index = ak.argsort(array)
>>> array[index]
<Array [-5, -1, 1, 4, 30] type='5 * int64'>
>>> ak.sort(array)
<Array [-5, -1, 1, 4, 30] type='5 * int64'>
```
I'd assume the same should work for an array containing `None`s:
```python
>>> array = ak.Array([None, None, 1, -1, 30])
>>> index = ak.argsort(array)
>>> array[index]
<Array [None, None, 1, None, None] type='5 * ?int64'>
>>> ak.sort(array)
<Array [-1, 1, 30, None, None] type='5 * ?int64'>
```
while it is possible to get sorted list from:
```python
>>> ak.to_list(array.layout.content[index])
[-1, 1, 30, None, None]
```

jpivarski:
Yes, that's wrong. It looks like `ak.argsort` is returning type `?int64`, rather than `int64`. Is this a manifestation of the fact that 1-dimensional arrays were not handled until recently? Does the same thing happen when argsorting deeper lists?

Since `ak.sort` puts missing values at the end (`None` is after all numbers), `ak.argsort` should put the index positions of those missing values last, just as it would with any other value.

ianna:
> Yes, that's wrong. It looks like `ak.argsort` is returning type `?int64`, rather than `int64`. Is this a manifestation of the fact that 1-dimensional arrays were not handled until recently? Does the same thing happen when argsorting deeper lists?
> 
> Since `ak.sort` puts missing values at the end (`None` is after all numbers), `ak.argsort` should put the index positions of those missing values last, just as it would with any other value.

Yes, it's the same for deeper lists.

tests/test_0074-argsort-and-sort.py
```diff
@@ -368,15 +466,25 @@ def test_RecordArray():
...
-        [],
+        [1],
+        [2, 2],
+        [3, 3, 3],
+        [4, 4, 4, 4],
+        [5, 5, 5],
+        [6, 6],
+        [7],
+        []
+        # [],
```
ianna, line 471: @jpivarski - this result at `axis=0` looks a bit odd to me. Though I have a feeling it it correct. It's more obvious if I modify the array to:
```python
>>> array = ak.Array([[], [-1], [2, 0.2], [0.3, 3.3, 3], [4, 4.4, 0.4, -4], [-5, 5, 5.5], [0.6, 6], [-7], []])
>>> ak.argsort(array, axis=0)
<Array [[], [7], [5, 2], ... [2, 6], [4], []] type='9 * var * int64'>
>>> ak.sort(array, axis=0)
<Array [[], [-7], [-5, 0.2, ... 2, 6], [4], []] type='9 * var * float64'>
```
and then try to access values by an argsorted index individually. They do correlate with the same values from a sorted array.

BTW, I can use the `argsort` at `axis=-1` to sort the initial array in the same axis, but not if I use `axis=0`:
```python
>>> array[ak.argsort(array)]
<Array [[], [-1], [0.2, 2, ... 6], [-7], []] type='9 * var * float64'>
>>> array[ak.argsort(array, axis=0)]
  File "/Users/yana/Projects/PR945/awkward-1.0/awkward/highlevel.py", line 993, in __getitem__
    tmp = ak._util.wrap(self.layout[where], self._behavior)
ValueError: in ListArray64 attempting to get 7, index out of range

(https://github.com/scikit-hep/awkward-1.0/blob/1.4.0rc2/src/cpu-kernels/awkward_ListArray_getitem_jagged_apply.cpp#L46)
```
Is it a feature or a bug? Thanks!

ianna (commented):
@jpivarski - I've scaled down `argsort` to use an `std` version as I have initially planned to modify it. I've taken your advise to post-process the indices - there is a new kernel for that.

src/libawkward/array/ListOffsetArray.cpp
```diff
@@ -1872,19 +1860,18 @@ namespace awkward {
...
       ContentPtr outcontent = trimmed.get()->sort_next(
         negaxis, util::make_starts(offsets_), nextparents,
-        offsets_.length() - 1, ascending, stable, false);
+        offsets_.length() - 1, ascending, stable);
+
+      Index64 outoffsets(offsets_.length());
+      // FIXME: move to kernel
+      for (int64_t i = 0; i < offsets_.length(); i++) {
+        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
+      }
```
ianna, line 1869: @agoose77 - same here: a temporary fix - I left it for you to provide a proper one :-)

src/libawkward/array/ListOffsetArray.cpp
```diff
@@ -2040,17 +2041,16 @@ namespace awkward {
 
       outcontent = outcontent.get()->carry(outcarry, false);
 
+      Index64 outoffsets(offsets_.length());
+      // FIXME: move to kernel
+      for (int64_t i = 0; i < offsets_.length(); i++) {
+        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
+      }
+
```
ianna: @agoose77 - same here: a temporary fix - I left it for you to provide a proper one :-)

src/libawkward/array/ListOffsetArray.cpp
```diff
@@ -1835,17 +1824,16 @@ namespace awkward {
...
```
ianna: @agoose77 - here is a temporary fix - I left it for you to provide a proper one :-)

src/libawkward/array/ListOffsetArray.cpp
ianna: @agoose77 - same here: a temporary fix - I left it for you to provide a proper one :-)

src/libawkward/array/ListOffsetArray.cpp
```diff
@@ -2074,22 +2074,25 @@ namespace awkward {
...
+        nextparents,
+        offsets_.length() - 1,
+        ascending,
+        stable);
+
+      Index64 outoffsets(offsets_.length());
+      // FIXME: move to kernel
+      for (int64_t i = 0; i < offsets_.length(); i++) {
+        outoffsets.data()[i] = offsets_.data()[i] - offsets_.data()[0];
+      }
```
ianna: @agoose77 - same here: a temporary fix - I left it for you to provide a proper one :-)

tests/test_0074-argsort-and-sort.py
jpivarski: This is the important thing:

```python
>>> ak.all(array[ak.argsort(array, axis=-1)] == ak.sort(array, axis=-1))
True
```

As for `axis=0`, the meaning of sorting with `axis=0` cuts across lists:

```python
>>> ak.sort(array, axis=0).tolist()
[[], [-7.0], [-5.0, 0.2], [-1.0, 3.3, 0.4], [0.3, 4.4, 3.0, -4.0], [0.6, 5.0, 5.5], [2.0, 6.0], [4.0], []]
```

That is, `-7.0` is the smallest number that is first in any list, `-5.0` is the second smallest number that is first in any list, and `0.2` is the smallest number that is second in any list. I don't know that it makes sense to maintain the shapes of all these lists when you're doing that, and I don't know what analysts would _want_ it to mean in the context of a data analysis.

At least, it's no mystery that `array[ak.argsort(array, axis=0)]` fails: the index positions returned by `ak.argsort` cut across all of the lists—they can be values that are much larger than the lengths of any of the lists.

```python
>>> ak.argsort(array, axis=0).tolist()
[[], [7], [5, 2], [1, 3, 4], [3, 4, 3, 4], [6, 5, 5], [2, 6], [4], []]
```

So what should we do about it? Maybe—without removing any of the the code that works—we should put in an exception that refuses to return results when `axis != -1` (or the equivalent positive `axis`, so this has to be in the C++ code, right after `posaxis` is determined from `wrap_axis_if_negative`). Then we're not promising that this is the behavior, such that someone can start relying on it before we know what the right policy should be. Also, if someone really does want `axis != -1 and axis is not None`, they'll complain that it's not working and we can ask them about their use-case and understand what the appropriate output would be.

What do you think?

jpivarski (commented):
So, this is correct:

```python
>>> ak.sort([1, 2, None, 3, 0, None], axis=0)
<Array [0, 1, 2, 3, None, None] type='6 * ?int64'>
>>> ak.sort([[1, 2, None, 3, 0, None]], axis=1)
<Array [[0, 1, 2, 3, None, None]] type='1 * var * ?int64'>
```

and this is consistent with the above:

```python
>>> ak.argsort([1, 2, None, 3, 0, None], axis=0)
<Array [4, 0, 1, 3, None, None] type='6 * ?int64'>
>>> ak.argsort([[1, 2, None, 3, 0, None]], axis=1)
<Array [[4, 0, 1, 3, None, None]] type='1 * var * ?int64'>
```

in the sense that the argsort can be used as a slice on the original array to effectively get the sort:

```python
>>> array = ak.Array([1, 2, None, 3, 0, None])
>>> ak.argsort(array)
<Array [4, 0, 1, 3, None, None] type='6 * ?int64'>
>>> array[ak.argsort(array)]
<Array [0, 1, 2, 3, None, None] type='6 * ?int64'>
>>> ak.sort(array)
<Array [0, 1, 2, 3, None, None] type='6 * ?int64'>
```

However, the fact that it's doing so by returning missing values in the argsort could be a problem for some applications. The way I would have expected this to go is:

```python
>>> ak.argsort(array)
<Array [4, 0, 1, 3, 2, 5] type='6 * int64'>
>>> array[ak.argsort(array)]
<Array [0, 1, 2, 3, None, None] type='6 * ?int64'>
>>> ak.sort(array)
<Array [0, 1, 2, 3, None, None] type='6 * ?int64'>
```

The two last items in the argsort, `2` and `5`, are the index positions of the two missing values in the `array`. You don't see the distinction when you do `array[ak.argsort(array)]` because whether you pick a `None` from the original array or you pass through a `None` in a slice with missing values, you get `None` in the output either way.

The reason people want to use argsort is to determine an ordering from one array (the result of computing a formula) and apply that ordering to another (may be complex objects). So, if there was some situation in which, say, `pt` had missing values but `eta` didn't, the (current) `?int64` output of argsort would differ from the (proposed) `int64` output in

```python
pt_order = ak.argsort(pt)
eta_sorted_by_pt = eta[pt_order]
```

With missing values in the argsort output, `eta_sorted_by_pt` would have missing values, which is strange because `eta` didn't. If the argsort output was strictly integers, some of which represent the positions of missing values, then `eta_sorted_by_pt` would not have missing values, just as `eta` does not.

Also, I think there's a strong expectation that argsort produces permutation arrays, and permutation arrays have no missing values. Even the integers with missing values removed from the (current) `?int64` behavior is not a permutation array.

As I understand it, the `Nones` are put into the output in a step that comes after descending into an option-type layout's `project()` result. For that, you only need to count how many `Nones` there should be and tack that many onto the end. What would it take to identify the locations of the `Nones` in the input (in order, a strictly monotonically increasing array of integers) and tack that onto the end instead?

tests/test_0074-argsort-and-sort.py
ianna: btw, NumPy handles it in a weird way:

```python
>>> a = np.array([[1,4],[3,1]])
>>> np.sort(a, axis=0)
array([[1, 1],
       [3, 4]])
>>> np.argsort(a, axis=0)
array([[0, 1],
       [1, 0]])
```
but
```python
>>> a[np.argsort(a, axis=0)]
array([[[1, 4],
        [3, 1]],

       [[3, 1],
        [1, 4]]])
```
The same in axis=-1:
```python
>>> np.sort(a, axis=-1)
array([[1, 4],
       [1, 3]])
>>> np.argsort(a, axis=-1)
array([[0, 1],
       [1, 0]])
>>> a[np.argsort(a, axis=-1)]
array([[[1, 4],
        [3, 1]],

       [[3, 1],
        [1, 4]]])
```

jpivarski:
I've linked this PR to issue #967 because merging this will resolve that. Was this PR intended to solve another issue, too? If so, we should use the "Linked issues" to ensure that the corresponding issues gets cleared when this is done. I find the closing keywords, such as "fixes #XYZ" to be pretty unreliable (I'll pick the wrong tense or something), but that "Linked issues" section is great.

ianna:
> I've linked this PR to issue #967 because merging this will resolve that. Was this PR intended to solve another issue, too? If so, we should use the "Linked issues" to ensure that the corresponding issues gets cleared when this is done. I find the closing keywords, such as "fixes #XYZ" to be pretty unreliable (I'll pick the wrong tense or something), but that "Linked issues" section is great.

Thanks! I think I never opened the issues this PR is trying to solve...

ianna:
@jpivarski - should this case be different, e.g. keeping `None`?
```python
    array = ak.Array([[3, 2, 1], [], None, [4, 5]])

    assert ak.to_list(ak.argsort(array, axis=1)) == [
        [2, 1, 0],
        [],
        None,
        [0, 1],
    ]
```
or should it be:
```python
   [
        [2, 1, 0],
        [],
        0,  # or 2?
        [0, 1],
    ]
```

jpivarski:
> @jpivarski - should this case be different, e.g. keeping `None`?
> 
> ```python
>     array = ak.Array([[3, 2, 1], [], None, [4, 5]])
> 
>     assert ak.to_list(ak.argsort(array, axis=1)) == [
>         [2, 1, 0],
>         [],
>         None,
>         [0, 1],
>     ]
> ```

The "None" is a level above the one where the sorting is happening, so yes, it stays.

For `axis=-1`, it seems pretty clear what sorting should do: the bottom-most lists get sorted (or argsorted, returning integer positions, even positions of Nones), but everything above this lowest level is left as-is. That's why this list-level None is fine.

ianna:
@jpivarski - ok, the `None`s are replaced by their indices in `argsort`. I was wondering if it should be an option for `argsort`? 

The `nan` values in an array would prevent correct sorting, so this is fixed as well. There are two (possibly one kernel function) to write - the `for` loops are marked with `FIXME`, but this is for tomorrow.

jpivarski:
@ianna I see in the tests that this is doing the right thing. This PR represents an important bug-fix (handling None at `axis=0` and in `argsort`), so it would be great to include it in 1.4.0, but the time for that is constrained by PyHEP. Perhaps this PR should go into 1.4.0 now (since it concerns user-visible changes) and the refactoring to put the inline loops into kernels can be done in another PR that gets merged after 1.4.0 (since it concerns important, but not user-visible, code organization)? Do you think it's in a stable state for that, notwithstanding the cleanup issues that need to eventually happen?

> I was wondering if it should be an option for `argsort`?

I think that would complicate the interface too much. It's not really a useful feature (someone could mask the output with `ak.is_none` to get the same effect).

jpivarski:
@ianna Let me know if I can merge this PR as-is and include it in 1.4.0. Thanks!

ianna:
> @ianna Let me know if I can merge this PR as-is and include it in 1.4.0. Thanks!

@jpivarski - if the last commit succeeds, that's all done. Please, merge it

jpivarski:
That was fast! I checked the diff, and it looks good, too. The kernel headers and ctypes should now be auto-generated, and if they're not, a failed test will tell us that. I'll enable auto-merge.

PR #948, opened 2021-06-23
pictures for a tutorial

ianna:
<strike>this one replaces https://github.com/scikit-hep/awkward-1.0/pull/926</strike>


ianna:
@jpivarski - I'm doe with this for today. Please, have a look. Thanks!

jpivarski (commented):
I can only give a general comment that these are like reference documentation, rather than tutorials. The reference docstrings for all the high-level functions focus on one function, what it does, and what its options and defaults are. That's what these new pages do for the reducers, but they should be broader (and less detailed) than that. A tutorial should instead focus on a problem (e.g. preparing data for plots), naming the functions that can be brought to bear on that problem, with examples, but without providing anything like a complete description of their options.

The [ak.sum](https://awkward-array.readthedocs.io/en/latest/_auto/ak.sum.html) reference documentation is already more tutorial-like in that it explains the general consequences of reducing with variable-length and missing data. Arguably, that should be taken out of the reference and put into this tutorial on reducers. The same is true of the [ak.behavior](https://awkward-array.readthedocs.io/en/latest/ak.behavior.html) reference, which is very tutorial-like, but it can't be moved because other websites already point to it (Coffea GitHub, for one).

The tutorials need to be from the data analysts' perspective. That's why there should be some extra discussion of `ak.any` and `ak.all`, because even though they're reducers, they're in a different mental category than the other reducers because they operate on booleans. When a data analyst is working through logic like, "I need any jet-muon pair to have ΔR < 0.2," that's mentally different from thinking, "I need to sum over photon energies," so we'd want some extra text to draw attention to the fact that formally, they're the same type of function.

The default `axis=None`, but that was chosen to agree with NumPy's functions of the same names. In most circumstances, users are going to want to set `axis=-1`, there should be a section drawing attention to that.

Also, there should be some explanation of what `axis=0` means in the context of arrays of lists of particles. The left-alignment of variable-length lists illustrated by your diagram implies that the most stably identifiable objects should be first in each nested list, and the tail of random junk should be last. If we had chosen to generalize `axis != -1` by right-alignment, the opposite would be true. The consequences of this left-alignment choice needs to be brought to the attention of data analysis users.

The argmin and argmax documentation is in a separate page from the other reducers because they're used in a special way: the output of argmin/argmax is almost always used as a slice, to select the "best" something-or-other from each list in an array of lists. That is, "find the particle in each event with the highest pT" or even "find the particle's eta in each event with the highest particle-pT." These common tasks require more than one step: computing the array of values to min/maximize, getting the array of integers from argmin/argmax, and then applying it as a slice. To do this, it is necessary to pass `keepdims=True`, which is not default (because of NumPy), in order to have the array of integers be at the right depth. Empty lists come out as `None`, though with `keepdims=True`, you might want to `ak.fill_none` them with empty lists. There's a lot to think about in this, and unlike [ak.sum](https://awkward-array.readthedocs.io/en/latest/_auto/ak.sum.html), the [ak.argmin](https://awkward-array.readthedocs.io/en/latest/_auto/ak.argmin.html) and [ak.argmax](https://awkward-array.readthedocs.io/en/latest/_auto/ak.argmax.html) reference documentation doesn't cover it.

I should also point out that the default reducer behavior for empty lists depends on the reducer: [ak.sum](https://awkward-array.readthedocs.io/en/latest/_auto/ak.sum.html) of empty lists results in `0`, [ak.prod](https://awkward-array.readthedocs.io/en/latest/_auto/ak.prod.html) of empty lists results in `1`, etc. It's only `None` for min/max/argmin/argmax because min/max are operations without an identity in number systems without infinity ("semigroups"), whereas the others, including `ak.any`/`ak.all`, have identities ("monoids").

I think this is the source of our disconnect about whether we can finish all of these tutorials before PyHEP. It generally takes me a day to write one page, to include all of these things I'm thinking that users might need to know. Writing a reference docstring is quite a bit faster, since that's like filling out a form: you have to say what the arguments mean, what the return value will be, and if there are any special cases—not how it will be used in the context of a data analysis.

Do you want to have another try at this? It's okay if you don't—the list of things that I'm looking for (above) are in my head and a "guess what I'm thinking" game isn't an efficient way to go about it. On the other hand, if you think you have an idea of what these are supposed to be like now and just weren't oriented before, go ahead!

ianna:
@jpivarski - thanks! You’ve got a point. The reference documentation is indeed more tutorial-like and since these pages refer to it I thought it is not necessary repeating it. It’s sort of like “ok, I’ve got an idea what it does and if I want to know more, I’ll check the reference guide”. I’ll give it another try if you don’t mind.

ianna:
@jpivarski - as discussed - only the pictures are left in this PR.

jpivarski (approved):
Thanks! (I just moved them to a different directory, that's all.)

Thanks also for the documentation fix in reducers.py.

I'll schedule this for auto-merging because it looks like there's nothing else that needs to be done.

PR #949, opened 2021-06-23
WIP: support _Arrow datasets

agoose77:


agoose77:
Let's close this given that we're astroturfing most of this code in v2

PR #952, opened 2021-06-23
Refactor: use `maybe_wrap` in source

agoose77:
This PR adds a new `maybe_wrap_like` to handle the altogether more common case that we want to `maybe_wrap` and take the behaviour from an existing array. 

@jpivarski you might feel this is starting to lead to creep of helper functions, but I think it's reasonable given that these two patterns are very common in the codebase.

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -449,6 +449,10 @@ def maybe_wrap(content, behavior, highlevel):
         return content
 
 
+def maybe_wrap_like(content, array, behavior, highlevel):
+    return maybe_wrap(content, behaviorof(array, behavior=behavior), highlevel)
+
+
 def extra(args, kwargs, defaults):
     out = []
     for i in range(len(defaults)):
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -162,10 +162,7 @@ def from_numpy(
             contents.append(recurse(array[name], mask))
         layout = ak.layout.RecordArray(contents, array.dtype.names)
 
-    if highlevel:
-        return ak._util.wrap(layout, behavior)
-    else:
-        return layout
+    return ak._util.maybe_wrap(layout, behavior, highlevel)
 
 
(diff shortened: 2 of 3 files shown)


jpivarski:
They're pretty common patterns. The `maybe_wrap_like` is only going to be applicable in cases with a single array argument and if the `behavior` is not needed (and extracted) earlier in the function, but the helper function creep doesn't bother me for these well-named functions.

I see helper functions as being a balance between avoiding repetitive (WET, non-DRY) code on the one hand versus the indirection of having to look up what all of these helper functions do on the other. There are cases when I'd actually prefer WET over DRY, and that is if the thing that you're Writing Every Time is short enough to be easily read as an "idiom" and it's so generic that the name is not immediately recognizable. Basically, the optimization metric is minimizing the time needed to understand a bit of code, not minimizing the number of characters in the code or any code-golf thing like that.

There's also been a lot more effort in beautifying the public programming interface that data analyst users will be interacting with every day in Jupyter than the internal machinations that we use to implement it, since we can improve the latter at any time (as you're doing right now).

agoose77:
> They're pretty common patterns. The maybe_wrap_like is only going to be applicable in cases with a single array argument and if the behavior is not needed (and extracted) earlier in the function, but the helper function creep doesn't bother me for these well-named functions.

I thought about making arrays support an iterable of Array or an Array, but I'm not sure whether that's just making things more complicated for a small number of cases.

> There's also been a lot more effort in beautifying the public programming interface that data analyst users will be interacting with every day in Jupyter than the internal machinations that we use to implement it, since we can improve the latter at any time (as you're doing right now).

I agree on these principles. For me, this function is beneficial more in that it reduces the likelihood of programming errors (though these kinds of constructs usually fail loudly), and should make it easier to read the code because this will be a frequently seen idiom. 

This PR is definitely take it or leave it - it doesn't add any new features ;)

jpivarski:
Does that mean it's done? Because I approve! Give me a checkmark and I'll enable auto-merge.

PR #953, opened 2021-06-23
tests: fix loading from any directory

henryiii:
The tests wouldn't run from inside the directory.


henryiii:
PS: As a safety feature, cibuildwheel runs from a different directory, so that doesn't pick up a local copy by accident. But the current form of the tests require only running from one directory (the repo directory). This enables them to be run anywhere.

jpivarski:
Only one test file assumed a directory? I'm surprised. But if the tests pass from a different directory, then that must have been it.

Thanks!

henryiii:
Note I am not running with any extras, like numba, etc, so this was just for the basic tests. (Adding a few extras might not be a bad idea, but we'll start with this).

PR #954, opened 2021-06-23
fix(setup): sync with cmake_example

henryiii:
The current version of the setup.py wasn't loading the correct MSVC reliably. I've synced this up with the work done on the `cmake_example`, which fixes this issue and is arguably a bit nicer with more docs.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -13,7 +13,6 @@ import setuptools.command.build_ext
 import setuptools.command.install
 from setuptools import setup, Extension
 
-
 if sys.version_info < (3,):
     # Note: never import from distutils before setuptools!
     from distutils.util import get_platform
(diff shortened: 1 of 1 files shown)


henryiii:
FYI, the current pybind/cmake_example was heavily influenced by awkward's setup.py, so it's coming full circle. :)

jpivarski (approved):
This looks good to me: adding more CMake environment variables and a mapping of platform names, presumably to build the more fine-grained matrix of wheels. The only thing that sticks out to me is a mixture of `+=` and `append` for accumulating arguments, but that's a minor stylistic thing. I'm relying heavily on the fact that the tests pass and that you definitely know what you're doing.

So I'd merge this as-is, and since your last commit was 19 hours ago (and you just asked me to review), I assume you're done and I'll merge it right away.

henryiii:
> The only thing that sticks out to me is a mixture of += and append

I had intended to make all the one-item additions an append, but got missed. Might try to clean that up later, then.

PR #955, opened 2021-06-23
fix: manylinux1 couldn't take None/newaxis

henryiii:
The current wheel building infrastructure does not test the wheels after building them. Manylinux1 wheels were broken due to a really weird bug - see https://github.com/pybind/pybind11/issues/3056 - that I don't understand yet. This should fix the wheels for now. The replacement wheel building infrastructure does test the wheels. :)


jpivarski:
Thanks!

PR #959, opened 2021-06-24
 C++ refactoring: _getitem_array implementation 

ioanaif:
- [x] `Content.__getitem__` in the right order
   - [x] `EmptyArray._getitem_array`
   - [x] `NumpyArray._getitem_array`
   - [x] `RegularArray._getitem_array`
   - [x] `ListArray._getitem_array`
   - [x] `ListOffsetArray._getitem_array`
   - [x] `RecordArray._getitem_array` handing the `allow_lazy` case
   - [x] `IndexedArray._getitem_array`
   - [x] `IndexedOptionArray._getitem_array`
   - [x] `ByteMaskedArray._getitem_array`
   - [x] `BitMaskedArray._getitem_array`
   - [x] `UnmaskedArray._getitem_array`
   - [x] `UnionArray._getitem_array`


jpivarski:
Copying from Slack, `Content.__getitem__` should be roughly:

   1. check int
   2. check slice
   3. check string
   4. check non-tuple iterable of strings
   5. ...
   6. check ak.Array, pull out its layout and call `__getitem__` on that so that it can be handled by rule 7 or 8
   7. check NumpyArray (to enter into `_getitem_array`)
   8. check generic Content (other cases, like jagged arrays and missing values)
   9. check tuple (for the generic `_getitem_next`)
   10. check remaining Iterable by calling nplike.`asarray` on it, wrapping it in NumpyArray and passing it into `__getitem__` again so that it gets picked up by rule 7 and goes into `_getitem_array`
   11. TypeError

_(roughly)_

jpivarski:
For testing, you can alternate between old-style and new-style arrays. The `ak._v2.tmp_for_testing` module defines `v1v2_equal`, which you can use to see if an old-style and a new-style array are exactly equal, and `v1_to_v2`/`v2_to_v1` for conversions.

When your project is done, we'll be removing the `ak._v2.tmp_for_testing` module as well as the old-style arrays, so while they're useful for ensuring that your new functions are compatible with the old ones, the final tests will have to be written in such a way that we can remove the old stuff and not lose the tests.

You can do this by developing with direct comparisons (`v1v2_equal`), then make the final tests look like

```python
result = your_v2_array.something()
assert ak.to_list(result) = [...]
assert result.form.tolist() == {...}
assert v1v2_equal(result, v2_to_v1(your_v2_array).something())
```

where you get `[...]` and `{...}` from the old-style array, and they're plain Python data. When we remove the old-style arrays, we can remove the last assertion, but the other two completely check the correctness of the array (values and detailed type).

jpivarski:
All Content subclass constructors (e.g. calling `ListArray`, `IndexedArray`, etc.) need to pass in the appropriate `identifiers` and `parameters`. We can't define `identifiers` yet, but I'll make utility functions to do the right thing when they're not all `None` (and pass through `None` for now). Passing through or creating empty `parameters` is very important, so the next PR should get started doing that.

Additionally, I'll make the very early (low numbered) tests start testing the new code soon.

This PR was in very good shape before I started messing with it. A few things were simplified by using kernels, rather than NumPy functions, which is something that we can do now, due to #966.

I'll try to get some underlying infrastructure done before you come back to this next week.

PR #960, opened 2021-06-24
ci: move to cibuildwheel on GHA

henryiii:
This PR moves the wheel building to cibuilldwheel, and performs it on GitHub Actions. One benefit of GitHub Actions is that we can keep this up-to-date with dependabot at some point; another is that the Azure 10 build queue and the GitHub Actions 10 build queue are not synced, so these can build in parallel with Azure ;). It's the same hardware and VMs, generally.

Features:
* Tests are now run on the wheels. This caught some bugs (#953, #954, #950, #955).
* SDists are build with pypa/build.
* Wheels can be directly downloaded from the GitHub Actions interface.
* Removed the manylinux1 32-bit wheels, which were segfaulting. Easy to re-add if the segfaults on running tests are fixed.
* Removed the CPython 3.10 builds - Python 3.10 will not have a stable ABI until 3.10.0b4, so PyPy wheels are not supposed to be uploaded yet. Cibuildwheel 2.0.0 alphas have 3.10 support protected by a flag CIBW_PRERELEASE_PYTHONS, so that users don't upload 3.10 wheels accidentally during the early beta period.
* Trimmed the overlapping manylinux wheels - it's now manylinux1 for 2.7, 3.5, and 3.6, and manylinux2010 for 3.6+. This is based on the Pip version analysis (performed using Hist ;) ) in https://github.com/pypa/manylinux/issues/994#issuecomment-792210874 
* Added Universal2 builds for Python 3.8 and 3.9 (TODO: I still need to test these on my Apple Silicon machine)
* Added Arm on Linux builds for 3.6+ (TODO: Should it be 3.7+? Are there old Arm users around? Due to emulation, these are _slow_. Travis can be used to get a faster native build, but not fond of messing with that. Building wheels should be pretty rare anyway.)
* Added PyPy builds for Linux (tested) and macOS (untested due to missing macOS wheels for NumPy)
* Builds manually triggered on GitHub, or on "Release" via GitHub (can be changed). TODO: Also on PR, so that this PR can build. Can be removed when the PR is ready.
* PyTest defaults to printing a summary.
* Removed almost 200 lines of code.

The CUDA build is currently untouched, still on Azure.


henryiii:
Ahh, actions don't run from a fork if not already in the repo. I can move the branch to the repo and try again.

PS: Unrelated, but the it seems GitHub has new link previews for cross-linked PRs. :)

PR #974, opened 2021-07-01
Bugfix: fix #973

### LINKED ISSUE #973: `ak.flatten` fails to flatten `ak.layout.NumpyArray` for `axis=None`
Completely flattening an n-D `ak.layout.NumpyArray` doesn't produce a flat layout:

```python3
>>> import awkward as ak
>>> import numpy as np
>>> ak.flatten(ak.from_numpy(np.zeros((3, 3, 5))), axis=None)
<Array [[[0, 0, 0, 0, 0], ... [0, 0, 0, 0, 0]]] type='3 * 3 * 5 * float64'>
```

agoose77:
Fix #973 by calling `array.reshape(-1)` before returning from `ak._util.completely_flatten`

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -571,9 +571,14 @@ def completely_flatten(array):
 
     elif isinstance(array, ak.layout.NumpyArray):
         if array.format.upper().startswith("M"):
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is correct and there's a test!

PR #976, opened 2021-07-01
Bugfix: support multidimensional NumPy mask arrays in `ak.mask`

### LINKED ISSUE #975: `ak.mask` fails for n-dim `NumpyArray` masks
The following example raises a `ValueError`
```python3

import awkward as ak
import numpy as np

array = ak.Array(
    ak.layout.RegularArray(ak.layout.NumpyArray(np.r_[1, 2, 3, 4, 5, 6, 7, 8, 9]), 3)
)
mask = ak.Array(ak.layout.NumpyArray(np.array([
    [True, True, True],
    [True, True, False],
    [True, False, True]
])))

ak.mask(array, mask)
```

agoose77:
Fixes #975

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -179,6 +179,7 @@ def mask(array, mask, valid_when=True, highlevel=True, behavior=None):
         [layoutarray, layoutmask],
         getfunction,
         behavior,
+        numpy_to_regular=True,
         right_broadcast=False,
         pass_depth=False,
     )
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Yes, that's all it needed. I'll try to get this into 1.4.0, too.

agoose77:
> Yes, that's all it needed. I'll try to get this into 1.4.0, too.

OK. I've modified the test to actually check the result of `ak.mask`, but it still passes.

PR #977, opened 2021-07-02
ArrayBuilder refactoring

ianna:
- [x] move all `*::snapshot` methods that depended on `Content` classes to a `builder_snapshot(const ak::Builder& builder)` function invoked in `content.cpp` to be handled via `pybind11` 
<strike>- [ ] move `ArrayBuilder::append` methods to be handled via `pybind11` </strike>
   as discussed, this one is needed for numba and will be revisited later
- [x] remove `RawArray` class
- [x] remove *.cpp tests
- [x] remove dependent project


ianna:
access to `ArrayBuilder::snapshot` from C++ causes failure, that is expected:
```c++
________________________________ test_producer _________________________________

    def test_producer():
        complicated_data = [1.1, 2.2, 3.3, [1, 2, 3], [], [4, 5], {"x": 12.3, "y": "wow"}]
    
>       assert ak.to_list(build.dependent.producer()) == complicated_data
E       ValueError: called obsolete 'UnionBuilder::snapshot'
E       
E       (https://github.com/scikit-hep/awkward-1.0/blob/1.4.0/src/libawkward/builder/UnionBuilder.cpp#L70)

test-python.py:13: ValueError

```

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -13,6 +14,8 @@
 
 #include "awkward/builder/ArrayBuilder.h"
 
+#include "awkward/refactoring.h"
```
ianna: All dependencies on arrays are moved out of the `ArrayBuilder` and other `Builder`s. In general, this file is needed only for a dependent project. The `refactoring.h` file contains the C++ equivalent of a `builder_snapshot` from `src/python/content.cpp`. The latter is used everywhere else.

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -45,32 +48,43 @@ namespace awkward {
 
   const ContentPtr
   ArrayBuilder::snapshot() const {
-    return builder_.get()->snapshot();
+    // FIXME: refactor
+    return ::builder_snapshot(builder());
```
ianna: this is called by a dependent project only.

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -45,32 +48,43 @@ namespace awkward {
...
   }
 
   const ContentPtr
   ArrayBuilder::getitem_at(int64_t at) const {
-    return snapshot().get()->getitem_at(at);
```
ianna, line 53: all `getitem`* are handled [here](https://github.com/scikit-hep/awkward-1.0/pull/977/files#diff-d5823d31f4669422f4f216ca1fc64471cb6f6993a30c08a9eb805ca019e9228fR1351-R1397)

src/libawkward/io/json.cpp
```diff
@@ -757,8 +758,9 @@ namespace awkward {
       return true;
     }
 
+    // FIXME: refactor
     const ContentPtr snapshot() const {
-      return builder_.snapshot();
+      return ::builder_snapshot(builder_.builder());
```
ianna, line 763: this `snapshot()` needs to be moved to python

ianna (commented):
@jpivarski - please, check if it's going in the right direction. Thanks.

src/libawkward/io/json.cpp
ianna: as discussed, rapidjson will be replaced by ijson

ianna:
@jpivarski - the last commit was a minor cleanup, so I assume the tests will pass. Please, have a look when you have time. Thanks!

jpivarski (approved):
Awesome! I scanned through everything and marvel at how many things used to refer to Content but didn't really need to (including in the documentation).

About deleting "dependent project," I had thought that we'd just disable it, but I like your idea of actually deleting it. It's best to be clear that the capability is going away.

I also like the idea of concentrating transitional code in include/awkward/refactoring.h, so that we can find it again when the transition is complete.

Let's merge it! You can do the honors (pushing the "enable squash-and-merge" button), in case you have any last changes.

PR #987, opened 2021-07-09
Feature: add GitHub Issue Forms

### LINKED ISSUE #986: Meta: use GitHub Issue Forms
GitHub now provides the ability to define Issue Forms, which establish a structured, validated form for creating Issues.

[Here](https://github.com/agoose77/test-issue-syntax/issues/new?assignees=&labels=bug+%28unverified%29&template=bug-report.yaml&title=Bug) is an example form for Awkward.

Do you think this would be a useful addition?

agoose77:
See [here](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) for more info

Here's an test repository with the same templates already configured:
https://github.com/agoose77/test-issue-syntax/issues/new/choose

Closes #986


agoose77:
OK, I've implemented equivalents of our existing issue forms using the GitHub Issue Forms schema. This is just a first draft!

jpivarski (approved):
Thanks for doing this!

I edited the text and structure of the forms; here's my mock-up: https://github.com/jpivarski/test-issue-syntax/issues/new/choose

In general, I'm minimizing the amount of explanatory text and reducing the structure to one textarea per form. The person filling out these forms is coming from some completely other task and having a lot of widgets on the screen is daunting.

If there's only one textarea, why use forms at all? Because the explanatory text can be outside the textarea, clearly separating the instructions we're giving them from the text they write. Even better: the instructions we're giving them don't disappear when they remove the original contents to start writing their text.

As a downside of the new forms, you can only drag & drop files into the textareas after you've clicked on them, so I've included a reminder in bold to click into the textarea first. It's also only possible to drop files or paste images if the textarea doesn't have a "render" field, so I've left that off.

It's cool that dedicated text fields for code to reproduce and log outputs will render the Python and tracebacks appropriately, but users are going to need more freedom to explain things, and one general-purpose textarea does that best. Some people are going to want to paste their error messages as screenshots because that's the only way they know how to do it. (I've included "Copy-pasted text is better than screenshots," but everyone is at a different skill level with respect to the terminal apps they're using.)

agoose77:
Could I make the case for restoring some of the structured inputs, but making them optional? As a user that is comfortable with making bug reports, these templates will improve my workflow by pre-defining a structure and adding the formatting etc as required. Additionally, by having these fields, our GitHub Issues will have a more consistent style to them.

I can see your point about users unfamiliar with creating Issues, but I wonder whether reducing the need for them to understand Markdown is a good thing here (by using rendered outputs)? So, in other words:

- Awkward version
- Description + any extra info (required)
- Code to reproduce (required), python
- Traceback (optional), pytb

Would this be a better compromise?

Merge as is if you remain unconvinced ;)

jpivarski:
(I've disabled auto-merge so that we can talk about this.)

Making them optional doesn't remove them from the screen, and I think the main issue is with a novice user **seeing something that looks complicated**. I think most people will be coming to this in the following context: "I ran into something that doesn't work; I'm going to be a good citizen and report it before I get back to my real work." If they see a bewildering array of widgets, that's going to slow them down, even if most of those widgets don't have a red star.

I can't emphasize enough how much of a step up these forms are in that they put our instructions to users outside of the text box where they write back to us. That's huge. The fact that the form also provides the ability to further structure the feedback would matter to big projects that get a lot of very similar requests, such as [PyPI's size limit requests](https://github.com/pypa/pypi-support/issues/new?assignees=&labels=limit+request&template=limit-request.yml&title=Limit+Request%3A+PROJECT_NAME+-+000+MB). For something like that, you really want "Project URL?" "Which limit?" "How big?", etc. to be in separate boxes because there are enough of them and they're routine enough that you'd want to automate some parts of that procedure. However, bugs in Awkward Array are idiosyncratic; each issue is addressed manually.

With separate boxes for "Code to reproduce?"  and "Traceback?", people would have a hard time explaining problems that involve more than one code block, e.g. "If I do _this_, I get _that_, but if I do _this other thing_, I get _that other thing_. It would also make it hard to explain issues that are not error messages but wrong answers (or suspected wrong answers). For an issue-filter who wants to get back to the physics project they were working on, having to decide what to do with a square peg and a round hole is extra hassle.

agoose77:
I can't really argue against your concern for adding too much structure. It's a shame that we can't mark the template as optional. 

I haven't changed my mind on the benefits of having a standard template, but we can always change it down the line as this isn't an API that we have to maintain :) let's merge your changes and revisit this again in future if needs be!

jpivarski:
Okay, sounds good. Let's start with this, as we can always change it, adding more structure, in the future.

PR #985, opened 2021-07-08
Feature: add `np.ravel`

agoose77:
As discussed in #984, this PR adds a new simplified `ak.ravel` counterpart to `np.ravel` for Awkward Arrays. This retains the record-losing behaviour of `ak.flatten(axis=None)`, and permits us to schedule the removal of this quirk from `ak.flatten` after a deprecation cycle. In the meantime, a new `axis="records"` axis will be added to `ak.flatten` to opt-in to this change.

```python3
>>> np.ravel(ak.Array([1, 2, [3, 4]]))
ak.Array([1, 2, 3, 4])
```

Implements NumPy `ak.ravel` with `ak.flatten(..., axis=None)` behaviour (closes #989)

Issue #984, opened 2021-07-08
Add record-preserving for behaviour to `ak.flatten`


agoose77:
As discussed in the [user docs](https://awkward-array.org/how-to-restructure-flatten.html#ak-flatten-with-axis-none), there is currently no straightforward way for a user to completely flatten an array without losing its record structure, e.g.

```python3
>>> array = ak.Array([{'x': 1, 'y': 'hi'}])
>>> ak.flatten(array, axis=None)
<Array [1, 104, 105] type='3 * int64'>
```

@jpivarski mentioned on Gitter the idea of adding a new `axis="records"` case to `ak.flatten` which preserves (nested) record fields.

jpivarski:
@agoose77 I changed my mind. This does not have to be a `next-release` issue. We can leave it for the far future.

PR #990, opened 2021-07-11
Feature: add `ak.flatten(axis="records")`

agoose77:
Building upon #985, this PR adds a new `axis="records"` axis will be added to `ak.flatten` to opt-in to this change.

- [ ] Add `axis="records"` to `ak.flatten`
- [ ] Add deprecation note to `axis=None` argument 
   The meaning of `axis=None` will change to *preserving* records.

Implements #984

### DIFF
--- a/docs-src/how-to-restructure-flatten.md
+++ b/docs-src/how-to-restructure-flatten.md
@@ -31,7 +31,9 @@ To destructure an array for plotting, you'll want to
    * remove record structures,
    * remove missing data
 
-There is a function that does all of these things in one call, [ak.flatten](https://awkward-array.readthedocs.io/en/latest/_auto/ak.flatten.html) with `axis=None`, but you don't want to apply that without thinking because structure is important to the meaning of your data and you want to be able to interpret the plot. Destructuring is an information-losing operation, so your guidance is required to eliminate exactly the structure you want to eliminate, and there are several ways to do that, depending on what you want to do.
+There is a function that does all of these things in one call, [ak.ravel](https://awkward-array.readthedocs.io/en/latest/_auto/ak.ravel.html), but you don't want to apply that without thinking because structure is important to the meaning of your data and you want to be able to interpret the plot. Destructuring is an information-losing operation, so your guidance is required to eliminate exactly the structure you want to eliminate, and there are several ways to do that, depending on what you want to do.
+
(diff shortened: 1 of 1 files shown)


jpivarski:
Since `ak.ravel` exists, we can go ahead with changing the `ak.flatten` with `axis=None` behavior to preserve records. For reasons described below, we can go straight to `axis=None`, without the intermediary `axis="records"`. (That would represent two deprecations, one to warn about `axis=None` and the other to remove `axis="records"`, and that's too complicated for something that can be done without a deprecation cycle, for reasons described below.)

Documentation will have to change at the same time as the feature itself (i.e. in this PR). `ak.flatten` with `axis=None` is mentioned in:

   * the `ak.run_lengths` docs, but what's said here will remain valid with the new behavior
   * the `ak.flatten` docs: _needs updates_
   * the `how-to-restructure-flatten.md` tutorial: _needs updates_
   * the `how-to-restructure-pad.md` tutorial: _needs updates_
   * notebook documentation in docs-jupyter, but that's dated (2020) and doesn't need to be updated

Since the new behavior would use an interface that currently has another meaning, there can't be a gradual deprecation cycle. However, this would break things in a way that draws attention to itself. If someone had been using `axis=None` without records, nothing changes, but if they had records, then instead of squashing all numbers into a one-dimensional NumPy-like array, they'll get a record, whose uses are more restricted than a one-dimensional array, so it will show up as an exception. If it shows up in an exception, the user would be able to investigate and find out that the behavior has changed (rather than just getting wrong answers or something).

The breaking change has to go in at a [minor version boundary](https://github.com/scikit-hep/awkward-1.0#roadmap), though we're mostly only doing minor version boundaries now. If we have any reason to release a 1.4.x, we'll have to disable the change.

jpivarski:
This is the oldest PR, it's about a policy change, and it would be hard to bring up-to-date, even without #1244.

Is `ak.flatten(axis="records")` obsolete, now that you've implemented `ak.ravel`? Did that fill the need that this would have filled?

It's been about a half year since we talked about this. Can we close the PR?

agoose77:
@jpivarski this is an old PR, and given my limited cycles right now I think it is a good idea to close it. This change was rather to change flatten now that ravel exists. I mostly want to drop all list structure whilst preserving records, which is currently a non-trivial thing to do. We can discuss it again when I make a new pr (with implementation!) :)

jpivarski:
Sounds good. There's a lot to focus on!

PR #991, opened 2021-07-12
Fixed bug #770, #930: not a policy issue; UnionForm::purelist_parameter was incorrectly comparing its contents' direct parameters, rather than their purelist_parameters.

### LINKED ISSUE #770: Union of indexed arrays loses its record type
If I define a simple mixin class and create a union array that can be an instance:
```python
import awkward as ak


@ak.mixin_class(ak.behavior)
class Blah:
    @property
    def blah(self):
        return self["x"]


a = ak.zip({"x": [[1, 2], [3]], "y": [[4, 5], [6]]})
b = ak.zip({"x": [[-1, -2, -3], [-4]], "z": [[-4, -5, -6], [-7]]})
c = ak.with_name(ak.concatenate([a, b], axis=1), "Blah")
```
then, as expected, `c.blah` returns `<Array [[1, 2, -1, -2, -3], [3, -4]] type='2 * var * union[int64, int64]'>` (it would be nice if it simplified, that's #714).
But if I first filter `a` and `b` (i.e. turning them into indexed arrays):
```python
a2 = a[a.x % 2 == 0]
b2 = b[b.x % 2 == 0]
c2 = ak.with_name(ak.concatenate([a2, b2], axis=1), "Blah")
```
the same `c2.blah` now raises an `AttributeError`. One also sees that `c2.layout.purelist_parameter("__record__")` is empty.

With ak version `1.2.0rc2`.

### LINKED ISSUE #930: UnionArrays must not hide a `__record__` parameter when all contents have the same `__record__` value
Maybe the new rule could be,

   1. if the UnionArray has a parameter with a given name, `purelist_parameter(name)` should return its value;
   2. if the UnionArray does not have this parameter, but every one of its contents has that parameter with the same value, then `purelist_parameter(none)` should return that common value.

None of the error cases that raised this issue are still errors, though. In all of those cases, it was wrong for the UnionArray to have existed: a `simplifiy_uniontype` call was missing. The above rule would make UnionArrays transparent to common parameters, but maybe that's bad because then we'd be less likely to catch missing `simplifiy_uniontype` calls, since that's a common symptom.

Are there any legitimate cases in which you'd have a UnionArray with all contents having the same parameter? I can think of one corner-case: you have an array of records that are all Lorentz vectors (`"__record__": "Momentum4D"`), but some of them have auxiliary data, like `"charge"` because they're electrons, while others don't. Then you'd want the record name to be visible above the UnionArray.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward-1.0/issues/770#issuecomment-861867937_

jpivarski:


jpivarski:
See https://github.com/scikit-hep/awkward-1.0/issues/770#issuecomment-878630514 for an explanation.

PR #1003, opened 2021-07-15
chore: fix spelling and check in pre-commit

henryiii:
Adds a spelling check, and I fixed (or added to the ignore file) everything it found.

Also added the miscapitalization check. I had to remove checking for "Numpy" instead of "NumPy" automatically, because quite a few things have "Numpy" burned into class names - I couldn't even check for "Numpy\b", since one class is named Numpy without anything else. I still manually fixed the ones I saw in comments.


jpivarski (approved):
It's especially helpful for comments and documentation. (There might be more in the docs* directories.)

Although it happens to be the case that I only found out that NumPy is supposed to have two capital letters after "NumpyArray" was defined, I intentionally kept the class name that way because "NumPyArray" is too bumpy; it would be easy to forget. In that name, "Numpy" is thought of as one part and "Array" is another: more nested structures would be a distraction. (Note "ByteMaskedArray" and "BitMaskedArray" vs "UnmaskedArray". All of those capitalizations were intentional.)

On the other hand, I can be persuaded to rename `nplike.NumPy` and `nplike.CuPy` (both of them together). There isn't a connection between these and `ak.layout.NumpyArray`; the different capitalization schemes wouldn't mentally conflict. I can make that part of this PR so that you can check for "Numpy\b". They're always fully qualified:

```bash
% egrep -r 'Numpy\b' src
src/awkward/__init__.py:    raise ImportError("Numpy 1.13.1 or later required")
src/awkward/_connect/_numba/builder.py:numpy = ak.nplike.Numpy.instance()
src/awkward/_connect/_numba/layout.py:numpy = ak.nplike.Numpy.instance()
src/awkward/_connect/_jax/jax_utils.py:numpy = ak.nplike.Numpy.instance()
src/awkward/_connect/_autograd.py:numpy = ak.nplike.Numpy.instance()
src/awkward/partition.py:numpy = ak.nplike.Numpy.instance()
src/awkward/nplike.py:        return Numpy.instance()
src/awkward/nplike.py:class Numpy(NumpyLike):
src/awkward/operations/describe.py:numpy = ak.nplike.Numpy.instance()
src/awkward/operations/convert.py:numpy = ak.nplike.Numpy.instance()
src/awkward/highlevel.py:numpy = ak.nplike.Numpy.instance()
src/libawkward/array/NumpyArray.cpp:        std::string("Numpy format \"") + format_
src/libawkward/array/NumpyArray.cpp:            std::string("cannot convert Numpy format \"") + format_
% egrep -r 'Cupy\b' src
src/awkward/nplike.py:        return Cupy.instance()
src/awkward/nplike.py:class Cupy(NumpyLike):
src/awkward/operations/convert.py:    cupy = ak.nplike.Cupy.instance()
```

jpivarski:
I don't remember what I thought I was waiting for, but whatever it was, it's better to get this merged now than let it get more out of date.

PR #1004, opened 2021-07-15
docs: touch up contributing

henryiii:
This corrects the outdated commend about formatting and linting.

It also removes something I found highly irritating - I'm sure you know what it
is :). Using new features in a language or a library does _not_ make your code
"less-simple" - it makes it harder to read and more complex because instead of
using a new feature that has a clear meaning that everyone can learn, it's
adding code unique to your library only every reader must parse to see what you
are doing. For example, take a few items from contextlib (which has had
fantastic improvements):

```python
with contextlib.redirect_stdout(sys.stderr):
    should_print_to_stderr_but_doesnt()

with contextlib.suppress(FileNotFoundError):
    os.remove('somefile.tmp')

with contextlib.ExitStack() as stack:
    stack.enter_context(context_manager_1())
    stack.enter_context(context_manager_2())
    stack.enter_context(context_manager_3())
```

I think you can read what those do without looking at the docs, and if you
can't, everyone knows how to get to the Python docs; finding local
documentation or reading the expanded try/catch or manually implemented
versions is much harder for readers. And it's easy to incorrectly implement;
for example, if `should_print_to_stderr_but_doesnt` throws an error, a hand
implementation might not restore the streams, causing important information
about what happened to be swallowed.

Also, why should someone upgrade awkward, then? Aren't older versions
"simpler", by this logic?

If/when you are ready to drop Python 2, remember that Python 2 _will_ keep
working! It will just pull older versions of awkward / uproot (as it already
does with NumPy, boost-histogram, and everything else). Binary distribution
will become much harder quite rapidly for Python 2 - pip, packaging, wheel, and
auditwheel have already dropped support, manylinux1 (the only remaining place
with Python 2) dies Jan 1, 2022, pybind11 drops support roughly Jan 1, 2022,
cibuildwheel drops it tomorrow, etc.

(PS: I originally did this with the GitHub interface and it committed to main
accidentally - immediately reverted, and making a proper PR. Huge apologies!).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -235,7 +235,7 @@ We adhere to an 80-character line width, which is a [standard in the industry](h
 
 Unit tests do not need to adhere to the 80-character limit.
 
-we use several linters and reformatters; all are supported through a single interface with the [pre-commit](https://pre-commit.com) tool. Automatic reformatting will be applied to PRs automatically using [pre-commit.ci](https://pre-commit.ci), though running pre-commit locally is still recommended.
+We use several linters and reformatters; all are supported through a single interface with the [pre-commit](https://pre-commit.com) tool. Automatic reformatting will be applied to PRs automatically using [pre-commit.ci](https://pre-commit.ci), though running pre-commit locally is still recommended.
 
 ### Fully qualified names
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -235,9 +235,7 @@ We adhere to an 80-character line width, which is a [standard in the industry](h
 
 Unit tests do not need to adhere to the 80-character limit.
 
-We don't, however, use automated formatters like clang-format or black. Automated formatting can rearrange code into something that is, paradoxically, harder to read, undermining the human element in programming.
-
-We don't currently use linters like clang-tidy or flake8, but I'm open to it.
+we use several linters and reformatters; all are supported through a single interface with the [pre-commit](https://pre-commit.com) tool. Automatic reformatting will be applied to PRs automatically using [pre-commit.ci](https://pre-commit.ci), though running pre-commit locally is still recommended.
 
 ### Fully qualified names
 
@@ -275,7 +273,7 @@ We only supprt Python 2.7 as much as is practical. For example, we ignore compil
 
 Python 2.7 and 3.5 have unstable dict order, which excludes them both from some tests.
 
-Allowing Python 2.7 after it has been [fully and finally discontinued](https://mail.python.org/archives/list/python-dev@python.org/message/OFCIETIXLX34X7FVK5B5WPZH22HXV342/) may sound anachronistic, but some of our users depend on legacy codebases forcing them in Python 2, and Python 2 limits our adoption of new language features, keeping the code "simple, rather than smart."
+Allowing Python 2.7 after it has been [fully and finally discontinued](https://mail.python.org/archives/list/python-dev@python.org/message/OFCIETIXLX34X7FVK5B5WPZH22HXV342/) may sound anachronistic, but some of our users depend on legacy codebases forcing them in Python 2, so we support that for now.
 
 ### Third party dependencies
 


CONTRIBUTING.md
```diff
@@ -235,9 +235,7 @@ We adhere to an 80-character line width, which is a [standard in the industry](h
 
 Unit tests do not need to adhere to the 80-character limit.
 
-We don't, however, use automated formatters like clang-format or black. Automated formatting can rearrange code into something that is, paradoxically, harder to read, undermining the human element in programming.
-
-We don't currently use linters like clang-tidy or flake8, but I'm open to it.
+we use several linters and reformatters; all are supported through a single interface with the [pre-commit](https://pre-commit.com) tool. Automatic reformatting will be applied to PRs automatically using [pre-commit.ci](https://pre-commit.ci), though running pre-commit locally is still recommended.
```
henryiii:
```suggestion
We use several linters and reformatters; all are supported through a single interface with the [pre-commit](https://pre-commit.com) tool. Automatic reformatting will be applied to PRs automatically using [pre-commit.ci](https://pre-commit.ci), though running pre-commit locally is still recommended.
```

jpivarski (approved):
The change in wording is fine. The "simple, rather than smart" text is inappropriate, since code that supports Python 2 doesn't look simpler.

On the larger topic of supporting Python 2, I understand your statement that "Dropping support for Python 2 doesn't mean that people won't be able to use Awkward Array with Python 2; they just won't be able to use the latest version of Awkward Array with Python 2." However, it's that last part that matters to me until the major physics collaborations are out of the Python 2 jungle. This is where we've misunderstood each other before: I meant "latest Awkward working in Python 2" while you meant "some version of Awkward working in Python 2," that was the disconnect.

As for the argument, "if they don't want Python 3, why would they want the latest Awkward features?" (i.e. if luddite in one area, why not others?), it's because the constraints on which Python you use are decoupled from the need for library features—they're separate concerns and having one restrict the other adds constraints where they didn't need to be. In fact, I use this personally: my OS, Gimp, and Inkscape come from LTS Ubuntu, which gets very old, but Python libraries come from Conda so that I can use the latest of these, because that's where it matters. The two kinds of software needs are decoupled. So that was the reason I wanted to keep Python 2 support, to simplify the life of some poor maintainer of a locked-down DAQ or something.

But now that the physics collaborations really are getting out of the Python 2 jungle (CMS is :tada:), and since manylinux1 dies on Jan 1, 2022, maybe that can be our Python 2 kill date. I think we can advertise, loudly, that Awkward Array won't support Python 2 past Jan 1, 2022. After that, old versions of Awkward Array will be available but no one will get bug-fixes or new features if they're using Python 2.

It might also be possible to split the build infrastructure so that updating it and moving it to GitHub Actions can be Python 3 only, while the unchanged build procedure for Python 2 still hobbles along in Azure until we kill it at the end of the year. To the extent that both go through some of the same files, such as setup.py, that could be complicated, but if adding cibuildwheel involves new files, it could be simpler to set them up for Python 3 only (especially since cibuildwheel support for Python 2 ends tomorrow!).

Maybe even Jan 1, 2022 is too far in the future; Dec 1, 2021 would be just as good and might avoid any edge effects. Dec 5, 2020 was when the awkward1 → awkward name change happened; a notorious anniversary. Do we know the Python 3 readiness of all the major LHC experiments?

   - [X] CMS
   - [ ] ATLAS
   - [ ] LHCb
   - [ ] ALICE

Smaller collaborations have more flexibility and are _more_ likely to be Python 3 ready than the above. If all of these collaborations have mainline (stable released) Python 3 support, then maybe we could pick an earlier month, but still far enough in the future that it's not dropped as a surprise. I'll note it on the [release history](https://github.com/scikit-hep/awkward-1.0#roadmap) section on the README.

jpivarski:
I've converted this into a Discussion: https://github.com/scikit-hep/awkward-1.0/discussions/1010 (with a curiously binary number).

PR #1005, opened 2021-07-15
Fixes #595 and #630; adds a default for NEP-18.

### LINKED ISSUE #595: Define wrappers for pure NumPy functions that don't automatically call '__array__'
We can override any NumPy function using our own internal `@ak._connect._numpy.implements` decorator, and this has been used to give NumPy functions Awkward generalizations. However, there are still a lot of NumPy functions that fail for Awkward Arrays, even flat ones, because NumPy isn't calling calling `__array__` on the `ak.Array`. Some NumPy functions do, others don't.

To make this more predictable, someone (me?) should do a survey of all functions in the `numpy` namespace find the ones that don't automatically call `__array__`, and define them in a submodule of their own, like this:

```python
@ak._connect._numpy.implements("whatever")
def whatever(array_arg, another_array_arg, non_array_arg):
    nplike = ak.nplike.of(array_arg, another_array_arg)._module
    return nplike.whatever(nplike.asarray(array_arg), nplike.asarray(another_array_arg), non_array_arg)
```

This could be done more systematically with `*args` and `**kwargs`, but we still need to know which arguments are supposed to be arrays to pass them through `nplike.asarray`. (Here, I'm using the direct `._module`, rather than going through the full [NumpyLike](https://github.com/scikit-hep/awkward-1.0/blob/68087850eabafd39f0c0de37ed96487fa44cc69f/src/awkward/nplike.py#L120) interface because I don't want to verify NumPy-CuPy agreement/lack-of-agreement for every one of these. The `nplike.asarray` function raises an exception if the Awkward Array isn't a rectilinear, numerical array, so this isn't generalizing NumPy's behavior so much as making Awkward-brand arrays usable as NumPy arrays in exactly the cases that NumPy itself covers (except `dtype="O"` and strings).

These functions will _not_ be used in the main Awkward codebase, they will _not_ be exposed to the top-level `awkward` namespace, and they will _not_ be documented on ReadTheDocs. Among them are some functions that we might want to generalize someday, and that would mean moving them from this submodule to one of the standard places and possibly narrowing the argument list.

This submodule should probably be called `ak._connect._numpy_compatibility` or something. It should be imported by `__init__.py` after `ak._connect._numpy`.

### LINKED ISSUE #630: Default __array_function__ behavior
I wonder if it is reasonable to attempt calling `ak.to_numpy` on each awkward array passed to a NEP18 `__array_function__` that doesn't have an explicit override already, and if the awkward array is not expressible in numpy then display a suitably descriptive error like "no implementation found and the argument <Array ...> was not coercible to a numpy array"

This would avoid errors such as:
```
TypeError: no implementation found for 'numpy.histogram' on types that implement __array_function__: [<class 'awkward.highlevel.Array'>, <class 'numpy.ndarray'>]
```
when in reality the awkward array may well be compatible with the NEP18 function.

jpivarski:


jpivarski:
@nsmith- This one had been waiting too long.

PR #1008, opened 2021-07-15
Bugfix: fix `ak.packed` for `RegularArray`s with `.size=0`

### LINKED ISSUE #1006: `ak.packed` fails for `RegularArray` with size 0
### Version of Awkward Array

1.4.0

Given this code
```python3
ak.packed(ak.from_numpy(np.zeros((1, 0), dtype=np.int32), regulararray=True))
```
`ak.packed` raises `ZeroDivisionError` when dividing the length of the content by the size.

agoose77:
This also catches some places where the `identities` and `parameters` were not retained from the original layout

Fixes #1006

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2174,7 +2174,10 @@ def _pack_layout(layout):
         new_index[~is_none] = nplike.arange(len(new_index) - nplike.sum(is_none))
 
         return ak.layout.IndexedOptionArray64(
-            ak.layout.Index64(new_index), layout.project()
+            ak.layout.Index64(new_index),
+            layout.project(),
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This looks right.

Keeping identities and parameters on all node types is also important (though unrelated to the bug-fix, might as well fix them now).

agoose77:
> This looks right.
> 
> Keeping identities and parameters on all node types is also important (though unrelated to the bug-fix, might as well fix them now).

My thoughts too. Not sure how that snuck through the original commit; originally I ensured these were all preserved.

agoose77:
I'm done with this PR now! :rocket:

PR #1009, opened 2021-07-15
Bugfix: support empty buffers in `from_buffers`

### LINKED ISSUE #1007: `ak.from_buffers` fails for n-dim `NumpyArray` with size 0
### Version of Awkward Array

1.4.0

The following code
```python3
>>> array = ak.from_numpy(np.zeros((1, 0), dtype=np.int32))
>>> ak.from_buffers(*ak.to_buffers(array))
```
fails with a `ZXeroDivisionError`:
```pytb
ZeroDivisionError                         Traceback (most recent call last)
/tmp/ipykernel_10330/380633746.py in <module>
----> 1 ak.from_buffers(*ak.to_buffers(array))

/opt/texat-venv/lib/python3.9/site-packages/awkward/operations/convert.py in from_buffers(form, length, container, partition_start, key_format, lazy, lazy_cache, lazy_cache_key, highlevel, behavior)
   5055 
   5056         else:
-> 5057             out = _form_to_layout(*(args + (None, None)))
   5058 
   5059     elif isinstance(length, Iterable):

/opt/texat-venv/lib/python3.9/site-packages/awkward/operations/convert.py in _form_to_layout(form, container, partnum, key_format, length, lazy_cache, lazy_cache_key)
   4712 
   4713         if length is not None:
-> 4714             actual = len(raw_array) // dtype_inner_shape.itemsize
   4715             if length > actual:
   4716                 raise ValueError(

ZeroDivisionError: integer division or modulo by zero
```

agoose77:
`ak.from_buffers` fails to reconstruct an array when the `NumpyForm` has a zero itemsize. This PR fixes this by only performing a length check if itemsize` is nonzero, and taking the leading dimension from the given `length`.

Fixes #1007

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -4710,7 +4710,9 @@ def _form_to_layout(
         else:
             dtype, inner_shape = dtype_inner_shape.subdtype
 
-        if length is not None:
+        # If the buffer is required to have a length, check that it's not too short.
+        # If the array has no itemsize, then the buffer can never be "too short"!
+        if (length is not None) and (dtype_inner_shape.itemsize > 0):
             actual = len(raw_array) // dtype_inner_shape.itemsize
             if length > actual:
                 raise ValueError(
@@ -4719,7 +4721,21 @@ def _form_to_layout(
                     + ak._util.exception_suffix(__file__)
                 )
 
-        array = raw_array.view(dtype).reshape((-1,) + inner_shape)
+        # NumPy can only infer the length of the array from the buffer
+        # if the inner shape has a nonzero item-size.
+        if dtype_inner_shape.itemsize:
+            leading_shape = (-1,)
+        elif length is not None:
+            leading_shape = (length,)
+        # Given that `length=None` only occurs when reading legacy data
+        # This already failed, so we can fail if `length` is not given
+        else:
+            raise ValueError(
+                "buffer is empty, and no length was given. It is not possible to "
+                "determine the correct size of the array from the buffer alone."
+            )
+
+        array = raw_array.view(dtype).reshape(leading_shape + inner_shape)
 
         return ak.layout.NumpyArray(array, identities, parameters)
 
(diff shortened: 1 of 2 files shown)


jpivarski:
In principle, the NumpyForm constructor should complain about `itemsize <= 0`, or even any `itemsize` that disagrees with the data type. (There are a lot of redundancies in v1 NumpyForm/NumpyArray's dtype handling, in large part because I discovered platform dependencies in the Python buffer's `format` after development was well underway.) The v2 constructor and `from_json` don't ask for the `itemsize`, since it can be derived from dtype.

Form JSON with `"itemsize": 0` is not legal, though I suppose the check against it hadn't been implemented.

agoose77:
> In principle, the NumpyForm constructor should complain about `itemsize <= 0`, or even any `itemsize` that disagrees with the data type. (There are a lot of redundancies in v1 NumpyForm/NumpyArray's dtype handling, in large part because I discovered platform dependencies in the Python buffer's `format` after development was well underway.) The v2 constructor and `from_json` don't ask for the `itemsize`, since it can be derived from dtype.
> 
> Form JSON with `"itemsize": 0` is not legal, though I suppose the check against it hadn't been implemented.

Ah, I'm not being clear enough in the discussion here - `itemsize` is referring to the `dtype.itemsize`, which counts the size of the inner sub array. That *can* be zero, if any of the inner dims are zero.

jpivarski:
If we're talking about a dtype's `itemsize` because some element in `dtype.shape` is zero, that's supposed to be captured in the `inner_shape`. (The `inner_shape` is exactly the `dtype.shape`; in other words, the array's `shape[1:]`.)

A zero could be hidden in the `inner_shape` and could lead to an untested error. It seems you've found one, then.

agoose77:
Yes, this is what I'm talking about. Given that it's legal for NumPy arrays to have such a dtype (or at least, be described by one), the only error here is that we currently implicitly expect the provided buffer to have non-zero length. This assumption fails in the case that `to_buffers` was called with something like `np.zeros((3, 0))`.

agoose77:
I think this is ready to test :wrench: and ship :rocket:, unless you have any style concerns.

jpivarski:
It's totally legal for a NumPy array to have zero size due to a zero-length dimension hidden somewhere in its shape.

Since you've given me the go-ahead and tests pass, I'll merge this now.

PR #1011, opened 2021-07-17
chore: bump pybind11 to 2.8.0

henryiii:


henryiii:
Okay, this one's ready too. pybind11 2.7+ is pickier about being explicit with the tuple constructor. On 32 bit Windows `py::ssize_t` is a 32 bit integer.

jpivarski:
> Okay, this one's ready too. pybind11 2.7+ is pickier about being explicit with the tuple constructor. On 32 bit Windows `py::ssize_t` is a 32 bit float.

32-bit _float_? Integer, right?

Thanks for this as well. I'll enable auto-merge, now that the required tests have been updated (and I see that they're right in the list above).

jpivarski:
The requirement to do 3.10 must have crossed in the mail. The previous run of the test didn't have them. It's up to date now; something just had to get flushed.

henryiii:
Yes, I meant integer. 🙃

Restarted the CI, it knows it needs the new checks but they need to run.

PR #1013, opened 2021-07-19
Bugfix: do not use `regular_to_jagged` in `ak.zip`

### LINKED ISSUE #1012: `ak.zip` converts `RegularArray`s to `ListOffsetArray64`s
### Version of Awkward Array

1.4.0

Here is a reproducer:
```python3
import awkward as ak
import numpy as np

parameters = ak.from_numpy(np.arange(8 * 8).reshape(-1, 8), regulararray=True)
array = ak.zip((parameters, parameters))
assert array.layout.purelist_isregular
```

This is caused by the change to `ak.zip` in 585ca1a0d0c6f8d8473c3f177f7ccf8546cea9a6

agoose77:
This fixes #1012 which was introduced by 585ca1a0d0c6f8d8473c3f177f7ccf8546cea9a6. Instead, we just use `right_broadcast=False`, and allow the user to override this value with a parameter.

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -477,6 +477,7 @@ def zip(
     with_name=None,
     highlevel=True,
     behavior=None,
+    right_broadcast=False,
 ):
     """
     Args:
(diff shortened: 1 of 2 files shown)


agoose77:
@jpivarski once tests pass, this is good on my end!

jpivarski (approved):

PR #1029, opened 2021-07-21
Fixed #1026; jagged slicing of multidim NumpyArray.

### LINKED ISSUE #1026: `NumpyArray::getitem_next_jagged` is undefined for `ndim == 2`
### Version of Awkward Array

1.4.0

I think we should support this given that it works if the `NumpyArray` is converted into a `RegularArray`:
```python3
array = ak.Array(
    ak.layout.ListOffsetArray64(
        ak.layout.Index64(np.r_[0, 4, 4, 6]),
        ak.layout.NumpyArray(np.arange(4 * 6).reshape(6, 4)),
    )
)
ix = ak.Array([
    [[0], [1], [2], [3]],
    [],
    [[3], [2]]
])
```
This fails:
```python3
array[ix]
```

This succeeds:
```python3
ak.to_regular(array, axis=-1)[ix]
```

I would expect the result of jagged indexing into a `NumpyArray` to produce a jagged result (i.e. behave as it currently does for `RegularArray`).

The simplest solution would be to re-use the `RegularArray` code here, i.e. `toRegularArray()->getitem_XXX` but I don't know what the protocol on this is vs doing less work by implementing a solution specifically for NumpyArrays. I suspect here it's not unreasonable because the conversion to `RegularArray` is cheap (?)

jpivarski:


agoose77:
@jpivarski I (hopefully) fixed the test for windows, but disabled auto-merge so that you'll get the final sign off.

jpivarski (commented):
Thanks!

PR #1031, opened 2021-07-22
C++ refactoring: starting _getitem_next.

jpivarski:
This PR implements all types of slices allowed by NumPy on Awkward Arrays. Awkward's extensions of NumPy's slicing rules will be covered in a future PR.

In principle, there are 12 classes to cover: BitMaskedArray, ByteMaskedArray, EmptyArray, IndexedArray, IndexedOptionArray, ListArray, ListOffsetArray, NumpyArray, RecordArray, RegularArray, UnionArray, and UnmaskedArray with 10 types of slice items: no-slice (`head == ()`), integer (`int`), range-slice (`slice`), field (`str`), fields (`list(str)`), `np.newaxis` (which is `None`), `Ellipsis` (`...`), flat array (`Index64`), jagged array (`ListOffsetArray`), and option-type array (`IndexedOptionArray`). **120 cases.**

However, putting off Awkward's extensions reduces this to only 8 types of slice by dropping jagged arrays and option-type arrays. (They can be left as `NotImplementedError`.) **96 cases.**

Also, I did the "weird" classes, so that what remains would be more straightforward: all of EmptyArray (because it's a leaf), NumpyArray (because it's a leaf and very different—much simpler—in Python), RecordArray (because fields are fundamentally different from other slices), and RegularArray (because fake RegularArrays are used to structure the getitem procedure). **64 cases.**

Also, I did the "weird" slice types, so that what remains would be more straightforward: no-slice, field, fields, `np.newaxis`, and `Ellipsis`. **24 cases.**

Here's what remains, for you to finish off this PR:

   - BitMaskedArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - ByteMaskedArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - IndexedArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - IndexedOptionArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - ListArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - ListOffsetArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - UnionArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array
   - UnmaskedArray
     - [x] integer slice
     - [x] range-slice
     - [x] flat array

As you finish implementing each one, uncomment the corresponding tests in test_1031-start-getitem_next.py. When the PR is done, all of the code should be uncommented. test_1031b-start-getitem_next-specialized.py is for any additional tests you need to add while working things out.

### DIFF
--- a/src/awkward/_v2/contents/bitmaskedarray.py
+++ b/src/awkward/_v2/contents/bitmaskedarray.py
@@ -222,11 +222,8 @@ class BitMaskedArray(Content):
         if head == ():
             return self
 
-        elif isinstance(head, int):
-            raise NotImplementedError
-
-        elif isinstance(head, slice):
-            raise NotImplementedError
+        elif isinstance(head, int) or isinstance(head, slice):
+            return self.toByteMaskedArray()._getitem_next(head, tail, advanced)
 
         elif ak._util.isstr(head):
             return self._getitem_next_field(head, tail, advanced)
--- a/src/awkward/_v2/contents/bitmaskedarray.py
+++ b/src/awkward/_v2/contents/bitmaskedarray.py
@@ -165,14 +165,14 @@ class BitMaskedArray(Content):
     def _getitem_at(self, where):
         if where < 0:
             where += len(self)
-        if 0 > where or where >= len(self):
+        if not (0 <= where < len(self)):
             raise ak._v2.contents.content.NestedIndexError(self, where)
         if self._lsb_order:
             bit = bool(self._mask[where // 8] & (1 << (where % 8)))
         else:
             bit = bool(self._mask[where // 8] & (128 >> (where % 8)))
         if bit == self._valid_when:
-            return self._content[where]
+            return self._content._getitem_at(where)
         else:
             return None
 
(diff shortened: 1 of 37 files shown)


jpivarski:
@ioanaif As a hint for how to go about translating and debugging, I showed some examples of my own debugging: 86377cc51b7ec42907a68297ee4d7749f4b55196 adds debugging code to compare the Python with the original C++ and 3878e10d3ecbbfa1ab8503f9a29f2473fbd2a170 is the subsequent commit that removes that debugging code (not sure which would be easier to read).

Once the C++ codebase is compiled, adding `std::cout` is fast to recompile (if you're using the `localbuild.py`), so you can add equivalent print statements on both sides when something is different, to try to figure out why something is different. The new Python code goes through all of the same steps as the C++ (in these "non-weird" cases), so you can check the intermediate Indexes to be sure that they're exactly the same at each step (or narrow in on where they're different).

jpivarski:
I should have done this when you first started!

@all-contributors please add @ioanaif  for code, test

jpivarski (commented):
It's great! I just made a few tweaks: "ISOPTION" was from when IndexedArray and IndexedOptionArray were templated versions of the same class; now they're distinct classes. The "nextcarry_outindex" function is only used for handling the option-type, so IndexedArray doesn't need it (and anyway, it's exactly the same function).

PR #1033, opened 2021-07-26
Script for inserting cout line in libawkward

SantamRC:
Run `gen.py` to insert `cout` in all kernel function calls in libawkward

### DIFF
--- a/src/libawkward/Identities.cpp
+++ b/src/libawkward/Identities.cpp
@@ -17,6 +17,7 @@
 #include "awkward/Identities.h"
 
 namespace awkward {
+  void printMe();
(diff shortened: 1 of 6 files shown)


src/libawkward/Identities.cpp
```diff
@@ -17,6 +17,7 @@
 #include "awkward/Identities.h"
 
 namespace awkward {
+  void printMe();
```
ianna: this function must be a method of `Identities` class. `Identities` class methods are defined in` Identities.h`.

ianna (changes requested):
@SantamRC please, move the function to the header file. Thanks

SantamRC:
Continuing work in #1044

Issue #176, opened 2020-03-18
Identities should not be assigned in place


jpivarski:
Instead of `setidentities`, the function should be `withidentities` and it should return a new array object.

Also, there's no high-level interface to identities and minimal testing. But identities are just a little more than a stub right now: they will be needed, will need to be enhanced when needed, and doing that will be much, much easier because they've been included all along and we won't be trying to fit them in after the fact.

jpivarski:
This is superseded by v2.

PR #111, opened 2020-02-06
Allow Awkward Arrays to be used as slices, including list-type and option-type.

jpivarski:
Addresses issue #67.

PR #1036, opened 2021-07-29
C++ refactoring: testing_starting _getitem_next

ioanaif:
test_0002-minimal-listarray.py 

- [x] test 
- [x] test_len 
- [x] test_members

test_0006-deep-iteration.py 

- [x] test_iterator

test_0008-slices-and-getitem.py 

- [x] test_numpyarray_getitem_bystrides 
- [x] test_numpyarray_getitem_next

test_0011-listarray.py 

- [x] test_listarray_basic 
- [x] test_listoffsetarray_basic 
- [x] test_listarray_at 
- [x] test_listoffsetarray_at 
- [x] test_listarray_slice 
- [x] test_listoffsetarray_slice 
- [x] test_listarray_slice_slice 
- [x] test_listoffsetarray_slice_slice 
- [x] test_listarray_ellipsis 
- [x] test_listoffsetarray_ellipsis 
- [x] test_listarray_array_slice 
- [x] test_listoffsetarray_array_slice 
- [x] test_listarray_array 
- [x] test_listoffsetarray_array

test_0013-error-handling-struct.py 

- [x] test_numpyarray 
- [x] test_listarray_numpyarray 
- [x] test_listarray_listarray_numpyarray

test_0014-finish-up-getitem.py 

- [x] test_basic 
- [x] test_listoffsetarrayA64 
- [x] test_listoffsetarrayA32 
- [x] test_listarrayA64
- [x] test_listarrayA32 
- [x] test_listoffsetarrayB64
- [x] test_listoffsetarrayB32 
- [x] test_listarrayB64 
- [x] test_listarrayB32

test_0020-support-unsigned-indexes.py 

- [x] test_listarray_basic 
- [x] test_listoffsetarray_basic 
- [x] test_listarray_at 
- [x] test_listoffsetarray_at 
- [x] test_listarray_slice 
- [x] test_listoffsetarray_slice 
- [x] test_listarray_slice_slice 
- [x] test_listoffsetarray_slice_slice 
- [x] test_listarray_ellipsis 
- [x] test_listoffsetarray_ellipsis 
- [x] test_listarray_array_slice 
- [x] test_listoffsetarray_array_slice 
- [x] test_listarray_array 
- [x] test_listoffsetarray_array

test_0021-emptyarray.py 

- [x] test_getitem

test_0023-regular-array.py 

- [x] test_iteration 
- [x] test_getitem_at 
- [x] test_getitem_range 
- [x] test_getitem 
- [x] test_getitem_deeper 
- [x] test_numpy

test_0024-use-regular-array.py 

- [x] test_empty_array_slice 
- [x] test_nonflat_slice 
- [x] test_newaxis

test_0025-record-array.py 

- [x] test_basic 
- [x] test_scalar_record 
- [x] test_getitem 
- [x] test_getitem_other_types 
- [x] test_getitem_next

test_0046-start-indexedarray.py 

- [x] test_basic 
- [x] test_null 
- [x] test_carry 
- [x] test_others 
- [x] test_missing

test_0084-start-unionarray.py 

- [x] test_getitem


jpivarski:
Good to see that this is started. Keep in mind that "refcount" tests can simply be deleted (it's in the Google Doc). The C++ code was heavily tested for reference counts because reference counting between Python and C++ had to be carefully managed. Since the new implementation is all in Python, reference counting is completely done for us and we can ignore those tests.

That's why it's a good idea to copy the old tests, rather than changing them in place. We still want the old tests to be there in the interim, but there's a lot that we can drop for v2.

src/awkward/_v2/contents/content.py
```diff
@@ -308,11 +329,11 @@ def format_slice(x):
             raise IndexError(
                 """cannot slice
 
-    {0}
+{0}
 
 with
 
-    {1}
+{1}
```
jpivarski: I wanted the two "semi-quoted" things (what you're slicing and what you're slicing it with) to be indented, to be separated from the rest of the text. The string representation of an `ak.Array` will never be multi-line, and it's hard to imagine a situation in which a slice would be. They should look okay indented.

src/awkward/_v2/contents/content.py
```diff
@@ -449,6 +470,31 @@ def _carry_identifier(self, carry, exception):
...
+        #             "awkward_new_Identities",
+        #             newidentifier.dtype.type
+        #         ](
+        #             newidentifier.to(nplike),
+        #             len(self),
+        #         )
+        # )
+        # self.setidentities(newidentities)
+        return self
+
```
jpivarski: There should not be a `setidentifier` interface because removing that is a long-standing issue: #176. Any tests that call `setidentifier` should just get removed. That way, the v1 → v2 transition would resolve that issue by simply not upgrading it.

Someday after that, I'll need to add a non-in-place `with_identifier`, but identifiers are a low priority now. (In the initial development, I thought they were going to be a lot more important.)

src/awkward/_v2/contents/listarray.py
```diff
@@ -151,14 +151,136 @@ def _carry(self, carry, allow_lazy, exception):
...
+        return ak._v2.contents.listoffsetarray.ListOffsetArray(
+            offsets, nextcontent, identifier, self._parameters
+        )
+
+    def toListOffsetArray64(self, start_at_zero):
+        offsets = self.compact_offsets64(start_at_zero)
+        out = self.broadcast_tooffsets64(offsets)
+        return out
+
+    def _getitem_next_jagged(self, slicestarts, slicestops, slicecontent, tail):
```
jpivarski: It's a better idea to not introduce `_getitem_next_jagged` until this PR, adding old tests and their consequences, is done. You don't want a single PR to have to deal with both updates.

None of the tests in the list should involve jagged slicing. If one does, then I've misidentified it. (And since jagged slicing was introduced in #111, it's hard to see how any test with a number smaller than that could include it.)

src/awkward/_v2/contents/listarray.py
```diff
@@ -151,14 +151,136 @@ def _carry(self, carry, allow_lazy, exception):
...
+
     def _getitem_next(self, head, tail, advanced):
         nplike = self.nplike  # noqa: F841
 
         if head == ():
             return self
 
         elif isinstance(head, int):
-            assert advanced is not None
+            assert advanced is None
```
jpivarski, line 187: Right.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -147,6 +147,95 @@ def _carry(self, carry, allow_lazy, exception):
...
+                self._offsets.dtype.type,
+            ](
+                out.to(nplike),
+                self._offsets.to(nplike),
+                self._length,
+            )
+        )
+        return out
+
+    def broadcast_tooffsets64(self, offsets):
```
jpivarski: If `compact_offsets64` and `broadcast_tooffset64` are very similar between ListArray and ListOffsetArray, we should find some way to combine them.

For example, the `minmax_depth` method on Form was repurposed in Content using a sneaky Python trick (the implementation duck-types to both classes, so it can just be referenced). This is a scale on which it will be beneficial to make the code DRY. (Sometimes people work too hard to make very short segments DRY, but it really matters on these 100 line functions.)

src/awkward/_v2/contents/recordarray.py
```diff
@@ -191,6 +194,12 @@ def content(self, index_or_key):
             )
         return self._contents[index][: self._length]
 
+    def contentitems(self):
+        pairs = []
+        for i in range(len(self._contents)):
+            pairs.append((self._keys[i], self._contents[i]))
+        return pairs
+
```
jpivarski: The test code might be calling this, but let's not reimplement it. In this translation, we'll have the opportunity to cut down on infrequently used methods.

The testing code can be changed from `array.contentitems()` to `zip(array.keys, array.contents)`. (Maybe `keys` is a method, not a property?)

src/awkward/_v2/contents/unionarray.py
```diff
@@ -81,6 +81,17 @@ def contents(self):
     def nplike(self):
         return self._tags.nplike
 
+    @property
+    def numcontents(self):
+        return len(self._contents)
```
jpivarski: This is another method that doesn't need to be translated. The testing code can be changed from `array.numcontents` to `len(array.contents)`. Many of these were in the old version because C++ doesn't have all the conveniences that Python has.

src/awkward/_v2/contents/content.py
```diff
@@ -214,10 +214,24 @@ def __getitem__(self, where):
...
             elif isinstance(where, tuple):
+
                 if len(where) == 0:
                     return self
+
+                prepare_tuple = []
+                for i in range(len(where)):
+                    x = self._prepare_tuple_item(where[i])
+                    if isinstance(x, list) and len(x) == 0:
+                        pass
```
jpivarski: Since `x` is the output of `_prepare_tuple_item`, `isinstance(x, list)` means that `x` is a list of strings, slicing fields. Skipping it would be equivalent to taking all of the fields (leaving the RecordArray as-is), but `x == []` would be equivalent to taking none of the fields (replacing the RecordArray with a RecordArray of the same `length` but no fields). Thus, this item should not be skipped; if there's an error in `_getitem_fields` that makes it behave badly, that error should be fixed in `_getitem_fields`. Which test instance motivated this?

src/awkward/_v2/contents/content.py
```diff
@@ -245,14 +259,17 @@ def __getitem__(self, where):
                     )
                     allow_lazy = "copied"  # True, but also can be modified in-place
                 elif issubclass(where.dtype.type, (np.bool_, bool)):
-                    carry = ak._v2.index.Index64(np.nonzero(where.data.reshape(-1))[0])
+                    nplike = ak.nplike.of(where)
+                    where = nplike.nonzero(where.data.reshape(-1))[0]
```
jpivarski: I think a better choice than `nplike = ak.nplike.of(where)` is `nplike = self.nplike`, which is to say, "regardless of whether the `where` array is on the CPU or GPU, move it to be on the same device as `self`." In fact, `where` could be a non-array, like a Python list, in which case `ak.nplike.of(where)` would return NumPy, but `self.nplike` could be CuPy. It will have to eventually be moved to the device `self` is on, so better to do that early.

Since this is only used once on the following line, `self.nplike` can just be used inline there.

src/awkward/_v2/contents/content.py
```diff
@@ -265,7 +282,11 @@ def __getitem__(self, where):
                 return self.__getitem__((where,))
 
             elif isinstance(where, Iterable) and all(ak._util.isstr(x) for x in where):
-                return self._getitem_fields(where)
+                if len(where) == 0:
+                    # FIXME
+                    return []
```
jpivarski: This is related to the comment on line 224 of this file. `Content.__getitem__` definitely should not be returning a Python list (the allowed return types are `Content`, `dtype.type` (i.e. int, bool, ...), and `None`). The `_getitem_fields` method should be handling an empty `where` by converting the first RecordArray into a RecordArray with no fields. If it's not, that's where it should be fixed.

RecordArrays can have zero `contents`; this is why they need a `length` argument:

```python
>>> RecordArray([], [], 10)
<RecordArray len='10' is_tuple='False'>
</RecordArray>
>>> ak.to_list(RecordArray([], [], 10))
[{}, {}, {}, {}, {}, {}, {}, {}, {}, {}]
```

src/awkward/_v2/contents/unionarray.py
```diff
@@ -81,6 +81,13 @@ def contents(self):
     def nplike(self):
         return self._tags.nplike
 
+    def keys(self):
+        keys = set()
+        for i in range(len(self._contents)):
+            if hasattr(self._contents[i], "keys"):
+                keys.update(self._contents[i].keys)
+        return keys
+
```
jpivarski: For this to work in general, all `Content` types need to have `keys` defined, and they need to be consistent about whether it's a method or a property. This implementation assumes that `self._contents[i]` has a `keys` property, which won't work if `self._contents[i]` is a UnionArray because you just defined this `keys` as a method. (No parentheses.) I think RecordArray has a `keys` property, so that will require all `Content` subclasses to have a `keys` property, not method.

The output should be a list. I don't think it was a set in the past—that would cause problems with Python 2.7 and 3.5, which don't have stable set order. I also think the `keys` of a Union _ought_ to be the intersection, rather than the union (as above), since the only keys one could pass to `_getitem_field` or `_getitem_fields` are the ones that exist on all `contents` of the UnionArray. Maybe it wasn't originally designed that way (in v1), but we have more experience with UnionArrays now: they're only useful in operations that assume the least common denominator of the `contents`, which would be an intersection here. Any test instances that require the output to be the union can be changed or deleted.

Independent of the above, this is a method that can be (and therefore should be) implemented on the Form subclasses, rather than the Content subclasses. What it returns only depends on the metadata, which is in the Form. By defining it on Forms, VirtualArray will be able to use it without unnecessarily reading data. That makes this method like `minmax_depth`, `purelist_depth`, and a few others that I implemented on the Forms a week or two ago. The Content implementations just link to the Form implementations, using a Python trick. This `keys` property should eventually become like that, but if you want to just delete the test instances that need it and move on, we can do it later.

src/awkward/_v2/contents/regulararray.py
```diff
@@ -165,6 +165,69 @@ def _carry(self, carry, allow_lazy, exception):
...
+        self._handle_error(
+            nplike["awkward_RegularArray_compact_offsets", out.dtype.type](
+                out.to(nplike),
+                self._length,
+                self._size,
+            )
+        )
+        return out
+
+    def broadcast_tooffsets64(self, offsets):
```
jpivarski: Quick request: let's make these (`compact_offsets64` and `broadcast_tooffsets64`) private methods (`_compact_offsets64` and `_broadcast_tooffsets64`). I'm doing the same in #1059.

src/awkward/_v2/contents/content.py
ioanaif: test_0008_v2-slices-and-getitem.py:74
test_0021_v2-emptyarray.py::test_getitem:39

src/awkward/_v2/contents/listarray.py
```diff
@@ -184,8 +184,11 @@ def _getitem_next(self, head, tail, advanced):
 
             nexthead, nexttail = self._headtail(tail)
 
-            start, stop, step = head.indices(self._stops[0])
+            start, stop, step = head.start, head.stop, head.step
+
             step = 1 if step is None else step
+            start = 0 if start is None else start
+            stop = ak._util.kSliceNone if stop is None else stop
```
jpivarski: @ioanaif If `step` is negative, `start is None` is not equivalent to `start == 0`; negative `step` means you're counting down from the length of the list, which isn't a single value that can be replaced here (all of the lists may have different lengths). In situations like this, it should be

```python
            step = 1 if step is None else step
            start = ak._util.kSliceNone if start is None else start
            stop = ak._util.kSliceNone if stop is None else stop
```

I'll do this in the next commit.

jpivarski (approved):
@ioanaif This is great, thank you! It looks like handling ListOffsetArray and IndexedOptionArray in `_getitem_next` is next.

I just made some tweaks and implemented enough for the skipped tests to pass. The skipped test "reasons" mentioned the lack of `toListOffsetArray`, but what was actually happening was that `Content.__getitem__` wasn't handling multi-dimensional NumPy arrays well: it was converting them into RegularArray of NumpyArray, but only NumpyArray was recognized as a rectilinear array type. If we let those RegularArrays be converted into ListOffsetArrays, they would do the slices incorrectly.

So now that's fixed and we _also_ have an implementation of `toListOffsetArray`.

Notice that `ListArray._broadcast_tooffsets64` and `ListOffsetArray._broadcast_tooffsets64` would be exactly the same, since the kernels only rely on the existence of `starts` and `stops`. (An `offsets` index can be zero-copy converted to `starts` and `stops`, but not the other way around.) Since this isn't C++, it's possible to implement it only on ListOffsetArray and have the ListArray class use ListOffsetArray's implementation, just as properties like `minmax_depth` reuse the Forms' implementations. This technique will likely help us minimize complexity in other methods, too.

Issue #1038, opened 2021-07-30
Extend the array-class syntax in `ak.behavior`


agoose77:
### Description of new feature

As discussed on Gitter, it may be useful for some people to be able to specify lower bounds on the dimensionality required for array classes to be chosen by their `behavior` name. 

# Syntax
## Option A
The `"+"` token was mused to reflect "one dimension":

* `ak.behavior["+", "point"]` would denote a one dimensional array of points.
* `ak.behavior["*", "point"]` would denote a one-plus dimensional array of points.
* `ak.behavior["*", "+", "point"]` would denote a two-plus dimensional array of points.

## Option B
A cleaner approach might instead use only `"*"`, and would allow users only to specify `ndim >= k` instead of `ndim == k`. This would look like:

* `ak.behavior["*", "point"]` would denote a one-plus dimensional array of points.
* `ak.behavior["*", "*", "point"]` would denote a two-plus dimensional array of points.

## Option C
Seeing as we already use `"."` to mean a single dimension array, the same behaviour in A could be implemented with `"."`:

* `ak.behavior[".", "point"]` would denote a one dimensional array of points.
* `ak.behavior["*", "point"]` would denote a one-plus dimensional array of points.
* `ak.behavior["*", ".", "point"]` would denote a two-plus dimensional array of points.

# Observations
One problem that I can see with option A is that it uses regular expressions qualifiers, but whose behaviours are inverted (`"*"` is 0+ in regex, and `"+"` is 1+).

agoose77:
I took a brief look at the `arrayclass` function to gauge what would need to be done here. It looks like this might require a little bit of thought, because we currently do not seem to expose a means to determine the depth at which a `__record__` parameter was found. For this reason, it might be prudent to wait until v2 is done.

jpivarski:
I guess it's not quite like regular expressions.

Since `"."` already means "one level" and `"*"` means "any number of levels," how about

   * `".", "."` for "two levels"
   * `".", ".", "."` for "three levels"
   * `"*", "."` for "two or more levels"
   * `"*", ".", "."` for "three or more levels"

? I don't think it would be necessary to introduce a number here, to simplify the problem of asking for "57 levels," since that kind of thing is very unlikely. I can't imagine requirements needing to specify anything bigger than 3.

agoose77:
> I guess it's not quite like regular expressions.

[...]

PR #1039, opened 2021-07-30
Extend arrayclass to depend on number of dimensions.

jpivarski:
@agoose77 I didn't want #1038 to have to wait for v2, so I wrote `nlist_parameter` to demonstrate how `purelist_parameter` can be generalized. This method is like `purelist_parameter` except that it will only report the parameter at a given depth. Unlike `purelist_parameter`, it "looks through" nodes at a higher depth than the one specified and like `purelist_parameter`, it will take the first node that has the requested parameter _at_ the specified depth.

Now that I look at the original problem, you also want "depth greater than or equal to," but that would be a simple modification of this method: after `n` reaches `0`, it can switch over to `purelist_parameter` to keep going. The main thing about these methods is that they encode the knowledge that ListArray, ListOffsetArray, and RegularArray are dimensions, RecordArrays are terminal, UnionArrays require agreement among all of their branches, and everything else is a non-dimension node.

You can use this PR as a base for solving #1038 or as an example of how a method _like_ `purelist_parameter` is implemented, since it's illustrated in the diff of this first commit. Whatever you come up with can easily be ported to v2.

### DIFF
--- a/include/awkward/Content.h
+++ b/include/awkward/Content.h
@@ -105,6 +105,11 @@ namespace awkward {
     virtual const std::string
       purelist_parameter(const std::string& key) const = 0;
 
+    /// @brief The parameter associated with `key` n levels of list
+    /// dimensions deep.
+    virtual const std::string
+      nlist_parameter(const std::string& key, int64_t n) const = 0;
+
     /// @brief Returns `true` if the parameter associated with `key` exists
     /// and is equal to `value`; `false` otherwise.
     ///
@@ -558,6 +563,11 @@ namespace awkward {
     /// RecordArray.
     virtual const std::string
       purelist_parameter(const std::string& key) const;
+
+    /// @brief The parameter associated with `key` n levels of list
+    /// dimensions deep.
+    virtual const std::string
+      nlist_parameter(const std::string& key, int64_t n) const;
 
     /// @brief Returns `true` if all nested lists down to the first RecordArray
     /// are RegularArray nodes; `false` otherwise.
--- a/include/awkward/array/BitMaskedArray.h
+++ b/include/awkward/array/BitMaskedArray.h
@@ -59,6 +59,9 @@ namespace awkward {
     const std::string
       purelist_parameter(const std::string& key) const override;
 
+    const std::string
+      nlist_parameter(const std::string& key, int64_t n) const override;
+
     bool
       purelist_isregular() const override;
 
--- a/include/awkward/array/ByteMaskedArray.h
+++ b/include/awkward/array/ByteMaskedArray.h
@@ -53,6 +53,9 @@ namespace awkward {
     const std::string
       purelist_parameter(const std::string& key) const override;
 
+    const std::string
+      nlist_parameter(const std::string& key, int64_t n) const override;
+
     bool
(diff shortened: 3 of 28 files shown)


agoose77:
Thanks for the starting point Jim. If you're happy that we need to extend the content API in order to facilitate #1038, that's helpful!

It looks like this particular PR would be used in `arrayclass` like
```python3

for i in range(layout.purelist_depth):
    record = layout.nlist_parameter(i)
    if record is not None:
        break
else:
    raise ValueError("no record found")
```

We would end up visiting the outer contents multiple times in this manner. Are you averse to having a `purelist_parameter_depth` which returns a `std::pair`with the parameter and the depth at which it was found?

jpivarski:
> We would end up visiting the outer contents multiple times in this manner. Are you averse to having a `purelist_parameter_depth` which returns a `std::pair`with the parameter and the depth at which it was found?

I'm not averse to that. (pybind turns `std::pair` into a Python tuple.) But also, there's no reason to avoid walking over a tree structure multiple times. The one large order parameter that we preemptively optimize for is the length of the arrays—that's why we have kernels separated out and rules against iterating over data in the middle layer, etc. But the size of a tree of layout nodes is not considered a parameter that can scale large, since it scales with the complexity of the data _type_, not the size of the dataset, so we don't need to optimize traversals over it. Walk over it 20 times if makes the code simpler.

(That was a hard thing for me to get used to when I started programming in Lisp. Lisp was gleefully traversing and copying data structures that would have been more carefully examined in C or Pascal, and at first it seemed terribly wasteful. But then I saw how much easier it was to do things, not having to worry about minimizing the computer's work all the time, and a lot of data really is small. So really, it's more about identifying the few things you _do_ need to worry about as order parameters that can scale large, and everything else is written for clarity, rather than speed.)

agoose77:
> But also, there's no reason to avoid walking over a tree structure multiple times.

I don't disagree here, but I think in this case it might make implementation simpler — is it more common to want a parameter at a particular depth, or at any depth (but the depth needs to be known)? If the latter, then extending the `purelist_parameter` to give the depth might be helpful.

jpivarski:
Okay, as long as it's simpler for humans. My main point being that this particular place is not one to consider performance-optimizing.

agoose77:
> Okay, as long as it's simpler for humans. My main point being that this particular place is not one to consider performance-optimizing.

I think it was a point worth stating — it's easy to get out of the mindset of "only the (important) loops matter", and in this case, the dimensions are *not* the important loops!

jpivarski:
This is the second-oldest PR, it's about a policy change, and I think it's been forgotten about. It would be difficult to get this up-to-date as things currently stand, let alone after #1244.

Is it an important feature? It's been half a year since we've talked about it, which is suggestive that it's not that necessary.

Can we close the PR?

agoose77:
I think we might still want this behaviour, but you make a good case for closing it for now. Let's discuss again in future if/when it is raised again?

jpivarski:
Right. The good news is that the mechanism for doing this is almost exactly the same in v2 as v1, so we can reference this PR when starting a new attempt at implementing it again.

PR #1040, opened 2021-07-30
Proposal to stop ak.broadcast_arrays at records.

jpivarski:
@agoose77 This is prompted by your question on Gitter:

> Hey @jpivarski , sorry for the visual noise - I've determined that you can't edit messages on Matrix once someone has replied to them (well, you can, but Gitter displays copied messages 🤮).
> 
> Should two arrays with different dimensions and record types be broadcastable?
> E.g.
> 
> ```
> >>> x.type
> 4056 * var * {"u": ?int64, "v": ?int64}
> >>> y.type
> 4056 * var * var * {"q": float64, "t": float64}
> >>> ak.broadcast_arrays(x, y)
> ValueError ...
> ```
> We currently require the record fields to match if we allow records to broadcast against one another.
> 
> This behavior initially surprised me - I expected the record-part to be irrelevant and successfully left-broadcast; the record is part of the type, and we allow different dtypes to broadcast:
> 
> ```
> >>> ak.broadcast_arrays(x.u, y.t)
> [<Array [[[66], [65]], ... 64], [68], [65]]] type='4056 * var * option[var * int64]'>,
>  <Array [[[32], [199]], ... [31], [275]]] type='4056 * var * option[var * float64]'>]
> ```
> However, I'm also not sure about the resulting type here - naively I'd expect the option to wrap the dtype, not the var dimension.
> 
> The broadcasting works if I use zip, because zip exits early from the broadcasting once the purelist_depth is 1 (i.e. it doesn't encounter the bare record).
> 
> ```
> >>> ak.zip((u, v)).type
> 4056 * var * var * ({"u": ?int64, "v": ?int64}, {"t": float64, "q": float64})
> ```
> Most likely I am missing something here, but perhaps you could shed some light on this? @jpivarski

What I've done here is to stop broadcasting at records when all inputs are records, so that non-records are still broadcasted into the fields of a record, but two records aren't broadcasted into each other (which is impossible, which is why you got the ValueError). Maybe the rule should be "any record" rather than "all records"? That's why this PR is a draft.

See tests/test_1040-proposal-to-stop-broadcast_arrays-at-records.py for examples.

Also note that this only affects `ak.broadcast_arrays`, not broadcasting in general.

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1258,11 +1258,12 @@ def broadcast_arrays(*arrays, **kwargs):
             right-broadcasting, as described below.
         highlevel (bool, default is True): If True, return an #ak.Array;
             otherwise, return a low-level #ak.layout.Content subclass.
-        promote_scalar_to_record (bool, default is True): If True,
-            scalar values will be promoted to records using the record keys,
+        stop_at_record (bool, default is False): If False,
+            broadcasting against a record will build a common record for the
+            given arrays. For scalars, this means promoting them to a record,
             e.g. `(10.0, {'eta': 1.4})` becomes `({'eta': 10}, {'eta': 1.4})`.
-            Otherwise, preserve scalars i.e. `(10.0, {'eta': 1.4})` remains
-            unchanged.
+            Otherwise, broadcasting will terminate when it encounters a record
+            i.e. `(10.0, {'eta': 1.4})` remains unchanged.
 
     Like NumPy's
     [broadcast_arrays](https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast_arrays.html)
@@ -1356,19 +1357,14 @@ def broadcast_arrays(*arrays, **kwargs):
     #ak.Array.type, but it is lost when converting an array into JSON or
     Python objects.
     """
-    (
-        highlevel,
-        left_broadcast,
-        right_broadcast,
-        promote_scalar_to_record,
-    ) = ak._util.extra(
+    (highlevel, left_broadcast, right_broadcast, stop_at_record,) = ak._util.extra(
         (),
         kwargs,
         [
             ("highlevel", True),
             ("left_broadcast", True),
             ("right_broadcast", True),
-            ("promote_scalar_to_record", True),
+            ("stop_at_record", False),
         ],
     )
 
@@ -1386,12 +1382,12 @@ def broadcast_arrays(*arrays, **kwargs):
     # broadcasting (A, B), where A and B have different fields. Therefore, we exit early if we
     # encounter (record, record, ..., record), or (scalar, record, ...) and the user has opted out
     # from this promotion behavior.
-    should_not_broadcast_fields = all if promote_scalar_to_record else any
+    should_stop_at_records = any if stop_at_record else all
 
     def getfunction(inputs):
         if all(isinstance(x, ak.layout.NumpyArray) for x in inputs):
             return lambda: tuple(inputs)
-        elif should_not_broadcast_fields(
+        elif should_stop_at_records(
             isinstance(x, ak.layout.RecordArray) for x in inputs
         ):
             return lambda: tuple(inputs)
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1369,6 +1369,8 @@ def broadcast_arrays(*arrays, **kwargs):
     def getfunction(inputs):
         if all(isinstance(x, ak.layout.NumpyArray) for x in inputs):
             return lambda: tuple(inputs)
+        elif all(isinstance(x, ak.layout.RecordArray) for x in inputs):
+            return lambda: tuple(inputs)
         else:
             return None
 
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -1258,6 +1258,11 @@ def broadcast_arrays(*arrays, **kwargs):
             right-broadcasting, as described below.
         highlevel (bool, default is True): If True, return an #ak.Array;
             otherwise, return a low-level #ak.layout.Content subclass.
+        promote_scalar_to_record (bool, default is True): If True,
+            scalar values will be promoted to records using the record keys,
+            e.g. `(10.0, {'eta': 1.4})` becomes `({'eta': 10}, {'eta': 1.4})`.
+            Otherwise, preserve scalars i.e. `(10.0, {'eta': 1.4})` remains
+            unchanged.
 
     Like NumPy's
     [broadcast_arrays](https://docs.scipy.org/doc/numpy/reference/generated/numpy.broadcast_arrays.html)
@@ -1351,10 +1356,20 @@ def broadcast_arrays(*arrays, **kwargs):
     #ak.Array.type, but it is lost when converting an array into JSON or
     Python objects.
     """
-    (highlevel, left_broadcast, right_broadcast) = ak._util.extra(
+    (
+        highlevel,
+        left_broadcast,
+        right_broadcast,
+        promote_scalar_to_record,
+    ) = ak._util.extra(
         (),
         kwargs,
-        [("highlevel", True), ("left_broadcast", True), ("right_broadcast", True)],
+        [
+            ("highlevel", True),
+            ("left_broadcast", True),
+            ("right_broadcast", True),
+            ("promote_scalar_to_record", True),
+        ],
     )
 
     inputs = []
@@ -1366,10 +1381,19 @@ def broadcast_arrays(*arrays, **kwargs):
             y = ak.layout.NumpyArray(ak.nplike.of(*arrays).array([y]))
         inputs.append(y)
 
+    # We do not want the (scalar, record, ...)  promotion behavior of `broadcast_and_apply` to occur
+    # in the case of (record, record) in this routine. Otherwise, we would raise an error when
+    # broadcasting (A, B), where A and B have different fields. Therefore, we exit early if we
+    # encounter (record, record, ..., record), or (scalar, record, ...) and the user has opted out
+    # from this promotion behavior.
+    should_not_broadcast_fields = all if promote_scalar_to_record else any
+
     def getfunction(inputs):
         if all(isinstance(x, ak.layout.NumpyArray) for x in inputs):
             return lambda: tuple(inputs)
-        elif all(isinstance(x, ak.layout.RecordArray) for x in inputs):
+        elif should_not_broadcast_fields(
+            isinstance(x, ak.layout.RecordArray) for x in inputs
+        ):
             return lambda: tuple(inputs)
         else:
             return None
--- a/tests/test_1040-proposal-to-stop-broadcast_arrays-at-records.py
+++ b/tests/test_1040-proposal-to-stop-broadcast_arrays-at-records.py
@@ -1,0 +1,32 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from __future__ import absolute_import
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import awkward as ak  # noqa: F401
+
+
+def test():
+    assert [x.tolist() for x in ak.broadcast_arrays([1, 2, 3], [[], [1], [1, 2]])] == [
+        [[], [2], [3, 3]],
+        [[], [1], [1, 2]],
+    ]
+
+    assert [
+        x.tolist()
+        for x in ak.broadcast_arrays([1, 2, 3], [[], [{"x": 1}], [{"x": 1}, {"x": 2}]])
+    ] == [
+        [[], [{"x": 2}], [{"x": 3}, {"x": 3}]],
+        [[], [{"x": 1}], [{"x": 1}, {"x": 2}]],
+    ]
+
+    assert [
+        x.tolist()
+        for x in ak.broadcast_arrays(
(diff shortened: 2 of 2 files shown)


agoose77:
That was quick, thanks Jim. I had not realised that we had this broadcasting behaviour between record arrays and non-record arrays. Seeing this PR encouraged me to verify the behaviour, and initially it took me by surprise.

> Maybe the rule should be "any record" rather than "all records"? That's why this PR is a draft.

Well, I keep flip-flopping on this question. I think we should perhaps offer a `promote_scalars_to_records=True` argument to `broadcast_arrays`, so that if users *don't* want this, they can opt-out? 

I was at one point concerned about the existing promotion behaviour, but after considering nested record arrays, it is the only sensible way to perform promotion, so I'm in favour of it.

agoose77:
@jpivarski I like your proposed change here. I *do* think that `any` vs `all` should be a user-parameter - I can imagine a time where transparently populating records when broadcasting a scalar against a record might be useful (`all`), but I mostly find myself wanting to preserve non-record-ness (`any` ). 

What do you think of these new commits?

agoose77:
Actually, I don't think this is quite right to call this `promote_scalar_to_record`. Perhaps `stop_at_record` is a better name?

agoose77:
Hmm, maybe this naming isn't the best, though ­— we're actually just testing whether the layouts at the current depth are records or not. They might be scalars, or they might be lists (of scalars / records). I've therefore changed the PR.

jpivarski:
Since this function is `broadcast_arrays`, it can have fine-grained arguments for "any" vs "all". This is what people should turn to if another function's broadcasting isn't fine-grained enough. I wouldn't use the word "stop", though: I doubt people are going to be thinking of this in terms of the process that implements it (which steps downward to a certain level and then "stops"). Maybe `broadcast_any_records` and `broadcast_all_records`?

jpivarski:
Also, I can override that one required-but-not-running test. It's because our set of tests has changed to accommodate Python 3.10; I moved the Windows ones to only test 32-bit on one Python version, which is 3.9 because that's currently the most mainstream.

agoose77:
> 

I see your point. It also might be worth pinning down some concrete examples to discuss; there are some places where this new behaviour might not be desirable.

## Broadcasting same records
Given these arrays,
```python
>>> left = ak.Array([[{"x": [1, 2]}]])
>>> right = ak.Array([[{"x": [[3, 4], [5, 6]]}]])
```
the current behaviour is as follows:
```python3
>>> ak.broadcast_arrays(left, right)
[<Array [[{x: [[1, 1], [2, 2]]}]] type='1 * var * {"x": var * var * int64}'>,
 <Array [[{x: [[3, 4], [5, 6]]}]] type='1 * var * {"x": var * var * int64}'>]
 ```
With `all`, we would have:
```python
[<Array [[{x: [1, 2]}]] type='1 * var * {"x": var * int64}'>,
 <Array [[{x: [[3, 4], [5, 6]]}]] type='1 * var * {"x": var * var * int64}'>]
```

 I think that this behaviour is wrong. My feeling is that we should try and broadcast through the records, and *only if they're not compatible* do we return early at this depth. A record being compatible with another is determined by, to my mind, `ak.fields(this) == ak.fields(that)`.  If `allow_incompatible_records=False` then encountering incompatible records would raise an `Exception` as it currently does in `main`.

## Promoting non-records to records
Maybe the non-record ­– record broadcasting could be a nice option. Unlike non-identical records, the current non-record-becomes-record behaviour is well defined, so it could be an opt in via `coerce_to_record=True`? 

e.g.
```python
>>> scalar = ak.Array([5])
>>> ak.broadcast_arrays(left, scalar, coerce_to_record=True)
[<Array [[{x: [1, 2]}]] type='1 * var * {"x": var * int64}'>,
 <Array [[{x: [5, 5]}]] type='1 * var * {"x": var * int64}'>]
```
and
```python
>>> ak.broadcast_arrays(left, scalar, coerce_to_record=False)
[<Array [[{x: [1, 2]}]] type='1 * var * {"x": var * int64}'>,
 <Array [[5]] type='1 * var * int64'>]
```

The overall new behaviour would then be:

- Identical (compatible) records are transparent to `broadcast_and_apply`
- Incompatible records don't raise a `ValueError`, but instead stop broadcasting at that layout (like `all` currently does). Maybe this becomes an error if `allow_incompatible_records` is `False`?
- Record – non-record cases stop broadcasting, unless `coerce_to_record` is set to `True`

As for the API, normally I advocate for new functions, but in this case I feel like it might be better to have parameters to `broadcast_arrays`. This is primarily because we currently expose the NEP 35 equivalents at the high-level API (i.e. `ak.ravel`), so if we have to have `broadcast_arrays`, we might as well let it have a broader API instead of multiple separate additional functions. Does this argument appeal to you?

jpivarski:
Transparent broadcasting of compatible records (all `inputs` have the same sets of fields) had to be disabled because functions with broadcasting, such as `+`, were doing bad things like:

```python
>>> ak.Array([{"r": 1, "phi": 0.1}]) + ak.Array([{"r": 2, "phi": 0.2}])   # real code raises error, now
<Array [{r: 3, phi: 0.3}] type='1 * {"r": int64, "phi": float64}'>
```

These arrays of polar-coordinate vectors are missing the `behavior` that would add them correctly, but that can happen accidentally. Coffea found that it would be better for such things to be errors than to be added naively (as a reminder to set the `behavior` properly). That means that broadcasting, in general, can't pass them through. Broadcasting has to stop at record boundaries.

I would have thought that `ak.broadcast_arrays` would raise an error just as attempting to add them does, but it passes the inputs back without modification:

```python
>>> ak.broadcast_arrays(ak.Array([{"r": 1, "phi": 0.1}]), ak.Array([{"r": 2, "phi": 0.2}]))
[<Array [{r: 1, phi: 0.1}] type='1 * {"r": int64, "phi": float64}'>,
 <Array [{r: 2, phi: 0.2}] type='1 * {"r": int64, "phi": float64}'>]
```

Is that different in this branch? (Oh yeah, that was the original point of this PR, I think.)

Broadcasting through records of identical fields could be enabled in `ak.broadcast_arrays` with an argument, but the default should be standard broadcasting, like what `+` does.

Broadcasting scalars into records looks interesting, but I can't think of a use-case. For most operations (like `+`, in which users can't override broadcasting behaviors), arrays with identical fields can't be used together, so I don't see why one would want to construct arrays with this property.

agoose77:
> Transparent broadcasting of compatible records (all `inputs` have the same sets of fields) had to be disabled because functions with broadcasting, such as `+`, were doing bad things like:

Right, I recall you mentioning this change. It might be that I don't have a good enough mental model of this particular code yet — I suggested the identical record case for where you might have sublists within record fields which differ, e.g. 
```python
(
    var * pos[x * var * float64],
    var * pos[x * float64]
)
```

I would naively want this to produce `var * pos[x * var * float64]` (and using the `parameters` of the deeper layout). 

In fact, this is what we have in `main` (v1):
```python3
>>> a = ak.Array([{'x': [1]}])
>>> b = ak.Array([{'x': [[3, 4]]}])
>>> ak.broadcast_arrays(a, b)
[<Array [{x: [[1, 1]]}] type='1 * {"x": var * var * int64}'>,
 <Array [{x: [[3, 4]]}] type='1 * {"x": var * var * int64}'>]
```

I'm probably misusing some terminology here, which might be where we are crossing wires.


> Broadcasting scalars into records looks interesting, but I can't think of a use-case. For most operations (like `+`, in which users can't override broadcasting behaviors), arrays with identical fields can't be used together, so I don't see why one would want to construct arrays with this property.

As for scalars-into-records, we currently also have
```python
>>> c = ak.Array([[3,4]])
>>>  ak.broadcast_arrays(a, c)
[<Array [[{x: [1]}, {x: [1]}]] type='1 * var * {"x": var * int64}'>,
 <Array [[{x: [3]}, {x: [4]}]] type='1 * var * {"x": var * int64}'>]
```
Do I understand correctly that you'd prefer to drop this?

jpivarski:
I'm coming back to this with too-large gaps in between.

Broadcasting and records is just confusing me, maybe they're incompatible concepts. One concept comes from APL/NumPy, land of purely numerical arrays, and the other comes from ALGOL/OOP. Awkward attempts to bring these worlds together as much as possible, but maybe this one just doesn't fit. NumPy-style reducers and sorting, which cross between levels of nested lists when `axis != -1`, also doesn't mix well with records, and I think the original implementation had to be revoked because, however it could be defined, it wasn't predictable. The same thing might be true here.

I'm not saying the old implementation was right—it was probably allowing too much. Could we just remove broadcasting through records, just treat them as solid objects?

Coming back to the original question,

> ```python
> >>> x.type
> 4056 * var * {"u": ?int64, "v": ?int64}
> >>> y.type
> 4056 * var * var * {"q": float64, "t": float64}
> >>> ak.broadcast_arrays(x, y)
> ValueError ...
> ```

broadcasting the `var` to `var * var` is fine, but if `"u"`, `"v"` is broadcasted somehow into `"q"`, `"t"` (or even if they had the same field names but different types, like one is scalar and the other is a list), then the records' types have changed. Whatever `parameters` they had would have to be invalidated: if one of them used to be a Momentum4D, it would have to lose that type. That's probably unexpected.

Taking a step back, languages could take records as being more or less "solid": ALGOL-like languages (C, Python, Java) consider records to be very solid: they have names, maybe methods, and constructing them or taking them apart is a laborious process. Structured arrays in NumPy are very loose, and Awkward Array is somewhere in between—it's the fact that they can have behaviors that tightens it somewhat, because you usually don't want Lorentz vectors to turn into non-vectors, but they are _somewhat_ softer than OOP objects. We just have to decide how soft.

jpivarski:
Oh, and the set of required tests can get fixed if you update to `main`.

agoose77:
I agree that this is a tricky area: I've rewritten this answer a couple of times. I generally like the feeling in Awkward that records are very loose - `ak.zip` and `ak.unzip` are wonderful features that really solve the awkward (no pun intended) feeling of mapping OO-logic onto array programming. I personally see types (`__record__`) as a secondary, more rigid system that is layered on top of these loose records. As we discussed, these types need to drop the moment that the structure changes, but I think that is a reasonable requirement. 

Let's assume that we always have two main cases:
- identical records with different substructures, i.e. same fields, but internal lists may be different.
- scalar-only with different structures

I think these cover many of the examples that we would encounter in the wild. 

There is a third case, for scalar-record, and I think that warrants consideration. Unlike the other two examples, where the operations to bring two arrays into compatible shapes is well posed, here it is not. The whole utility of broadcasting is to make operations between two arrays "just work". In the case of scalar vs record, what does "just work" even mean? There is no primitive operation that can directly operate between a record and a scalar, unless the record implements a behaviour for it. 

I feel that we should prohibit broadcasting scalars against records. It will be annoying for my use case, but equally, to replicate the behaviour I wanted earlier, you can just pull out the leaf that you need, e.g.
```python
scalar, _ = ak.broadcast_arrays(scalar, record.this.that.leaf)
```
This _is_ well defined.

Then, I think we should require that any (record, record) case should permit _only_ matching records. Either we do this by taking only the common fields, *or* we require identical fields. Obviously, we have to drop parameters in the former case.

How does this sound? I am not sure if I'm missing something

jpivarski:
I think we're on the same page; let me repeat it back in my words to see if we are. (I'm going to avoid the word "scalar" because the non-records we're talking about here are lists/arrays.)

   * Non-records and non-records are unchanged: uncontroversial case.
   * Non-records and records are to be prohibited because it's probably a mistake and `record.this.that.leaf` is easy enough to type. Perhaps the error message should suggest this. (Here, we're taking advantage of the "softness" of records, but asking users to explicitly use that softness, rather than making assumptions about what they want.)
   * Records and records with different sets of fields should be prohibited. Whatever they thought they meant by that, it was wrong.
   * Records and records with the same sets of fields could have each individual field get broadcasted and if any records change their structure, they have to lose their `parameters` because they are no longer the same type.

Records and records with the same sets of fields still can't be used in any mathematical operations, even if the types of all fields exactly match. But now that I'm thinking of it, that prohibition is not a prohibition in broadcasting, but in function application: `broadcast_and_apply` does _two_ things; mathematical operations are not allowed to do the second of those two things when records are involved. The `ak.broadcast_arrays` function is not doing the second of those two things; it's not applying anything. The output has the same number of arrays as the input.

Is that right?

agoose77:
@jpivarski right, that's exactly as I see it. AFAICT, if someone *wants* to do something that operates on two different record types (here "type" means the record datashape) then they need to handle that in the getfunction that handles control flow. Awkward code can just treat that as an error.
> Records and records with the same sets of fields still can't be used in any mathematical operations, even if the types of all fields exactly match. 

Yes, unless they override the operators via a behaviour(?)

How do you feel about taking the intersection of record keys vs requiring them to be identical? I like the idea that it *might* be useful to take the intersection instead of requiring hard compatibility, but perhaps that's a user option with a sane default?

Maybe scalar is a poor choice of words, as it connotes singular values. Perhaps a better word would be "primitive" arrays vs "record" arrays? Here I mean any array that contains a record vs any array that does not.

jpivarski:
This is the third-oldest PR, it's about a policy change, and we're in agreement about what the `ak.broadcast` behavior ought to be. Something has been implemented for v1 (is it the full change?) but not v2.

It looks like it won't be hard to merge with #1244. If #1244 goes into main first (it probably will), I think it would be relatively easy to merge main into this branch and continue. So there's no hurry to finish this, but you'll want to be aware of that dependency.

(To be clear, I don't think this PR should be closed!)

jpivarski:
This one is old and is a policy change in v1, which won't be changing like that anymore—only bug-fixes. It is a policy that we might pick up in v2, and we've been addressing a lot of questions about how broadcasting works recently (e.g. #1679).

PR #1045, opened 2021-08-03
LayoutBuilder refactoring

ianna:


ianna:
@jpivarski - `LayoutBuilder` code is clean from dependencies on the `Content` classes, `Type`, and `Form` classes. It takes a JSON form. I will need to revisit `pybind11` part that still have some dependencies on the `Content` classes.

jpivarski (approved):
This looks great! It seems to have also simplified in the translation. I hadn't been thinking that each node would return the pieces that make up its form, I had been thinking that each node would hold a reference to a part of a RapidJSON DOM, but now that I think of it, the nested contents would get duplicated that way. I like it the way it is.

I also noticed a spelling correction (thanks!) and you're using the British "initialise." I have no objection (never set a spelling standard, and C++ would catch differences; a standard might be needed in Python if flake8 doesn't catch them).

jpivarski:
Oh, you can do the honors and press "squash and merge" (so that I don't press it while you're adding one last commit or something).

agoose77:
Congrats on merging this, big PR!

PR #1050, opened 2021-08-09
Allow to override build date with SOURCE_DATE_EPOCH

bmwiedemann:
Allow to override build date with `SOURCE_DATE_EPOCH`
in order to make builds reproducible.
See https://reproducible-builds.org/ for why this is good
and https://reproducible-builds.org/specs/source-date-epoch/
for the definition of this variable.

This PR was done while working on reproducible builds for openSUSE.

### DIFF
--- a/dev/generate-kernel-signatures.py
+++ b/dev/generate-kernel-signatures.py
@@ -14,7 +14,7 @@ CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 def reproducible_datetime():
 
     build_date = datetime.datetime.utcfromtimestamp(
-        int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+        int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
     )
     return build_date.isoformat().replace("T", " AT ")[:22]
 
--- a/dev/generate-kernel-signatures.py
+++ b/dev/generate-kernel-signatures.py
@@ -4,10 +4,19 @@ from __future__ import absolute_import
 
 import os
 import datetime
+import time
 
 import yaml
 
 CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
+
+
+def reproducible_datetime():
+
+    build_date = datetime.datetime.utcfromtimestamp(
+        int(os.environ.get('SOURCE_DATE_EPOCH', time.time()))
+    )
+    return build_date.isoformat().replace("T", " AT ")[:22]
 
 
 def type_to_ctype(typename):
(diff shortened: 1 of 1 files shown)


dev/generate-kernel-signatures.py
```diff
@@ -4,12 +4,21 @@
...
 
 CURRENT_DIR = os.path.dirname(os.path.realpath(__file__))
 
 
+def reproducible_datetime():
+
+    build_date = datetime.datetime.utcfromtimestamp(
+        int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
+    )
+    return build_date.isoformat().replace("T", " AT ")[:22]
```
agoose77, line 19: Small question (probably @jpivarski), can we just specify [`timespec`](https://docs.python.org/3/library/datetime.html#datetime.datetime.isoformat) instead of truncating to 22 characters? If this time is not parsed anywhere, then `timespec="milliseconds"` would probably suffice?

jpivarski: `timespec` was introduced in Python 3.6, and (at the moment) we're still supporting Python 2.7. The ISO format ensures that cutting the string at a specific number of characters is always correct.

jpivarski:
This seems sensible; I can understand why you'd want the timestamp in the source code to not vary with time. The documentation on reproducible builds talk about reproducing the application binary, and both of these timestamps go into comments, which cannot affect the binaries, but I can see it being good to make it regular, just on principle.

I see that there's a fallback to `time.time()` if `SOURCE_DATE_EPOCH` is not defined, which is good. I don't know under what conditions the variable is defined (only during compilation?), but the fallback just reproduces the previous behavior, so I have no complaints.

Anyway, I'm happy with this PR and will merge it if I'm sure you're done. If you don't plan to make any more changes, let me know in a comment and I'll merge it into the codebase. Thanks!

bmwiedemann:
Being able to trust binaries is the primary goal, but in the context of the openSUSE distribution, we also have python scripts and documentation all bundled into rpm packages and it is easier to reason for trust of our users if we are able to reproduce every bit in our packages and long term even the whole images.

dev/generate-kernel-signatures.py
agoose77: I *always* forget Py2, that's not good!

jpivarski: For Awkward Array, it won't matter next year: #1010.

jpivarski:
@all-contributors please add @bmwiedemann for code

PR #1063, opened 2021-08-18
LayoutBuilder template using either ForthMachine32 or ForthMachine64 and a tutorial

ianna:


docs-src/how-to-create-layoutbuilder.md
```diff
@@ -14,10 +14,91 @@ kernelspec:
...
+builder.tag(1)
+builder.boolean(True)
+builder.tag(0)
+builder.float64(-2.2)
+```
+
+Snapshot
+--------
+
+To turn a #ak::LauoutBuilder into a layout, call `snapshot`. This is an inexpensive operation (may be done multiple times; the builder is unaffacted).
```
agoose77:
```suggestion
To turn a #ak::LayoutBuilder into a layout, call `snapshot`. This is an inexpensive operation (may be done multiple times; the builder is unaffected).
```

agoose77: I took a look over this out of curiosity - let me know if you'd rather wait for review before any suggestions. I can't do much besides code review at the moment so I'm keen to help!

ianna: thanks, I'll pin you when I'm finished!

ianna: @agoose77 - Can I take you up on your kind offer of code review? Thanks!

agoose77:
Absolutely, I'll take a look tomorrow!

ianna:
@jpivarski - I think, I'm done with this PR. It introduces `LayoutBuilder32` and `LayoutBuilder64`, plus the error handling is improved, I think.

docs-src/how-to-create-layoutbuilder.md
```diff
@@ -14,10 +14,303 @@ kernelspec:
...
 
-**This is a stub:** I intend to write this article, but haven't yet.
+What is LayoutBuilder
+---------------------
 
-If you need it soon, create an issue saying so and I'll make it a higher priority.
+`ak.layout.LayoutBuilder` is the low-level LayoutBuilder that builds layouts, or `ak.layout.Content` arrays. It must be initialized by a JSON string that represents a valid `ak.forms.Form`.
 
-[![](img/github-issues-documentation.png)](https://github.com/scikit-hep/awkward-1.0/issues/new?assignees=&labels=docs&template=documentation.md&title=)
+A layout is the composable `ak.layout.Content` elements that determine how an array is structured. The layout may be considered a "low-level" view, as it distinguishes between arrays that have the same logical meaning (i.e. same JSON output and high-level `type`) but different
```
agoose77: Maybe clearer to say
```suggestion
A layout is build from composable `ak.layout.Content` elements that determine how an array is structured. The layout may be considered a "low-level" view, as it distinguishes between arrays that have the same logical meaning (i.e. same JSON output and high-level `type`) but different
```

docs-src/how-to-create-layoutbuilder.md
```diff
@@ -14,10 +14,303 @@ kernelspec:
...
+  * node types, such as `ak.layout.ListArray64` and `ak.layout.ListOffsetArray64`,
+  * integer type specialization, such as `ak.layout.ListArray64` and `ak.layout.ListArray32`,
+  * or specific values, such as gaps in a `ak.layout.ListArray64`.
 
-The text of your issue doesn't have to be much more than a link to this page, so I can be sure which page you're referring to. If you add details about how and why you need it, however, I may be able to tailor the text to help you more.
+`ak.forms.Form` describes a low-level data type or "form". There is an exact one-to-one relationship between each `ak.layout.Content` class and each Form.
```
agoose77, line 27: Maybe lowercase form, or link to `ak.forms.Form` again?
```suggestion
`ak.forms.Form` describes a low-level data type or "form". There is an exact one-to-one relationship between each `ak.layout.Content` class and each form.
```

docs-src/how-to-create-layoutbuilder.md
```diff
@@ -14,10 +14,303 @@ kernelspec:
...
+```
+
+If record contents have the same type, the fields alternate:
+
+```{code-cell}
+builder.float64(1.1)  # "one"
+builder.float64(2.2)  # "two"
+builder.float64(3.3)  # "one"
+builder.float64(4.4)  # "two"
+```
```
agoose77, line 234: Can we still specify the field if we want? If so, could we mention that?

````suggestion
```
However, the fields can be explicitly specified if required:
``{code-cell}
builder.field("two").float64(1.1)  # "two"
builder.field("one").float64(2.2)  # "one"
builder.field("one").float64(3.3)  # "one"
builder.field("two").float64(4.4)  # "two"
```
````

File: src/python/content.cpp [hunk 5]
```diff
@@ -2124,60 +2125,123 @@ getitem<ak::LayoutBuilder>(const ak::LayoutBuilder& self, const py::object& obj)
   return box(unbox_content(::layoutbuilder_snapshot(self.builder(), self.vm().get()->outputs())).get()->getitem(toslice(obj)));
 }
 
-py::class_<ak::LayoutBuilder>
+template <>
+py::object
+getitem<ak::LayoutBuilder64>(const ak::LayoutBuilder64& self, const py::object& obj) {
```
agoose77, line 2130: Would it be reasonable here to just alias the snapshot once, e.g.
```cpp
auto snapshot = box(unbox_content(::layoutbuilder_snapshot(self.builder(), self.vm().get()->outputs())).get();
...
if (...) {
    return snapshot->getitem_range(start, stop);
} else (...) {
    return snapshot->getitem_field(...);
}

docs-src/how-to-create-layoutbuilder.md
jpivarski: I think "is" is correct: a layout is the tree of nested `ak.layout.Content` nodes. To say that it's built from them implies that there's another step, analogous to the step from a Builder (ArrayBuilder or LayoutBuilder) to an array through `snapshot`.

"Layout" may simply be a synonym for `Content`, though it may be distinguished in that "layout" often refers to the whole tree and a `Content` object is specifically one node, since it has a meaning in code. Maybe that's what you meant by "built from."

In that case, how about "A layout consists of composable `ak.layout.Content`..."? The word "consists" is describing a static thing, whereas "built from" could describe a static thing or could describe a process like the one involving `snapshot`.

docs-src/how-to-create-layoutbuilder.md
jpivarski: I'd prefer capitals on "Form" because it's such a common word and easily mis-seen or even mistyped as "from." I try to use capitals with "Form" because it clues people in to "this is some special thing with a specific meaning."

All of these references to classes could point to the documentation: instead of "Form" or even "`ak.forms.Form`", you could say [ak.forms.Form](https://awkward-array.readthedocs.io/en/latest/ak.forms.Form.html). Similarly with their subclasses, including Content subclasses like [ak.layout.ListArray](https://awkward-array.readthedocs.io/en/latest/ak.layout.ListArray.html). I think the other tutorials do this.

docs-src/how-to-create-layoutbuilder.md
jpivarski: I don't remember whether `builder.field("xyz")` acts as an assertion or not. It will not allow you to write them in a different order: that's a performance thing.

jpivarski (approved):
The documentation looks good to me, and I'm glad that writing it inspired some tweaks to the code. The references to class names could be links, as I've said in a comment above, and that would be an improvement.

Thanks @agoose77 for the review! You brought up some good points.

jpivarski:
Oh, and I'm holding off on merging until you decide what you want to do about @agoose77's comments, @ianna. Although when you've addressed everything, you can do the merge. That's what my sign-off above means.

docs-src/how-to-create-layoutbuilder.md
agoose77: That is better still (and doesn't have a typo)!

ianna: Thanks! It's fixed.

docs-src/how-to-create-layoutbuilder.md
ianna: No, there is no command to set a field. I've changed the wording to make it more clear. Thanks!

src/python/content.cpp
ianna: Thanks, it's a good idea! I'll be revisiting this (and ArrayBuilder) code for v2. I think, there will be fewer boxing/unboxing needed - if any.

docs-src/how-to-create-layoutbuilder.md
ianna: @agoose77 - Thanks for the review - very much appreciated

docs-src/how-to-create-layoutbuilder.md
agoose77: Ah, good to know!

PR #1065, opened 2021-08-18
Unit Tests

SantamRC:
To run :
```
python dev/generate-tests.py
python -m pytest -vv unit-tests
```


dev/generate-tests.py
```diff
@@ -692,8 +700,46 @@ class Error(ctypes.Structure):
...
+                        file.write("\tassert " + key + " == " + "pytest_" + key + "\n")
+                file.write("\n\n")
+
+
 if __name__ == "__main__":
     specdict = readspec()
     genspectests(specdict)
     gencpukerneltests(specdict)
     gencudakerneltests(specdict)
+    generateTests()
```
ianna: @SantamRC - this should probably follow previous function names convention: `genunittests`?

test-data.yml
```diff
@@ -0,0 +1,2326 @@
...
+        starts_out: [0, 1, 1, 1, 1, 1, 1, 1]
+        stops_out: [1, 1, 1, 1, 1, 1, 1, 1]
+  - name: awkward_content_reduce_zeroparents_64
+    tests:
+    - error: true
+      inputs: {}
+      outputs:
+        toparents: [0]
+    - error: false
+      inputs: {length: 10}
```
ianna: btw, why does this input look different?

test-data.yml
```diff
@@ -0,0 +1,2326 @@
...
+        length: 8
+        offsets_in: [0, 1, 1, 1, 1, 1, 1, 1]
+      outputs:
+        mask_out: [0, 1, 2, 3, 4, 5, 6, 7]
+        starts_out: [0, 1, 1, 1, 1, 1, 1, 1]
+        stops_out: [1, 1, 1, 1, 1, 1, 1, 1]
+  - name: awkward_content_reduce_zeroparents_64
+    tests:
+    - error: true
+      inputs: {}
```
ianna: I think, `length` in this case is `0` and the test should pass.

ianna (commented):
It looks great, thanks! I've added some minor comments.

Also from pre-commit:
`dev/generate-tests.py:728:34: E712 comparison to True should be 'if cond is True:' or 'if cond:'`

test-data.yml
SantamRC: The format is generated like this. Looking into it.

test-data.yml
SantamRC: The definition is,
```
def awkward_content_reduce_zeroparents_64(toparents, length):
    for i in range(length):
        toparents[i] = 0
```
So `length=0` will not change `toparents` but if `length=1` then `toparents=[0]` and the test will pass. Should I put the length as 0 or 1 or should I keep it as an error case?

jpivarski:
@all-contributors please add @SantamRC  for test

PR #1074, opened 2021-08-25
C++ refactoring: Implementing combinations

sw5sh:
This PR implements the `_combinations` function for all the Awkward Array types except `VirtualArray`.

* `_combinations` implementation
- [x] BitMaskedArray
- [x] ByteMaskedArray
- [x] EmptyArray
- [x] IndexedArray
- [x] IndexedOptionArray
- [x] ListArray
- [x] ListOffsetArray
- [x] NumpyArray
- [x] RecordArray
- [x] RegularArray
- [x] UnionArray
- [x] UnmaskedArray

* `_combinations` tests
- [x] BitMaskedArray
- [x] ByteMaskedArray
- [x] EmptyArray
- [x] IndexedArray
- [x] IndexedOptionArray
- [x] ListArray
- [x] ListOffsetArray
- [x] NumpyArray
- [x] RecordArray
- [x] RegularArray
- [x] UnionArray
- [x] UnmaskedArray


src/awkward/_v2/contents/content.py
```diff
@@ -504,6 +507,75 @@ def _localindex_axis0(self):
...
+                len(self),
+                1,
+            )
+        )
+        contents = []
+        for ptr in tocarry:
+            contents.append(ak._v2.contents.IndexedArray(ptr, self))
+        return ak._v2.contents.recordarray.RecordArray(
+            contents, recordlookup, parameters=parameters
+        )
```
sw5sh: Hi @jpivarski , could you please help me by reviewing this function? The corresponding C++ code is [here](https://github.com/scikit-hep/awkward-1.0/blob/b1e342d2d22818daa8d63fa21fe163902b9cb94e/src/libawkward/Content.cpp#L1322-L1384). Thank You!

jpivarski: <div>Anish  10:20 AM</div>
<div>Hi Jim, before moving further ahead with combinations it would help me out a lot if you could review on a function I wrote. I have put it as a comment on GitHub. Just a bit of what I am trying to do here, The C++ code uses a pointer to pointer to do things, and I am trying to abstract everything into python lists and nplike arrays for the python version.</div><br>

<div>Jim Pivarski  10:22 AM</div>
<div>Is the reduce doing array-scale iteration in Python?</div></br>

<div>10:22</div>
<div>How large can thisn get?</div></br>

<div>10:23</div>
<div>Oh, yeah, that's safe. I see it's being iterated over in C++.</div></br>

<div>Anish  10:23 AM</div>
<div>thisn is the n of the combinations.</div></br>

<div>Jim Pivarski  10:23 AM</div>
<div>(The C++ code wasn't allowed to iterate over array-scale things, either.)</div></br>

<div>10:24</div>
<div>This is a tricky one, with that array of pointers.</div></br>

<div>Anish  10:24 AM</div>
<div>Yeah, that's why I wanted someone else to look at it too :P</div></br>

<div>Jim Pivarski  10:24 AM</div>
<div>Right. Good call.</div></br>

<div>10:25</div>
<div>That n is also not array-scale: it's passed in by the user, rarely larger than 2, but we want to at least be able to accept generic n.</div></br>

<div>10:25</div>
<div>As I've said elsewhere, the GPU version is only going to accept n<=5 for purely mathematical reasons!</div></br>

<div>10:26</div>
<div>So it makes n Indexes in a list tocarry. That's like the C++ std::vector.</div></br>

<div>Anish  10:26 AM</div>
<div>Yup, the malloc is abstracted into a Index64 type.</div></br>

<div>Jim Pivarski  10:27 AM</div>
<div>Yeah, I was just looking at that. Technically, the type is size_t, but it looks like we're up-casting 32-bit systems to a common size, and will be in trouble if 128-bit systems ever become common.</div></br>

<div>Anish  10:28 AM</div>
<div>I think if that ever happens, the kernels itself will need changing too, right?</div></br>

<div>Jim Pivarski  10:28 AM</div>
<div>Where's the definition of NumpyKernel? I want to check on the definition of _cast.</div></br>

<div>10:29</div>
<div>Most of the Awkward kernels won't need to be changed if addressing becomes 128-bit, but this one will. However, most of the rest of the world will be going through similar pains, so it's not a major worry. Maybe there should just be a comment about it.</div></br>

<div>Anish  10:30 AM</div>
<div>Here's _cast</div></br>

https://github.com/scikit-hep/awkward-1.0/blob/b1e342d2d22818daa8d63fa21fe163902b9cb94e/src/awkward/nplike.py#L365-L372

<div>Jim Pivarski  10:30 AM</div>
<div>Thanks. You're calling it without a second argument; how is that possible?</div></br></div></br>

<div>Jim Pivarski  10:32 AM</div>
<div>Is it just not being tested yet?</div></br>

<div>Anish  10:32 AM</div>
<div>Oh, my bad. I haven't tested the function yet. It should by POINTER(ctypes.c_int64). My bad I'll correct that</div></br>

<div>Jim Pivarski  10:33 AM</div>
<div>That looks right. That would give you a pointer to the underlying data.</div></br>

<div>10:34</div>
<div>This looks good! I think it's what you want, though there may be minor issues when you actually do try to test it. You may need to do a lot of print-outs of pointers before you get it working.</div></br>

<div>Anish  10:35 AM</div>
<div>Understood. Thanks! I'll go ahead with this for now and catch the pointer bugs during testing</div></br>

<div>Jim Pivarski  10:35 AM</div>
<div>My only comment is a minor one: the use of reduce is nice and functional and all, but doesn't make it clearer to me what's going on. I think I'd prefer the plain for loop as in C++, just for readability. That's a minor comment, since some people might find the reduce version easier to read—it could be a toss-up.</div></br>

<div>Anish  10:36 AM</div>
<div>Oh okay, I intentionally did that, because Python for loops are slow, and I read a StackOverflow answer where this guy compares reduce to for loop and it's very fast compared to it.</div></br>

<div>10:38</div>
<div>I think I'll revert to a for loop since thisn can't be very large anyways</div></br>

<div>Jim Pivarski  10:39 AM</div>
<div>For us, the dividing line between fast and slow is whether something's in a kernel or not, and then we just need to make sure that anything array-scaled (len of any Index or Content) is in a kernel. The rest we can presume to be slow "O(1) bookkeeping" and not worry about optimizing. In particular, deciding between an explicit loop and functional reduce shouldn't be made on the basis of performance (and performance measurements for reducing a large list probably doesn't apply to our short ones, anyway).</div></br>

<div>10:42</div>
<div>The benefit of having a rule for what gets optimized and what doesn't ("all array-scale iterations must be in kernels") is that everything that doesn't come under this rule can be completely relaxed about performance, allowing us to focus on readability for these parts.
That said, sometimes functional or at least list comprehension syntax is more readable, but subjectively, to me in this instance, reduce isn't helping to make it clearer.</div></br>

<div>10:43</div>
<div>So, would it be okay if I copy-paste this Slack conversation into GitHub? Ideally, we would have had this back-and-forth exchange in public, but Slack is faster. I would start from 20 minutes ago.</div></br>

<div>Anish  10:49 AM</div>
<div>Yeah sure, no problem. And thanks for the comparison between readibility vs performance.</div></br>

src/awkward/_v2/contents/content.py
```diff
@@ -504,6 +504,68 @@ def _localindex_axis0(self):
...
+        fromindex = ak._v2.index.Index64.empty(n, self.nplike, dtype=np.int64)
+
+        self._handle_error(
+            self.nplike[
+                "awkward_RegularArray_combinations_64",
+                np.int64,
+                toindex.to(self.nplike).dtype.type,
+                fromindex.to(self.nplike).dtype.type,
+            ](
+                tocarryraw.data,
```
jpivarski: You don't need to put `nplike_tocarryraw` into `tocarryraw` just to take it out again as `tocarryraw.data`.

src/awkward/_v2/contents/listarray.py
```diff
@@ -425,3 +426,91 @@ def _localindex(self, axis, depth):
...
+            self._handle_error(
+                self.nplike[
+                    "awkward_ListArray_combinations",
+                    np.int64,
+                    toindex.to(self.nplike).dtype.type,
+                    fromindex.to(self.nplike).dtype.type,
+                    self._starts.to(self.nplike).dtype.type,
+                    self._stops.to(self.nplike).dtype.type,
+                ](
+                    tocarryraw.data,
```
jpivarski: Same point about `nplike_tocarryraw` and `tocarryraw`. I'm just trying to reduce indirection so that this will be easier to read.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -458,3 +458,95 @@ def _localindex(self, axis, depth):
...
+            self._handle_error(
+                self.nplike[
+                    "awkward_ListArray_combinations",
+                    np.int64,
+                    toindex.to(self.nplike).dtype.type,
+                    fromindex.to(self.nplike).dtype.type,
+                    starts.to(self.nplike).dtype.type,
+                    stops.to(self.nplike).dtype.type,
+                ](
+                    tocarryraw.data,
```
jpivarski: Again with `nplike_tocarryraw` and `tocarryraw`.

src/awkward/_v2/contents/regulararray.py
```diff
@@ -459,3 +459,87 @@ def _localindex(self, axis, depth):
...
+
+            if self._size != 0:
+                self._handle_error(
+                    self.nplike[
+                        "awkward_RegularArray_combinations_64",
+                        np.int64,
+                        toindex.to(self.nplike).dtype.type,
+                        fromindex.to(self.nplike).dtype.type,
+                    ](
+                        tocarryraw.data,
```
jpivarski: Again with `nplike_tocarryraw` and `tocarryraw`.

jpivarski (changes requested):
We talked on Slack (code reviews are much better when interactive, but unfortunately, we don't get a public log). Some of the things I was worried about were unfounded (declaring the Index as `np.intp` is exactly the right thing to do, and we do in fact need that `ptr` property, at least for this exceptional double-pointer case). There were only a few remaining items in an overall excellent PR.

  - [x] centralize the `n < 1` check in `Content.combinations`, so that it doesn't have to be in every `_combinations` method
  - [x] see if `ListArray._combinations` can reuse `ListOffsetArray._combinations` by calling it with itself (its `self`) as the `self` argument, and if so, do so
  - [x] add a `pytestmark` to skip the tests if Python 2

That's it! When you're done, let me know that you're not planning on committing any more changes and I'll merge it when the tests pass.

Thanks again; this is great!

src/awkward/_v2/contents/listoffsetarray.py
sw5sh: My bad. Rectified this in the most recent commit.

src/awkward/_v2/contents/regulararray.py
sw5sh: Yup, rectified the issue.

sw5sh:
I am done with all the changes, so you can merge this, after the tests pass. Thank You!

jpivarski (approved):

PR #1079, opened 2021-08-31
C++ refactoring: Implementing VirtualArray in Awkward v2.

jpivarski:


jpivarski:
@ioanaif, I'm telling pytest to skip all _v2 related code in Python 2.7: see 623508f554e250e20a857b9a12dfd212d4c8a03a . You might want to include this one commit in your PR, too, if it passes tests. That way, you don't have to skip individual lines or functions that fail in Python 2.7.

Discussion #1010 says we'll be dropping Python 2.7 support by the end of this year, but maybe the more appropriate dividing line would be 1.x vs 2.x.

ioanaif:
Great! I'll add it to the PR.

PR #1081, opened 2021-09-03
Data for Remaining Kernel Functions

SantamRC:
To run
```
python dev/generate-tests.py
python -m pytest tests-spec-explicit
```


jpivarski (approved):
This is looking good! I ran all of the tests and checked the YAML; there were no errors or linting warnings. I saw @ianna's comment that there are some duplicates, so I did a quick check. Right before @SantamRC's latest commit, I found a lot of them, now there are none:

```python
>>> import yaml, json
>>> test_data = yaml.safe_load(open("test-data.yml"))
>>> for kernel in test_data["unit-tests"]:
...     signatures = set()
...     for test in kernel["tests"]:
...         signature = json.dumps(test)
...         if signature in signatures:
...             print(kernel["name"], signature)
...         signatures.add(signature)
... 
```

This wouldn't catch cases in which the input or output argument names are listed in a different order, but I'm going to assume that this doesn't happen. (The arguments have a natural ordering; I'll assume they're all in that order.)

As it is now, I'd say this is good and ready to merge. After this PR, I'd like to change the name from "test-data.yml" to something that doesn't start with "test" because I get into the tests directory by tab-completion. I'll probably rename it to "kernel-tests.yml", but that's after merging.

Also, all the tests pass (on my local machine) and the linters are happy.

Are you done now, @SantamRC? Any reason I shouldn't merge it?

Thanks!

SantamRC:
Yeah @jpivarski I am done working on this PR. Also the arguments are all placed in the same order ( Alphabetical ).
Thanks!!

PR #1085, opened 2021-09-08
Fix: add utility to check whether a string is a filepath

### LINKED ISSUE #1084: "path too long for Windows" when calling `ak.from_json` with a long json string on Windows
### Version of Awkward Array

not sure

Discovered by a student in a course (who was working on Windows). We were going through the bikeroutes example:

```python
import awkward as ak
import urllib
url = "https://raw.githubusercontent.com/Chicago/osd-bike-routes/master/data/Bikeroutes.geojson"
bikeroutes_json = urllib.request.urlopen(url).read()
bikeroutes = ak.from_json(bikeroutes_json)
```

Gave him the following error:

```
ValueError Traceback (most recent call last)
<ipython-input-25-171b38169686> in <module>
----> 1 bikeroutes = ak.from_json(bikeroutes_json)

~\Anaconda3\lib\site-packages\awkward\operations\convert.py in from_json(source, nan_string, infinity_string, minus_infinity_string, complex_record_fields, highlevel, behavior, initial, resize, buffersize)
1054 is_path, source = ak._util.regularize_path(source)
1055
-> 1056 if os.path.isfile(source):
1057 layout = ak._ext.fromjsonfile(
1058 source,

~\Anaconda3\lib\genericpath.py in isfile(path)
28 """Test whether a path is a regular file"""
29 try:
---> 30 st = os.stat(path)
31 except OSError:
32 return False

ValueError: stat: path too long for Windows
```

unfortunately i currently have no windows Installation available to reproduce it

agoose77:
I didn't follow the `isint` / `isstr` convention: I think we should rename those in a subsequent PR

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -5,6 +5,7 @@ from __future__ import absolute_import
 import re
 import sys
 import os
+import os.path
 import warnings
 import itertools
 import numbers
@@ -39,6 +40,13 @@ kMaxUInt32 = 4294967295  # 2**32 - 1
 kMaxInt64 = 9223372036854775806  # 2**63 - 2: see below
 kSliceNone = kMaxInt64 + 1  # for Slice::none()
 kMaxLevels = 48
+
+
+def is_file_path(x):
+    try:
+        return os.path.isfile(x)
+    except OSError:
+        return False
 
 
 def isint(x):
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -45,7 +45,7 @@ kMaxLevels = 48
 def is_file_path(x):
     try:
         return os.path.isfile(x)
-    except OSError:
+    except ValueError:
         return False
 
 
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -1075,7 +1075,7 @@ def from_json(
 
     is_path, source = ak._util.regularize_path(source)
 
-    if os.path.isfile(source):
+    if ak._util.is_file_path(source):
         layout = ak._ext.fromjsonfile(
             source,
             nan_string=nan_string,
--- a/tests/test_1084-from-json-win-path.py
+++ b/tests/test_1084-from-json-win-path.py
@@ -7,13 +7,13 @@ import urllib.request
 import urllib.error
 
 
-@pytest.mark.skipif(not ak._util.win, reason="requires Windows")
+@pytest.mark.skipif(ak._util.win, reason="requires Windows")
 def test():
-    url = "https://raw.githubusercontent.com/Chicago/osd-bike-routes/5f556dc/data/Bikeroutes.geojson"
+    url = "https://raw.githubusercontent.com/Chicago/osd-biske-routes/5f556dc/data/Bikeroutes.geojson"
     try:
         bikeroutes_json = urllib.request.urlopen(url).read()
     except urllib.error.URLError:
-        pytest.skip(reason="couldn't download sample dataset")
+        pytest.skip(msg="couldn't download sample dataset")
 
     # This shouldn't fail (see #1084)
     ak.from_json(bikeroutes_json)
--- a/tests/test_1084-from-json-win-path.py
+++ b/tests/test_1084-from-json-win-path.py
@@ -9,7 +9,7 @@ import urllib.error
 
 @pytest.mark.skipif(not ak._util.win, reason="requires Windows")
 def test():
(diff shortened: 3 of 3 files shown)


agoose77:
This might fix #1084 (but we need to add a test)

jpivarski:
There's an `ak._util.iswin` to check to see if you're on Windows and we know that the bikeroutes JSON is too long to be a Windows path. Such a test would require an HTTP connection, which should probably skip if the attempt to download with

```python
import urllib.request

url = "https://raw.githubusercontent.com/Chicago/osd-bike-routes/master/data/Bikeroutes.geojson"
bikeroutes_json = urllib.request.urlopen(url).read()
```

fails.

agoose77:
> There's an `ak._util.iswin` to check to see if you're on Windows and we know that the bikeroutes JSON is too long to be a Windows path. Such a test would require an HTTP connection, which should probably skip if the attempt to download with
> 
> ```python
> import urllib.request
> 
> url = "https://raw.githubusercontent.com/Chicago/osd-bike-routes/master/data/Bikeroutes.geojson"
> bikeroutes_json = urllib.request.urlopen(url).read()
> ```
> 
> fails.

Let's use that instead — it's a more direct test.

jpivarski:
I didn't merge this because (as I understand it), it's waiting for a test. If I should add it to 1.5.0, let me know. Otherwise, there will probably be only one release candidate before the final release (but technically two "rc" releases in GitHub because it looks like the first one didn't trigger).

jpivarski:
Nope; that's because Azure isn't configured to deploy anymore—it's handled by GitHub Actions now, and I had forgotten.

jpivarski:
Oh, I was thinking that this could just be a test that attempts to download through HTTP, and skips if it fails. I don't think it would be the only test that requires network access.

That way, we don't need to add a copyright message for Chicago's data. We're really just wanting it for its size (which could also be an auto-generated JSON, if we really want to avoid network access).

agoose77:
I've added a test, but I can't test the previous behaviour against this test unless I poke around with the CI. Do you happen to have a Windows machine to hand?

agoose77:
> I've added a test, but I can't test the previous behaviour against this test unless I poke around with the CI. Do you happen to have a Windows machine to hand?



> Oh, I was thinking that this could just be a test that attempts to download through HTTP, and skips if it fails. I don't think it would be the only test that requires network access.
> 
> That way, we don't need to add a copyright message for Chicago's data. We're really just wanting it for its size (which could also be an auto-generated JSON, if we really want to avoid network access).

Right, I forgot to motivate why I downloaded a sample; the data are quite large. It's not a *slow* download, but it's enough that I don't think we want to add it to the test suite unless required. Are you not comfortable with including the license file? It's not infectious, and it's only in the test suite.

jpivarski:
> Do you happen to have a Windows machine to hand?

The CI is our only access (both of us).

> Right, I forgot to motivate why I downloaded a sample; the data are quite large. It's not a slow download, but it's enough that I don't think we want to add it to the test suite unless required.

I considered it small enough to use in a classroom setting, and the downloads in CI are a lot faster than they are at home. (The CI machines are well-connected!) I just downloaded it in 0.3 seconds at home.

Generating simple JSON that is as large as the bikeroutes file would be sufficient. The bikeroutes sample is 2.3 MB, and a list of numbers up to 1 million is 7.8 MB:

```python
json.dumps(np.arange(1000000).tolist())
```

Generating it and interpreting it as an Awkward Array takes 0.2 seconds:

```python
ak.Array(json.dumps(np.arange(1000000).tolist()))
```

Without your fix, the above line should fail in Windows (CI).

agoose77:
If we want to test the Windows case, then we only need 260 chars of JSON, although I can see the merit in re-using a "large" dataset. I will update with your suggestions.

jpivarski:
Oh yeah, and although the license is not infectious, simplifying things to the extent that no one has to evaluate whether that is the case is beneficial. I have a bias against including anything that requires a license, which can be overcome by a good enough reason: Lark has a good enough reason, as does pybind11, as well as RapidJSON... I just want to keep the list short, and this one test is doable without having to add the dataset.

jpivarski:
> 260 chars of JSON

I didn't know the threshold, but I did know that the bikeroutes exceeded it. That's the thinking behind the 7 MB of JSON.

agoose77:
Oh, I ignored py2 here because I was thinking about v2. However, it makes sense to merge this now - thanks for backporting it

PR #1089, opened 2021-09-12
Revert 'NumPy' spelling to 'Numpy' in code only.

jpivarski:


agoose77:
Hey Jim, what's the motivation for this? Are many relying on these APIs?

jpivarski:
As it turns out, there are. Technically, we shouldn't have changed this spelling without a deprecation cycle because it's not a hidden (underscored) interface, so the fact that Coffea was using it (see IRIS-HEP Slack #coffea) is legitimate.

Since there haven't been little releases since 1.4.0, this change comes as a "big bang" to them. The motivation for the breaking change ("change spelling to be the way the NumPy team capitalizes for `nplike` singletons but not NumpyArray, NumpyForm, and NumpyType") was not strong enough to cause that pain: they'd have to update Coffea and put in a version dependence, and then we don't know how many data analyses also access `nplike` directly (I didn't _expect_ them to...).

If there's some pain point in the other direction: someone (you?) have code that relies on the "NumPy" spelling, then we can expose both spellings in this PR. But since the "NumPy" spelling was never in any non-RC release, it seems safe to me to not introduce such duplication.

I've disabled auto-merge for now, but if there's no problem, I'll reenable it.

lgray:
Not so much a big bang to coffea in particular - names can be fixed. The major problem is that if this change in capitalization were introduced it would break all previous versions of coffea currently in "production" (i.e. actively used in analysis). That would be Very Bad. 

If you'd really like this spelling change I'm completely for doing it, we just have to be mindful of the analysts (including ourselves) when rolling it out.

jpivarski:
The appropriate way to roll it out would be to have both spellings for a while and anything that uses the old spelling should get warnings with a deadline for when it would be removed.

But the main thing is that the motivation for changing it wasn't very strong—for something even as painful as a deprecation cycle (better than a sudden breaking change!), someone would have to be asking for it. That wasn't the case: @henryiii added a spell-checker, I thought it was a good idea, and I saw that it made all the "NumPy" spellings consistent. Then I decided to extend that consistency to the `nplike` singletons, which I had in my mind as "internal," even though they don't start with an underscore. Nobody asked for this change, and it's even arguable that it shouldn't happen because we never considered changing NumpyArray, NumpyForm, and NumpyType. It is especially true that NumpyArray is part of the public API (for developers of downstream libraries), even though it's not the recommended data analyst user interface.

So I'll merge this and make an RC2 release. We still have time before the final release to expose both spellings, if there's any reason for it, but I don't think there is.

PR #1092, opened 2021-09-13
C++ refactoring: argsort

ianna:


src/awkward/_v2/contents/indexedoptionarray.py
```diff
@@ -314,11 +314,272 @@ def _localindex(self, axis, depth):
...
+                self._index.to(nplike),
+                len(self._index),
+                parents.to(nplike),
+                starts.to(nplike),
+            )
+        )
+
+        ind = ak._v2.contents.NumpyArray(nulls_index)
+
+        # # FIXME: work around for if mergeable: test it on two 1D arrays
```
ianna, line 448: @jpivarski - this is a workaround in absence of `merge`

ianna (commented):
@jpivarski - I think, it's done for now. Please, review when you have time. Thanks.

src/awkward/_v2/contents/content.py
```diff
@@ -588,6 +588,72 @@ def min(self, axis=-1, mask=True, keepdims=False, initial=None):
...
+            kind ({‘quicksort’, ‘mergesort’, ‘heapsort’, ‘stable’}, optional):
+                an optional sorting algorithm. This is similar to NumPy
+                sort.
+            order (str or list of str, optional): similar to NumPy.
+
+        Puts indices of Content elements in an ordered sequence: numeric or lexicographical,
+        ascending or descending.
+
+        This function is similar to NumPy [sort](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html).
+        """
```
jpivarski: This isn't the high-level function, so it's not the place for the full docstring. We shouldn't be obligated to put such nice docstrings on all of these `Content` methods, so this will get removed eventually, or reduced to a minimal level of docstring that we're able to maintain on all of the `Content` methods uniformly. I'm going to leave it here for now because this information will be needed in the high-level function, when that gets updated.

src/awkward/_v2/contents/bytemaskedarray.py
```diff
@@ -325,32 +325,40 @@ def _localindex(self, axis, depth):
...
-                self._mask.to(nplike),
-                len(self._mask),
-                self._valid_when,
-            )
-        )
-        return ak._v2.contents.IndexedOptionArray(
-            index,
-            self._content,
-            self._identifier,
-            self._parameters,
```
jpivarski, line 347: At first, I was worried about this function disappearing, but then I saw that `toIndexedOptionArray`/`toIndexedOptionArray64` is a duplication in the main branch. Thanks for removing it here!

src/awkward/_v2/contents/content.py
```diff
@@ -588,6 +588,72 @@ def min(self, axis=-1, mask=True, keepdims=False, initial=None):
...
+        else:
+            if negaxis <= 0:
+                negaxis = negaxis + depth
+            if not (0 < negaxis and negaxis <= depth):
+                raise ValueError(
+                    "axis="
+                    + str(axis)
+                    + " exceeds the depth of the nested list structure "
+                    "(which is " + str(depth) + ")"
+                )
```
jpivarski: These kinds of things might be better as `.format(...)` than `+`. For anything but the most minimal cases, `.format(...)` is easier to read than `+`, and it makes conversions like `str(axis)` implicit. It's not reason enough to change this now, though.

src/awkward/_v2/contents/indexedoptionarray.py
jpivarski: Okay. The use of the searchable "`FIXME`" text is good; we'll look for it after `mergeable` is done. Verifying that your tests for argsort still work after this block is replaced by `mergeable` will be a good test that we all understand and agree on what `mergeable` is supposed to do.

src/awkward/_v2/contents/indexedoptionarray.py
```diff
@@ -341,9 +602,9 @@ def _sort_next(
         )
 
         next_length = index_length - numnull[0]
-        nextparents = ak._v2.index.Index64.zeros(next_length, nplike)
-        nextcarry = ak._v2.index.Index64.zeros(next_length, nplike)
-        outindex = ak._v2.index.Index64.zeros(index_length, nplike)
+        nextparents = ak._v2.index.Index64.empty(next_length, nplike)
+        nextcarry = ak._v2.index.Index64.empty(next_length, nplike)
+        outindex = ak._v2.index.Index64.empty(index_length, nplike)
```
jpivarski, line 607: This and the above is a good catch. I also use `zeros` (or more often `ones` or `full` with value `123`) for debugging, but then we want to switch over to `empty` for production code.

tests/test_1072_v2-sort.py
```diff
@@ -69,14 +77,21 @@ def test_masked():
         [1, 2, None, None, None],
     ]
```
jpivarski, line 78: Do you know why it's saying "'ListOffsetArray' has no 'mask'"? That's a bad sign, that some code thinks it's working with a ByteMaskedArray (or less likely, a BitMaskedArray), but it actually has a ListOffsetArray.

At first, I thought this was a regression, since the `pytest.mark.skip` is new, but now I see that it's replacing a commented-out version, which is always better. At least we'll see the fact that there's something wrong here in the list of skipped tests.

jpivarski (approved):
I think it's ready to be merged. I made some comments inline about things that can be tidied up, but they're not necessary for this PR. I'll let you do the merge so that you can decide how much to add, if anything.

tests/test_1072_v2-sort.py
ianna: I think, `mask` suppose to be applied at a high level `ak.Array`, not a layout. I left the test as a reminder to check it after we promote v2 to a high level. I’ll check the layout before I do a v1 to v2 conversion.

jpivarski: If any object is passed into the `ak.mask` function, `ak.to_layout` will attempt to convert it into a layout, which means stripping off an `ak.Array` if that's what the object is, a no-op if it's already a layout, and a conversion if it's something else. However, it might not be able to do the conversion if it's a v2 layout (i.e. if `v1_to_v2` is not called in `ak.to_layout`).

src/awkward/_v2/contents/content.py
ianna: done

src/awkward/_v2/contents/bytemaskedarray.py
ianna: thanks!

src/awkward/_v2/contents/content.py
ianna: done

PR #1095, opened 2021-09-14
C++ refactoring: Prepare Awkward for Dask (kernel-level DAG)

jpivarski:


jpivarski:
I used "cherry-pick" to select the commits from this PR that are relevant beyond the prototype. Since the "kernel-level thing" will not be a DAG (https://github.com/ContinuumIO/dask-awkward/discussions/2), we don't want all of these changes, but a few of them are general fix-ups, not specific to the prototype.

PR #1101, opened 2021-09-20
C++ Refactoring: Implement validityerror for all array types.

sw5sh:


jpivarski (approved):
Looks great, except that you were doing comparisons that looked like this:

```python
self.parameter("__array__") == '"string"'
```

That was correct in C++ because all of the parameters were JSON-quoted. In principle, parameters can include lists and dicts (though none of them yet have), so they had to be passed around in C++ in JSON serialization. Therefore, a parameter that is just a string, such as `"string"`, would actually show up in C++ as `"\"string\""`. In principle, there could be spaces around it, like `"   \"string\" "`, but JSON stringifiers (such as Python's `json.dumps`) wouldn't do that. More general tests of parameter equality converted the data into JSON documents (using RapidJSON), but these particular tests didn't need that, and so they were simple in C++ (no dicts to be out of order or potential spaces between elements).

In Python, however, we're leaving any lists and dicts in parameters as they are, not serialized into JSON. Thus, the correct comparison is to `"string"`, not `'"string"'`.

Just in case @ioanaif or @ianna run into this, I'm pinging you here.

PR #1108, opened 2021-09-27
Fixed #1071: mask_identity=False should not return option type.

### LINKED ISSUE #1071: `mask_identity=False` should never introduce option-types
### Version of Awkward Array

1.4.0

Given this array:
```python3
layout = ak.layout.ListOffsetArray64(
    ak.layout.Index64(np.r_[0, 1]),
    ak.layout.IndexedArray64(
        ak.layout.Index64(np.r_[0,1,2,3]),
        ak.layout.RegularArray(
            ak.layout.NumpyArray(np.r_[1,2,3,4]),
            4
        )
    )
)
array = ak.Array(layout)
```
the result of reducing with `mask_identity=False` is an array with an option type:
```pycon
>>> ak.argmin(array, axis=-1, mask_identity=False)
<Array [[0]] type='1 * var * ?int64'>
```

If I were to hazard a guess, I suspect it is something to do with the way we handle indexed types vs indexed option types.

jpivarski:


jpivarski:
Same thing in v2: https://github.com/scikit-hep/awkward-1.0/pull/1099#discussion_r717090955.

PR #1111, opened 2021-10-06
C++ refactoring: unique and is_unique

ianna:


jpivarski:
In PR #1117 (which will auto-merge soon), all v2 tests are being moved into a new `tests/v2 directory`, and the "v2" qualifier is being removed from their filenames. Please do that for the tests in this PR, too!

The other directory restructuring that I'll be doing today will be to copy high-level code into `src/awkward/v2`. At first, it will be commented out, and then I'll slowly uncomment things as they're made to work with v2 only. But since that involves new files, it won't interfere with any of your ongoing work.

ianna:
@jpivarski - there are a couple of places where for loops should be replaced by newly written kernel functions - they are marked with FIXME. Apart from that I think, it's done.

src/awkward/_v2/contents/content.py
```diff
@@ -924,6 +924,51 @@ def validityerror(self, path="layout"):
...
+                            )
+                        )
+                else:
+                    if negaxis <= 0:
+                        negaxis = negaxis + depth
+                    if not (0 < negaxis and negaxis <= depth):
+                        raise ValueError(
+                            "axis={0} exceeds the depth of the nested list structure "
+                            "(which is {1})".format(axis, depth)
+                        )
```
jpivarski: All errors in which a user specifies an unusable axis (e.g. the three above) should use `np.AxisError` now—it's a convention I'm trying to establish, and in overlaps with NumPy, it's what NumPy does.

The "axis too deep" error should have this text:

```python
np.AxisError("axis={0} exceeds the depth of this array ({1})".format(axis, depth))
```

Cross-linking the comment to @stormiestsin: https://github.com/scikit-hep/awkward-1.0/pull/1135#discussion_r744811036

src/awkward/_v2/contents/content.py
```diff
@@ -924,6 +924,51 @@ def validityerror(self, path="layout"):
...
+
+            starts = ak._v2.index.Index64.zeros(1, self.nplike)
+            parents = ak._v2.index.Index64.zeros(len(self), self.nplike)
+
+            return self._unique(negaxis, starts, parents, 1)
+
+        raise ValueError(
+            "unique expects axis 'None' or '-1', got axis={0} that is not supported yet".format(
+                axis
+            )
```
jpivarski: I missed this one; it's another `np.AxisError`. (There might be others, in addition to what I catch.)

src/awkward/_v2/contents/indexedoptionarray.py
```diff
@@ -578,6 +578,210 @@ def _localindex(self, axis, depth):
...
+            for i in range(len(starts)):
+                for _j in range(out._offsets[i], out._offsets[i + 1]):
+                    newindex[k] = ll
+                    k += 1
+                    ll += 1
+                if parents[starts[i]] in nulls:
+                    newindex[k] = -1
+                    k += 1
+                    shift += 1
+                newoffsets[i + 1] = out._offsets[i + 1] + shift
```
jpivarski: Reminder to move to a kernel (can resolve this comment when done).

src/awkward/_v2/contents/indexedoptionarray.py
```diff
@@ -578,6 +578,210 @@ def _localindex(self, axis, depth):
...
+                self._parameters,
+            )
+            return out
+
+        if isinstance(out, ak._v2.contents.NumpyArray):
+            nextoutindex = ak._v2.index.Index64.empty(len(out) + 1, nplike)
+            # FIXME: move to kernel
+            for i in range(len(out)):
+                nextoutindex[i] = i
+            nextoutindex[-1] = -1
```
jpivarski: Another reminder to move to a kernel.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -570,6 +570,220 @@ def _localindex(self, axis, depth):
...
+            return self
+
+        branch, depth = self.branch_depth
+
+        if (
+            self.parameter("__array__") == "string"
+            or self.parameter("__array__") == "bytestring"
+        ):
+            if branch or (negaxis != depth):
+                raise ValueError("array with strings can only be sorted with axis=-1")
```
jpivarski: I think this also should be a `np.AxisError`.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -570,6 +570,220 @@ def _localindex(self, axis, depth):
...
+                    None,
+                    self._parameters,
+                )
+
+        if not branch and (negaxis == depth):
+            if (
+                self.parameter("__array__") == "string"
+                or self.parameter("__array__") == "bytestring"
+            ):
+                raise ValueError("array with strings can only be sorted with axis=-1")
```
jpivarski: `np.AxisError`

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -570,6 +570,220 @@ def _localindex(self, axis, depth):
...
+            if len(self._offsets) - 1 != len(parents):
+                raise ValueError(
+                    "length of self._offsets_ - 1 is not equal to parents length"
+                )
```
jpivarski: This looks more like it would catch an internal programming error than a user mistake. ValueErrors are for user mistakes (and the error message would have to be more helpful, to tell the user what to do instead) and AssertionErrors are for programming mistakes.

If you're thinking that it really shouldn't be possible to get `len(self._offsets) - 1 != len(parents)`, which would be my expectation, too, you can just do

```python
assert len(self._offsets) - 1 == len(parents)
```

and we'll deal with it if it ever happens in practice. No need to come up with a good error message for an assertion unless if provides some new information—if we ever hit it, we'll be getting into the code, anyway. The same doesn't apply to ValueErrors because user mistakes should not be resolved by delving into the code, but by changing inputs.

src/awkward/_v2/contents/numpyarray.py
```diff
@@ -395,6 +395,300 @@ def iscontiguous(self):
...
+                len(self._data),
+                offsets.to(nplike),
+                length,
+                nextoffsets.to(nplike),
+            )
+        )
+
+        out2 = self[0 : nextoffsets[-1]]
+        # FIXME: trim nextoffsets
+        return out2, nextoffsets
```
jpivarski: Reminder of this FIXME: it looks important and easily overlooked.

tests/v2/test_0404-array-validity-check.py
```diff
@@ -0,0 +1,861 @@
...
+    #     1,
+    #     0,
+    #     1,
+    #     0,
+    #     0,
+    #     0,
+    #     1,
+    #     1,
+    #     1,
+    # ]
```
jpivarski: `BitMaskedArray.bytemask` is now named `mask_as_bool` and takes a `validity_error` argument.

jpivarski (commented):
Great!

I know this is a lot of work, and I see your FIXMEs of the last issues to fix before it's done. I added to that some choices of exception class: C++ without access to pybind forced us to put a lot of different types of exceptions into only one class, ValueError (`std::invalid_argument` in C++), but some of the ones labeled below should instead be `np.AxisError`. We also want to maintain the distinction between user errors (ValueError, TypeError, `np.AxisError`) and internal programming errors, which are AssertionErrors (RuntimeErrors in Awkward 1, as this is what `std::runtime_error` produced, and that was our closest C++ equivalent).

One thing that I didn't mention below is to add type-tracer tests after each value test, which I also mentioned to @stormiestsin in https://github.com/scikit-hep/awkward-1.0/pull/1135#discussion_r744849527. They can usually be added in a semi-automated way, and they ensure that the code is Dask-ready.

Other than that, thanks for all of this; I can see the intricacy of a lot of these cases!

src/awkward/_v2/contents/content.py
ianna: done

src/awkward/_v2/contents/content.py
ianna: done

src/awkward/_v2/contents/listoffsetarray.py
ianna: done

src/awkward/_v2/contents/listoffsetarray.py
ianna: done

tests/v2/test_0404-array-validity-check.py
ianna: done

src/awkward/_v2/contents/indexedoptionarray.py
ianna: done

src/awkward/_v2/contents/indexedoptionarray.py
ianna: done

src/awkward/_v2/contents/numpyarray.py
ianna: the function is not used, I removed it

src/awkward/_v2/contents/listoffsetarray.py
ianna: done

jpivarski (approved):

PR #1116, opened 2021-10-15
C++ refactoring: project - bit/byte/unmaskedarray

ioanaif:
project():

- [x] bitmaskedarray
- [x] bytemaskedarray
- [x] unmaskedarray
- [x] indexedarray
- [x] indexedoptionarray
- [x] testing


jpivarski:
In PR #1117 (which will auto-merge soon), all v2 tests are being moved into a new `tests/v2 directory`, and the "v2" qualifier is being removed from their filenames. Please do that for the tests in this PR, too!

The other directory restructuring that I'll be doing today will be to copy high-level code into `src/awkward/v2`. At first, it will be commented out, and then I'll slowly uncomment things as they're made to work with v2 only. But since that involves new files, it won't interfere with any of your ongoing work.

PR #1120, opened 2021-10-18
C++ refactoring: Enable codecov.

jpivarski:


jpivarski:
@ianna @ioanaif @stormiestsin @agoose77 Coverage for v2 developments are being tracked here: https://app.codecov.io/gh/scikit-hep/awkward-1.0

There might be a way to integrate it into this page, though I don't know it just yet. Anyway, I'm hoping to keep the code coverage monitor at a "minimum noise" level: it's more often something we'll want to dive into as a diagnostic for specific functions, rather than an overall score to get concerned about if it decreases. (The baseline is starting out at 77% without having ever looked at it before. The v1 is higher, but we know we have some "skips" in v2.)

Tomorrow, I'll copy high-level files into the _v2 directory.

PR #1123, opened 2021-10-22
C++ refactoring: Renamed record 'key' -> 'field' to be consistent with high-level.

jpivarski:


jpivarski:
@ioanaif @ianna @stormiestsin @agoose77: the names associated with each of a Record/RecordArray's children are now called "fields" instead of "keys", which were the last two conflicting names for this concept. (Previously, there was also "recordlookup," in addition to "keys".) Some time ago, @nsmith- and I deliberated over the multiplicity of names and narrowed in on "fields," which was applied to the high-level (in PR #464), but only now to the mid-level.

There was also some confusion among classes as to whether this should be a property or a zero-argument method (like a Mapping's `keys()`); it's a property, and all array nodes have this property. (It goes down to the first nested record, returning an empty list if there isn't any. The property is always a list, though a RecordArray/RecordForm/RecordType with `_fields is None` means that that record is a tuple. The non-underscored `fields` in that case is `[str(i) for i in range(len(self._contents))]`.

Form keys (`form_key`) and parameter keys are separate concepts, and they're still named "keys." But all of the method and function argument names like `index_or_key`, referring to record fields, have been named `index_or_field` (etc.).

This PR touches a lot of v2 code, so it could be hard to merge (sorry)! We'll be doing more of these "name alignments," getting mid-level names in sync with the high-level names later, but this one is particularly pervasive and a common word in Python like "key" is hard to fix in only the places where it ought to be fixed, so I think this will be easier to do now than later.

PR #1125, opened 2021-10-25
C++ refactoring: to_arrow and from_arrow in v2

jpivarski:
As part of this, I'm using Arrow's ExtensionType to add metadata so that Awkward → Arrow → Awkward conversions are fully round-trip. That is,

   * never lose option-type, even if an array is not nested within another (corner case of Arrow's nullable-handling)
   * never lose any parameters; they get JSON-encoded

As long as `register_extension_type` is called before opening a file (i.e. the new `ak.from_parquet` uses `ak._v2._connect.pyarrow.import_pyarrow_parquet` as it should), all Awkward metadata will be round-trippable through Parquet, too.


jpivarski:
Oh, and this will require Arrow 6.0.0 as a minimum (so that `ExtensionArray.to_pylist()` works). Hopefully, that will come out in the next few days.

jpivarski:
@ianna, @ioanaif, and @stormiestsin, I think you all have open PRs. Arrow 6.0.0 came out today, which introduced [a bug](https://issues.apache.org/jira/browse/ARROW-14485) that [broke our builds](https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=8055&view=logs&jobId=4f922444-fdfe-5dcf-b824-02f86439ef14&j=4f922444-fdfe-5dcf-b824-02f86439ef14&t=982d46f3-a8a2-5060-e80f-0c9fd0589210).

The single commit 477581d9937776a9fa984a5c278d1eccf14f25d6 is a work-around that avoids it (not a proper fix, but the `from_parquet` code will be rewritten in the Dask context, anyway). You may want to apply this work-around to your branches if you're hitting this test failure.

PR #1132, opened 2021-11-03
C++ refactoring: utility methods for high-level functions in v2.

### LINKED ISSUE #1037: Add `axis=None` functionality to `ak.to/from_regular`

I believe we discussed this on Gitter (or perhaps elsewhere) @jpivarski. I'm opening this issue to track the feature as I don't have time just now to work on it, but want to keep tabs on it.

It would be useful if there were an `axis=None` variant of the to/from regular functions such that the entire layout may be converted to/from regular contents.

jpivarski:


jpivarski:
`recursively_apply` and `broadcast_and_apply` haven't been fully tested, but that will come when we reintroduce all the high-level functions. Whatever bugs are left are corner-cases. (It's translated from old code that worked and the basic cases work in the new code.)

So I'll merge this as soon as the tests pass, and then move on to to/from_buffers.

PR #1134, opened 2021-11-04
C++ refactoring: to_buffers and from_buffers

jpivarski:


jpivarski:
Next after this is `ak._v2.highlevel.__getstate__`/`__setstate__` (pickle). Then uncomment the tests (tests/v2/test_0348-form-keys.py, tests/v2/test_1007-from_buffers-empty-ndarray.py) and that should be it for this PR. It may need a pass-through `ak._v2.operations.structure.packed` (for pickle).

PR #1135, opened 2021-11-08
C++ Refactoring: Implement rpad and rpad_and_clip

sw5sh:
This PR adds the `rpad` and `rpad_and_clip` funtions to `content.py`. Internally, these two have been morphed into a single function `_rpad` with an additional parameter `clip`. Two kernels were replaced with their `nplike` alternatives.


src/awkward/_v2/contents/content.py
```diff
@@ -960,6 +960,38 @@ def fields(self):
...
+        #         "awkward_index_rpad_and_clip_axis0",
+        #         index.dtype.type,
+        #     ](index.to(self.nplike), target, len(self))
+        # )
+        shorter = min(target, len(self))
+        index = ak._v2.index.Index64(
+            self.nplike.append(
+                self.nplike.arange(shorter), self.nplike.full(target - shorter, -1)
+            )
+        )
```
jpivarski: Does `nplike.arange(shorter)` + `nplike.full(target - shorter, -1)` replace `awkward_index_rpad_and_clip_axis0`? If so, great! Let's get rid of the commented-out code and just go forward with the NumPy-like code.

One thing, though: `nplike.append` should probably be banned. Appending arrays always involves allocating a new array and copying into it, so we should approach it like that from the beginning. The following should do the same thing with one fewer copy:

```python
npindex = self.nplike.full(target, -1, dtype=np.int64)
npindex[:shorter] = self.nplike.arange(shorter, dtype=np.int64)
index = ak._v2.index.Index64(npindex)
```

(and also, let's be explicit about integer dtypes to avoid Windows-vs-everything else differences. Windows defaults to `np.int32`, which is a problem because some kernel arguments only accept `np.int64`. In the code above, where you use `Index64` instead of just `Index`, it will be converted into `np.int64`, but on Windows that would mean that the whole array would be constructed in 32-bit and then converted/copied into 64-bit; let's avoid that because we can.)

src/awkward/_v2/contents/content.py
```diff
@@ -960,6 +960,38 @@ def fields(self):
...
+            None,
+            self._parameters,
+        )
+        return next.simplify_optiontype()
+
+    def rpad(self, length, axis):
+        return self._rpad(length, axis, 0, clip=False)
+
+    def rpad_and_clip(self, length, axis):
+        return self._rpad(length, axis, 0, clip=True)
```
jpivarski: Names will get rearranged. The mid-level/internal names will be brought in line with the high-level names, which in this case is `pad_none` with a `clip=True/False` argument. So this layer of making it two separate functions in between the high-level, where it's not, and the low-level, where it's not, will go away.

However, that doesn't have to happen in this PR (unless you want to do that and update the tests now, in anticipation of it eventually happening). For simple renamings, we should plan on doing that in bulk before Awkward 2.0 gets released. We'll wait for a moment when no one is working on any active PRs, gather a list of names to change, and then use some combination of IDE tools and Emacs keyboard macros to do them all efficiently.

So this comment is just for your information (FYI), not an action item.

src/awkward/_v2/contents/emptyarray.py
```diff
@@ -205,6 +205,13 @@ def _reduce_next(
     def _validityerror(self, path):
         return ""
 
+    def _rpad(self, target, axis, depth, clip):
+        posaxis = self.axis_wrap_if_negative(axis)
+        if posaxis != depth:
+            raise ValueError("axis exceeds the depth of this array")
```
jpivarski: Errors in which `axis=XYZ` and `XYZ` is not a valid choice are handled all over the place in the codebase, but they mean the same thing to the user. I've been trying to consolidate their messages, so that they get to be more immediately recognizable, in part by using NumPy's `np.AxisError`, and by using the same message text.

```python
raise np.AxisError("axis={0} exceeds the depth of this array ({1})".format(axis, depth))
```

The `axis` is the `XYZ` that the user specifies (may be negative) and `depth` has a consistent meaning in all of these functions (fortunately!). In principle, one might think this calls for handling all axis errors in a subroutine, but then the last line number of the stack trace would be in that subroutine; to debug it, we'd have to remember to go one level up, and that's a little inconvenient. Also, _calling_ such a subroutine would be something we'd have to do consistently everywhere, and I don't see how it's any harder to just use consistent text. (The only argument would be if we planned on ever changing that text—I think we can safely ignore that possibility.)

So, please do change the exception class and message for this and any other "bad axis" exceptions.

src/awkward/_v2/contents/listarray.py
```diff
@@ -959,6 +959,92 @@ def _validityerror(self, path):
...
+                #         self._starts.dtype.type,
+                #         self._stops.dtype.type,
+                #     ](
+                #         min.to(self.nplike),
+                #         self._starts.to(self.nplike),
+                #         self._stops.to(self.nplike),
+                #         len(self._starts),
+                #     )
+                # )
+                min_ = self.nplike.min(self._stops.data - self._starts.data)
```
jpivarski: Same as in a previous comment. It's great to replace kernels with NumPy functions. Let's also get rid of the commented-out code. If you're thinking about the problem of finding unused kernels (knowing, for instance, that `awkward_ListArray_min_range` is no longer used and can be removed), I plan on making a sweep to search for all of those in bulk, very much like "garbage collection," but for dead code. For the same reason as garbage collection of unreferenced data, it will be easier to do that in bulk.

src/awkward/_v2/contents/numpyarray.py
```diff
@@ -691,6 +691,22 @@ def _validityerror(self, path):
...
         return ""
 
+    def _rpad(self, target, axis, depth, clip):
+        if len(self.shape) == 0:
+            raise ValueError("cannot rpad a scalar")
+        elif len(self.shape) > 1 or not self.iscontiguous():
+            return self.toRegularArray()._rpad(target, axis, depth, clip)
+        posaxis = self.axis_wrap_if_negative(axis)
+        if posaxis != depth:
+            raise ValueError("axis exceeds the depth of this array")
```
jpivarski: Here's another `np.AxisError`. Please search for all of these in your PR; I'm not necessarily going to find them all. Thanks!

tests/v2/test_1135-rpad-operation.py
```diff
@@ -0,0 +1,1297 @@
...
+        [[15, 16, 17, 18, 19], [20, 21, 22, 23, 24], [25, 26, 27, 28, 29]],
+    ]
+
+    assert ak.to_list(array.rpad_and_clip(5, 0)) == [
+        [[0, 1, 2, 3, 4], [5, 6, 7, 8, 9], [10, 11, 12, 13, 14]],
+        [[15, 16, 17, 18, 19], [20, 21, 22, 23, 24], [25, 26, 27, 28, 29]],
+        None,
+        None,
+        None,
+    ]
```
jpivarski: If you're planning on converting all `rpad_and_clip(target, axis)` into `rpad(target, axis, clip=True)`, in preparation for the name change (turning function names into argument values is more awkward than just changing names), know that there's an easier way to do it than doing it all by hand.

In Emacs, this is an ideal candidate for a keyboard macro:

```lisp
M-x start-kbd-macro            ; C-x (
M-x search-forward             ; C-s
.rpad_and_clip(
M-x backward-char              ; C-b or left arrow
M-x delete-backward-char × 9   ; backspace 9 times to delete "_and_clip"
M-x forward-sexp               ; C-M-f to go forward over parenthesized expression
M-x backward-char              ; C-b or left arrow
, clip=True
M-x end-kbd-macro              ; C-x )
```

Then you can run the above keyboard macro by `M-x call-last-kbd-macro` or `C-x e` (and repeatedly by pressing `e` thereafter). It would search forward to all ".rpad_and_clip", erase the "_and_clip", jump to the end of the parenthesized expression (regardless of whatever's in it), and add the ", clip=True". It's the sort of thing that I would never do through `M-x` commands—I'd use the keyboard shortcuts; I'm just being explicit to be understood. Since some of those shortcuts navigate syntactically through the code (jumping over parentheses), it's possible to do sophisticated, syntax-aware edits, not just find-and-replace. Since it plays them back only once per `C-x e` (or `e`), you get to keep an eye on it and make sure that it's not messing things up—semi-automated can be overall faster than manual _and_ faster than full automation, since a fully automated script can fail to handle unforeseen cases and then you have to go back and fix the mess it's made.

I'm telling you all of this so that we globally optimize our use of time. If you would have done this edit manually or by writing a script, note that it can be done in less than a minute with a keyboard macro. Either I'll do it later, as one of the tasks in a future renaming pass, or you might want to try it in Emacs, or—best yet—you can tell me if a feature like this exists in any modern (i.e. non-Emacs) IDE. I'd love to know. I've been searching for it, wondering how users of modern IDEs get along without having this feature, but maybe it exists by another name.

(Actually, I don't remember what your preferred IDE was. If I'm telling you something you already know; sorry!)

tests/v2/test_1135-rpad-operation.py
```diff
@@ -0,0 +1,1297 @@
...
+
+    assert ak.to_list(listoffsetarray.rpad_and_clip(3, 0)) == [
+        [0.0, 1.1, 2.2],
+        [],
+        [3.3, 4.4],
+    ]
+    # assert str("option[") + str(ak.type(listoffsetarray)) + str("]") == str(
+    #     ak.type(listoffsetarray.rpad_and_clip(3, 0))
+    # )
```
jpivarski: Why are all of these type assertions commented out? Do they not work anymore? If so, why?

Another thing that should be done in the tests (new constraint): after each value test, such as

```python
    assert ak.to_list(listoffsetarray.rpad_and_clip(3, 0)) == [
        [0.0, 1.1, 2.2],
        [],
        [3.3, 4.4],
    ]
```

we now need to do a type-tracer test. The equivalent for the above is

```python
    assert (
        listoffsetarray.typetracer.rpad_and_clip(3, 0).form
        == listoffsetarray.rpad_and_clip(3, 0).form
    )
```

The first part, `listoffsetarray.typetracer`, makes an equivalent of `listoffsetarray` with non-concrete arrays in place of the NumPy arrays. We want to pass this type-tracer version of the Content through the same operation that we pass the Content through, then verify that they both have the same Form. That ensures that it will be possible to Daskify this operation, and that users will be able to work with lazy data with full type information. (Note: I've usually added these to tests with a keyboard macro—an imperfect one that I run and then fix the errors, which is faster than typing out the whole thing manually each time.)

The type-tracer test is different from the ones you've commented out: the ones you've commented out seem to be verifying that the type of padded data is equal to the type of the original data + option type (because now `None` is a possible value). That test is specific to what rpad does; the type-tracer tests are general and we have to apply them to all operations.

jpivarski (changes requested):
Thanks for doing this! I just have some inline comments and requests.

tests/v2/test_1135-rpad-operation.py
sw5sh: Thanks for this! This might be cheating but I am currently using `Neovim` which is a `modern` Vim fork. I followed similar steps to register a macro and then execute it for every instance in one go. This is how I did it in Vim, there could be better ways of doing this, but I am not currently very aware of it (I started shifting my workflow to `Neovim`, only like a week back)
```
qa                          "Start recording a macro and store it in the register 'a'
/.rpad_and_clip(            "Find the first instance of "rpad_and_clip(" from the current line
f_                          "jump to the first "_"
de                          "delete till end of the work i.e _and_clip
%                           "go to the matching closing parenthesis i.e )
<InsertMode>                       
, clip=True                 "Insert the kwarg
<EscToNormalMode>
q                           "Stop recording the macro
```
Then we just do a `@a` for the first time and then just spam `@@` for every subsequent macro execution. It's pretty neat. For like proper modern IDEs / Text Editors, it becomes a bit difficult. I am not sure if there's a better way, but in VS Code, you'll have to write some editor specific commands in a JSON object and install an [extension](https://marketplace.visualstudio.com/items?itemName=geddski.macros) to get the macro to work.

src/awkward/_v2/contents/content.py
sw5sh: I have refactored `rpad_and_clip` into `rpad` with an argument `clip` in this PR itself, so this should be resolved.

src/awkward/_v2/contents/content.py
sw5sh: Understood. I have updated the code snippet with this.

src/awkward/_v2/contents/listarray.py
sw5sh: My thinking was that it would be easier to catch commented out kernels later, and remove their implementations permanently. But as you suggested, an automated grep search could also help catch every unused kernel. I have removed the commented out code in the latest commit.

tests/v2/test_1135-rpad-operation.py
sw5sh: I used a macro and semi-automated all the `typetracer` stuff using vim. Now all the tests pass. Regarding the commented out `str` tests, those tests use `ak.type` which hasn't been implemented for `v2` if I am not wrong.

src/awkward/_v2/contents/numpyarray.py
sw5sh: I caught all the bad axis errors and I am raising `np.AxisError` for them.

src/awkward/_v2/contents/emptyarray.py
sw5sh: This is done in the latest commits.

tests/v2/test_1135-rpad-operation.py
jpivarski: All right! Then it can be done easily in Emacs and Vim (I'm guessing the above works in normal Vim, not just Neovim). It seems to be the killer app for programmatic text editing, and it's crucial that there's a low barrier to entry (i.e. not having to install an extension, not having to write it out, but "programming by example"). I'm still surprised that it's not ubiquitous among IDEs.

tests/v2/test_1135-rpad-operation.py
jpivarski: Given a `ak._v2.contents.Content` object named `x`, you can get its type via `x.form.type`.

jpivarski (approved):

jpivarski:
I agree; it's done now, so I'll enable auto-merge. Thanks again!

PR #1140, opened 2021-11-10
C++ refactoring: argsort - remove workaround, use merge

ianna:


ianna:
@jpivarski - this PR is complete. Please, check when you have time. Thanks!

jpivarski (approved):
The first time I wanted to actually use this CodeCov report, and it's not working! https://app.codecov.io/gh/scikit-hep/awkward-1.0/compare/1140/overview

I just wanted to see that the two new lines (`self.mergeable`) are covered, which would indicate that using mergeable/merge is as correct as the work-around. Anyway, if I have your word for it that this part of the code is covered, then you can merge this PR.

jpivarski:
It is tested. (I checked out your branch, ran all the tests—success—then added an exception on the line I wanted to see if it was covered—it raised that exception—so we're good.)

PR #1141, opened 2021-11-10
C++ refactoring: UnionArray argsort and sort and enable tests

ianna:


src/awkward/_v2/contents/unionarray.py
```diff
@@ -701,7 +701,13 @@ def _argsort_next(
         kind,
         order,
     ):
+        if len(self) == 0:
+            return ak._v2.contents.NumpyArray(self.nplike.empty(0, np.int64))
+
         simplified = self.simplify_uniontype(mergebool=True)
+        if len(simplified) == 0:
+            return ak._v2.contents.NumpyArray(self.nplike.empty(0, np.int64))
```
jpivarski, line 709: (This, and a similar statement above.) A value, such as the length being zero, should not determine the type of an array, such as "int". This would rightly fail a type-tracer test, which might not know that the length is zero. Simplify turns `union[X, X]` into `X` and `union[union[X, Y], Z]` into `union[X, Y, Z]`, but only things if those types, determined by type.

jpivarski: Sorry—I didn't realize this was in argaort. The above code is fine if they're returning the expected result of the argsort: an empty int-type array. If that's the case you can resolve this.

src/awkward/_v2/contents/unionarray.py
```diff
@@ -712,7 +718,13 @@ def _argsort_next(
     def _sort_next(
         self, negaxis, starts, parents, outlength, ascending, stable, kind, order
     ):
+        if len(self) == 0:
+            return self
+
         simplified = self.simplify_uniontype(mergebool=True)
+        if len(simplified) == 0:
+            return simplified
```
jpivarski, line 726: Same here.

jpivarski (commented):
This is probably right: I wrote the first comment thinking it was for a different function—if I was wrong about that and there's nothing wrong with the code I pointed out, then this PR is 100% ok and if ready to be merged.

src/awkward/_v2/contents/unionarray.py
ianna: Thanks! Yes I enabled the union array tests and saw that it needs fixing.

src/awkward/_v2/contents/unionarray.py
ianna: this enable sorting of zero length arrays

https://github.com/scikit-hep/awkward-1.0/pull/1141/files/b1bb82db80d16875bafead8467b6d01d2ae3d3d8#diff-3943a97658fec00f9029d697ac275074416842b955db8f0d66c200ae2f20bc5cL820-L822

ianna (commented):
@jpivarski - thanks for checking. This PR is a short one, I'm done with it.

PR #1142, opened 2021-11-10
C++ refactoring: prepared high-level ArrayBuilder and reducer functions, though still untested.

ianna:


tests/v2/test_0835-datetime-type.py
```diff
@@ -47,454 +53,692 @@ def test_date_time_minmax():
...
+    assert ak.to_list(ak._v2.operations.reducers.min(array, axis=0)) == [
+        np.datetime64("2020-02-27T10:41:11"),
+        np.datetime64("2020-01-27T10:41:11"),
+        np.datetime64("2020-01-27T10:41:11"),
+        np.datetime64("2020-01-27T10:41:11"),
+        np.datetime64("2020-04-27T10:41:11"),
+    ]
+    assert ak.to_list(ak._v2.operations.reducers.max(array, axis=0)) == [
+        datetime.datetime(2020, 4, 27, 0, 0),
+        datetime.datetime(2020, 3, 27, 10, 41, 11),
```
ianna, line 599: something to look into: `max` returns `datetime.datetime` after update to master while `min` returns `numpy.datetime64`

tests/v2/test_0835-datetime-type.py
```diff
@@ -47,454 +53,692 @@ def test_date_time_minmax():
...
+        akarray[1:] + akarray[:-1]
+
+
+def test_ufunc_mul():
+    nparray = np.array(
+        [np.datetime64("2021-06-03T10:00"), np.datetime64("2021-06-03T11:00")]
+    )
+    akarray = ak._v2.highlevel.Array(nparray)
+
+    with pytest.raises(np_exception._UFuncBinaryResolutionError):
```
ianna: this does not work with earlier versions of Numpy. In any case, it should be a `ValueError` exception.

ianna (commented):
@jpivarski - I'm not sure what to do with this PR: put it hold or merge?  It implements support for `datetime` and `timedelta` in all reducers. It adds 26 tests that pass plus 6 that currently skipped. However, the reducers can not handle axis `None` due to an `AttributeError: ``module 'awkward._v2._util' has no attribute 'completely_flatten'`

There is an odd behaviour of `max` vs `min` after a recent update. Also, I think, a `Numpy` exception mentioned in the comments will be handled elsewhere.

ianna:
The following test fails with Numpy 1.3:
```python
def test_ufunc_sum():
    nparray = np.array(
        [np.datetime64("2021-06-03T10:00"), np.datetime64("2021-06-03T11:00")]
    )
    akarray = ak._v2.highlevel.Array(nparray)

    with pytest.raises(np.core._exceptions.UFuncTypeError):
        akarray[1:] + akarray[:-1]

```
due to an import error:
```
tests/v2/test_0835-datetime-type.py:12: in <module>
    import numpy.core._exceptions as np_exception
E   ImportError: No module named 'numpy.core._exceptions'
```

jpivarski:
`completely_flatten` is now a `Content` method. (Since it only involves one array, turning it into a method lets us do OOP-style small functions instead of `if`-`elif` chains, which were only there because I was in a hurry and wanted to avoid writing C++ code.)

jpivarski:
> ```python
>     with pytest.raises(np.core._exceptions.UFuncTypeError):
>         akarray[1:] + akarray[:-1]
> ```

Relying on a private module of another library is not a good idea, for exactly this reason. (It could disappear again in a future version, without warning.) Checking for TypeError (UFuncTypeError's immediately superclass) would work just as well. That's a public API.

tests/v2/test_0835-datetime-type.py
jpivarski: Same comment as below. It's because `np.core._exceptions` is private. This one descends from TypeError, not ValueError, though:

```python
>>> np.core._exceptions._UFuncBinaryResolutionError.__mro__
(<class 'numpy.core._exceptions._UFuncBinaryResolutionError'>,
 <class 'numpy.core._exceptions.UFuncTypeError'>,
 <class 'TypeError'>,
 <class 'Exception'>,
 <class 'BaseException'>,
 <class 'object'>)
```

src/awkward/_v2/_reducers.py
```diff
@@ -44,14 +44,22 @@ class ArgMin(Reducer):
 
     @classmethod
     def apply(cls, array, parents, outlength):
+        dtype = (
+            np.dtype(np.int64)
+            if (
+                str(array.form.type).startswith("datetime64")
+                or str(array.form.type).startswith("timedelta64")
+            )
```
jpivarski: We should not be using the string representations of objects to determine what type they are. Currently, it would be

```python
isinstance(t, ak._v2.types.NumpyType) and t.primitive.startswith("datetime64")
```

but this should be in a helper function. Hold on a moment, I'll fix this FIXME:

https://github.com/scikit-hep/awkward-1.0/blob/f795658559d61f1fd394c828e95608c9de8ee027/src/awkward/_v2/types/numpytype.py#L14-L65

(In the original design, I hadn't counted on datetimes and timedeltas having units, much less _arbitrary_ units, so I've got to turn that dict into a function. I'll do it now as a tiny PR.)

tests/v2/test_0835-datetime-type.py
ianna: done

src/awkward/_v2/operations/reducers/ak_all.py
```diff
@@ -9,53 +9,50 @@
...
-#             [ak.nplike.of(x).all(x) for x in ak._v2._util.completely_flatten(layout)]
-#         )
-#     else:
-#         behavior = ak._v2._util.behaviorof(array)
-#         return ak._v2._util.wrap(
-#             layout.all(axis=axis, mask=mask_identity, keepdims=keepdims), behavior
-#         )
+    """
+    Args:
+        array: Data to combine with "logical all."
```
jpivarski: The text to use now for an "array" argument is

https://github.com/scikit-hep/awkward-1.0/blob/f795658559d61f1fd394c828e95608c9de8ee027/src/awkward/_v2/operations/convert/ak_to_buffers.py#L21

Somebody once complained that this argument is described in a variety of ways—it wasn't standardized the way some of the other arguments are. Having standard text tells a user that a given argument is interpreted the same way in all functions with that standard text.

src/awkward/_v2/highlevel.py
```diff
@@ -2085,726 +2085,721 @@ def __bool__(self):
...
+        can be a more time and memory-efficient way to put old data in a
+        new structure, since it avoids copying and even walking over the
+        old data structure (matters more when the structures are large).
+
+        If `obj` is an arbitrary Python object, this is equivalent to
+        #ak.from_iter except that it fills an existing #ak.ArrayBuilder,
+        rather than creating a new one.
+
+        If `obj` is an #ak.Array and `at` is an int, this method fills the
+        ArrayBuilder with a reference to `obj[at]` instead of `obj`.
```
jpivarski: This part of the documentation is no longer true; these paragraphs can just be dropped. If you want to mention that Python iterables (including dicts) are interpreted with #ak.from_iter, then that would be good.

For reference, here is the old, removed behavior:

```python
>>> import awkward as ak
>>> array = ak.Array([[1, 2, 3], [], [4, 5]])
>>> builder = ak.ArrayBuilder()
>>> builder.append(array)

>>> builder.snapshot()
<Array [[[1, 2, 3], [], [4, 5]]] type='1 * var * var * int64'>

>>> builder.snapshot().layout
<ListOffsetArray64>
    <offsets><Index64 i="[0 3]" offset="0" length="2" at="0x55ceeb42f9f0"/></offsets>
    <content><IndexedArray64>
        <index><Index64 i="[0 1 2]" offset="0" length="3" at="0x55ceeb847d60"/></index>
        <content><ListOffsetArray64>
            <offsets><Index64 i="[0 3 3 5]" offset="0" length="4" at="0x55ceeb117a30"/></offsets>
            <content><NumpyArray format="l" shape="5" data="1 2 3 4 5" at="0x55ceeb1143d0"/></content>
    </IndexedArray64></content>
</ListOffsetArray64>
```

and here is the new behavior:

```python
>>> import awkward as ak
>>> array = ak.Array([[1, 2, 3], [], [4, 5]])
>>> builder = ak.ArrayBuilder()
>>> builder.append(array)

>>> builder.snapshot()
<Array [[[1, 2, 3], [], [4, 5]]] type='1 * var * var * int64'>

>>> builder.snapshot().layout
<ListOffsetArray64>
    <offsets><Index64 i="[0 3]" offset="0" length="2" at="0x56492e998cb0"/></offsets>
    <content><ListOffsetArray64>
        <offsets><Index64 i="[0 3 3 5]" offset="0" length="4" at="0x56492efd8e20"/></offsets>
        <content><NumpyArray format="l" shape="5" data="1 2 3 4 5" at="0x56492efd8f20"/></content>
</ListOffsetArray64>
```

The difference is that there's no longer any link to the original array (through an IndexedArray); the `builder.append(array)` treats the `array` like any Python iterable and copies its data.

src/awkward/_v2/operations/reducers/ak_all.py
```diff
@@ -9,53 +9,50 @@
...
+
+        def reduce(xs):
+            if len(xs) == 1:
+                return xs[0]
+            else:
+                return xs[0] and reduce(xs[1:])
+
+        return reduce(
+            [ak.nplike.of(x).all(x) for x in ak._v2._util.completely_flatten(layout)]
+        )
```
jpivarski: Actually, I had been planning to work on these reducers, though you're more than welcome to do it. I'll find something else that needs doing. But, something I had in mind about all of these reducers: you might want to create a helper function to handle the `axis=None` case because it's very repetitive; there's a lot of duplicated code (which is now less evident because each of these functions is in a different file). The helper function can live in `ak._v2._util`, where all of these functions can see it.

ianna: Thanks! I don't mind if you take over the branch - I will not be working on it during the weekend. I'll see were we stand on Monday. :-)

src/awkward/_v2/_reducers.py
jpivarski: PR #1146 will merge as soon as its tests pass, but it provides functional access to this information. The check (which you can put into a helper function) would then be

```python
(
    isinstance(t, ak._v2.types.NumpyType)
    and ak._v2.types.numpytype.primitive_to_dtype(t.primitive).kind.upper() == "M"
)
```

src/awkward/_v2/operations/reducers/ak_all.py
jpivarski: I don't entirely know what you were going to do with the datetimes/timedeltas—is there kernel-level work needed? If so, I'll just do the part I was intending, which was the high-level interface. I can get the high-level ArrayBuilder updated, too. By Monday, I'll leave this in a state where it's only needing the datetimes/timedeltas and tests.

ianna: Thanks! I don't think any kernel-level work is needed.

src/awkward/_v2/_reducers.py
jpivarski: Done in 883f0bb3e9ef277e0d6513521b700cac0ae963eb (not yet checked in).

tests/v2/test_0835-datetime-type.py
jpivarski: Note: I didn't look into this. It's still an issue.

jpivarski:
I've changed the title of this PR to better reflect what has been updated. I'd also like to merge this as soon as it passes tests to keep PRs granular. (I've made some code cleanliness fixes, and if I'm working on something else in another PR, I might get confused that they're not already done. Thus, I want to start any PRs I do later today on this one.)

Note that I haven't moved any new tests into tests/v2 and I haven't changed the "skip" status of any that are there. The things you were planning to do with this PR aren't done yet. You'll want to open a new PR, probably with this one's old name, to address those things.

ianna:
Awesome! Thanks, @jpivarski !

PR #1150, opened 2021-11-16
C++ refactoring: to_numpy()

ianna:
- [x] reducers support complex numbers and tests
- [x] to_numpy and tests
- [x] validity error enabled and tests
- [x] tests for highlevel reducers enabled


ianna:
@jpivarski - nearly done. There are two issues to fix and I'll work on them tomorrow.

The first is that validity check that expects `dtype == uint8`. This is in `tests/v2/test_0089-numpy-functions.py:345: ValueError: at layout ("<class 'awkward._v2.contents.listoffsetarray.ListOffsetArray'>"): __array__ = "char" requires dtype == uint8`

The second is a `ValueError: ak.size is ambiguous due to variable-length arrays at axis 2 (try ak.flatten to remove structure or ak.to_numpy to force regularity, if possible)` in `tests/v2/test_0089-numpy-functions.py:345`

jpivarski:
We should drop `ak.size`. It was only added to support Pandas, and that's no longer a goal (#350). It makes no sense when arrays can be non-rectilinear; for that, you need all the information in `ak.num`. Such a test can probably be dropped as well.

ianna:
@jpivarski - please, review. I've enabled most of the tests. I've partially enabled the code for `Char` and `Byte` behaviour, however, these functions can be copied to `to_numpy` directly if nothing else is using it.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -1788,6 +1788,25 @@ def _to_arrow(self, pyarrow, mask_node, validbytes, length, options):
...
+                ]
+            )
+
+        if self.parameter("__array__") == "string":
+            return self.nplike.array(
+                [
+                    ak._v2.behaviors.string.CharBehavior(self[i]).__str__()
+                    for i in range(len(self))
+                ]
+            )
```
jpivarski: These are loops over all the elements in the array. Ideally, we'd pad them all to the maximum size (`pad_none` followed by `fill_none`) and take the padded buffer directly into NumPy in a vectorized operation, but I think `fill_none` isn't ready yet.

The next-best thing is this 40× faster `to_list` implementation; it iterates over all the elements in Python, making an intermediate Python list (as the original code did), but it does so only with built-in Python types, and that speeds it up a lot.

```suggestion
        array_param = self.parameter("__array__")
        if array_param == "bytestring" or array_param == "string":
            return self.nplike.array(self.to_list())
```

jpivarski (approved):
Great! I had only one suggested change, which I'll put in, let the tests run, and squash-and-merge if it passes.

Thanks!

jpivarski:
`to_list` might also fix the failing tests:

```
E       AssertionError: assert False
E        +  where False = <function array_equal at 0x7fcf985b1160>(array(['[111, 110, 101]', '[116, 119, 111]', '[116, 104, 114, 101, 101]'],\n      dtype='<U25'), array(['one', 'two', 'three'], dtype='<U5'))
```

Somewhere, it's converting "[111, 110, 101]" (which is `[ord(x) for x in "one"]`) into a string, `'[111, 110, 101]'`, instead of just `"one"`. I don't know why that is because I don't see that error when I run the tests locally, but switching to `to_list` would change the code path, at least.

PR #1161, opened 2021-11-22
C++ refactoring: ak.concatenate

ioanaif:


src/awkward/_v2/_util.py
```diff
@@ -23,6 +24,14 @@
...
 np = ak.nplike.NumpyMetadata.instance()
 
+py27 = sys.version_info[0] < 3
+py35 = sys.version_info[0] == 3 and sys.version_info[1] <= 5
+
+# to silence flake8 F821 errors
+if py27:
+    unicode = eval("unicode")
+else:
+    unicode = None
```
jpivarski: Awkward v2 isn't supporting Python 2.7 or 3.5, so these won't be needed for anything.

(All of the Python 2.7 checks at the beginning of each v2 test will be removed when Python 2.7 is removed from CI.)

src/awkward/_v2/_util.py
```diff
@@ -627,594 +636,536 @@ def wrap(content, behavior=None, highlevel=True, like=None):
...
-
-#                 if first is None:
-#                     secondround = True
-#                     for x in inputs:
-#                         if isinstance(x, listtypes) and not isinstance(
-#                             x, ak._v2.contents.RegularArray
-#                         ):
-#                             first = x
-#                             break
+def broadcast_and_apply(  # noqa: C901
```
jpivarski: `broadcast_and_apply` is in `_broadcasting.py` now. That's the one that has a "depth_context" and "nested_context" now. I'm sorry—I should have removed the commented out version from this file.

src/awkward/_v2/_util.py
```diff
@@ -1316,204 +1267,118 @@ def wrap(content, behavior=None, highlevel=True, like=None):
...
-#         return ak._v2.contents.BitMaskedArray(
-#             layout.mask,
-#             transform(layout.content, depth, user),
-#             layout.valid_when,
-#             len(layout),
-#             layout.lsb_order,
-#             layout.identities,
-#             layout.parameters if keep_parameters else None,
-#         )
+def transform_child_layouts(transform, layout, depth, user=None, keep_parameters=True):
```
jpivarski: The `transform_child_layouts` was once part of `recursively_apply`, which is now a method on Content. It should be possible to express what you're using `transform_child_layouts` for with `recursively_apply` (and thus reduce the number of code paths to support—that's why they were consolidated).

src/awkward/_v2/contents/content.py
```diff
@@ -607,7 +607,7 @@ def _localindex_axis0(self):
     # FIXME: Temporary fix for concatenate (only working for axis = 0)
     def concatenate(self, others):
         if not isinstance(others, list):
-            others = [others]
+            others = [ak._v2.contents.NumpyArray([others])]
```
jpivarski: With `concatenate` having been implemented, this method can be removed now, right? If it still has use as a helper function, then the "FIXME" can be removed, but also the arguments should be normalized to Content subclasses in the high-level function (which would ensure that any raw NumPy arrays are already converted to NumpyArray via a single call to `to_layouts`—by the time we get here, it should already be a Content subclass).

src/awkward/_v2/operations/structure/ak_concatenate.py
```diff
@@ -4,524 +4,201 @@
...
+    #             a low-level #ak.layout.Content subclass.
+    #         behavior (None or dict): Custom #ak.behavior for the output array, if
+    #             high-level.
+
+    #     Returns an array with `arrays` concatenated. For `axis=0`, this means that
+    #     one whole array follows another. For `axis=1`, it means that the `arrays`
+    #     must have the same lengths and nested lists are each concatenated,
+    #     element for element, and similarly for deeper levels.
+    #     """
+
```
jpivarski, line 40: If the docstring is still correct, it can be uncommented.

src/awkward/_v2/operations/structure/ak_concatenate.py
```diff
@@ -4,524 +4,201 @@
...
+    contents = [
+        ak._v2.operations.convert.to_layout(
+            x, allow_record=False if axis == 0 else True, allow_other=True
+        )
+        for x in arrays
+    ]
```
jpivarski, line 46: This is where all inputs have been normalized. From this point on, we can assume that they're not raw NumPy arrays anymore (oy Python lists, etc.).

src/awkward/_v2/operations/structure/ak_concatenate.py
```diff
@@ -4,524 +4,201 @@
...
+                batch = [collapsed.merge_as_union(x)]
+
+        out = batch[0].mergemany(batch[1:])
+        if isinstance(out, ak._v2.contents.unionarray.UnionArray):
+            # if isinstance(out, ak._v2._util.uniontypes):
+            out = out.simplify_uniontype(merge=merge, mergebool=mergebool)
+
+    else:
+
+        def getfunction(inputs, depth):
```
jpivarski: This can be replaced with an `action` that passes the depth down through a `depth_context`, following the example in the v2 `from_regular`. That one uses `recursively_apply`, rather than `broadcast_and_apply`, but the two interfaces have been made the same now. There are also examples of v2 `broadcast_and_apply` in `_connect/numpy.py`.

src/awkward/_v2/operations/structure/ak_concatenate.py
```diff
@@ -4,524 +4,201 @@
...
+                return None
+
+        out = ak._v2._util.broadcast_and_apply(
+            contents,
+            getfunction,
+            behavior=ak._v2._util.behavior_of(*arrays, behavior=behavior),
+            allow_records=True,
+            right_broadcast=False,
+            pass_depth=True,
+        )[0]
```
jpivarski, line 198: This is where the `broadcast_and_apply` from `_broadcasting.py` would go. It didn't have a custom switch for `pass_depth`; that information is one of the many keyword arguments to your `action` function. Use `**kwargs` to absorb all the keyword arguments you don't use (as in `from_regular`).

src/awkward/_v2/operations/structure/ak_fill_none.py
```diff
@@ -2,125 +2,133 @@
...
+                return layout
+
+            if posaxis + 1 == depth:
+                layout = maybe_fillna(layout)
+
+            return ak._v2._util.transform_child_layouts(
+                transform, layout, depth, posaxis
+            )
+
+    out = transform(arraylayout, 1, axis)
```
jpivarski: Instead of these three calls to `transform`, use `recursively_apply`. (The intention was to reduce all these different ways of doing essentially the same thing with a single interface—that's why I don't want all the old interfaces brought back.)

jpivarski (changes requested):
I can see how much went into this—thank you! However, I wasn't clear enough about the new `broadcast_and_apply` and `recursively_apply` interfaces, which are in `_broadcasting.py` and the Content class, respectively. What you've done below takes us to 90% of the implementation, but some arguments need to change to use the new interfaces. After that, the `broadcast_and_apply` and `transform`/`transform_layouts` functions you ported in `_v2/_util.py` won't be needed.

jpivarski (approved):

jpivarski (approved):
Thank you very much for fixing that (and finding a few untested parts of _broadcasting.py). The commit I just made was to remove some commented-out code, that's all. It's set to auto-merge when the tests pass again.

ioanaif:
> Thank you very much for fixing that (and finding a few untested parts of _broadcasting.py). The commit I just made was to remove some commented-out code, that's all. It's set to auto-merge when the tests pass again.

Thank you! Great! Happy to have this merged!

douglasdavis:
Woohoo! Thanks for this 😄

PR #1162, opened 2021-11-22
ak.from_json_schema as a demonstration of generating AwkwardForth from a type-schema

jpivarski:


jpivarski:
Code generated for `test_number`:

```forth
input source
output node0-data float64

source skipws
source enumonly s" [" drop
source skipws
0
source enum s" ]"
begin
while
  source textfloat-> node0-data
  1+
  source skipws
  source enumonly s" ]" s" ,"
  source skipws
repeat
source skipws
```

Code generator for `test_option_integer`:

```forth
input source
output node0-mask int8
output node1-data int64

source skipws
source enumonly s" [" drop
source skipws
0
source enum s" ]"
begin
while
  source enum s" null" dup if
    source textfloat-> node1-data
  else
    0 node1-data <- stack
  then
  node0-mask <- stack
  1+
  source skipws
  source enumonly s" ]" s" ,"
  source skipws
repeat
source skipws
```

Code generated for `test_string`:

```forth
input source
output node0-offsets int64
output node1-data uint8
0 node0-offsets <- stack

source skipws
source enumonly s" [" drop
source skipws
0
source enum s" ]"
begin
while
  source quotedstr-> node1-data
  node0-offsets +<- stack
  1+
  source skipws
  source enumonly s" ]" s" ,"
  source skipws
repeat
source skipws
```

Code generated for `test_array_integer`:

```forth
input source
output node0-offsets int64
output node1-data int64
0 node0-offsets <- stack

source skipws
source enumonly s" [" drop
source skipws
0
source enum s" ]"
begin
while
  source enumonly s" [" drop
  source skipws
  0
  source enum s" ]" 
  begin
  while
    source textfloat-> node1-data
    1+
    source skipws
    source enumonly s" ]" s" ," 
    source skipws
  repeat
  node0-offsets +<- stack
  1+
  source skipws
  source enumonly s" ]" s" ,"
  source skipws
repeat
source skipws
```

Code generated for `test_array_array_integer`:

```forth
input source
output node0-offsets int64
output node1-offsets int64
output node2-data int64
0 node0-offsets <- stack
0 node1-offsets <- stack

source skipws
source enumonly s" [" drop
source skipws
0
source enum s" ]"
begin
while
  source enumonly s" [" drop
  source skipws
  0
  source enum s" ]" 
  begin
  while
    source enumonly s" [" drop
    source skipws
    0
    source enum s" ]" 
    begin
    while
      source textfloat-> node2-data
      1+
      source skipws
      source enumonly s" ]" s" ," 
      source skipws
    repeat
    node1-offsets +<- stack
    1+
    source skipws
    source enumonly s" ]" s" ," 
    source skipws
  repeat
  node0-offsets +<- stack
  1+
  source skipws
  source enumonly s" ]" s" ,"
  source skipws
repeat
source skipws
```

Code generated for `test_option_record`:

```forth
input source
output node0-index int64
output node1-data int64
output node2-data float64
variable node0-count

source skipws
source enumonly s" [" drop
source skipws
0
source enum s" ]"
begin
while
  source enumonly s" null" s" {" if
    node0-count @ node0-index <- stack
    1 node0-count +!
    source skipws
    source enum s" }" 
    begin
    while
      source enumonly s" \"x\"" s" \"y\""
      source skipws
      source enumonly s" :" drop 
      source skipws
      case
        0 of
          source textfloat-> node1-data
        endof
        1 of
          source textfloat-> node2-data
        endof
      endcase
      source skipws
      source enumonly s" }" s" ," 
      source skipws
    repeat
  else
    -1 node0-index <- stack
  then
  1+
  source skipws
  source enumonly s" ]" s" ,"
  source skipws
repeat
source skipws
```

Unlike ArrayBuilder, which has to anticipate that any level of nesting might become a record, all of these option-types are implemented with ByteMaskedArray except for optional records, which are IndexedOptionArray. That's because making a dummy-value record to hide behind a ByteMaskedArray mask would be complicated to do and could be arbitrarily large. Only the IndexedOptionArray needs a variable to keep a global count: everything else is stacky.

Note that this does not _validate_ the JSON against its schema; it _assumes_ the schema to parse the data more quickly.

jpivarski:
Tomorrow I should write a docstring and do some performance tests.

After that, plain old JSON (no schema).

jpivarski:
Well, using a JSONSchema to build a Forth machine often helps a little with performance, maybe 20%, over the untyped case, but not a lot. It depends on the data structure, of course, but there isn't a clear pattern where the specialized code approach shines.

Taking this to be a standard example:

```python
MULTIPLIER = int(10e6)
json_string = b"[" + b", ".join([
    b'[{"x": 1.1, "y": [1]}, {"x": 2.2, "y": [1, 2]}, {"x": 3.3, "y": [1, 2, 3]}],' +
    b'[],' +
    b'[{"x": 4.4, "y": [1, 2, 3, 4]}, {"x": 5.5, "y": [1, 2, 3, 4, 5]}]'
] * MULTIPLIER) + b"]"
```

(from the one-slide-summary), the schema-specialized Forth code takes 31.5 seconds and the untyped ArrayBuilder takes 38.6 seconds (average of 3 runs; standard deviations of 0.5 seconds). For very simple types, the ArrayBuilder wins, and for complex types, Forth's gains are only at this 20% level. It might not justify maintaining it as a separate code path.

Pulling it apart is revealing: with all ArrayBuilder code removed, just the RapidJSON parsing is 6.0 seconds while the Forth is still 30.8 seconds. RapidJSON is 5× faster at parsing JSON than the schema-specialized Forth, and I should have checked the RapidJSON-only figure first because this is consistent with Forth's ~5 ns/instruction rule: in these runs, it was 7.69 ns/instruction.

Absorbing the `skipws` (skip whitespace) commands into the subsequent command could shave off 20%, but it's not clear that that's worthwhile.

While AwkwardForth may be a good idea for general binary formats, especially ROOT, it's not great for JSON. Ideally, we'd want to attach the dynamic execution to RapidJSON's own parsing (and therefore not have to compete in that department). Just something to think about.

PR #1164, opened 2021-11-23
C++ refactoring: to and from json

ianna:


ianna:
@jpivarski - I've moved the snapshot() generation out of the json functions. The changes are minimal, however this is causing a problem. Please, have a look at a failing `tests/test_0384-lazy-arrayset.py`.

It looks like that the contents order is not preserved thus a generated `form_key` is different.
Screenshot 2021-11-23 at 18 11 36

new vs old:
```diff
< [('get', 'kitty-node13-tags'), ('get', 'kitty-node13-index'), ('get', 'kitty-node16-offsets'), ('get', 'kitty-node15-data'), ('get', 'kitty-node2-data'), ('get', 'kitty-node3-data')]
---
> [('get', 'kitty-node11-tags'), ('get', 'kitty-node11-index'), ('get', 'kitty-node14-offsets'), ('get', 'kitty-node13-data'), ('get', 'kitty-node6-data'), ('get', 'kitty-node7-data')]
```

src/libawkward/io/json.cpp
```diff
@@ -772,7 +772,7 @@ namespace awkward {
   };
 
   template<typename HANDLER, typename STREAM>
-  const ContentPtr
+  const std::pair<int, const BuilderPtr>
   do_parse(HANDLER& handler, rj::Reader& reader, STREAM& stream) {
```
jpivarski, line 776: A Builder knows its own `length()`, so I think this extra integer is not necessary, which would simplify these signatures quite a lot. Also, we don't use bare `int` in Awkward C++; we always want to know the integer size (and integers outside of arrays are always `int64_t`).

jpivarski:
The JSON representation of RecordForms is an object (curly brackets); I wish I had chosen two JSON arrays, but it has to remain backward-compatible. When that passes through Python dicts in Python < 3.6, the order may be lost. However, you're seeing errors in all the Python versions and the stage at which `form_keys` are associated with these Forms is much earlier. This one way in which the order might not be preserved is not what's going on in the latest round of failing tests.

In `test_0384-lazy-arrayset.py:test_lazy_buffers`, the JSON has `"listcollection"` before `"collection"`, so the ArrayBuilder must parse it first and it must be the first key in the RecordBuilder. RecordBuilder, RecordForm (when not serialized), and RecordArray all have a well-defined order for their keys and contents. Of your two screenshots, the one on the left is wrong.

I can't see what's going wrong here, but there must be an exact point in the process when it goes wrong. Are the records already out of order in the `array` defined on lines 25-65 (e.g. `"collection"` is before `"listcollection"` in the `array.layout.recordlookup`)? If so, at what point do they get out of order? They ought to all be in order in the RecordBuilder; the only code that has changed is code between the RecordBuilder and the RecordArray. What about the RecordForm that comes between those two? (As opposed to the various Forms that the later `ak.from_buffers` call makes on line 76.)

ianna:
Yes, mode debugging is needed... I'm on it.

src/libawkward/io/json.cpp
ianna: Currently the integer is needed because it is an indicator of multiple JSON streams for the following cases: `tests/test_0437-stream-of-many-json-files.py`

jpivarski: From what I saw of the code, I thought the integer was the length of the Builder. If so, `Builder::length` can produce that on demand.

If it's not the length, what is it? The JSON streams, or "JSONS" or "[ndjson](http://ndjson.org/)" would produce a Builder with a `length` just as the "one big JSON array" case would. Is the integer 0 for non-stream, 1 for stream, or something like that?

src/python/io.cpp
```diff
@@ -15,6 +13,55 @@
...
+  /// Content. Hence, taking a snapshot is a constant-time operation.
+  ///
+  /// It is safe to take multiple snapshots while accumulating data. The
+  /// shared buffers are only appended to, which affects elements beyond
+  /// the limited view of old snapshots.
+  py::object
+  builder_snapshot(const ak::BuilderPtr builder) {
+    ::NumpyBuffersContainer container;
+    int64_t form_key_id = 0;
+    std::string form = builder.get()->to_buffers(container, form_key_id);
```
ianna: @jpivarski - here: it looks like the keys get sorted. Enforcing an alphabetical order (as a test, not a solution) seems to fix the problem:
```diff
def test_lazy_buffers():
+
     array = ak.from_json(
         """
     [
         {
-            "listcollection": [
+            "a_listcollection": [
                 {"item1": 1, "item2": 2},
                 {"item1": 2, "item2": 4},
                 {"item1": 3, "item2": 6}
             ],
-            "collection": {"item1": 3, "item2": 4},
-            "singleton": 5,
-            "listsingleton": [1, 2, 3],
-            "unioncollection": {"item1": 3},
-            "masked": null
+            "b_collection": {"item1": 3, "item2": 4},
+            "c_singleton": 5,
+            "d_listsingleton": [1, 2, 3],
+            "e_unioncollection": {"item1": 3},
+            "f_masked": null
         },
         {
-            "listcollection": [
+            "a_listcollection": [
                 {"item1": 1, "item2": 2},
                 {"item1": 2, "item2": 4},
                 {"item1": 3, "item2": 6}
             ],
-            "collection": {"item1": 3, "item2": 4},
-            "singleton": 5,
-            "listsingleton": [1, 2, 3],
-            "unioncollection": [{"item1": 2}],
-            "masked": 4
+            "b_collection": {"item1": 3, "item2": 4},
+            "c_singleton": 5,
+            "d_listsingleton": [1, 2, 3],
+            "e_unioncollection": [{"item1": 2}],
+            "f_masked": 4
         },
         {
-            "listcollection": [
+            "a_listcollection": [
                 {"item1": 1, "item2": 2},
                 {"item1": 2, "item2": 4},
                 {"item1": 3, "item2": 6}
             ],
-            "collection": {"item1": 3, "item2": 4},
-            "singleton": 5,
-            "listsingleton": [1, 2, 3],
-            "unioncollection": {"item1": 4},
-            "masked": 4
+            "b_collection": {"item1": 3, "item2": 4},
+            "c_singleton": 5,
+            "d_listsingleton": [1, 2, 3],
+            "e_unioncollection": {"item1": 4},
+            "f_masked": 4
         }
     ]"""
     )

```

ianna: ok, getting warmer:
```c++
      .def(py::init([](const std::map<std::string,
                                      std::shared_ptr<ak::Form>>& contents,

```

jpivarski: Sorting might have been a way to get some Python 2.7/3.5 tests to be stable, but `from_json` should not be sorting keys. The JSON text has a particular order, and all steps through to the creation of the array should be maintaining that order.

Feel free to impose a Python version >= 3.6 constraint on any tests that fail. We won't be keeping these Python versions in CI for very long.

Do you think it's the C++ `std::map` that's sorting these keys? I think the v1 Forms have a constructor that takes two `std::vectors` (one for key names, the other for contents) instead of the `std::map`.

ianna: yes, `std::map` is a sorted associative container. I'll try to switch to `std::unordered_map`.

ianna: `ArrayBuilder` test shows that the `Record`s get sorted:

```Python
def test_array_builder():
    b = ak.ArrayBuilder()
    b.begin_list()

    b.begin_record()
    b.field("listcollection")

    b.begin_list()

    b.begin_record()
    b.field("item1")

    b.begin_list()
    b.integer(1)
    b.integer(3)
    b.integer(2)
    b.end_list()

    b.end_record()

    b.begin_record()
    b.field("item2")

    b.begin_list()
    b.integer(2)
    b.integer(6)
    b.integer(4)
    b.end_list()

    b.end_record()

    b.end_list()
    b.end_record()


    b.begin_record()
    b.field("collection")
    b.begin_list()
    b.begin_record()
    b.field("item1")
    b.integer(3)
    b.end_record()
    b.begin_record()
    b.field("item2")
    b.integer(4)
    b.end_record()

    b.end_list()
    b.end_record()

    b.begin_record()
    b.field("singleton")
    b.integer(5)
    b.end_record()

    b.begin_record()
    b.field("listsingleton")
    b.begin_list()
    b.integer(1)
    b.integer(3)
    b.integer(2)
    b.end_list()
    b.end_record()

    b.begin_record()
    b.field("unioncollection")
    b.begin_record()
    b.field("item1")
    b.integer(3)
    b.end_record()
    b.end_record()

    b.begin_record()
    b.field("masked")
    b.null()
    b.end_record()
    b.end_list()

    print(b.snapshot().layout)
```

```
<ListOffsetArray64>
    <offsets><Index64 i="[0 6]" offset="0" length="2" at="0x7fecbdc45f00"/></offsets>
    <content><RecordArray length="6">
        <field index="0" key="collection">
            <IndexedOptionArray64>
                <index><Index64 i="[-1 0 -1 -1 -1 -1]" offset="0" length="6" at="0x7fec5dd3d180"/></index>
                <content><ListOffsetArray64>
                    <offsets><Index64 i="[0 2]" offset="0" length="2" at="0x7fec5dd3cf00"/></offsets>
                    <content><RecordArray length="2">
                        <field index="0" key="item1">
                            <IndexedOptionArray64>
                                <index><Index64 i="[0 -1]" offset="0" length="2" at="0x7fec5dd3d520"/></index>
                                <content><NumpyArray format="l" shape="1" data="3" at="0x7fecbdc45f10"/></content>
                            </IndexedOptionArray64>
                        </field>
                        <field index="1" key="item2">
                            <IndexedOptionArray64>
                                <index><Index64 i="[-1 0]" offset="0" length="2" at="0x7fec5dd3d4c0"/></index>
                                <content><NumpyArray format="l" shape="1" data="4" at="0x7fecbdc067b0"/></content>
                            </IndexedOptionArray64>
                        </field>
                    </RecordArray></content>
            </IndexedOptionArray64>
        </field>
        <field index="1" key="listcollection">
            <IndexedOptionArray64>
                <index><Index64 i="[0 -1 -1 -1 -1 -1]" offset="0" length="6" at="0x7fec5dd3ce90"/></index>
                <content><ListOffsetArray64>
                    <offsets><Index64 i="[0 2]" offset="0" length="2" at="0x7fecbdd91f90"/></offsets>
                    <content><RecordArray length="2">
                        <field index="0" key="item1">
                            <IndexedOptionArray64>
                                <index><Index64 i="[0 -1]" offset="0" length="2" at="0x7fec5dd3cff0"/></index>
                                <content><ListOffsetArray64>
                                    <offsets><Index64 i="[0 3]" offset="0" length="2" at="0x7fec5dd3a3a0"/></offsets>
                                    <content><NumpyArray format="l" shape="3" data="1 3 2" at="0x7fecbdc4c160"/></content>
                            </IndexedOptionArray64>
                        </field>
                        <field index="1" key="item2">
                            <IndexedOptionArray64>
                                <index><Index64 i="[-1 0]" offset="0" length="2" at="0x7fec5dd3cf70"/></index>
                                <content><ListOffsetArray64>
                                    <offsets><Index64 i="[0 3]" offset="0" length="2" at="0x7fec5dd3d020"/></offsets>
                                    <content><NumpyArray format="l" shape="3" data="2 6 4" at="0x7fecbdc45f20"/></content>
                            </IndexedOptionArray64>
                        </field>
                    </RecordArray></content>
            </IndexedOptionArray64>
        </field>
        <field index="2" key="listsingleton">
            <IndexedOptionArray64>
                <index><Index64 i="[-1 -1 -1 0 -1 -1]" offset="0" length="6" at="0x7fec5dd3d900"/></index>
                <content><ListOffsetArray64>
                    <offsets><Index64 i="[0 3]" offset="0" length="2" at="0x7fec5dd3d9f0"/></offsets>
                    <content><NumpyArray format="l" shape="3" data="1 3 2" at="0x7fecbdc4cd00"/></content>
            </IndexedOptionArray64>
        </field>
        <field index="3" key="masked">
            <IndexedOptionArray64>
                <index><Index64 i="[-1 -1 -1 -1 -1 -1]" offset="0" length="6" at="0x7fec5dd3d280"/></index>
                <content><EmptyArray/></content>
            </IndexedOptionArray64>
        </field>
        <field index="4" key="singleton">
            <IndexedOptionArray64>
                <index><Index64 i="[-1 -1 0 -1 -1 -1]" offset="0" length="6" at="0x7fec5dd3d770"/></index>
                <content><NumpyArray format="l" shape="1" data="5" at="0x7fecbdc3d1d0"/></content>
            </IndexedOptionArray64>
        </field>
        <field index="5" key="unioncollection">
            <IndexedOptionArray64>
                <index><Index64 i="[-1 -1 -1 -1 0 -1]" offset="0" length="6" at="0x7fec5dd3da90"/></index>
                <content><RecordArray length="1">
                    <field index="0" key="item1">
                        <NumpyArray format="l" shape="1" data="3" at="0x7fecbddaeab0"/>
                    </field>
                </RecordArray></content>
            </IndexedOptionArray64>
        </field>
    </RecordArray></content>
</ListOffsetArray64>

```

ianna: @jpivarski - is it expected?

jpivarski: It shouldn't be sorting the fields of the records—they should come out in the order they were first encountered. RecordBuilder maintains that order—in fact, it's a performance thing, since it is optimized under the assumption that the order doesn't change. (A different order doesn't break it, but it does make it slower.)

`to_buffers` outputs JSON as already-serialized text, so that maintains the order, too. I don't know if Python's `json.loads` maintains order, but it should, in Python 3.6+.

The next step is the RecordForm, which also maintains order.

I don't know that it could be. I'll test it as soon as I get to a computer.

jpivarski: It's not the `Form.fromjson` step (the only one I was worried about:

```python
>>> import awkward as ak
>>> ak.forms.Form.fromjson('''{"class": "RecordArray", "contents": {"x": "int32", "y": "float64"}}''')
{
    "class": "RecordArray",
    "contents": {
        "x": "int32",
        "y": "float64"
    }
}
>>> ak.forms.Form.fromjson('''{"class": "RecordArray", "contents": {"y": "float64", "x": "int32"}}''')
{
    "class": "RecordArray",
    "contents": {
        "y": "float64",
        "x": "int32"
    }
}
```

I'll find out what's happening in the full ArrayBuilder example.

jpivarski: Both the high-level and the low-level ArrayBuilders sort the fields:

```python
>>> b = ak.ArrayBuilder()
>>> with b.record():
...     b.field("y").real(3.14)
...     b.field("x").integer(123)
... 
>>> b.snapshot().layout
<RecordArray length="1">
    <field index="0" key="x">
        <NumpyArray format="l" shape="1" data="123" at="0x55fe20dd85a0"/>
    </field>
    <field index="1" key="y">
        <NumpyArray format="d" shape="1" data="3.14" at="0x55fe20e49ae0"/>
    </field>
</RecordArray>
>>> b = ak.layout.ArrayBuilder()
>>> b.beginrecord()
>>> b.field("y")
>>> b.real(3.14)
>>> b.field("x")
>>> b.integer(123)
>>> b.endrecord()
>>> b.snapshot()
<RecordArray length="1">
    <field index="0" key="x">
        <NumpyArray format="l" shape="1" data="123" at="0x55fe20b26dc0"/>
    </field>
    <field index="1" key="y">
        <NumpyArray format="d" shape="1" data="3.14" at="0x55fe20b26de0"/>
    </field>
</RecordArray>
```

but the more basic `to_buffers` does not:

```python
>>> form, length, content = b.to_buffers()
>>> print(form)
{"class": "RecordArray", "contents": {"y": {"class": "NumpyArray", "primitive": "float64", "form_key": "node1"}, "x": {"class": "NumpyArray", "primitive": "int64", "form_key": "node2"}}, "form_key": "node0"}
```

And as I checked above, the `Form.fromjson` method doesn't do the sorting.

```python
>>> ak.forms.Form.fromjson(form)
{
    "class": "RecordArray",
    "contents": {
        "y": {
            "class": "NumpyArray",
            "itemsize": 8,
            "format": "d",
            "primitive": "float64",
            "form_key": "node1"
        },
        "x": {
            "class": "NumpyArray",
            "itemsize": 8,
            "format": "l",
            "primitive": "int64",
            "form_key": "node2"
        }
    },
    "form_key": "node0"
}
```

So the problem is bracketed. Narrowing in on it.

jpivarski: `ak.from_buffers` is sorting it:

```python
>>> ak.from_buffers(form, length, content, key_format="{form_key}-{attribute}").layout
<RecordArray length="1">
    <field index="0" key="x">
        <NumpyArray format="l" shape="1" data="123" at="0x55fe20dd85a0"/>
    </field>
    <field index="1" key="y">
        <NumpyArray format="d" shape="1" data="3.14" at="0x55fe20e49ae0"/>
    </field>
</RecordArray>
```

jpivarski: The v2 version is not sorting it:

```python
>>> ak._v2.operations.convert.from_buffers(form, length, content).layout
<RecordArray is_tuple='false' len='1'>
    <content index='0' field='y'>
        <NumpyArray dtype='float64' len='1'>[3.14]</NumpyArray>
    </content>
    <content index='1' field='x'>
        <NumpyArray dtype='int64' len='1'>[123]</NumpyArray>
    </content>
</RecordArray>
```

So the v1 implementation is wrong.

jpivarski: Here's the problem: the v1 `RecordForm.contents` sorts the fields.

```python
>>> form.contents
{'x': {
    "class": "NumpyArray",
    "itemsize": 8,
    "format": "l",
    "primitive": "int64",
    "form_key": "node2"
}, 'y': {
    "class": "NumpyArray",
    "itemsize": 8,
    "format": "d",
    "primitive": "float64",
    "form_key": "node1"
}}
```

It's because this property uses `std::map` in C++:

https://github.com/scikit-hep/awkward-1.0/blob/e50a2ed5fba578f3916d15da07b678f1b2a17b65/src/python/forms.cpp#L601-L608

But just switching the `std::map` to a `std::unordered_map` doesn't fix it...

jpivarski: Just avoiding the `RecordForm.contents` property fixes it:

```diff
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -4782,7 +4782,9 @@ def _form_to_layout(
         return ak.layout.NumpyArray(array, identities, parameters)
 
     elif isinstance(form, ak.forms.RecordForm):
-        items = list(form.contents.items())
+        items = []
+        for i in range(form.numfields):
+            items.append((form.key(i), form.content(i)))
         if form.istuple:
             items.sort(key=lambda x: int(x[0]))
         contents = []
```

```python
>>> import awkward as ak
>>> b = ak.ArrayBuilder()
>>> with b.record():
...     b.field("y").real(3.14)
...     b.field("x").integer(123)
... 
>>> b.snapshot().layout
<RecordArray length="1">
    <field index="0" key="y">
        <NumpyArray format="d" shape="1" data="3.14" at="0x56139e6fe8e0"/>
    </field>
    <field index="1" key="x">
        <NumpyArray format="l" shape="1" data="123" at="0x56139e9a0880"/>
    </field>
</RecordArray>
```

The problem is that `std::map` explicitly sorts and `std::unsorted_map` is an unstable hashmap, like Python before version 3.6, so we can't fix it by just replacing the `std::map`. I looked for another `std::map` subclass that maintains insertion order, but didn't find any. The answers to [this StackOverflow question](https://stackoverflow.com/q/1098175/1623645) suggest work-arounds (rather like the one I used above). In the case above, the `std::map` is just used to communicate to pybind11 that we want the output to be a `py::dict`. I could have filled a `py::dict` directly in src/python/forms.cpp, and that would solve the issue for C++/Python interface in new Pythons, but not internal C++ usages.

Where are you replacing `std::map` with `std::unsorted_map` in C++? Is it in internal C++ routines or in the C++/Python interface?

jpivarski: To expand on the previous comment, the following is a non-workaround solution:

```diff
--- a/src/python/forms.cpp
+++ b/src/python/forms.cpp
@@ -598,11 +598,10 @@ make_RecordForm(const py::handle& m, const std::string& name) {
           py::arg("has_identities") = false,
           py::arg("parameters") = py::none(),
           py::arg("form_key") = py::none())
-      .def_property_readonly("contents", [](const ak::RecordForm& self)
-           -> std::map<std::string, std::shared_ptr<ak::Form>> {
-        std::map<std::string, std::shared_ptr<ak::Form>> out;
+      .def_property_readonly("contents", [](const ak::RecordForm& self) -> py::dict {
+        py::dict out;
         for (int64_t i = 0;  i < self.numfields();  i++) {
-          out[self.key(i)] = self.content(i);
+          out[py::str(self.key(i))] = py::cast(self.content(i));
         }
         return out;
       })
```

I'll make it a PR. It's #1166; you might want to merge it into this branch.

ianna: @jpivarski - dictionaries are insertion ordered from Python 3.6. That is the reason why the test still fails on some architectures. I can either disable the test for earlier versions of Python or try to switch to a vector of pairs.

jpivarski: You can disable old versions of Python out of hand, just as we've been doing for v2.

We will limit our support to Python 3.6 and higher "pretty soon." There will be one more minor release of Awkward and Uproot with Pythons 2.7 and 3.5, and then we'll drop them. The looming deadline is December 31, when most of the testing and packaging infrastructure will go away (see #1010), so we'll be doing that in less than a month, hopefully the week after ACAT.

agoose77: @ianna @jpivarski Is there plan with v2 to drop (`<=3.6`)  (or even 3.7) in line with NEP 29?

jpivarski: I'm not motivated by NEP 29. Personally, I think the right approach to take is, "keep open as many user options as is easy to do so," which is a two-parameter optimization: ease of users to connect versions X and Y of two libraries and our ease of providing version X that works with Y. That means there's some sort of weighting between user hardship and developer hardship that shouldn't be taken too extreme in either direction. But with most of the tools we use dropping Python 2 support at the end of this year, our developer hardship would go through the roof if we tried to continue maintaining it. That's the reason we'll be giving up on Python 2 users, if there are any of them left. (I don't have good survey data for this year, but last year it was still significant.)

By putting the support window on a schedule, NEP 29 is very friendly toward downstream dependencies, since they can plan for loss of support. However, most of our users are not developers and likely wouldn't know where to look for that schedule if it even occurs to them that there might be one. I think as long as we're not more restrictive than NEP 29 (supporting fewer versions), then we won't be a burden on anyone who does want to follow it (specifically, developers of downstream dependencies). Surprising them with a lack of support would be a problem, but I don't see how it could be bad to surprise them with more support than they want.

jpivarski: Oh, I should add that this is the reason for 3.6+, as opposed to 3.7+ or 3.8+. Python 3.5 has that pesky unstable-dict thing, as well as libraries like `json` still not having a consistent str-vs-bytes policy. I've never encountered any difficulties (developer burdens) specific to Python 3.6. The fact that we've never been prompted to make an `ak._util.py36` flag for customizing code is an indication of how benign it is. I know there are language features in each new version of Python, many of which are very nice (that `match`/`case` in 3.10 looks wonderful), but that's why I'm framing this in terms of "burden," how difficult it is for us to live without new features versus how difficult it is for users to use our software in some overconstrained version hell.

src/libawkward/io/json.cpp
ianna: yes, it counts the number of JSON streams. There is an extra wrapping of the result in a list since it's not known in advance if it's going to be one or many JSON fragments.

jpivarski: When the "json" versus "ndjson" interfaces are well separated, the number of JSON streams is something that only applies to the "ndjson" case, and then it _is_ the `ArrayBuilder::length`. So in the final form of this PR (separating those interfaces is a goal of this PR, right?), the extra integer won't be necessary.

All we need to know at the end is the length of the array that is produced (`ak.from_buffers` needs it), not the lengths of any substructure within that.

src/python/io.cpp
```diff
@@ -15,6 +13,55 @@
...
+
+    void
+      full_buffer(const std::string& name, int64_t length, int64_t value, const std::string& dtype) override {
+        py::object pyarray = py::module::import("numpy").attr("full")(py::int_(length), py::int_(value), py::str(dtype));
+        container_[py::str(name)] = pyarray;
+      }
+
+  private:
+    py::dict container_;
+  };
```
jpivarski: This is a direct copy-paste of

https://github.com/scikit-hep/awkward-1.0/blob/6f622b4ba609e04ea8eb5416d1025990d9c9f4e6/src/python/content.cpp#L979-L1002

Can they be consolidated? Does io.cpp and content.cpp have a common .h file that this class can be defined in?

We don't want it to be a visibly exported symbol, so its (tiny) implementations should remain inline.

jpivarski (changes requested):
We can make the separation of "json" and "ndjson" (either through a boolean argument or through two methods) be a separate PR, after this one. When that's done, the `std::pair<int, const BuilderPtr>` return type can become just `BuilderPtr`, as discussed in a previous comment (above, not part of this review).

The only thing I'd ask for in this PR is to avoid duplicating the NumpyBuffersContainer class. It can go in a header file (with implementation). It's more than a code-cleanliness/DRY thing (although that is important): when working on the specialized JSON parser last week, I found that Windows _randomly_ segfaulted when two classes in different parts of the codebase had the same name. It was a hard-to-diagnose crash because it was failing in a test for the other class with this name—the mere existence of the new class crashed it (because it was existing with a bad name!).

src/python/io.cpp
ianna: Thanks, it's done. I'd prefer to have a separate PR for `json` vs `ndjson` separation. I'd like to check if this is already sufficient to unblock the `ArrayBuilder` pending PR. Thanks!

jpivarski (approved):

src/python/io.cpp
jpivarski: That's much better—thanks!

PR #1166, opened 2021-11-29
Preserve order in v1 RecordForm.contents.

jpivarski:


agoose77:
I ran into this again the other day, and wondered if we'd placed it on the v2 roadmap. Nice!

jpivarski:
For context, this is the conversation where it came up: https://github.com/scikit-hep/awkward-1.0/pull/1164#issuecomment-976964033.

agoose77:
> For context, this is the conversation where it came up: [#1164 (comment)](https://github.com/scikit-hep/awkward-1.0/pull/1164#issuecomment-976964033).

Ah, looks like I should have filed an issue at the time! Nice catch.

PR #1186, opened 2021-12-15
feat: bump cibuildwheel, add Python 3.10

henryiii:


henryiii:
IIRC, llvmlite supports 3.10, but numba only supports 3.10 in master (and only in the last few days). The wheels build, it's just not possible to run the normal tests.

jpivarski:
On the calls, they've still been having troubles with 3.10's bytecode optimization. It's harder to tell where the syntax boundaries are.

henryiii:
`buildtest-awkward (Windows py39-32bit)` needs to be removed or replaced with `buildtest-awkward (Windows py310-32bit)` in the branch protection settings, otherwise this is ready. (there are also a few new 310 checks) :)

jpivarski:
Got it, thanks! I added all the Python 3.10 tests to the requirements.

Great!

henryiii:
This was hard to do before due to Python 2.7 requiring cibuildwheel 1. The newer toml config is nice too. :)

PR #1195, opened 2021-12-28
Fix: _pack_layout should also pack projected index arrays

agoose77:
When _pack_layout is passed an indexed array, it returns `layout.project()`. However,  `_pack_layout` is *supposed* to return a layout whereby only its children may need packing. In this case, by returning `layout.project()`, we are not guaranteed to satisfy this expectation.

This PR fix the issue by ensuring that we pack the result of `project`.

### DIFF
--- a/src/awkward/operations/structure.py
+++ b/src/awkward/operations/structure.py
@@ -2244,7 +2244,7 @@ def _pack_layout(layout):
 
     # Project indexed arrays
     elif isinstance(layout, ak._util.indexedtypes):
-        return layout.project()
+        return _pack_layout(layout.project())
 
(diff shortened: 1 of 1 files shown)


jpivarski:
ak.packed has already been translated to v2; is this issue present in the translated function? We don't want the bug to regress when we update. Thanks!

agoose77:
This should already be fixed in v2 after looking at the source: we immediately call `packed()` on the `project()`ed result

jpivarski (approved):
Great! Then if you think it's done, you can go ahead and "squash and merge" it.

(I just checked: you should have access to do so.)

PR #1199, opened 2022-01-04
chore: bump to pybind11 2.9.0

henryiii:
I forgot there was a 2.8.1 last time - but now there's a 2.9.0, so let's jump to that.


jpivarski (approved):
If the tests pass (and it looks like they will), then I approve. You can squash-and-merge when you're done, or auto-merge now.

PR #1215, opened 2022-01-11
C++ refactoring: Fixed handling of list-nested boolean slices.

jpivarski:
@douglasdavis This fixes the NotImplementedError you mentioned in Slack and also an error with greater-than-depth-1 nested booleans:

```python
>>> a = ak._v2.highlevel.Array([[[0, 1, 2]], [[], [3, 4]], [], [[5], [6, 7, 8, 9]]])
>>> a[a % 2 == 0].show()
```

should not be

```python
[[[1, 0, 1]],
 [[], [3, 4]],
 [],
 [[5], [7, 6, 7, 6]]]
```

but should be

```python
[[[0, 2]],
 [[], [4]],
 [],
 [[], [6, 8]]]
```

(The booleans were being interpreted as integer positions, `0` and `1`, for lists of lists of numbers and deeper. A recursion wasn't getting passed on as it should be.)

### DIFF
--- a/src/awkward/_v2/_slicing.py
+++ b/src/awkward/_v2/_slicing.py
@@ -111,7 +111,11 @@ def prepare_tuple_item(item):
         return item.data
 
     elif isinstance(item, ak._v2.contents.Content):
-        return prepare_tuple_bool_to_int(prepare_tuple_nested(item))
+        out = prepare_tuple_bool_to_int(prepare_tuple_nested(item))
+        if isinstance(out, ak._v2.contents.NumpyArray):
+            return out.data
+        else:
+            return out
 
     elif isinstance(item, Iterable) and all(ak._util.isstr(x) for x in item):
         return list(item)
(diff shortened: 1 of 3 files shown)


src/awkward/_v2/_slicing.py
```diff
@@ -359,52 +378,54 @@ def prepare_tuple_bool_to_int(item):
 
     elif isinstance(item, ak._v2.contents.ListOffsetArray):
         return ak._v2.contents.ListOffsetArray(
-            item.offsets,
-            item.content,
-            identifier=item.identifier,
-            parameters=item.parameters,
+            item.offsets, prepare_tuple_bool_to_int(item.content)
         )
```
jpivarski, line 382: This is the bug-fix beyond just passing through the typetracer.

src/awkward/_v2/_slicing.py
```diff
@@ -359,52 +378,54 @@ def prepare_tuple_bool_to_int(item):
...
 
     elif isinstance(item, ak._v2.contents.IndexedOptionArray):
         if isinstance(item.content, ak._v2.contents.ListOffsetArray):
-            content = prepare_tuple_bool_to_int(item.content)
             return ak._v2.contents.IndexedOptionArray(
-                item.index,
-                content,
-                identifier=item.identifier,
-                parameters=item.parameters,
+                item.index, prepare_tuple_bool_to_int(item.content)
```
jpivarski, line 387: And so is this.

src/awkward/_v2/_slicing.py
```diff
@@ -413,15 +434,12 @@ def prepare_tuple_bool_to_int(item):
 
         else:
             return ak._v2.contents.IndexedOptionArray(
-                item.index,
-                item.content,
-                identifier=item.identifier,
-                parameters=item.parameters,
+                item.index, prepare_tuple_bool_to_int(item.content)
             )
```
jpivarski, line 438: And so is this.

src/awkward/_v2/_slicing.py
```diff
@@ -413,15 +434,12 @@ def prepare_tuple_bool_to_int(item):
...
 
     elif isinstance(item, ak._v2.contents.NumpyArray):
         assert item.data.shape == (item.length,)
-        return item.data
+        return item
```
jpivarski, line 442: This one resolves a type inconsistency. The function should always return a Content, rather than sometimes returning an array. It's been handled at the call site as well.

src/awkward/_v2/contents/regulararray.py
```diff
@@ -33,12 +33,15 @@ def __init__(
...
-        if not (ak._util.isint(size) and size >= 0):
-            raise TypeError(
-                "{0} 'size' must be a non-negative integer, not {1}".format(
-                    type(self).__name__, size
+        if not isinstance(size, ak._v2._typetracer.UnknownLengthType):
+            if not (ak._util.isint(size) and size >= 0):
+                raise TypeError(
+                    "{0} 'size' must be a non-negative integer, not {1}".format(
+                        type(self).__name__, size
+                    )
```
jpivarski, line 41: It's okay for this RegularArray to have an unknown `size` because it's a temporary RegularArray. The final Form will have a known `size`.

PR #1217, opened 2022-01-12
C++ refactoring: ak.flatten

ioanaif:


jpivarski:
`completely_flatten` is a method on `Content` now:

https://github.com/scikit-hep/awkward-1.0/blob/82fda7394418bcae2e4c967f58f9fd5196660803/src/awkward/_v2/contents/content.py#L1231-L1233

It's because it only affects one array. Things that operate on only one array should be methods with overloads on each subclass (the OOP way), rather than functions that `if .. elif .. else` them all. We only resort to `if .. elif .. else` style for functions of more than one array, because then we have to deal with the fact that you can get different combinations of node types. With one array, single-dispatch works well.

So `completely_flatten` should not be reintroduced to `ak._v2._util`. Thanks!

src/awkward/_v2/operations/structure/ak_flatten.py
```diff
@@ -8,163 +8,153 @@
...
+
+    if axis is None:
+        out = (layout.completely_flatten(),)
+
+        assert isinstance(out, tuple) and all(isinstance(x, np.ndarray) for x in out)
+
+        if any(isinstance(x, nplike.ma.MaskedArray) for x in out):
+            out = ak._v2.contents.NumpyArray(nplike.ma.concatenate(out))
+        else:
+            out = ak._v2.contents.NumpyArray(nplike.concatenate(out))
```
jpivarski: v1 made this same mistake:

https://github.com/scikit-hep/awkward-1.0/blob/4071495c93fe50443827bbc286ace45ae5fa7ba0/src/awkward/operations/structure.py#L1903

a NumPy MaskedArray can't be ingested directly into a NumpyArray; it has to use `from_numpy` instead.

On the other hand, I don't think that `completely_flatten` can return a NumPy MaskedArray. Nope, it can't ([v1 implementation](https://github.com/scikit-hep/awkward-1.0/blob/4071495c93fe50443827bbc286ace45ae5fa7ba0/src/awkward/_util.py#L578-L631), which is all in one place).

It should work if you just click this suggestion in. The code will be more maintainable if it's not checking for impossible cases.

```suggestion
        out = ak._v2.contents.NumpyArray(nplike.concatenate(out))
```

tests/v2/test_0150-flatten.py
```diff
@@ -545,43 +667,119 @@ def test_flatten0():
...
+    assert ak._v2.operations.structure.flatten(empty, axis=2).tolist() == []
+
+
+@pytest.mark.skip(reason="ak.repartition() not implemented yet")
+def test_0612():
+    assert ak._v2.operations.structure.flatten(
+        ak.repartition(ak.Array([[0, 1, 2], [], [3, 4], [5], [6, 7, 8, 9]]), 3)
+    ).tolist() == [0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+
```
jpivarski: `ak.repartition` won't be a feature of v2. Just click in this suggestion to delete this test.

```suggestion
```

tests/v2/test_0150-flatten.py
```diff
@@ -545,43 +667,119 @@ def test_flatten0():
...
+    a = ak._v2.contents.IndexedOptionArray(idx, a)
+    idx = ak._v2.index.Index64([0])
+    a = ak._v2.contents.ListOffsetArray(idx, a)
+    idx = ak._v2.index.Index64([175990832])
+    a = ak._v2.contents.ListOffsetArray(idx, a)
+
+    assert ak._v2.operations.structure.flatten(a, axis=2).tolist() == []
+    assert (
+        str(ak._v2.operations.structure.flatten(a, axis=2).type) == "0 * var * ?float64"
+    )
```
jpivarski: This test does not actually use `ak.repartition`. Does it work if you click in this suggestion?

```suggestion
def test_0724():
    a = ak._v2.contents.NumpyArray(np.empty(0))
    idx = ak._v2.index.Index64([])
    a = ak._v2.contents.IndexedOptionArray(idx, a)
    idx = ak._v2.index.Index64([0])
    a = ak._v2.contents.ListOffsetArray(idx, a)
    idx = ak._v2.index.Index64([175990832])
    a = ak._v2.contents.ListOffsetArray(idx, a)

    assert ak._v2.operations.structure.flatten(a, axis=2).tolist() == []
    assert (
        str(ak._v2.operations.structure.flatten(a, axis=2).type) == "0 * var * ?float64"
    )
```

tests/v2/test_0150-flatten.py
```diff
@@ -545,43 +667,119 @@ def test_flatten0():
...
+    a2 = ak.zip({"a": [[7, 8], [9], []], "b": [[10, 11], [12], []]}, with_name="a2")
+    union = ak.where([True, False, True], a1, a2)
+    assert set(ak.flatten(union, axis=None)) == set([1, 2, 3, 4, 5, 6, 9, 12])
+
+
+@pytest.mark.skip(reason="ak.from_numpy not implemented yet")
+def test_0973():
+    array = ak.from_numpy(np.zeros((3, 3, 5)))
+    flattened = ak.flatten(array, axis=None)
+    assert flattened.ndim == 1
```
jpivarski: `ak.from_numpy` is implemented. Try it:

```suggestion
def test_0973():
    array = ak._v2.operations.convert.from_numpy(np.zeros((3, 3, 5)))
    flattened = ak._v2.operations.structure.flatten(array, axis=None)
    assert flattened.ndim == 1
```

jpivarski (changes requested):
Some suggested changes, but they're all clickable. Some of them I'm quite sure will work, and others I'm not so sure. When you click them all in, they'll get commit hashes in case you need to revert any (like, tests I think will work but maybe don't).

jpivarski (approved):

PR #1220, opened 2022-01-13
C++ refactoring: ak.where

ioanaif:


jpivarski (changes requested):
The tests have a few calls to v1 functions, so they're testing v1 functionality, rather than v2. Any that don't exist yet can be moved into `@pytest.mark.skip` tests as a future reminder.

This is what I've been using to check for these:

```bash
egrep 'ak\.[a-zA-Z0-9_]+\(' tests/v2/*.py
```

If we wanted to be super-secure about this, the `egrep` could be added to `pre-commit` somehow, though I don't know how and I think it would be fine to just do it manually.

tests/v2/test_0866-getitem_field-and-flatten-unions.py
```diff
@@ -0,0 +1,25 @@
...
+    union = ak.where([True, False, True], a1, a2)
+    assert str(union.a.type) == "3 * var * int64"
+
+
+@pytest.mark.skip(reason="FIXME: ak.zip needs to be implemented")
+def test_flatten_axis_none():
+    a1 = ak.zip({"a": [[1], [], [2, 3]], "b": [[4], [], [5, 6]]}, with_name="a1")
+    a2 = ak.zip({"a": [[7, 8], [9], []], "b": [[10, 11], [12], []]}, with_name="a2")
+    union = ak.where([True, False, True], a1, a2)
+    assert set(ak.flatten(union, axis=None)) == set([1, 2, 3, 4, 5, 6, 9, 12])
```
jpivarski, line 25: Just be careful when you un-skip these that you also convert them to use v2.

jpivarski (approved):
Great! It's done. I'll merge it.

PR #1224, opened 2022-01-13
Build wheels for ppc64le

chrisburr:
I've found myself having a need for POWER wheels for awkward array. What do you think about including them in the PyPI releases? I'm excluding Python 3.6 and 3.7 to shrink the build matrix but it's up to you if you want to keep that part.

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -73,6 +73,12 @@ jobs:
       matrix:
         python: [36, 37, 38, 39, 310]
         arch: [aarch64, ppc64le]
+        exclude:
+          - arch: ppc64le
+            python: 36
+          - arch: ppc64le
+            python: 37
+
     steps:
 
     - uses: actions/checkout@v2
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -72,7 +72,7 @@ jobs:
     strategy:
       matrix:
         python: [36, 37, 38, 39, 310]
-        arch: [aarch64]
+        arch: [aarch64, ppc64le]
     steps:
 
     - uses: actions/checkout@v2


jpivarski:
I'm fine with this, including the restriction on Python version for particular hardware.

It should probably also get tested in regular CI. That could be hard to do with the Azure tests, but we also run some GitHub Actions tests in CI: `.github/workflows/tests.yml`. Since wheels.yml and tests.yml are on the same system, is it easy to add this as a test?

chrisburr:
It seems like you currently don't do any CI for the aarch64 builds except for when deploying releases?

It should be possible to set up on either GitHub actions or Azure however it will be quite slow as it's using emulation (from experience elsewhere it'll be ~5x slower).

jpivarski:
As I understand it, we're moving from Azure to GitHub Actions, and that's partially done through some GitHub Actions tests in CI that aren't currently required. (I could make them required.)

The current Azure tests do the combinations of {x86_64 Linux, Mac, Windows} × {Python versions}, which are all the same hardware (x86_64). Azure only covers continuous testing, not deployment.

The GitHub Actions that deploy (wheels.yml) have a variety of architectures, recently added by @henryiii.

There are a few GitHub Actions that run in continuous testing (test.yml). I would think that they should test all the same architectures that get deployed. The advantages of doing it in test.yml are that this is what we're moving to and the syntax will be the same as for deployment. These tests are not currently required, but I can make them so.

I didn't realize that there are architectures that get deployed as wheels without testing. The deployment process itself invokes the tests, but wheels.yml only gets triggered manually or in an actual release. If a bug breaks an architecture, we'd want to know that sooner.

jpivarski:
I'm doing a prerelease now, so let's take this as-is to see if it works in the full deployment machinery. I'll revisit this question of whether these hardware architectures (ARM, PowerPC) get regularly tested.

jpivarski:
Hi Chris,

This was a few years ago, but somehow I missed adding you to our list of contributors for this work. Sorry!

@all-contributors please add @chrisburr for infra

PR #1225, opened 2022-01-14
C++ refactoring: ak.zip

ioanaif:


src/awkward/_v2/_util.py
```diff
@@ -479,6 +479,11 @@ def behavior_of(*arrays, **kwargs):
 
 # maybe_wrap and maybe_wrap_like go here
 def wrap(content, behavior=None, highlevel=True, like=None):
+    assert content is None or isinstance(
+        content, (np.generic, ak._v2.contents.Content, ak._v2.record.Record)
+    )
```
jpivarski: `content` should never be `np.generic`. Did you put this in because a test required it? That could be an incorrect test.

`wrap` is _only_ ever used to put `Content` layouts in `highlevel.Array` and `Record` layouts in `highlevel.Record`. It was added because we noticed the same code being written at the end of each `ak.this`/`ak.that` function and consolidated it somewhat. In v1, there were several of these functions with "maybe" in their names; in v2, they're consolidated into one.

src/awkward/_v2/contents/unionarray.py
```diff
@@ -353,11 +353,7 @@ def _nested_tags_index(self, offsets, counts):
 
     def _getitem_next_jagged_generic(self, slicestarts, slicestops, slicecontent, tail):
         simplified = self.simplify_uniontype()
-        if (
-            simplified.index.dtype == np.dtype(np.int32)
-            or simplified.index.dtype == np.dtype(np.uint32)
-            or simplified.index.dtype == np.dtype(np.int64)
-        ):
+        if hasattr(simplified, "index"):
             raise NestedIndexError(
```
jpivarski: In the original v1 code, was this actually asking, "Is `simplified` still a UnionArray after attempting to `simplify_uniontype`?" If so, the v2 code should check to see if isinstance UnionArray, not whether it has `index` as a property. The error message is saying "can't apply jagged slices to irreducible union arrays," and an "irreducible union array" is one that's still a UnionArray after `simplify_uniontype`.

The code, as-is, would match IndexedArrays and IndexedOptionArrays, and I don't think they should be error cases. I think it would be possible to construct such a case and do a test: if you had a union of `?int64` and `int64`, I think that `simplify_uniontype` would turn that into a non-union of `?int64` as an IndexedOptionArray, and then this `hasattr(simplified, "index")` would match when it shouldn't.

src/awkward/_v2/contents/unionarray.py
```diff
@@ -1192,9 +1188,9 @@ def _to_numpy(self, allow_missing):
 
     def _completely_flatten(self, nplike, options):
         out = []
-        for content in self._contents:
+        for i in range(len(self._contents)):
             out.extend(
-                content[: self._tags.length]._completely_flatten(nplike, options)
+                self._contents[i][self._tags[i] :]._completely_flatten(nplike, options)
```
jpivarski: This might be passing tests, but I think it's wrong. The number of items in the list `self._contents` is the number of types in the union, like two or three. The number of items in the array `self._tags` is the length of the union array itself, like a million. They can't both be indexed by the same `i`.

I think the original code you changed this from was wrong, too, so let's think it through. The purpose of this is to only descend the `_completely_flatten` on items in `self._contents[i]` that are reachable, i.e. to exclude the "hidden" data. After picking the single `Content` with `self._contents[i]`, the ones that are accessible are the ones that are picked out by index. The `self._index` includes all contents, not just `self._contents[i]`, so that, too, needs to be selected. The index values that you want correspond to `self._tags == i`, so I think the

```python
self._contents[i][self._tags[i]:]._completely_flatten(nplike, options)
```

has to be replaced by

```python
self._contents[i][self._index[self._tags == i]]._completely_flatten(nplike, options)
```

or maybe more directly

```python
self._contents[i]._carry(self._index._carry(self._tags.data == i, ...), ...)._completely_flatten(nplike, options)
```

with the appropriate arguments for `_carry` (to save it from having to go through all the user-friendliness checks that `__getitem__` does before it finds out that slicing by an array ought to be a call to `_carry`. Though `_carry` is somewhat restrictive: I think you have to give it an `Index`, rather than an array. Maybe you'd want to do this in several steps, getting arrays from `self._index` and `self._tags` and doing the slices in NumPy space before making an `Index` and carrying `self._contents[i]`.

jpivarski (changes requested):
Okay, the `zip` implementation looks good! My corrections are for other parts that were updated along with it.

src/awkward/_v2/_util.py
ioanaif: Yes, it was added to deal with 11 failing tests.  

Among which, **tests/v2/test_0410-fix-argminmax-positions-for-missing-values.py:471**

  Here, `ak._v2.operations.reducers.max` https://github.com/scikit-hep/awkward-1.0/blob/5f5687fde4bd15ba67707c12370b7cc15bcdae43/src/awkward/_v2/operations/reducers/ak_max.py#L70 calls layout.max which returns an np.generic. Transforming to a content type before passing to `wrap` causes awkward/_v2/highlevel.py:1021 to fail as `self._layout` is a boolean array which is passed again to `wrap` after being sliced (passing a boolean).
  
The other tests that fail in this case: 
    
- tests/v2/test_0013-error-handling-struct.py:39: 
- tests/v2/test_0013-error-handling-struct.py:115: 
- tests/v2/test_0089-numpy-functions.py:86: 
- tests/v2/test_0089-numpy-functions.py:290: 
- tests/v2/test_0111-jagged-and-masked-getitem.py:1160: 
- tests/v2/test_0315-integerindex.py:133: 
- tests/v2/test_0410-fix-argminmax-positions-for-missing-values.py:439: 
- tests/v2/test_0835-datetime-type.py:661: 
- tests/v2/test_0835-datetime-type.py:685: 
- tests/v2/test_0959-_getitem_array-implementation.py:17:

src/awkward/_v2/contents/unionarray.py
ioanaif: Indeed, it should have checked if it is an instance of UnionArray, not if it has the index attribute. I updated it accordingly.

jpivarski (approved):

jpivarski:
It looks good now (your changes in addition to my restriction of `wrap` to Content/Record).

I'm going to enable auto-merge. If you have anything else to add (I assume you don't), let me know before the tests pass. Thanks!

src/awkward/_v2/_util.py
jpivarski: Fixed in 80cc45decc319d899018ba6bfcce33a9ec1d61be.

jpivarski:
This and #1226 conflicted in a test (because #1226 got committed while this PR was still testing), so I had to resolve the conflict and start a new test-merge cycle.

jpivarski:
`ak.num` also has the possibility of returning a number, rather than a Content/Record/Array, so it failed the new `wrap` type-check as well. 92eee96ddd6c28f08b5e95503b869c2e26e2bac5 fixes it.

Even though this is starting to look like an annoyance, I'd rather have it work this way: `wrap` is acting as a (non-static) type-check, ensuring that we really know whether the output is always an array or can sometimes be scalar. Not all uses of `wrap` have to be guarded—most functions return arrays for all inputs. The ones that had to be updated were reducers, `__getitem__`, `__iter__`, and now `num`, which have known cases that can return scalars. The guard makes that more apparent, and it could be a stepping-stone to someday typing the whole codebase. Someday.

PR #1231, opened 2022-01-19
Updated the generate-cuda script. Works for py >= 3.8

sw5sh:
The `ast` library had an API change for python >= 3.8. https://bugs.python.org/issue40430. I couldn't update the script to work with all versions of Python, since doing that leads to more dead ends. After talking to Jim, the generated kernels will be checked in to the source tree very soon, so getting the script to work on any recent Python version will work for now.

### DIFF
--- a/dev/generate-cuda.py
+++ b/dev/generate-cuda.py
@@ -199,15 +199,9 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
     elif node.__class__.__name__ == "Mult":
         code = "*"
     elif node.__class__.__name__ == "Subscript":
-        if (
-            node.slice.__class__.__name__ == "Name"
-            and node.slice.id in forvars
-        ):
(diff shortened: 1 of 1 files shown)


jpivarski:
(First question: is this meant to be a draft? I don't know if the changes here are intended to be final. The tests pass! `:)`)

As I said in Slack, this script is not intended to have a long lifetime. Pretty soon, we'll generate the CUDA source code one last time and then check it into git for maintenance by hand. Then the script will just be removed. Therefore, it doesn't need to run in all versions of Python; just get it to run in the most convenient version of Python, probably 3.10 because that's the one you're using on your laptop. (I'm using 3.9, but I think all the changes to the ast module are between 3.7 and 3.8.)

sw5sh:
This is it! You can go ahead and merge it. The script works on my machine which has Python 3.10. And I'll try to check in the auto-generated kernels by the end of next week so this script would have served it's purpose by then.

PR #1232, opened 2022-01-19
macos segfault bugfix

### LINKED ISSUE #1209: MacOS tests randomly segfault
### Version of Awkward Array

HEAD

For a while now (weeks?), the MacOS `tests/**/*.py` have been failing with low probability—there's about a 1 in 5 chance that one of the tests (based on Python 3.6, 3.7, 3.8, 3.9, and 3.10) raises a segfault. [Here is an example.](https://dev.azure.com/jpivarski/Scikit-HEP/_build/results?buildId=8629&view=logs&j=a8b274c7-ba97-5314-7052-52a2bff17ff3&t=70094b08-aa59-59b5-e85c-b680ecf0ab0f) When that happens, the error code may be one of several signals, such as 6, 10, or 11. It often fails while running v2 tests, but that might just be because the v2 tests are after than the v1 tests. The fact that it hasn't been failing consistently in the same test suggests that the signal is invoked during a garbage collection, but the actual error is earlier.

If this really is due to a v1 segfault, it's rare enough that we could even ignore it—there won't be many more v1 versions released. If it's due to something in v2, then we really must take care of it. (And if it's narrowed down to an easy v1 issue, then we'd also want to fix it in that case.)

If this had been a Linux segfault, I would run the tests locally many times in order to reproduce it (e.g. run it 100 times to virtually guarantee the P ≈ 1/25 case, then I would bisect the tests to find at least one that raises it. It might be necessary to insert explicit `gc.collect()` calls to trigger it in absence of many other tests. However, the bug only manifests itself on MacOS, and I don't have access to a Mac to do that intensive testing on. (Doing it through CI would be awful because it would add the compilation time to each run.)

Does anyone (@ianna, @ioanaif, @swishdiff, @agoose77, ...) have a Mac and is willing to try that bisection search? Maybe my first question should be if any Mac users have encountered this in their local tests (assuming you've been running enough tests recently).

ianna:
address issue #1209

### DIFF
--- a/src/cpu-kernels/awkward_IndexedArray_numnull_parents.cpp
+++ b/src/cpu-kernels/awkward_IndexedArray_numnull_parents.cpp
@@ -16,6 +16,9 @@ ERROR awkward_IndexedArray_numnull_parents(
       numnull[i] = 1;
       *tolength = *tolength + 1;
     }
+    else {
+      numnull[i] = 0;
+    }
   }
   return success();
 }
--- a/src/cpu-kernels/awkward_unique_offsets.cpp
+++ b/src/cpu-kernels/awkward_unique_offsets.cpp
@@ -14,17 +14,15 @@ ERROR awkward_unique_offsets(
   int64_t j = 0;
   for (int64_t i = 0;  i < length;  i++) {
     tooffsets[j] = fromoffsets[i];
-    while (starts[j] == starts[j + 1]) {
-      tooffsets[j + 1] = fromoffsets[i];
-      if (j < startslength - 1) {
+    for (int64_t k = j; k < startslength - 1; k++) {
+      if (starts[j] == starts[j + 1]) {
+        tooffsets[j + 1] = fromoffsets[i];
         j++;
-      } else {
-        break;
       }
     }
     j++;
   }
-  tooffsets[j - 1] = fromoffsets[length - 1];
+  tooffsets[startslength] = fromoffsets[length - 1];
 
   return success();
 }
--- a/kernel-specification.yml
+++ b/kernel-specification.yml
@@ -1331,6 +1331,8 @@ kernels:
               if fromindex[i] < 0:
                   numnull[i] = 1
                   tolength[0] = tolength[0] + 1
+              else:
+                  numnull[i] = 0
     automatic-tests: false
     manual-tests: []
 
--- a/src/awkward/_v2/contents/bitmaskedarray.py
+++ b/src/awkward/_v2/contents/bitmaskedarray.py
@@ -354,7 +354,7 @@ class BitMaskedArray(Content):
         if isinstance(
             other,
             (
-                ak._v2.contents.emptyArray.EmptyArray,
+                ak._v2.contents.emptyarray.EmptyArray,
                 ak._v2.contents.unionarray.UnionArray,
             ),
         ):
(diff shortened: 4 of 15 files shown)


ianna:
@jpivarski - there was a bug in a cpu kernel. It did not manifest itself in v1 because the kernel was not used there. Now, I'm running the tests with a higher statistic to see if the segfault is indeed gone - there are no crashes so far.

ianna:
@jpivarski - I'm done with this PR. There are no crashes reproduced so far..

src/cpu-kernels/awkward_IndexedArray_numnull_parents.cpp
```diff
@@ -16,6 +16,9 @@ ERROR awkward_IndexedArray_numnull_parents(
       numnull[i] = 1;
       *tolength = *tolength + 1;
     }
+    else {
+      numnull[i] = 0;
+    }
   }
   return success();
 }
```
jpivarski, line 24: So it was either this...

src/cpu-kernels/awkward_unique_offsets.cpp
```diff
@@ -14,17 +14,15 @@ ERROR awkward_unique_offsets(
...
-        break;
       }
     }
     j++;
   }
-  tooffsets[j - 1] = fromoffsets[length - 1];
+  tooffsets[startslength] = fromoffsets[length - 1];
 
   return success();
 }
```
jpivarski, line 28: Or this? Either way, that's fantastic! Thank you so much for finding it!

jpivarski (approved):
> @jpivarski - there was a bug in a cpu kernel. It did not manifest itself in v1 because the kernel was not used there. Now, I'm running the tests with a higher statistic to see if the segfault is indeed gone - there are no crashes so far.

:tada: Wonderful!

The two kernel changes below look very plausibly like they could cause this. The awkward_IndexedArray_numnull_parents.cpp is obvious in retrospect; if we use `numnull` at all after this kernel (which we must; otherwise, why fill it?), then it would be a problem that every `numnull[i]` for which `fromindex[i] >= 0` is undefined. I don't understand the awkward_unique_offsets.cpp change as well, but it looks significant.

Most importantly, these were bugs on all platforms, and for v1 as well as v2. It's not the case that v2 was adding some new instability; it was doing better testing on an old bug (or two old bugs, if I understand correctly).

I just updated the kernel-specification.yml (one of these two kernels had a Python definition; I don't want it to get out of date). When it's done with its new tests, I'll squash-and-merge, and then start a new release, now that this blocker has been fixed.

Again, thanks for all the work that went into this! Seeing the glaring bug, it was well worth it!

jpivarski:
I'm also running https://github.com/scikit-hep/awkward-1.0/actions/runs/1719297682 to make sure that the deployment process won't run into any segfaults (as it did last time).

jpivarski:
@henryiii, this PR definitely fixes a segfault and it seems to be the one @ianna saw on her Mac. (She could reproduce the segfault before fixing the bug, but not after, so that looks good.) So it was good to merge the PR.

However, the wheels.yml test failed, this time in Windows 32-bit: https://github.com/scikit-hep/awkward-1.0/runs/4871371245?check_suite_focus=true

If I'm reading this log correctly, it failed to install NumPy. It doesn't look like a segfault. (I also looked at the raw logs, in case the formatting is hiding something.) Is that your interpretation of it as well? Is it just another case that we should exclude because NumPy is not supporting the platform/Python version combination?

henryiii:
NumPy does not ship windows 32-bit wheels for Python 3.10: https://pypi.org/project/numpy/1.22.1/#files

jpivarski:
Okay, thanks! I'm following up on #1237.

PR #1234, opened 2022-01-19
C++ refactoring: ak.type and ak.values_astype

ioanaif:


src/awkward/_v2/contents/numpyarray.py
```diff
@@ -581,13 +581,20 @@ def _as_unique_strings(self, offsets):
...
-            self._identifier,
-            self._parameters,
-            self._nplike,
-        )
+        if (
+            self.parameter("__array__") == "string"
+            or self.parameter("__array__") == "bytestring"
+            or self.parameter("__array__") == "char"
+        ):
+            return self
```
jpivarski:
```suggestion
        if (
            self.parameter("__array__") == "string"
            or self.parameter("__array__") == "bytestring"
            or self.parameter("__array__") == "char"
            or self.parameter("__array__") == "byte"
        ):
            return self
```

These kinds of checks may be proliferating, so they should probably become a utility function in `ak._v2._util`. That would best be handled in another PR, though.

src/awkward/_v2/operations/describe/ak_type.py
```diff
@@ -2,147 +2,154 @@
...
+            np.uint32,
+            np.uint64,
+            np.float32,
+            np.float64,
+            np.complex64,
+            np.complex128,
+            np.datetime64,
+            np.timedelta64,
+        ),
+    ):
```
jpivarski:
```suggestion
    elif isinstance(
        array,
        [x.type for x in ak._v2.types.numpytype._dtype_to_primitive_dict],
    ):
```

This _should_ work. There's a canonical list of handled dtypes in `ak._v2.types.numpytype._dtype_to_primitive_dict` and we should use that, rather than introduce another one here.

src/awkward/_v2/operations/describe/ak_type.py
```diff
@@ -2,147 +2,154 @@
...
+    np.uint16: "uint16",
+    np.uint32: "uint32",
+    np.uint64: "uint64",
+    np.float32: "float32",
+    np.float64: "float64",
+    np.complex64: "complex64",
+    np.complex128: "complex128",
+    np.datetime64: "datetime64",
+    np.timedelta64: "timedelta64",
+}
```
jpivarski: The equivalent of this is in `ak._v2.types.numpytype._dtype_to_primitive_dict`, though that one takes dtypes to strings, rather than types to strings.

(It's necessary to use dtypes, rather than types, because a type like `np.datetime64` doesn't carry units, but a dtype like `np.dtype("datetime64[us]")` does.)

src/awkward/_v2/operations/describe/ak_type.py
```diff
@@ -2,147 +2,154 @@
...
+            np.uint64,
+            np.float32,
+            np.float64,
+            np.complex64,
+            np.complex128,
+            np.datetime64,
+            np.timedelta64,
+        ),
+    ):
+        return ak._v2.types.PrimitiveType(type.dtype2primitive[array.dtype.type])
```
jpivarski:
```suggestion
        return ak._v2.types.PrimitiveType(ak._v2.types.numpytype._dtype_to_primitive_dict[array.dtype])
```

Both this and the above change have to go together. I should have made it one edit.

src/awkward/_v2/operations/structure/ak_values_astype.py
```diff
@@ -8,90 +8,86 @@
...
+    np.dtype(np.datetime64): "datetime64",
+    np.dtype(np.timedelta64): "timedelta64",
+}
+
+if hasattr(np, "float16"):
+    _dtype_to_string[np.dtype(np.float16)] = "float16"
+if hasattr(np, "float128"):
+    _dtype_to_string[np.dtype(np.float128)] = "float128"
+if hasattr(np, "complex256"):
+    _dtype_to_string[np.dtype(np.complex256)] = "complex256"
```
jpivarski: This should also be removed, in favor of `ak._v2.types.numpytype._dtype_to_primitive_dict`.

jpivarski (changes requested):
On the whole, it's looking good, but I have a few requests. Most of them are about using centralized utilities, reducing the proliferation of type-lists that were in v1.

The type name for leaves of the type tree changed from PrimitiveType to NumpyType. As far as I can see, there is no `ak._v2.types.PrimitiveType`, so how are these tests working? Virtually every attempt to build a type would have to terminate on these leaf-nodes, so you ought to be getting

```
AttributeError: module 'awkward._v2.types' has no attribute 'PrimitiveType'
```

(I don't see a PrimitiveType being defined in this diff.)

Anyway, in an attempt to use terminology more consistently, the NumpyArray has NumpyForm and NumpyType. We still use the word "primitive," but it refers to the _string_ naming the type in Forms. v1 and v2 have to be able to accept each others' Forms so that pickled data from each version can be used in the other (that was a goal, so that users could build work-arounds if they need to), and that means the JSON field name "primitive" in the Form can't be changed. (Otherwise, a better name _might_ be "dtype", though that would also be misleading because dtypes are Python objects, and this is a string. So it's just going to be "primitive".)

jpivarski (approved):

PR #1237, opened 2022-01-19
Remove Windows 32-bit from the Python 3.10 build.

jpivarski:


jpivarski:
Testing the deployment here, by the way: https://github.com/scikit-hep/awkward-1.0/actions/runs/1719986605

henryiii:
You could skip it instead (or also) in pyproject.toml; the benefit of the skip is that it works locally too. Technically, we don't really support local Windows runs yet... But the next cibuildwheel release should. Quite minor, this is fine.

jpivarski:
Now the MacOS build [gets this far before failing](https://pipelines.actions.githubusercontent.com/ErlVSyz7hapZD5oDnavYEewPSZubvgO3c79oCbrFWfDzi2sAj4/_apis/pipelines/1/runs/858/signedlogcontent/11?urlExpires=2022-01-19T22%3A20%3A38.4990933Z&urlSigningMethod=HMACV1&urlSignature=3gcX1tCu56ANRYConwgqhMHIq2gerdZfelrIIt5G2cw%3D):

```
2022-01-19T21:11:38.4243380Z   build/lib.macosx-10.9-x86_64-3.9/awkward/libawkward-cpu-kernels.dylib
2022-01-19T21:11:38.4243790Z   build/lib.macosx-10.9-x86_64-3.9/_ext.cpython-39-darwin.so
2022-01-19T21:11:38.4244150Z   build/lib.macosx-10.9-x86_64-3.9/libawkward.dylib
2022-01-19T21:11:38.4244530Z   build/lib.macosx-10.9-x86_64-3.9/libawkward-cpu-kernels.dylib
2022-01-19T21:11:38.4244930Z   --- copying includes ------------------------------------------
2022-01-19T21:11:38.4245350Z   error: [Errno 17] File exists: 'build/lib.macosx-10.9-x86_64-3.8/awkward/include'
2022-01-19T21:11:38.4245790Z   Building wheel for awkward (pyproject.toml): finished with status 'error'
2022-01-19T21:11:38.4246070Z   ERROR: Failed building wheel for awkward
2022-01-19T21:11:38.4246280Z Failed to build awkward
2022-01-19T21:11:38.4246500Z ERROR: Failed to build one or more wheels
```

I suppose we don't really need any includes, now that we're not supporting downstream dependencies of the C++ part.

They'd be needed in the SDist, but not in wheels.

I'm going to try this:

```diff
--- a/setup.py
+++ b/setup.py
@@ -193,11 +193,6 @@ class Install(setuptools.command.install.install):
         print("--- build directory -------------------------------------------")
         tree("build")
 
-        print("--- copying includes ------------------------------------------")
-        shutil.copytree(
-            os.path.join("include"), os.path.join(outerdir, "awkward", "include")
-        )
-
         print("--- outerdir after copy ---------------------------------------")
         tree(outerdir)
```

By the same logic, we also don't need to bundle any static libraries, just the dynamic ones. I'll start with this, though.

jpivarski:
Here's the next one: https://github.com/scikit-hep/awkward-1.0/actions/runs/1720551239

henryiii:
I think the problem is setuptools doesn't understand that PyPy and CPython are different. If you split the CPython and PyPy jobs, I think it will be fine. Not sure why PyPy 3.7 doesn't show that problem, and 3.8 does, though - but it might be "better" (more CPython-like) handling of build paths for PyPy 3.8.

See https://github.com/pypa/setuptools/issues/2912

jpivarski:
I keep doing other things and then coming back to check up on this. You have something specific in mind—could you set it up the way you're thinking?

henryiii:
I've reverted the include file removal to verify my suspicion was correct, but if it's not useful, the static files should be removed, as they make the wheels larger (libs much more than include).

jpivarski:
Great, **thank you very much**! I see that it has been tested here: https://github.com/scikit-hep/awkward-1.0/actions/runs/1721784145 (the commit hex matches), so we know that it will work in a real release. I'll get started on that release right now.

jpivarski:
> if it's not useful, the static files should be removed, as they make the wheels larger (libs much more than include).

The static libraries are not useful, but the next release will be a prerelease. The frequency of non-prereleases has gone down considerably in the past year, so that alleviates some of the disk space pressure. Nevertheless, before 1.8.0, I'll try removing the static libraries.

PR #1239, opened 2022-01-20
Revert "Build wheels for ppc64le"

jpivarski:
Reverts scikit-hep/awkward-1.0#1224

@chrisburr I should have done a manual CD test, first. It didn't deploy: https://github.com/scikit-hep/awkward-1.0/actions/runs/1724731128

I'm reverting this to release 1.8.0rc3. We'll revisit it!

### DIFF
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -77,13 +77,7 @@ jobs:
     strategy:
       matrix:
         python: [36, 37, 38, 39, 310]
-        arch: [aarch64, ppc64le]
-        exclude:
-          - arch: ppc64le
-            python: 36
-          - arch: ppc64le
-            python: 37
-
+        arch: [aarch64]
     steps:
 
     - uses: actions/checkout@v2


chrisburr:
Ah I think it fails to test because numpy doesn't have ppc64le wheels (when I tested numpy was already installed so I didn't notice). It seems it's on [the roadmap](https://numpy.org/neps/roadmap.html#platform-support) but I guess it'll have to wait for now unless you'd consider releasing untested ppc64le wheels?

jpivarski:
Okay, that explains it. If NumPy doesn't have PowerPC wheels, I don't think it helps much for Awkward to have PowerPC wheels because PowerPC users would have to compile _something_, anyway. (I'm assuming that it's at least as hard to compile NumPy as Awkward.) When they do, we can reinstate this, though.

chrisburr:
> If NumPy doesn't have PowerPC wheels, I don't think it helps much for Awkward to have PowerPC wheels because PowerPC users would have to compile something, anyway.

Numpy is a fairly standard thing to find pre-installed so I think in most cases it's probably not an issue.

jpivarski:
I see—you're thinking it would come from the OS distribution's package manager. (I used to do that, but now I always get it from pip or conda. I don't trust my OS distribution's package manager for any Python packages.)

If it's reintroduced into the Awkward build, then GitHub Actions would have to compile NumPy, which isn't terrible, but it's a slow-down. It's not something we'd want to have happen in regular CI builds, though maybe a CD release.

PR #1242, opened 2022-01-21
C++ refactoring: ak.parameters

ioanaif:


jpivarski (approved):
Looks great!

We talked about these things on Slack, but for posterity:

   * The new `merge_equal` flag for `merge_parameters` is good, as it bring the use of this function in line with one of the same name in v1. The other use of this function is in Arrow conversions, which has the different merging behavior, and that's why you need a flag.
   * A lot of the changes in the tests are replacing `XYZ.parameters` with `ak._v2.operations.describe.parameters(XYZ)`, which is fine. The _behavior_ defined by the tests has not changed.

PR #1243, opened 2022-01-21
style: add shellcheck

henryiii:
This cleans up the recommendations produced for cuda-build.sh. I don't think this is checked in CI, and I haven't run the changes. Not sure the old one wwas working properly either. This really needs to be further redone, it should not be injecting things like the WHEEL file; wheels are hashed and can't be edited like this. And there are ways to indicate they are not pure Python other than hacking it like this. But this is a start. CC @swishdiff

### DIFF
--- a/cuda-build.sh
+++ b/cuda-build.sh
@@ -6,41 +6,35 @@ set -e
 PLATFORM="manylinux2014_x86_64"
 CUDA_VERSION=9.0
 
-AWKWARD_VERSION=`cat VERSION_INFO | tr -d '[:space:]'`
+AWKWARD_VERSION=$(tr -d '[:space:]' < VERSION_INFO)
 
-if [[ "$CUDA_VERSION" == "11.0" ]]; then
-    export DOCKER_IMAGE_TAG="11.0-devel-ubuntu18.04"
-elif [[ "$CUDA_VERSION" == "10.2" ]]; then
-    export DOCKER_IMAGE_TAG="10.2-devel-ubuntu18.04"
-elif [[ "$CUDA_VERSION" == "10.1" ]]; then
-    export DOCKER_IMAGE_TAG="10.1-devel-ubuntu18.04"
-elif [[ "$CUDA_VERSION" == "10.0" ]]; then
-    export DOCKER_IMAGE_TAG="10.0-devel-ubuntu18.04"
-elif [[ "$CUDA_VERSION" == "9.2" ]]; then
-    export DOCKER_IMAGE_TAG="9.2-devel-ubuntu18.04"
-elif [[ "$CUDA_VERSION" == "9.1" ]]; then
-    export DOCKER_IMAGE_TAG="9.1-devel-ubuntu16.04"
-elif [[ "$CUDA_VERSION" == "9.0" ]]; then
-    export DOCKER_IMAGE_TAG="9.0-devel-ubuntu16.04"
-elif [[ "$CUDA_VERSION" == "8.0" ]]; then
-    export DOCKER_IMAGE_TAG="8.0-devel-ubuntu16.04"
-else
-    echo "Docker image for CUDA version" $CUDA_VERSION "is not known"
+case $CUDA_VERSION in
+  "11.0" | "10.2" | "10.1" | "10.0" | "9.2")
+    export DOCKER_IMAGE_TAG="$CUDA_VERSION-devel-ubuntu18.04"
+    ;;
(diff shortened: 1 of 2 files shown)


jpivarski:
It's good that you cc'ed Anish. Anish, have you started work on this file? Would it be easy or difficult to merge in these changes? This looks to me like a much-improved style, and a good point to start from.

cuda-build.sh
```diff
@@ -97,17 +93,17 @@ EOF
...
-rm -f awkward_cuda_kernels-$AWKWARD_VERSION-py3-none-$PLATFORM.whl
+rm -f "awkward_cuda_kernels-$AWKWARD_VERSION-py3-none-$PLATFORM.whl"
 
-unzip awkward_cuda_kernels-$AWKWARD_VERSION-py3-none-any.whl
+unzip "awkward_cuda_kernels-$AWKWARD_VERSION-py3-none-any.whl"
 
-cp awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL tmp_WHEEL
-cat tmp_WHEEL | sed "s/Root-Is-Purelib: true/Root-Is-Purelib: false/" | sed "s/Tag: py3-none-any/Tag: py3-none-"$PLATFORM"/" > awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL
+cp "awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL" tmp_WHEEL
+sed "s/Root-Is-Purelib: true/Root-Is-Purelib: false/" < tmp_WHEEL | sed "s/Tag: py3-none-any/Tag: py3-none-$PLATFORM" > "awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL"
```
sw5sh: This line raises an ``sed: -e expression #1, char 54: unterminated `s' command``. Here's the fixed line:
```sh
sed "s/Root-Is-Purelib: true/Root-Is-Purelib: false/" < tmp_WHEEL | sed "s/Tag: py3-none-any/Tag: py3-none-$PLATFORM/" > "awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL"
```

sw5sh:
This is a good start, I can work with this. I have run the script on my laptop (Python 3.10 CUDA 11.5) and the build works after fixing the minor `issue` I raised in the code review! 
PS: I am currently reading up on GitHub Actions and Python Wheels. I would love some article recommendations on best practices. Thanks!

henryiii:
https://scikit-hep.org/developer is a good start, IMO, though I'm biased. :)

henryiii:
Is there something I should be addressing? I don't see a code comment.

sw5sh:
image
This is weird. It shows up in my page but not yours. But I'll put the concerned thing down here:

Regarding Line 101 in cuda-build.sh: 
> This line raises an ``sed: -e expression #1, char 54: unterminated `s' command``. Here's the fixed line:
> ```sh
> sed "s/Root-Is-Purelib: true/Root-Is-Purelib: false/" < tmp_WHEEL | sed "s/Tag: py3-none-any/Tag: py3-none-$PLATFORM/" > > "awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL"
> ```

Also, thanks for the developer page link!

henryiii:
There's a (pending); you need to "submit" the review after you start one (it does that so it doesn't notify multiple times).

henryiii:
And you can use the "suggest changes" button to make it easier to see the difference and apply it. ;)

sw5sh (commented):
This should be it! My bad :)

cuda-build.sh
henryiii:
```suggestion
sed "s/Root-Is-Purelib: true/Root-Is-Purelib: false/" < tmp_WHEEL | sed "s/Tag: py3-none-any/Tag: py3-none-$PLATFORM/" > "awkward_cuda_kernels-$AWKWARD_VERSION.dist-info/WHEEL"
```

jpivarski:
This looks done. @swishdiff, if you're happy with it in its current state, you may squash-and-merge it. (I get the impression that @henryiii considers it done, so I'm asking you because it affects your work.)

sw5sh:
Yup, I am actually building on this version of the script in my branch, so getting it merged into main would ease my job.

jpivarski:
Oh, then let's do it. Not just now, but in general let me know if an outstanding PR is a blocker.

PR #1214, opened 2022-01-11
Fix: drop parameters for flattened RecordArray

agoose77:

PR #1200, opened 2022-01-05
replace shared_ptr with unique_ptr in Forth buffers

ianna:

PR #1244, opened 2022-01-21
style: pyupgrade to 3.6

henryiii:
This is completely automated. There are lots of further cleanups, but this is the automated stuff.

### DIFF
--- a/dev/generate-cuda.py
+++ b/dev/generate-cuda.py
@@ -94,7 +94,7 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
         else:
             raise Exception("Cannot handle more than triply nested loops")
         if len(node.iter.args) == 1:
-            code = "if ({0} < {1}) {{\n".format(thread_var, traverse(node.iter.args[0]))
+            code = f"if ({thread_var} < {traverse(node.iter.args[0])}) {{\n"
         elif len(node.iter.args) == 2:
             code = "if (({0} < {1}) && ({0} >= {2})) {{\n".format(
                 thread_var, traverse(node.iter.args[1]), traverse(node.iter.args[0])
@@ -107,17 +107,17 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
     elif node.__class__.__name__ == "While":
         assert node.test.__class__.__name__ == "Compare"
         assert len(node.test.ops) == 1
-        code = "while ({0}) {{\n".format(traverse(node.test))
+        code = f"while ({traverse(node.test)}) {{\n"
         for subnode in node.body:
             code += traverse(subnode, args, copy.copy(forvars), declared)
         code += "}\n"
     elif node.__class__.__name__ == "Raise":
         if sys.version_info[0] == 2:
-            code = 'err->str = "{}";\n'.format(node.type.args[0].s)
+            code = f'err->str = "{node.type.args[0].s}";\n'
         elif sys.version_info[0] == 3 and sys.version_info[1] in [5, 6, 7]:
-            code = 'err->str = "{}";\n'.format(node.exc.args[0].s)
+            code = f'err->str = "{node.exc.args[0].s}";\n'
         else:
-            code = 'err->str = "{}";\n'.format(node.exc.args[0].value)
+            code = f'err->str = "{node.exc.args[0].value}";\n'
         code += "err->filename = FILENAME(__LINE__);\nerr->pass_through=true;\n"
     elif node.__class__.__name__ == "If":
         code = ""
@@ -159,7 +159,7 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
         elif node.value is False:
             code = "false"
         else:
-            raise Exception("Unhandled NameConstant value {}".format(node.value))
+            raise Exception(f"Unhandled NameConstant value {node.value}")
     elif node.__class__.__name__ == "Num":
         code = str(node.n)
     elif node.__class__.__name__ == "BinOp":
@@ -295,7 +295,7 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
                 operator = "<<="
             else:
                 raise Exception(
-                    "Unhandled AugAssign node {}".format(node.op.__class__.__name__)
+                    f"Unhandled AugAssign node {node.op.__class__.__name__}"
                 )
             targetnode = node.target
         left = traverse(targetnode, args, copy.copy(forvars), declared)
@@ -397,7 +397,7 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
                     traverse(node.value, args, copy.copy(forvars), declared),
                 )
     else:
-        raise Exception("Unhandled node {}".format(node.__class__.__name__))
+        raise Exception(f"Unhandled node {node.__class__.__name__}")
     return code
 
 
--- a/dev/generate-cuda.py
+++ b/dev/generate-cuda.py
@@ -1,6 +1,5 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
-from __future__ import absolute_import
 
 import argparse
 import ast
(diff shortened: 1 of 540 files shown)


dev/generate-cuda.py
```diff
@@ -233,7 +232,7 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
             casttype = "uint8_t"
         else:
             casttype = node.func.id
-        code = "({0})({1})".format(
+        code = "({})({})".format(
             casttype, traverse(node.args[0], args, copy.copy(forvars), declared)
         )
```
henryiii, line 237: Something like this is not covered to an f-string by pyupgrade - it only does simple strings that are clearly readable as f-strings. Personally, I'd rewrite it this way:

```python
traversed = traverse(node.args[0], args, copy.copy(forvars), declared)
code = f"({casttype})({traversed})"
```

dev/generate-cuda.py
```diff
@@ -170,23 +169,23 @@ def traverse(node, args={}, forvars=[], declared=[]):  # noqa: B006
...
-            code = "!{0}".format(
+            code = "!{}".format(
                 traverse(node.operand, args, copy.copy(forvars), declared)
             )
         else:
             raise Exception(
-                "Unhandled UnaryOp node {0}. Please inform the developers.".format(
+                "Unhandled UnaryOp node {}. Please inform the developers.".format(
                     node.op.__class__.__name__
                 )
```
henryiii, line 190: This would have been just fine as an f-string IMO, but pyupgrade is conservative about conversions.

henryiii: Also, Exceptions should always be more specific - if not using a custom one, at least RuntimeError. :)

I also try to always use `msg = <stuff>`, then `raise Exception(msg)`, otherwise your message gets repeated twice, once in the line print and once in the message printout.

jpivarski: I was going to say that I only ever use "Exception" in debugging (just like "print", it's easy to search for when cleaning up). This was not written by me. The three exceptions look like they ought to be "NotImplementedError".

However, dev/generate-cuda.py is also on the chopping block for code that will be removed soon. After @swishdiff finalizes the form of the CUDA source code that dev/generate-cuda.py creates, we'll check in that source code and remove the generating script. Henceforth, the source code would be maintained by hand. (This is just seeding many but not all of the initial implementations to save typing for the easy ones.)

henryiii: > it's easy to search for when cleaning up

That's a nice idea. Not worried about changing these, just noting them in passing.

jpivarski:
From what I see, these are a lot of good updates. I like what it has done with the format strings, but they could be hard to merge with other PRs that are currently in development. From my scan, it looks like the non-format string changes would be easy to merge, so maybe non-format string and format string updates can be two PRs?

**These are the currently outstanding PRs:**

   * #1242 (@ioanaif): mostly touches test files, which are not heavily affected by the style upgrade. I think this one would be compatible.
   * #1240 (@jpivarski): mostly creates new files (well, filling in placeholders). This would be compatible, too—if this were in main and I pulled that into my PR, the update to .pre-commit-config.yaml would beautify the code that I'm working on, and that would be fine.
   * #1234 (@ioanaif): requested a review from me, so this one should probably merge into main _first_, before the style updates (and then this branch should bring in those changes and beautify them).
   * #1214 (@agoose77): this one was done, but failed an auto merge. It was a MacOS timeout—I bet it was related to the segfault. I'm updating that to main to fix the segfault and trying again.
   * #1207 (@ianna): this is almost entirely C++, so it's safe.
   * #1200 (@ianna): also entirely C++.
   *#1040 (@agoose77 and me): this only touches a couple of files. I don't think it would be too hard to merge.
   * #1039 (@agoose77 and me): this was an old proposal that got forgotten. @agoose77 and I should reconsider whether we want it, and if so, reimplement it starting from a more recent base. If not, it should be closed. (Discussion should be on #1039, not here.)
   * #990 (@agoose77 and me): same as above.

So after going through all the open PRs, it looks like we _will_ be able to do this style upgrade in one PR, **though the "almost there" and "should we drop it?" questions need to be dealt with first.**

Getting back to this PR, here's a list of the things that I noticed were changed:

   * `from __future__ import absolute_import` removed from every file, which totally makes sense and should happen all at once.
   * Inheritance from `object` is now implicit, naturally. Also easy to merge because it doesn't affect that many lines.
   * `open` arguments, such as `"r"` looks like a good thing.
   * `super` arguments; it will be easy to get used to this style. I always had to look up the argument order in the past. Now there aren't any!
   * `yield from` is an old Python feature that I just hadn't known about until recently (late last year). It's good to do all of those consistently.
   * `set` constructors with curly brackets make me a little nervous because it's hard to see that they're not dicts, at a glance. But okay.
   * dict comprehensions were something that I avoided because the syntax would absolutely break Python 2.6 (as opposed to just being a runtime error). But we shouldn't even be thinking about such things anymore.
   * The use of `"{0} {1} {2}"` in format strings instead of `"{} {} {}"` was also for Python 2.6 friendliness. It was because I had to go through a huge codebase once, turning all the `"{} {} {}"` into `"{0} {1} {2}"` to make it run on some system, and that was such a painful memory. But again, we shouldn't even be thinking about such things anymore.
   * The format strings, even though the changes are conservative, are much better, particularly when they make code fit onto one line. Even though it affects a huge number of lines, it's a very positive change and worth doing.

Some things that didn't get _automatically_ changed:

   * I didn't see all of the `try ... except ImportError` blocks get simplified to just the `try` clause, but all of them are there to handle changed module or class names in Python 3. The `try` block (the spelling that is tried first) is the Python 3 version. These would be easy to fix by hand.
   * The format strings that require `repr`, rather than `str` conversion of an object should all be expressed with the `!r` syntax instead of explicitly using the `repr` function. These will be easy to find and fix, too (assuming `!r` is Python 3.6 compatible, but I bet it is). I only used the `repr` function because I couldn't remember the `!r` syntax (the fact that it's an exclamation point), but we can just memorize that and use it from here on out.

dev/generate-cuda.py
jpivarski: I agree, though that's a very case-by-case thing.

henryiii: I think all of them could be written as f-string one way or the other, though (but not in an automated way, yes). In a recent project (Particle, I think), I ended up converting everything - only "format later" cases remain as `.format`, which is the main use for it now. If you assign `something=<long expression>` inside `.format`, then it's always just as readable or more readable to just use normal Python assignment before the f-string.

The call for whether `f(x)` or `x.thing.otherthing` should be pre-assigned or inlined is likely preference, though - probably based on character length + number of brackets required. Pyupgrade just doesn't convert it for you at all if it's not simple.

henryiii:
I can split it up if you'd like - pyupgrade `--py3-plus` will skip the f-strings, then we can do a second pass with `--py36-plus` to get the f-strings.

Pyupgrade has a "garbage-in, garbage-out" policy (yes, I've asked Anthony about at least the repr cleanup). `try`/`except`'s should have been `sys.version_info` comparisons - these tell both the tool and the reader exactly _why_ this is being done, which side is new/old, and are also understood by mypy.

Same is true for `"{x}".format(x=repr(x))`, this should have been `"{x:!r}".format(x=x)`. This is really a cleanup, not a python upgrade, since the first version of `.format` supported this (in Python 2.6). It's just more noticeable with f-strings, since extra brackets are uglier inline. I _think_ I have a sed expression somewhere that cleans these up that I used in something like Particle.

It is completely trivial to rerun this branch after PRs get merged, as it's entirely automated (except for the change to .pre-commit-config.yaml, of course). I can break it into commits to make it easier to see the non-f-string changes, and possibly apply them either as two PRs or as a rebase and merge PR when ready.

jpivarski:
Let's rerun it (with f-strings and everything—same settings that it has now) after #1214 and #1234 merge into main. The other outstanding PRs are

   * easy to adjust after this goes into main (#1242, #1240, #1040), or
   * completely transparent to it because they are editing C++ (#1207, #1200), or
   * should probably be closed, anyway (#1039, #990).

As for the things that didn't automatically convert, they'll be easy to do by hand, too. I just mentioned them because it's what I noticed while scanning over the changes.

henryiii:
Okay, ping me here when that happens. :)

jpivarski:
It has happened! The little icons next to the bullet points are useful:

image

However, I should resolve this conflict before you start that.

jpivarski:
Hopefully, I haven't introduced any errors in that merge conflict resolution.

Or maybe it doesn't matter? If you pull from main and re-run from scratch... Yeah, I didn't need to resolve that by hand.

Anyway, @henryiii, it's all yours.

henryiii:
Every change is automatically generated, conflicts don't matter, I can just rerun. I can do this in about an hour (in mtg).

henryiii:
@jpivarski The ball in your court now. :) I have a followup ready that does the (somewhat) by-hand edits that I would normally do.

jpivarski:
Although this happened in the European evening on a Friday, I know that @ioanaif and @agoose77 have been active and have seen this discussion. Also our PRs (mine included) don't look like they'll have difficulty merging with it. I don't know whether @ianna was online and I wouldn't assume so for this time of day, but all of her PRs are completely independent because they only modify C++ files.

I didn't check with @swishdiff, for whom it's even later, but he's working on CUDA compilation and although there are some Python files involved in that, they are developer scripts that never had Python 2 as a constraint. (The dev/generate-cuda.py file only has format-string updates, nothing else. If he's working on a copy of that file offline, he could overwrite the git copy with his own, then let pre-commit apply the formats to his copy—if it's even an issue.)

I don't think this is going to be a problem for anybody, so I'm going to merge it.

jpivarski:
Just to keep an eye on everything, to catch any conceivable problems: https://github.com/scikit-hep/awkward-1.0/actions/runs/1731094339

henryiii:
If anyone needs help rebasing/merging with this, please let me know!

PR #1246, opened 2022-01-21
style: further cleanup for Python 3.6+

henryiii:
Followup to #1244 with some simi-manual and manual changes.

- chore: drop sys.version_info checks
- chore: drop py27,py35,py36 checks
- chore: clean up things that mention 'ython 2/3'
- chore: clean up ImportErrors
- chore: replace repr with !r
- chore: remove `__div__`

I'd recommend always using the idiom `sys.version_info >= (3, 8)` or similar - never define a `py36` constant, it's less readable and far less automatable / statically typeable. Especially since there was one for >= 3.6 and one for <= 3.6! :)

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -269,15 +269,15 @@ Templating is only used for integer specialization.
 
 We target Python 3.6 and above. Import statements can assume Python 3 names, string-checking can assume Python 3 meanings of `str` and `bytes`, Unicode literals don't need to be prefixed by `u`, and dict order can be assumed to be stable. Python's f-strings can now be used, but not with equals signs (e.g. `f"{something = }"` rather than `f"something = {something}"`) because that's a Python 3.8 feature (its main use is in debugging, anyway).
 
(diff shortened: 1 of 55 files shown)


jpivarski (approved):
This is great; I would accept it as-is. Respond with a short comment (so I get an email) if you're done and I'll merge it.

I need to get used to using `sys.version_info >= (3, 8)`; it doesn't _look_ like something that's easy to automate around, since it has a lot of variations, such as `(3, 8) <= verinf[:2]` with `import version_info from sys as verinf`. But if expressions like the one you gave are The Standard, then that's what we should use. (The reason for `py38` and friends was precisely so that they could be swapped and searched easily, but it was a convention that didn't extend beyond this codebase.)

henryiii:
Should be done. I'll make a new PR if I think of anything else to check for.

There shouldn't be any variations; `sys.version_info >= (3, 8)` is The Standard, as used in python/typeshed, like here: https://github.com/python/typeshed/blob/fabd8428572dcc20b7ae7919e16d9b4c87cf3da2/stdlib/__future__.pyi#L19 

Using an AST, it's possible to include some of the variations - but I'd not use them, just `sys.version_info`, easier to quickly search.

PR #1248, opened 2022-01-24
Fix: support mixed array types in `NumpyLike.to_rectilinear`

### LINKED ISSUE #1247: `NumpyLike.to_rectilinear` fails for NumPy arrays
### Version of Awkward Array

1.8.3rc0

This bug is most easily triggered when invoking a NumPy function with a rectilinear Awkward argument and NumPy rectilinear argument, e.g. `np.isin`:
```pycon
>>> reference = np.r_[1,2,3,4]
>>> test = ak.Array([1,2,9,0])
>>> np.isin(test, reference)
```

which gives the following traceback:

```pytb
  File "<__array_function__ internals>", line 5, in isin
  File "/home/angus/Git/awkward-1.0/awkward/highlevel.py", line 1440, in __array_function__
    return ak._connect._numpy.array_function(func, types, args, kwargs)
  File "/home/angus/Git/awkward-1.0/awkward/_connect/_numpy.py", line 40, in array_function
    args = tuple(_to_rectilinear(x) for x in args)
  File "/home/angus/Git/awkward-1.0/awkward/_connect/_numpy.py", line 40, in <genexpr>
    args = tuple(_to_rectilinear(x) for x in args)
  File "/home/angus/Git/awkward-1.0/awkward/_connect/_numpy.py", line 32, in _to_rectilinear
    return nplike.to_rectilinear(arg, allow_missing=False)
  File "/home/angus/Git/awkward-1.0/awkward/nplike.py", line 422, in to_rectilinear
    return [self.to_rectilinear(x, *args, **kwargs) for x in array]
  File "/home/angus/Git/awkward-1.0/awkward/nplike.py", line 422, in <listcomp>
    return [self.to_rectilinear(x, *args, **kwargs) for x in array]
  File "/home/angus/Git/awkward-1.0/awkward/nplike.py", line 425, in to_rectilinear
    raise TypeError("to_rectilinear argument must be iterable")
TypeError: to_rectilinear argument must be iterable
```

I think we need to handle the case where an array is `np.ndarray`.

agoose77:
Fixes #1247

### DIFF
--- a/src/awkward/nplike.py
+++ b/src/awkward/nplike.py
@@ -403,6 +403,7 @@ class Numpy(NumpyLike):
         if isinstance(
             array,
             (
+                numpy.ndarray,
                 ak.Array,
                 ak.Record,
                 ak.ArrayBuilder,
--- a/src/awkward/nplike.py
+++ b/src/awkward/nplike.py
@@ -400,10 +400,12 @@ class NumpyKernel:
 
 class Numpy(NumpyLike):
     def to_rectilinear(self, array, *args, **kwargs):
-        if isinstance(
+        if isinstance(array, numpy.ndarray):
+            return array
+
+        elif isinstance(
             array,
             (
-                numpy.ndarray,
                 ak.Array,
                 ak.Record,
                 ak.ArrayBuilder,
--- a/tests/test_1247-numpy-to_rectilinear-ndarray.py
+++ b/tests/test_1247-numpy-to_rectilinear-ndarray.py
@@ -6,6 +6,8 @@ import awkward as ak  # noqa: F401
 
 
 def test():
-    reference = np.r_[1, 2, 3, 4]
-    test = ak.Array([1, 2, 9, 0])
-    assert ak.to_list(np.isin(test, reference)) == [True, True, False, False]
+    array = np.array([1, 2, 9, 0])
+    nplike = ak.nplike.of(array)
+
+    ak_array = ak.from_numpy(array)
+    assert ak.to_list(nplike.to_rectilinear(array)) == ak.to_list(ak_array)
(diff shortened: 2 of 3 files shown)


src/awkward/nplike.py
```diff
@@ -406,6 +406,7 @@ def to_rectilinear(self, array, *args, **kwargs):
         if isinstance(
             array,
             (
+                numpy.ndarray,
```
jpivarski: Maybe this should be in its own `if` statement? The implementation would be a pass-through, rather than going through `to_numpy`. This works, but it might be more comprehensible, from a maintenance perspective, if it's singled out as a distinct case from the Awkward classes.

jpivarski (approved):
You're right about the issue and this is the right way to address it. The one comment below is stylistic.

You're also pulling in `pybind11` again. (Personally, I do a new `git clone --recursive` when a submodule version changes, which is not very often.) I've written a recipe for fixing a PR branch here: https://github.com/scikit-hep/awkward-1.0/pull/1207#issuecomment-1018670709.

src/awkward/nplike.py
agoose77: This is a good suggestion!

agoose77:
> You're right about the issue and this is the right way to address it. The one comment below is stylistic.
> 
> You're also pulling in `pybind11` again. (Personally, I do a new `git clone --recursive` when a submodule version changes, which is not very often.) I've written a recipe for fixing a PR branch here: [#1207 (comment)](https://github.com/scikit-hep/awkward-1.0/pull/1207#issuecomment-1018670709).

Ugh, I didn't see this in my changelist, yikes. Will fix

PR #1249, opened 2022-01-24
Fix: support nested option types in `ak.is_none`

### LINKED ISSUE #1193: `ak.is_none` does not descend into option contents
### Version of Awkward Array

1.7.0

Currently, `ak.is_none` returns during the result building stage once it encounters an option type. This produces some interesting results when `ak.is_none` over nested option types:
  ```python3
  >>> index_of_index = ak.Array(
  ...     ak.layout.IndexedOptionArray64(
  ...         ak.layout.Index64(np.r_[0, 1, 2, 3]),
  ...         ak.layout.IndexedOptionArray64(
  ...             ak.layout.Index64(np.r_[0, -1, 2, 3]),
  ...             ak.layout.NumpyArray(np.r_[1, 2, 3, 4]),
  ...         ),
  ...     )
  ... )
  >>> index_of_index
  <Array [1, None, 5, 6] type='4 * ?union[?int64, ?int64]'>
  >>> ak.is_none(index_of_index)
  <Array [False, False, False, False] type='4 * bool'>
  ```

In these cases, it's visible in the type string that there is something funky going on with the layout, but the correct result would be a `NumpyArray` layout of union of bytemask arrays

It might be that the simplest solution here is to use the `transform_child_layouts` utility, and return the union of the option-layout bytemask and the result from `transform_child_layouts`, e.g.
```python3
def is_none(array, axis=0, highlevel=True, behavior=None):
    """
    Args:
        array: Data to check for missing values (None).
        axis (int): The dimension at which this operation is applied. The
            outermost dimension is `0`, followed by `1`, etc., and negative
            values count backward from the innermost: `-1` is the innermost
            dimension, `-2` is the next level up, etc.
            a low-level #ak.layout.Content subclass.
            high-level.

    Returns an array whose value is True where an element of `array` is None;
    False otherwise (at a given `axis` depth).
    """

    def apply(layout, depth, posaxis):
        posaxis = layout.axis_wrap_if_negative(posaxis)
        if posaxis == depth - 1:
            nplike = ak.nplike.of(layout)
            if isinstance(layout, ak._util.optiontypes):
                mask = nplike.asarray(layout.bytemask()).view(np.bool_)

                inner_mask = ak.to_numpy(
                    ak._util.transform_child_layouts(apply, layout, depth, posaxis)
                )
                return ak.layout.NumpyArray(mask | inner_mask)

            if isinstance(
                layout,
                (
                    ak._util.unknowntypes,
                    ak._util.listtypes,
                    ak._util.recordtypes,
                    ak.layout.NumpyArray,
                ),
            ):
                return ak.layout.NumpyArray(nplike.zeros(len(layout), dtype=np.bool_))

        return ak._util.transform_child_layouts(apply, layout, depth, posaxis)

    layout = ak.operations.convert.to_layout(array)

    out = apply(layout, 1, axis)

    return ak._util.maybe_wrap_like(out, array, behavior, highlevel)
```

This feels a bit clunky, but it makes the point :)

Alternatively, we could just call `ak.packed` here, which will ensure that merge-able layouts are merged. This solution won't fix other cases, though, e.g.
```python3
    ak.layout.Index64(np.r_[0, 1, 2, 3]),
    ak.layout.UnionArray8_64(
        ak.layout.Index8(np.r_[0, 0, 1, 1]),
        ak.layout.Index64(np.r_[0, 1, 0, 1]),
        [
                ak.layout.Index64(np.r_[0, -1, 2, 3]),
                ak.layout.NumpyArray(np.r_[1, 2, 3, 4]),
            ),
            ak.layout.NumpyArray(np.r_[4, 5, 6]),
        ],
    ),
)
```

Related to this: at present, multple option layouts grow the type string `??` corresponds to two nested options. Is this what we want? I would expect to see a single logical option (`any(l.optional for l in layouts_at_depth)`).

agoose77:
Fixes #1193
Requires #1279

To implement this feature, I introduce a temporary union. Let me know if there's a more elegant way to do this! Given that, unlike most other layout transformers, we need to modify several layers, I've split the transformation process into two different transforms - one to locate the first layout, another to actually perform the transformation.

I also wondered whether it would be possible to write the new recursive `getfunction`s with an early-return style instead of explicitly handling each case? In the past, the return value had a lot of interpretation modes, so it was more dangerous. With v2, I wonder if we could re-consider this position?

### DIFF
--- a/src/awkward/_v2/operations/structure/ak_is_none.py
+++ b/src/awkward/_v2/operations/structure/ak_is_none.py
@@ -38,9 +38,7 @@ def is_none(array, axis=0, highlevel=True, behavior=None):
                     ak._v2.index.Index8(tag),
                     ak._v2.index.Index64(index),
                     [
-                        layout.content.recursively_apply(
-                            getfunction_inner
-                        ),
+                        layout.content.recursively_apply(getfunction_inner),
                         ak._v2.contents.NumpyArray(
                             nplike.array([True], dtype=np.bool_)
                         ),
@@ -66,7 +64,7 @@ def is_none(array, axis=0, highlevel=True, behavior=None):
         if posaxis != depth - 1:
             return
 
-        return getfunction_inner(layout, depth)
+        return layout.recursively_apply(getfunction_inner)
 
     layout = ak._v2.operations.convert.to_layout(array)
     behavior = ak._v2._util.behavior_of(array, behavior=behavior)
--- a/src/awkward/_v2/operations/structure/ak_is_none.py
+++ b/src/awkward/_v2/operations/structure/ak_is_none.py
@@ -23,8 +23,6 @@ def is_none(array, axis=0, highlevel=True, behavior=None):
     False otherwise (at a given `axis` depth).
     """
 
-    layout = ak._v2.operations.convert.to_layout(array)
-
     # Determine the (potentially nested) bytemask
     def getfunction_inner(layout, depth, **kwargs):
         if not isinstance(layout, ak._v2.contents.Content):
@@ -65,6 +63,7 @@ def is_none(array, axis=0, highlevel=True, behavior=None):
         if depth_context["posaxis"] == depth - 1:
             return layout.recursively_apply(getfunction_inner)
 
+    layout = ak._v2.operations.convert.to_layout(array)
     behavior = ak._v2._util.behavior_of(array, behavior=behavior)
     depth_context = {"posaxis": axis}
     out = layout.recursively_apply(getfunction_outer, depth_context=depth_context)
--- a/src/awkward/_v2/operations/structure/ak_is_none.py
+++ b/src/awkward/_v2/operations/structure/ak_is_none.py
@@ -23,54 +23,49 @@ def is_none(array, axis=0, highlevel=True, behavior=None):
     False otherwise (at a given `axis` depth).
     """
 
+    layout = ak._v2.operations.convert.to_layout(array)
+
+    # Determine the (potentially nested) bytemask
     def getfunction_inner(layout, depth, **kwargs):
-        nplike = ak.nplike.of(layout)
-        if isinstance(layout, ak._v2.contents.Content):
-            if layout.is_OptionType:
-                layout = layout.toIndexedOptionArray64()
-
-                # Convert the option type into a union, using the mask
-                # as a tag.
-                tag = nplike.asarray(layout.bytemask())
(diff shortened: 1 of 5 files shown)


agoose77:
NB - I haven't enabled any tests that currently skip due to missing `is_none` - I noticed that one failed, and this is all the time I have to look at this for now.

jpivarski:
> To implement this feature, I introduce a temporary union. Let me know if there's a more elegant way to do this! Given that, unlike most other layout transformers, we need to modify several layers, I've split the transformation process into two different transforms - one to locate the first layout, another to actually perform the transformation.

Temporary UnionArrays are fine; they're used in a few other cases.

One transformation/action invoking another is also precedented. In fact, if we know that it can go exactly two levels of transforming deep, then I think it's more readable this way than trying to feed it into itself with a different parameter, just because having different states in a finite state machine be laid out as different code on the screen is easier on human brains. "Action 1: locate, action 2: transform" is very clear.

> I also wondered whether it would be possible to write the new recursive `getfunction`s with an early-return style instead of explicitly handling each case? In the past, the return value had a lot of interpretation modes, so it was more dangerous. With v2, I wonder if we could re-consider this position?

Early return is fine. I tend to not use it because of FP influences, but avoiding stuff like this:

https://github.com/scikit-hep/awkward-1.0/blob/6638fc7d6ad082c4441598a803d93f0687ecfba1/src/awkward/_v2/operations/structure/ak_is_none.py#L57-L60

is worthwhile, I agree. On the other hand, for something like this:

https://github.com/scikit-hep/awkward-1.0/blob/6638fc7d6ad082c4441598a803d93f0687ecfba1/src/awkward/_v2/operations/structure/ak_is_none.py#L66-L69

I'd invert the predicate and only explicitly return the non-trivial case:

```python
        if depth_context["posaxis"] == depth - 1:
            return layout.recursively_apply(getfunction_inner)
```

As you noticed, information for the next level is passed through `depth_context` and `lateral_context`, and the returned data is either the transformed Content or None. My intent with that was to make "action" functions be ignore the cases they don't care about (implicitly return None) instead of the `pass_user` stuff, where the user data might end up being None in such a case.

> NB - I haven't enabled any tests that currently skip due to missing `is_none` - I noticed that one failed, and this is all the time I have to look at this for now.

Okay! Let me know when that changes and I'll review it.

agoose77:
The broken test is caused by #1260, I believe

agoose77:
@jpivarski I think this is also good to merge!

jpivarski (approved):
Indeed! Looks great!

PR #1255, opened 2022-01-27
chore: update pytest config, 6.0+

henryiii:


henryiii:
I'm fine with this when you are, I'll play with the numpy import issue later in a new PR. It sounds like some large libraries can trigger this (like PyTorch), so it's probably not our fault.

jpivarski (approved):
I agree. If this adds failing-on-warnings (except the strange NumPy one), that's an improvement for sure!

PR #1259, opened 2022-01-28
Fix: fix `ByteMaskedArray.simplify_optiontype()`

agoose77:
Currently `ByteMaskedArray.simplify_optiontype()` has a few typos that mean it raises an `AttributeError` instead of returning a simplified layout.


jpivarski (approved):
Thanks for fixing this; these are unquestionably bugs.

PR #1262, opened 2022-01-28
chore: initial nox and pylint support

henryiii:
This adds an initial noxfile. No "fly by" contributor docs for it yet, that
will probably come a bit later, after we get editable installs working.

This adds configuration and a nox job for PyLint. I've avoided touching the
code, and instead added suppression with a "TODO" on anything I think we should
work on fixing. I think we could make PRs once in a while removing some of
these suppressions. (Otherwise this becomes too many changes at once). Some are
minor stylistic improvements, sometimes related to Python 3, some are
simplifications, and some are highly bug-prone that probably should be checked
sooner rather than later (marked with `investigate!`).

The syntax for disabling inline is `# pylint: disable=<name>`, or
`disable-next` to put it on the line before, or it can be at the top of a file.
This serves as nice documentation for a user to explain you know you are doing
something a bit naughty.

PyLint also has a bunch of over-the-top checks like `too-many-*` that I'd
expect we would always leave off. This is intentional - it's easier to see and
turn off a check vs. find it and turn it on. I _think_ many of them are in the
"design" group, but I can't see how to tell what is in a group, I only know how
to list all groups (`pylint --list-groups`).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -76,3 +76,88 @@ filterwarnings = [
   "ignore:The NumPy module was reloaded:UserWarning",
 ]
 log_cli_level = "info"
+
+[tool.pylint]
+master.py-version = "3.6"
+master.jobs = "0"
+master.ignore-paths = [
+  "src/awkward/_typeparser/generated_parser.py"
+]
+reports.output-format = "colorized"
+messages_control.enable = [
+  "useless-suppression",
+]
+messages_control.disable = [
+  "abstract-method",  # TODO: check but might need to keep disabled
+  "arguments-differ",  # TODO: some of these can be cleaned up
+  "attribute-defined-outside-init",  # TODO: These should be avoided!
+  "bad-super-call",  # TODO: investigate!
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -76,3 +76,88 @@ filterwarnings = [
...
+  "cell-var-from-loop",  # TODO: investigate and locally disable if needed (unlikely)
+  "chained-comparison", # TODO
+  "consider-iterating-dictionary",  # TODO
+  "consider-merging-isinstance",  # TODO
+  "consider-using-enumerate", # TODO
+  "consider-using-f-string",  # TODO
+  "consider-using-in",  # TODO
+  "consider-using-max-builtin",  # TODO
+  "consider-using-min-builtin",  # TODO
+  "duplicate-code",  # Hahahaha
```
jpivarski, line 105: Hmmm...

jpivarski (approved):
I approve of this; feel free to merge it when you're done with it.

PR #1263, opened 2022-01-28
chore: add dependabot for actions

henryiii:
I noticed that our cibuildwheel action was really old in #1262, which made me
realise nothing was keeping these up to date. Here's a dependabot that will do
that.


jpivarski (approved):
:+1:

PR #1267, opened 2022-01-29
style: update to first non-pre-release black!

henryiii:
The exponent change is really nice, it also seems a bit more compact for really long asserts and if's.


jpivarski (approved):
I've always thought that `**` should be an exception to the "spaces around binary operators" rule.

All the other changes are more compact and look more readable, too. This is definitely an improvement.

It touches a lot of files, 35, but only 8 of them are not tests. Tests generally don't overlap people's active work—I don't think that merging this would be disruptive.

jpivarski:
(I think it's increasing test coverage because the more compact representation means fewer lines of code to test!)

PR #1270, opened 2022-01-31
Add GHA to build CUDA Wheels and update the cuda build script

sw5sh:
This PR adds a GitHub Action to build CUDA Wheels and uploads it as an artifact (I think this wheel will be uploaded to pypi once we have the first release of the kernels?). The build script has been updated to be built for `manylinux` systems and the wheel is not `PureLib`, without unzipping, changing things and zipping the wheel back. Although this has caused the wheel to be built for specific python version (`cp310`) instead of the earlier more general `py3`. 
Since Awkward's CUDA backend infrastructure relies heavily on Cupy. I have restricted `awkward_cuda_kernels` to be built for only the `CUDA` versions that `Cupy` supports. The library name is dependent on the `CUDA` version instead of the awkward version as earlier. `awkward_cuda_kernels-<cuda-version>-cp310-cp310-<platform>.whl`


jpivarski (approved):
This works, but only for Python 3.10. Setting `distclass = BinaryDistribution` solves the

```bash
sed "s/Root-Is-Purelib: true/Root-Is-Purelib: false/" < tmp_WHEEL
```

problem we used to have! There's still the problem that this wheel is labeled as being Python version dependent when it isn't (since the compiled code is accessed through ctypes, not any compiled pybind11 bindings).

We don't want to reintroduce the unzipping, editing metadata by hand, rezipping to label it as Python version independent (that information needs to go into the `Tag` section of the `WHEEL` file inside the zipped whl). I don't know of a solution for that (@henryiii?). In the meantime, we can just extend the matrix over all Python versions: 7 CUDA versions × 5 Python versions = 35 wheel files, which is still just 11.5 MB.

Another request: during ordinary CI testing, let's just do the minimum CUDA version and the maximum CUDA version for _one_ version of Python (it can be 3.10), so only 2 GitHub Action VMs are expended on this. During deployment, all 35 jobs can run, and they should be defined in wheels.yml so that `upload_all` will wait for them and include them in the upload (though they'll be uploaded to the `awkward-cuda-kernels` package name, not the `awkward` package name).

henryiii:
I've got a PR to wheel that would add a utility to do this (https://github.com/pypa/wheel/pull/422), but it's on hold for other changes in wheel. However, you can do it in the setup.py directly (which is probably better for this sort of case): see https://github.com/scikit-build/ninja-python-distributions/pull/85/files

sw5sh:
Thanks for links Henry! I made the changes and this should be ready after the tests pass.

jpivarski (approved):
Great! Just two things before merging:

  * rename "cuda_wheel.yml" to "cuda_build_test.yml" so that we know it's for testing and only testing that the CUDA compiles
  * only the minimum and maximum CUDA version in that one (10.2 and 11.5.1), so that only 2 VMs run in each commit-test
  * maybe, if you want to, attempt to pull one function out of the shared library in test_1270-install-test.py (to verify that it has the symbols you're expecting, and they're visible)

Also, verify that the artifact from https://github.com/scikit-hep/awkward-1.0/actions/runs/1783596775 has all of the wheels that you wanted (all 7) with the right names (no Python dependence).

Once all that's done, you can squash-and-merge!

Issue #579, opened 2020-12-08
Remember to remove unnecessary intermediate arrays from most reducers


jpivarski:
Only `ak.argmin` and `ak.argmax` need _all_ of them; the rest should allow for `Index64(0)` to be passed through instead (without segfaults, obviously).

If we know that it will be applied at the bottom level (i.e. `negaxis == 1`), there are more opportunities to fast-track it. Rather than making tentative `parents` that get expanded at each level, the bottom level can be implemented by just making the `parents` at the bottom level alone.

There are enough tests of the "full" solution that exists now; switching to optimized versions will be well-tested by the existing suite of tests.

This optimization potential has been known for a while; I just need to remember to do it (and have the satisfaction of popping an issue when it's done).

jpivarski:
This was addressed (as much as is possible) in v2.

PR #1274, opened 2022-02-01
Fixing #1266 by reordering nextparents (in v2; v1 is NOT FIXED!).

### LINKED ISSUE #1266: `ak.sum` produces incorrect structure for outer dimension
### Version of Awkward Array

56312ea2

This is seen in both v1 and v2 in master

Given this layout:
```python3
x = ak._v2.Array([
    [
        [
            [1, 2, 3],
            [4, 0, 0]
        ],
        [
            [5, 6, 7],
            [8, 0, 0]
        ]
    ],
    [
        [
            [9, 10, 11],
            [12, 0, 0]
        ],
        [
            [13, 14, 15],
            [16, 0, 0]
        ]
        
    ]
])
```

Calling `ak.sum(x, axis=0)` does not match (in structure) the NumPy result:
```pycon
>>> ak._v2.to_numpy(x).sum(axis=0).tolist()
[[[10, 12, 14], [16, 0, 0]], [[18, 20, 22], [24, 0, 0]]]
>>> ak._v2.sum(x, axis=0).tolist()
[[[10, 12, 14], []], [[16, 0, 0, 18, 20, 22], []]]
```

jpivarski:
@agoose77, just reordering the `parents` to be monotonic fixes the case you found, but it breaks others. I'm going to leave this open until we know what to do about it.

I included @ianna in the branch name, thinking that this PR in v1 might involve a kernel equivalent to `np.argmax`, but it might not.


agoose77:
Yes, clearly this isn't as simple a fix as I had first hoped. Maybe I've just missed something, but I'll take another look at some point :)

agoose77:
This fails on the positional reducers, I believe, because `argsort` isn't guaranteed to preserve the order between equal elements. A quick check with `sorter = np.argsort(nextparents, kind="mergesort")` seems to confirm this.

I'd quite like to try re-writing the kernel(s) in `listoffsetarray` so that we can make use of the guarantee that `parents` is ordered, but that's a bigger task than I have time for right now.

agoose77:
Now the test-failures are just typetracer failures (I think, I haven't run all the tests locally).

jpivarski:
All the tests pass!

Wait—is it the case that the only thing we needed was to sort the `parents` with a stable sort? That's all that was wrong?

agoose77:
> Wait—is it the case that the only thing we needed was to sort the `parents` with a stable sort? That's all that was wrong?

Yep! I was thinking about it whilst trying to redesign the kernel to compute nextparents and nextcarry directly, and realised that there was probably no guarantee of stability in the sort that we were using. It was the fact that it only failed for positional reducers that made me suspicious though.

agoose77:
Note to self - if we do merge this, then we should definitely explain _why_ we are using `mergesort`.

jpivarski:
No kidding—so that somebody doesn't just say, "Oh, here's a faster sorting algorithm; I'll use that" (and maybe doesn't get segfaults because they're not on a Mac...).

This is definitely a faceplam moment. I'm still astonished.

So yes, we should think about merging this. I'm still keeping in mind your initial assessment that it's a band-aid on a system that's grown pretty complicated. I was talking with @ianna earlier today—she has a different solution so we should take a look at that.

Also, all of these intermediate arrays have a noticeable impact on performance (`ak.sum` is significantly slower than `np.sum`), and I've been thinking for a while that we should perhaps have a separate code path for the `axis=-1` case. In #579, I was thinking that some of these arrays could be skipped without a separate code path, but we could skip a lot more of them with one. The downside is that looks like more complexity, having two ways to do something, but that all-important `axis=-1` case would be less difficult to understand if it's handled in isolation. (If, for instance, someone is looking at it with a profiling tool, trying to see what can be trimmed.)

Then we'd be presenting only the `axis != -1` case as something unavoidably complex and slow. As we've learned this week, it takes great effort to understand what it's doing.

agoose77:
> This is definitely a faceplam moment. I'm still astonished.

I don't think this is a facepalm moment! There is a *lot* going on here, and it's not always clear what our invariants are. 

>  (and maybe doesn't get segfaults because they're not on a Mac...).

Thankfully I think even if the sorter were changed, this would only break positional reducer results, rather than any allocation-related intermediary arrays. It just so happens that if the sorter isn't stable, the values in the reduction groups are visited in different orders, so positional reducers produce incorrect results. I never expected to care about sort stability!

I agree about performance. I was thinking about it abstractly that each kernel is doing an allocation and some work, and as we get larger and deeper arrays this will add up. If we can do things in one (or fewer) passes then it means we can claw back some performance. 

RE optimising `axis=-1` reductions. I suspect these are _very_ common. It would be interesting to see how much faster they are than the slow path. My only worry is *more* code to debug (e.g. this reduction bug that only shows up at n>=4), but I am also cognizant of how even a 10% perf gain is substantial. This is certainly more your area of experience in the Awkward domain, so I'm not going to try and push my opinion around!

ianna:
Great! Well done @agoose77 ! Thanks!

jpivarski:
I think we're all happy about this fix. There might be a more efficient solution, but this one is correct. I'll be merging it.

agoose77:
Thanks all! Note that we didn't fix v1 here, but maybe that's another PR.

jpivarski:
I missed that. Now the title is less misleading.

Of course: to do v1, we'd need to run a sorting algorithm in C++. Maybe one of the `argsort` kernels can be used for that.

jpivarski:
But yes, another PR. Not this one.

PR #1275, opened 2022-02-01
style: pylint 1

henryiii:
This hides a false positive (that's now fixed in the main branch of PyLint).

Then many of the `consider-*` checks are addressed. I can point them out in more detail inline in a review later. I didn't orignally realize that `x in {y, z}` and `x in (y, z)` are different - they both get simplified in bytecode, not creating the container (which is more expensive for a set), but the set version checks equality via hashes, and the tuple version checks via `==`. I think I've corrected any I mistakenly used sets on, but worth a once-over.

- style: global-variable-not-assigned false positive
- chore: include consider-style checks


src/awkward/_connect/_autograd.py
```diff
@@ -15,7 +15,8 @@
 def register():
     import autograd
 
-    global NEP13Box
+    # PyLint false positive: https://github.com/PyCQA/pylint/issues/5756
+    global NEP13Box  # pylint: disable=global-variable-not-assigned
```
henryiii, line 19: Will be removable, see https://github.com/PyCQA/pylint/issues/5756

henryiii:
I also put in several TODO's for things that needed some input; I went with matching the old code technically, even if something else might have been meant.

src/awkward/_connect/_numba/arrayview.py
```diff
@@ -40,9 +40,9 @@ def find(x):
             for i, array in enumerate(arrays):
                 if x is array:
                     return i
-            else:
```
henryiii, line 43: `for` loops do not need an `else` unless there's a `break` in them. It makes the code harder to read and makes the user thing something special is happening. This goes really well with the guard-style (no `else` after `return`/`break`/`raise`), but that's a bigger change for later.

src/awkward/_connect/_autograd.py
jpivarski: This autograd module will not survive into version 2.0. JAX can be seen as an update of autograd, and @swishdiff will be integrating Awkward with that instead. (I originally went for autograd because its pure-Pythonness made it an easier target.)

src/awkward/_util.py
```diff
@@ -521,8 +521,7 @@ def maybe_wrap_like(content, array, behavior, highlevel):
 
 def extra(args, kwargs, defaults):
     out = []
-    for i in range(len(defaults)):
-        name, default = defaults[i]
+    for i, (name, default) in enumerate(defaults):
```
henryiii, line 524: I mostly used enumerate unless there was an explicit reason to assign via `[I]`.

src/awkward/_v2/_reducers.py
```diff
@@ -11,26 +11,11 @@ class Reducer:
...
-        ):
-            if ak._util.win or ak._util.bits32:
-                return np.uint32
-            else:
-                return np.uint64
+        if given_dtype in (np.bool_, np.int8, np.int16, np.int32):
+            return np.int32 if ak._util.win or ak._util.bits32 else np.int64
+
+        if given_dtype in (np.uint8, np.uint16, np.uint32):
+            return np.uint32 if ak._util.win or ak._util.bits32 else np.uint64
```
henryiii, line 18: :D

src/awkward/_v2/_reducers.py
```diff
@@ -544,16 +529,16 @@ def __del__(self):
...
-                type == np.int8
-                or type == np.int16
-                or type == np.int32
-                or type == np.int64
-                or type == np.uint8
-                or type == np.uint16
-                or type == np.uint32
-                or type == np.uint64
-            ):
+            if type in {
```
henryiii: Oops, `{}` should be `()`. Forget when grepping that this could be on the next line.

src/awkward/_connect/_jax/jax_utils.py
```diff
@@ -76,9 +76,9 @@ def find_nparray_node(
 
         def recurse(outarray, recurse_where):
 
-            if isinstance(recurse_where, numbers.Integral) or isinstance(
-                recurse_where, str
-            ):
+            # TODO: capture makes this really confusing. A new parameter should
+            # be used instead if capture is really needed.
+            if isinstance(recurse_where, (numbers.Integral, str)):
```
jpivarski, line 81: jax_utils.py is also on the chopping block. It's from the v1 JAX integration attempt, and @swishdiff will be doing this using an entirely different technique.

src/awkward/_v2/_typetracer.py
```diff
@@ -559,7 +559,7 @@ def broadcast_arrays(self, *arrays):
                 shape = [1] * (len(thisshape) - len(shape)) + shape
             elif len(shape) > len(thisshape):
                 thisshape = (1,) * (len(shape) - len(thisshape)) + thisshape
-            for i in range(len(shape)):
+            for i in range(len(shape)):  # pylint: disable=consider-using-enumerate
```
henryiii, line 584: Here's one I left as range(len

src/awkward/_v2/contents/numpyarray.py
```diff
@@ -367,25 +367,28 @@ def mergeable(self, other, mergebool):
...
+                self._data.dtype,
+                other._data.dtype,
+            }:
                 return False
 
-            if self._data.dtype != other._data.dtype and (
-                self._data.dtype == np.timedelta64
-                or other._data.dtype == np.timedelta64
-            ):
+            if not matching_dtype and np.timedelta64 in {
```
henryiii: I'll grep correctly for these. :(

src/awkward/_v2/forms/recordform.py
```diff
@@ -150,20 +150,19 @@ def __eq__(self, other):
                 and _parameters_equal(self._parameters, other._parameters)
             ):
                 if self.is_tuple:
-                    for i in range(len(self._contents)):
-                        if self._contents[i] != other._contents[i]:
+                    # TODO: could the lengths be different?
+                    for self_cont, other_cont in zip(self._contents, other._contents):
```
henryiii: I also tried to notice when zip could be used.

src/awkward/_v2/operations/convert/ak_from_json_schema.py
```diff
@@ -147,7 +147,7 @@ def build_assembly(schema, container, instructions):
         if len(tpe) == 1:
             tpe = tpe[0]
 
-    if tpe == "boolean" or tpe == "integer" or tpe == "number":
+    if tpe in {"boolean", "integer", "number"}:
```
henryiii, line 150: FYI, these are fine as a set, since the hash of a string should be what you are checking for.

jpivarski:
These are generally good improvements (the diffs I've seen so far). In many of them, I'd be on the fence:

   * between a standard for loop filling an array with `append` versus a list comprehension (I make the choice based on the obfuscation that black's line width limit causes),
   * between `for i in range(len(...))` and `enumerate(...)` (I go for the dummy variable `i` if there's a symmetry between the things that are being selected with square brackets),
   * between `x == this or x == that` and `x in (this, that)` (I usually choose equality, especially if the set is small, because I don't want Python building data structures in tight loops; a performance consideration)

But if the linter says, "do this, not that," then sure.

henryiii:
In some languages, it's faster to do the comprehension, because the language knows you are not done filling the loop, and can even sometimes take length hints, while appending has to just keep making new larger arrays. I don't know how true that is in Python, but I tend to highly favor comprehensions if you are making a new structure or changing every element, and can reasonably write one. It also used to be much better for static typing to do the comprehension, because then the checker knows what the type in the list is. Though checkers are getting smart enough to backtrack in many cases.

If a loop has `x[i] = f(x[i])`, _and `x[i]` can change inside the loop_, then this is best. Most of the time, `x[i]` on the right will not change inside the loop; then it's much, much better to use `for i, x_i = enumerate(x)` and `x[i] = f(x_i)`, because you are conveying that fact to the reader. Otherwise, the reader has check every assignment and see if `x[i]` could change (I know, because I had to do this!)

henryiii:
> I usually choose equality, especially if the set is small, because I don't want Python building data structures in tight loops; a performance consideration

Python explicitly optimizes the `in` case with a literal set, list, or tuple on the right. It's both better, and avoids repeating the left hand side. Also, `thing == 'x' or thimg == 'y'` visually takes more effort than `thing in {'x', 'y'}`, you have to read and think about more conditions, and you have to visually check to see if the condition really is over the same variable each time.

Correction: it only optimizes for literals. So there might be a small cost for a tuple of numpy types.

jpivarski:
> Python explicitly optimizes the `in` case with a literal set, list, or tuple on the right.

That does it for me—I'm always going to do that from now on. Of my three bullet points, this one was the least-well justified because it's purely a performance consideration. But since the Python compiler is smart enough to catch that case, I'll use that pattern all the time, now.

For the list comprehensions, they can get unreadable if they're wide enough for black to split onto many lines. It is true that a list comprehension gives more information to Python about the size, though so would

```python
big_list = [None] * known_size
for i, x in enumerate(some_input):
    big_list[i] = f(x)
```

henryiii:
Here's a few examples to illustrate my note above:

```pycon
>>> x = 2

>>> # OR chains
>>> %timeit x == 1 or x == 2 or x == 3
49.1 ns ± 1.34 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> # Compared to a a list of literals
>>> %timeit x in (1, 2, 3)
32.8 ns ± 0.828 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> # Or a set of literals
>>> %timeit x in {1, 2, 3}
27.9 ns ± 0.748 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

>>> # However, if these are not literals, you lose the optimization:
>>> y = 1
>>> %timeit x in {y, 2, 3}
93.7 ns ± 4.15 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
>>> # Sets are expensive, but you gain if it grows - while tuples are cheap but linear
>>> %timeit x in (y, 2, 3)
58.3 ns ± 1.19 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
>>> # In fact, you lose a LOAD_CONST even for the if chain:
>>> %timeit x == y or x == 2 or x == 3
53.9 ns ± 0.775 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)
```

Personally, I'd err toward readability in all but the hottest loops. And those can have a note about being hot loops, and a comment.

henryiii:
Anyway, let me know what you like / don't like and we can can turn off any linting rules you I added that you don't like, etc.

Also please check the TODO's, those are things you likely can quickly answer.

henryiii:
(The _new_ TODO's on code, not the TODOs in the linting list, that's just what I intend to do eventually)

src/awkward/_connect/_jax/jax_utils.py
henryiii: Ahh, this was the main one that was driving me crazy. The capture makes this really hard to follow. And capture causes late binding, so it can be really tricky to correctly put this into a nicer form without knowing if the late binding is actually intended.

src/awkward/_v2/forms/recordform.py
henryiii: Ah, there is a length check above. Is there a reason this isn't just `return self._contents == other._contents`?

src/awkward/_connect/_jax/jax_utils.py
jpivarski: Since this is code that will be replaced soon, the main thing is to learn from its example. We don't need to change it in this PR, but we'll just avoid this pattern in the future. (So I think I can resolve this one.)

src/awkward/_connect/_numba/arrayview.py
jpivarski: I was thinking of the `else` as helpful information for humans reading the code: it's a signal that one can either exit the function in the loop or handle the "no matches" case. (I wouldn't, for instance, use a `for`-`else` patterns for a loop without any nonlocal jumping: `break`, `return`, `raise`.)

It's true that it only makes a semantic difference for `break`, and if that's a reason for not using the `else`, then sure. All of these are weak preferences for me, in which I could have gone either way. If the linter defines a standard, I have no complaints about following it.

src/awkward/_util.py
jpivarski: This one really should be `enumerate`. Agreed.

src/awkward/_v2/_reducers.py
jpivarski: Yeah.

src/awkward/_v2/forms/recordform.py
jpivarski: The `self.is_tuple` case could be `return self._contents == other._contents`. I was just looking over it, trying to find out why it wasn't—the reason is for symmetry with the `not self.is_tuple` case. In that case, you do need to align the keys of `self` with the keys of `other`. Although that case could be

```
return dict(zip(self._fields, self._contents)) == dict(zip(other._fields, other._contents))
```

and there would still be symmetry, yet the whole thing is much simpler.

src/awkward/_v2/operations/convert/ak_from_json_schema.py
jpivarski: True.

src/awkward/_v2/contents/numpyarray.py
```diff
@@ -623,6 +626,8 @@ def _unique(self, negaxis, starts, parents, outlength):
         if negaxis is None:
             contiguous_self = self if self.is_contiguous else self.contiguous()
             flattened_shape = 1
+            # TODO: could this be contiguous_self.shape[i]?
+            # pylint: disable-next=consider-using-enumerate
             for i in range(len(contiguous_self.shape)):
                 flattened_shape = flattened_shape * self.shape[i]
```
jpivarski: It could. `self` and `contiguous_self` are guaranteed to have the same `shape`.

```suggestion
            flattened_shape = 1
            for x in enumerate(contiguous_self.shape):
                flattened_shape = flattened_shape * x
```

src/awkward/_v2/forms/unionform.py
```diff
@@ -102,11 +102,11 @@ def __eq__(self, other):
...
             ):
-                for i in range(len(self._contents)):
-                    if self._contents[i] != other._contents[i]:
+                # TODO: ._contents is a list. Why not just use ==?
+                for self_cont, other_cont in zip(self._contents, other._contents):
+                    if self_cont != other_cont:
                         return False
-                else:
-                    return True
+                return True
```
jpivarski: Yes, for the same reason as in recordform.py.

src/awkward/_connect/_numba/arrayview.py
henryiii: An else after a loop should tell the reader that there's control flow here. Returns are already control flow. The else is meaningless, it has no impact, but a reader has to read the code above and prove that - plus they also have to try to align the else with the for and not with the nested ifs. Plus they have to realize that the else does nothing. Many Python users don't even know it's valid to use an else with a for; there are special cases (with break) where it's really nice and solves a problem with a teachable structure instead of injecting loop control variables, which is why it's there, but it shouldn't be used if there's no break.

Eventually you'll probably see me argue against else's and elif's that don't have an effect in control flow either, but this one is easier due to the fact it's a rare and specialized feature for `for` and `break`.

henryiii: (If there were a reverse check, one that forced else's after loops that always break, it might be different, because then a reader could take semantic meaning from the else. Currently an always breaking for might not have an else, so every usage has to be read carefully. But it's not what the else was for, and it adds indentation and complexity for the reader, so it's not likely to exist unless you write it yourself. Which is not too bad, from what I'm told.)

src/awkward/_v2/contents/numpyarray.py
henryiii: This could be `math.prod(contiguous_self.shape)` in Python 3.8+. Is there a reason this isn't `np.prod(contiguous_self.shape)`? (small array in a hot loop?)

henryiii:
I've addressed (most ?) comments in 250ba00.

src/awkward/_v2/contents/numpyarray.py
jpivarski: `np.prod` could do it, but in this context, we want to avoid confusion between this small-scale tuple arithmetic and arrays that are dealt with via nplike. This is one that would have to happen through actual NumPy, not CuPy or whatever, but then it would stand out as a possible mistake of using NumPy instead of nplike.

There are other, similar, situations that aren't as simple as a straight product, too. There's some value for this one to look like the others.

henryiii: It being a (small) tuple vs. an array would be enough for me to avoid `np.prod`. Though `math.prod` would be nice, eventually. There's value in simple things looking and reading simply, with special cases standing out. For example, when I read it, I wasn't sure if it could be reduced to a simple product.

henryiii:
PS: I'm quite happy to revert or change anything you want changed, after I make my point, anyway. ;)

jpivarski:
I think we discussed everything here. I'd better merge it before it gets too out of date (already a little hard to merge).

jpivarski:
If `pyarrow` is installed, you get a lot of warnings about objects not existing in `pyarrow.lib` (though they do). However, these are green warnings, not yellow warnings, and so they aren't blockers. It's just a lot of noise when running it locally.

henryiii:
Hmm, that sounds like a compiled extension. There's supposed to be a way to have load the compiled extension and detect the contents, but I haven't been able to get it working properly.

Generally, I'd recommend always running pylint via nox; nox will create and control the environment so it is consistent.

PR #1276, opened 2022-02-03
This PR adds support to call kernels in CUDA from v2 Awkward Arrays.

sw5sh:
Tasks to complete:

- [x] Implement `copy_to` for all array classes
- [x] Implement `to_backend` and `backend(*arrays)`
- [x] Write a minimal test using `ak.num`
- [x] Call a kernel whose implementation is not in CUDA and see if it's being handled correctly.
~~- [ ] Don't upload artifacts during CI testing.~~


sw5sh:
The functions `_getitem_next*`(many `_getitem_next_jagged`) are passing mixed `nplike` because I think they are passing slices and slices are `Numpy nplike` through and through, but when the `typetracers` tests come in, these two `nplikes` get mixed and hence they raise the assertion errors introduced in Wednesday's refactoring. 

For now, I have commented out the assertions and they should be easy to find by doing a ``fgrep -rl "# assert" src``. 

Also, now `to` has been replaced by `raw` and `ak.to_cupy` and `ak.from_cupy` functions have also have been implemented.

src/awkward/_cuda_kernels.py
```diff
@@ -0,0 +1,33 @@
...
+    except ModuleNotFoundError:
+        raise ModuleNotFoundError(
+            """to use CUDA awkward arrays in Python, install the 'awkward[cuda]' package with:
+    pip install awkward[cuda] --upgrade
+or
+    conda install awkward[cuda]"""
+        ) from None
+
+lib = ctypes.cdll.LoadLibrary(awkward_cuda_kernels.shared_library_path)
+kernel = awkward._kernel_signatures_cuda.by_signature(lib)
```
jpivarski: So if this is on Linux, it will raise a ModuleNotFoundError upon import? Oh, I see: unlike `ak._cpu_kernels`, this one is not imported preemptively.

I'm not sure that `conda install awkward[cuda]` is a thing. That recommendation may need to be

```
conda install -c conda-forge awkward-cuda-kernels
```

Also, we probably shouldn't assume that users already have conda-forge added as a channel (though they should!). Hence the `-c` option, just in case.

src/awkward/_v2/_util.py
```diff
@@ -29,6 +29,8 @@
 kSliceNone = kMaxInt64 + 1  # for Slice::none()
 kMaxLevels = 48
 
+backend = {ak.nplike.Numpy: "cpu", ak.nplike.Cupy: "cuda"}
```
jpivarski: This lookup table is going the opposite way of what you want: users give you a string and you need to return the singleton. Also, if the lookup dict (data) is defined here, the function that does the lookup (methods) should be here, too. That's the spirit of OOP, even if we're not going to the trouble of making actual objects.

So

```python
_backends = {
    "cpu": ak.nplike.Numpy,
    "cuda": ak.nplike.Cupy,
}

def regularize_backend(backend):
    if backend in _backends:
        return _backends[backend].instance()
    else:
        complain!
```

I realized while writing this that calling `ak.nplike.Cupy.instance()` might try to import `awkward_cuda_kernels`, so it should be delayed until the user is actually asking for it.

src/awkward/_v2/contents/listarray.py
```diff
@@ -264,6 +269,13 @@ def _getitem_next_jagged(self, slicestarts, slicestops, slicecontent, tail):
             outoffsets = ak._v2.index.Index64.empty(
                 slicestarts.length + 1, self._nplike
             )
+            # assert (
+            #     outoffsets.nplike is self._nplike
+            #     and slicestarts.nplike is self._nplike
+            #     and slicestops.nplike is self._nplike
+            #     and self._starts.nplike is self._nplike
+            #     and self._stops.nplike is self._nplike
+            # )
```
jpivarski: Some of these are commented out (here and following). What's happening?

Is it because the `slicer` in

```python
array[slicer]
```

might not have the same nplike as the `array`? If that's a possibility, then the nplikes that differ should be converted (`_to_backend`), rather than let the difference be ignored. They're about to be used in the kernel that follows, and their `.data` would point to the wrong place (CPU vs GPU).

src/awkward/_v2/_util.py
jpivarski: I just realized that if you're using this internally, you'll have an `nplike` singleton, rather than a string naming it. Somewhere, you'll want a convenience function that passes through that `nplike`—maybe here, maybe not.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -321,16 +329,21 @@ def _getitem_next(self, head, tail, advanced):
             nexthead, nexttail = ak._v2._slicing.headtail(tail)
             nextcarry = ak._v2.index.Index64.empty(lenstarts, self._nplike)
 
+            # assert (
+            #     nextcarry.nplike is self._nplike
+            #     and starts.nplike is self._nplike
+            #     and stops.nplike is self._nplike
+            # )
```
jpivarski: More commented-out assertions, like in ListArray.

src/awkward/_v2/contents/content.py
```diff
@@ -1299,6 +1312,18 @@ def flatten(self, axis=1, depth=0):
...
+        if backend == ak._v2._util.backend[ak.nplike.Numpy]:
+            if isinstance(self.nplike, ak.nplike.Numpy):
+                return self
+            return self._to_backend(ak.nplike.Numpy.instance())
+        elif backend == ak._v2._util.backend[ak.nplike.Cupy]:
+            if isinstance(self.nplike, ak.nplike.Cupy):
+                return self
+            return self._to_backend(ak.nplike.Cupy.instance())
+        else:
+            raise ValueError("Possible values for backends are `cpu` or `cuda`.")
```
jpivarski: What if the `Content.to_backend` (public) takes `backend` as a string, then converts it into an `nplike` singleton using the function in _util.py, and then all of the `_to_backend` (private) methods in each Content subclass expects an `nplike` singleton? That way, you'd be able to use the `_to_backend` methods internally when you need to make sure that a `slicer` from

```python
array[slicer]
```

has the same nplike as the `array`. Also, if `_to_backend` is taking an nplike singleton as input, maybe it should be called `_to_nplike`. That defines the layers: "backends" (strings) come from users into the public API, and "nplikes" (singleton objects) are passed around internally.

src/awkward/_v2/contents/regulararray.py
```diff
@@ -392,14 +398,18 @@ def _getitem_next(self, head, tail, advanced):
                 nextadvanced = ak._v2.index.Index64.empty(
                     self._length * nextsize, self._nplike
                 )
+                # assert (
+                #     nextadvanced.nplike is self._nplike
+                #     and advanced.nplike is self._nplike
+                # )
```
jpivarski: More commented-out assertions. I'm getting the impression that these are all in the list-like arrays (ListArray, ListOffsetArray, and RegularArray), and that they all have to do with `array[slicer]`.

src/awkward/_v2/index.py
```diff
@@ -165,6 +157,21 @@ def __copy__(self):
...
+        if backend == ak._v2._util.backend[ak.nplike.Numpy]:
+            if isinstance(self.nplike, ak.nplike.Numpy):
+                return self
+            return self._to_backend(ak.nplike.Numpy.instance())
+        elif backend == ak._v2._util.backend[ak.nplike.Cupy]:
+            if isinstance(self.nplike, ak.nplike.Cupy):
+                return self
+            return self._to_backend(ak.nplike.Cupy.instance())
+        else:
+            raise ValueError("Possible values for backends are `cpu` or `cuda`.")
```
jpivarski: This, and in Content, is where some code duplication can be reduced by having the function in _util.py handle all the different nplike cases. You want to isolate that in one place so that when we need to add more nplikes to the list, we don't forget it in one of these other places—only one place to fix when it needs fixing.

src/awkward/nplike.py
```diff
@@ -13,26 +13,31 @@
 
 
 def of(*arrays):
-    libs = set()
+    backends = set()
```
jpivarski: Good!

src/awkward/nplike.py
```diff
@@ -458,13 +468,40 @@ def char(self):
...
+            )
+
 
 class Cupy(NumpyLike):
     def to_rectilinear(self, array, *args, **kwargs):
         return ak.operations.convert.to_cupy(array, *args, **kwargs)
 
     def __getitem__(self, name_and_types):
-        raise NotImplementedError("no CUDA in v2 yet")
+        import sys
```
jpivarski: Since `sys` is a standard library, there's no reason to not put its import up at the top of the file.

jpivarski (changes requested):
It's fantastic that this is working! On review, I found a few organizational things that, fixing them now, will save us trouble later. See the inline comments.

Thanks!

jpivarski (approved):

PR #1279, opened 2022-02-08
Fix: simplify output in {Byte,Bit}MaskedArray

### LINKED ISSUE #1260: Simplify output of {Bit,Byte}MaskedArray
At the moment, v1 calls `simplify_optiontype` before returning from particular `getitem` methods in these content types, whilst v2 does not. Is this a planned change, or do we need to restore these?

https://github.com/scikit-hep/awkward-1.0/blob/fca80a69b23b2f5b2d952cc03fee722d9a867665/src/awkward/_v2/contents/bytemaskedarray.py#L161-L210

vs

https://github.com/scikit-hep/awkward-1.0/blob/fca80a69b23b2f5b2d952cc03fee722d9a867665/src/libawkward/array/ByteMaskedArray.cpp#L664-L752

agoose77:
Fixes #1260

### DIFF
--- a/src/awkward/_v2/contents/bitmaskedarray.py
+++ b/src/awkward/_v2/contents/bitmaskedarray.py
@@ -271,7 +271,7 @@ class BitMaskedArray(Content):
             self._field_identifier(where),
             None,
             self._nplike,
-        )
+        ).simplify_optiontype()
 
     def _getitem_fields(self, where, only_fields=()):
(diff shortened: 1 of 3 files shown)


agoose77:
@jpivarski I think this is good to go!

jpivarski (approved):
This is super clear and straightforward. Thanks for the fix!

PR #1289, opened 2022-02-15
C++ refactoring: ak.with_parameter

ioanaif:


src/awkward/_v2/contents/content.py
```diff
@@ -1360,3 +1360,10 @@ def to_json(
             complex_real_string,
             complex_imag_string,
         )
+
+    def withparameter(self, key, value):
+        out = copy.copy(self)
+        parameters = copy.copy(self.parameters)
+        parameters[key] = value
+        out._init(self.identifier, parameters, self.nplike)
```
jpivarski: It would be acceptable to access `_parameters` directly from within this class's own methods.

```suggestion
        out._parameters = parameters
```

There's no need to reset the `identifier` and `nplike` to the values they already have. This isn't just a performance thing; I think setting only the `_parameters` would make the intent clearer for maintenance. Future readers of this code might think it's important that `identifier` and `nplike` are getting involved.

Extracting `parameters` from the property (`self.parameters` and not `self._parameters`) is important because `self._parameters` might be `None`, and if so, `self.parameters` first promotes it to `{}`, so you can be sure that what you get is a dict.

For code at this level of abstraction, it would also be acceptable to replace

```python
        parameters = copy.copy(self.parameters)
```

with

```python
        if self._parameters is None:
            parameters = {}
        else:
            parameters = copy.copy(self._parameters)
```

so that `self._parameters` doesn't get changed by looking at it, but I'm only putting in a suggested change for the first one.

src/awkward/_v2/forms/numpyform.py
```diff
@@ -106,7 +106,19 @@ def _type(self, typestrs):
...
+            out._parameters = (
+                {"__array__": self._parameters["__array__"]}
+                if "__array__" in self._parameters.keys()
+                else out._parameters
+            )
+            out._parameters = (
+                {"__record__": self._parameters["__record__"]}
+                if "__record__" in self._parameters.keys()
+                else out._parameters
+            )
```
jpivarski: This probably isn't right. Which test is this motivated by?

The logic of "if parameters contains `__array__` or `__record__`, then only include that parameter, otherwise include all of them" would be very unexpected. The existence of an `__array__` or `__record__` parameter makes the others not show up in the high-level type, and, in this implementation, the existence of a `__record__` parameter would exclude the `__array__` parameter.

That's not what v1 does:

```python
>>> ak.layout.NumpyArray([1, 2, 3], parameters={"__doc__": "hey"}).type({})
int64[parameters={"__doc__": "hey"}]

>>> ak.layout.NumpyArray([1, 2, 3], parameters={"__doc__": "hey", "__array__": "something"}).type({})
int64[parameters={"__array__": "something", "__doc__": "hey"}]
```

Let me know what motivated this and we'll come up with a better algorithm.

jpivarski (changes requested):
Hi! One suggested change (that you can just click on) and another... I don't know what motivated it, but we very likely want a different implementation. See below.

src/awkward/_v2/forms/numpyform.py
ioanaif: The logic was: for type only include `__array__` or `__record__` if present, otherwise out._parameters will stay None for `__array__` and for `__record__` can be None or containing `__array__`

This was motivated by:

https://github.com/scikit-hep/awkward-1.0/issues/459#issuecomment-694941328
So the rules would be,
    * if arrays have different `__array__` or `__record__` parameters, they are not equal;
    * if they otherwise have different parameters, the types can be equal, but merging (concatenation, option-simplify, or union-simplify) removes parameters other than `__array__` and `__record__`._


In **test_0527-fix-unionarray-ufuncs-and-parameters-in-merging**

```

plain_plain = ak._v2.highlevel.Array([0.0, 1.1, 2.2, 3.3, 4.4])
array_plain = ak._v2.operations.structure.with_parameter(plain_plain, "__array__", "zoinks")
plain_isdoc = ak._v2.operations.structure.with_parameter(plain_plain, "__doc__", "This is a zoink.")
array_isdoc = ak._v2.operations.structure.with_parameter(array_plain, "__doc__", "This is a zoink.")

assert ak._v2.operations.describe.type(plain_plain) == ak._v2.operations.describe.type(plain_isdoc)

assert ak._v2.operations.describe.type(plain_isdoc) == ak._v2.operations.describe.type(plain_plain)

assert ak._v2.operations.describe.type( array_plain) == ak._v2.operations.describe.type(array_isdoc)

assert ak._v2.operations.describe.type(array_isdoc) == ak._v2.operations.describe.type(array_plain)
```

jpivarski (approved):

PR #1292, opened 2022-02-16
C++ refactoring: ak.with_field

ioanaif:


src/awkward/_v2/operations/structure/ak_with_field.py
```diff
@@ -1,134 +1,137 @@
...
+    change the array in-place.
+
+    See #ak.Array.__setitem__ and #ak.Record.__setitem__ for a variant that
+    changes the high-level object in-place. (These methods internally use
+    #ak.with_field, so performance is not a factor in choosing one over the
+    other.)
+    """
+
+    # if isinstance(what, (ak._v2.highlevel.Array, ak._v2.highlevel.Record)):
+    #     what_caches = what.caches  # noqa: F841
```
jpivarski: Should just be deleted. `ak._v2.Array` has no caches.

src/awkward/_v2/operations/structure/ak_with_field.py
```diff
@@ -1,134 +1,137 @@
...
+        if not (isinstance(where, str) or where is None):
+            where = where[0]
+
+        behavior = ak._v2._util.behavior_of(base, what, behavior=behavior)
+        base = ak._v2.operations.convert.to_layout(
+            base, allow_record=True, allow_other=False
+        )
+
+        if len(base.fields) < 0:
+            raise ValueError("no tuples or records in array; cannot add a new field")
```
jpivarski: Shouldn't this be `== 0`? The length of a list can never be negative.

jpivarski (approved):
Everything looks good, though it would be better if there was a test of adding a field to a tuple (i.e. the original `fields` property is `None`).

PR #1293, opened 2022-02-17
C++ refactoring: ak.full_like, ak.zeros_like, ak.ones_like

ioanaif:


src/awkward/_v2/operations/structure/ak_full_like.py
```diff
@@ -1,173 +1,167 @@
...
+    """
+    Args:
+        array: Array to use as a model for a replacement that contains only
+            `fill_value`.
+        fill_value: Value to fill new new array with.
+        highlevel (bool, default is True): If True, return an #ak.Array;
+            otherwise, return a low-level #ak.layout.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+        dtype (None or type): Overrides the data type of the result.
```
jpivarski: A user would reasonably expect to be able to pass anything that NumPy would accept as a `dtype` here, including things like `np.int32`, `np.dtype(np.int32)`, or even `"i4"`. Expand the docstring (here and in `ones_like` and `zeros_like`) and promote it in one of the first lines of this function.

```python
if dtype is not None:
    dtype = np.dtype(dtype)
```

The `fill_value` can then be cast to it using `fill_value = dtype.type(fill_value)` if it's a number, but datetime64/timediff64 don't have enough information. I don't know what to do about those cases, but surely there's a way.

jpivarski (approved):
Just one request to loosen the allowed inputs for `dtype`, then (with a test), you can merge it.

jpivarski (approved):
We talked about three changes in Slack,

   * use `fill_value = nplike.array([fill_value], dtype=dtype)[0]` so that dates get the right granularity
   * `dtype == np.dtype(np.bool_)` instead of `dtype is bool`
   * `fill_value.view(np.uint8)` in the boolean case (or should it be `np.int8`? maybe it doesn't matter) so that `fill_value` always has a NumPy type and not a Python type.

When those are done, you can squash and merge!

PR #1296, opened 2022-02-18
C++ refactoring: ak.fields

ioanaif:


jpivarski (approved):
This function's existence made more sense before `fields` became an argument of both `ak.Array` and all the `Content` subclasses. But I see no reason not to keep it.

PR #1303, opened 2022-02-21
Fix: do not increment field index for nested lists

### LINKED ISSUE #1302: `RecordArrayBuilder` does not support nested lists correctly
### Version of Awkward Array

2759ac81

Consider this reproducer:
```python3
import awkward as ak

builder = ak.layout.LayoutBuilder64("""
{
    "form_key": "root",
    "class": "RecordArray",
    "contents": {
        "u": {
            "form_key": "u",
            "class": "ListOffsetArray64",
            "offsets": "i64",
            "content": {
                "form_key": "u-content",
                "class": "RecordArray",
                "contents": {
                    "i": {
                        "class": "NumpyArray",
                        "primitive": "int64",
                        "form_key": "u-content.i"
                    },
                    "j": {                    
                        "class": "ListOffsetArray64",
                        "offsets": "i64",
                        "content": "int64",
                        "form_key": "u-content.j"
                    }
                }
            }
        },
        "v": {
            "class": "NumpyArray",
            "primitive": "float64",
            "form_key": "v"
        },
        "w": {
            "class": "NumpyArray",
            "primitive": "int64",
            "form_key": "w"
        },
        "x": {
            "class": "NumpyArray",
            "primitive": "int64",
            "form_key": "x"
        }
    }
}
""")

builder.begin_list()  # u

builder.int64(1)  # i
builder.begin_list()  # j
builder.int64(2)
builder.end_list()  # j

builder.end_list()  # u

builder.float64(3.0)  # v
builder.int64(4)  # w 
builder.int64(5)  # x

print(ak.Array(builder.snapshot()))

```
This example fails with an error 
```pytb
  File ..., line 36, in <module>
    result.int64(2)
ValueError: ListOffsetArray Builder node-id4-offsets needs begin_list
```
The reason that this happens is that currently the `RecordArrayBuilder` primitive operations (`int64`, `float64`) post-increment the field index. This means that when the user has `RecordArray`s _inside_ lists _inside_ `RecordArray`s, operations that the user expects to apply to the innermost `RecordArrayBuilder` are also modifying the outermost builder. This ultimately leads to obscure error messages.

The control flow is a little tricky to follow, so I'll outline what's happening below:
- The root `RecordArrayBuilder` has `field_index() == 0` after the first `builder.begin_list()` on line 49.
- The root `RecordArrayBuilder` has `field_index() == 1` after the first `builder.int64(1)` on line 51.

This is where the bug happens. The first call to `builder.int64` is only supposed to modify the inner `RecordArrayBuilder`, but it also modifies the root `RecordArrayBuilder`. This is because we always increment the index even when the `RecordArrayBuilder` is inside a list. It follows that the subsequent call to `builder.begin_list()` operates on field 1, which is the `NumpyArrayBuilder` "v", instead of the `ListOffsetArray64` "j".

agoose77:
This would fix #1302 by holding the `field_index_` constant whilst in an `inside-list` state.

### DIFF
--- a/tests/test_1302-layout-builder-nested-list.py
+++ b/tests/test_1302-layout-builder-nested-list.py
@@ -35,7 +35,7 @@ def test():
             },
             "v": {
                 "class": "NumpyArray",
-                "primitive": "float64",
+                "primitive": "int64",
                 "form_key": "v"
             },
             "w": {
@@ -57,30 +57,33 @@ def test():
 
     builder.int64(1)  # i
     builder.begin_list()  # j
-    builder.int64(2)
+    builder.int64(9)
+    builder.int64(8)
+    builder.int64(7)
     builder.end_list()  # j
 
     builder.end_list()  # u
 
-    builder.float64(3.0)  # v
-    builder.int64(4)  # w
-    builder.int64(5)  # x
+    builder.int64(2)  # v
+    builder.int64(3)  # w
+    builder.int64(4)  # x
 
-    # Second pass
     builder.begin_list()  # u
 
     builder.int64(1)  # i
     builder.begin_list()  # j
-    builder.int64(2)
+    builder.int64(9)
+    builder.int64(8)
+    builder.int64(7)
     builder.end_list()  # j
 
     builder.end_list()  # u
 
-    builder.float64(3.0)  # v
-    builder.int64(4)  # w
-    builder.int64(5)  # x
+    builder.int64(2)  # v
+    builder.int64(3)  # w
+    builder.int64(4)  # x
 
     assert ak.to_list(builder.snapshot()) == [
-        {"u": [{"i": 1, "j": [2]}], "v": 3, "w": 4, "x": 5},
-        {"u": [{"i": 1, "j": [2]}], "v": 3, "w": 4, "x": 5},
+        {"u": [{"i": 1, "j": [9, 8, 7]}], "v": 2, "w": 3, "x": 4},
+        {"u": [{"i": 1, "j": [9, 8, 7]}], "v": 2, "w": 3, "x": 4},
     ]
--- a/tests/test_1302-layout-builder-nested-list.py
+++ b/tests/test_1302-layout-builder-nested-list.py
@@ -22,13 +22,13 @@ def test():
                         "i": {
                             "class": "NumpyArray",
                             "primitive": "int64",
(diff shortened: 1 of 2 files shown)


jpivarski:
I was just going to suggest @ianna to review this. But you've got everything all set here, so I'll just leave it to you two!

tests/test_1302-layout-builder-nested-list.py
```diff
@@ -0,0 +1,71 @@
...
+    builder.begin_list()  # j
+    builder.int64(2)
+    builder.end_list()  # j
+
+    builder.end_list()  # u
+
+    builder.float64(3.0)  # v
+    builder.int64(4)  # w
+    builder.int64(5)  # x
+
```
ianna, line 85: @agoose77 -Thanks! Well done at spotting the bug!

Please, extend the test to fill in more then one value:
```python
       builder.begin_list()  # u (0, 0)
    
        builder.int64(1)  # i (1, 1)
        builder.begin_list()  # j
        builder.int64(2)
        builder.end_list()  # j
    
        builder.end_list()  # u
    
        builder.float64(3.0)  # v
        builder.int64(4)  # w
        builder.int64(5)  # x
    
    
        builder.begin_list()  # u (0, 0)
    
        builder.int64(1)  # i (1, 1)
        builder.begin_list()  # j
>       builder.int64(2)
E       ValueError: ListOffsetArray Builder u-content.j-offsets needs begin_list
```

ianna (dismissed):

tests/test_1302-layout-builder-nested-list.py
agoose77: I thought I'd tested this case locally - apparently not! Thanks for spotting this.

agoose77:
OK, looks like it was a bug with the index wrapping.
Here, when `field_index_ == contents_size_ - 1`, both the returned value and the "next index" (`field_index_`) have the same value:
```cpp
return (field_index_ < contents_size_ - 1) ?
      field_index_++ : (field_index_ = 0);
```
I think we just want to do something like
```cpp
auto result = field_index_;
field_index_ = (field_index_ + 1) % contents_size_;
return result;
```
which is done in b31fee7, with a common-case shortcut.

ianna (approved):
@agoose77 - Thanks! It looks good. Please, go ahead and squash&merge it.

PR #1306, opened 2022-02-22
Awkward CUDA Kernel's version should be tied with Awkward

sw5sh:


jpivarski (approved):
Manually tested: running

```bash
./cuda-build.sh -c 11.4.2 -i
```

installs awkward-cuda-kernels with version 1.8.0rc5 (latest `VERSION_INFO` at the time of the test). This package's version number is _not_ 11.4.2, the CUDA version.

PR #1307, opened 2022-02-22
C++ refactoring: ak.argcombinations, ak.combinations

ioanaif:


jpivarski (approved):
Fantastic! I guess `combinations` was already implemented as a layout method, and just needed to be passed through to highlevel (and tested)? It's great to know that it's tested now.

PR #1309, opened 2022-02-23
C++ refactoring: ak.sort

ioanaif:


src/awkward/_v2/operations/structure/ak_sort.py
```diff
@@ -7,35 +7,33 @@
...
+    #     For example,
+
+    #         >>> ak.sort(ak.Array([[7, 5, 7], [], [2], [8, 2]]))
+    #         <Array [[5, 7, 7], [], [2], [2, 8]] type='4 * var * int64'>
+    #"""
+    layout = ak._v2.operations.convert.to_layout(
+        array, allow_record=False, allow_other=False
+    )
+    out = layout.sort(axis, ascending, stable)
+    return ak._v2._util.wrap(out, behavior, highlevel)
```
jpivarski, line 39: This would drop any `behavior` that the original `array` had. That is, if `array.behavior` is not None, then the array that this function returns should have that same `behavior`. It can only be overridden if the user-specified `behavior` is not None.

These rules are codified in utility functions `behavior_of` and `wrap` with the `like` argument. See their [implementation here](https://github.com/scikit-hep/awkward-1.0/blob/982b507094bb568f724991c39c3b528abb8c99eb/src/awkward/_v2/_util.py#L459-L502).

I'm only realizing this now because Coffea tests failed when this was accidentally removed from v1 `ak.unzip`. Here's me putting it back earlier today: #1310. (I explicitly reassigned `behavior`; the same thing could be accomplished in v2 by passing a `like=array` argument to `ak._v2._util.wrap`.) Since we don't test `behavior`-forwarding carefully but Coffea uses it a lot, there may be a lot of v2 functions that are missing this forwarding. I haven't been thinking about it while reviewing these PRs.

Since this is probably an across-the-board problem, let's not worry about it here, but fix it in a sweep across all the `ak._v2.*` functions, all at once. Here's the test that I used for `ak.unzip` (v1):

```python
>>> ak.unzip(ak.Array([{"x": 1}], behavior={}))[0].behavior
{}
```

The `ak.Array` is constructed with a `behavior` that is not equal to None, but is anyway simple (empty dict). The `ak.unzip` is _not_ given a `behavior` argument, so the `behavior` property of the output array must be that empty dict. When the `behavior` is not forwarded, the output is None.

We don't need to test the fancy rules (how an explicit `behavior` overrides the forwarded one) because the most likely mistake is that forwarding was forgotten.

I'm merging this PR because it's fine, apart from the behavior-forwarding, and I'm making a [Project](https://github.com/scikit-hep/awkward-1.0/projects/4) card to check it for all functions.

jpivarski (approved):
Great, except for one thing (which will be dealt with later, after this PR). See below:

Issue #1311, opened 2022-02-23
len of typetracer array raises TypeError (cannot be interpreted as integer)


douglasdavis:
### Version of Awkward Array

1.8.0rc4

### Description and code to reproduce

This looks to be related to #1216 -- the typetracer length issue causing the repr to break is now raising if I try to compute the meta of an object that has had len map partitioned across it:

```python
x = dak.from_json(...)
x.map_partitions(len)
```
will fail because `map_partitions` will try to call `len(x.meta)` to compute a new meta.

```python
In [95]: x
Out[95]: dask.awkward<getitem, npartitions=2>

In [96]: x.meta
Out[96]: <Array-typetracer type='?? * var * int64'>

In [97]: x.compute()
Out[97]: <Array [[1, 2, 3], [1, 2], ..., [3, ..., 4], [1, 2, 3]] type='40 * var * int64'>

In [98]: len(x.meta)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Input In [98], in <module>
----> 1 len(x.meta)

File ~/.pyenv/versions/3.10.2/envs/dask-awkward/lib/python3.10/site-packages/awkward/_v2/highlevel.py:537, in Array.__len__(self)
    527 def __len__(self):
    528     """
    529     The length of this Array, only counting the outermost structure.
    530 
   (...)
    535     is `3`, not `5`.
    536     """
--> 537     return len(self._layout)

TypeError: 'UnknownLengthType' object cannot be interpreted as an integer
```

I traced the `__len__(self)` call for `x.meta` to be rely on the `length` attr of its `layout` (which is an `awkward._v2._typetracer.UnknownLengthType`):
```
In [111]: x.meta._layout.length
Out[111]: UnknownLength
```

jpivarski:
What JSON data did `x` come from and what Awkward operations get applied to do the `x.map_partitions(len)`? I'm trying to reproduce this with Awkward-only, so that I can narrow down to find out what's putting an `UnknownLength` on the typetracer.

douglasdavis:
[...]

PR #1312, opened 2022-02-23
Keep as much length knowledge as possible in typetracers.

jpivarski:
I have no idea whether this addresses #1311, @douglasdavis, but it does maximize the amount of length information available by turning concrete arrays of length `N` into typetracer arrays of length `N`, rather than `UnknownLength`. It will definitely happen that intermediate calculations will create arrays of unknown length, which could legitimately result in the final array having unknown length. This PR should just make that happen less often, in fewer circumstances.

### DIFF
--- a/src/awkward/_v2/_broadcasting.py
+++ b/src/awkward/_v2/_broadcasting.py
@@ -130,7 +130,7 @@ def all_same_offsets(nplike, inputs):
             if x.size == 0:
                 my_offsets = nplike.empty(0, dtype=np.int64)
             else:
-                my_offsets = nplike.arange(0, x.content.length, x.size)
+                my_offsets = nplike.arange(0, x.content.length, x.size, dtype=np.int64)
 
             if offsets is None:
                 offsets = my_offsets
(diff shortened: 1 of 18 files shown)


jpivarski:
Oh, and I'm going to keep this open until you tell me your opinion on it. It has some general cleanup, so I'll want to merge something very much like it.

So let me know!

jpivarski:
To do: add a method to "un-know" the length of a typetracer (both low-level and high-level), so that a length derived from the first partition is not incorrectly imputed to be the length of each partition (or the array as a whole).

jpivarski:
> To do: add a method to "un-know" the length of a typetracer (both low-level and high-level), so that a length derived from the first partition is not incorrectly imputed to be the length of each partition (or the array as a whole).

@martindurant and @douglasdavis, this adds a method so that you can

```python
ak._v2.Array(array.layout.forget_length())
```

to forget a typetracer's length. It only applies one level deep. If need be, we can add a `recursive=True` option, but I assume that this is what you want.

This PR will merge after the tests pass.

martindurant:
Correct, it's only the outer-most length that we are concerned with. Keeping the other ones is probably useful.

PR #1317, opened 2022-02-24
C++ refactoring: ak.cartesian, ak.argcartesian

ioanaif:


src/awkward/_v2/operations/structure/ak_argcartesian.py
```diff
@@ -14,101 +14,93 @@ def argcartesian(
...
+        >>> one[one_index]
+        <Array [1.1, 1.1, 2.2, 2.2, 3.3, 3.3] type='6 * float64'>
+        >>> two[two_index]
+        <Array ['a', 'b', 'a', 'b', 'a', 'b'] type='6 * string'>
+
+    All of the parameters for #ak.cartesian apply equally to #ak.argcartesian,
+    so see the #ak.cartesian documentation for a more complete description.
+    """
+    if axis < 0:
+        raise ValueError("the 'axis' of argcartesian must be non-negative")
```
jpivarski: I see no reason why the `axis` can't be negative. It's like that in v1, also:

https://github.com/scikit-hep/awkward-1.0/blob/5e4e4b984a62baf4075ab96113903f49de058ada/src/awkward/operations/structure.py#L3598-L3602

I'm guessing it's an artifact from a very early time. Anyway, I just committed this fix and added a test, so this PR will relax that constraint.

jpivarski (approved):
Everything here looks great, including the very complicated `cartesian` function.

I noticed that `argcartesian` is excluding negative `axis`; I don't see a reason for that, so I'm removing that restriction and adding a test. Once that passes, I'll merge this.

Thanks!

PR #1322, opened 2022-02-25
chore: wheel not required for setuptools PEP 517 (all-repos)

jpivarski:
@henryiii I'm following the example of https://github.com/scikit-hep/uproot4/pull/565 and the other Scikit-HEP packages. I think this one didn't match your `sed` string because it specifies a version.

If this is the right thing to do for Awkward as well, go ahead and merge this PR.

Closes #1315.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
     "setuptools>=42.0",
-    "wheel>=0.36.0",
     "cmake>=3.13",
     "PyYAML",
 ]


henryiii:
Not sure this is any different than #1315 ?

henryiii:
(And I noticed it wouldn't match, as I stared with an all-repos grep, which is why I did that manually. It's particularly bad to specify versions, as setuptools's PEP 517 hook also does this, which can cause conflicts).

jpivarski:
This is completely identical to #1315; sorry I missed it! You can merge either one of these and close the other.

PR #1326, opened 2022-02-28
Docs: fix typo in documentation

agoose77:
- [x] Drop documentation of `ak.highlevel.ArrayBuilder.extend`
- [x] Drop documentation of `ak.layout.ArrayBuilder.extend`
- [x] Drop documentation of `ak.layout.ArrayBuilder.append`
- [x] Fix typo of `ArrayBuilder`


jpivarski:
Yes, please drop references to the `extend` method, since that's an old one that was _removed_. It used to have a different meaning than just calling `append` many times, now it wouldn't, if it were to exist.

agoose77:
I should have realised that it was just as likely that it was _removed_ as added!

jpivarski (approved):
Thanks! If it's ready to go, you can merge it.

PR #1331, opened 2022-03-02
What a delayed nplike would look like

jpivarski:
@swishdiff This is based on the `ErrorContext` from #1327. Nothing has been tested yet, just trying out ideas.

This asynchronous nplike could be used with any other nplike performing the work, so I'm trying things out with `ak.nplike.Numpy.instance()` as the backend.


agoose77:
Could we use `concurrent.futures.Future` instead of rolling our own? We'd split the Awkward-specific code from the existing `Future` into a `_WorkItem` which holds a reference to the future, and sets the exception / completed status. The future would then just be a standard interface into the executing task.

The benefit of this is a standard `Future` API, which I think is worth supporting despite the fact that this is going to be a fairly internal detail.

jpivarski:
> Could we use `concurrent.futures.Future` instead of rolling our own?

I thought about that (just as I looked into asyncio), and we do use a method of the same name for the same concept, `result()` (so it's duck-type friendly!). However, if we use the Future that is built into Python's Executors, then we get a lot of other concepts along for the ride that we have no intention of implementing, like `cancel`/`cancelled`, `add_done_callback`, `set_running_or_notify_cancel`, ... Using `concurrent.futures.Future` has the advantage of parsimony (not inventing a second thing when a first thing exists) and the disadvantage of having to decide what to do with these other methods (should they all raise exceptions?) and possible bugs because `concurrent.futures.Future` does some things that we don't know about yet.

"Use vs make anew" is always a trade-off, and usually in favor of anew (why we have a nice, modular ecosystem). But I think it should always be reevaluated in each context. In this one, I don't see how the trade-off is in our favor. At the end of stubbing out all of the methods we don't use and understanding `concurrent.futures.Future`'s details well enough to know that it's not going to bite us, we'll be at the same place where we are now, with a simple `Future` we know how to use.

(The trade-off would be different if these were public objects, because it would add into the mix the possibility that users would confuse this `Future` with `concurrent.futures.Future`. If we're worried about developers making that mistake, we could rename our Future. This kind of thing is also called a "[Promise](https://en.wikipedia.org/wiki/Futures_and_promises).")

jpivarski:
This message is mostly for @agoose77's information, since @swishdiff and I have just been talking about it for the last two hours: we're not going to follow this route of background workers tied to CUDA streams. He's been trying to make it work since last week, and there are a lot of hard problems like how to decide which CUDA stream to run an Awkward kernel on: the one attached to the background thread that launched it or the one assigned by [CuPy's context manager](https://docs.cupy.dev/en/stable/reference/generated/cupy.cuda.Stream.html)? Why wouldn't they be the same? Well, our background thread delays Python code execution, so an operation could get executed after CuPy's context manager exits. It's becoming clear that this approach has Awkward Array and CuPy fighting for control, which is not what we want.

Why did we start on the idea of a background thread in the first place? Because every Awkward kernel allocates memory for its output—nothing is in-place. Calling cuMalloc is a blocking operation, so if we did it naively, all of our work on GPUs would effectively be synchronizing every step of the way, which is not a good way to use a parallel device (and there's no point in having a _slow_ CUDA backend). The idea was that, with a background thread, we could do all the CUDA-blocking operations and users could fill up the GPU by running multiple such threads.

However, CuPy needs to do this, too. How did we think CuPy manged it without a background thread? My first thought was that CuPy users are more careful to do in-place operations. While that might be the case, @swishdiff informed me that CuPy also avoids cuMalloc: calling `cp.empty(1000000)` does not call cuMalloc to get a big array; it preallocates for all the arrays it thinks it's going to need and `cp.empty(1000000)` draws from its own allocation. (That is, CuPy has its own, internal "malloc," what's sometimes called an "[arena](https://en.wikipedia.org/wiki/Region-based_memory_management).")

So... CuPy avoids blocking operations if you use its API. We want to avoid blocking operations without reinventing the wheel, and especially without having to make our wheel work with CuPy's wheel. (To extend the analogy: you really don't want to add wheels to a car that are different sizes or rotate at different rates! That's analogous to the confusion that would come if Awkward-CUDA streams were somehow different from CuPy streams.) We can ensure that every time we want a new Index or NumpyArray, we use a CuPy call to get it, which we were planning on doing, anyway. As long as Awkward-CUDA is implemented entirely in terms of CuPy API, code wrapped in `with cp.cuda.Stream():` would behave as users expect, as this would control which streams both CuPy and Awkward use.

The only trouble then is that we have device kernels that CuPy doesn't know about—CuPy wouldn't know that one of our kernels is finished processing and it is allowed to use the output of an array. @swishdiff then introduced me to [cp.RawKernel](https://docs.cupy.dev/en/stable/reference/generated/cupy.RawKernel.html): if we put the CUDA/C++ source code of our kernels into a RawKernel, CuPy compiles it and then treats it as one of its own operations—presumably using the right stream, device, and aware of when it finishes. This would be a way for us to use the CuPy API _exclusively_.

It also solves another problem: we won't need a second Python package named `awkward-cuda-kernels`. CuPy comes with a just-in-time compiler of CUDA code. Source code is much more portable than compiled code. Using CuPy's JIT also eliminates the problem of needing different Python modules compiled for different CUDA versions (9.x, 10.x, 11.x, ...), which is also a problem I'd rather not have. So this approach of doing everything through CuPy, which includes using it for JIT compilation, solves _two_ major problems.

Therefore, I'm very pleased to close this PR and give up on the background thread idea. I'm fully acknowledging how much time @swishdiff spent in trying to make it work, in how much time @agoose77 and I have talked about it, as well as the time I put into it. But in software design, it's good to recognize a sunk cost early. (The whole v1 → v2 transition is a case of recognizing a simplification late.)

--------

Beyond GPUs, @agoose77 did point out that these background threads can be useful for parallelizing on the CPU, a use-case that hadn't occurred to me when I started this direction. But we will also be using dask-awkward to parallelize on the CPU, and I'd rather not have more than one way of doing a thing. The background thread (as I mentioned when first introducing it) is like v1 VirtualArray (in some but not all aspects). VirtualArray-style laziness allows for parallelization to be transparent: users don't know when a computation happens, but if they ask for the result, it will happen before they get the result. Dask-style laziness makes parallelization a little more visible: there's an explicit `compute` method that makes the computation happen. As I said in [my talk at the Dask summit](https://github.com/jpivarski-talks/2021-05-21-dasksummit-awkward-collection/blob/main/degrees-of-transparency-2.pdf), the transparency of VirtualArrays was a liability. Users wanted more control over when (and if) computations happen.

So we'll keep in mind that there's this potential method of transparent parallelization, in case we find that the dask-awkward pattern, with a `compute` method, does not fit all problems. We can always come back to this and put the relevant bits into a new PR.

------

Since this PR also motivated #1327, we might ask if that should be reverted, too. Although #1327 became more urgent when the possibility of background threads was raised, it's still useful to have error messages reported at the granularity of `ak.*` operations and I wouldn't want to go back to not having that feature. And really, most of what #1327 did is it put exception handling in a centralized place so that we can more easily revisit the question of what error messages look like. @henryiii's flake8 check ensures that it's used consistently. These are all good things.

agoose77:
What a journey! To be honest, this sounds like a good conclusion to the problem of CuPy integration. Background threads would be really interesting, but I am sure we all anticipated a new frontier of headaches that would better be avoided if possible. In particular, I anticipate far fewer benefits from local CPU task parallelism than distributed parallelism via Dask, given that Dask also solves the memory problem at the same time. Adding code solely to solve that problem would probably be a long-term footgun.

Solving another problem of packaging is also a big win, and given that we're already planning on jitting with RDataFrame, a little _jit_ more can't hurt :laughing: 

#1327 is absolutely still relevant - it will be useful when/if the other exception PEP is landed to be able to formally move notes into the exception, which with #1327 we can do in a single place. Furthermore, as we often discuss, it's an internal API that we can be fairly blasé about changing down the road.

Issue #1333, opened 2022-03-02
Rejection of invalid layouts.


agoose77:
In #1308 there was a conversation about the policy for accepting redundant (unsimplified) layouts. There have been at least two cases (#713, #1308 discussion) where non-simplified layouts produce strange results when passed to high level functions, and until now I've raised these as bugs.

After the discussion with @jpivarski in #1308 I've realised that this is a policy issue, as Awkward functions themselves do/should not produce such layouts. Accepting redundant layouts is transparent in some functions (like `ak.unflatten` and `ak.packed`), but not those where the high-level function needs to match a particular layout structure. Such functions are made far simpler by the requirement of simplified layouts.

The take-away is:
- Awkward should not produce non-simplified Arrays
- Users of the mid-layer should check that they don't produce non-simplified Arrays

To support external users of the mid layer, I think it would be helpful to catch cases where this invariant fails; bugs in the Awkward code-base will more easily found than those in external users' code (because we will have a greater number of users overall).

agoose77:
We discussed the idea of a per-layout flag that determines whether a layout is in its simplest form. This would add a reasonable amount of code to check something that internally should never happen. 

A less robust approach would be to tackle #516 at the same time and create a mid-level API that resembles `ak._util`, e.g.
```python3
def wrap(layout, behavior):
    assert_simplified(layout)
    return ak._v2._util.wrap(layout, behavior)

...
```

The issue with *this* approach is that it only checks for validity at `wrap` time, and doesn't remember layouts it has visited before. Users that pass a layout to `ak.XXX` functions with `highlevel=True` would then receive Arrays that have not been checked.

The more I think about it, the less confident I am that we *want* to go to all this trouble. We don't enforce `check_valid` by default, and that is the same class of error. *Maybe* we should include a simplification check when `check_valid` is set to `True`, and call it a day. Perhaps this is just a documentation problem?

I appreciate that I've gone round in a few circles here, thanks for coming for the ride!

jpivarski:
> The more I think about it, the less confident I am that we _want_ to go to all this trouble. We don't enforce `check_valid` by default, and that is the same class of error. _Maybe_ we should include a simplification check when `check_valid` is set to `True`, and call it a day.

At least in the two Content subclasses I checked yesterday, the `check_valid` implementation also checks for this redundancy. So if a user is concerned about the validity of a layout, this option will fix it.

All in all, there are three ways for a layout to be invalid:

[...]

PR #1334, opened 2022-03-02
Feat: validate Array simplification

agoose77:
As outlined in #1333, I think it would be useful to make it safer for users of the middle layer `ak._v2.contents` to create layouts. At present, it's possible (and not altogether hard) to create layouts that are _not_ simplified. This PR will not solve the problem of layouts with invalid offsets, but these sorts of mistakes are usually much easier to discover because things tend to fail quite noisily.

The purpose of this PR is to look at what this API might look like.

### DIFF
--- a/src/awkward/_v2/forms/indexedform.py
+++ b/src/awkward/_v2/forms/indexedform.py
@@ -133,6 +133,32 @@ class IndexedForm(Form):
             form_key=None,
         )
 
+    def simplify_optiontype(self):
+        if isinstance(
+            self._content,
+            (
+                ak._v2.forms.indexedoptionform.IndexedOptionForm,
+                ak._v2.forms.bytemaskedform.ByteMaskedForm,
+                ak._v2.forms.bitmaskedform.BitMaskedForm,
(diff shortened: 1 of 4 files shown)


agoose77:
Ideally to test whether one layout is simplified, I'd check whether the _form_ is simplified, rather than the array. Let's implement that for `ak._v2.forms.IndexedForm`.

agoose77:
As you can see, I thought that it would be useful to build out the `simplify_` methods for `Form` so that one does not need to actually simplify the array to test whether it *can* be simplified. This would require a number of functions that we currently implement on `ak._v2.contents.Content` to be duplicated for `ak._v2.forms.Form`. 

However, by doing this we end up with a lot of duplicated code between `Form` and `Content`. Furthermore, most of these methods do not depend upon their contents, hence `TypeTracer`. It would therefore seem that `ak._v2.forms.Form` is most useful in representing the form of an array, whereas the typetracer is more useful in performing transformations of the array. 

I think I'm going to close this, and further discuss in #1333

PR #1340, opened 2022-03-03
ak.flatten and ak.ravel should test for nplike.ndarray, not np.ndarray.

### LINKED ISSUE #1339: ak.flatten axis=None fails on typetracer array
### Version of Awkward Array

1.8.0

Minimal reproducer:

(I stepped through `ak._v2.flatten` until line line 101 from the traceback and `all(isinstance(x, np.ndarray) for x in out)` is evaluating to False since the elements of the `out` tuple in this case are instances of `TypeTracerArray`.)

```python
In [1]: import awkward._v2 as ak

In [2]: x = [[1, 2, 3], [2]]

In [3]: ak.Array(x)
Out[3]: <Array [[1, 2, 3], [2]] type='2 * var * int64'>

In [4]: x = ak.Array(x)

In [5]: x.layout.typetracer
Out[5]: 
<ListOffsetArray len='2'>
    <offsets><Index dtype='int64' len='3'>[?? ... ??]</Index></offsets>
    <content><NumpyArray dtype='int64' len='4'>[?? ... ??]</NumpyArray></content>
</ListOffsetArray>

In [6]: xtt = ak.Array(x.layout.typetracer)

In [7]: ak.flatten(xtt)
Out[7]: <Array-typetracer type='?? * int64'>

In [8]: ak.flatten(xtt, axis=None)
AssertionError                            Traceback (most recent call last)
Input In [8], in <cell line: 1>()
----> 1 ak.flatten(xtt, axis=None)

File ~/.pyenv/versions/3.10.2/envs/dask-awkward/lib/python3.10/site-packages/awkward/_v2/operations/structure/ak_flatten.py:101, in flatten(array, axis, highlevel, behavior)
     99 if axis is None:
    100     out = layout.completely_flatten(function_name="ak.flatten")
--> 101     assert isinstance(out, tuple) and all(isinstance(x, np.ndarray) for x in out)
    103     out = ak._v2.contents.NumpyArray(nplike.concatenate(out))
    105 elif axis == 0 or layout.axis_wrap_if_negative(axis) == 0:

AssertionError: 

In [9]: ak.flatten(x, axis=None)
Out[9]: <Array [1, 2, 3, 2] type='4 * int64'>
```

jpivarski:


agoose77:
@jpivarski Not sure if this is helpful, but I did a quick grep and found a few other places in `v2` where we reference `np.ndarray`, e.g. `ak.type`. I didn't check whether they're actually resolving against `numpy.ndarray`, or if the usage is acceptable.

jpivarski:
Do you mean these two?

https://github.com/scikit-hep/awkward-1.0/blob/61596516352066437e602040f0e89eb2d8a78a79/src/awkward/_v2/operations/structure/ak_fill_none.py#L54-L74

Actually, these are the `fill_value` for `ak.fill_none`, and they're coming from the user—NumPy arrays are one of the possible input types. The other cases I came across are similar.

Which cases are you talking about?

jpivarski:
`ak.type` is also just checking possible user inputs, which includes `np.ndarray`.

The thing that had to be fixed here is that these arrays are coming out of a layout (`layout.completely_flatten(...)`), and so they'd have whatever array type the `nplike` says. That's a different context from data that comes from a user.

agoose77:
Sure, I might be missing the obvious here. Is it possible that a user will knowingly call `ak.XXX` functions on `TypeTracerArray`? Is that a useful thing that a keen user might wish to do?

jpivarski:
It shouldn't be possible. The TypeTracer was created just to support Dask, though you may have found another good use for it in avoiding duplication between Content methods and Form methods.

It's in a module named with an underscore—there's no way to make one without invoking private code (so not even "mid-level," really "low-level"!). A user might get one from a Dask Awkward Array's `meta`, but hopefully this is considered private, too.

That doesn't preclude the possibility of accidents (bugs), but users aren't _supposed_ to be able to see them.

jpivarski:
It looks like this is safe for me to merge (no more work to do).

PR #1347, opened 2022-03-08
C++ refactoring: ak.run_lengths

ioanaif:


src/awkward/_v2/behaviors/string.py
```diff
@@ -250,7 +254,7 @@ def register(behavior):
     # behavior[ak.nplike.numpy.equal, "bytestring", "bytestring"] = _string_equal
     # behavior[ak.nplike.numpy.equal, "string", "string"] = _string_equal
     # behavior[ak.nplike.numpy.not_equal, "bytestring", "bytestring"] = _string_notequal
-    # behavior[ak.nplike.numpy.not_equal, "string", "string"] = _string_notequal
+    behavior[ak.nplike.numpy.not_equal, "string", "string"] = _string_notequal
```
jpivarski, line 257: With `_string_equal` implemented, it should be possible to enable all four of these. I guess only one was needed for the test. (Enabling these string and categorical behaviors is one of the cards.)

jpivarski (approved):
Also looks good. I see that you've enabled a string behavior (vectorized equality). That should allow tests like

```python
assert ak.all(ak.Array(["one", "two", "three"]) == ak.Array(["one", "two", "three"]))
assert ak.any(ak.Array(["one", "two", "three"]) != ak.Array(["one", "two", "3"]))
assert not ak.all(ak.Array(["one", "two", "three"]) != ak.Array(["one", "two", "3"]))
```

(Without the equality behavior, it would try to compare `"t"` with `"3"` and complain about the different number of characters; with the behavior, it would compare `"three"` with `"3"` as whole units.)

The `run_lengths` implementation also looks good.

PR #1348, opened 2022-03-08
Feat: add stack implementation

agoose77:
### Feature :sparkles: 
This PR adds an implementation of `np.stack`. This can already be done by users without *too* much difficulty, using `np.concatenate` and `np.newaxis`, but I think it's a fundamental enough function to warrant inclusion.

### Method :blue_book: 
The approach taken for `posaxis > 0` is to build a union over the existing lists, re-index their sublists, and then add a new list layout i.e.
`[A1 A2 A3 ... An]`, `[B1 B2 B3 ... Bn]` → `Union[A1 B1 A2 B2 A3 B3 ... An Bn]` →`List[[A1 B1] [A2 B2] [A3 B3] ... [An Bn]]`
For `posaxis == 0 `, we merge the contents into a single `Content`, and then create a new list layout over this content. This is simpler because we do not need to re-order the sublists as we do for `posaxis>0`

### Discussion :left_speech_bubble:  
I have not used `RegularArray` for the `posaxis > 0` case here, despite the fact that we *know* that the layout is regular. This is because I would like some input from the rest of the team as to what *should* happen. As we've discussed previously, the `RegularArray` layouts often introduce two sometimes conflicting behaviors - maintaining length information and informing behaviour of broadcasting, indexing.

My gut feeling is that we _should_ introduce a `RegularArray` layout, because we *know* it will always have a guaranteed constant length. 

I'd also be interested to know whether people think I should write a kernel for generating the tags and indices.

### DIFF
--- a/src/awkward/_v2/operations/structure/__init__.py
+++ b/src/awkward/_v2/operations/structure/__init__.py
@@ -23,6 +23,7 @@ from awkward._v2.operations.structure.ak_broadcast_arrays import (  # noqa: F401
     broadcast_arrays,
 )
 from awkward._v2.operations.structure.ak_concatenate import concatenate  # noqa: F401
+from awkward._v2.operations.structure.ak_stack import stack  # noqa: F401
 from awkward._v2.operations.structure.ak_where import where  # noqa: F401
 from awkward._v2.operations.structure.ak_flatten import flatten  # noqa: F401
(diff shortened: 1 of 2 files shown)


agoose77:
Closing for now as not a priority. Implementation is mostly there, will revisit in near future.

PR #1351, opened 2022-03-08
Feat: add `is_tuple` describe operation

agoose77:
### Feature :sparkles: 
This PR 
- adds a `ak.is_tuple` function to test whether a layout behaves as a tuple or a record.  
- adds `is_tuple`/`istuple` to all `Form` and `Content` types (C++ and Python) to handle recursion.

### Method :blue_book: 
To keep things simple, this just uses a recursive layout visitor. There is now a `_recursively_apply` API that supports early termination (by returning the same layout), but it doesn't support return values. The boilerplate to do this (`nonlocal` et al.) seems more indirected than a direct recursive visitor.

### Discussion :left_speech_bubble: 
I don't think we should add an `is_tuple` property to `ak.Array`, for the same reason that most of the Awkward API is in the form of free functions.

### DIFF
--- a/src/awkward/_v2/forms/unionform.py
+++ b/src/awkward/_v2/forms/unionform.py
@@ -235,6 +235,10 @@ class UnionForm(Form):
         return list(set.intersection(*[set(x) for x in fieldslists]))
 
     @property
+    def is_tuple(self):
+        return all(x.is_tuple for x in self._contents)
+
+    @property
     def dimension_optiontype(self):
         for content in self._contents:
             if content.dimension_optiontype:
--- a/src/awkward/_v2/forms/unionform.py
+++ b/src/awkward/_v2/forms/unionform.py
@@ -236,7 +236,7 @@ class UnionForm(Form):
 
     @property
     def is_tuple(self):
-        return all(x.is_tuple for x in self._contents)
+        return all(x.is_tuple for x in self._contents) and (len(self._contents) > 0)
 
     @property
     def dimension_optiontype(self):
--- a/src/libawkward/array/UnionArray.cpp
+++ b/src/libawkward/array/UnionArray.cpp
@@ -272,11 +272,12 @@ namespace awkward {
 
   bool
   UnionForm::istuple() const {
-  bool is_tuple = false;
+    bool all_contents_are_tuple = true;
     for (auto content : contents_) {
-        is_tuple = is_tuple && content.get()->istuple();
+        all_contents_are_tuple = all_contents_are_tuple && content.get()->istuple();
     }
-    return is_tuple;
+    //
+    return all_contents_are_tuple && (!contents_.empty());
   }
 
   bool
@@ -1456,11 +1457,12 @@ namespace awkward {
   template <typename T, typename I>
   bool
   UnionArrayOf<T, I>::istuple() const {
-  bool is_tuple = false;
+    bool all_contents_are_tuple = true;
     for (auto content : contents_) {
-        is_tuple = is_tuple && content.get()->istuple();
+        all_contents_are_tuple = all_contents_are_tuple && content.get()->istuple();
     }
-    return is_tuple;
+    //
+    return all_contents_are_tuple && (!contents_.empty());
   }
 
   template <typename T, typename I>
--- a/src/libawkward/array/UnionArray.cpp
+++ b/src/libawkward/array/UnionArray.cpp
@@ -276,7 +276,6 @@ namespace awkward {
     for (auto content : contents_) {
         all_contents_are_tuple = all_contents_are_tuple && content.get()->istuple();
     }
-    //
     return all_contents_are_tuple && (!contents_.empty());
   }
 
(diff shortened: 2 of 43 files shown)


agoose77:
Actually, I'm going to use the form for this, so that we aren't arbitrarily restricted to contents

jpivarski:
(Internal) methods like `purelist_depth` and `fields` can also be implemented using only Form information, too, but rather than construct an entire Form tree just to get the answer to a simple question, these methods are duck-typed between the concrete Form and corresponding Content. Like all of these:

https://github.com/scikit-hep/awkward-1.0/blob/2e8ce81763bfd263ff55e79838c221658ae9a4c0/src/awkward/_v2/contents/content.py#L1201-L1223

A similar pattern would work for `is_tuple` here.

Actually, instead of making a function for `recursively_apply`, what I would do is open all the src/awkward/_v2/forms/*.py files and put an `is_tuple` right after every `fields`, then add the duck-typed entry point in src/awkward/_v2/contents/content.py (next to `fields`).

There are competing advantages between the "put a method in every subclass" and "write a function that knows about all classes" techniques. Arguably, it's better to have the information associated with the subclass itself (making it easier to create a new subclass), and arguably, it's better to have the information for all subclasses in one place, for a particular function. Basically, we have a 2D table of classes and functionality, and sometimes it seems better to have whole rows together (all information about one class in one place) and sometimes it seems better to have whole columns together (all information about one function in one place). Since we have to make a choice, the choice for v2 has been to put it in the classes wherever possible, and in a function when it is not possible, which usually means more than one array. This is why `recursively_apply` is a method on all classes and `broadcast_and_apply` is a standalone function. `is_tuple` is very similar to `fields`, so I would do the same thing with it.

agoose77:
Right, the same thought occurred to me. I quite like the high-level separation of "core" vs "extra" traits & methods, but I can also see the benefit from a maintenance/style perspective of doing this in-line. I'll take your preference into account in future! :)

agoose77:
@jpivarski I've implemented this per-form now, and added a `Content` property that refers to the `Form`.

jpivarski (approved):
This is good—it fits with our current way of doing things. The ideas about factorizing between Contents and Forms are good ones, but this is the best way to implement `is_tuple` _right now_. Thank you!

Is this PR done? (It looks done.) If so, you can merge it.

agoose77:
@jpivarski it is done, and it's v2 only. Is it reasonable to only target v2 for new features? I can't recall where we stand on this.

jpivarski:
It's a small feature and there's no reason to _stop_ ourselves from adding v1 features if they're easy. The "new features in v2 only" is a policy that saves us from having to do twice the work. @masonproffitt's use-case is in v1 (I'm pretty sure).

agoose77:
@jpivarski could you give a once-over for this PR with its v1 changes?

src/libawkward/array/UnionArray.cpp
```diff
@@ -270,6 +270,15 @@ namespace awkward {
...
   }
 
+  bool
+  UnionForm::istuple() const {
+  bool is_tuple = false;
+    for (auto content : contents_) {
+        is_tuple = is_tuple && content.get()->istuple();
+    }
+    return is_tuple;
+  }
```
jpivarski: I just noticed this, not sure if it's in v2 also, but how could the result of this ever be true? `false && X` is `false`.

src/libawkward/array/UnionArray.cpp
```diff
@@ -1444,6 +1453,16 @@ namespace awkward {
...
 
+  template <typename T, typename I>
+  bool
+  UnionArrayOf<T, I>::istuple() const {
+  bool is_tuple = false;
+    for (auto content : contents_) {
+        is_tuple = is_tuple && content.get()->istuple();
+    }
+    return is_tuple;
+  }
```
jpivarski: Same here.

src/awkward/_v2/forms/unionform.py
```diff
@@ -234,6 +234,10 @@ def fields(self):
         fieldslists = [cont.fields for cont in self._contents]
         return list(set.intersection(*[set(x) for x in fieldslists]))
 
+    @property
+    def is_tuple(self):
+        return all(x.is_tuple for x in self._contents)
```
jpivarski: This is a good definition, but the v1 logic doesn't match it. (Unless I'm just unable to think straight on short notice. Anyway, this is a prompt to check it.)

jpivarski (changes requested):
Everything looks good except that I think the v1 UnionArray/UnionForm is wrong. See below.

src/libawkward/array/UnionArray.cpp
agoose77: @jpivarski Yep, that's wrong (I inverted the condition). I should add a test for the union too.

jpivarski (approved):

jpivarski:
The implementations of `Union*::is_tuple` appear to be correct now. Whenever you're done with the PR, go ahead and squash-and-merge!

PR #1352, opened 2022-03-08
C++ refactoring: ak.nan_to_num

ioanaif:


jpivarski (approved):
Great, it's using the new `_impl` and everything! It looks good to me (and the v1 implementation is still in recent memory), so I'll merge it.

PR #1355, opened 2022-03-09
This PR sets up the architecture to call CuPy Raw Kernels from Awkward.

sw5sh:
- [x] Convert awkward_ListArray_num to a CuPy raw kernel
- [x] Update the kernel dict to use these new Python files
- [x] Define a CuPy Kernel class to call these kernels
- [x] Load all kernels at the first invoke (make sure they are compiled and cached)


.github/workflows/cuda_build_test.yml
```diff
@@ -1,6 +1,6 @@
 name: CUDA Kernels Build
 
-on: [pull_request]
+on: []
 
 jobs:
   build-awkward-cuda-kernels:
```
jpivarski: Actually, you can get rid of the cuda_build_test entirely. All it did before was check that the kernels compile; now we can't even do that because we'd need CuPy to be installed.

We'll need something else for testing, though I don't know what that can be. I've brought up the question with colleages.

.gitignore
```diff
@@ -9,9 +9,10 @@ docs/demos/test-program.cpp
 include/awkward/kernels.h
 src/awkward/_kernel_signatures.py
 src/awkward/_kernel_signatures_cuda.py
+src/awkward/_v2/_connect/cuda/_kernel_signatures.py
 
 # Generated CUDA kernels
-src/cuda-kernels/awkward_*.cu
+# src/cuda-kernels/awkward_*.cu
```
jpivarski: It can be deleted, rather than commented out. Git is providing the history of it having been there. (It will be easier to find it in the blame log if it's one change, existence → non-existence, than two: existence → commented out → non-existence.)

dev/generate-kernel-signatures.py
```diff
@@ -12,61 +12,61 @@
...
+    # "awkward_RegularArray_broadcast_tooffsets",
+    # "awkward_NumpyArray_fill_tobool",
+    # "awkward_NumpyArray_reduce_adjust_starts_64",
+    # "awkward_NumpyArray_reduce_adjust_starts_shifts_64",
+    # "awkward_regularize_arrayslice",
+    # "awkward_RegularArray_getitem_next_at",
+    # # "awkward_ListOffsetArray_compact_offsets", Need to tune tests
     "awkward_BitMaskedArray_to_ByteMaskedArray",
-    "awkward_BitMaskedArray_to_IndexedOptionArray",
+    # "awkward_BitMaskedArray_to_IndexedOptionArray",
```
jpivarski: This is the end of a long commented-out block. For git-related reasons, it's better to delete than comment-out.

I still use commenting-out a lot in short-term work and in the midst of a PR that isn't ready to be reviewed, but when these things get committed to history, let's do the simple thing and delete them outright.

src/awkward/_v2/_connect/cuda/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+kernel_specializations = {
+    "awkward_ListArray32_num_64": "cuda_ListArray_num<int32_t, int64_t>",
+    "awkward_ListArrayU32_num_64": "cuda_ListArray_num<uint32_t, int64_t>",
+    "awkward_ListArray64_num_64": "cuda_ListArray_num<int64_t, int64_t>",
+    "awkward_RegularArray_num_64": "cuda_RegularArray_num<int64_t>",
+    "awkward_BitMaskedArray_to_ByteMaskedArray": "cuda_BitMaskedArray_to_ByteMaskedArray",
+    "awkward_ListArray32_validity": "cuda_ListArray_validity<int32_t>",
+    "awkward_ListArrayU32_validity": "cuda_ListArray_validity<uint32_t>",
+    "awkward_ListArray64_validity": "cuda_ListArray_validity<int64_t>",
+}
```
jpivarski: Should this be hand-written (and therefore in git) or auto-generated? It looks like something that would be easier to auto-generate.

Also, for the prefix ("`cuda_`" rather than "`awkward_`"), the reason we have a prefix is to avoid collisions in a namespace—suppose that a user is also writing `cupy.RawKernel` and they use the same names. If we don't have actual C++ namespaces with "`::`", we use a prefix. (The cpu-kernels have a C interface, so no namespaces.)

  1. Perhaps you want the prefix to be "`awkward`", rather than "`cuda`", since you're trying to disambiguate these functions from other CUDA functions, but not from other Awkward functions. (One of the benefits of a library with an odd name is that no one else is likely to pick the same thing.)
  2. Can you use real C++ namespaces now that we're JIT-compiling?

For the automated mapping of names, your starting point is what the kernel calls are doing:

```python
nplike["name_of_kernel_without_specialization", np.uint32, np.int64](arguments)
```

In your bookkeeping, you need to get from a kernel name without specialization plus the dtype of every array (pointer-based) argument passed into or out of it. If you arrange your kernels to be templated on all the pointer-based arguments, regardless of whether it needs to be (you might have only one specialization), then the mapping can be automated.

src/awkward/_v2/_connect/cuda/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def initialize_cuda_kernels(cupy):
+    if cupy is not None:
+        global kernel
+
+        if kernel is None:
+            import awkward._v2._connect.cuda._kernel_signatures
+
+            cuda_src = f"#define ERROR_BITS {error_bits}\n #define MAX_NUMPY_INT {numpy.iinfo(numpy.int64).max}"
```
jpivarski: Early decision to make: should the error integer be `int64` or `uint64`? That will determine what the maximum value is.

  * It's pretty certain that we'd never have more than `np.iinfo(np.int64).max` kernel calls, so we don't _need_ it to be unsigned int.
  * But this number will never be negative, and (the reason that I usually pick signed anyway) it will never be subtracted or compared with any value that might be negative. There's no reason _not_ to make it unsigned.

Whatever you do, you want to be consistent with that.

Also, the name `MAX_NUMPY_INT` is not saying what this is about. It has nothing to do with NumPy (though that's how you come up with the value) and everything to do with being the `NO_ERROR` case. The code will be more understandable if the name reflects its role.

src/awkward/_v2/_connect/cuda/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def synchronize_cuda(stream):
+    cupy = import_cupy("Awkward Arrays with CUDA")
+
+    stream.synchronize()
+    invocation_index = cuda_streamptr_to_contexts[stream.ptr][0]
+    contexts = cuda_streamptr_to_contexts[stream.ptr][1]
+
+    if invocation_index != numpy.iinfo(numpy.int64).max:
```
jpivarski: Create a Python constant with the same name as a the C++ constant, such as `NO_ERROR`, so that the "why?" of this comparison is more clear. Also, it keeps the choice of `int64` vs `uint64` from being spread everywhere.

src/awkward/_v2/_connect/cuda/__init__.py
```diff
@@ -0,0 +1,142 @@
...
+
+    pip install cupy
+
+or
+
+    conda install -c conda-forge cupy
+"""
+
+cuda_streamptr_to_contexts = {}
+error_bits = 8
```
jpivarski: Constants like this are usually all caps in Python: `ERROR_BITS`.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_BitMaskedArray_to_ByteMaskedArray.cu
```diff
@@ -0,0 +1,50 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+__global__ void
```
jpivarski, line 3: Would `__device__` work as well? None of these kernels are going to run on the CPU, right? Declaring them as GPU-only should cut the compilation time in half, and now that matters because they're being compiled while the user waits.

Also, if there's some reason why one of these kernels can't be compiled on the CPU—because you use some intrinsic, for instance—then avoiding the attempt removes that difficulty.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_BitMaskedArray_to_ByteMaskedArray.cu
```diff
@@ -0,0 +1,50 @@
...
+cuda_BitMaskedArray_to_ByteMaskedArray(int8_t* tobytemask,
+  const uint8_t* frombitmask,
+  int64_t bitmasklength,
+  bool validwhen,
+  bool lsb_order,
+  int64_t invocation_index,
+  int64_t* err_code) {
+  int64_t thread_id = blockIdx.x * blockDim.x + threadIdx.x;
+  if (thread_id < bitmasklength && err_code[0] == MAX_NUMPY_INT) {
```
jpivarski: If it's more clear or easier to automate, this could become two separate `if` statements.

Every kernel will need to check `err_code[0] == NO_ERROR`. That becomes something you'd want to copy-paste without thinking.

The `if (thread_id < bitmasklength)` is something you have to think about when writing every kernel: which function argument corresponds to the array length you don't want to overstep can have a different name in every kernel.

I'm 99% sure that two nested `if` statements either generates the same code as one `if` with an `&&` or is otherwise just as fast.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_ListArray_validity.cu
```diff
@@ -0,0 +1,34 @@
...
+    C stop = stops[thread_id];
+    if (start != stop) {
+      if (start > stop) {
+        atomicMin(err_code, invocation_index * (1 << ERROR_BITS) + ERROR_START_STOP); // failure("start[i] > stop[i]", i, kSliceNone, FILENAME(__LINE__));
+      }
+      if (start < 0) {
+        atomicMin(err_code, invocation_index * (1 << ERROR_BITS) + ERROR_START_ZERO); // failure("start[i] < 0", i, kSliceNone, FILENAME(__LINE__));
+      }
+      if (stop > lencontent) {
+        atomicMin(err_code, invocation_index * (1 << ERROR_BITS) + ERROR_STOP_CONTENT); // failure("stop[i] > len(content)", i, kSliceNone, FILENAME(__LINE__));
```
jpivarski: The commented-out parts can be removed now.

Since this error-setting operation is one that would be done all over the place, maybe you want a utility function for it? Maybe even a [preprocessor macro](https://gcc.gnu.org/onlinedocs/cpp/Function-like-Macros.html#Function-like-Macros)? I don't know what `cupy.RawKernel` lets you do. Ideally, you'd like to be able to write these lines as

```c++
raise_error(ERROR_STOP_CONTENT)
```

and the `raise_error` macro would define once-and-for-all the `atomicMin`, the fact that we're overwriting `err_code`, the arithmetic on `invocation_index` and `ERROR_BITS`, etc. If it can't be a preprocessor macro but can be a function, you'd have to pass `err_code` and `invocation_index` as arguments, along with the specific error.

src/awkward/nplike.py
```diff
@@ -422,6 +420,60 @@ def __call__(self, *args):
         )
 
 
+class CupyKernel(NumpyKernel):
+    def max_length(self, args):
+        cupy = ak._v2._connect.cuda.import_cupy("Awkward Arrays with CUDA")
+        max_length = numpy.iinfo(numpy.int64).min
```
jpivarski, line 426: This is why you want that to be a Python constant. It's showing up everywhere.

tests-cuda/test_1270-install-test.py
```diff
@@ -3,6 +3,7 @@
 import pytest  # noqa: F401
 
 
+@pytest.mark.skip("We don't use shared library for accessing CUDA Kernels anymore.")
```
jpivarski: Rather than skipping, this test can be removed. We'd skip it if our intention was to reinstate it in the future, but if doing everything through CuPy is the right way to go (and it looks like it is), then we'd never have this library and wouldn't have to test loading it.

tests-cuda/v2/test_1276-cuda-num.py
```diff
@@ -40,7 +40,7 @@ def test_num_3():
         ak.contents.numpyarray.NumpyArray(np.array([0.0, 1.1, 2.2, 3.3, 4.4, 5.5]))
     )
     cuda_array = ak.to_backend(array, "cuda")
-    ak.num(cuda_array, 0)
+    assert ak.num(cuda_array, 0) == ak.num(array, 0)
```
jpivarski: Although Doug in dask-awkward is using `import awkward._v2 as ak`, we shouldn't use it in the Awkward codebase because we don't want one codebase to mix both `import awkward._v2 as ak` and `import awkward as ak`. That would get confusing!

I was about to say, "this should be `ak._v2.num`, rather than `ak.num`," until I noticed how it was imported.

jpivarski (changes requested):
I have a lot of comments (easier to do this on the code itself, rather than in person on Zoom), but they're all small things. The structure of it is looking good.

One more thing: you didn't remove the src/awkward_cuda_kernels directory. That should go, too. Also, cuda-build.sh.

I don't know how we're going to do continuous testing of the CUDA kernels, but the tests made by [gencudakerneltests](https://github.com/scikit-hep/awkward-1.0/blob/def3e69a7f1f21f7141b822205db8c85b10f213b/dev/generate-tests.py#L537-L705) will no longer work. Until further notice, you should delete that function, too. And the [place where it's called](https://github.com/scikit-hep/awkward-1.0/blob/def3e69a7f1f21f7141b822205db8c85b10f213b/dev/generate-tests.py#L751).

That's everything that I can see!

src/awkward/_v2/_connect/cuda/__init__.py
sw5sh: There were a lot of ambiguity when I tried to go down this route. I came up with an alternate solution of opening each and every `cpu-kernel` and extracting the specializations from there. There are only a handful of ambiguous cases (5), which can be handwritten when those respective kernels are implemented.

1. Yes, the kernels are now prefixed with `awkward_`.
2. I can use real C++ namespaces now, is `cupy`  a good namespace since these kernels are actually `cupy` RawKernel's?

src/awkward/_v2/_connect/cuda/__init__.py
sw5sh: I decided to move ahead with `np.uint64`.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_BitMaskedArray_to_ByteMaskedArray.cu
sw5sh: A `__device__` function can only be called from the GPU itself. We define `__global__` so that the kernel can be invoked from the CPU and the GPU (The first call to the kernel is always from the CPU). Regardless of the flag, these kernels cannot run on a CPU. These flags exist as invocation rules.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_ListArray_validity.cu
sw5sh: Turned this into a `preprocessor macro` `RAISE_ERROR(ERROR)`

src/awkward/nplike.py
sw5sh: This is for a different use-case.

jpivarski: Aha! You're right, it is. (That's more reason for wanting the `NO_ERROR` to look different.)

jpivarski (approved):

src/awkward/_v2/_connect/cuda/__init__.py
jpivarski: (We talked about it in Zoom.)

Removing the renaming-dict will be in the next PR. This one is ready to go.

PR #1360, opened 2022-03-10
C++ refactoring: ak.unflatten

ioanaif:


src/awkward/_v2/operations/structure/ak_unflatten.py
```diff
@@ -1,204 +1,200 @@
...
+    else:
+        counts = ak._v2.operations.convert.to_layout(
+            counts, allow_record=False, allow_other=False
+        )
+
+        if counts.is_OptionType:
+            mask = counts.mask_as_bool(valid_when=False)
+            counts = counts.to_numpy(allow_missing=True)
+            counts = ak.nplike.numpy.ma.filled(counts, 0)
+        elif counts.is_NumpyType or counts.is_UnkownType:
```
jpivarski:
```suggestion
        elif counts.is_NumpyType or counts.is_UnknownType:
```

src/awkward/_v2/operations/structure/ak_unflatten.py
```diff
@@ -1,204 +1,200 @@
...
+                positions[0] = 0
+
+                return ak._v2.contents.ListOffsetArray(
+                    ak._v2.index.Index64(positions), content
+                )
+
+            else:
+                return layout
+
+        out = transform(layout, depth=1, posaxis=axis)
```
jpivarski, line 188: Wrapping this up as a `transform` function that is only ever used once is now superfluous, but not harmful. In v1, it was needed so that the same procedure could be applied to each partition, but now we don't do PartitionedArrays anymore.

I also don't see a strong motivation to refactor it.

jpivarski (approved):
Great! I'll click the box that fixes the typo in the spelling of `is_UnknownType`.

PR #1362, opened 2022-03-10
Feat: add cumsum

agoose77:
This PR adds `np.cumsum` support to v2


agoose77:
Rebased onto https://github.com/scikit-hep/awkward-1.0/pull/1365

agoose77:
This is functionally mostly there, but I'm closing it for now as it's not a priority. I'll revisit in the near future!

PR #1365, opened 2022-03-10
Refactor: cleanup reducer

agoose77:
`ListOffsetArray` and the `Indexed[Option]Array` types have a non-negligible amount of boilerplate from v1 that we can eliminate. In addition, there is a reasonable overlap between the `reduce`, `[arg]sort`, and `unique` pathways that we can move into preparatory functions.

I've changed the logic here quite substantially (in the case of `IndexedOptionArray`), so it would be good to validate that the test suite is covering these changes.

### DIFF
--- a/src/awkward/_v2/contents/indexedarray.py
+++ b/src/awkward/_v2/contents/indexedarray.py
@@ -821,7 +821,7 @@ class IndexedArray(Content):
             if isinstance(unique, ak._v2.contents.RegularArray):
                 unique = unique.toListOffsetArray64(True)
 
-            elif isinstance(unique, ak._v2.contents.ListOffsetArray):
+            if isinstance(unique, ak._v2.contents.ListOffsetArray):
                 if starts.nplike.known_data and starts.length > 0 and starts[0] != 0:
                     raise ak._v2._util.error(
                         AssertionError(
--- a/src/awkward/_v2/contents/indexedarray.py
+++ b/src/awkward/_v2/contents/indexedarray.py
@@ -998,13 +998,19 @@ class IndexedArray(Content):
             keepdims,
         )
 
+        # If we are reducing the contents of this layout,
+        # then we do NOT want to return an optional layout
         if not branch and negaxis == depth:
             return out
         else:
             if isinstance(out, ak._v2.contents.RegularArray):
                 out = out.toListOffsetArray64(True)
 
-            elif isinstance(out, ak._v2.contents.ListOffsetArray):
+            # If the result of `_reduce_next` is a list, and we're not applying at this
+            # depth, then it will have offsets given by the boundaries in parents.
+            # This means that we need to look at the _contents_ to which the `outindex`
+            # belongs to add the new index
+            if isinstance(out, ak._v2.contents.ListOffsetArray):
                 if starts.nplike.known_data and starts.length > 0 and starts[0] != 0:
                     raise ak._v2._util.error(
                         AssertionError(
@@ -1032,7 +1038,7 @@ class IndexedArray(Content):
                     )
                 )
 
-                tmp = ak._v2.contents.IndexedArray(
+                inner = ak._v2.contents.IndexedArray(
                     outindex,
                     out._content,
                     None,
@@ -1042,7 +1048,7 @@ class IndexedArray(Content):
 
                 return ak._v2.contents.ListOffsetArray(
                     outoffsets,
-                    tmp,
+                    inner,
                     None,
                     None,
                     self._nplike,
--- a/src/awkward/_v2/contents/listoffsetarray.py
+++ b/src/awkward/_v2/contents/listoffsetarray.py
@@ -899,82 +899,14 @@ class ListOffsetArray(Content):
             if self._nplike.known_shape and parents.nplike.known_shape:
                 assert self._offsets.length - 1 == parents.length
 
-            nextlen = self._offsets[-1] - self._offsets[0]
-            maxcount = ak._v2.index.Index64.empty(1, self._nplike)
-            offsetscopy = ak._v2.index.Index64.empty(self._offsets.length, self._nplike)
-            assert (
-                maxcount.nplike is self._nplike
-                and offsetscopy.nplike is self._nplike
-                and self._offsets.nplike is self._nplike
-            )
-            self._handle_error(
-                self._nplike[
-                    "awkward_ListOffsetArray_reduce_nonlocal_maxcount_offsetscopy_64",
-                    maxcount.dtype.type,
-                    offsetscopy.dtype.type,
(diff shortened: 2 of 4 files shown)


jpivarski:
Looking good so far. A definition of `_prepare_next` wasn't included (so I don't see how the tests could pass). Should the name of that helper function be so generic? (I saw what you said on Gitter about naming things!) Is it just for reducers and operations that handle `axis` in the same way as reducers but don't change the number of dimensions (e.g. sort, cumsum, ...)?

On naming things, "reducers" and "rearrangers"? "Rearrange" sounds like it's only permutations/sorting, which cumsum is not, but it's in the same spirit.

agoose77:
> Looking good so far. A definition of `_prepare_next` wasn't included

Weird, I wonder why we can't see it from the UI? It's here: https://github.com/scikit-hep/awkward-1.0/blob/bf25bee5a2e8b66c41cfb7a5d8b9681acd5c5fdd/src/awkward/_v2/contents/indexedoptionarray.py#L1172-L1214

agoose77:
At this stage, I'm mainly just pulling things here and there to see what's fundamental vs a variation. It's tricky to fully refactor this because a lot of the SLOC are just kernel calls which are bulky but "primitive"

agoose77:
>  naming things, "reducers" and "rearrangers"? "Rearrange" sounds like it's only permutations/sorting, which cumsum is not, but it's in the same spirit.

I like this name!

agoose77:
@ianna would you be able to look over [cb3758c](https://github.com/scikit-hep/awkward-1.0/pull/1365/commits/cb3758c3e10de381cb16da600f319392a5c0c3cd) and see whether I've done anything stupid / breaking?

I've tried to rearrange the conditions to reduce the amount of code needed here, but there's a possibility that I've reasoned about it wrong.

agoose77:
OK, I've addressed `IndexedOptionArray`. I *think* we can really only factor out the _preparatory_ logic into its own function. The rest of the routines are either sufficiently bespoke or terse that it doesn't add much to make a new function. 

I have *not* replicated the original logic. It looks like we had some branches that would never be visited (e.g., anything in the rearranger routines would only a non-`IndexedOptionArray` if `depth != 1`, because we explicitly have `out = IndexedOptionArray(...)`). These changes pass the test suite, but reducer logic is non-trivial, so I would not be surprised if somewhere I've missed something. Now that I understand this logic a bit better, I think it should be easier to extend to other layout types.

src/awkward/_v2/contents/indexedarray.py
```diff
@@ -998,13 +998,19 @@ def _reduce_next(
...
         else:
             if isinstance(out, ak._v2.contents.RegularArray):
                 out = out.toListOffsetArray64(True)
 
-            elif isinstance(out, ak._v2.contents.ListOffsetArray):
+            # If the result of `_reduce_next` is a list, and we're not applying at this
+            # depth, then it will have offsets given by the boundaries in parents.
+            # This means that we need to look at the _contents_ to which the `outindex`
+            # belongs to add the new index
+            if isinstance(out, ak._v2.contents.ListOffsetArray):
```
jpivarski, line 1013: How are you sure that this needs to be an `if` and not an `elif`?

We have no tests that satisfy the `if isinstance(out, ak._v2.contents.RegularArray)` predicate. (I put an Exception in there and ran the tests: none of them failed.)

Aha: it's what happened in v1:

https://github.com/scikit-hep/awkward-1.0/blob/34094aa26e498210047428b2ecde6d0b1aa1fc7f/src/libawkward/array/IndexedArray.cpp#L2260-L2265

Although it would be better to have tests, conformance with v1 is a winning argument.

src/awkward/_v2/contents/indexedarray.py
```diff
@@ -821,7 +821,7 @@ def _unique(self, negaxis, starts, parents, outlength):
             if isinstance(unique, ak._v2.contents.RegularArray):
                 unique = unique.toListOffsetArray64(True)
 
-            elif isinstance(unique, ak._v2.contents.ListOffsetArray):
+            if isinstance(unique, ak._v2.contents.ListOffsetArray):
```
jpivarski, line 824: I had the same question about this as the previous one, but I can see that they're symmetric bits of code.

src/awkward/_v2/contents/indexedarray.py
jpivarski: Also, it's what we have in IndexedOptionArray, and that's very similar to the IndexedArray case. In fact, the IndexedArray _might_ be superfluous, since they had to share an implementation in v1, when IndexedOptionArray was a templated variation on IndexedArray. So this _might_ be a historical artifact.

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -1778,15 +1500,15 @@ def _reduce_next(
...
                     gaps.data,
                     outlength,
                 )
             )
 
             if reducer.needs_position:
-                nextshifts = ak._v2.index.Index64.empty(nextlen, self._nplike)
-                nummissing = ak._v2.index.Index64.empty(maxcount[0], self._nplike)
+                nextshifts = ak._v2.index.Index64.empty(nextcarry.length, self._nplike)
+                nummissing = ak._v2.index.Index64.empty(maxcount, self._nplike)
```
jpivarski, line 1511: I was wondering about how `maxcount` lost its `[0]`, but now I see that `_rearrange_prepare_next` returns the scalar value, rather than the one-element array.

This might come up again if `maxcount` is ever a TypeTracerArray. But the beauty of this solution is that we'd only have to fix it in one place (in the `_rearrange_prepare_next` function) if we do have to change it.

jpivarski (approved):
That's a lot of changes, and it's hard to tell from this altitude that they're right.

This is one of those times when you wish you had 100% test coverage.  `:(`

I scanned through it a second time, more carefully, and noted the things that bothered me most, but in each case, I could convince myself that what you've written is correct.

Thanks a lot—these are some deep and detailed edits!

agoose77:
> That's a lot of changes, and it's hard to tell from this altitude that they're right.
> 
> This is one of those times when you wish you had 100% test coverage. `:(`
> 
> I scanned through it a second time, more carefully, and noted the things that bothered me most, but in each case, I could convince myself that what you've written is correct.
> 
> Thanks a lot—these are some deep and detailed edits!

Thanks Jim. I noticed a couple of docs typos while reading the diff again (it's amazing what coming back to something can do for perspective), but I can fix those up in a later PR. 

I agree that this is not something that we can grep by the diffs alone. I did make some (albeit reasoned) big changes here, so I hope we've not introduced any new bugs. Still, the tests passing suggests we might be OK, and we can be suspicious of this PR as a default position if anything does come up!

PR #1367, opened 2022-03-13
C++ refactoring: ak.copy

ioanaif:


jpivarski (commented):
`Index.__copy__` is misnamed and doesn't handle `metadata` and `nplike`; expand into two: `__copy__` and `__deepcopy__`, where only the latter copies the array.

ak._v2.highlevel.Array and Record can get `__copy__` and `__deepcopy__` methods now, too.

jpivarski (approved):
Everything here looks good. Thanks!

PR #1368, opened 2022-03-13
C++ refactoring: ak.broadcast_arrays

ioanaif:


jpivarski (approved):
Great as-is, but maybe move `test_broadcast_arrays` before `pytest.importorskip("numexpr")` to make sure it gets tested (this one only depends on NumPy, not numexpr).

Go ahead and merge when ready.

PR #1369, opened 2022-03-14
C++ refactoring: ak.to_pandas

ioanaif:


src/awkward/_v2/contents/regulararray.py
```diff
@@ -1076,6 +1076,10 @@ def _rpad(self, target, axis, depth, clip):
             )
 
     def _to_numpy(self, allow_missing):
+        array_param = self.parameter("__array__")
+        if array_param in {"bytestring", "string"}:
+            return self._nplike.array(self.to_list())
```
jpivarski, line 1081: My eye jumped to this because it's converting the all the Awkward Array data into Python objects, and then into NumPy, but it's only for the case of strings. It could be done a little better: we could `ak.pad_none`, `ak.fill_none` the character data to have equal length and then cast the result as NumPy, but for strings (not bytestrings), there's the added complication that Awkward's unicode representation is UTF-8 and NumPy's is UTF-32, and that's not worth getting into.

jpivarski (approved):
Everything here looks good, too. Thanks for this!

PR #1373, opened 2022-03-17
Feat: add `depth_limit` to `ak.broadcast_arrays`

agoose77:
This behaves identically to the `depth_limit` argument of `ak.zip`, which arrests broadcasting at a given depth. Because most of the code is directly copied from `ak.zip`, it should be a fairly safe new feature.


jpivarski (approved):
Nice!

(This one was a lot easier to review.)

PR #1381, opened 2022-03-23
This PR adds the generated kernels and simplifies the template specialization generation process.

sw5sh:
The `cuda-kernels` were compared to the `cpu-kernels` and all kernels compile, however they are not yet tested using some `test-data`


sw5sh:
This PR should be done with the cleanups discussed in the Wednesday meeting.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_ListArray_compact_offsets.cu
```diff
@@ -1,26 +1,25 @@
 // BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
 enum class LISTARRAY_COMPACT_OFFSETS_ERRORS {
-    ERROR_START_STOP,    // message: "start[i] > stop[i]"
+  ERROR_START_STOP,  // message: "start[i] > stop[i]"
 };
```
jpivarski, line 5: This module is not sharing a namespace with all the other modules, is it? The use of `enum class` and verbose names is to keep names from colliding, but if (as I suspect) the names in this .cu are not visible in the other .cu files (through `cupy.RawKernels`), then it's safe to be less verbose. That's not a necessity, but it will reduce the work needed to scale up the number of .cu files—a lot of them have errors.

For for example, the phrase "LISTARRAY_COMPACT_OFFSETS" doesn't have to appear here, nor does the word "ERROR" have to be repeated. You can still use an `enum class` to isolate them in a namespace, but it could be

```c++
enum class ERROR {
  START_STOP,   // message: "start[i] > stop[i]"
};
```

and then used as `ERROR::START_STOP`. Or just

```c++
enum {
  ERROR_START_STOP,   // message: "start[i] > stop[i]"
};
```

and used as `ERROR_START_STOP`.

Again, all of this is assuming that when some other `cupy.RawModule` defines `ERROR::START_STOP` or `ERROR_START_STOP` as some other constant integer, there won't be any confusion about which one is meant. I'm expecting the symbols defined in different `cupy.RawModules` to be inaccessible in each other: that they can only see classes, functions, and enums that are explicitly shared in common header files. If I'm wrong about that, then you do need to disambiguate them like this.

jpivarski (approved):
This looks good! I'm quibbling about enums (below), but this is in a good state to merge as-is.

src/awkward/_v2/_connect/cuda/cuda_kernels/awkward_ListArray_compact_offsets.cu
sw5sh: The way we put our code in `RawModule` means that all the declarations can see each other, so making a distinction with this verbose name became a necessity.

PR #1384, opened 2022-03-24
Fix _prettyprint after 'for i in range' changed to 'for i, val in enumerate'.

jpivarski:
Just FYI @henryiii: changing

```python
for i in range(len(out)):
    if i > 0:
        out[i] = " " + out[i]
    else:
        out[i] = "[" + out[i]
    if i < len(out) - 1:
        out[i] = out[i] + ","
    else:
        out[i] = out[i] + "]"
```

into a for loop with

```python
for i, val in enumerate(out):
```

and all `out[i]` on the right-hand side of assignments converted into `val` broke this functionality. The original for loop relied on a side-effect in which `out[i]` was changed twice. Using an assigned-once name, `val`, prevented the first change from being seen in the final result.

I repeatedly run into cases in which side-effect turn out to be a bad idea (there's really something to this pure functional paradigm), and this could be counted as one of them. But side-effect based programming is a part of ordinary Python use and I wanted to bring to your attention that there are corner-cases like this in which `for i in range(len(XYZ))` → `for i, val in enumerate(XYZ)` can be dangerous. Not to avoid these refactorings, but to keep an eye out for them!

(Also, I haven't yet checked the history to be sure that that's how this bug crept in, I'm just guessing that because it looks like it.)

### DIFF
--- a/src/awkward/_v2/_prettyprint.py
+++ b/src/awkward/_v2/_prettyprint.py
@@ -180,13 +180,14 @@ def valuestr(data, limit_rows, limit_cols):
         out = front + back
         for i, val in enumerate(out):
             if i > 0:
-                out[i] = " " + val
+                val = out[i] = " " + val
             else:
-                out[i] = "[" + val
+                val = out[i] = "[" + val
             if i < len(out) - 1:
                 out[i] = val + ","
             else:
                 out[i] = val + "]"
+
         return "\n".join(out)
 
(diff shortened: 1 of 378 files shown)


henryiii:
I tried to watch for this, I know I avoided converting at least 1-2 because this side effect was being used. It's quite subtle - the new version is much better, IMO, since the reader can immediately see it's being intentionally reassigned.

Issue #1283, opened 2022-02-10
ak.sum segfault: backport v2 solution into v1


alexander-held:
### Version of Awkward Array

1.7.0

### Description and code to reproduce

```python
import awkward as ak
import numpy as np

v = ak.from_iter([[[1, 2, 3], [4, 5]], [[3, 4, 5], [6, 7]], [[5, 6, 7], [8, 9]]])
M = np.asarray([[1, 1.4, -0.3], [1.4, 1, 1.2], [-0.3, 1.2, 1]])
M_times_v = M[..., np.newaxis, np.newaxis] * v[np.newaxis, ...]
v_times_M_times_v = v[:, np.newaxis, ...] * M_times_v

# frequently (not always) segfaults with axis=0 or 1
np.sum(v_times_M_times_v, axis=0)

array = ak.from_iter([[[[1], [4, 9]], [[5.6], [14, 25.2]]], [[[5.6], [14, 25.2]], [[16], [25, 36]]]])
# always(?) segfaults with axis=0 or 1, also when using np.sum
ak.sum(array, axis=1)
```

The above code causes frequent segmentation faults in the first `np.sum` call, and always (at least in ~100 attempts) segfaults for the second `ak.sum` call (same when replacing this by `np.sum`).

I can reproduce this with two setups:
- Python 3.9.10 on M1 macOS with `awkward` 1.7.0 and `numpy` 1.22.2
- Python 3.8.10 on Intel macOS with `awkward` 1.7.0 and `numpy` 1.22.1

I cannot reproduce the segmentation fault when using the following:
- `python:3.9-slim` container with `awkward` 1.7.0 and `numpy` 1.22.2

In case it is useful, I rarely get the following output as well:
```
python(14372,0x102a64580) malloc: Heap corruption detected, free list is damaged at 0x600000358260
*** Incorrect guard value: 8
python(14372,0x102a64580) malloc: *** set a breakpoint in malloc_error_break to debug
```

In the original setup where I came across this (uses a larger array as input) I also see the following:
```
python(12981,0x105568580) malloc: *** error for object 0x3d: pointer being freed was not allocated
python(12981,0x105568580) malloc: *** set a breakpoint in malloc_error_break to debug
```
The original setup also crashes less frequently (did not measure this very precisely though).

I am not sure how to approach debugging this, but am happy to help with some pointers for what to do. I have also not tried other combinations of `numpy` and `awkward` yet or systematically investigated which kinds of arrays work and which cause crashes.

[...]

PR #1388, opened 2022-03-29
If this test passes, #1283 is fixed.

jpivarski:


jpivarski:
Right: it fails because #1274 fixes v2 without fixing v1. I just wanted to confirm that this is where it stands for now.

jpivarski:
This PR is included within #1389 and doesn't need to stay open. #1389 takes over its role as the one to close #1283.

PR #1389, opened 2022-03-30
Fix: sort v1 parents

### LINKED ISSUE #1283: ak.sum segfault: backport v2 solution into v1

agoose77:
This is a very hacky fix that just runs `std::stable_sort` on the CPU. Given that we won't (?) improve CUDA support in v1, and this is a v1-only fix, I think it's reasonable to (a) assume we're on the CPU, and (b) do this inline instead of with a CPU kernel.


jpivarski:
That's true: we won't be supporting the GPU in v1 (the experimental support has been removed), so a direct call to `std::stable_sort` is a perfectly acceptable solution (and a good idea!).

agoose77:
I naively expected this to fix the MacOS tests. It *seems* like it _maybe_ fixes 3.6, but not 3.7+. Perhaps this is a fluke, or maybe it reveals that there is more than one contributing bug here. 

First, I'll re-run the 3.6 test.

jpivarski:
We talked about this at today's meeting; it's old, addresses v1 only, and there haven't been any complaints about it not being done. I'm closing it. If need be, it can be reopened as a bug-fix in the 1.10.x series (v1 bug-fixes).

PR #1390, opened 2022-03-31
This PR adds all the remaining kernels in the studies directory

sw5sh:


jpivarski (approved):
That's a lot of kernels!

I tested it (because tests-cuda are not part of CI, yet) and made a correction to get it working. (The `pytest.raises` context manager returns an ExceptionInfo whose `__str__` is not the same as that of the ValueError.)

This is a nice conclusion to your semester's work on the GPU infrastructure. Congrats!

PR #1397, opened 2022-04-04
split up functions

martindurant:
Quite aside from the options we talked about for various ways to handle file layout, this kind of splitting up of functionality is very helpful for reusing the code in dask. I haven't checked whether this actually works yet, I probably mistyped along the way, but I wanted to show the idea.


jpivarski:
It's a good factorization. Be aware that your git submodules are out of date, and that's why they're being counted as diffs in this PR. @agoose77 also suggested turning the ErrorContext manager into a decorator, which is something we should do across all the functions, not in this PR.

martindurant:
> your git submodules are out of date

Remind me how to make this go away?

> something we should do across all the functions, not in this PR.

Agreed - it's here because it helped me reason about what the code was doing.

jpivarski:
> > your git submodules are out of date
> 
> Remind me how to make this go away?

I don't remember the git commands, so I always just clone a new repo.

I just [looked it up](https://stackoverflow.com/a/10916398/1623645): you `cd` into the submodule directory and then `git checkout` using the submodule's commit hash. It will be a detached HEAD; that's fine.

  * dlpack: e1e11e0d555c08bec08a6c7773aa777dfcaae9da
  * pybind11: ffa346860b306c9bbfb341aed9c14c067751feb8
  * rapidjson: f54b0e47a08782a6131cc3d60f94d038fa6e0a51

martindurant:
OK, @jpivarski , this actually runs. It still does not handle _metadata or directory partitioning, but there should be enough here for dask-awkward to use.

(cc @douglasdavis , since we talked about this)

jpivarski:
I'll be looking it over more carefully soon. In the meantime, I just want to make sure that this PR wouldn't change the pybind11 version.

jpivarski (approved):
This looks done to me! @martindurant, is there anything else you want to do in this PR? Otherwise, I'll be merging it.

PR #1399, opened 2022-04-06
This PR adds JAX as a new nplike

sw5sh:


src/awkward/nplike.py
```diff
@@ -22,6 +22,10 @@ def of(*arrays):
                 nplikes.add(ak.nplike.Numpy.instance())
             elif type(array).__module__.startswith("cupy."):
                 nplikes.add(ak.nplike.Cupy.instance())
+            elif type(array).__module__.startswith("jaxlib"):
```
jpivarski: Include a dot, `startswith("jaxlib.")`, so that it can't match a library named `jaxlibwhatever`.

Is it possible for relevant arrays to come from the `jax` library, rather than `jaxlib`? (I don't understand why JAX is split into two libraries like that, especially since the conda-forge recipe is broken: jax depends on a version of jaxlib that raises an exception on startup.)

If it's possible for the array to come from either `jaxlib` or `jax`, you'll need to check both `"jaxlib."` and `"jax."`.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -1 +1,33 @@
...
+
+
+def register_jax():
+    import awkward as ak
+    from awkward._v2._connect.jax.nplike import Jax
+
+    # JAX Specific Configurations
+    jax.config.update("jax_enable_x64", True)
+
+    ak._v2._util._backends["jax"] = Jax
```
jpivarski: It would be better to not need a `register` function. (The `ak._v2.numba.register()` is temporary; it exists because v1 is already registered so v2 has to be opt-in.)

This particular function doesn't do anything that seems to be necessary. Awkward should not be setting any global JAX settings. If the tests need that, then the tests should call `jax.config.update("jax_enable_x64", True)` directly—it shouldn't be hiding inside this innocent-looking function.

Is there a reason that `ak._v2._util._backends` can't be initialized with `Jax` in it? Couldn't it be made possible to declare the `Jax` class without importing `jax`/`jaxlib`? (It's possible for `Cupy`.) What is the `ak._v2._util._backends` needed for, anyway?

tests-jax/test_1399-from-jax.py
```diff
@@ -0,0 +1,50 @@
...
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import cupy as cp  # noqa: F401
+import awkward as ak  # noqa: F401
+import jax
+
+import awkward._v2._connect.jax
+
+awkward._v2._connect.jax.register_jax()
```
jpivarski: See my comment on the definition of this function: we'd much rather do without it!

tests-jax/test_1399-from-jax.py
```diff
@@ -0,0 +1,50 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import pytest  # noqa: F401
+import numpy as np  # noqa: F401
+import cupy as cp  # noqa: F401
+import awkward as ak  # noqa: F401
+import jax
```
jpivarski:
```suggestion

jax = pytest.importorskip("jax")
```

src/awkward/_v2/_connect/jax/nplike.py
```diff
@@ -0,0 +1,170 @@
...
+    def array_equal(self, array1, array2):
+        return self._module.array_equal(array1, array2)
+
+    def repeat(self, array, repeats, axis=None):
+        return self._module.repeat(array, repeats, axis)
+
+    def nan_to_num(self, *args, **kwargs):
+        return self._module.nan_to_num(*args, **kwargs)
+
+    # For all reducers: JAX returns zero-dimensional arrays like CuPy
```
jpivarski: I think it's fine to have code-duplication at this level, as you have it: implementing the reducers separately for `Cupy` and `Jax`.

src/awkward/_v2/_connect/jax/nplike.py
```diff
@@ -0,0 +1,170 @@
...
+        return self._module.asarray(np_array)
+
+    def array_equal(self, array1, array2):
+        return self._module.array_equal(array1, array2)
+
+    def repeat(self, array, repeats, axis=None):
+        return self._module.repeat(array, repeats, axis)
+
+    def nan_to_num(self, *args, **kwargs):
+        return self._module.nan_to_num(*args, **kwargs)
```
jpivarski: Calling `self._module.function_name(*args, **kwargs)` is what the superclass of all the nplikes does, so if you leave it unimplemented here, it will get that default implementation.

The `frombuffer` is a little different: by going through `numpy.frombuffer`, this will only work for JAX arrays on the CPU. Is it possible to test whether a Python buffer (a.k.a. `<memory>`) is in main memory or a GPU? Without importing CuPy?

Wait a minute: you know that `jax.numpy.frombuffer` exists, right?

jpivarski (changes requested):
Thanks for this!

Some large-scale refactoring things:

   * The `Jax` nplike is not much different from the `Cupy` one, and the `Cupy` one is in the `ak.nplike` module. I think the `Jax` one could go here (and it needs to be possible to declare the class without importing `jax`. Whenever you have something that depends on two attributes, like "is an nplike" and "is for JAX", there's always a question of whether to organize them such that all the nplikes are together or to organize them such that all the JAX-related stuff is together. Let's organize it with all nplikes together, because that will enforce that they behave the same way with respect to when the third-party library gets imported. Some nplikes, like the TypeTracer, are in other places, but that's because the TypeTracer is very different from the nplikes for third-party libraries.
   * Modules in `ak._v2.operations.*.*` should each provide one high-level function. See, for instance, `ak_from_parquet` and `ak_metadata_from_parquet`. There's a lot of overlap between these two, but they're in separate files to maintain that structure. So please split `ak_from_lib` back into `ak_from_cupy` and `ak_from_jax` (same for the "to" direction).
   * Sharing code in the _implementations_ of `from_X` and `to_X` is a good idea, and maybe it would be possible to extend this shared code to NumPy as well as CuPy and JAX. The shared implementation sounds important enough to go into `ak._v2._util` as functions named something like "`from_arraylib`" and "`to_arraylib`". Passing in the `module` object is a good idea, too: it could be done in both, the "from" as well as the "to". If it ever becomes necessary to handle special cases for differences between NumPy, CuPy, and JAX, we can check the `module.__name__` to know which one we have (without accidentally triggering an import of one of the others).

src/awkward/nplike.py
sw5sh: It's just `jaxlib.` since arrays come from XLA and `jaxlib.` is a wrapper around that. `jax` is a library that uses `jaxlib` so I think, just checking for `jaxlib.` will be sufficient for `JAX DeviceArrays`

src/awkward/_v2/_connect/jax/__init__.py
sw5sh: Yeah, we don't need a `register` function. I removed that. The `ak._v2._util._backends` is needed to map a user-level name to an nplike so that in the future if any of the backend names need to change we just change the name in the dict and we won't have to touch anything else in the codebase :)

src/awkward/_v2/_connect/jax/nplike.py
sw5sh: Oh I missed that function while writing this PR. I made the correction now.

src/awkward/_v2/_util.py
```diff
@@ -959,3 +960,221 @@ def expand_braces(text, seen=None):
...
+            raise ak._v2._util.error(
+                ValueError(f"cannot convert {array} into {type(module.array([]))}")
+            )
+
+    if module.__name__.startswith("jax") or module.__name__.startswith("cupy"):
+        return _impl(array)
+    elif module.__name__.startswith("numpy"):
+        layout = ak._v2.operations.convert.to_layout(
+            array, allow_record=True, allow_other=True
+        )
```
jpivarski:
```suggestion
    if module == "jax" or module.__name__ == "cupy":
        return _impl(array)
    elif module.__name__ == "numpy":
        layout = ak._v2.operations.convert.to_layout(
            array, allow_record=True, allow_other=True
        )
```

jpivarski (approved):
You'll probably need to merge main to get the Windows fix. But when tests pass, you can merge it!

PR #1401, opened 2022-04-06
Implement `recursively_apply` for `Record`

### LINKED ISSUE #1400: `with_name` fails for `ak._v2.Record`
### Version of Awkward Array

278dccee

```python
import awkward._v2 as ak

ak.Record({'x': 10.0}, with_name="X")
```

agoose77:
Fixes #1400

As per the discussion below, `Record.recursively_apply` is a top-level function that just forwards the `_recursively_apply` call to the content.

### DIFF
--- a/src/awkward/_v2/record.py
+++ b/src/awkward/_v2/record.py
@@ -193,3 +193,61 @@ class Record:
 
     def deep_copy(self):
         return Record(self._array.deep_copy(), copy.deepcopy(self._at))
+
+    def recursively_apply(
+        self,
+        action,
+        depth_context=None,
+        lateral_context=None,
+        keep_parameters=True,
+        numpy_to_regular=True,
+        return_array=True,
+        function_name=None,
+    ):
+        options = {
+            "keep_parameters": keep_parameters,
+            "numpy_to_regular": numpy_to_regular,
+            "return_array": return_array,
+            "function_name": function_name,
+        }
+
+        if return_array:
+
+            def continuation():
+                return Record(
+                    self._array._recursively_apply(
+                        action,
+                        1,
+                        copy.copy(depth_context),
+                        lateral_context,
+                        options,
+                    ),
+                    self._at,
+                )
+
+        else:
+
+            def continuation():
+                self._array._recursively_apply(
+                    action,
+                    1,
+                    copy.copy(depth_context),
+                    lateral_context,
+                    options,
+                )
+
+        result = action(
+            self,
+            depth=1,
+            depth_context=depth_context,
+            lateral_context=lateral_context,
+            continuation=continuation,
+            options=options,
+        )
+
+        if isinstance(result, Record):
+            return result
+        elif result is None:
+            return continuation()
+        else:
+            raise ak._v2._util.error(AssertionError(result))
--- a/src/awkward/_v2/record.py
+++ b/src/awkward/_v2/record.py
@@ -212,42 +212,23 @@ class Record:
         }
 
         if return_array:
-
-            def continuation():
-                return Record(
-                    self._array._recursively_apply(
-                        action,
-                        1,
-                        copy.copy(depth_context),
-                        lateral_context,
-                        options,
-                    ),
-                    self._at,
-                )
-
-        else:
-
-            def continuation():
+            return Record(
                 self._array._recursively_apply(
                     action,
                     1,
-                    copy.copy(depth_context),
+                    depth_context,
                     lateral_context,
                     options,
-                )
+                ),
+                self._at,
+            )
 
-        result = action(
-            self,
-            depth=1,
-            depth_context=depth_context,
-            lateral_context=lateral_context,
-            continuation=continuation,
-            options=options,
+        self._array._recursively_apply(
+            action,
+            1,
+            depth_context,
+            lateral_context,
+            options,
         )
 
-        if isinstance(result, Record):
-            return result
-        elif result is None:
-            return continuation()
-        else:
-            raise ak._v2._util.error(AssertionError(result))
+        return None
--- a/src/awkward/_v2/record.py
+++ b/src/awkward/_v2/record.py
@@ -204,31 +204,29 @@ class Record:
         return_array=True,
         function_name=None,
     ):
-        options = {
-            "keep_parameters": keep_parameters,
-            "numpy_to_regular": numpy_to_regular,
-            "return_array": return_array,
-            "function_name": function_name,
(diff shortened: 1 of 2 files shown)


agoose77:
@ioanaif I've tagged you here because I think this is your area of expertise?

Here's an initial implementation of `recursively_apply` for `ak._v2.Record`.  I think that we want to support the `recursively_apply` interface, because (a) it's useful, and (b) it's currently supported (via `ak._util`) in v1. 

However, whilst 3c79afd nearly passes the simple test that I've added, existing code does not (in many places) expect a non-Content argument in `apply`. So, if you are happy that we want to implement this function for `Record`, I think we'd also need to address how to fix these other functions. 

On this note, @jpivarski my main concern is what to do about the `is_OptionType` etc. flags. Right now, users can pass a `Record` object to a high level function that does not expect a non-Content, and it will throw an Exception if it attempts to lookup these flags. Some `apply` functions check `isinstance(..., Content)`, but I haven't looked into whether that is to handle some other non-content case besides `Record` or not.

<details> 
<summary>Rambling about types</summary>

As we've touched on before, `ak._v2.Record` and `ak._v2.Content` are a bit strange because they are really specialised implementations of a more abstract interface, but we never formally define this via inheritance / protocols (in the same way that we don't for `Form` and `Content`). 

It's funny, I am in general not a fan of strict inheritance hierarchies, but I hadn't realised how much nicer structural subtyping is when there are formal mechanisms to check them (e.g. mypy). I think you already raised the goal of implementing mypy support at a future date, so that's something to look forward to working on. 

</details>

My first instinct here is that `Record` already partially quacks like a `Content`, so we should also implement these `is_` flags. It would then require us to decide whether to add an `is_RecordItem` (terrible name), or leave this type omitted.

jpivarski:
This is my first impression, so take it with that in mind, but I _don't_ think we want Record to look too much like Content. The thing about the Content hierarchy is that any Content subclass can be a `content` or one of the `contents` of any other—not so with Record. (Specifically, I mean `ak._v2.contents.Content` and `ak._v2.record.Record`.) A Record can only ever be at the root of a tree, and the `array` that it contains must be a RecordArray.

As such, the value of getting Record into a `recursively_apply` call is limited: it can avoid a special case in the first step, before descent, but that's it. The value of the Content subclasses having a `recursively_apply` call is much stronger: it's the _only_ way to address the case of Content subclass "XYZArray" at some deep node in the hierarchy. A Record's `recursively_apply` saves one special case, XYZArray's `recursively_apply` saves infinitely many.

It would save that one special case. Implementations of `ak.*` functions that can take either an array or a record get a little simpler by not having to check `isinstance(layout, ak._v2.record.Record)` and doing something with that. Are there a lot of such `ak.*` functions? They can be identified by the fact that they pass `allow_record=True` when they internally call `to_layout`. If they're rare (I think they might be rare), then having _more_ ceremony to call them out might actually be a benefit.

Streamlining is not always a good idea. If we minimize the differences between Content and Record enough, Records might start getting _accidentally_ passed through as Contents in places where we don't want them to be. The symptom of that would be exceptions getting raised downstream of where the actual mistake happens, and those can be hard to trace back.

I can say that some of the connection between low-level Record and Content is historical: in C++, some functions could return arrays or scalars, such as `getitem`. To pass the type-checker, these functions had to return something like `std::variant<numbers, None, Record, Content>`, but the C++ was C++11, not 17. (Also, what's "`numbers`"?) So scalars returned from C++ had to be implemented as Content subclasses even when they logically aren't: numbers are NumpyArray with zero dimensions, a case that has to be forbidden as input to many functions, None is a Content subclass with almost all methods unimplemented, and Record is a subclass of Content, but only on the C++ side. pybind11 makes it out to be not a subclass of Content on the Python side. So, some of the method names being the same were because of this undesired constraint.

-------------------------

It takes so long to write a response that sometimes I'll have a change of heart while writing it. It might be alright for Record to have a `recursively_apply` (to save that one case, mentioned above), but its `recursively_apply` may need to be implemented in an entirely different way. I don't think the `action` function should ever match it. If `action` is going to specially check for the Record case, that case could just as easily be checked outside of `action` and `recursively_apply`.

The main benefit of Record's `recursively_apply` is to get through that one layer, which is always on the root of the layout tree, into the part that the `action` actually cares about. This `recursively_apply` could probably be written as a short-circuit to passing the `action` down to the Record's `array`.

agoose77:
> but I _don't_ think we want Record to look too much like Content.

Ha, this was my first thought too. Then I flip-flopped after looking at v1. 

> As such, the value of getting Record into a recursively_apply call is limited ...

Agreed. It mainly replaces the burden of checking for `isinstance(layout, Record)` with the need to support `Record` and `Content` in the `apply` function...

>  Implementations of ak.* functions that can take either an array or a record get a little simpler by not having to check 

as you put in the this paragraph!

> It might be alright for Record to have a recursively_apply ...  I don't think the action function should ever match it. 

I considered this solution earlier, but I wasn't sure whether this conditional skip would be a bad thing. On balance, I think this *might* be the best solution - the `to_layout` caller determines whether they wish to support `Record`s or not, and the `apply` function is guaranteed only to receive a `Content`.

> It takes so long to write a response that sometimes I'll have a change of heart while writing it.

I'm glad it's not just me. Sometimes I'll write the same answer three or four times before removing half the text in favour of reading the replies!

src/awkward/_v2/record.py
```diff
@@ -193,3 +193,40 @@ def to_list(self, behavior=None):
...
+            action,
+            depth_context,
+            lateral_context,
+            keep_parameters,
+            numpy_to_regular,
+            False,
+            function_name,
+        )
+
+        return None
```
jpivarski: This is minor stuff, but I just thought it would be a little clearer this way. You've added a test, so that will ensure that this is not a broken commit.

```suggestion
        out = self._array.recursively_apply(
            action,
            depth_context,
            lateral_context,
            keep_parameters,
            numpy_to_regular,
            return_array,
            function_name,
        )

        if return_array:
            return Record(out, self._at)
        else:
            return None
```

jpivarski (approved):
Great! I just wanted to tweak it a bit, and I'll use GitHub's suggested change to do that. If it passes the test with the change, then it will be merged. (I'll turn on the auto-merge.)

src/awkward/_v2/record.py
agoose77: Huh that was the kind of thing I usually look for in a PR... thanks for making the change!

PR #1407, opened 2022-04-11
Windows builds stopped working; be looser about directory name.

jpivarski:


jpivarski:
@henryiii FYI: it looks like setuptools (or some similar part of the process) updated and changed the working directory name on Windows, but not the other platforms. The reason I say that is because Windows builds started failing due to not finding a directory, and our print-outs show a different from expected name, on all except Python 3.6. Probably what happened is that the newly updated build tool only has releases for Python 3.7 and up, so that acts as a good indicator.

The `temp-...` directory now has a longer platform name, it includes "`cpython`" in its name, and it has removed the dot between major and minor Python version numbers. If this PR succeeds, I'll merge it into main and all of the affected PRs can merge from main to get it.

jpivarski:
FYI @swishdiff: this is done and merged.

henryiii:
This was an important fix I was pushing for (and it should have affected all systems, not just Windows); before PyPy 3.x would try to reuse CPython 3.x'd cached build (or vice versa) and break all sorts of things. It really started hitting us when PyPy was added to manylinux and cibuildwheel would build all in one job.

henryiii:
And it's in distutils which is vendored into setuptools, so yes.

PR #1410, opened 2022-04-12
Removed bytemask() in favour of mask_as_bool()

ioanaif:


src/awkward/_v2/_util.py
```diff
@@ -1123,7 +1123,7 @@ def _impl(array):
         elif isinstance(array, ak._v2.contents.IndexedOptionArray):
             content = _impl(array.project())
 
-            mask0 = module.asarray(array.bytemask()).view(np.bool_)
+            mask0 = array.mask_as_bool(valid_when=False)
```
jpivarski, line 1126: This last case illustrates why we want to avoid having both `bytemask` and `mask_as_bool`. You didn't see it, @ioanaif, because it was added by @swishdiff, and @swishdiff didn't know that `bytemask` was being removed.

jpivarski (approved):
Great! I did a `grep` and found one more case, and I think I know why that one was there. (See below.)

This is ready to merge!

PR #1413, opened 2022-04-13
replace llvmlite.ir instead of llvmlite.llvmpy.core

### LINKED ISSUE #1411: Use llvmlite.ir instead of llvmlite.llvmpy.core in Numba implementation
### Version of Awkward Array

HEAD

llvmlite, which underlies Numba, has a shim layer that was used to transition away from a previous project called llvmpy. It was never documented, but Awkward's Numba implementation uses a lot of undocumented stuff, and I didn't recognize that `llvmlite.llvmpy` was a shim layer that is going away. Now it's being decremented: numba/llvmlite#826.

Fortunately, it will be easy: there are only two ways we depend on it.

```bash
% fgrep -r llvmpy src/
src/awkward/_connect/_numba/builder.py:        ptr, llvmlite.llvmpy.core.Type.pointer(llvmlite.llvmpy.core.Type.int(8))
src/awkward/behaviors/string.py:    import llvmlite.llvmpy.core
src/awkward/behaviors/string.py:        llvmlite.llvmpy.core.Type.pointer(
src/awkward/behaviors/string.py:            llvmlite.llvmpy.core.Type.int(numba.intp.bitwidth // 8)
src/awkward/_v2/_connect/numba/builder.py:        ptr, llvmlite.llvmpy.core.Type.pointer(llvmlite.llvmpy.core.Type.int(8))
src/awkward/_v2/behaviors/string.py:    import llvmlite.llvmpy.core
src/awkward/_v2/behaviors/string.py:        llvmlite.llvmpy.core.Type.pointer(
src/awkward/_v2/behaviors/string.py:            llvmlite.llvmpy.core.Type.int(numba.intp.bitwidth // 8)
```

All instances of `llvmlite.llvmpy.core.Type.int` can be replaced with `llvmlite.ir.IntType` and all instances of `llvmlite.llvmpy.core.Type.pointer` can be replaced by `llvmlite.ir.PointerType`. See:

```python
>>> import llvmlite.llvmpy.core
>>> import llvmlite.ir
>>> llvmlite.ir.IntType(8) == llvmlite.llvmpy.core.Type.int(8)
True
>>> llvmlite.ir.PointerType(llvmlite.ir.IntType(8)) == llvmlite.llvmpy.core.Type.pointer(llvmlite.llvmpy.core.Type.int(8))
True
```

This will be really easy, and it can be done at any time before Numba drops `llvmlite.llvmpy` (i.e. it will work now; no reason to coordinate with Numba's release, but we do need to do it soon enough). In particular, we need to ensure that there's a (non-rc) 1.x version with this fix.

Ahmad-AlSubaie:
issue #1411 

This is my first PR and my first open source contribution.

### DIFF
--- a/src/awkward/_connect/_numba/builder.py
+++ b/src/awkward/_connect/_numba/builder.py
@@ -21,7 +21,7 @@ def globalstring(context, builder, pyvalue):
         context.add_dynamic_addr(builder, buf.ctypes.data, info=f"str({pyvalue!r})")
     ptr = context.get_constant(numba.types.uintp, dynamic_addrs[pyvalue].ctypes.data)
     return builder.inttoptr(
-        ptr, llvmlite.llvmpy.core.Type.pointer(llvmlite.llvmpy.core.Type.int(8))
+        ptr, llvmlite.ir.PointerType(llvmlite.ir.IntType(8))
     )
 
 
--- a/src/awkward/_connect/_numba/builder.py
+++ b/src/awkward/_connect/_numba/builder.py
@@ -20,9 +20,7 @@ def globalstring(context, builder, pyvalue):
         buf = dynamic_addrs[pyvalue] = numpy.array(pyvalue.encode("utf-8") + b"\x00")
         context.add_dynamic_addr(builder, buf.ctypes.data, info=f"str({pyvalue!r})")
     ptr = context.get_constant(numba.types.uintp, dynamic_addrs[pyvalue].ctypes.data)
-    return builder.inttoptr(
-        ptr, llvmlite.ir.PointerType(llvmlite.ir.IntType(8))
-    )
+    return builder.inttoptr(ptr, llvmlite.ir.PointerType(llvmlite.ir.IntType(8)))
 
 
 class ArrayBuilderType(numba.types.Type):
--- a/src/awkward/_v2/_connect/numba/builder.py
+++ b/src/awkward/_v2/_connect/numba/builder.py
@@ -19,7 +19,7 @@ def globalstring(context, builder, pyvalue):
         context.add_dynamic_addr(builder, buf.ctypes.data, info=f"str({repr(pyvalue)})")
     ptr = context.get_constant(numba.types.uintp, dynamic_addrs[pyvalue].ctypes.data)
     return builder.inttoptr(
-        ptr, llvmlite.llvmpy.core.Type.pointer(llvmlite.llvmpy.core.Type.int(8))
+        ptr, llvmlite.ir.PointerType(llvmlite.ir.IntType(8))
     )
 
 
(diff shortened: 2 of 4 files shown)


jpivarski:
Thank you very much for taking this on!

I see that you've added the import for `llvmlite.ir`, and you've correctly done it only in scopes (function-scope or module-scope, but in your case, just function-scopes) in which we already know that Numba is installed.

What's missing is the replacement of

   * `llvmlite.llvmpy.core.Type.int` → `llvmlite.ir.IntType`
   * `llvmlite.llvmpy.core.Type.pointer` → `llvmlite.ir.PointerType`

This is a good issue to get started on because these can be string-substitutions: the arguments to these functions are the same in both interfaces, so just the fully-qualified names of the functions (i.e. with all the dots) need to be replaced. Also, I can tell you that these functions were never imported using Python's `from` (e.g. `from llvmlite.llvmpy.core import Type`), so you don't need to worry about them appearing in places as "`Type.int`" without qualification.

These are all the files where the old functions appear:

```bash
% fgrep -rl llvmpy src/
src/awkward/_connect/_numba/builder.py
src/awkward/behaviors/string.py
src/awkward/_v2/_connect/numba/builder.py
src/awkward/_v2/behaviors/string.py
```

Although your imports are in the right scope, I don't think they're in exactly the right places, since they should be replacing imports for `llvmlite.llvmpy.core`. Instead of the two additional imports, I think you want to replace these imports:

   * `import llvmlite.llvmpy.core` → `import llvmlite.ir`

Again, string substitution is good enough because they're always fully qualified.

When pre-commit changes the branch, you'll have to `git pull` those changes into your local git repo. Oh wait, it looks like you've done that already.

In fact, you've made a lot of the changes I described above while I was typing them. (I should have just waited!)

Ahmad-AlSubaie:
Glad to be of help. Thank you for the detailed breakdown of the issue. I'm still learning the details of using Git and it's workflow.

`% fgrep -rl llvmpy src/` returns nothing now.

and all
`import llvmlite.llvmpy.core → import llvmlite.ir`

I haven't added any new import calls. Unless something else come up, this issue should be closed?

jpivarski (approved):
Yes, this is perfect!

There shouldn't be any new imports, since each `llvmlite.llvmpy.core` becomes a `llvmlite.ir`. So, no problems there.

I've attached issue #1411 to this PR (in the "Development" box on the right), so when this PR is merged, the issue will automatically be closed.

I've approved this PR, and that means that I'll be merging it when I'm certain that you're not still working on it. It looks to me like you're done. If you just give me a thumbs-up or a short comment saying you're done, I'll press the merge button.

Thanks for contributing!

Ahmad-AlSubaie:
Yup its all done. And thank you for being patient with me. Hopefully I can help contribute more.

jpivarski:
@all-contributors please add @Ahmad-AlSubaie for code

PR #1415, opened 2022-04-13
Passing behaviour in ak._v2 functions

ioanaif:


jpivarski (approved):
Nice! I did a sweep, just to check that every appearance of a `highlevel` variable has a corresponding `behavior` variable, and that all `ak._v2._util.behavior_of` passes in the given `behavior` if it has one.

Along the way, I found out that none of the reducers have `highlevel` or `behavior`. Perhaps they should, but at least this is consistent.

PR #1421, opened 2022-04-14
Enabled string/categorical behavior

ioanaif:


tests/v2/test_0028-add-dressed-types.py
```diff
@@ -75,22 +74,22 @@ def test_string2():
     assert str(ak._v2.operations.describe.type(a[0])) == "3 * uint8"
     assert str(ak._v2.operations.describe.type(a[1])) == "0 * uint8"
     assert str(ak._v2.operations.describe.type(a[2])) == "5 * uint8"
-
+    print(repr(a))
```
jpivarski:
```suggestion
```

Removing print statement. (I would have thought pre-commit would catch this, but maybe it's not applied to the tests.)

jpivarski (approved):
Great! This should merge well with #1427. I made some of the same changes there (without all of the other tests).

As we agreed on Slack, this doesn't mean that categorical is done, but you have some questions for @ianna about `is_unique`.

PR #1435, opened 2022-04-20
Remove duplicated import of to/from-parquet

douglasdavis:
I noticed when using the new `from_parquet` (implemented in `awkward._v2.operations.convert`) that the top level `awkward._v2` namespace imports the parquet read/write functions twice.


douglasdavis:
Perhaps this PR should also remove the `ak_from_parquet.py` and `ak_to_parquet.py` files in the `io` module?

jpivarski:
> Perhaps this PR should also remove the `ak_from_parquet.py` and `ak_to_parquet.py` files in the `io` module?

Yes, that would be helpful, thank you! This PR illustrates the fact that the "io" submodule has to be removed because I forgot and created `from_parquet`/`to_parquet` in two places. When I revamp the `from_json`/`to_json`, I'll move them to "convert" also.

(There's not enough mental distinction between "io" and "convert" to be clear on which should go where. Users don't see it, anyway—everything is just `ak.*`—it was just for our internal organization. Maybe the whole level of hierarchy should be reduced by one, so that all of those `ak_*` submodules are directly in "operations". Almost everything is in "structure", anyway. I may do that at the same time as internal renaming, to get the layout method names consistent with the high-level function names, but that would be a separate PR, timed to not interfere with other work.)

douglasdavis:
Sounds good! this narrow focused PR should be good to go then

jpivarski (approved):
Thanks!

PR #1437, opened 2022-04-21
Enable mixins behavior

ioanaif:


src/awkward/_v2/contents/unionarray.py
```diff
@@ -241,7 +241,7 @@ def _getitem_fields(self, where, only_fields=()):
             self._tags,
             self._index,
             [x._getitem_fields(where, only_fields) for x in self._contents],
-            self._fields_identifer(where),
+            self._fields_identifier(where),
```
jpivarski, line 244: `:)`

src/awkward/_v2/_util.py
```diff
@@ -513,8 +513,6 @@ def custom_ufunc(ufunc, layout, behavior):
     custom = layout.parameter("__array__")
     if not isstr(custom):
         custom = layout.parameter("__record__")
-    if not isstr(custom):
-        custom = layout.purelist_parameter("__record__")
```
jpivarski, line 517: Why was it once necessary to descend through lists to identify a record but now isn't?

I saw that there are some arrays of lists of records in the tests, which I'd think this is relevant for. If any of those tests override ufuncs, which I think they don't.

jpivarski: Well, I tried it, and apparently this is _fixing_ what I was thinking about, rather than breaking it.

In main:

```python
>>> ak._v2.behavior[np.absolute, "Point"] = lambda data: np.sqrt(data.x**2 + data.y**2)
>>> array = ak._v2.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}], with_name="Point")
>>> np.absolute(array)
<Array [1.49, 2.97] type='2 * float64'>
>>> array2 = ak._v2.Array([[{"x": 1.0, "y": 1.1}], [], [{"x": 2.0, "y": 2.2}]], with_name="Point")
>>> np.absolute(array2)
...
  File "/home/jpivarski/mambaforge/lib/python3.9/site-packages/awkward/_v2/_connect/numpy.py", line 109, in _array_ufunc_adjust_apply
    out = apply_ufunc(ufunc, method, nextinputs, kwargs)
TypeError: <lambda>() takes 1 positional argument but 4 were given
```

but in this branch:

```python
>>> ak._v2.behavior[np.absolute, "Point"] = lambda data: np.sqrt(data.x**2 + data.y**2)
>>> array = ak._v2.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}], with_name="Point")
>>> np.absolute(array)
<Array [1.49, 2.97] type='2 * float64'>
>>> array2 = ak._v2.Array([[{"x": 1.0, "y": 1.1}], [], [{"x": 2.0, "y": 2.2}]], with_name="Point")
>>> np.absolute(array2)
<Array [[1.49], [], [2.97]] type='3 * var * float64'>
```

So this must be the right thing to do. The thing I was worried about was it not finding the overloaded record when it's buried inside of a list like that.

But, uh-oh: there's another problem—if the array _does not_ have the `with_name="Point"`, ufuncs are not supposed to work on records. In v1:

```python
>>> array = ak.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}])
>>> np.absolute(array)
...
  File "/home/jpivarski/irishep/awkward-1.0/awkward/_util.py", line 1058, in apply
    raise ValueError(
ValueError: cannot broadcast records in this type of operation

(https://github.com/scikit-hep/awkward-1.0/blob/1.9.0rc2/src/awkward/_util.py#L1060)
```

(that's good), but in v2:

```python
>>> array = ak._v2.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}])
>>> np.absolute(array)
<Array [{x: 1, y: 1.1}, {x: 2, ...}] type='2 * {x: float64, y: float64}'>
```

(that's bad). But I tried this in main as well as this branch and it's the same in both, so it's unrelated to this change. I'll make it a new issue and prioritize it.

jpivarski (approved):
Great! I found an issue, but it's not an issue with this PR.

PR #1442, opened 2022-04-25
from_parquet: simplify and add cases as todo

martindurant:
Cutting out some unnecessary stuff based on our conversation, as well as listing all the file handling cases we should support.

### DIFF
--- a/src/awkward/_v2/operations/convert/ak_from_parquet.py
+++ b/src/awkward/_v2/operations/convert/ak_from_parquet.py
@@ -70,9 +70,6 @@ def from_parquet(
             storage_options,
             row_groups,
             columns,
-            max_gap,
-            max_block,
-            footer_sample_size,
         )
         return _load(
             actual_paths,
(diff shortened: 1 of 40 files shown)


src/awkward/_v2/operations/convert/ak_from_parquet.py
```diff
@@ -70,9 +70,6 @@ def from_parquet(
             storage_options,
             row_groups,
             columns,
-            max_gap,
-            max_block,
-            footer_sample_size,
```
jpivarski, line 75: Users will sometimes get a warning that the `footer_sample_size` is too small. That's fine, but it would be annoying if there is no way to set the `footer_sample_size` and silence the warning.

martindurant: This is not used, because fsspec.parquet doesn't offer any particular speedup for just reading the footer. The code just uses fs.open now.

jpivarski:
Is this superseded by #1476?

jpivarski:
Thanks!

PR #1444, opened 2022-04-26
Enable ak.singletons

ioanaif:


jpivarski:
To implement an axis, maybe use `:, :, :, np.newaxis` (if nonnegative) or `..., np.newaxis, :, :, :` (if negative) and then `drop_none`?

As for whether singletons should include Nones, I think it shouldn't, because otherwise it would be completely identical to the above slide (and it wouldn't be doing what singletons was invented to do: to switch between a token-based and a 0/1-length list-based representation of missing data).

jpivarski:
`drop_none` is #832.

jpivarski (approved):
Approving, but there needs to be another PR that adds an `axis` parameter to the new `singletons`. That can be implemented another PR, especially if you're going the route of doing it through `drop_none`.

Issue #1105, opened 2021-09-23
ak.array_equal to override np.array_equal


martindurant:
### Description of new feature

It would be nice to have convenience functions for asserting the equivalence of ak arrays.

We can compare an array's:
- values
- type (i.e., datashape description)
- form (including lengths of sub structures, if known)

Currently, tests tend to compare via `.tolist()` (resulting in python lists and dicts) which can compare the array values. The result of ak.type is already comparable. There are, however, well-defined rules of whether an array can be considered equal as a whole to another array.

Note that this is not the same as `__eq__`, which is an element-wise operation.

jpivarski:
And one more level: layouts can differ even when the Form is known (a `content` can be longer than than it needs to be, but such "hidden" elements are not considered to impact the equality of two arrays).

For example, these are to be considered the same, even though they have different NumpyArrays (the first counts up to 5 and the second counts up to 10):

```python
>>> ak.Array(ak.layout.ListOffsetArray64(
...     ak.layout.Index64(np.array([0, 3, 3, 5])),
...     ak.layout.NumpyArray(np.arange(5) * 1.1)))
<Array [[0, 1.1, 2.2], [], [3.3, 4.4]] type='3 * var * float64'>

>>> ak.Array(ak.layout.ListOffsetArray64(
...     ak.layout.Index64(np.array([0, 3, 3, 5])),
...     ak.layout.NumpyArray(np.arange(10) * 1.1)))
<Array [[0, 1.1, 2.2], [], [3.3, 4.4]] type='3 * var * float64'>
```

Both of these arrays have the same Type (`3 * var * float64`) and the same Form (`{"class": "ListOffsetArray64", "offsets": "i64", "content": "float64"}`), but they have different buffers. If you ran them through [ak.to_buffers](https://awkward-array.readthedocs.io/en/latest/_auto/ak.to_buffers.html) and compared the results, those results would be different from each other.

The "lists of rules" of what is considered a valid array are given in the [documentation of each Content subclass](https://awkward-array.readthedocs.io/en/latest/ak.layout.Content.html), in the constructor of each Pythonic illustration. These rules are also checked for a given array in the implementation of [ak.is_valid](https://awkward-array.readthedocs.io/en/latest/_auto/ak.is_valid.html). While there are these explicit lists of what is valid, there aren't any explicit lists of what is equal, though two arrays that have the same `.tolist()` representation and the same type _and are both valid_ are equal to each other.

ivirshup:
[...]

PR #1447, opened 2022-04-27
This PR attempts to add autodifferentiation support for Awkward Arrays using JAX pytrees.

sw5sh:


src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -16,8 +18,196 @@
...
+    jax.tree_util.register_pytree_node(
+        ak._v2.highlevel.Record,
+        ak._v2.highlevel.Record.jax_flatten,
+        ak._v2.highlevel.Record.jax_unflatten,
+    )
+
 
 def import_jax(name="Awkward Arrays with JAX"):
     if jax is None:
         raise ImportError(error_message.format(name))
```
jpivarski: Since we're all Python 3 now, we should be raising `ModuleNotFoundError`, rather than its superclass, `ImportError`. If there are other cases of raising `ImportError` (which you used as a model for this one), then we should change that one to `ModuleNotFoundError`, too.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -16,8 +18,196 @@
...
+
+def _find_numpyarray_nodes(layout):
+
+    data_ptrs = []
+
+    def find_nparray_ptrs(node, **kwargs):
+        if isinstance(node, ak._v2.contents.numpyarray.NumpyArray):
+            data_ptrs.append(node.data)
+
+    layout.recursively_apply(action=find_nparray_ptrs, return_array=False)
```
jpivarski, line 73: This works, and it's fine, but just to let you know: `lateral_context` does the same thing. The fact that this enclosed variable (`data_ptrs`) is a single object mutated in place is what makes it "lateral", as opposed to "depth".

No action item, though: nothing needs to be changed here. You can just resolve this when you read it.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -16,8 +18,196 @@
...
+                            for i in range(len(self_layout.contents))
+                        ]
+                    )
+                )
+            elif isinstance(self_layout, ak._v2.contents.NumpyArray):
+                return len(self_layout) == len(other_layout)
+            elif isinstance(self_layout, ak._v2.index.Index):
+                return self_layout.nplike.array_equal(
+                    self_layout.data, other_layout.data
+                )
```
jpivarski: This is okay because `np.array_equal` allows the Indexes to have different dtypes. (I didn't know that, but I just tested it.)

It reveals what a problem it's going to be to have `Index.data` use NumPy functions and `NumpyArray.data` use JAX functions, though: nearly every use of `nplike` would have to be `index_nplike`. We'll have to brainstorm together on what would be a good strategy to transition that.

src/awkward/_v2/contents/regulararray.py
```diff
@@ -44,7 +44,7 @@ def __init__(
             else:
                 size = int(size)
         if not isinstance(zeros_length, ak._v2._typetracer.UnknownLengthType):
-            if not (ak._util.isint(zeros_length) and zeros_length >= 0):
+            if not (ak._v2._util.isint(zeros_length) and zeros_length >= 0):
```
jpivarski, line 47: Good catch! (This and the above.)

src/awkward/_v2/highlevel.py
```diff
@@ -243,6 +243,12 @@ def __init__(
         if with_name is not None:
             layout = ak._v2.operations.with_name(layout, with_name, highlevel=False)
 
+        if (
+            backend is not None
+            and ak._v2._util.regularize_backend(backend) is ak.nplike.Jax.instance()
+        ):
+            layout = ak._v2.packed(layout, highlevel=False)
```
jpivarski: The reason for packing was to ensure that a primal and a tangent have the same array tree (apart from NumpyArray leaves). Packing every array is built with `backend="jax"` as an argument to the `ak._v2.Array` constructor is too much, and it also wouldn't catch cases in which the array is constructed with `backend=None` and then explicitly converted by the user using `ak._v2.to_backend`. So it shouldn't be done here because it catches too many cases and doesn't catch all the cases.

Is there a way to ensure that it only happens when a user supplies both a primal and a tangent? That only happens for one of `jax.jvp` or `jax.vjp`, right? It doesn't happen at all for `jax.grad`.

If there's no way for us to capture when users supply a tangent vs when it's created by us or JAX, then we could just not apply `packed` and consider it a rule that the tangent must have exactly the same detailed structure (Form and offsets) as the primal. We do _check_ it with `is_layout_equal`.

jpivarski: Oh, and we check for nplike with `isinstance(x, NplikeType)`, not `is NplikeType.instance()`.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -16,8 +18,196 @@
...
+    jax.tree_util.register_pytree_node(
+        ak._v2.contents.bitmaskedarray.BitMaskedArray,
+        ak._v2.contents.bitmaskedarray.BitMaskedArray.jax_flatten,
+        ak._v2.contents.bitmaskedarray.BitMaskedArray.jax_unflatten,
+    )
+    jax.tree_util.register_pytree_node(
+        ak._v2.contents.bytemaskedarray.ByteMaskedArray,
+        ak._v2.contents.bytemaskedarray.ByteMaskedArray.jax_flatten,
+        ak._v2.contents.bytemaskedarray.ByteMaskedArray.jax_unflatten,
+    )
```
jpivarski: These _could_ have been more succinct with something like

```python
for cls in [
    ak._v2.contents.bitmaskedarray.BitMaskedArray,
    ak._v2.contents.bytemaskedarray.ByteMaskedArray,
    ...
]:
    jax.tree_util.register_pytree_node(cls, cls.jax_flatten, cls.jax_unflatten)
```

with special cases for the high-level `ak._v2.Array`, `ak._v2.Record` and for the layout `ak._v2.record.Record`.

I don't feel strongly about this, since all of the Content subclasses have to be named somewhere, but if it's a list in a for loop, there's less repetition.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -16,8 +18,196 @@
...
+    jax.tree_util.register_pytree_node(
+        ak._v2.highlevel.Array,
+        ak._v2.highlevel.Array.jax_flatten,
+        ak._v2.highlevel.Array.jax_unflatten,
+    )
+    jax.tree_util.register_pytree_node(
+        ak._v2.record.Record,
+        ak._v2.record.Record.jax_flatten,
+        ak._v2.record.Record.jax_unflatten,
+    )
```
jpivarski: It's important to not let `jax_flatten` and `jax_unflatten` be public on the high-level `ak._v2.Array` and `ak._v2.Record`. That high-level namespace has to be kept uncluttered. Having this in the mid-level Content subclasses and `ak._v2.record.Record` is fine, but for consistency, let's just make them all private: `def _jax_flatten`, `def _jax_unflatten`.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -16,8 +18,196 @@
...
+class AuxData:
+    def __init__(self, layout):
+        self._layout = layout
+
+    @property
+    def layout(self):
+        return self._layout
+
+    def __eq__(self, other):
+        def is_layout_eq(self_layout, other_layout):
```
jpivarski: This should be a method on Content subclasses, Index, and `ak._v2.record.Record`, so that you don't need the `isinstance` chain. We've been trying to make every `isinstance` chain that _can_ be a method into a method. (It _can't_ be for `broadcast_and_apply`, for instance.)

Actually, this is related to #1105, which is asking for different kinds of equality. What you want here is two steps removed from very strict equality, so if you define a slightly more general function that includes absolutely strict equality:

```python
def layout_equal(self, other, index_dtype=True, numpyarray=True):
    ...
```

you can then use it in the `numpyarray=False` case. When `numpyarray=True`, it would require all nodes to have exactly the same properties, but when `numpyarray=False`, it would ignore NumpyArray leaves. The `index_dtype` is something only `Index.layout_equal` checks, but if True, the Indexes should have the same dtype; if False, they can have different dtypes. You want `index_dtype=False`. All `True` values are strict equality, but you want to be two steps removed from that.

Also, it's important for this function (including your case!) to assert equality of all properties, including Bit/ByteMaskedArray `valid_when`, RegularArray `size`, identifiers, parameters, etc. There's a [_parameters_equal](https://github.com/scikit-hep/awkward/blob/405c4d492298d262d7278d109b10891aa95448b1/src/awkward/_v2/forms/form.py#L171) function you can use (with `only_array_record=False`) to ignore the `None` case. There isn't an equivalent `_identifiers_equal`, but you can add one [here](https://github.com/scikit-hep/awkward/blob/main/src/awkward/_v2/identifier.py).

```python
def _identifiers_equal(one, two):
    if one is None and two is None:
        return True
    elif isinstance(one, Identifier) and isinstance(two, Identifier):
        return one.layout_equal(two, index_dtype=True)
    else:
        return False

class Identifier:
    ...

    def layout_equal(self, other, index_dtype=True):
        if self._ref != other._ref or self._fieldloc != other._fieldloc:
            return False
        if index_dtype and self._data.dtype != other._data.dtype:
            return False
        return self._nplike.array_equal(self._data, other._data)
```

src/awkward/_v2/_util.py
```diff
@@ -1168,6 +1168,24 @@ def _impl(array):
         else:
             return module.asarray(array)
     else:
-        ak._v2._util.error(
+        raise ak._v2._util.error(
```
jpivarski, line 1171: Good catch!

jpivarski (changes requested):
This is a big step forward!

The process of making `isinstance(nplike, Jax)` cases use `Jax.index_nplike` for all Index manipulating code and JAX with its tracers for the NumpyArray manipulating code is something we'll have to brainstorm, but it will be after this PR.

The pybind11 submodule will have to be brought in line with `main` before merging (that is, this PR should not try to change pybind11), and if you're having trouble with that, I'll try to remember how to do it.

Beyond that, there are some things to be refactored, none of them major, described inline.

src/awkward/_v2/highlevel.py
sw5sh: I haven't come across a mechanism that would allow us to intercept JAX during it's operation. JAX however will complain as soon as the primals and tangents don't match early on during the execution.

jpivarski: Okay, then we shouldn't pack it. Let the user get errors.

src/awkward/_v2/_connect/cuda/__init__.py
```diff
@@ -141,7 +141,7 @@ def __init__(self, name, error_context):
 
 def import_cupy(name="Awkward Arrays with CUDA"):
     if cupy is None:
-        raise ImportError(error_message.format(name))
+        raise ModuleNotFoundError(error_message.format(name))  # noqa: AK101
```
jpivarski: Use `ak._v2._util.error`

src/awkward/_v2/_connect/cuda/__init__.py
```diff
@@ -187,7 +187,9 @@ def initialize_cuda_kernels(cupy):
 
         return kernel
     else:
-        raise ImportError(error_message.format("Awkward Arrays with CUDA"))
+        raise ModuleNotFoundError(  # noqa: AK101
+            error_message.format("Awkward Arrays with CUDA")
+        )
```
jpivarski: Same.

src/awkward/_v2/_connect/jax/__init__.py
```diff
@@ -22,86 +22,37 @@
...
+            cls,
+            cls._jax_flatten,
+            cls._jax_unflatten,
+        )
 
 
 def import_jax(name="Awkward Arrays with JAX"):
     if jax is None:
-        raise ImportError(error_message.format(name))
+        raise ModuleNotFoundError(error_message.format(name))  # noqa: AK101
```
jpivarski: Same.

jpivarski (approved):
Only thing: you don't need to skip AK101.

jpivarski:
Do you have an "Enable auto-merge (squash)" button?

image

If so, you can press that and not have to watch it run the tests. If not, I'll press it and try to figure out why you don't see it.

jpivarski:
I just enabled it.

PR #1468, opened 2022-05-13
Rename low-level methods to match high-level function names.

jpivarski:
There were fewer of these than I thought there was going to be.

  * localindex → local_index
  * validityerror → validity_error
  * rpad → pad_none
  * withparameter → with_parameter

I changed mid-level names (i.e. on the layouts) because that's a public interface for downstream developers. I didn't change any kernel function names because those are totally internal and can be changed at any time. Also, those wouldn't be across-the-board changes that affect mergeability. Also, also, the kernel names aren't very good as they are, but they're defined in terms of their function (the Python snippets in kernel-signatures.yml), rather than a meaningful name.


jpivarski:
There were fewer of those than I thought there was going to be.

PR #1472, opened 2022-05-14
start working on growable buffer

ManasviGoyal:


ManasviGoyal:
Hi @ianna! I have added a standalone test for Panel implementation using the vector approach. For `getitem_at_nowrap()`, I have assumed that the size of all panels is same and equal to the number reserved initially. But I have also added the function for the case where the size of each panel is different (it is commented out). There is some issue with the `concatenate()` function and it is not able to access `ptr_` and stops after copying (I think it might be because of the default deleter, not sure). Please check.

studies/growable_buffer/growable_buffer_panels.cpp
```diff
@@ -8,54 +8,54 @@
 template <typename PRIMITIVE>
 class GrowableBuffer {
 public:
-  GrowableBuffer(size_t initial) 
-      : initial_(initial) { 
+  GrowableBuffer(size_t initial)
+      : initial_(initial) {
     initial_ = initial;
-    ptr_.push_back(std::make_unique<PRIMITIVE>(initial));
+    ptr_.push_back(std::unique_ptr<PRIMITIVE>(new PRIMITIVE[initial]));
```
ianna, line 14: @ManasviGoyal - please, check if the proposed changes work for you. Thanks!

ianna (commented):
The changes to CMakeLists.txt is a temporary one - just to make sure it builds on my computer. The tests run.

studies/growable_buffer/growable_buffer_panels.cpp
ManasviGoyal: Hi @ianna. The concatenate function still doesn't work but otherwise the changes work for me. Does it work for you?
Also, should I add any other methods from the original GrowableBuffer?

ianna: > Hi @ianna. The concatenate function still doesn't work but otherwise the changes work for me. Does it work for you?

Yes, please, have a look at the last commit. Also, IMHO we have to move out the index calculations per item. It would be better to make the buffer contiguous first.

> Also, should I add any other methods from the original GrowableBuffer?

I think, you need one more method to update the `ptr_` with a contiguous panel.

ManasviGoyal: > Yes, please, have a look at the last commit. Also, IMHO we have to move out the index calculations per item. It would be better to make the buffer contiguous first.

Okay. So will the concatenation be done externally like it is done here in the `main()` or should I just add it inside the `getitem_at_nowrap()`?

> I think, you need one more method to update the `ptr_` with a contiguous panel.

But that can be done by calling `concatenate()` inside the `ptr()` method that we already have just like you suggested earlier. So I don't think we need a separate method for that.

jpivarski:
Same question: is this superseded by #1494? I want to know which PRs to pay attention to because they're intended to be merged.

ManasviGoyal:
> Same question: is this superseded by #1494? I want to know which PRs to pay attention to because they're intended to be merged.

Yes. It is no longer needed since we switched to the header only implementation. I'll close this too.

PR #1476, opened 2022-05-18
parquet redux

martindurant:
cc @douglasdavis


jpivarski:
AK101: `raise Exception` → raise `ak._v2._util.error(Exception)`

martindurant:
> AK101: raise Exception → raise ak._v2._util.error(Exception)

Not too worried about the lint for the moment :)

More worried about [this kind of fudge](https://github.com/scikit-hep/awkward/pull/1476/files#diff-847463b485f259f948720b1a228edea9ec8757bd4d7a1a7aa4048b092ffda326R236).

martindurant:
@jpivarski , ready for you to have a look at this, so we can discuss.

Two outstanding issues I haven't addressed:
- what happens when you have one Record from each row-group. I suppose this is an array, but exactly how to concat is different
- I am not checking the column names reported in the metadata, as I'm not convinced that the writing arm is producing the right thing. We should come up with concrete tests to make certain.

jpivarski (approved):
In addition to the CI tests, I also tested it with some big files locally, and varied the options. All is good.

Reading the code itself, everything looks good to me!

I'll set this to auto-merge if all tests pass.

PR #1484, opened 2022-05-31
LayoutBuilder migration to v2

ianna:


ianna:
@jpivarski - Please, have a look when you have time. This PR completes migration of LayoutBuilder to v2. Although it is not intended to be a long term solution, I think, it is useful to have it in the main branch. We want to compare performance of the old implementation with the new one. The PR implements a high level interface to a current LayoutBuilder that we will swap with the one in development. Thanks!

jpivarski (approved):
> We want to compare performance of the old implementation with the new one. The PR implements a high level interface to a current LayoutBuilder that we will swap with the one in development.

Okay, that makes sense, though it would also work to do the performance study before v1 is removed. But more flexibility is better than less.

It's a good update, so I'll merge it!

Issue #503, opened 2020-10-30
Move libawkward.so (etc) into its own package


jpivarski:

henryiii:
Let's do this together, and we can use it as the point where we migrate to cibuildwheel. I expect together we'll be faster than if we did it separately.

jpivarski:
We can share screen in Zoom, but we might want to also use Atom/TeleType or the VSCode equivalent you use. (I haven't used VSCode before.) Otherwise, we could pass suggestions back and forth on a Google Doc, while only one of us is entering changes.

henryiii:
Yes, this will be more complicated than it was at the beginning of the year...

jpivarski:
@agoose77 This is the issue I was thinking of.

To catch up @henryiii on something we talked about in the Awkward-Dask meeting: @martindurant is thinking about a suite of string functions, like everything in `np.char.*` but with UTF-8 capabilities, vectorized across Awkward Arrays. (This might be a killer app for Awkward in industry.) The short path to implementing this is through Rust because Rust has a good UTF-8 library, in which questions like "what does it mean to capitalize?" have already been thought through. and it would be higher performance than multi-pass NumPy-based implementations [like this equality operation](https://github.com/scikit-hep/awkward-1.0/blob/0b0fb3f1c18b8acf5ab664ffefe513f67c82076b/src/awkward/behaviors/string.py#L124-L149).

Strings are a high-level `ak.behavior` in Awkward Array, but foundational ones that should included when one does `pip install awkward` (unlike, say, Lorentz vectors, for which there's `pip install vector awkward`). Rust complicates the build procedure, and a clean way of avoiding that would be to put them in a separate PyPI (and conda-forge) package: `awkward-string-kernels`. Since strings are foundational, `awkward` would _strictly depend_ on `awkward-string-kernels`.

Naturally, that reminded me of this project, of externalizing the compiled code into a package that `awkward` strictly depends on. An advantage that @martindurant brought up, which I hadn't thought of, is that any potential contributor who doesn't need to touch kernels or C++ would get those binaries from PyPI/conda-forge and be able to operate on a pure Python `awkward` codebase. With v2, that includes a much larger fraction of the codebase.

As an overview, the eventual project would involve the following pieces, which could all be separate packages:

   * `awkward-string-kernels`, as described above; probably a Rust-compiled shared object library with a ctypes interface that a Python part interacts with: that Python provides behaviors for `ak.beahviors`.
[...]

PR #1485, opened 2022-06-01
Add the beginning of an example of templated LayoutBuilder.

jpivarski:
This is for @ianna and @ManasviGoyal. (It's what I showed in our meeting.)

### DIFF
--- a/studies/templated-LayoutBuilder/LayoutBuilder.h
+++ b/studies/templated-LayoutBuilder/LayoutBuilder.h
@@ -1,0 +1,80 @@
+// only needed for 'dump', which is a temporary debugging tool
+#include <iostream>
+#include <string>
+
+// really needed
+#include <stdint.h>
+#include <vector>
+
+template <typename PRIMITIVE>
+class GrowableBuffer {
+public:
+  void append(PRIMITIVE datum) {
+    data_.push_back(datum);
+  }
+
+  int64_t length() const {
+    return data_.size();
+  }
+
+  void dump() const {
+    for (auto x : data_) {
+      std::cout << x << " ";
+    }
+  }
+
+private:
+  std::vector<PRIMITIVE> data_;
+};
+
+
+template <typename PRIMITIVE>
+class NumpyLayoutBuilder {
+public:
+  void append(PRIMITIVE x) {
+    data_.append(x);
+  }
+
+  int64_t length() const {
+    return data_.length();
+  }
+
+  void dump(std::string indent) const {
+    std::cout << indent << "NumpyLayoutBuilder" << std::endl;
+    std::cout << indent << "    data ";
+    data_.dump();
+    std::cout << std::endl;
+  }
+
+private:
+  GrowableBuffer<PRIMITIVE> data_;
+};
+
+
+template <typename INDEX, typename BUILDER>
+class ListOffsetLayoutBuilder {
+public:
+  ListOffsetLayoutBuilder() {
+    offsets_.append(0);
+  }
+
+  BUILDER* begin_list() {
+    return &content_;
+  }
+
+  void end_list() {
+    offsets_.append(content_.length());
+  }
+
+  void dump(std::string indent) const {
+    std::cout << indent << "ListOffsetLayoutBuilder" << std::endl;
+    std::cout << indent << "    offsets ";
+    offsets_.dump();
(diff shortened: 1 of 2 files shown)


ianna:
@jpivarski - shall the `GrowableBuffer.h` be placed in `src/awkward/` or `src/awkward/growable_buffer/` directory?

ianna:
@jpivarski - shall the `GrowableBuffer.h` be placed in `src/awkward/` or `src/awkward/growable_buffer/` directory?

ianna:
@jpivarski - shall the `GrowableBuffer.h` be placed in `src/awkward/` or `src/awkward/growable_buffer/` directory?

jpivarski:
How about

  * `src/awkward/cpp-headers/GrowableBuffer.h`
  * `src/awkward/cpp-headers/LayoutBuilder.h`

The use of a hyphen would be a bad idea if this would ever contain Python code (it won't). It should be a generic name, to allow for other header-only additions in the future. Maybe `include` would make sense, but we already have an `include/awkward` at top-level, and I think having two of them with similar function (containing C++ header files) would be confusing. Maybe all of the headers should eventually go into here, but only if the C++ code itself becomes header-only.

After dropping v1, we could entertain the notion of converting the remaining C++ code into header-only and having the .cpp part be nothing more than template instantiations and pybind11 wrappers. The reason we _didn't_ do that before is because it increases the compilation time (with localbuild.py, we can edit .cpp files and only recompile what we change, not the whole codebase), but that becomes less relevant when the v1 C++ code is dropped (see https://github.com/scikit-hep/awkward/issues/503#issuecomment-1106750209).

ianna:
> How about
> 
> * `src/awkward/cpp-headers/GrowableBuffer.h`
> * `src/awkward/cpp-headers/LayoutBuilder.h`
> 
> The use of a hyphen would be a bad idea if this would ever contain Python code (it won't). It should be a generic name, to allow for other header-only additions in the future. Maybe `include` would make sense, but we already have an `include/awkward` at top-level, and I think having two of them with similar function (containing C++ header files) would be confusing. Maybe all of the headers should eventually go into here, but only if the C++ code itself becomes header-only.
> 
> After dropping v1, we could entertain the notion of converting the remaining C++ code into header-only and having the .cpp part be nothing more than template instantiations and pybind11 wrappers. The reason we _didn't_ do that before is because it increases the compilation time (with localbuild.py, we can edit .cpp files and only recompile what we change, not the whole codebase), but that becomes less relevant when the v1 C++ code is dropped (see [#503 (comment)](https://github.com/scikit-hep/awkward/issues/503#issuecomment-1106750209)).

How about 
 * `src/awkward/interface/GrowableBuffer.h`
* `src/awkward/interface/LayoutBuilder.h`

jpivarski:
What's the reasoning behind the word "interface"?

Traditionally, .h files only describe interfaces, with the implementations in .cpp files, but the idea of a header-only library is that the implementation is included in the .h file (for downstream convenience). Since this directory is only for header-only libraries (which have to be shipped in the Python module, so that they can be used by Cling at runtime), this traditional use of "interface" for .h files would be misleading. It's the opposite of that: they're only mixing interface with implementation.

Were you thinking of "interface" in a different context?

ianna:
> What's the reasoning behind the word "interface"?
> 
> Traditionally, .h files only describe interfaces, with the implementations in .cpp files, but the idea of a header-only library is that the implementation is included in the .h file (for downstream convenience). Since this directory is only for header-only libraries (which have to be shipped in the Python module, so that they can be used by Cling at runtime), this traditional use of "interface" for .h files would be misleading. It's the opposite of that: they're only mixing interface with implementation.
> 
> Were you thinking of "interface" in a different context?

I think, I'm influenced by CMSSW. Originally, it seemed odd to use `interface` directory, but, we do include an interface to...
```c++
#include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
```

jpivarski:
> ```c++
> #include "DataFormats/TrackerCommon/interface/TrackerTopology.h"
> ```

Those really are interfaces: the implementation of TrackerTopology is in a different file. This directory is to contain only header-only code; that's why it needs to be in this directory instead of `include/awkward`.

jpivarski:
The tests got stuck, but these are just new files in the `studies` directory—the tests are irrelevant.

PR #1487, opened 2022-06-02
Adding repr overriden behavior

ioanaif:


src/awkward/_v2/highlevel.py
```diff
@@ -2312,7 +2315,7 @@ def snapshot(self):
         """
         formstr, length, container = self._layout.to_buffers()
         form = ak._v2.forms.from_json(formstr)
-        return ak._v2.operations.from_buffers(form, length, container)
+        return ak._v2.operations.from_buffers(form, length, container, highlevel=True)
```
jpivarski, line 2318: `highlevel=True` by default (or at least it should be, for all functions that define a `highlevel` parameter).

src/awkward/_v2/operations/ak_type.py
```diff
@@ -109,7 +109,7 @@ def _impl(array):
                 out = ak._v2.types.RegularType(out, x)
             return ak._v2.types.ArrayType(out, array.shape[0])
 
-    elif isinstance(array, ak.layout.ArrayBuilder):
+    elif isinstance(array, ak._v2.highlevel.ArrayBuilder):
```
jpivarski, line 112: Yikes! That's a good catch.

src/awkward/_v2/types/recordtype.py
```diff
@@ -148,6 +148,14 @@ def _str(self, indent, compact):
...
+        if (
+            self._typestr != "override"
+            and self._parameters is not None
+            and len(self._parameters) == 1
+            and "__record__" in self._parameters
+        ):
+            args = dict(zip(self._fields, [x.primitive for x in self._contents]))
+            return "{}[{}]".format(self._parameters["__record__"], str(args)[1:-1])
         args = [repr(self._contents), repr(self._fields)] + self._repr_args()
         return "{}({})".format(type(self).__name__, ", ".join(args))
```
jpivarski: This is a style thing, but I think it's better to be explicit about the if-else structure. Early returns are better when the rest of the code in the function is much longer than the early return.

```suggestion
            args = dict(zip(self._fields, [x.primitive for x in self._contents]))
            return "{}[{}]".format(self._parameters["__record__"], str(args)[1:-1])
        else:
            args = [repr(self._contents), repr(self._fields)] + self._repr_args()
            return "{}({})".format(type(self).__name__, ", ".join(args))
```

src/awkward/_v2/_prettyprint.py
```diff
@@ -84,11 +84,19 @@ def valuestr_horiz(data, limit_cols):
...
-                    )
+                    cols_taken, strs = valuestr_horiz(current, limit_cols - for_comma)
+
+                    if (
+                        ak._v2.highlevel.Record in current.__class__.__bases__
+                        or ak._v2.highlevel.Array in current.__class__.__bases__
+                        and not type(current).__repr__
+                        is ak._v2.highlevel.Array.__repr__
+                    ):
+                        strs = type(current).__repr__(current)
```
jpivarski, line 98: Great: this is the key thing.

src/awkward/_v2/highlevel.py
```diff
@@ -1141,7 +1141,10 @@ def __repr__(self):
     def _repr(self, limit_cols):
         import awkward._v2._prettyprint
 
-        pytype = type(self).__name__
+        try:
+            pytype = super().__getattribute__("__name__")
+        except AttributeError:
+            pytype = type(self).__name__
```
jpivarski, line 1147: This, too.

jpivarski (approved):
Fantastic, thanks! I have one suggested change (stylistic) that I'm applying and then enabling auto-merge.

PR #1490, opened 2022-06-06
Refactoring to include index_nplike and reducers

sw5sh:


jpivarski (approved):
I like this approach!

agoose77:
I'm a :+1: on this as well, it will be good to formally distinguish between index and content nplikes. Minor thought - is it worth using a slightly shorter name like `index` instead of `index_nplike`? It would make the code slightly more readable, and I think it's safe enough given that there is no `np.index`.

jpivarski:
> I'm a +1 on this as well, it will be good to formally distinguish between index and content nplikes. Minor thought - is it worth using a slightly shorter name like `index` instead of `index_nplike`? It would make the code slightly more readable, and I think it's safe enough given that there is no `np.index`.

We could, although I think that would be easy to do at any time; nplikes are rather internal. Perhaps the `nplike` module ought to be made private (renamed to `_nplike`) so to _formally_ say that we have this flexibility.

At this time, though, @swishdiff needs to finish his project (it's the last week!) and we can consider renaming later.

"`index`" might be a little understated, though. I would have trouble guessing what it means. Something like "`for_index`" would be more helpful, but that's getting back to the original length.

agoose77:
> "`index`" might be a little understated, though. I would have trouble guessing what it means. Something like "`for_index`" would be more helpful, but that's getting back to the original length. 

Actually, it's not just a length issue (which, in reality, I don't tend to weight that heavily). I should not have said "shorter", as it implied that length is an important metric.

Rather, I find `nplike.index_nplike` less readable for repetition than `nplike.index.zeros` or `nplike.for_index.zeros`. 

> We could, although I think that would be easy to do at any time; nplikes are rather internal. Perhaps the `nplike` module ought to be made private (renamed to `_nplike`) so to _formally_ say that we have this flexibility.

OK, no problem - it won't take long to do should we see fit at a later date. Good luck @swishdiff!

jpivarski:
JAX segment_prod has some subtleties: https://github.com/google/jax/issues/9296 We're using the log → sum → exp that they recommend there.

sw5sh:
@jpivarski please merge this when the tests pass. There are two cases in `_slicing.py`, that probably need special treatment for Jax, for now they raise an error if the `nplike` is `Jax`

jpivarski (approved):
I approve! You can squash-and-merge.

PR #1491, opened 2022-06-06
Adding a Forth Based Avro Reader

aryan26roy:


aryan26roy:
Some Performance Indicators when compared to uavro:
- avro file with records of 3 floats (~100Mb) : uavro- 1.645 s, forth- 0.717 s
- avro file with records of 46 floats (~90 Mb) : uavro- 1.244 s, forth- 0.490 s

aryan26roy:
Some performance indicators when compared to fastavro:
- avro file with records of 46 floats (~90 Mb) : fastavro- 3.2529 s, forth- 0.490 s
- avro file with 3-level jagged array of floats (~688 Mb) : fastavro- 22.0425 s, forth- 2.699 s

src/awkward/_v2/_connect/avro.py
```diff
@@ -0,0 +1,1440 @@
...
+            #             '''
+            #         _exec_code = _exec_code+gg
+            #         _exec_code = _exec_code+hh
+            #         _exec_code = _exec_code+jj
+            #         _exec_code = _exec_code+kk
+            raise NotImplementedError
+
+
+class LengthError(Exception):
+    pass
```
jpivarski: Internal: should start with underscore.

Should be more descriptive. More like `ReachedEndOfArray`?

src/awkward/_v2/_connect/avro.py
```diff
@@ -0,0 +1,1440 @@
...
+
+
+class read_avro_ft:
```
jpivarski: Python class names are CapitalCamelCase.

src/awkward/_v2/_connect/avro.py
```diff
@@ -0,0 +1,1440 @@
...
+    def __init__(self, file, show_code=False):
+        # self.file_name = file_name
+        # self._data = np.memmap(file_name, np.uint8)
+        self._data = file
+        self.field = []
+        self.blocks = []
```
jpivarski: Only need the total number, not a list of the number in each block.

src/awkward/_v2/_connect/avro.py
```diff
@@ -0,0 +1,1440 @@
...
+        self._marker = 0
+        self.is_prim = False
```
jpivarski: Use the whole word, "primitive".

src/awkward/_v2/_connect/avro.py
```diff
@@ -0,0 +1,1440 @@
...
+            self.metadata["avro.schema"], _exec_code, ind, 0, dec, keys, _init_code, con
+        )
+        first_iter = True
+        _init_code.append(";\n")
+        self.update_pos(17)
+        header_code = header_code + "".join(dec)
+        header_code = header_code + "".join(_init_code)
+        _exec_code.insert(0, "0 do \n")
+        _exec_code1 = "".join(_exec_code)
+        forth_code = header_code + _exec_code1 + "\nloop"
```
jpivarski: Single format string?

jpivarski: Suggestion:

```python
forth_code = f"""
{header_code}
{init_code}
{exec_code}
"""
```

jpivarski (changes requested):
We talked in person. Most of the things are names, which will be easy to change.

aryan26roy:
The latest commit adds proper handling for negative length lists. This slows down the performance on the 3-level jagged array of floats from 2.7 s to 3.2 s. However, this slowdown can be avoided my generating three forth codes, one assuming no negative lengths, another assuming all to be negative lengths and third to halt after encountering the first length. This would (probably) give nearly the same performance as the one without checks.

PR #1494, opened 2022-06-08
Templated LayoutBuilder

ManasviGoyal:
To develop a templated GrowableBuffer<PRIMITIVE> and the Layout builders that use this buffer. There will be two headers: GrowableBuffer.h and LayoutBuilder.h. This will allow using them in an external project without linking to the awkward libraries.

- [x] Numpy
- [x] ListOffset
- [x] EmptyRecord
- [x] Record
- [x] Tuple
- [x] List
- [x] Indexed
- [x] IndexedOption
- [x] Empty
- [x] Unmasked
- [x] Regular
- [x] BitMasked
- [x] ByteMasked
- [x] Union
- [x] Unit Tests for all Layout Builders

This can be a separate PR:
Documentation of LayoutBuilder.h


src/awkward/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,105 @@
+#ifndef GROWABLEBUFFER_H_
```
ianna: I think, we should add Awkward here:
```c++
#ifndef AWKWARD_GROWABLEBUFFER_H_
```

src/awkward/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,105 @@
...
+#define GROWABLEBUFFER_H_
+
+#include <iostream>
+#include <stdint.h>
+#include <vector>
+#include <memory>
+#include <numeric>
+#include <cmath>
+
```
ianna: and put it in the awkward namespace:
```c++
namespace awkward {
```

src/awkward/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,105 @@
...
+    return (ptr_.size() == 1);
+  }
+
+private:
+  size_t initial_;
+  std::vector<std::unique_ptr<PRIMITIVE>> ptr_;
+  std::vector<size_t> length_;
+  std::vector<size_t> reserved_;
+};
+
```
ianna:
```cpp
}
```

ianna (changes requested):
Great! Please, see minor comments. Thanks.

ManasviGoyal:
> Great! Please, see minor comments. Thanks.

Done. Should I add the descriptions of the methods too?

ianna:
> > Great! Please, see minor comments. Thanks.
> 
> Done. Should I add the descriptions of the methods too?

Yes, please, it's a good idea!

jpivarski:
This is looking good! One thing, though: the GrowableBuffer will always (?) be used by external code that wants to do "copy to an external buffer" and "concatenation" as one operation. That's why we think the panels approach is a step up in performance: we know that external code will have to copy the data out of a GrowableBuffer so that that external code can own it, and if we have to do a copy anyway, we might as well concatenate in the same operation. With concatenation being "free," having no extra performance cost, that's why we can let the data be discontiguous as it grows and letting it be discontiguous as it grows lets us avoid more expensive ways of growing a buffer (reallocating and copying).

So the class needs to have an interface like

```c++
void concatenate(PRIMITIVE* external_pointer) const noexcept {
    // always iterates over panels, copying data into external_pointer
    // assumes that the buffer; external code should have allocated it with the right length()
    // this function does not change any of the GrowableBuffer's internal panels; they're left as-is
}
```

for external code to be able to use concatenate-and-copy in one operation.

For use in ArrayBuilder, it will need `zeros`, `empty`, `arange` constructors (see original GrowableBuffer). For use in AwkwardForth as a replacement for ForthOutputBuffer, it will need another suite of methods.

ArrayBuilder will be an immediate use-case, AwkwardForth is down the road. Only worry about the ArrayBuilder methods for now.

src/awkward/_v2/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,263 @@
...
+      reserved_.clear();
+      reserved_.push_back(initial_);
+      ptr_.clear();
+      ptr_.push_back(std::unique_ptr<PRIMITIVE>(new PRIMITIVE[initial_]));
+    }
+
+    /// @brief Inserts one `datum` into the panel.
+    void
+    fill_panel(PRIMITIVE datum) {
+      if (length_[ptr_.size()-1] < reserved_[ptr_.size()-1]) {
```
ianna: Should this member function be private? Also, this if statement looks like a double check. Should it be removed?

src/awkward/_v2/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,263 @@
...
+      }
+      fill_panel(datum);
+    }
+
+    /// @brief Inserts an entire array into the panel(s), possibly triggering
+    /// allocation of a new panel.
+    void
+    append(PRIMITIVE* ptr, size_t size) {
+      for (size_t i = 0; i < size; i++) {
+        append(ptr[i]);
```
ianna: `append` would trigger if on each element. Would it be better to check the size beforehand and allocate a new panel (possibly a `size` size if it is bigger that `reserved`?

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,534 @@
...
+      visit_impl<sizeof...(RECORDs)>::visit(contents, index, fun);
+  }
+
+  template <typename FUNCTION, typename... RECORDs>
+  void
+  visit_at(std::tuple<RECORDs...>& contents, size_t index, FUNCTION fun)
+  {
+      visit_impl<sizeof...(RECORDs)>::visit(contents, index, fun);
+  }
+
```
ianna: this implementation is rather generic. Should it be moved to `utils.h`?

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,534 @@
...
+      std::stringstream form_key;
+      form_key << "node" << id_;
+      if (std::is_arithmetic<PRIMITIVE>::value) {
+        return "{ \"class\": \"NumpyArray\", \"primitive\": \""
+                  + type_to_name<PRIMITIVE>() + "\", " + "\"form_key\": \"" + form_key.str() + "\" }";
+      } else if (is_specialization<PRIMITIVE, std::complex>::value) {
+        return "{ \"class\": \"NumpyArray\", \"primitive\": \"complex128\", \"form_key\": \""
+                  + form_key.str() + "\" }";
+      }
+      return "unsupported type";
```
ianna: perhaps, it would be better to add more information on which type is not supported yet? ` typeid(PRIMITIVE).name()`?

ianna (changes requested):
This looks good! Please, see my comments inlined.

ianna (approved):
Thanks! This looks good. The tests run.
@jpivarski - please, have a look. Thanks!

ianna:
@jpivarski - I've added a `cmake` configuration to run the cpp-header only tests [here](https://github.com/scikit-hep/awkward/pull/1528/commits/933765f493559acc9721e356656f5c31b8b011b7). It's a separate PR because there are no hooks to `pytest` yet.

ianna:
@ManasviGoyal - please, cherry-pick [this commit](https://github.com/scikit-hep/awkward/pull/1528/commits/765e6a9f28879cb93c90e6b47337ea2997c189a3) that clears the warnings when the tests are built with the project flags on MacOS. Thanks.

ManasviGoyal:
> @ManasviGoyal - please, cherry-pick [this commit](https://github.com/scikit-hep/awkward/pull/1528/commits/765e6a9f28879cb93c90e6b47337ea2997c189a3) that clears the warnings when the tests are built with the project flags on MacOS. Thanks.

Done!

ianna:
@ManasviGoyal -  I've stared testing the new `GrowableBuffer` implementation on an `ArrayBuilder`: see #1529

I think, we need to extend the `GrowableBuffer` API with a `fill` method that would take one (possibly multi-pannel) `GrowableBuffer<FROM_PRIMITIVE>` and make another (one panel) `GrowableBuffer<TO_PRIMITIVE>`.

There are several cases where it is needed for building arrays. Here is one example how a current one is used:
```c++
    GrowableBuffer<std::complex<double>> buffer =
      GrowableBuffer<std::complex<double>>::empty(initial, old.reserved());
    int64_t* oldraw = old.ptr().get();
    std::complex<double>* newraw = buffer.ptr().get();
    for (size_t i = 0;  i < 2*old.length();  i++) {
      newraw[i] = {static_cast<double>(oldraw[i]), 0};
    }
    buffer.set_length(old.length());
    old.clear();
```

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+    std::string
+    form() {
+      return builder.form();
+    }
+
+    field_name field_;
+    BUILDER builder;
+  };
```
jpivarski: This is not describing a record, but a single field of a record. Having this layer of indirection will force users to say

```c++
actual_record_builder.field("which_field").field()
```

(one extra, unnecessary `.field()` call).

I was at first going to say that we should use classes everywhere, rather than structs, so that all data members can be hidden. (Also, no surprises, no needing to look up whether something is a struct or a class.) I can see that this one was a struct because it's so simple that you don't want much fanfare around it, but it can be eliminated entirely, reducing boilerplate for users.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+
+    void
+    append(PRIMITIVE x) {
+      data_.append(x);
+    }
+
+    void
+    append(PRIMITIVE* ptr, size_t size) {
+      data_.append(ptr, size);
+    }
```
jpivarski: I was going to say that the Pythonic word for this is "extend", but since methods can be distinguished in C++ by type signature, it's not necessary to use a different word. I have a slight preference for distinguishing between "append" and "extend", but using "append" and "append" is fine enough.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+      std::stringstream form_key;
+      form_key << "node" << id_;
+      if (std::is_arithmetic<PRIMITIVE>::value) {
+        return "{ \"class\": \"NumpyArray\", \"primitive\": \""
+                  + type_to_name<PRIMITIVE>() + "\", " + "\"form_key\": \"" + form_key.str() + "\" }";
+      } else if (is_specialization<PRIMITIVE, std::complex>::value) {
+        return "{ \"class\": \"NumpyArray\", \"primitive\": \"complex128\", \"form_key\": \""
+                  + form_key.str() + "\" }";
+      }
+      return "type " + std::string(typeid(PRIMITIVE).name()) + "is not supported yet";
```
jpivarski: I have a style preference for (a) `else if` on a new line and (b) conditional chains to be complete: if there's a default case, it should be in an `else` clause. We're emulating a [pattern matching syntax](https://docs.scala-lang.org/tour/pattern-matching.html) in a language that doesn't have one. Relying on early returns is unstructured programming; the same arguments against `goto` apply. (There are cases in which an early `return` makes code easier to read, like checking for an error case at the beginning of a function, getting rid of it right away, so that the rest of the function can be considered without thinking about that error case. But this is much more like a pattern match block.

Also, if a type is not supported, this function should raise an exception—maybe the constructor should raise an exception, so that you don't have to run over a lot of data before finding out—it shouldn't just return invalid JSON. Returning invalid JSON means that the error will be raised downstream, not here; the stack trace will point to the code where `json.loads` is called, not where the LayoutBuilder snapshot happens.

```suggestion
      if (std::is_arithmetic<PRIMITIVE>::value) {
        return "{ \"class\": \"NumpyArray\", \"primitive\": \""
                  + type_to_name<PRIMITIVE>() + "\", " + "\"form_key\": \"" + form_key.str() + "\" }";
      }
      else if (is_specialization<PRIMITIVE, std::complex>::value) {
        return "{ \"class\": \"NumpyArray\", \"primitive\": \"complex128\", \"form_key\": \""
                  + form_key.str() + "\" }";
      }
      else {
        throw std::runtime_error("type " + std::string(typeid(PRIMITIVE).name()) + "is not supported");
      }
```

Is `complex128` the only complex supported? Not `complex64`? The set of primitive types that Awkward Array supports is all listed here:

https://github.com/scikit-hep/awkward/blob/17fd7c0688da29c484ac814154acbff4c7800f04/src/awkward/_v2/types/numpytype.py#L66-L82

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+    awkward::GrowableBuffer<PRIMITIVE> data_;
+    unsigned id_ = ID;
+  };
+
+  template <unsigned ID, unsigned INITIAL, typename BUILDER>
+  class ListOffsetLayoutBuilder {
+  public:
+    ListOffsetLayoutBuilder()
+        : offsets_(awkward::GrowableBuffer<int64_t>(INITIAL))
+        , begun_(false) {
```
jpivarski: There should be no `begun_` member. See below.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+    void
+    append(PRIMITIVE x) {
+      content_.append(x);
+    }
+
+    template<typename PRIMITIVE>
+    void
+    append(PRIMITIVE* ptr, size_t size) {
+      content_.append(ptr, size);
+    }
```
jpivarski: ListOffsetArray should have no `append` method:

https://github.com/scikit-hep/awkward/blob/17fd7c0688da29c484ac814154acbff4c7800f04/studies/templated-LayoutBuilder/LayoutBuilder.h#L54-L80

The reason we want `begin_list` to return a raw pointer to `content_` is because the only way to get new data into the content is to `append` it directly on the content node, not through the ListOffset node. Like this:

https://github.com/scikit-hep/awkward/blob/17fd7c0688da29c484ac814154acbff4c7800f04/studies/templated-LayoutBuilder/test-stuff.cpp#L8-L14

That way, the ListOffset node does not need to keep track of internal state, this `begun_`, which ArrayBuilder needed because it was dealing with a much more complex case: not knowing what the nested content trees are and allowing them to change dynamically. The opportunity that we gain from JIT-compiling LayoutBuilder is that we get to make it simpler and therefore faster.

The single `begun_` boolean is a small amount of internal state for ListOffset, but it gets a lot larger and more complex for records and unions. Let's eliminate it here and consistently have no internal state, so that the interface is similar for all nodes.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+      return &content_;
+    }
+
+    void
+    end_list() {
+      if (!begun_) {
+        throw std::invalid_argument(
+          std::string("called 'end_list' without 'begin_list' at the same level before it"
+          "in ListOffsetLayoutBuilder"));
+      }
```
jpivarski: Don't even check. `begin_list` and `end_list` can be `noexcept`. We want to make all functions that are called _O(n)_ times for an array of length _n_ be `noexcept` and streamline them as much as possible. (We _can_ make them all `noexcept`, and only the GrowableBuffers have mutable state.)

All `begin_list` should do is return a raw pointer to the `content_` (what it has to do with "beginning a list" is just to write idiomatic code that way) and all `end_list` should do is append `content_.length()` to `offsets_`. So, `begin_list` can even be `const` (though `end_list` can't).

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+    }
+
+  private:
+    GrowableBuffer<int64_t> offsets_;
+    BUILDER content_;
+    unsigned id_ = ID;
+    bool begun_;
+  };
+
+  template <unsigned ID, typename... RECORD>
```
jpivarski: You can remove the middleman: the struct that you're calling `RECORD` could actually be the `const char*` and `BUILDER` pairs...

Oh wait—is it not possible to make a variable number of arguments in a template that alternates between two types, such that

```c++
RecordLayoutBuilder<"one", NumpyBuilder<int32_t>, "two", NumpyBuilder<double>>
```

is a valid instantiation? Oh, in that case, I guess you need the small type that just pairs a `const char*` with a `BUILDER`. But you only need that for building the type, right? You could still make the interface involve less boilerplate:

```c++
RecordLayoutBuilder<Field<"one", NumpyBuilder<int32_t>>, Field<"two", NumpyBuilder<double>>> builder;

builder.field("one").append(123);
builder.field("two").append(3.14);
```

by having the `RecordLayoutBuilder<...>::field(const char* field)` method return the `Field<...>::field()`. There's no reason users should have to

```c++
builder.field("one").field().append(123);
builder.field("two").field().append(3.14);
```

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+  class RecordLayoutBuilder {
+  public:
+    RecordLayoutBuilder()
+        : contents({new RECORD}...)
+        , length_(0)
+        , begun_(false) { }
```
jpivarski: Again, the `begun_` mutable boolean is no longer necessary.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+    end_record() {
+      if (!begun_) {
+        throw std::invalid_argument(
+        std::string("called 'end_record' without 'begin_record' at the same level "
+                    "before it"));
+      } else {
+        length_++;
+        begun_ = false;
+      }
+    }
```
jpivarski: In LayoutBuilder's new interface, there's no reason to have a `begin_record` or `end_record` at all. The best they could do is ensure that all fields have been updated exactly once between `begin_record` and `end_record`, but that would require so much machinery (mutable counters for the old lengths of all fields, checking that they differ by one from the new lengths of all fields, incrementing them all) that it would not be appropriate in high-performance situations. This is analogous to `std::vector` not doing runtime index checking, while Python lists do. LayoutBuilder is more like a `std::vector` and ArrayBulider is more like a Python list.

We can and probably should check that all fields have the same `length()` _after_ the loop over all data, perhaps in the `form` method, which must be called once at the end and should be able to throw exceptions.

For the `length()` of the RecordLayoutBuilder, you can just return the `length()` of the first field.

This only leaves out records that have no fields, which we could deal with later—the implementation would be so different from this RecordLayoutBuilder that perhaps that (weird) case can be another class. The template can be constructed in such a way that the RecordLayoutBuilder is guaranteed to have at least one field.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+      form_key << "node" << id_;
+      std::stringstream out;
+      out << "{ \"class\": \"RecordArray\", \"contents\": { ";
+      for (size_t i = 0;  i < std::tuple_size<decltype(contents)>::value;  i++) {
+        if (i != 0) {
+          out << ", ";
+        }
+        auto contents_form = [&out] (auto record) { out << "\"" << record->field() << + "\": ";
+                                                    out << record->form(); };
+        visit_at(contents, i, contents_form);
```
jpivarski: Why do this? Is it to get around a type constraint?

src/awkward/_v2/cpp-headers/LayoutBuilder.h
jpivarski: Actually, the error-checking should be in a separate method from `form`, so that one _can_ call it on every entry if they want to trade safety for speed. I'll write more about that later.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+      content_.append(x);
+    }
+
+    template<typename PRIMITIVE>
+    void
+    append(PRIMITIVE* ptr, size_t size) {
+      for (size_t i = 0; i < size; i++) {
+        append(ptr[i]);
+      }
+    }
```
jpivarski: These methods assume that the content (type `BUILDER`) has an append method, which can't be true of RecordLayoutBuilders. Template instantiation would fail. IndexedOptionArrays of RecordArrays is a major use-case (it's the only way to have a missing value cost 4 or 8 bytes, rather than however many bytes a record has), so that would need to work.

These append methods are a convenience: they "pass through" an append operation to the underlying content and update the `index_` in the same step. The coordination of `index_` and `content_` is a nice service, but can't be performed in general. You can, however, make helper functions:

```c++
    void
    append_index() {
      index_.append(content_.length());
    }

    void
    append_index(size_t size) {
      size_t start = content_.length();
      for (size_t i = start; i < start + size; i++) {
        index_.append(i);
      }
    }

    void
    append_null() {
      index_.append(-1);
    }

    void
    append_null(size_t size) {
      for (size_t i = 0; i < size; i++) {
        index_.append(-1);
      }
    }
```

Then a user could get the `content()`, update that and `append_index` together, or not get the `content()` and only `append_null`, either once or `size` times.

Like RecordLayoutBuilder, this will require some error-checking, since it's possible to get the index and the content into different lengths. That's the danger of using LayoutBuilder, but the consistency can be checked at any time during the accumulation or at the end.

src/awkward/_v2/cpp-headers/LayoutBuilder.h
```diff
@@ -0,0 +1,499 @@
...
+    void
+    append(PRIMITIVE x) {
+      content_.append(x);
+    }
+
+    template<typename PRIMITIVE>
+    void
+    append(PRIMITIVE* ptr, size_t size) {
+      content_.append(ptr, size);
+    }
```
jpivarski: Same issue as for the IndexedLayoutBuilder: the `content_` (`BUILDER`) might not have an `append` method.

tests-cpp/test-layout-builder.cpp
```diff
@@ -0,0 +1,1000 @@
...
+       "node1", ptr1, builder2->length());
+  std::cout<<std::endl;
+}
+
+void
+test_list_offset_of_list_offset() {
+  auto builder = awkward::ListOffsetLayoutBuilder<0,
+      initial, awkward::ListOffsetLayoutBuilder<1,
+          initial, awkward::NumpyLayoutBuilder<2, initial, double>
+  >>();
```
jpivarski: Having to manually set ids is pretty user-unfriendly. They only have to be unique, and they don't have to be created quickly. Couldn't there be a global

```c++
static std::atomic<uint64_t> layoutbuilder_next_id = 0;
```

with helper functions to give every new LayoutBuilder instance a new id? They only have to be unique per-tree and that would make them unique per-process, but being _too_ unique isn't a problem.

Alternatively, each LayoutBuilder constructor could initiate a recursive walk that reassigns ids in-place to be unique per-tree (with the added benefit of being repeatable per-tree). It's okay if the act of inserting a LayoutBuilder into a tree changes its id because LayoutBuilders are mutable objects that can't be in more than one tree, anyway. (On second thought, this recursive-replace method is sounding better than the global atomic integer, since building the same tree would produce the same Form, and that's a reproducibility benefit.)

jpivarski (changes requested):
The GrowableBuffer is great and I would sign off on it right now.

The LayoutBuilder, on the other hand, is about halfway between the simple/fast tool that it could be and where ArrayBuilder is now. I have a lot of inline comments, but I think I could explain what I'm thinking a little better with a prototype. Most of this code can be used as-is—what I'll show with the prototype is a way to prune off some methods that are getting in the way and show how it can all fit together. I'll write it up in Python because it will be faster for me to write it that way; you have a good handle on the C++isms, and I think you'll have no trouble translating.

I'll post that here as soon as I can.

If you want to split this PR into one PR for GrowableBuffer and another for LayoutBuilder, then we'll be able to merge the GrowableBuffer part right away.

ManasviGoyal:
> The GrowableBuffer is great and I would sign off on it right now.
> 
Great! Thanks.

> The LayoutBuilder, on the other hand, is about halfway between the simple/fast tool that it could be and where ArrayBuilder is now. I have a lot of inline comments, but I think I could explain what I'm thinking a little better with a prototype. Most of this code can be used as-is—what I'll show with the prototype is a way to prune off some methods that are getting in the way and show how it can all fit together. I'll write it up in Python because it will be faster for me to write it that way; you have a good handle on the C++isms, and I think you'll have no trouble translating.
> 
Thanks. The prototype really gives a better idea to improve upon the implementation. I'll update the PR with the changes that you suggested.
> If you want to split this PR into one PR for GrowableBuffer and another for LayoutBuilder, then we'll be able to merge the GrowableBuffer part right away.

Yes, it will be better to split the PR so that GrowableBuffer can be merged. I'll do the same. Thanks.

jpivarski:
I finished writing the prototype. Along the way, I realized that there are some subtle issues, such as dealing with records that have no fields and regular arrays with inner size 0. There also needs to be a different builder for records with named fields ("records") and records whose fields are not named ("tuples").

I ran into some hard choices when trying to write Python that is intended to illustrate how it _would_ be implemented in C++. Specifically, there were a few places where the C++ should pass references to integers and strings to be able to modify them remotely. You just can't do that in Python. To simulate this, I made a

```python
class Ref:
    def __init__(self, value):
        self.value = value
```

because Python passes all mutable objects as references, so "`reference.value = 3`" means

```c++
// reference has C++ type &int
reference = 3;
```

I hope my meaning is clear.

I'm looking forward to the full C++ version of all of this!

ManasviGoyal:
> I finished writing the prototype. Along the way, I realized that there are some subtle issues, such as dealing with records that have no fields and regular arrays with inner size 0. There also needs to be a different builder for records with named fields ("records") and records whose fields are not named ("tuples").
> 
> I ran into some hard choices when trying to write Python that is intended to illustrate how it _would_ be implemented in C++. Specifically, there were a few places where the C++ should pass references to integers and strings to be able to modify them remotely. You just can't do that in Python. To simulate this, I made a
> 
> ```python
> class Ref:
>     def __init__(self, value):
>         self.value = value
> ```
> 
> because Python passes all mutable objects as references, so "`reference.value = 3`" means
> 
> ```c++
> // reference has C++ type &int
> reference = 3;
> ```
> 
> I hope my meaning is clear.
> 
> I'm looking forward to the full C++ version of all of this!

Yes. It is clear. I'll make the requested changes and update the soon. Thanks!

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
```diff
@@ -0,0 +1,1706 @@
...
+    BUILDER content_;
+    size_t id_;
+    size_t length_;
+    std::string parameters_;
+    size_t size_ = SIZE;
+  };
+
+  // FIXME: template index: int64_t and int32_t
+  template <unsigned INITIAL,  typename... BUILDERS>
+  class Union {
```
ianna: I think, `Union` should take an additional index type parameter to allow its `GrowableBuffer<int32_t> index_`

ianna (commented):
@ManasviGoyal and @jpivarski - the PR looks good. It implements all types and their tests defined in Jim's prototype. There are some places where I'd suggest to use a type aliases. Such as:
```c++
using Contents = typename std::tuple<BUILDERS...>;
```
and declare:
```c++
Contents contents_;
```
and more tests are needed to check that parameters are correctly placed in the forms.

I'll do a more thorough review after @ManasviGoyal has a chance to do any final touches.

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -278,6 +279,23 @@ namespace awkward {
       ptr_ = panel_.get();
     }
 
+    /// @brief Last element in last panel
+    PRIMITIVE
+    last() const {
+      if (ptr_->length() == 0) {
+        throw std::runtime_error("Buffer is empty");
```
jpivarski, line 283: Eventually, these functions should be `noexcept` and have some default action for such cases. However, doing that after a performance-test framework is set up would be more useful, so that we can see how much it matters.

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
```diff
@@ -0,0 +1,1706 @@
...
+
+    std::string index_as_field() const {
+      return std::to_string(static_cast<int>(index));
+    }
+
+    template<typename PRIMITIVE>
+    void
+    append(PRIMITIVE x) {
+      builder.append(x);
+    }
```
jpivarski: This method assumes that `Builder::append` exists, but that's not true of all builders. It would fail to compile if a record field's type is ever a list, for instance.

If this is a convenience method (not the only way to do it), then it should be removed.

I think it's not the only way to do it, since `builder` is public.

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
jpivarski: Union should be parameterized by a TAG type, though that has always been `uint8`, and an INDEX type, `int32`, `uin32`, or `int64`.

See v1 UnionArray: it's specialized for `UnionArray8_32`, `UnionArray8_U32`, and `UnionArray8_64`.

(Actually, I think that RNTuple unions have equal-sized TAG and INDEX types, and both of them are 32-bit. But they're also interleaved, and for an Index, we'd have to copy to make it contiguous, anyway, so the motivation to match RNTuple's TAG and INDEX types is gone.)

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
```diff
@@ -0,0 +1,1706 @@
...
+
+    const std::size_t index = ENUM;
+    Builder builder;
+  };
+
+  template <unsigned INITIAL, typename PRIMITIVE>
```
jpivarski, line 33: Why is the initial size of the buffer a compile-time constant?

This should become a runtime value, passed down through instance constructors, not template parameters.

With it in the template parameter, every C++ user is going to have to specify `INITIAL` at every node of the tree. That's a lot of boilerplate for a value they usually don't care about. They'll usually take a default that gets passed in from above, or through a `BuilderOptions`.

The reason it's a parameter at all, instead of hard-coded into the code, is so that we have some leverage when doing performance tests. Users, even C++  users, will trust whatever optimal value we come up with from those performance tests. They're not going to want to have to type it.

src/awkward/_v2/cpp-headers/awkward/utils.h
```diff
@@ -87,6 +89,18 @@ type_to_name<char>() {
...
+const std::string
+type_to_name<std::complex<float>>() {
+    return "complex64";
+}
+
+template <>
+const std::string
+type_to_name<std::complex<double>>() {
+    return "complex128";
+}
```
jpivarski, line 104: Explicitly setting these is a good idea. If the types other than complex are assigned through `typeid(T).name()`, it would be better to make them explicit overrides, too. The typeid name is platform-dependent: it might come out to be the right name on one system, but it then be the wrong name on another.

tests-cpp/test_1494-layout-builder.cpp
```diff
@@ -0,0 +1,1559 @@
...
+  std::vector<std::string> fields {"one", "two", "three"};
+  auto names = builder.field_names();
+
+  for (size_t i = 0; i < names.size(); i++) {
+    assert(names[i] == fields[i]);
+  }
+
+  auto& one_builder = builder.field<Field::one>();
+  auto& two_builder = builder.field<Field::two>();
+  auto& three_builder = builder.field<Field::three>();
```
jpivarski: I like this interface. This is a good way to solve the problem in C++!

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
```diff
@@ -0,0 +1,1706 @@
...
+    void
+    _append_end() {
+      current_index_ += 1;
+      if (valid_when_) {
+        current_byte_ref_ = current_byte_;
+      }
+      else {
+        current_byte_ref_ = ~current_byte_;
+      }
+    }
```
jpivarski: These functions didn't have to start with underscores because C++ is already marking them as private.

It also doesn't hurt, though.

jpivarski (changes requested):
Very nice! There are just a few things to fix, described inline.

src/awkward/_v2/cpp-headers/awkward/utils.h
ManasviGoyal: @jpivarski All the other primitive types are set explicitly too. The `typeid(T).name()` is just for the case when the user tries to use a type other than the primitive and it throws an exception. Should we assume that the user will always use a primitive type and make it `noexcept`?

jpivarski: Okay, good! I forgot that I was looking at a diff.

ManasviGoyal:
> @ManasviGoyal and @jpivarski - the PR looks good. It implements all types and their tests defined in Jim's prototype. There are some places where I'd suggest to use a type aliases. Such as:
> 
> ```c++
> using Contents = typename std::tuple<BUILDERS...>;
> ```
> 
> and declare:
> 
> ```c++
> Contents contents_;
> ```
> 
> and more tests are needed to check that parameters are correctly placed in the forms.
> 
> I'll do a more thorough review after @ManasviGoyal has a chance to do any final touches.

@ianna What kind of parameters the users might use? One I think will be `"__array__": "string"`.  Are there any other?

jpivarski:
I forgot to check that all the node types can have parameters, all different from each other. Some of them, like `__array__ == "string"`, are functional, having meanings that will be used by Awkward Array, but they can be anything at all. Some applications put documentation strings in there, and others have general metadata, like units.

At the C++ level, the parameters can just be a string. In Python, it will have to resolve to something JSON-formatted, but the C++ doesn't have to enforce that.

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -304,9 +304,9 @@ namespace awkward {
     void
     append(PRIMITIVE datum) {
       if (ptr_->length() == ptr_->reserved()) {
-        add_panel((size_t)ceil(ptr_->reserved() * 1.5));
+        _add_panel((size_t)ceil(ptr_->reserved() * 1.5));
```
ianna: @ManasviGoyal - Certain sets of names and function signatures are always reserved to the implementation in C++. I’d avoid using leading underscore for both function names and private members. When in doubt, please, use Google C++ style guide: https://google.github.io/styleguide/cppguide.html

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
```diff
@@ -464,33 +651,34 @@ namespace awkward {
...
     UserDefinedMap content_names_;
+    std::string parameters_;
+    size_t id_;
 
     static constexpr size_t fields_count_ = sizeof...(BUILDERS);
 
     template <std::size_t... S>
     void
-    map_fields(std::index_sequence<S...>) {
+    _map_fields(std::index_sequence<S...>) noexcept {
```
ianna: Same here

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
```diff
@@ -1572,59 +1634,31 @@ namespace awkward {
...
-              << " but index length " << last_valid_index_[tag];
+              << " but index length " << last_valid_index_[tag] << "\n";
           error.append(out.str());
+
           return false;
         }
       }
 
-      std::vector<bool> valid_contents = content_is_valid(index_sequence, error);
+      std::vector<bool> valid_contents = _content_is_valid(index_sequence, error);
```
ianna: Same here

ianna (changes requested):
@ManasviGoyal - beginning identifiers with an underscore is considered poor programming style.

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
ianna: The users will not see it - we can type alias it as in the test examples. The reason why we need it as a template parameter - we use the default constructors. The builders are allocated on the stack and so their GrowableBuffers. We could change it, but let’s discuss it on Wednesday first. Thanks

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
ManasviGoyal: @ianna Hi. I think you checked one of the previous commits. I have removed the underscores already. Also I had added some tests for form parameters. Please check.

ManasviGoyal:
Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?

jpivarski:
> Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?

The types of all of these Indexes (offsets, starts, stops, mask, etc.) should be templated. Most of them can be `int32`, `uint32`, or `int64`, so LayoutBuilder needs the flexibility to switch between these. The exceptions are ByteMasked masks, which only have one option, `int8`, BitMasked masks, which have only one option, `uint8`, and Union tags, which only have one option, `int8`.

You could check the v1 Content subclasses, in src/libawkward/array/*.cc, to see which templates are actually instantiated. LayoutBuilder should be reproducing the same set of types.

ianna:
> > Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?
> 
> The types of all of these Indexes (offsets, starts, stops, mask, etc.) should be templated. Most of them can be `int32`, `uint32`, or `int64`, so LayoutBuilder needs the flexibility to switch between these. The exceptions are ByteMasked masks, which only have one option, `int8`, BitMasked masks, which have only one option, `uint8`, and Union tags, which only have one option, `int8`.
> 
> You could check the v1 Content subclasses, in src/libawkward/array/*.cc, to see which templates are actually instantiated. LayoutBuilder should be reproducing the same set of types.

I'd suggest that it goes to the next PR.

ManasviGoyal:
> > > Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?
> > 
> > 
> > The types of all of these Indexes (offsets, starts, stops, mask, etc.) should be templated. Most of them can be `int32`, `uint32`, or `int64`, so LayoutBuilder needs the flexibility to switch between these. The exceptions are ByteMasked masks, which only have one option, `int8`, BitMasked masks, which have only one option, `uint8`, and Union tags, which only have one option, `int8`.
> > You could check the v1 Content subclasses, in src/libawkward/array/*.cc, to see which templates are actually instantiated. LayoutBuilder should be reproducing the same set of types.
> 
> I'd suggest that it goes to the next PR.

Actually, I already added the changes.

ianna:
> > > > Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?
> > > 
> > > 
> > > The types of all of these Indexes (offsets, starts, stops, mask, etc.) should be templated. Most of them can be `int32`, `uint32`, or `int64`, so LayoutBuilder needs the flexibility to switch between these. The exceptions are ByteMasked masks, which only have one option, `int8`, BitMasked masks, which have only one option, `uint8`, and Union tags, which only have one option, `int8`.
> > > You could check the v1 Content subclasses, in src/libawkward/array/*.cc, to see which templates are actually instantiated. LayoutBuilder should be reproducing the same set of types.
> > 
> > 
> > I'd suggest that it goes to the next PR.
> 
> Actually, I already added the changes.

Fantastic! Is it ready for review? Thanks!

ManasviGoyal:
> > > > > Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?
> > > > 
> > > > 
> > > > The types of all of these Indexes (offsets, starts, stops, mask, etc.) should be templated. Most of them can be `int32`, `uint32`, or `int64`, so LayoutBuilder needs the flexibility to switch between these. The exceptions are ByteMasked masks, which only have one option, `int8`, BitMasked masks, which have only one option, `uint8`, and Union tags, which only have one option, `int8`.
> > > > You could check the v1 Content subclasses, in src/libawkward/array/*.cc, to see which templates are actually instantiated. LayoutBuilder should be reproducing the same set of types.
> > > 
> > > 
> > > I'd suggest that it goes to the next PR.
> > 
> > 
> > Actually, I already added the changes.
> 
> Fantastic! Is it ready for review? Thanks!

Besides the BitMasked part that we were discussing, yes, everything else is ready for review.

src/awkward/_v2/cpp-headers/awkward/LayoutBuilder.h
jpivarski: Okay, the initial GrowableBuffer size has to be a compile-time constant for a C++ reason: you need to use default constructors. (Why is that?)

And if GrowableBuffers are parameterized by more than one parameter—if they need to have a different first panel size from subsequent panel sizes as we discussed in the last meeting—do all of those have to become template parameters?

If this is the case, then we need to provide type aliases as part of the header-only package, with some choice of GrowableBuffer parameters hard-coded. C++ users of this package need to be able to say `LayoutBuilder::Numpy<float>` without having to set up an alias on their own.

ianna:
> > > > > > Hi, @jpivarski. Should we keep the type of offsets, starts, stops, mask, etc. fixed in the code or should we make them templated parameters so that users can use different types as needed?
> > > > > 
> > > > > 
> > > > > The types of all of these Indexes (offsets, starts, stops, mask, etc.) should be templated. Most of them can be `int32`, `uint32`, or `int64`, so LayoutBuilder needs the flexibility to switch between these. The exceptions are ByteMasked masks, which only have one option, `int8`, BitMasked masks, which have only one option, `uint8`, and Union tags, which only have one option, `int8`.
> > > > > You could check the v1 Content subclasses, in src/libawkward/array/*.cc, to see which templates are actually instantiated. LayoutBuilder should be reproducing the same set of types.
> > > > 
> > > > 
> > > > I'd suggest that it goes to the next PR.
> > > 
> > > 
> > > Actually, I already added the changes.
> > 
> > 
> > Fantastic! Is it ready for review? Thanks!
> 
> Besides the BitMasked part that we were discussing, yes, everything else is ready for review.

Thanks! Yes, I have a solution for that, but let's do it in the next PR.

jpivarski (approved):
I still have a question about including GrowableBuffer parameters as compile-time template parameters, but this is looking good for merging now. Any minor issues can be picked up in later PRs.

ianna:
> I still have a question about including GrowableBuffer parameters as compile-time template parameters, but this is looking good for merging now. Any minor issues can be picked up in later PRs.

Thanks, I'll merge this one and rebase https://github.com/scikit-hep/awkward/pull/1560 that restores `BuilderOptions`.

ianna (approved):
Looks good! Thanks.

PR #1497, opened 2022-06-10
Fix/Fixed slicing shape for array of booleans

### LINKED ISSUE #1495: Boolean array slice followed by integer slice in v2
### Version of Awkward Array

HEAD

This is how things are supposed to work (and do, in v1):

```python
>>> import awkward as ak
>>> array = ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]])
>>> array[[True, False, True]]
<Array [[1.1, 2.2, 3.3], [4.4, 5.5]] type='2 * var * float64'>
>>> array[[True, False, True], 1]
<Array [2.2, 5.5] type='2 * float64'>
```

but in v2,

```python
>>> array = ak._v2.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]])
>>> array[[True, False, True]]
<Array [[1.1, 2.2, 3.3], [4.4, 5.5]] type='2 * var * float64'>
>>> array[[True, False, True], 1]
<Array [] type='0 * float64'>
```

Found by @swishdiff. Maybe @ioanaif is in the best position to address it?

ioanaif:
#1495

### DIFF
--- a/src/awkward/_v2/_slicing.py
+++ b/src/awkward/_v2/_slicing.py
@@ -59,8 +59,9 @@ def getitem_broadcast(items):
                     out[-1].metadata["shape"] = x.shape
                 elif issubclass(x.dtype.type, (np.bool_, bool)):
                     if len(x.shape) == 1:
-                        out.append(ak._v2.index.Index64(nplike.nonzero(x)[0]))
-                        out[-1].metadata["shape"] = x.shape
+                        current = ak._v2.index.Index64(nplike.nonzero(x)[0])
+                        out.append(current)
+                        out[-1].metadata["shape"] = current.data.shape
                     else:
                         for w in nplike.nonzero(x):
                             out.append(ak._v2.index.Index64(w))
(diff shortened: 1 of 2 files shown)


src/awkward/_v2/_slicing.py
```diff
@@ -59,8 +59,9 @@ def getitem_broadcast(items):
                     out[-1].metadata["shape"] = x.shape
                 elif issubclass(x.dtype.type, (np.bool_, bool)):
                     if len(x.shape) == 1:
-                        out.append(ak._v2.index.Index64(nplike.nonzero(x)[0]))
-                        out[-1].metadata["shape"] = x.shape
+                        current = ak._v2.index.Index64(nplike.nonzero(x)[0])
+                        out.append(current)
+                        out[-1].metadata["shape"] = current.data.shape
```
jpivarski, line 64: That makes a lot of sense: the shape had been wrong because at some point, we made the assumption that `x` is an array of integer positions, but here it's an array of booleans, unlike what goes into `out`. If this was the case, the following should have worked (pre-bug-fix), because the shape of the booleans and the shape of the integers is the same:

```python
>>> ak._v2.Array([[1.1, 2.2, 3.3], [4.4], [5.5, 6.6]])[[True, True, True], 0]
<Array [1.1, 4.4, 5.5] type='3 * float64'>
```

and it did.

jpivarski (approved):
Thanks for the quick fix, @ioanaif! If this is done (it looks like it is), please do squash-and-merge it.

PR #1505, opened 2022-06-15
Feat: add missing `_like` methods to `TypeTracer`

### LINKED ISSUE #1504: ak.* support for typetracer arrays

Running list of (discovered by trying in dask-awkward) missing functions:
- [ ] `ones_like`
- [ ] `zeros_like`
- [ ] `firsts`

Example failure:
```python
>>> import awkward._v2 as ak
>>> ak.ones_like(ak.Array(ak.Array([1, 2, 3]).layout.typetracer))
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/operations/ak_ones_like.py", line 30, in ones_like
    return _impl(array, highlevel, behavior, dtype)
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/operations/ak_ones_like.py", line 34, in _impl
    return ak._v2.operations.ak_full_like._impl(array, 1, highlevel, behavior, dtype)
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/operations/ak_full_like.py", line 177, in _impl
    out = layout.recursively_apply(action)
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/contents/content.py", line 1326, in recursively_apply
    return self._recursively_apply(
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/contents/numpyarray.py", line 1296, in _recursively_apply
    result = action(
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/operations/ak_full_like.py", line 164, in action
    nplike.ones_like(original),
  File "/Users/ddavis/.pyenv/versions/dask-dev/lib/python3.10/site-packages/awkward/_v2/_typetracer.py", line 531, in ones_like
    raise ak._v2._util.error(NotImplementedError)
NotImplementedError:
```

agoose77:
This PR just implements some of the missing `TypeTracer` methods to fix #1504

Changes:
- [x] Remove `unset` from `TypeTracer` signatures - we should support explicit `None`
- [x] Extend `from_array` to handle literal sequence of primitives
- [x] Add `_like` methods

### DIFF
--- a/src/awkward/_v2/_typetracer.py
+++ b/src/awkward/_v2/_typetracer.py
@@ -206,7 +206,15 @@ class TypeTracerArray:
 
         # not array-like
         if not hasattr(array, "shape"):
-            array = np.array(array)
+            sequence = list(array)
+            array = np.array(sequence)
+            if array.dtype is object:
+                raise ak._v2._util.error(
+                    ValueError(
+                        f"bug in Awkward Array: attempt to construct `TypeTracerArray` "
+                        f"from a sequence of non-primitive types: {sequence}"
+                    )
+                )
 
         if dtype is None:
             dtype = array.dtype
--- a/src/awkward/_v2/_typetracer.py
+++ b/src/awkward/_v2/_typetracer.py
@@ -432,9 +432,6 @@ class TypeTracerArray:
         return self
 
 
-unset = object()
-
-
 class TypeTracer(ak.nplike.NumpyLike):
     known_data = False
     known_shape = False
@@ -483,21 +480,21 @@ class TypeTracer(ak.nplike.NumpyLike):
 
     ############################ array creation
 
-    def array(self, data, dtype=unset, **kwargs):
+    def array(self, data, dtype=None, **kwargs):
         # data[, dtype=[, copy=]]
-        if dtype is unset:
+        if dtype is None:
             dtype = data.dtype
         return TypeTracerArray.from_array(data, dtype=dtype)
 
-    def asarray(self, array, dtype=unset, **kwargs):
+    def asarray(self, array, dtype=None, **kwargs):
         # array[, dtype=][, order=]
-        if dtype is unset:
+        if dtype is None:
             dtype = array.dtype
         return TypeTracerArray.from_array(array, dtype=dtype)
 
-    def ascontiguousarray(self, array, dtype=unset, **kwargs):
+    def ascontiguousarray(self, array, dtype=None, **kwargs):
         # array[, dtype=]
-        if dtype is unset:
+        if dtype is None:
             dtype = array.dtype
         return TypeTracerArray.from_array(array, dtype=dtype)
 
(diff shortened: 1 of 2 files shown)


agoose77:
The modified `from_array` impl doesn't handle a sequence of non-numpy-friendly value types. We now raise an error for this case, rather than permitting `'O'` dtypes.

At this point I think we need to decide *how* flexible `from_array` needs to be. Namely:
- should we handle sequence of mixed `Unknown(Scalar|Length|MaybeNone)`, or only accept numpy-friendly types?
- do we want to accept scalars here like NumPy does?

My thought is to keep it restricted for now. `nplike` is an internal API (in the sense that users are not expected to work with it), so we can restrict ourselves to sequence-of-known-scalars.

jpivarski:
This function creates TypeTracerArrays _from_ NumPy arrays to act as buffers within an Awkward Array, so they need to be the kinds of NumPy arrays that can be buffers in an Awkward Array. Those NumPy arrays can't have `dtype=object`, which means MaybeNone is off the list, though UnknownScalars and UnknownLength both represent integers, and UnknownScalar says what `dtype` it will be (useful for making the array). UnknownLength is like a Python `int`, so 64-bit (maybe 32-bit on Windows?).

Is this PR done? It looks done, and I'd like to make a new release very soon (definitely by tomorrow).

Why did you remove the `unset = object()`? I created that to make a distinction between a keyword argument that wasn't explicitly set and a keyword argument that may have been set to `None`. Is there no need for such a distinction? (I can't think of any—I was just being cautious.)

agoose77:
@jpivarski my concern was if we use it internally for anything: I recall doing `nplike.asarray` on literals in Awkward at one point. 
If we're confident this should only receive "valid" values, then I'm happy with what's here currently.

`unset = object()` is fine if we need it, but we had lots of `if arg is unset` which fails for when the user passes in the signature-specified default e.g. `None`. Not sure if that makes too much sense - I have COVID brain right now :upside_down_face:

jpivarski:
I'm sorry to hear that—I hope you feel better soon!

I think `nplike.asarray` might be used on a user-provided input in `ak.fill_none` (on the value that will fill the array). The first attempt is to make that fill value an Awkward Array, though.

The `nplike` objects are strictly for our internal use, and we can mandate that internal code always specifies arguments. Maybe that would be too annoying. Anyway, I had been thinking that as we learn how our own code calls these `nplike` functions, we'd put appropriate defaults in there, one at a time replacing the `unset` objects.

But that's okay; I think that's minor. Thanks for the new methods and I'll include them in the pre-release!

douglasdavis:
Thanks @agoose77! `ones_like` and `zeros_like` are working.

edit: I had originally pointed out a new issue with `ak.to_numpy` but I think it's actually OK.

PR #1508, opened 2022-06-16
from rdataframe for awkward arrays

ianna:


src/awkward/_v2/_connect/cling.py
```diff
@@ -134,11 +142,22 @@ class ArrayView {
...
+    }
+
   protected:
     ssize_t start_;
     ssize_t stop_;
     ssize_t which_;
     ssize_t* ptrs_;
+    PyObject* lookup_;
+    PyObject* generator_;
+
```
ianna: @jpivarski - I'm not sure how to avoid adding both lookup and generator pointers. The generator needs a Form  of a lookup to convert it to a layout.

ianna (commented):
@jpivarski - I think, it's done. `Numba` code is not affected. Anything that relies on `cling` does not need to specify `lookup` nor `generator`. `RDataFrame` source owns them both, and its lifetime is longer than `from_rdataframe` operation that needs them.

src/awkward/_v2/_connect/cling.py
```diff
@@ -54,11 +56,11 @@ def generate_ArrayView(compiler, use_cached=True):
   template <typename ARRAY, typename VALUE>
   class Iterator {
   public:
-    Iterator(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs)
-      : start_(start), stop_(stop), which_(which), ptrs_(ptrs) { }
+    Iterator(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs, PyObject* lookup = nullptr, PyObject* generator = nullptr)
```
jpivarski: Since the `Iterators` are internally generated by us (not user-facing code), they don't have to have default arguments.

In fact, it's better if they don't because we would want to know immediately if we're ever creating an Iterator (or RIterator, or Array, or Record) that doesn't have these arguments set.

I can see how it's convenient to include a default value for a new argument while you're adding it, but eventually the default value should be taken off so that we can confirm that these objects are never getting `nullptr`s.

src/awkward/_v2/_connect/cling.py
```diff
@@ -84,16 +86,18 @@ class Iterator {
...
+    PyObject* lookup_;
+    PyObject* generator_;
   };
 
   template <typename ARRAY, typename VALUE>
   class RIterator {
   public:
-    RIterator(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs)
-      : start_(start), stop_(stop), which_(which), ptrs_(ptrs) { }
+    RIterator(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs, PyObject* lookup = nullptr, PyObject* generator = nullptr)
```
jpivarski: See comment about Iterators (above).

src/awkward/_v2/_connect/cling.py
```diff
@@ -119,12 +123,16 @@ class RIterator {
...
+    PyObject* lookup_;
+    PyObject* generator_;
+
   };
 
   class ArrayView {
   public:
-    ArrayView(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs)
-      : start_(start), stop_(stop), which_(which), ptrs_(ptrs) { }
+    ArrayView(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs, PyObject* lookup = nullptr, PyObject* generator = nullptr)
```
jpivarski: See comment about Iterators (above).

src/awkward/_v2/_connect/cling.py
```diff
@@ -163,13 +182,15 @@ def generate_RecordView(compiler, use_cached=True):
 namespace awkward {
   class RecordView {
   public:
-    RecordView(ssize_t at, ssize_t which, ssize_t* ptrs)
-      : at_(at), which_(which), ptrs_(ptrs) { }
+    RecordView(ssize_t at, ssize_t which, ssize_t* ptrs, PyObject* lookup = nullptr, PyObject* generator = nullptr)
```
jpivarski: See comment about iterators (above).

src/awkward/_v2/_connect/cling.py
```diff
@@ -583,9 +604,9 @@ def generate(self, compiler, use_cached=True):
 namespace awkward {{
   class {self.class_type()}: public ArrayView {{
   public:
-    {self.class_type()}(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs)
-      : ArrayView(start, stop, which, ptrs) {{ }}
-    {self.class_type()}() : ArrayView(0, 0, 0, 0) {{ }}
+    {self.class_type()}(ssize_t start, ssize_t stop, ssize_t which, ssize_t* ptrs, PyObject* lookup = nullptr, PyObject* generator = nullptr)
```
jpivarski: See comment about Iterators (above).

After this, there are a lot more cases of these for all the concrete classes. It _should_ be safe to just remove the defaults now that everything has been written. If there's a compiler error, then that's great—it would otherwise have been a segfault.

src/awkward/_v2/_connect/cling.py
jpivarski: As I (re-)learned on Zoom last week, a Generator is a subclass of a Lookup, so it could be one and the same object. That's not how it was implemented in Python: the `generator` is made first, from a Form, and the `lookup` is created afterward, from an Array. (In Numba, there's only one object.)

But it doesn't have to stay that way: the Python `generator` is made first, followed by a `lookup`. You could attach the original `generator` to the `lookup` as an attribute:

```python
lookup._generator = generator
```

and then only pass around `PyObject* lookup`. When you need to reconstitute a layout from a Lookup, you can assume that the `lookup._generator` exists, to get the Form.

Alternatively, you could attach only what you need:

```python
lookup._form = generator.form
```

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -157,7 +157,16 @@ def form_dtype(form):
 
         # Triggers event loop and execution of all actions booked in the associated RLoopManager.
         cpp_reference = result_ptrs.GetValue()
+        lookup = cpp_reference[0].lookup()
+        generator = cpp_reference[0].generator()
```
jpivarski: This is what could be

```python
lookup = cpp_reference[0].lookup()
generator = lookup._generator
```

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -157,7 +157,16 @@ def form_dtype(form):
...
+        array_out = generator[column].tolayout(lookup[column], 0, ())
 
-        return _wrap_as_array(column, ak._v2.from_iter(cpp_reference))
+        return ak._v2._util.wrap(
+            ak._v2.contents.RecordArray(
+                fields=[column],
+                contents=[array_out],
+            ),
+            highlevel=True,
+        )
```
jpivarski: Isn't this the same as `_maybe_wrap` (which is no longer "maybe")?

And `_wrap_as_array` doesn't seem to be used anymore (i.e. dead code).

Always returning a RecordArray of one field is a step toward returning a collection of columns as a RecordArray of many fields.

src/awkward/_v2/_connect/rdataframe/to_rdataframe.py
```diff
@@ -231,27 +232,30 @@ class {array_data_source} final
...
-            }}
+        ~{array_data_source}() {{
+            Py_DECREF(fPyLookup);
+            Py_DECREF(fPyGenerator);
+        }}
 
-            void SetNSlots(unsigned int nSlots) {{
-                fNSlots = nSlots; // FIXME: always 1 slot for now
+        void SetNSlots(unsigned int nSlots) {{
+            fNSlots = nSlots; // FIXME: always 1 slot for now
```
jpivarski: The comment is not true: this RDataSource is fully parallelizable. When RDataFrame asks for it to have `nSlots`, it uses all those slots, not just 1.

```suggestion
            fNSlots = nSlots;
```

tests/v2/test_1508-awkward-from-rdataframe.py
```diff
@@ -0,0 +1,83 @@
...
+
+    array_out = generator.tolayout(lookup, 0, ())
+
+    # data_frame = ak._v2.to_rdataframe({"x": array})
+    #
+    # assert data_frame.GetColumnType("x") == "double"
+    #
+    # ak_array_out = ak._v2.from_rdataframe(
+    #     data_frame, column="x", column_as_record=False
+    # )
```
jpivarski: Is this test still valid? (If not, we can delete it, rather than comment it out.)

jpivarski (changes requested):
This PR adds the ability to pass ak.Arrrays through RDF in a zero-copy way (by reference). It doesn't handle the case of an RDF with a Filter, though that should probably be another PR.

I don't think it's necessary to pass through both `lookup` and `generator`, since a `generator` (which has less information) can be attached to a `lookup` (which has more information) as an attribute. See inline.

Also, we should be able to remove the default ` = nullptr` now (or after dropping the `PyObject* generator`).

A test that should be included in this PR (not noted below) is something that ensures that the Lookup object reference count is not changed. For example,

https://github.com/scikit-hep/awkward/blob/49a3ff13ef90b8778a80573211d58c544729eaa5/tests/v2/test_0118-numba-cpointers.py#L92-L157

runs compiled code that (`f1`) consumes a Lookup reference and has no array output, code that (`f2`) consumes a Lookup reference and produces an array, and code that (`f3`) consumes a Lookup reference and returns two arrays. They all compile the code 3 times and run over it 10 times, checking `sys.getrefcount` of anything that is referenced on the compiled side. For this implementation, that would be the `lookup` and `generator`, but if `generator` is attached to the `lookup` and only the `PyObject* lookup` is passed around, it would be enough to just check the `lookup`. While working on the test, I also verified that removing the `Py_DECREF`

https://github.com/scikit-hep/awkward/blob/0eb61fe9b01b696facf8a82b86c07b41e6785ceb/src/awkward/_v2/_connect/rdataframe/to_rdataframe.py#L249-L252

caused the `sys.getrefcount`s to increase linearly—to know that I was monitoring the right Python objects and would have seen a memory leak if it existed. (That's why the loop up to 10 is useful: just getting a wrong refcount could be a lot of different problems, but a linearly increasing refcount as it's passed to and from C++ repeatedly is a smoking gun.)

src/awkward/_v2/_connect/cling.py
ianna: done

src/awkward/_v2/_connect/cling.py
ianna: done

src/awkward/_v2/_connect/cling.py
ianna: done

src/awkward/_v2/_connect/cling.py
ianna: done

src/awkward/_v2/_connect/cling.py
ianna: done

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py

src/awkward/_v2/_connect/rdataframe/to_rdataframe.py
ianna: done

tests/v2/test_1508-awkward-from-rdataframe.py
ianna: done

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py

src/awkward/_v2/_connect/cling.py
ianna: I've added a `generator` as a `Lookup` instance attribute because its other attributes do not have an `_` in their names.

ianna:
@jpivarski - I think, I've addressed all your comments. Please, have a look when you have time. Thanks!

jpivarski (approved):
Everything here looks great! Yes, let's merge it!

PR #1516, opened 2022-06-23
Package: fix MANIFEST to include cpp-headers

agoose77:
We can write any number of patterns here. I chose `recursive-include src/awkward/_v2/cpp-headers *.h` because it's explicit and simple.

We could just include any headers found in the source like we do for `cpp` files. Right now, I'm tempted to keep it very narrow, as I imagine we'll modify this build tooling in the near future once `scikit-build` gets Hatch(?) support?

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@ include dev/generate-kernel-signatures.py
 include kernel-specification.yml
 recursive-include include *.h
 recursive-include src *.cpp *.py *.cu
+recursive-include src/awkward/_v2/cpp-headers *.h
 recursive-include tests *.cpp *.py samples/*
 recursive-include tests-cpu-kernels *.py
 recursive-include tests-cuda-kernels *.py


jpivarski (approved):
I checked out the [artifact](https://github.com/scikit-hep/awkward/suites/7068802197/artifacts/278918030) that the [packaging test](https://github.com/scikit-hep/awkward/actions/runs/2552007126) produced and both the source tarball and one of the wheels contain the desired file:

```bash
% ls awkward-1.9.0rc6/src/**/*.h   # inflated tarball
awkward-1.9.0rc6/src/awkward/_v2/cpp-headers/rdataframe_jagged_builders.h

% ls awkward/**/rdataframe_jagged_builders.h   # inflated wheel
awkward/_v2/cpp-headers/rdataframe_jagged_builders.h
```

So it's good! I'll merge it now.

jpivarski:
Fixes the issue described in https://github.com/scikit-hep/awkward/pull/1295#issuecomment-1164617443.

PR #1518, opened 2022-06-29
chore: remove unneeded lines

henryiii:
These aren't needed anymore, `v2` will no longer update to `v3.0.0`, but only to `v3`, etc.


jpivarski:
Okay, thanks!

PR #1523, opened 2022-06-30
add C++ headers-only distribution configuration

ianna:
as discussed with @jpivarski this restores a config.py, but it defines only header part as follows (no libraries):

```
% python -m awkward._v2.config
% python -m awkward._v2.config --cflags
-std=c++17 -I/Users/yana/Projects/PR1522/awkward/awkward/src/_v2/cpp-headers
% python -m awkward._v2.config --cflags-only-I
-I/Users/yana/Projects/PR1522/awkward/awkward/src/_v2/cpp-headers
% python -m awkward._v2.config --incdir       
/Users/yana/Projects/PR1522/awkward/awkward/src/_v2/cpp-headers
```


jpivarski (approved):
Only one thing: I think this file should have path `src/awkward/_v2/config.py` so that it can't be confused with the old one. When `/_v2/` → `/`, it _will_ occupy the old space, but then it will be a new major version of Awkward and we expect things like that to change.

Other than that, it's great! Go ahead and merge it when it's moved into the `_v2` directory.

ianna:
> Only one thing: I think this file should have path `src/awkward/_v2/config.py` so that it can't be confused with the old one. When `/_v2/` → `/`, it _will_ occupy the old space, but then it will be a new major version of Awkward and we expect things like that to change.
> 
> Other than that, it's great! Go ahead and merge it when it's moved into the `_v2` directory.

I did move the file. Now the CI does not trigger the tests...

src/awkward/_v2/config.py
```diff
@@ -0,0 +1,44 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
```
jpivarski, line 2:
```suggestion
```

PR #1524, opened 2022-07-01
restructure cpp headers

ianna:
extract common code to a separate header file


ianna:
@ManasviGoyal - FYI. When this is merged you can update your PR with:
```
#include "utils.h"
```
and remove the duplicated code from `LayoutBuilder.h`. Thanks!

ManasviGoyal:
> @ManasviGoyal - FYI. When this is merged you can update your PR with:
> 
> ```
> #include "utils.h"
> ```
> 
> and remove the duplicated code from `LayoutBuilder.h`. Thanks!

Sure. Thanks!

jpivarski (approved):
Moving generic things like this out to a new utils.h is a good idea.

PR #1532, opened 2022-07-06
Fix: Error when using ak.copy in v2

### LINKED ISSUE #1482: Error when using ak.copy in v2
### Version of Awkward Array

1.9.0rc4

```
import awkward._v2 as ak

array = ak.Array([
    [{"x": 1.1, "y": [1]}, {"x": 2.2, "y": [1, 2]}, {"x": 3.3, "y": [1, 2, 3]}],
    [],
    [{"x": 4.4, "y": [1, 2, 3, 4]}, {"x": 5.5, "y": [1, 2, 3, 4, 5]}]
])

ak.copy(array)
```

Expect a copy of the array, but I get this:

```
File ~/miniconda3/envs/research/lib/python3.9/site-packages/awkward/_v2/operations/ak_copy.py:57, in copy(array)
     10 """
     11 Returns a deep copy of the array (no memory shared with original).
     12 
   (...)
     51 a mutable third-party library, so this function allows you to make a true copy.
     52 """
     53 with ak._v2._util.OperationErrorContext(
     54     "ak._v2.fill_none",
     55     dict(array=array),
     56 ):
---> 57     return _impl(array)
...
...
...
TypeError: cannot pickle 'module' object
```

ioanaif:
This PR aims to fix #1482


src/awkward/_v2/record.py
```diff
@@ -191,8 +191,8 @@ def to_list(self, behavior=None):
 
         return self._array[self._at : self._at + 1]._to_list(behavior, None)[0]
 
-    def deep_copy(self):
-        return Record(self._array.deep_copy(), copy.deepcopy(self._at))
+    def __deepcopy__(self):
+        return Record(self._array.__deepcopy__(), copy.deepcopy(self._at))
```
jpivarski:
```suggestion
    def __deepcopy__(self, memo=None):
        return Record(self._array.__deepcopy__(memo), self._at)
```

src/awkward/_v2/contents/content.py
```diff
@@ -1480,15 +1480,13 @@ def with_parameter(self, key, value):
 
         return out
 
-    def deep_copy(self, memo=None):
+    def __deepcopy__(self, memo=None):
         cls = self.__class__
         new_instance = cls.__new__(cls)
         memo = {"id(self)": new_instance}
```
jpivarski:
```suggestion
        memo = {id(self): new_instance}
```

src/awkward/_v2/contents/content.py
```diff
@@ -1480,15 +1480,13 @@ def with_parameter(self, key, value):
...
         for k, v in self.__dict__.items():
             if k == "_nplike":
                 setattr(new_instance, k, v)
```
jpivarski:
```suggestion
            if k == "_nplike":
                new_instance._nplike = v
```

jpivarski (approved):
This should do it, thanks!

src/awkward/_v2/contents/content.py
jpivarski: Proof:

```python
>>> import copy
>>> one = object()
>>> two = object()
>>> three = object()
>>> four = object()
>>> five = object()
>>> x = [one, two, three, four, five]
>>> x.append(x)
>>> y = copy.deepcopy(x, memo)
>>> memo[id(x[0])] is y[0]
True
```

src/awkward/_v2/contents/content.py
jpivarski: Proof:

```python
>>> import copy
>>> one = object()
>>> two = object()
>>> three = object()
>>> four = object()
>>> five = object()
>>> x = [one, two, three, four, five]
>>> x.append(x)
>>> y = copy.deepcopy(x, memo)
>>> memo[id(x[0])] is y[0]
True
```

PR #1533, opened 2022-07-06
Fix: numba pre-commit issues

ioanaif:


tests/test_0057b-virtual-array-numba.py
```diff
@@ -74,8 +74,8 @@ def test_listarray():
         counter = [0]
 
         def materialize():
-            counter[0] += 1
-            return layout
+            counter[0] += 1  # noqa: B023
+            return layout  # noqa: B023
```
jpivarski, line 78: Safe because `materialize` functions are not collected in a list and used outside the loop. They're only used in the loop.

src/awkward/_connect/_numba/arrayview.py
```diff
@@ -907,20 +907,22 @@ class type_method(numba.core.typing.templates.AbstractTemplate):
...
                                     x.literal_type
                                     if isinstance(x, numba.types.Literal)
                                     else x
                                     for x in args
                                 ],
-                            )(lower)
+                            )(
+                                lower  # noqa: B023
+                            )
                             return sig
```
jpivarski, line 926: Safe because `generic` is returned as soon as it's created; it never sees a different value of `args` and `methodname` than the ones it has at the time of its construction.

src/awkward/_v2/_connect/numba/arrayview.py
```diff
@@ -663,20 +663,22 @@ class type_method(numba.core.typing.templates.AbstractTemplate):
...
                                     x.literal_type
                                     if isinstance(x, numba.types.Literal)
                                     else x
                                     for x in args
                                 ],
-                            )(lower)
+                            )(
+                                lower  # noqa: B023
+                            )
                             return sig
```
jpivarski, line 682: Same here.

jpivarski (approved):
Thanks! I'll merge it so that all the other active PRs can use it.

PR #1535, opened 2022-07-06
Growable Buffer header

ManasviGoyal:
New header-only implementation of Growable Buffer using panels. Contains the following :-

- [x]  GrowableBuffer.h
- [x]  Unit Tests for GrowableBuffer
- [x]  Documentation of GrowableBuffer.h

### DIFF
--- a/src/awkward/_v2/cpp-headers/GrowableBuffer.h
+++ b/src/awkward/_v2/cpp-headers/GrowableBuffer.h
@@ -238,27 +238,26 @@ namespace awkward {
       return (ptr_.size() == 1);
     }
 
-    /// @brief Takes one (possibly multi-panels) GrowableBuffer<FROM_PRIMITIVE>
+    /// @brief Takes this (possibly multi-panels) GrowableBuffer<PRIMITIVE>
     /// and makes another (one panel) GrowableBuffer<TO_PRIMITIVE>.
     ///
-    /// Used to change the data type of buffer content from `FROM_PRIMITIVE`
+    /// Used to change the data type of buffer content from `PRIMITIVE`
     /// to `TO_PRIMITIVE` for building arrays.
-    template<typename FROM_PRIMITIVE>
-    static GrowableBuffer<PRIMITIVE>
-    fill_from(GrowableBuffer<FROM_PRIMITIVE>& from_buffer) {
-      auto length = from_buffer.length();
-      auto ptr = std::unique_ptr<PRIMITIVE>(new PRIMITIVE[length]);
-      PRIMITIVE* rawptr = ptr.get();
+    template<typename TO_PRIMITIVE>
+    GrowableBuffer<TO_PRIMITIVE>
+    copy_as() {
+      auto len = length();
+      auto ptr = std::unique_ptr<TO_PRIMITIVE>(new TO_PRIMITIVE[len]);
+      TO_PRIMITIVE* rawptr = ptr.get();
       int64_t k = 0;
-      for (size_t i = 0;  i < from_buffer.ptr_.size(); i++) {
-        for (int64_t j = 0; j < from_buffer.length_[i]; j++) {
-          rawptr[k] = static_cast<PRIMITIVE>(from_buffer.ptr_[i].get()[j]);
+      for (size_t i = 0;  i < ptr_.size(); i++) {
+        for (int64_t j = 0; j < length_[i]; j++) {
+          rawptr[k] = static_cast<TO_PRIMITIVE>(ptr_[i].get()[j]);
           k++;
         }
       }
-      return GrowableBuffer<PRIMITIVE>(length, std::move(ptr), length, length);
+      return GrowableBuffer<TO_PRIMITIVE>(len, std::move(ptr), len, len);
     }
-
   private:
     /// @brief Inserts one `datum` into the panel.
     void
--- a/src/awkward/_v2/cpp-headers/GrowableBuffer.h
+++ b/src/awkward/_v2/cpp-headers/GrowableBuffer.h
@@ -1,0 +1,298 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+#ifndef AWKWARD_GROWABLEBUFFER_H_
+#define AWKWARD_GROWABLEBUFFER_H_
+
+#include <vector>
+#include <memory>
+#include <numeric>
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Great! The location of this file will move and it the class will get new methods, but this is a good checkpoint.

To get the tests to pass, merge with `main`. @ioanaif's #1533 fixes the pre-commit issues that you see. Once it does pass tests, you can squash-and-merge it yourself.

ManasviGoyal:
> Great! The location of this file will move and it the class will get new methods, but this is a good checkpoint.
> 
> To get the tests to pass, merge with `main`. @ioanaif's #1533 fixes the pre-commit issues that you see. Once it does pass tests, you can squash-and-merge it yourself.

Sure. Thanks. I have added the function that is needed for use in Array Builder. So, it should be able to work with Array Builders and the new Layout Builder. New methods can be added as and when they are needed.

jpivarski:
At some point, we should be able to remove the old GrowableBuffer, too.

Is that currently dead code? (I.e. if you just remove the old GrowableBuffer, can it compile?) When it is possible to remove, let's do that right away so that we don't end up with a name conflict somehow.

Fortunately, GrowableBuffer does not have a pybind11 interface, so there won't be anything in the src/python directory to dismantle.

src/awkward/_v2/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,298 @@
...
+    /// Used to change the data type of buffer content from `FROM_PRIMITIVE`
+    /// to `TO_PRIMITIVE` for building arrays.
+    template<typename FROM_PRIMITIVE>
+    static GrowableBuffer<PRIMITIVE>
+    fill_from(GrowableBuffer<FROM_PRIMITIVE>& from_buffer) {
+      auto length = from_buffer.length();
+      auto ptr = std::unique_ptr<PRIMITIVE>(new PRIMITIVE[length]);
+      PRIMITIVE* rawptr = ptr.get();
+      int64_t k = 0;
+      for (size_t i = 0;  i < from_buffer.ptr_.size(); i++) {
```
ianna: I think, you need to add accessors to private members: ptr_ and length_

ianna (changes requested):
Please, see my comment about accessing private members from a static function. Thanks.

src/awkward/_v2/cpp-headers/GrowableBuffer.h
ManasviGoyal: Do we need to expose the length_ (and ptr_) of every panel? As currently, we already have a method that returns accumulated length.

jpivarski: Isn't the fact that GrowableBuffer is implemented by panels an internal detail that you don't want other code depending on? (For instance, if we find in the future that some mix of panels and reallocation is ideal... not that I think that will happen.) Exposing either `ptr_` or `length_` would reveal the panels.

ManasviGoyal: Yes, I agree. Exposing `ptr_` and `length_` should be avoided.

src/awkward/_v2/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,298 @@
...
+        for (int64_t j = 0; j < from_buffer.length_[i]; j++) {
+          rawptr[k] = static_cast<PRIMITIVE>(from_buffer.ptr_[i].get()[j]);
+          k++;
+        }
+      }
+      return GrowableBuffer<PRIMITIVE>(length, std::move(ptr), length, length);
+    }
```
ianna, line 260:
```suggestion
    /// @brief Takes one (possibly multi-panels) GrowableBuffer<FROM_PRIMITIVE>
    /// and makes another (one panel) GrowableBuffer<TO_PRIMITIVE>.
    ///
    /// Used to change the data type of buffer content from `FROM_PRIMITIVE`
    /// to `TO_PRIMITIVE` for building arrays.
    template<typename FROM_PRIMITIVE>
    static GrowableBuffer<PRIMITIVE>
    fill_from(GrowableBuffer<FROM_PRIMITIVE>& from_buffer) {
      auto length = from_buffer.length();
      auto ptr = std::unique_ptr<PRIMITIVE>(new PRIMITIVE[length]);
      PRIMITIVE* rawptr = ptr.get();
      int64_t k = 0;
      for (size_t i = 0;  i < from_buffer.ptr_.size(); i++) {
        for (int64_t j = 0; j < from_buffer.length_[i]; j++) {
          rawptr[k] = static_cast<PRIMITIVE>(from_buffer.ptr_[i].get()[j]);
          k++;
        }
      }
      return GrowableBuffer<PRIMITIVE>(length, std::move(ptr), length, length);
    }
```

src/awkward/_v2/cpp-headers/GrowableBuffer.h
```diff
@@ -0,0 +1,298 @@
...
+
```
ianna:
```suggestion
    /// @brief Takes this (possibly multi-panels) GrowableBuffer<PRIMITIVE>
    /// and makes another (one panel) GrowableBuffer<TO_PRIMITIVE>.
    ///
    /// Used to change the data type of buffer content from `PRIMITIVE`
    /// to `TO_PRIMITIVE` for building arrays.
    template<typename TO_PRIMITIVE>
    GrowableBuffer<TO_PRIMITIVE>
    copy_as() {
      auto len = length();
      auto ptr = std::unique_ptr<TO_PRIMITIVE>(new TO_PRIMITIVE[len]);
      TO_PRIMITIVE* rawptr = ptr.get();
      int64_t k = 0;
      for (size_t i = 0;  i < ptr_.size(); i++) {
        for (int64_t j = 0; j < length_[i]; j++) {
          rawptr[k] = static_cast<TO_PRIMITIVE>(ptr_[i].get()[j]);
          k++;
        }
      }
      return GrowableBuffer<TO_PRIMITIVE>(len, std::move(ptr), len, len);
    }
```

ianna (changes requested):
Ok, my compiler does not like that the static function accesses the private members. Here is an alternative implementation.

ianna (approved):
This looks good! Thanks. Please, merge it when the tests pass.

jpivarski:
@all-contributors please add @ManasviGoyal for code

PR #1538, opened 2022-07-07
Solving the endian bug on MacOS

aryan26roy:


jpivarski:
The bug is actually not related to endianness, but some compiler optimization that optimizes away our `byteswap` function. Mac has the same endianness as all other x86's, but the function that actively swaps was somehow not being invoked.

jpivarski:
@all-contributors please add @aryan26roy for code

PR #1539, opened 2022-07-08
Fix: ak._v2.is_none check for axis value 

### LINKED ISSUE #1417: ak._v2.is_none should reject axis values that are too large
### Version of Awkward Array

HEAD

All of the following are fine because `array` has an `axis=0`, `axis=1`, `axis=2`:

```python
>>> import awkward as ak
>>> array = ak._v2.Array([[[1, 2, None], None], None])
>>> ak._v2.is_none(array, axis=0).show()
[False,
 True]
>>> ak._v2.is_none(array, axis=1).show()
[[False, True],
 None]
>>> ak._v2.is_none(array, axis=2).show()
[[[False, False, True], None],
 None]
```

But this shouldn't be allowed to work:

```python
>>> ak._v2.is_none(array, axis=3).show()
[[[1, 2, None], None],
 None]
>>> ak._v2.is_none(array, axis=3)
<Array [[[1, ...], ...], ...] type='2 * option[var * option[var * ?int64]]'>
```

The output type isn't right, either: it should only ever be `bool` at the level of the `axis`.

ioanaif:
#1417


jpivarski (approved):
Great, thank you! (I also tested the negative axis cases, and everything is as it should be.)

PR #1548, opened 2022-07-14
DeprecationWarning fix; sre_{constants,parse} API made private in 3.11

douglasdavis:
Long thread here: https://github.com/python/cpython/issues/91308


douglasdavis:
I see this an automatically generated file now!

PR #1550, opened 2022-07-16
ci: use GHA instead of Azure :hammer: 

agoose77:
This PR moves the buildtest CI to GitHub Actions, and drops the doctest workflow.

My motivation for tackling this is that I am also moving the docs to GHA in a separate PR.

I've not used Azure before, and so this might not 100% match on the trigger logic. The exclusion logic looks as though it should never run, but clearly it does in `main`, so I must be missing something.

This PR sets the action to run on PR, like the tests.


agoose77:
@jpivarski not sure who else to tag on this. Thought perhaps Henry, but I see he's marked himself as "busy".

jpivarski:
Does this entirely switch from Azure to GitHub Actions? I had thought there were some blockers preventing us from doing that, but it looks like all of the tests are running here. I'll need to change the required tests in `main`'s branch protection before we can merge it, and the other active PRs should adopt this before they merge, so this will take a little time.

As such, let's ask @henryiii to review this. He's busy this week with Snowmass, but maybe he'll get a chance to look at it next week. Meanwhile, I'll spread the word that this switchover is happening and to be ready for it. This would be a good reason for Awkward Array project meetings (or just "Awkward Meetings"?), for us all to keep in touch with what everyone else is doing. We might have a regular time for that by the end of August.

henryiii:
I'll look at it later, but pretty sure there weren't blockers, it was just work that had never been done. Azure is just a earlier design of GHA.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,176 @@
...
+  cancel-in-progress: true
+
+jobs:
+  Windows:
+    strategy:
+      matrix:
+        python-version:
+        - '3.9'
+        - '3.10'
+        - '3.9'
```
henryiii: `3.9` dup?

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,176 @@
...
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: true
+    - name: Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v4
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install Pip
+      run: python -m pip install --upgrade pip
```
henryiii: Pip version is pretty well kept up to date, IMO.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,176 @@
...
+  MacOS:
+    strategy:
+      matrix:
+        python-version:
+        - '3.10'
+        - '3.9'
+        - '3.8'
+        - '3.7'
+        - '3.6'
+    runs-on: macOS-10.15
```
henryiii:
```suggestion
    runs-on: macOS-11
```

10.15 is deprecated.

.github/workflows/build-test.yml
agoose77: Ah - missed the arch :warning: 

Do you know why we had an x86 test?

agoose77:
> Does this entirely switch from Azure to GitHub Actions? 

Only buildtest, but the doctest will also be moved in a separate docs PR

agoose77:
I think I'll need to revisit this in the near future, once the docs are being built on GHA too. However, I think at that point we can look at whether an optimisation pass is necessary; right now, we will build one architecture 3X: once for docs, once for tests, and once for the wheel build test. Maybe that is OK, but regardless, a future problem!

agoose77:
I believe this is nearly ready to go. Just needs some input as to the triggers - do we want to run this on every PR push? Right now, that is what's happening on Azure, though I'm not sure if it is intentional

jpivarski:
This is great, but let's actually merge it on an off-week. At the same time that it is merged, we'll have to switch the `main` branch protection to expect these tests, rather than the Azure ones, so everyone will need to merge this PR to get passing tests.

The first week of August (1‒5) is an off-week because several of us will be at CoDaS-HEP.

People with open PRs are you, me, @ianna, @ManasviGoyal, @aryan26roy (though he'll be done by August), and @martindurant.

martindurant:
Mine can be merged, and I can iterate, right?

jpivarski:
> Mine can be merged, and I can iterate, right?

Okay, I didn't know it was declared ready. I'll try merging it with main to run the tests again, and I'll give it another scan, then merge it.

jpivarski:
I just brought this up to date because it will be merged next week, probably right after our first group meeting. That would be a good way to warn everybody.

jpivarski:
@agoose77 Actually, I should be able to remove the `azure-doctest-awkward.yml` from this also (i.e. no more `.ci` directory and no Azure tests at all). The documentation tests are minimal compare with what you're doing in the `docs` branch, and that's where all the new development is going.

Any reason why you kept it? If you're leaving it up to me, I'll delete it and we'll go entirely onto GitHub Actions (good for simplicity).

agoose77:
I removed the doctest CI in the docs branch, but I was leaving it here for now. I don't mind if you want to remove it here too!

jpivarski:
I forgot about this; I'll have to do it on the weekend.

PR #1552, opened 2022-07-19
Docs: merge JB and Sphinx documentation

agoose77:
This is the first step in merging these two sites. As such, the PR is quite small - later work will add content and further restructuring

Flowchart of PRs in this project:
```mermaid
graph LR;
integrate[Integrate JB & Sphinx]
improve[Improve UX of integration]
content[Expand user-guide content]
integrate --> improve --> content
style integrate stroke:#f66,stroke-width:2px
click integrate "https://github.com/scikit-hep/awkward/pull/1552" "This PR"
click improve "https://github.com/scikit-hep/awkward/pull/1563" "Improvement PR"
click content "https://github.com/scikit-hep/awkward/pull/1575" "Content PR"
```

## Meaningful changes
- [x] Remove references to awkward-array.org
- [x] Move `docs-src` under `user-guide` section of `docs-sphinx`
- [x] Use `pydata` Sphinx theme
- [x] Deploy using GHA (remove Azure) 
- [x] Update PAT on RTD to extend lifetime & make note in future to renew.
- [x] Deploy branch rather than PR
- [ ] Remove existing RTD webhook

```mermaid
graph LR;

build["Build on GHA"]
upload["Upload .jupyter_cache to artefacts"]
webhook["Trigger RTD via webhook"]
pull["Pull artefact from GHA"]

build --> upload --> webhook
webhook --> pull
```

https://awkward-array.readthedocs.io/en/feat-merge-docs/


agoose77:
So, it's looking like RTD doesn't expose environment variables to custom builds; this will prevent us from pulling the artefacts from GHA.

I think at this stage it might be more straightforward to perform the build on RTD, but pre-execute the notebooks and package building on the CI, i.e. publish _those_ artefacts and then use them on RTD.

That would mean publishing a built wheel(s) and the JB execution cache. Note that MyST-nb also supports and _inline_ execution mode which does not work with caching currently. We don't use that anywhere at the moment, though.

jpivarski:
One thing: don't plan on compiling Awkward on the RTD server. It runs over an hour and the server gets shut down, in a way that doesn't feedback to the CI check.

I liked your idea about compiling Awkward and then building the docs on a CI machine (GitHub Actions), especially if that job fails if the documentation can't be built. Couldn't we have everything run on the CI machine, produce static HTML as an artifact, and then pick up that artifact on the RTD server? Can RTD be triggered by a CI job finishing (e.g. Linux commandline), rather than a GitHub commit or release? That way, we can ensure that the artifact is populated before starting RTD.

That wouldn't involve any environment variables, and RTD would be used minimally; we won't have to rely on its capabilities at all. Maybe it would involve two Sphinx configurations, one to really build it in CI and the other to take the static HTML verbatim?

(Inline execution, and even execution of notebooks for JupyterBooks, are second priority to having the documentation built at all. If you're trying to compile Awkward in RTD, it will not get built at all.)

agoose77:
> One thing: don't plan on compiling Awkward on the RTD server

Yes, I can reassure you that that's not on the table!

The original plan was a full HTML build on the CI, and then have a very simple hook expose that to RTD: https://github.com/scikit-hep/awkward/blob/e0f38ed5b5624b1b2a61054414ad8c69aaf29632/.readthedocs.yml

That's where the webhook comes in - the CI triggers the RTD "build" (artefact download) once it has finished.

Unfortunately, [a bug in RTD](https://github.com/readthedocs/readthedocs.org/issues/9438#issuecomment-1189345456) means that we can't actually retrieve the built assets - we need an access token, and there's no safe way to inject the token without env-var support. At least, I can't think of a workaround without a third-party intermediary

The other option is to make the needed assets available for fast builds. Right now (ignoring my bigger PR that will use autodoc), we can get away with executing the jupyter-book on the CI, and then passing the cache to RTD. That will avoid executing any notebooks and obviate the need for the awkward library. Because this operates within the usual machinery, we can use the aforementioned mechanism to retrieve the built assets. 

There are benefits to this approach - this lets RTD integrate more nicely with our docs, where I was originally thinking we'd need to replicate the search functionality etc.

agoose77:
OK. This PR is now ready for review. It notifies RTD to perform a full Sphinx build. The jupyter-cache output is generated before this notification and uploaded to the workflow artefacts. Additionally, Sphinx is set to fail if any notebooks fail to execute.

agoose77:
I've done very little content work besides the bare minimum in order to guarantee that this is a small PR that we can merge more easily. 

See https://awkward-array.readthedocs.io/en/feat-merge-docs/ for the current build

agoose77:
Note that, we could later decide to only build on GHA if RTD fix the bug, or we use an intermediary store for the artefact. I would like to exclusively build on GHA in the long run, as it simplifies certain things.

But, this is deliverable "today" :)

dev/download-github-artefact
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
```
jpivarski: Perhaps this script filename should end with .py and it should be invoked by an explicit `python` call, so that we control which Python executable runs it?

jpivarski (approved):
Do I understand correctly that this PR does all of the following?

   * Moves doc-testing from Azure to GitHub Actions.
   * A GitHub Actions worker does the actual Sphinx-build and triggers readthedocs to take its artifact.
   * Converts all URL references from `https://awkward-array.org` to `https://awkward-array.readthedocs.io`. After this PR, there's no reference to `https://awkward-array.org`.
   * Removes the big, image-based panels that send the user to Python reference or C++ reference from GitHub README, PyPI README, and the quickstart page.
   * Moves most (all) of the `docs-src` (JupyterBook) into `docs-sphinx` (Sphinx) to run the whole thing from Sphinx.

If I merge this PR, what happens to the currently live documentation? Is it the case that `https://awkward-array.readthedocs.io` will have all of the v1 documentation, as it is now (I don't see source file changes) but `https://awkward-array.org` will become stale? Would it be appropriate for me to redirect `https://awkward-array.org` to point to `https://awkward-array.readthedocs.io` so that they become synonyms, and I can cancel Netlify?

Once `https://awkward-array.org`, you may need to convert all of the URLs back, since this is the advertised URL. (`https://awkward-array.org/quickstart.html` should also point at the landing page, even if it becomes a synonym for `https://awkward-array.org`.)

When would new documentation go to readthedocs? With every commit, with every PR, or with every release/tag? Ideally, it should be tested in every commit and deployed to readthedocs with every release/tag.

Can I see the new documentation before committing? Is it on an unlisted readthedocs URL?

(Beyond that, bravo—this is a big undertaking!)

dev/download-github-artefact
agoose77: > Do I understand correctly that this PR does all of the following?

Yes, although we also update the theme to the new pydata theme. We don't yet take full advantage of this / make any sweeping design changes; that's down the road.

> If I merge this PR, what happens to the currently live documentation? 

Any activity on a PR will update that PR branch on RTD. The per-branch visibility is controlled on RTD side. However, I haven't yet set up the building from `main` - I wanted to run the rest of this PR past you first! It might make sense to trigger this workflow on `branch` activity rather than PRs.

All doc-building is configured from GitHub via our CI and the `.readthedocs.yaml` configuration. 

> Would it be appropriate for me to redirect https://awkward-array.org to point to https://awkward-array.readthedocs.io so that they become synonyms, and I can cancel Netlify?

Yes - and I'd suggest that we make use of custom domains to have `awkward-array` point to the RTD: https://docs.readthedocs.io/en/stable/custom-domains.html

> When would new documentation go to readthedocs? With every commit, with every PR, or with every release/tag? Ideally, it should be tested in every commit and deployed to readthedocs with every release/tag.

Currently with every "event" on a PR. This should probably change to a per-branch workflow which is more intuitive (and will then naturally pick-up merges). TODO...

 > Can I see the new documentation before committing? Is it on an unlisted readthedocs URL?
 
 The docs are live, but unlisted: https://awkward-array.readthedocs.io/en/feat-merge-docs/index.html

agoose77: > Perhaps this script filename should end with .py and it should be invoked by an explicit python call, so that we control which Python executable runs it?

Sure! I have a preference for executables, but it's really no big deal either way.

agoose77: NB: we'll have 50% failing builds until we remove the built-in webhook from RTD, as it does not wait for the artefacts to be deployed by the docs workflow

agoose77:
@jpivarski (for Monday during work hours) - how does this PR look at the moment? I'm currently finished on the first set of deliverables. It's not finished in the sense that these are "final" docs, but I think it's mergeable, and I intend to rebase my larger PR on top of this (well, probably manually cherry-pick) in order to start on phase 2.

jpivarski:
I only haven't done it yet because I was nervous about what would happen to the docs when this goes in. The CoDaS-HEP tutorial is a little more than a week away, and I'll want to make frequent reference to the docs.

From a discoverability standpoint, I was at first at a loss because I couldn't find the API documentation:

image

I eventually found it in the bar at the top, and I knew what I was looking for. Could the three main books, "User Guide", "API Reference", and "C++ Reference", also be in the left-bar of the landing page? Usually, I don't like having multiple ways of doing the same thing, but being able to find the right book on the landing page is much more important than being able to switch between books on any other page. (I like the fact that we _can_ switch between books on other pages, but getting to the right one quickly is important on the landing page. Some users will have never seen this page before and probably don't even know that what they're looking for is called "API Reference", while others will be hitting it for the 20th time today and need to get to what they're looking for quickly. That's why I originally made them such big image-buttons, and why NumPy/SciPy/Pandas do the same with those four icons.)

(... Have to go now! Arg! ...)

"C++ Reference" doesn't have to be at the same level as the other two, and removing it may make it easier to see the others—less noise. A link to the C++ reference can be hidden at the end of the quickstart, in a bullet point or something. The names can be simplified, too: "Tutorials" and "Reference"? (Not everybody would immediately recognize "API".)

In the User Guide/Tutorials, the subsections in the left-bar are not expanded by default (as in the old awkward-array.org), and I think it would help a lot if they were.

In the API Reference/Reference, the subsections in the left-bar never expand, and they're pretty important for some of the pages (ak.Array, ak.behavior, Forth, ...).

Sorry for so much back and forth, but changing the docs is like unveiling a new website!

agoose77:
> Could the three main books, "User Guide", "API Reference", and "C++ Reference", also be in the left-bar of the landing page?

I don't think this will be very easy. The top bar in this theme is the *root* toc-tree. The left bar is then responsible for displaying the second-level toctree, which changes per-page. I don't believe that it's possible in Sphinx (I believe it's Sphinx, anyway) to set this second-level toctree for the root document (index). That's why (I assume) all pydata themes don't have any side-bar content for the index. I don't believe this *will* be a problem, as I'll discuss in response to another point!

> Usually, I don't like having multiple ways of doing the same thing, but being able to find the right book on the landing page is much more important than being able to switch between books on any other page.

I agree with this assessment

> That's why I originally made them such big image-buttons, and why NumPy/SciPy/Pandas do the same with those four icons.)

That's what I am suggesting is the solution here instead of the left-bar (see #1563), This PR is *just* about the bare minimum to take what we currently have and port it to a unified theme. I was originally doing everything in a single PR, but I didn't feel it was accessible as a reviewer, so I've split it. If you'd feel happier delaying it until after CoDaS HEP, that's no problem!

> "C++ Reference" doesn't have to be at the same level as the other two, and removing it may make it easier to see the others—less noise. 

I decided on putting C++ reference in the top bar after deciding what needed to go in the cards view shown in #1563. I'd be interested to hear your thoughts given the proposed cards UX.

> The names can be simplified, too: "Tutorials" and "Reference"? (Not everybody would immediately recognize "API".)

I'm leaning heavily into the "standard" established by PyData here:

image
image
image

My guiding principle here is align with the scipy ecosystem unless we need to do something different. Does that sit with you?

> In the User Guide/Tutorials, the subsections in the left-bar are not expanded by default (as in the old awkward-array.org), and I think it would help a lot if they were.

Agreed. That's a todo. I can make it a todo of this PR if required.

> Sorry for so much back and forth, but changing the docs is like unveiling a new website!

No, quite right. I actually held off creating #1563 over the weekend, but I realise now that it just created more confusion as to where this PR was going.

agoose77:
Aside: I've been discussing certain parts of this on Slack with you. Now that there's a direction, I'll lean towards GH PR discussions so that
- It's more transparent
- Less duplication between conversation threads

:)

agoose77:
OK - I've created a `docs` branch that this will merge into. Now, we can feel more confident about merging iterative PRs that might undergo reasonable amounts of change over the course of their development, without worrying about what `latest` users will see.

agoose77:
@jpivarski are you happy to merge this into `docs`, and move to the second PR to discuss further iterations?

The purpose of this PR is just to build the scaffolding (CI integration, JB support, etc.) I picture the docs branch as being a moving target, with small PRs that let us discuss small groups of changes. We could do this with one *large* PR, but I think that can get quite unwieldy w.r.t review.

agoose77:
Oh, wait - you've already approved this :facepalm: Not sure why I didn't see the notification ...

I'll merge.

PR #1554, opened 2022-07-19
layout builder python bindings

ianna:


tests-cpp/test_1494-layout-builder.cpp
```diff
@@ -0,0 +1,1375 @@
...
+    {Field::one, "one"},
+    {Field::two, "two"},
+    {Field::three, "three"}});
+
+  RecordBuilder<
+      lb::Field<Field::one, NumpyBuilder<double>>,
+      lb::Field<Field::two, NumpyBuilder<int64_t>>,
+      lb::Field<Field::three, NumpyBuilder<char>>
+  > builder(fields_map);
+
```
ianna, line 387: @jpivarski - I wonder if this is more convenient for a user

tests-cpp/test_1494-layout-builder.cpp
```diff
@@ -0,0 +1,1375 @@
...
+  auto names = builder.field_names();
+  for (auto i : names) {
+    std::cout << "field name " << i << std::endl;
+  }
+
+  auto& one_builder = builder.field<Field::one>();
+  auto& two_builder = builder.field<Field::two>();
+  auto& three_builder = builder.field<Field::three>();
+
```
ianna, line 396: this is how the record builders would be accessed

tests-cpp/test_1494-layout-builder.cpp
```diff
@@ -0,0 +1,1375 @@
...
+          lb::Field<Field2::w, NumpyBuilder<char>>>>
+  > builder;
+  builder.set_field_names(fields_map0);
+
+  auto& x_builder = builder.field<Field0::x>();
+  x_builder.set_field_names(fields_map1);
+
+  auto& y_builder = builder.field<Field0::y>();
+  y_builder.set_field_names(fields_map2);
+
```
ianna, line 603: The drawback is setting the maps for each record builder individually.

ianna (commented):
@jpivarski - I guess, I can have one map with offsets to pass to each record builder. Any thoughts? Thanks.

jpivarski:
I'm confused about what this PR is doing: the title is "layout builder python bindings," but it doesn't define Python bindings for LayoutBuilder. (Such a thing wouldn't be all that useful, since LayoutBuilder must be used in compiled C++ to be useful.)

ianna:
close for now

Issue #1439, opened 2022-04-22
ufuncs on records should not be allowed unless they are explicitly overridden.


jpivarski:
Unless explicitly overridden, ufuncs are not supposed to work on records. In v1:

```python
>>> array = ak.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}])
>>> np.absolute(array)
Traceback (most recent call last):
...
  File "/home/jpivarski/irishep/awkward-1.0/awkward/_util.py", line 1058, in apply
    raise ValueError(
ValueError: cannot broadcast records in this type of operation

(https://github.com/scikit-hep/awkward-1.0/blob/1.9.0rc2/src/awkward/_util.py#L1060)
```

(that's good), but in v2:

```python
>>> array = ak._v2.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}])
>>> np.absolute(array)
<Array [{x: 1, y: 1.1}, {x: 2, ...}] type='2 * {x: float64, y: float64}'>
```

(that's bad).

In v1, this was issue #457 and PR #510, which added:

https://github.com/scikit-hep/awkward-1.0/blob/779819e0405cd481dbae0976fc311f2f1d91d9b5/src/awkward/_util.py#L1057-L1061

It looks like the check was copied into v2:

https://github.com/scikit-hep/awkward-1.0/blob/779819e0405cd481dbae0976fc311f2f1d91d9b5/src/awkward/_v2/_broadcasting.py#L653-L656

but it's not being triggered in the above example for some reason.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward-1.0/pull/1437#discussion_r856339232_

PR #1559, opened 2022-07-21
Fix: ufuncs on records should not be allowed unless overridden

ioanaif:
#1439


ioanaif:
Now that ufuncs are no longer allowed for records in v2, test `0355-mixins` fails with`Error details: cannot broadcast records in absolute` @jpivarski

jpivarski:
While it's possible that we'll find some tests that assume you can apply ufuncs (like `np.abs`) to generic records and those tests have to be changed or removed, this particular test _should_ still work.

In the definition of Point, a method was defined for `np.abs`:

https://github.com/scikit-hep/awkward/blob/afe5fc166a78b2d062d59811e91a0a0c3fc17dad/tests/test_0355-mixins.py#L18-L20

The rule should be that records can't be used in ufuncs unless a method has been defined for that particular ufunc, as is the case here. That's what `broadcast_and_apply` does; `recursively_apply` should follow the same rule. (That is, `np.add(record, record)` should fail if `record` does not have an implementation of `np.add`, but it should use that implementation if it does.)

tests/v2/test_1559_fix_ufuncs_records_1439.py
```diff
@@ -0,0 +1,37 @@
...
+import awkward as ak  # noqa: F401
+import numpy as np  # noqa: F401
+
+to_list = ak._v2.operations.to_list
+
+
+def test_ufuncs_on_records_1439():
+    array = ak._v2.Array([{"x": 1.0, "y": 1.1}, {"x": 2.0, "y": 2.2}])
+    with pytest.raises(ValueError):
+        np.absolute(array)
```
jpivarski: It's good that this complains if `np.absolute` is not overloaded, but it should also not complain if `np.absolute` is overloaded. (One more test.)

jpivarski (approved):
Yes! `recursively_apply` should be just like `broadcast_and_apply` in requiring `behavior` and not allowing records unless the records are overloaded.

Just need a test to be sure that `np.absolute` will work if it is overloaded (like `np.add`).

Once that test has been added and it passes, go ahead and merge.

PR #1560, opened 2022-07-22
builder options and layout builder updates

ianna:
`ArrayBuilder`, `LayoutBuilder`, and `GrowableBuffer` are using `BuildOptions` now. There two options defined at the moment - an `initial` which is `int64_t` and a `resize` factor which is `double`. The `Options` is a tuple and can accept other parameters. A test is added.

This PR also drops requirements on C++ version for `LayoutBuilder` from C++17 to C++14. Both `GrowableBuffer` and `BuildOptions` can be used with C++11. 

`BitMasked` mask is fixed.

### DIFF
--- a/src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
+++ b/src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
@@ -20,26 +20,22 @@ namespace awkward {
   class Panel {
   public:
     Panel(size_t reserved)
-      : ptr_(std::unique_ptr<PRIMITIVE[]>(new PRIMITIVE[reserved])),
-        length_(0),
-        reserved_(reserved),
-        next_(nullptr) {  }
+        : ptr_(std::unique_ptr<PRIMITIVE[]>(new PRIMITIVE[reserved])),
+          length_(0),
+          reserved_(reserved),
+          next_(nullptr) {}
 
     Panel(std::unique_ptr<PRIMITIVE[]> ptr, size_t length, size_t reserved)
-      : ptr_(std::move(ptr)),
-        length_(length),
-        reserved_(reserved),
-        next_(nullptr) {
-    }
+        : ptr_(std::move(ptr)),
+          length_(length),
+          reserved_(reserved),
+          next_(nullptr) {}
 
-    PRIMITIVE &operator[](size_t i) {
-      return ptr_.get()[i];
-    }
+    PRIMITIVE& operator[](size_t i) { return ptr_.get()[i]; }
 
     Panel*
     append_panel(size_t reserved) {
-      next_ = std::move(std::unique_ptr<Panel>(
-        new Panel(reserved)));
+      next_ = std::move(std::unique_ptr<Panel>(new Panel(reserved)));
       return next_.get();
     }
 
@@ -66,13 +62,15 @@ namespace awkward {
 
     void
     concatenate_to(PRIMITIVE* to_ptr, size_t offset) const noexcept {
-      memcpy(to_ptr + offset, reinterpret_cast<void*>(ptr_.get()), length_ * sizeof(PRIMITIVE));
+      memcpy(to_ptr + offset,
+             reinterpret_cast<void*>(ptr_.get()),
+             length_ * sizeof(PRIMITIVE));
       if (next_) {
         next_->concatenate_to(to_ptr, offset + length_);
       }
     }
 
-    template<typename TO_PRIMITIVE>
+    template <typename TO_PRIMITIVE>
     void
     copy_as(TO_PRIMITIVE* to_ptr, size_t offset) {
       for (size_t i = 0; i < length_; i++) {
@@ -131,9 +129,11 @@ namespace awkward {
       if (actual < minreserve) {
         actual = minreserve;
       }
-      return GrowableBuffer(options,
-        std::unique_ptr<PRIMITIVE[]>(new PRIMITIVE[(size_t)actual]),
-        0, actual);
+      return GrowableBuffer(
+          options,
+          std::unique_ptr<PRIMITIVE[]>(new PRIMITIVE[(size_t)actual]),
+          0,
+          actual);
     }
 
     /// @brief Creates a GrowableBuffer in which all elements are initialized to `0`.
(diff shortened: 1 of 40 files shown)


ianna:
@jpivarski - please, check when you have time. Thanks!

ianna:
@jpivarski - I’m done for today. If you have time to review it, please do. Thanks

ianna:
@jpivarski - this PR is complete. Please, check when you have time. Thanks.

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -302,7 +334,7 @@ namespace awkward {
     void
     append(PRIMITIVE datum) {
       if (ptr_->current_length() == ptr_->reserved()) {
-        add_panel((size_t)ceil(ptr_->reserved() * 1.5));
+        add_panel((size_t)ceil(ptr_->reserved() * options_.resize()));
```
jpivarski, line 337: So right now, the algorithm is that each panel is a constant factor larger than the previous, so they'll grow exponentially. That made sense for the replacing-panel algorithm (so that there are logarithmically many replacements), but for the new discontiguous-panel algorithm, we want the panel size to grow less rapidly.

I think the best algorithm will probably be:

   1. Initial panel size is small, like 1024.
   2. Subsequent panel size (second onward) is large, like 1024*1024.

The actual values of the parameters (`initial` and `subsequent`) would be determined by performance tests: the criteria are (1) `initial` should be small enough that we can make a lot of them without running out of memory (~1 GB) and large enough that small arrays like unit tests don't trigger a second panel, and (2) `subsequent` should be large enough that << 1% of the `GrowableBuffer::append` time is spent managing panels; >> 99% is spent just copying data.

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -320,21 +352,21 @@ namespace awkward {
         for (size_t i = 0; i < empty_slots; i++) {
           fill_panel(ptr[i]);
         }
-        add_panel(size - empty_slots > ptr_->reserved() ?
-                  size - empty_slots : ptr_->reserved());
+        add_panel(size - empty_slots > ptr_->reserved() ? size - empty_slots
```
jpivarski: You know, the use of the word "slots" here is very different from the use of the word "slots" in RDataFrame. Here, it means "filled items," but in RDataFrame, it means "parallel processors."

Since these parts of the codebase are nearly adjacent (RDataFrame would sometimes use LayoutBuilder, which uses GrowableBuffer), a maintainer could get confused about these different meanings of the word "slots." Not necessarily now, but we should plan on replacing this usage of "slots" with another word. `empty_slots` → `unfilled_items`?

tests-cpp/test_1560-builder-options.cpp
```diff
@@ -0,0 +1,14 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+#include "awkward/BuilderOptions.h"
+#include <cassert>
+
+int main(int /* argc */, char ** /* argv */) {
+
+  awkward::Options<int64_t, double> options(1024, 1.0);
```
jpivarski: For a C++ user, this should have a name (or be in a namespace) that indicates that this object represents options for builders, so `awkward::BuilderOptions` or `awkward::LayoutBuilder::Options`.

jpivarski (commented):
As far as I can see, this PR does a few things:

   * It passes BuilderOptions through a lot of constructors, rather than a single integer (`initial`). That will allow us to expand the options, such as a distinction between the initial GrowableBuffer panel size and subsequent ones. (We'll want to make those arguments not template parameters in LayoutBuilder, since I doubt an object like BuilderOptions can be a compile-time constant. ~~If LayoutBuilder needs to have a default constructor, maybe the default constructor can supply default BuilderOptions?)~~ I see that you did that already.
   * Various clean-ups, which cause indentation diffs.

On closer inspection, it looks to me like the only substantial change is that `initial` → `options`, with all the supporting changes that needs. Is this right? (Have I missed anything?)

The only thing I would say needs to be changed is that a C++ user sees these options with a class name like `awkward::BuilderOptions`, not `awkward::Options`, because these options are specifically for builders.

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
ianna: The default for `resize` is 1.

jpivarski: Okay, good: so for now the "exponential growth" is a constant ($\exp(t)$ is an exponential!). But the thing to get addressed later is that it's the wrong algorithm. For later, anyway.

ianna:
> As far as I can see, this PR does a few things:
> 
> * It passes BuilderOptions through a lot of constructors, rather than a single integer (`initial`). That will allow us to expand the options, such as a distinction between the initial GrowableBuffer panel size and subsequent ones. (We'll want to make those arguments not template parameters in LayoutBuilder, since I doubt an object like BuilderOptions can be a compile-time constant. ~If LayoutBuilder needs to have a default constructor, maybe the default constructor can supply default BuilderOptions?)~ I see that you did that already.
> * Various clean-ups, which cause indentation diffs.
> 
> On closer inspection, it looks to me like the only substantial change is that `initial` → `options`, with all the supporting changes that needs. Is this right? (Have I missed anything?)
> 

just small stuff:
- `std::void_t` has been replaced as it is C++17
- `BitMasked` mask was fixed. That required additional `GrowableBuffer` concatenate method with an offset defined.
- cleanup LayoutBuilder API - remove temporary/legacy `to_buffer` method

> The only thing I would say needs to be changed is that a C++ user sees these options with a class name like `awkward::BuilderOptions`, not `awkward::Options`, because these options are specifically for builders.

tests-cpp/test_1560-builder-options.cpp
ianna: It's actually a tuple of options with an `awkward::BuilderOptions` specialisation that is a `awkward::Options<int64_t, double>`

ianna: eventually we can type alias other specialisations and the getters

src/awkward/_v2/cpp-headers/awkward/GrowableBuffer.h
ianna: done

tests-cpp/test_1560-builder-options.cpp
ianna: done

jpivarski (approved):
Okay, great—it's ready to merge, then!

PR #1563, opened 2022-07-24
Docs: rework content for usability + UX

agoose77:
This PR builds upon #1552 to make the docs more discoverable and provide helpful answers to "what is an awkward array", etc. It addresses some of the tasks in https://github.com/scikit-hep/awkward/projects/7#card-84465881

Flowchart of PRs in this project:
```mermaid
graph LR;
integrate[Integrate JB & Sphinx]
improve[Improve UX of integration]
content[Expand user-guide content]
integrate --> improve --> content
style improve stroke:#f66,stroke-width:2px
click integrate "https://github.com/scikit-hep/awkward/pull/1552" "Improvement PR"
click improve "https://github.com/scikit-hep/awkward/pull/1563" "This PR"
click content "https://github.com/scikit-hep/awkward/pull/1575" "Content PR"
```

The scope for this PR is small:
- Make better use of PyData theme
- Rework existing pages to better reflect their new locations e.g. API-reference, remove duplicate icons
- Preliminary work on beginner guides and contribution guides

Notably, this adds a panel-based front page:
image
image

https://awkward-array.readthedocs.io/en/docs-phase-2/


jpivarski:
This is looking really good (especially those four icons!), though I'm even more confused about which PRs are now active. (Does this supercede #1552?)

Since this is changing our public website, perhaps we should get a pool of people to review it, especially users of different experience levels. If there's anyone we know who haven't used Awkward Array before, it could be good to point them at the new docs and see if they find what they're looking for.

From my side (I'm looking at this with a laptop again), I still think it's important to load up the left-bar. A complex front page with lots of detail is self-defeating, but a long list of things in the left-bar is not: it's off to the side, out of the way, but available for an in-page search or direct click for people who know what they're looking for. Also, I think the set of things in that left-bar should be stable: the same set, regardless of which page you're looking at. Although the four icons on the front page divide the docs into Beginner's Guide, User's Guide, API Reference, and Contributor's Guide, it can be a single document from the perspective of the left-bar. It's possible to put section-headings in the left-bar (as you have: Beginner's Guide and User's Guide share a left-bar, but they all could). Strangely, clicking on Beginner's Guide goes to the _second_ page in that first group (skips "What is a...").

With the four icons, we have four major divisions, but the top bar still says User Guide, API Reference, C++ Reference. Having the top bar be these four major divisions would cement the distinctions. And again, C++ Reference shouldn't be so highlighted.

Within the Guides, the "Converting arrays," "Creating new arrays," sub-sections are not expanded by default. If that's hard to do, it would be better to not have sub-sections: it would be better to have the pages like "Converting to/from NumPy," "Converting to/from Python objects" be top-level within the User's Guide. It's more important for a word like "NumPy" to be findable by an in-page search or at a glance when scrolling the left-bar than it is for "Converting" to be. The user with a NumPy array might not think of or recognize "Converting" as what they want, but would recognize "NumPy" (it's more specific).

Super-minor, but in

image

I don't think the badges need to be boxed. All my eyes notice is that the box doesn't line up with the four icons boxes.

In fact, we don't even need the badges there. (I'm not very convinced of the usefulness of the badges in general, since they all look the same to me, but maybe they can be on the GitHub README only, and not the other front-pages.) [NumPy](https://numpy.org/doc/stable/), [SciPy](https://docs.scipy.org/doc/scipy/), and [Pandas](https://pandas.pydata.org/docs/) all use this space for version, date, download docs, and useful links. The less we have above the four icons, the better, because that improves the chances that all four icons will be visible on someone's browser window. (I, for one, will always be clicking through to API reference because I'm giving people links to which function they should use. If there's too much stuff above the icons, I have to scroll down to get to that box. Even the text under each of the four icons can be minimized.)

agoose77:
> This is looking really good (especially those four icons!), though I'm even more confused about which PRs are now active. (Does this supercede #1552?)

I'm rolling out a series of focused PRs that we can review independently. I was originally working on a single branch but it would never really be "done", and it would be nearly impossible to review from the diffs alone. Once #1552 is merged, this changeset will get smaller :) You can see in the issue description a diagram that indicates where in the "series of PRs" each PR fits.

> Since this is changing our public website, perhaps we should get a pool of people to review it, especially users of different experience levels. If there's anyone we know who haven't used Awkward Array before, it could be good to point them at the new docs and see if they find what they're looking for.

I agree in principle. My thought is that the earliest of PRs are just "boilerplate" / getting the plumbing in. The latter PR (#.3) is more content focused, and would be better for externals to review.

> From my side (I'm looking at this with a laptop again), I still think it's important to load up the left-bar. A complex front page with lots of detail is self-defeating, but a long list of things in the left-bar is not: it's off to the side, out of the way, but available for an in-page search or direct click for people who know what they're looking for. 

So (forgive me if I'm repeating parts of another thread, I'm not sure what we've both written and read) this is 50% design choice and 50% constraint. The theme that we're using here has merit because it's used by the pydata stack that users will already know. It follows that these sites do not have any (primary) sidebar content on their index page. Now, as it turns out, it is simple to add this: https://github.com/holoviz/holoviews/blob/master/doc/_templates/sidebar-nav-bs.html
You can see an example of what this looks like [here](https://holoviews.org/reference_manual/holoviews.core.data.html):
image

Also, to add - PyData will be releasing a new update to the theme that will remove the sidebar entirely on the index page:
image

So, the real question is whether to do it. I'm personally not a fan; content duplication is one thing, but the real reason I like this theme is that it makes the menu more discoverable by avoiding overloading the user with content. Putting everything in a single sidebar feels _less_ discoverable because you need to know where to look - if you have to scroll through a large list of items, the individual items are less noticeable. However, recognising that some users might prefer a "flat" list, I added the toctree to the index of the user guide which lists _all_ pages.

In particular, I feel that the different sections of the documentation as seen [here](https://awkward-array.readthedocs.io/en/docs-phase-2/index.html) serve different purposes. On the _api reference_ I agree that CTRL+F is a bonus, because I often know what I'm looking for and want to quickly check some parameters. Meanwhile, the user-guide should be more of a browsing experience - reading prose instead of scan-reading function signatures. By isolating the toctrees into per-navbar sections, we better reflect this.

E.g. I find this 
image
less readable than 
image

As for search, we have a good Sphinx-search which is the usual method for finding things on documentation sites powered by Sphinx.

> Also, I think the set of things in that left-bar should be stable: the same set, regardless of which page you're looking at. Although the four icons on the front page divide the docs into Beginner's Guide, User's Guide, API Reference, and Contributor's Guide, it can be a single document from the perspective of the left-bar. It's possible to put section-headings in the left-bar (as you have: Beginner's Guide and User's Guide share a left-bar, but they all could). 

If we do this (Sphinx calls these non-link headings "captions"), then we lose the ability to use captions to group anything else on the site. I feel like that might be a step backwards. I actually had these captions as links before, but I found captions more readable:
- we no longer need an "Index" link that doesn't do anything for each caption
- the text is bold and clearly a grouping 

> Strangely, clicking on Beginner's Guide goes to the _second_ page in that first group (skips "What is a...").

We can change this! But, [NumPy](https://numpy.org/doc/stable/) does this too. I can see the merit in linking to what-is-an-awkward-array because it's part of that section, and I would logically click "get started" If I didn't know what Awkward did.

However, I wonder whether we should carry on as we are, but instead link to "what is an awkward array" from a button on the index:
image


> With the four icons, we have four major divisions, but the top bar still says User Guide, API Reference, C++ Reference. Having the top bar be these four major divisions would cement the distinctions. And again, C++ Reference shouldn't be so highlighted.

Having the top bar forcibly reflect the four-card layout in the index feels like a reasonably severe constraint to me. It's already standard in PyData to have the primary nav (top bar) be separate to the cards (which are like "hot links"). Are you willing to move on this?

> Within the Guides, the "Converting arrays," "Creating new arrays," sub-sections are not expanded by default. If that's hard to do, it would be better to not have sub-sections: it would be better to have the pages like "Converting to/from NumPy," "Converting to/from Python objects" be top-level within the User's Guide. It's more important for a word like "NumPy" to be findable by an in-page search or at a glance when scrolling the left-bar than it is for "Converting" to be. The user with a NumPy array might not think of or recognize "Converting" as what they want, but would recognize "NumPy" (it's more specific).

I think I've mostly spoken to this a bit earlier w.r.t information overload & the usage differences between API ref and the user guide. I'd be happy to keep discussing it though :)

> Super-minor, but in
> 
> image
> 
> I don't think the badges need to be boxed. All my eyes notice is that the box doesn't line up with the four icons boxes.
> 
> In fact, we don't even need the badges there. (I'm not very convinced of the usefulness of the badges in general, since they all look the same to me, but maybe they can be on the GitHub README only, and not the other front-pages.) [NumPy](https://numpy.org/doc/stable/), [SciPy](https://docs.scipy.org/doc/scipy/), and [Pandas](https://pandas.pydata.org/docs/) all use this space for version, date, download docs, and useful links. The less we have above the four icons, the better, because that improves the chances that all four icons will be visible on someone's browser window. (I, for one, will always be clicking through to API reference because I'm giving people links to which function they should use. If there's too much stuff above the icons, I have to scroll down to get to that box. Even the text under each of the four icons can be minimized.)

Yes, I was hoping you'd say this! I've made use of the reclaimed space in the above mockup showing the suggested button to "what is an awkward array"

agoose77:
At a high-level, we're not restricted to only mimic PyData, despite that being something I reference a lot here. It's rather that I think there's a lot of value in keeping things as reasonably similar as possible.

For example, I actually don't love how NumPy's docs hide content by not showing a drop down:
image

Some of these options are actually sections with their own subtrees, but you can't as a user tell that because the drop-down is disabled.

Equally, for the API reference NumPy does not provide a flat list of contents, which impedes the "what else can I do?" workflow. I am on the fence as to whether we should change our API reference to also be sectioned, and have the generated toctree as the index. What we have for the api-reference index is already fairly flat. But, that's a later discussion!

jpivarski:
> E.g. I find this
>
> less readable than

Actually, I find the expanded-out more readable—for the navbar, at least. I don't think the content (central column) should duplicate it.

What's bothering me about it is that the names of individual pages are descriptive and helpful, and they've been grouped into (I think) meaningful sections, but the names of those sections were made up to fill in a blank. I don't think "Converting arrays" (and the others) is a helpful name. If the two-level structure is technically difficult, unaesthetic, or hard to read, then we should flatten it to just the pages.

All of this comes down to my belief (from experience of being a user) that users don't know exactly what they're looking for. That's why control-F is more useful than a search bar: incremental search is different from search in that it can help you find things that you don't know the whole word for. You type a little, backspace, type a little, backspace, and based on the whole-screen context of what the incremental search is interactively finding, you come to realize what it is you want to search for. In a search bar, you only find what you ask for.

(Also, that search facility is not great. It's better than GitHub's, but that's a low bar. It's rare for me to find what I'm looking for through a Sphinx search.)

The [original Awkward documentation](https://github.com/scikit-hep/awkward-0.x#readme) (and [Uproot](https://github.com/scikit-hep/uproot3#readme)) was one long page, to support that kind of discovery. Now there's too much to tell to actually get it on one page—zero click-throughs—but we can still get it pretty flat, like one or two click-throughs. At every click-through, a fraction of readers give up or follow the wrong path.

So I won't be a stickler on any visual arrangement, but it has to satisfy the constraint of minimizing the number of clicks from landing page to the information (flattening the tree). The current docs are failing at this, too: the fact that you have to click from awkward-array.org through "Python reference" to get to a page that has all of the functions listed (and then click on the function to learn about) is currently hiding information from users. A lot of user questions are answered by just saying that such-and-such a function exists—they're not finding it, and presumably, they're looking. I don't remember that happening as much with the single-page docs. (Then it was _other_ problems; Awkward 0.x had a lot of quirks.)

Even if the top bar is not the four sections, "C++ Reference" should not be up there. Having prominent links to places most people won't want to go takes away from getting them to where they do want to go. We should be thinking of this as funnelling.

All in all, I'm probably being too specific in my comments, since I'm probably missing good points from you by going round-robin between this and other things. I should leave you to have more autonomy in designing this. That's why in this comment, I tried to focus more on the high-level: why I'm wanting to flatten the tree/minimize the number of click-throughs. I hope it's more helpful than bewildering.

agoose77:
There's a lot to reply to here, so I think it's worth stating my assumptions up front
- We want to standardise our docs with big players
- We want to strike a balance between zero structure (and single click to any page) and click-fatigue / poor discoverability

Before we get into the weeds, here's an alternative API doc that uses the JB theme: https://docs.xarray.dev/en/stable/api.html

I like this theme _in general_, but I miss the ability to drop a level of nesting which the top bar provides. I see that as a bonus, not a disadvantage

> What's bothering me about it is that the names of individual pages are descriptive and helpful, and they've been grouped into (I think) meaningful sections, but the names of those sections were made up to fill in a blank. I don't think "Converting arrays" (and the others) is a helpful name. If the two-level structure is technically difficult, unaesthetic, or hard to read, then we should flatten it to just the pages.

So, yes - a big part of this is the actual names and grouping. I do think the current hierarchy is needed - there's just so much content to discuss. If we place all `how-to-convert` contents into a single document, we end up with a very large secondary sidebar (right hand side) that is less readable. Whereas, for xarray, which has fewer headings, it's fine: https://docs.xarray.dev/en/stable/user-guide/interpolation.html

I actually like how we currently break down topics. It might not be the "right" breakdown, but the amount of content per-documents feels digestible yet comprehensive. I don't think eliminating a level of toctree nesting would be a good fit here in terms of document size.

But, I haven't yet tackled these documents with any serious overhaul - again, trying to keep things lightweight & focused per PR. I picture doing this in a later (third) PR that looks mainly at content, and will probably be a longer one (unless I manage to break it down per-section). I'm aiming to have small, short lived PRs as I know the bigger they get, the longer they take to merge & review.

> All of this comes down to my belief (from experience of being a user) that users don't know exactly what they're looking for. That's why control-F is more useful than a search bar: incremental search is different from search in that it can help you find things that you don't know the whole word for. You type a little, backspace, type a little, backspace, and based on the whole-screen context of what the incremental search is interactively finding, you come to realize what it is you want to search for. In a search bar, you only find what you ask for.

I acknowledge this case, and mentioning incremental search is an important nuance that I'll keep in future discussions! Maybe this is something we should consider for our sphinx search too, e.g.: https://readthedocs-sphinx-search.readthedocs.io

I'll reply to the rest of this below.

> (Also, that search facility is not great. It's better than GitHub's, but that's a low bar. It's rare for me to find what I'm looking for through a Sphinx search.)

Huh, I have not had this experience with Sphinx, but who knows as to which software we're both searching! In any case, it's clear that we don't want to exclusively rely on search for discoverability (and i wouldn't propose it either!)

> The [original Awkward documentation](https://github.com/scikit-hep/awkward-0.x#readme) (and [Uproot](https://github.com/scikit-hep/uproot3#readme)) was one long page, to support that kind of discovery. Now there's too much to tell to actually get it on one page—zero click-throughs—but we can still get it pretty flat, like one or two click-throughs. At every click-through, a fraction of readers give up or follow the wrong path.

Yes, as discussed above, I think we need to find a compromise between these competing goals. I found the awkward 1.0 docs a challenge before we removed the private namespaces. But, now we have much more content thanks to merging the two sites.

I really want to make a case for "user interaction" differences between API docs which need to be easy to scan, and guides which need to be easy to read. I want to give the most help to the user that _doesn't_ know where to look; experienced users will remember where to navigate to (but we shouldn't make this harder than needs be).

I've seen some docs _somewhere_ with a page that creates user "case studies", e.g.
> I am a HEP researcher who wants to perform combinatorics on my data ...
>
> I am a data scientist working with ragged JSON data. I want to perform ...

We might want something like this (again, another PR) to help make the docs accessible from different user bases. 


> So I won't be a stickler on any visual arrangement, but it has to satisfy the constraint of minimizing the number of clicks from landing page to the information (flattening the tree). The current docs are failing at this, too: the fact that you have to click from awkward-array.org through "Python reference" to get to a page that has all of the functions listed (and then click on the function to learn about) is currently hiding information from users. A lot of user questions are answered by just saying that such-and-such a function exists—they're not finding it, and presumably, they're looking. I don't remember that happening as much with the single-page docs. (Then it was _other_ problems; Awkward 0.x had a lot of quirks.)

Going further with my "two kinds of user", I feel it's mainly the API reference that needs to be fast. In my experience, when I'm needing an API reference, I want it fairly rapidly. When I want to understand how something works, I am happier to take a bit more time. That's why I like having the API reference on the top navbar - it's a single click away. I don't like how collapsed NumPy and SciPy's docs are in this regard, but I recognise that they have even bigger challenges regarding API size.

For the user guide, it's currently three clicks from the landing page to a leaf. I think this is reasonable (minimum one click to go *somewhere* (the guide), another click to navigate to a topic. If we want to make the average case "2 clicks", then we could look to make the entry pages ("how-to-convert.md" etc) give a high-level overview (like a more detailed generated-toctree) of their section, e.g.
> ## I want to *create Awkward Arrays from existing data*
> Awkward Arrays can be created from a wide variety of Python data types, including NumPy arrays, Python objects, JSON, ROOT, Apache Arrow/Parquet, pandas, and generic (flat) buffers with metadata. 

Then these "parent" pages actually have merit

> Even if the top bar is not the four sections, "C++ Reference" should not be up there. Having prominent links to places most people won't want to go takes away from getting them to where they do want to go. We should be thinking of this as funnelling.

You did mention this already, I just haven't done it yet :turtle: It is certainly worth discussing what *should* go here. Maybe we link to it from the API reference. My only reservation here is that docs which stray too far from a common navigation structure get tricky to navigate (I'm looking at you, NumPy). Maybe it's fine for us, because really very few people should be needing these docs. 

I don't yet have a clear idea of what *should* go in the navbar besides API ref and user guide, so it will be nice to bikeshed that at some point. I did originally have a "high level" and "Low level" guide, but it felt like an arbitrary distinction, at least for now whilst it doesn't feel like we have a huge amount of content besides "create arrays with layout classes" (and maybe extending numba?). Maybe a design decision is thinking about what we want to write for "library authors" (which I'm going to call the "integration guide").

> All in all, I'm probably being too specific in my comments, since I'm probably missing good points from you by going round-robin between this and other things. I should leave you to have more autonomy in designing this. That's why in this comment, I tried to focus more on the high-level: why I'm wanting to flatten the tree/minimize the number of click-throughs. I hope it's more helpful than bewildering.

agoose77:
Some minor additions:

I don't want to weight clicks from the navbar too strongly as "clicks" - there's a fixed, small number of links here, and users will end up muscle memory / bookmarking this. E.g., I personally don't type the full URL to the docs, I either navigate to GitHub and click the link (because my muscle memory is quickest to type GH links, and I know where the docs button is), or using a Chrome Search Engine that queries the RTD i.e. this first navigation is automatic for familiar users, I suspect. 

For new users, I don't care so much about raw clicks as I do about whether they knew they were making the right clicks, i.e. if the structure was helpful rather than opaque (sorry NumPy!)

Given how much is yet undecided with respect to structure, I vote that we down-weight it as a priority for this first couple of PRs. Once the content PRs get underway, we might see a nicer structure emerge.

Actually, thinking about it - if you're concerned about the docs changing too much in the intervening period, we could make a "docs" branch that these PRs merge against. That way, the "latest" docs do not change until we're ready. This should be fairly easy to merge back into master as we're nearly entirely touching docs-only content.

jpivarski:
Two very good ideas above:

   1. Putting documentation improvements into a `docs` ranch so we can develop them in more than just a single PR.
   2. Grouping documentation pages with names beginning with "I want...", as in "I want to make an Awkward Array from existing data" instead of generic verbs like "Converting." As long as the docs are consistent about using the first person like this, I think that could be a perfect mental cue for someone who's thinking about that.

I had almost forgotten about my desire to divide things into high level and low level, and I agree that it could be hard for people to see when the first coming into it. I have, however, been gravitating toward that terminology when explaining things to people. It's really a analyst user versus downstream developer distinction, but the latter could just be a late section of the user's guide.

agoose77:
> It's really a analyst user versus downstream developer distinction, but the latter could just be a late section of the user's guide.

That's the direction I'm currently leaning into; the "Fundamentals of Awkward Array" should be followed by "Building libraries with Awkward Array"

I will get on the `docs` branch

agoose77:
I've added the release log to the navbar instead of the C++ reference. I think this should be fairly prominent, but feel free to suggest moving it.
image

I've moved the C++ link to a card in the API reference section:
image

API reference links now use the Sphinx `py:class` and `py:func`. To support this, the auto-generated (and manually curated) ReST files now have `py:currentmodule` and `py:class` directives.
image

I'm not going to go too much further with this work - we will eventually be dropping v1 docs anyway IIUC. This just makes things consistent between the pybind11 classes and those spidered by the `generate_docstrings` function. In the long run, we'll likely use autodoc instead of this tool.

jpivarski:
This is along the lines of what I had in mind for "burying" the entry point into doxygen. It's not on the main screen, a user has already clicked through to API reference, this is a last check "Did you mean C++?" Passing this block, it's all Python. It's like an early return in a function. I like it.

agoose77:
## New `reference` Endpoint
Now the API docs sit under `/reference`. This means that we can clean-up the docs directory, but _more_ importantly, it means that our generated and curated API docs are both clearly API documents according to the URL, e.g
- `ak.ArrayBuilder.html`
- `_auto/ak.all.html`
vs
- `reference/ak.ArrayBuilder.html`
- `reference/_auto/ak.all.html`

To guarantee that _existing_ URLs don't break, I've added a redirection list that maps to these locations.

## New Release Section
Release notes now sit under a release-notes section. 
image
This is preparing for using a tool like `github-activity` as part of our release process down the line (no rush on this). I intended to remove the changelog parsing from the build step, which should reduce build times. Ultimately I envisage us having a web-based release workflow like the JupyterLab project does. They have jupyter-releaser to automate this, it may/may not be useful for us.

agoose77:
## Redundant Headers
I've removed the section titles from the API reference docs. They take up a lot of space:
Before:
image

After:
image

There's a lot less visual noise here now. We might want to do some more formatting / theming work, but I think this is a good starting point. I've added a hidden anchor so that existing links don't break.

However, this removes the secondary sidebar links: 
image

It turns out that it is really hard to get Sphinx to put non-headers into the table of contents (and therefore, the right bar). I wrote a local extension to take more control over the table of contents, which gives this:
image

It's not beautiful code, but that's because the underlying Sphinx code is gnarly too. So, the question is
```mermaid
graph LR;
want_headers1{Want headers?}
want_headers2{Want headers?}
need_sidebar{Need sidebar links?}
use_extension[Use extension]
use_headers[Use headers]
no_headers[No headers]

need_sidebar --yes--> want_headers1
want_headers1--no-->use_extension
want_headers1 --yes-->use_headers

need_sidebar --no-->want_headers2
want_headers2 --no-->no_headers
want_headers2 --yes-->use_headers
```


@jpivarski do you have any strong feelings about this?

I feel like the answer might be "yes, we want sidebar links, because there are far more classes in v2". If so, it's then a question of "do we want headers". In general, I don't love writing fragile code to implement something non-essential, so it would be good to measure how essential it is to avoid using headers too, and how fragile the code is!

jpivarski:
Fixing the redundant headers looks good when unlinked, but I didn't like the way that they'd be highlighted when linking to them. I preferred linking to the header, rather than the autodoc, because linking to the autodoc would turn it yellow and scroll almost past it. It was counterproductive, but could be fixed with styles.

I see that it's still a page per function _or_ class, and that's good. You're asking about subsections (on the right) for methods of classes? They're not very necessary. If we had a lot of classes, they'd be important, but Awkward is mostly functions. Also, there's a lot of content under `ak.Array.__getitem__` that should be its own page (or pages). People are having trouble finding the slicing documentation that's in there.

So if you don't like what you had to do to get the method subsections to appear, it overrides my slight preference to have them—you can leave them out.

agoose77:
Thanks Jim. It seems like there are other people with an interest in being able to modify the toc tree generation, so this might end up being something Sphinx can provide. In the practical timeline, it would need to be a nicer extension. I can keep an eye on it there. In some ways, the case that is most problematic is classes with empty docstrings - then the docstring to header ratio is very small. But, those aren't "good docs" to begin with - empty docstrings are only very slightly useful. So, maybe having  a header is fine here. I'll play it out tomorrow :)

agoose77:
## Class Definitions

Class definitions are now inline for the curated class docs (`ak.layout`):
image

This will not look reasonable as-is for `ak.Array`, but as Jim notes above - this probably needs splitting into separate pages, or some other kind of rework. I tentatively suggest that the details in `Array.__getitem__` are more like usage-documentation than API documentation. There's no hard-and-fast rule here, but I'm thinking that these docs can go into the user-guide, with a ``:doc:`...` `` role linking from `Array.__getitem__`. I'd do this in PR 3.

jpivarski:
> This will not look reasonable as-is for `ak.Array`, but as Jim notes above - this probably needs splitting into separate pages, or some other kind of rework.

It would be advantageous to move much of the property and method docs for `ak.Array` (and other classes) out of docstrings into specialized pages, such as "Slicing", instead of the docstring for "__getitem__". All of these classes, including layouts, forms, and types, should have short docstrings—many of them are currently empty.

But beyond the information about slicing being more discoverable if it's in its own page, having shorter docstrings would reduce the output of error messages, when the error message include the source code of the function that failed (mostly in pytest), because that docstring is miles long.

agoose77:
> It would be advantageous to move much of the property and method docs for ak.Array (and other classes) out of docstrings into specialized pages

Agreed. I'm thinking of the user-guide for this, so that we can keep a very clear distinction between the more black-and-white docstrings that describe usage, and the user-level information that describes semantics & motivation

agoose77:
I'm not going to touch much more in this PR. Structurally I want to address how the auto-generated docs are handled, but that will probably need to be its own PR too. Moreover, it doesn't need to happen *right now*.

The objectives of this PR were mainly structural. The next PR in this series will probably focus more on content, including reworking signatures to be less verbose.

Preview of this PR: https://awkward-array.readthedocs.io/en/docs-phase-2/

ianna:
@agoose77 - @ManasviGoyal is going to write a new LayoutBuilder documentation: both the reference guide and the user guide. Could you please, coordinate with each other? Thanks

jpivarski:
> @agoose77 - @ManasviGoyal is going to write a new LayoutBuilder documentation: both the reference guide and the user guide. Could you please, coordinate with each other? Thanks

A good place for it would be in the doxygen—in other words, in comments in the header files themselves (with appropriate formatting). That way, if anybody includes the header files in their own project, they get the documentation along with it.

But also having dedicated documents in Sphinx would also be good.

As for content, the most important thing would be "how to use it." Primarily, we should be thinking of the ctapipe project. We should also have a coordinated meeting with them.

ManasviGoyal:
> > @agoose77 - @ManasviGoyal is going to write a new LayoutBuilder documentation: both the reference guide and the user guide. Could you please, coordinate with each other? Thanks
> 
> A good place for it would be in the doxygen—in other words, in comments in the header files themselves (with appropriate formatting). That way, if anybody includes the header files in their own project, they get the documentation along with it.
> 
> But also having dedicated documents in Sphinx would also be good.
> 
> As for content, the most important thing would be "how to use it." Primarily, we should be thinking of the ctapipe project. We should also have a coordinated meeting with them.

@jpivarski Yes. A meeting with the astrophysicists will be really helpful to get their feedback and inputs.

For the dedicated document, I was thinking that we can have a tutorial kind of thing in which we can add some examples from the tests with steps-by-step instructions for the users. We can focus on just Numpy, Listoffset and Record Layout Builders as the ctapipe project will use these mainly.

I was planning to add it in `docs-src` with other `how-to` documents. Would that be the right place to add it? Also what should be the file type, Markdown or something else?

jpivarski:
Any new meeting would be after this upcoming week, but I'll get that planned as soon as I can.

The how-to docs would be with the other tutorial docs, which used to be in docs-src (as Jupytext), but they're being heavily reorganized and restructured by @agoose77. The best thing to do would be to write the how-to instructions independently of where they'll eventually be going, and @agoose77 will find a place for them.

Markdown is a good neutral format. We have doxygen configured to interpret markdown. To preview the formatting, you could write it directly in GitHub. You could even do drafts in comments on a pull request or on a new page of the wiki.

agoose77:
Thanks @jpivarski . I replied over on #1574, but I'm not hugely worried as to which approach @ManasviGoyal takes - as long as it's Markdown, I can integrate it :)

RE this PR, I'd be happy for you to review it at some point. I'm away at a wedding from 4th August - 9th, so if you are super busy at the moment then don't worry - we can pick up when I get back!

There are some outstanding tasks that were raised here, or elsewhere, that I've [moved to "todo"](https://github.com/scikit-hep/awkward/projects/7#column-19008268). I think the most pressing issue is probably the documentation contents now - filling out the existing user guide. I believe that the current set of topics are good places to start - constrained, focused, and in some cases requested by users. In the next PR, then, I'll prioritise working on _content_ rather than doc layouts.

jpivarski:
I'll review this on Monday. Do any other PRs need to go together with this one?

agoose77:
Oh yes, I don't expect people to do any work on weekends; I was just sneaking in some commits. I hadn't thought about the optics of pinging you for a review, I apologise for that.

No, this builds upon the `docs` branch, which has already had the previous PR merged into it. I took your :+1: as approval [there](https://github.com/scikit-hep/awkward/pull/1552).

jpivarski (approved):
This is definitely a good phase 2. Since there are so many changes, I'm reviewing it by looking at the output on readthedocs.

The front page divides the world into "Beginner's Guide", "User Guide", "API Reference", and "Contributing Guide", labels that come from NumPy/SciPy/Pandas. Within the Beginner's and User's Guides, however, the left-bar headings are "Getting Started" and "Fundamentals of Awkward Array". Having them be different is a little disorienting; makes it hard to tell where we are and how to get back here if needed. Could these be "Beginner's Guide" and "User Guide", for consistency with the first four panels? I recognize that the "Beginner's Guide" is a page, distinct from the "What is an Awkward Array?" page, whereas the "User's Guide" is a collection, but this is some complicated navigation—it makes it feel like a maze.

I like how the "Release Notes" are prominent in the top-bar. However, the notes for each version are on separate pages. They were not written with the assumption that the ones before and after it would not be on the same screen. There are major, very important version milestones that have almost no text, with all of the incremental changes on the "rc" releases before it. If the release notes were a single continuous page (as I've seen elsewhere), then it would be clear that the information one is looking for is under one of the other version headings, but not if it's click-through. When trying to find out when a particular feature was added, I wouldn't know how many versions to click through. This needs to be one page.

But the PR represents an immense amount of work, is a huge improvement, and should be merged. I'll let you press the button, in case you have one or two more things planned.

agoose77:
@jpivarski I started writing an answer outlining options here, but I looked again at Pandas and I'm more convinced in their approach here.

> The front page divides the world into "Beginner's Guide", "User Guide", "API Reference", and "Contributing Guide", labels that come from NumPy/SciPy/Pandas. Within the Beginner's and User's Guides, however, the left-bar headings are "Getting Started" and "Fundamentals of Awkward Array". Having them be different is a little disorienting; makes it hard to tell where we are and how to get back here if needed. Could these be "Beginner's Guide" and "User Guide", for consistency with the first four panels? 
I recognize that the "Beginner's Guide" is a page, distinct from the "What is an Awkward Array?" page, whereas the "User's Guide" is a collection, but this is some complicated navigation—it makes it feel like a maze.

Yes, this is based on NumPy's approach. For context:
- NumPy has "Getting Started" as a sub-section of "User Guide"
- SciPy has "Getting Started" as its own top-level section with links to tutorials and install instructions.
- Pandas has "Getting Started" as its own top-level section, with some quick-links that answer rhetorical questions.

I think our design, based upon NumPy, is more confusing than it needs to be. I think ultimately I like the Pandas "User Guide" design which doesn't have the categorical headers. Something like:
image

Then we can move the "What is an Awkward Array?" to a separate "Getting Started" guide. This guide would need to be added to the top-level navbar, as it is a top-level section. NB - the index card would still be linking to a sub-page of a top-level document (Getting Started > What is an Awkward Array?) But I don't think that will be nearly as problematic, as it looks like a link button rather than a card.

So, our index would become
image

The main content of the getting started guide could be based upon the Pandas implementation, which I think is really good:
https://pandas.pydata.org/pandas-docs/stable/getting_started/index.html#intro-to-pandas

I especially like the fact that their "Intro to Pandas" subsection is basically a series of questions. I'm not sure whether the fact that they have so many tutorials is a good fit for us, but I like the idea that we could have common prospective user questions here with links to the relevant docs & a short inline answer:
image

@jpivarski one outstanding issue is where to put the advanced user (library author) docs. The constraints are:
- Any top-level pages must be added to the navbar
  This can get unwieldy if we have too many links, e.g. 
image
- Standalone pages with little content should probably just be subsumed into another section

I'm actually violating one of these rules at the moment (the contributor docs), which I do want to resolve.

I am wondering how much material corresponds to library author docs? Right now, I feel like it's just the "how to use layouts" section. If so, would you be happy to make "Integrating a library with Awkward Array" (or something like that) a subtopic in the user-guide? e.g.
image

In summary:
- Move "what is an awkward array" to a new "getting started" section
- Add install instructions to this section
- Add Q&A cards to this section
- Rename "Beginners guide to Awkward Array" to "10 Minutes to Awkward Array"
- Remove user-guide headings
- Add library integration to an "Integrating another Library"

agoose77:
I based the release notes change upon https://docs.scipy.org/doc/scipy/release.1.9.0.html

This is for two orthogonal goals:
- Move release-note generation from docs building to release time (e.g. via [jupyter-releaser](https://github.com/jupyter-server/jupyter_releaser)). This will speed up docs building & be more standardised via a changelog tool (e.g. [github-activity](https://github.com/executablebooks/github-activity))
- Make release notes easier to navigate by loading less HTML!

If you're on-board with using a tool for release note generation, then what we're discussing here is a temporary measure - rewriting release notes from our existing `make_changelog` tool whilst we're not using a standalone tool. Or, if you want to keep the published release notes verbatim as they are now, then we could use this new workflow only for new releases.

That said, whilst this is placeholder for that future change, I think you're right that it's currently too granular; whilst I strongly argue in favour of making these pages, the RCs need to be aggregated.

Here's what an aggregated log looks like:
image

How do you feel about the [SciPy](https://docs.scipy.org/doc/scipy/release.html) or [Pandas](https://pandas.pydata.org/pandas-docs/stable/whatsnew/v1.3.1.html) approach? The non-RC releases are given release note pages. I feel that this is sensible - the only RCs that users should really care about are the bleeding edge RCs. We can have a singular "latest-RC" release that corresponds to this, perhaps? Or, do we even need to list the RCs on the docs? Aren't they already visible from the GitHub side of things? Saying yes would make this a lot easier, but sometimes we have to take the hard road ;)

Prior art:
- [NumPy](https://numpy.org/doc/stable/release.html) has both per-release pages and a summary. The summary is just the toctree summary that nests up to a single header. This would be useful for searching headings within releases, but less so for  small fixes. I found the sphinx search was able to locate strings of text within these release notes.
- [SciPy](https://docs.scipy.org/doc/scipy/release.html) has per-release pages, with a summary list that contains just the links to the release notes.
- [xarray](https://docs.xarray.dev/en/stable/whats-new.html#) has a single release page. Note that they use the JB theme rather than pydata.
- [Pandas](https://pandas.pydata.org/pandas-docs/stable/whatsnew/index.html) has both per-release pages and a summary. The summary is grouped by major release, and doesn't contain any grep-able text.

jpivarski:
Your deep study of the Pandas docs is convincing—go ahead and make ours similar, as you suggest. In particular, I like the pages being questions, too.

For granularity of release notes, yes, it could be one large per non-rc release, as long as that page includes all of the rc notes leading up to it. The rc notes contain most of the important updates found in the non-rc, since that one usually says, "this has no changes since the last (rc) release."

The pattern was less regular in early history, but maybe all of the pre-1.0 releases can be grouped onto one page, while subsequent releases follow a more regular pattern.

agoose77:
Thanks for that quick follow up Jim.

I'll address the points in the comment concerning our Getting Started page.

If you're happy with grouping the releases to the release version (i.e. RCs are bundled together with the release), then I'll get on that. I think it's a good idea to merge the pre 1.0 releases too, I'll make note of that!

agoose77:
@jpivarski would you be able to author something like 
 in the style of our existing diagrams? I think you mentioned that you used inkscape? If it will take me the same amount of time to do it, I'm happy to do the work myself - just thought you might already have a template somewhere :)

agoose77:
I've reverted the release-note changes in favour of a future PR to integrate the changelog generation. That requires some further discussion, so I've kicked it down the road.

I've started adding some simple question-style feature discussions in the getting-started section: 
image

These are mainly rough drafts that I'll improve in subsequent commits.

agoose77:
There's now a single-column layout index page for small screens too:
image

jpivarski:
> I think you mentioned that you used inkscape? If it will take me the same amount of time to do it, I'm happy to do the work myself - just thought you might already have a template somewhere :)

I don't have templates, but there is a collection of diagrams in [docs-img](https://github.com/scikit-hep/awkward/tree/main/docs-img) that you can copy and modify as a starting point, which would get the same look-and-feel.

This ragged array example is most like example-reduction.svg:

[](https://github.com/scikit-hep/awkward/blob/main/docs-img/diagrams/example-reduction.svg)

which originally came from the PDF of a talk created by minted in LaTeX. It didn't translate well (SVG doesn't save fonts, and if you pick one that isn't on all/most systems, it can look bad), so I redid it for a tutorial in example-reduction-sum.svg:

[](https://github.com/jpivarski-talks/2022-07-11-scipy-loopy-tutorial/blob/main/img/example-reduction-sum.svg)

I replaced the text in a more standard font. (I need to get my act together about knowing which fonts are standard and using only them; I usually only want "something sans-serif," "something monospace," or rarely, "something serif." I should make some good choices and stick to them, otherwise old diagrams are hard to reuse.)

In this context, you don't have anything to say about the sum/product at `axis=0` or `axis=1`, so you can just remove those elements in Inkscape. I don't know if you want to remove the `None`, too, because you don't want to introduce missing data yet.

Inkscape is great: [download](https://inkscape.org/); if you just focus on learning one or two tools (just the arrow, text, raise/lower selection, fill/stroke style, and alignment to start with), you'll be as productive as me at that. Easy things are easy if you ignore all the buttons you don't know (there are a lot of buttons I don't know) and all things are possible with some web-searching.

Please do try that out, use "Inkscape SVG" as the primary file format (it's a _subtype_ of "Plain SVG", thanks to SVG's extensibility with xlinks), and don't worry about the overall scale of the image. Select everything and use "File → Document properties → Custom size → Resize page to content" with top-left-right-bottom padding of 5, 5, 5, 5 to make the page fit whatever you draw.

Without being dogmatic about it, I generally follow Edward Tufte's advice on making plots and diagrams use just enough elements to get the point across. In this diagram, I made it look like horizontally-formatted Python because that's how people usually see the data in context. Line colors should generally be bold and distinct like Matplotlib's color sequence (medium blue and dark orange are the most distinct), and fill colors should generally be light, like pastel. There's a palette of good-enough choices on the bottom of the Inkscape window; no need to tweak anything with the color wheel. Your example numbers are very detailed (3 digits of precision); I think everyone would get the point if they were 1.1, 2.2, 3.3, etc. without getting caught up in "why those numbers?" The prime numbers I used to demonstrate `ak.prod` and the powers of 2 I used to demonstrate `ak.sum` are to ensure that the results are unique, so that people can't misinterpret the order of the reducer—you don't need that and can simplify the numbers for this purpose.

agoose77:
Thanks Jim. I hadn't realised that Inkscape actually have a derived version of SVG, curious. Thankfully I've used Inkscape before, so I'm hopeful it will be a productive process in future.

For context - the image was entirely placeholder - I much prefer our examples. In fact, I've just taken the reduction diagram and split it into the reduction part and the ragged part:
image

agoose77:
I've added a Community Tutorials section that maintains an aggregated list of community tutorials. Most of these are yours, Jim, so we should be able to ensure that they survive long term:
image


I'd be interested on what your feelings are for this first-pass of presenting users with the necessary information RE what the tutorial details and how it is presented.

agoose77:
Pandas' user guide has these short panels 
image

that link to a long-form tutorial:
https://pandas.pydata.org/pandas-docs/stable/getting_started/intro_tutorials/01_table_oriented.html

I'm not convinced whether we need/want to duplicate this information in a tutorial *and* the user-guide. The user-guide is probably verbose enough, and I'm not a fan of having the same content repeated slightly differently in separate parts of the site.

However, if you think that Pandas does this well and want to repeat it ourselves, we can think about it.

jpivarski:
> Most of these are yours, Jim, so we should be able to ensure that they survive long term

Don't worry too much about preserving my content. That's an extra constraint on making this into a good overall documentation system (e.g. the conflict of "duplicate this information in a tutorial and the user-guide"), and the problem will be easier to solve with fewer constraints.

As the original author of the library, my perspective is often _not_ the right one for a new user encountering it. That's why I want to bring in people who have trained new users, to at least review it. There are things we can probably say more simply. If that means tearing up old documentation, please do it: the goal should be for it to be a sensible whole in the future.

agoose77:
> > Most of these are yours, Jim, so we should be able to ensure that they survive long term
> 
> Don't worry too much about preserving my content. That's an extra constraint on making this into a good overall documentation system (e.g. the conflict of "duplicate this information in a tutorial and the user-guide"), and the problem will be easier to solve with fewer constraints.
> 
> As the original author of the library, my perspective is often _not_ the right one for a new user encountering it. That's why I want to bring in people who have trained new users, to at least review it. There are things we can probably say more simply. If that means tearing up old documentation, please do it: the goal should be for it to be a sensible whole in the future.

My idea for the _community_ tutorials is that they will better showcase an end-to-end usage of Awkward rather than the details of particular features. The user-guide, meanwhile, should focus on helping users find features and explaining their use in more detail than the API reference (with executable examples, etc).

For now, I think I'll leave your tutorials in this short list. We can accept PRs to build out / curate this list as time goes on. For now, I think it better to have something to link to vs leaving users guessing?

It seems like you're also leaning towards not having mini-tutorials in addition to the user guide. As such, I won't do that!

agoose77:
Given that you've approved this (albeit slightly earlier), I think it would be best if I merge this (into docs) and continue iterating in the next PR cycle.

Issue #1551, opened 2022-07-18
Layout Record should have an 'axis_wrap_if_negative' method


jpivarski:
### Version of Awkward Array

HEAD

### Description and code to reproduce

Otherwise, any method that takes an `axis` fails like this:

```
AttributeError: 'Record' object has no attribute 'axis_wrap_if_negative'
```

This affects both v1 and v2. It's a bug, so it should be fixed in both, but v2 is higher priority if there's only time to fix one of them. The fix in v2 can be in Python; the fix in v1 has to be C++ code.

Fields of a record don't count as an `axis`, so the implementation would be to just pass it to `self._array.axis_wrap_if_negative`.

However, if the `axis` that this function returns is equal to `0`, that should be an error. The user is trying to compute something on a top-level array, but the top-level is a scalar, not an array. We don't want to pass that through `self._array.axis_wrap_if_negative` and have, for instance, `ak.to_regular(record, axis=0)` return an _array_, especially if that ignores the Record's `_at` attribute and returns an output containing what had been hidden data.

agoose77:
Huh, I thought we had an issue tracking this, but apparently not!

jpivarski:
I don't remember it. I just did a search and didn't find it. If it rings a bell, that's a good sign for consistency, at least.

PR #1565, opened 2022-07-25
Fix: Missing `axis_wrap_if_negative`  Record method in both v1 and v2

ioanaif:
#1551


jpivarski (approved):
Looks great! #1559 does the other half of this: refusing to allow records in the error cases that came up.

PR #1567, opened 2022-07-26
Fix: tolist() bug from Uproot (longlong type for Index) 

### LINKED ISSUE #1530: `tolist` bug revealed by Uproot 5 testing
### Version of Awkward Array

HEAD

In scikit-hep/uproot5#620, a test named `test_gohep_output_file` is failing in `tolist`:

```
    with uproot.open(skhep_testdata.data_path("uproot-issue413.root"))[
        "mytree"
    ] as tree:
        assert tree["Str"].array().tolist()

                                   ^^^^^^^^
```

It looks like an Awkward bug, to be fixed here. (Uproot has a work-around, so we won't see it again unless we actively look for it. Hence this bug report.)

ioanaif:
#1530


src/awkward/_v2/index.py
```diff
@@ -27,6 +27,9 @@ def __init__(self, data, metadata=None, nplike=None):
         if len(self._data.shape) != 1:
             raise ak._v2._util.error(TypeError("Index data must be one-dimensional"))
 
+        if self._data.dtype.type == np.longlong:
+            self._data = self._data.astype(np.int64)
```
jpivarski: Since longlong is just another word for 64-bit (or 32-bit?) integers, we don't need `astype`, which converts. It should be enough to use `view`, which is zero-copy.

In fact, we'd like to know if there are any systems for which longlong is not 64-bit, so we can put in an assertion with a question and hope someone answers it. Oh, [longlong is _at least_ 64-bits](https://stackoverflow.com/questions/5221615/c-long-long-always-64-bit), but my NumPy, at least, doesn't have an `np.int128` type.

Finally, `np.longlong` is a type, so we want to check for it with `issubclass`.

This should apply all of those:

```suggestion
        if issubclass(self._data.dtype.type, np.longlong):
            assert np.dtype(np.longlong).itemsize == 8, "longlong is always 64-bit, right?"
            self._data = self._data.view(np.int64)
```

jpivarski (approved):
Just click on the substitution, and if it passes tests, it's done (you can merge it).

jpivarski:
It looks like you're still debugging this. Debugging through CI is hard, and I've never managed to get a Windows virtual machine that I could test on.

Often, though, Windows fails because it has different C `int`, `long`, `long long` assignments to `int32` and `int64`. That is very likely what's going on here. It is possible to simulate this kind of thing in a 32-bit Linux Docker container. What I used to do (I don't have one set up on my current computer to use as an example; sorry!) is make an instance of

https://hub.docker.com/r/i386/debian/

and then use the latest 32-bit Linux Miniconda, which is one of these:

https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86.sh
https://repo.anaconda.com/miniconda/Miniconda3-4.5.4-Linux-x86.sh

Install Miniconda with the conda-forge channel inside the Docker container. Because this is old stuff, you have to explicitly set the channel priority to strict or ensure that conda-forge is the only channel, no "defaults".

https://conda-forge.org/docs/user/introduction.html#how-can-i-install-packages-from-conda-forge

Don't try to install everything (a lot of it won't work), just the minimum to get the test running, which might be just `numpy` and `pytest`. Maybe not even `pytest` (just run what you need to on a `python -c` line).

Along the way, you'll see some warnings saying that you're trying to do 32-bit stuff on a 64-bit computer. Docker containers are a little leaky: it can see that you have a 64-bit OS underneath, but it worked anyway.

When you start the Docker container, you can to symlink it to your git directory so that you can use a normal editor outside the 32-bit world, and any changes made to source files aren't deleted when the container shuts down. It goes without saying that you'll need to do a clean recompilation of the C++ part of Awkward. You might consider doing this on a different git clone than the main one you normally use.

It's a bit of work to set it up, but it makes debugging a lot easier than "submit a change, wait 15 minutes, see how Windows CI failed...". The 32-bit Linux test is likely to work because you're dealing with a "how is long long defined?" issue.

PR #1568, opened 2022-07-26
Fix: Lengths of empty regular slices

### LINKED ISSUE #1557: Lengths of empty regular slices (in v2)
### Version of Awkward Array

HEAD

At least for v2 and variable-length arrays, we get the right zero-slicing edge case:

```python
>>> import awkward._v2 as ak
>>> ak.Array([[1, 2, 3], [4, 5]])[:, []]
<Array [[], []] type='2 * var * int64'>
```

but v2 regular arrays have a zero-length error:

```python
>>> ak.to_regular(ak.Array([[1, 2, 3], [4, 5, 6]]), axis=1)[:, []]
<Array [] type='0 * 3 * int64'>
```

What we want to have happen is for this slice to build a RegularArray with an explicit `zeros_length` argument, which is the only way to make a RegularArray with non-zero length yet contain lists of zero length:

```python
>>> ak.Array(
...     ak.contents.RegularArray(
...         ak.contents.NumpyArray(np.arange(1, 7)),
...         size=0,
...         zeros_length=2,
...     )
... )
<Array [[], []] type='2 * 0 * int64'>
```

Pointed out by @grst.

v1 arrays are also incorrect, but this is an edge-case bug that only really needs to get fixed in v2. (v1 has only 4 more months left...)

ioanaif:
#1557


tests/v2/test_0024-use-regular-array.py
```diff
@@ -52,9 +52,14 @@ def test_empty_array_slice():
...
     assert to_list(listarray[2, 1, []]) == []
     assert listarray.typetracer[2, 1, []].form == listarray[2, 1, []].form
+<<<<<<< HEAD
+    assert to_list(listarray[2, [1], []]) == []
+    assert listarray.typetracer[2, [1], []].form == listarray[2, [1], []].form
+=======
     assert to_list(listarray[2, [1], []]) == [[]]
     # FIXME
     # assert listarray.typetracer[2, [1], []].form == listarray[2, [1], []].form
+>>>>>>> 0c477f49cdcb3ed31fb70655391633b007e739fb
```
agoose77: @ioanaif it looks like a sneaky merge conflict made its way through!

What did you want to do about this test? Is it broken for another reason, and therefore you're commenting it out here? Maybe we should skip it instead of comment it out, so that we have a high-level indication it's not working?

```suggestion
assert to_list(listarray[2, [1], []]) == [[]]
# FIXME
# assert listarray.typetracer[2, [1], []].form == listarray[2, [1], []].form
```

ioanaif: @agoose77 It got in there by mistake while the PR was still a draft; I reverted all the changes to that test before marking the PR as ready

agoose77: Oh, strange. Maybe my client didn't refresh! Ignore me :smile:

jpivarski (commented):
This doesn't look like a general solution: if there are two sliced dimensions, should there be two RegularArrays? What if the sliced dimension is not first? Why is `start is None or stop is None` the condition for determining whether there should be a RegularArray? For positive `step`, `start is None` is equivalent to `start == 0` and `stop is None` is equivalent to `stop == len(self)`.

I totally believe that a RegularArray wrapper is needed here, but not under these conditions. Maybe always? Maybe if there's at least one slice somewhere in the tuple? Maybe the number of RegularArrays depends on the number of slices, and their sizes depend on which dimensions they're at?

The way to find out is to see what NumPy does: fortunately this test doesn't depend on dimensions being irregular. You can make a NumPy array with a lot of dimensions and try slices on it with various combinations of `:` and `[0, 1, 2]`, and `[]` to see what happens to the output `shape`.

ioanaif:
> What if the sliced dimension is not first?  

Good point, I assumed in an empty slice that it would go first. 

> Why is start is None or stop is None the condition for determining whether there should be a RegularArray?

This was the pattern I found from the tests, the `zeros_length` needs to be adjusted only when a slice with this properties was passed  

 > The way to find out is to see what NumPy does: fortunately this test doesn't depend on dimensions being irregular. You can make a NumPy array with a lot of dimensions and try slices on it with various combinations of `:` and `[0, 1, 2]`, and `[]` to see what happens to the output `shape`.

I noticed this behaviour is only consistent when slice.start/slice.stop is None, here are some of the NumPy tests I did: 

```
>>> d = np.arange(3 * 3 * 2).reshape(3,3,2)
>>> e = ak._v2.contents.NumpyArray(d)
>>> c1 = np.array([], np.int64)
>>> ak._v2.to_list(d[[2],[1],c1]) == ak._v2.to_list(e[[2],[1],c1]) == []
True
>>> d = np.arange(1 * 3 * 3 * 2).reshape(1,3,3,2)
>>> e = ak._v2.contents.NumpyArray(d)
>>> ak._v2.to_list(d[c1,c1]) == ak._v2.to_list(e[c1,c1]) == []
True
>>> d = np.arange(2 * 3).reshape(2, 3)
>>> e = ak._v2.contents.NumpyArray(d)
>>> ak._v2.to_list(d[:,[]]) == ak._v2.to_list(b[:,[]]) == [[], []]
True 
>>> ak._v2.to_list(d[1:,[]]) == ak._v2.to_list(e[1:,[]]) == [[]]
True
>>> ak._v2.to_list(d[:-2,[]]) == ak._v2.to_list(e[:-2,[]]) == []
True
>>> ak._v2.to_list(d[:-1,[]]) == ak._v2.to_list(e[:-1,[]]) == [[]]
True
```

src/awkward/_v2/contents/content.py
```diff
@@ -519,10 +519,19 @@ def _getitem(self, where):
...
+                if isinstance(items[0], slice):
+                    if items[0].start is None or items[0].stop is None:
+                        return ak._v2.contents.RegularArray(
+                            out,
+                            0,
+                            out.size,
+                            None,
+                            None,
+                            self._nplike,
+                        )
```
jpivarski: The first issue with generality (described above) can be fixed like this:

```suggestion
                if isinstance(items[0], slice):
                    return ak._v2.contents.RegularArray(
                        out,
                        0,
                        out.size,
                        None,
                        None,
                        self._nplike,
                    )
```

jpivarski (changes requested):
Okay, so this isn't a general solution to the problem.

Although this test case works (in your unit tests):

```python
>>> ak._v2.to_regular(ak._v2.Array([[1, 2, 3], [4, 5, 6]]), axis=1)[:, []]
<Array [[], []] type='2 * 0 * 2 * 3 * int64'>
```

There's nothing special about the slice's `start` and `stop` being `None`. For instance, the first dimension has length 2, so `:` is completely equivalent to `0:2`, and yet:

```python
>>> ak._v2.to_regular(ak._v2.Array([[1, 2, 3], [4, 5, 6]]), axis=1)[0:2, []]
<Array [] type='0 * 3 * int64'>
```

That's easily fixed by removing the if-statement as suggested below.

But here's a deeper issue (pun intended): I can make the example 3-dimensional, rather than 2, by putting the whole thing into a length-1 list, replacing `axis=1` with `axis=2`, and putting another `:` in the slice:

```python
>>> ak._v2.to_regular(ak._v2.Array([[[1, 2, 3], [4, 5, 6]]]), axis=2)[:, :, []]
<Array [[]] type='1 * var * 3 * int64'>
```

But it's not returning

```python
>>> np.array([[[1, 2, 3], [4, 5, 6]]])[:, :, []].tolist()
[[[], []]]
```

as it should.

It has something to do with the nested RegularArrays. If I make the Awkward Array out of RegularArrays, it doesn't work, but if I make it out of a (multidimensional) NumpyArray, it does (although that may just be because we pass it off to NumPy, which is correct by definition).

```python
>>> ak._v2.from_numpy(np.array([[[1, 2, 3], [4, 5, 6]]]), regulararray=True)[:, :, []]
<Array [[]] type='1 * 0 * 1 * 2 * 3 * int64'>
>>> ak._v2.from_numpy(np.array([[[1, 2, 3], [4, 5, 6]]]), regulararray=False)[:, :, []]
<Array [[[], []]] type='1 * 2 * 0 * int64'>
```

I thought maybe I would figure it out, but no—I poked around with it, but I don't understand why the original solution worked for the test example. Do you see what's going on here?

src/awkward/_v2/contents/content.py
```diff
@@ -520,6 +520,13 @@ def _getitem(self, where):
                 self._nplike,
             )
 
+            if (
+                any(isinstance(x, list) and len(x) == 0 for x in items)
+                and any(isinstance(x, slice) for x in items)
+                and next.maybe_toNumpyArray() is not None
+            ):
+                next = next.maybe_toNumpyArray()
+
```
jpivarski:
```suggestion
            if (
                any(isinstance(x, list) and len(x) == 0 for x in items)
                and any(isinstance(x, slice) for x in items)
            ):
                attempt = next.maybe_toNumpyArray()
                if attempt is not None:
                    next = attempt

```

jpivarski: This just avoids doing the `maybe_toNumpyArray` twice. It is a deep recursive algorithm.

ioanaif: True. Thank you!

jpivarski (approved):

jpivarski (approved):

PR #1573, opened 2022-07-29
Feat: raise `AttributeError` for public Array attributes

### LINKED ISSUE #1511: Raise warning on setattr for assigning arrays to records

Take the following rather surprising but currently correct behavior:
```
>>> import awkward as ak
>>> a = ak.zip({"x": ak.Array([[1], [2, 3], [4, 5, 6]]), "y": ak.Array([[10], [20, 30], [40, 50, 60]])})
>>> a.x = a.x *3
>>> a.x
<Array [[3], [6, 9], [12, 15, 18]] type='3 * var * int64'>
>>> a["x"]
<Array [[1], [2, 3], [4, 5, 6]] type='3 * var * int64'>
```
It is surprising to users that `a.x` and `a["x"]` are different, but expected because the array set by `setattr` ends up masking the underlying array already in the layout of `a`. The recommended way to assign arrays to existing records is to use `setitem` as shown:
```
>>> a["y"] = a.y *3
>>> a.y
<Array [[30], [60, 90], [120, 150, 180]] type='3 * var * int64'>
>>> a["y"]
<Array [[30], [60, 90], [120, 150, 180]] type='3 * var * int64'>
```
then the `getattr` and `getitem` methods behave the same.

I don't see a lot of legitimate use cases for using `setattr` with array arguments. In these cases I think we should raise a warning or error.

agoose77:
Fixes #1511 

As discussed in the issue, the new behaviour is:
- existing public _class_ attributes can be assigned to (though the attribute itself may not be writable, e.g. with a read-only data descriptor)
- private attributes can be assigned to, and will never attempt to set a field
- fields can never be set

We might have some hidden issues if users employ underscore-prefixed fields. Right now, setting these attributes will write to the `ak.Array` object. However, this shadowing is already accepted when we read from an attribute - class attributes take precedence over fields, just as they do in this PR for writing. Are we happy with this?

### DIFF
--- a/src/awkward/_v2/highlevel.py
+++ b/src/awkward/_v2/highlevel.py
@@ -1044,7 +1044,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
     def __delitem__(self, where):
         """
         Args:
-            where (str): Field name to add to records in the array.
+            where (str): Field name to remove from records in the array.
 
         Removes a field from records in the array.
         """
@@ -1062,6 +1062,9 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
     def __getattr__(self, where):
         """
+        Args:
+            where (str): Attribute name to lookup
+
         Whenever possible, fields can be accessed as attributes.
 
         For example, the fields of an `array` like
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is good; I'd like to merge it, but I'm going to update the branch to main (after #1576 is done) before doing so.

agoose77:
I might change `... in dir(type(self))` to `hasattr(self.__class__, ...)` for simplicity, but I'll do it at a later date.

jpivarski:
Good point. Or maybe `hasattr(type(self))` because if there's a public function that reveals a dunder-attribute, I think it should be preferred.

PR #1587, opened 2022-08-13
Docs: add replite support

agoose77:
[replite](https://github.com/jtpio/replite) is a standalone code console, powered by jupyterlite, that runs in the webbrowser. 

This PR builds Awkward as an out-of-tree pyodide module and exposes it to end-users of the documentation within a replite console. This should permit users to test-drive Awkward without needing to pip install it, and could (in future) be extended to permit interactive examples (via `thebe` integration) in our user guide.

Namely, this PR:
- Builds Awkward Array as an WASM wheel via pyodide (emscripten). 
- Adds a "Try Awkward Array in the browser" example page with an embedded replite code-console (via `jupyterlite-sphinx`)

image

We can drop the `await piplite.install(...)` if we build in-tree, but that's a lot slower and heavier.

There are a couple of pain points that I'd like to resolve first before merging:
- Support for latest pyodide (https://github.com/jupyterlite/jupyterlite/pull/756)
   - Pyodide 0.21.0 makes it a lot easier to compile a shared library like Awkward's C++ component, but jupyterlite is yet to support it (see PR)
   - Currently we therefore build jupyterlite from GIT, but once there's an RC/pre-release we can easily just `pip install` on both our CI and RTD.
- Move WASM building into separate job, and use ccache.


agoose77:
Closed for branch renaming. Sorry for the noise all.

PR #1596, opened 2022-08-15
fix:  in RegularArray `getitem_next` should have an option when `maybe_toNumpy` is not None

### LINKED ISSUE #1595: Slicing RegularArrays inside of something else, such as IndexedOptionArray (v2)
### Version of Awkward Array

HEAD

After #1568, this works:

```python
>>> array = ak._v2.from_numpy(np.array([[[1, 2, 3], [4, 5, 6]]]), regulararray=True)
>>> array[:, :, []]
<Array [[[], []]] type='1 * 2 * 0 * int64'>
```

However, if the RegularArrays are inside of something else, such as an IndexedOptionArray (which doesn't change the dimensionality and it's not as complicated as a RecordArray or UnionArray), then it doesn't work.

```python
>>> array = ak._v2.Array(
...     ak._v2.contents.IndexedOptionArray(
...         ak._v2.index.Index64([0, -1]),
...         ak._v2.from_numpy(np.array([[[1, 2, 3], [4, 5, 6]]]), regulararray=True, highlevel=False),
...     )
... )
>>> array[:, :, []]
```
```
  File "/home/jpivarski/irishep/awkward/awkward/_v2/highlevel.py", line 1175, in __repr__
    return self._repr(80)
  File "/home/jpivarski/irishep/awkward/awkward/_v2/highlevel.py", line 1199, in _repr
    valuestr = " " + awkward._v2._prettyprint.valuestr(self, 1, strwidth)
  File "/home/jpivarski/irishep/awkward/awkward/_v2/_prettyprint.py", line 211, in valuestr
    _, strs = valuestr_horiz(data, limit_cols)
  File "/home/jpivarski/irishep/awkward/awkward/_v2/_prettyprint.py", line 87, in valuestr_horiz
    current = get_at(data, index)
  File "/home/jpivarski/irishep/awkward/awkward/_v2/_prettyprint.py", line 42, in get_at
    out = data._layout._getitem_at(index)
  File "/home/jpivarski/irishep/awkward/awkward/_v2/contents/indexedoptionarray.py", line 160, in _getitem_at
    return self._content._getitem_at(self._index[where])
  File "/home/jpivarski/irishep/awkward/awkward/_v2/contents/regulararray.py", line 177, in _getitem_at
    raise ak._v2._util.indexerror(self, where)
IndexError: cannot slice RegularArray with 0
```

This was the case before #1568; that PR didn't introduce it.

Alternatively, we could turn the outermost regular-sized lists into variable-sized lists with `ak.from_regular`, perform the same slice, and expect the same result, namely `[[[], []]]`. But instead we get a different result:

```python
>>> array = ak._v2.from_regular(
...     ak._v2.from_numpy(np.array([[[1, 2, 3], [4, 5, 6]]]), regulararray=True),
...     axis=1,
... )
>>> array[:, :, []]
<Array [[]] type='1 * var * 3 * int64'>
```

ioanaif:


jpivarski:
This is out of date and the issue that it addresses was closed by #1597. In particular, https://github.com/scikit-hep/awkward/pull/1597/files#r955454563 points out the removal of `maybe_toNumpyArray`, which we discussed at length because it earlier (this PR) seemed to be necessary. Hopefully these links will help someone in the future who's trying to understand the choices we made!

PR #1597, opened 2022-08-15
fix: empty slice lists of record arrays (#1593)

### LINKED ISSUE #1593: RecordArray loses its field in a slice that doesn't select fields (v2)
### Version of Awkward Array

HEAD

Here's a good slice:

```python
>>> ak._v2.Array([[[1, 2, 3], [4, 5, 6]]])[:, []]
<Array [[[], []]] type='1 * var * var * int64'>
```

There's no record, no field being picked out, and it says, "all of the first dimension, none of the second." The result is correct.

Here's another good slice:

```python
>>> ak._v2.Array([{"x": [[1, 2, 3], [4, 5, 6]]}])[:, [0]]
<Array [{x: [[1, 2, 3]]}] type='1 * {x: 1 * var * int64}'>
```

There's a record array, but a field is not picked out, and it says, "all of the first dimension, one item from the second." The result is correct.

But here's a bad slice:

```python
>>> ak._v2.Array([{"x": [[1, 2, 3], [4, 5, 6]]}])[:, []]
<Array [{}] type='1 * {}'>
```

There's a record array, but a field is not picked out, and it says, "all of the first dimension, none of the second." The result is a record array with no fields: that's wrong—there should still be an _x_ field, even though it will be empty.

Somewhere, an `ak._v2.contents.RecordArray` is not being populated correctly.

### LINKED ISSUE #1601: Advanced indexing doesn't follow NumPy behaviour
### Version of Awkward Array

HEAD

In fixing #1593, I realised that my understanding of NumPy's advanced slicing was missing some important context concerning separated advanced indices:
https://numpy.org/doc/stable/user/basics.indexing.html#basics-indexing:~:text=In%20the%20first,just%20like%20slicing).

Namely, 
> Two cases of index combination need to be distinguished:
> - The advanced indices are separated by a slice, [Ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis) or [newaxis](https://numpy.org/doc/stable/reference/constants.html#numpy.newaxis). For example `x[arr1, :, arr2]`.
> - The advanced indices are all next to each other. For example `x[..., arr1, arr2, :]` but not `x[arr1, :, 1]` since 1 is an advanced index in this regard.
>
> In the first case, the dimensions resulting from the advanced indexing operation come first in the result array, and the subspace dimensions after that. In the second case, the dimensions from the advanced indexing operations are inserted into the result array at the same spot as they were in the initial array (the latter logic is what makes simple advanced indexing behave just like slicing).

It turns out that we don't handle this either. Our current solution in `main` only produces valid solutions in the case where the array can be converted to a regular array, and the advanced slices are passed as empty lists (e.g. `[]`) (thanks to #1568). 
```python
>>> d = np.arange(5 * 7 * 11 * 13 * 14).reshape(5, 7, 11, 13, 14)      
>>> e = ak._v2.contents.NumpyArray(d)
>>> f = ak._v2.operations.to_regular(e, axis=1)
>>> d[:, [], :, []].shape
(0, 5, 11, 14)
>>> f[:, [], :, []].type.show()
0 * 5 * 11 * 14 * int64
```

However, we don't handle the jagged / non-empty case properly:
```python
>>> d[:, [0], :, [0]].shape
(1, 5, 11, 14)
>>> f[:, [0], :, [0]].type.show()
5 * 1 * 11 * 14 * int64
```

I haven't yet thought of what we need to do here. I would assume that we need to visit the index tuple and determine ahead of time whether we will be advanced indexing. It might be that we need another pathway for this.

agoose77:
This PR fixes #1593, and closes #1601, in the following ways:
- Normalises the `[]` index into an integer array, so that it passes through the correct indexing logic
- Ensures the zero-length is set for advanced slices
- Prevents [NumPy advanced-basic mixed indexing](https://docs.scipy.org/doc/numpy/reference/arrays.indexing.html#combining-advanced-and-basic-indexing) (case 1)
- Hardens the `_getitem` logic against non-sized iterables.
   - Normally this is best to let fail with the `len()` check, but in this case we are effectively implementing a `match` and it's much more elegant to ask for permission.

The existing test suite has been modified with this PR. Only one test has a modified expected output:
https://github.com/scikit-hep/awkward/blob/456bae3de7d0edf82c34a2e948b76a7a786cd2ac/tests/v2/test_0024-use-regular-array.py#L23

Some tests now raise intentional errors, so these are required using `pytest.raises`.


src/awkward/_v2/contents/recordarray.py
```diff
@@ -315,7 +315,11 @@ def _getitem_field(self, where, only_fields=()):
                 return self.content(where)._getitem_fields(nexthead, nexttail)
 
     def _getitem_fields(self, where, only_fields=()):
-        indexes = [self.field_to_index(field) for field in where]
+        if len(where) == 0:
+            indexes = range(len(self._contents))
+        else:
+            indexes = [self.field_to_index(field) for field in where]
```
ioanaif:
```suggestion
            indexes = [self.field_to_index(field) for field in where]
            if self._fields is None:
                fields = None
            else:
                if len(where) == 0:
                    fields = [self._fields[i] for i in range(len(self._contents))]
                else:
                    fields = [self._fields[i] for i in indexes]
```

In here `indexes` should not be changed, because they are used later on (L333+) to pick up correct array elements accordingly. 

In your above changes, this test will fail: 

`assert ak._v2.to_list(ak._v2.Array([{"x": [[1, 2, 3], [4, 5, 6]]}])[:, []]) == [{}]` 

Because the slice will output the following:

`ak._v2.to_list(ak._v2.Array([{"x": [[1, 2, 3], [4, 5, 6]]}])[:, []])
[{'x': [[1, 2, 3], [4, 5, 6]]}]`


However, this will mean that when slicing with empty arrays `len(fields)` will be > 0 while `len(contents) == 0` and will cause a `ValueError` at L96 which can be fixed with `if len(contents) > 0 and not len(contents) == len(fields):` (in the issue, @jpivarski pointed out that even if the length is 0, the field should be passed on). 

Moreover, accepting different lengths will require another fix in File "/awkward/_v2/highlevel.py", line 1175, in __repr__ , into
`return self._contents[index] IndexError: list index out of range`

tests/v2/test_1593-empty-slice-list-record.py
```diff
@@ -0,0 +1,6 @@
+import awkward as ak
+
+
+def test():
+    array = ak._v2.Array([{"x": [[1, 2, 3], [4, 5, 6]]}])
+    assert array[:, []].fields == ["x"]
```
ioanaif, line 6:
```suggestion
    assert array[:, []].fields == ["x"]
    assert ak._v2.to_list(ak._v2.Array([{"x": [[1, 2, 3], [4, 5, 6]]}])[:, []]) == [{}]
```

src/awkward/_v2/contents/recordarray.py
agoose77: Ah, thanks for this. I've misunderstood the intent of Jim's test case. The empty slice should still drop the contents, but keep the fields (whereas I was leaving the contents for subsequent slicing (that never happens)).

agoose77:
Following @ioanaif's comments I realise what we're actually needing to fix here. It turns out that we already handle the non-tuple case (`array[[]]`, which performs a naive carry). The simplest way to ensure this for a tuple slice is to just normalise `[]` →`slice(0, 0)` and handle it as normal.

This means that `_getitem_fields` can be left solely to deal with fields, rather than empty slices.

I want to try and unify the normalisation of tuple elements with the fast path.

agoose77:
@jpivarski @ioanaif 

As mentioned briefly in the commits, the current approach I'm taking here is to convert `[]` into `array([], dtype=int64)` as early as possible. This means we can scope `_getitem_fields` to _only_ working with field slices (which is much easier than having to add special case logic to every layout, for which we currently assume that `_getitem_fields` only handles non-empty lists of fields).

This being my first time properly looking at the v2 getitem logic, I wondered if I could make a strong case for dropping the non-tuple (fast-path) codepath in `Content._getitem`? It's not a huge issue, but I believe that this bug probably came about because we have these two separate code paths (we handled empty lists in the fast-path). I'm wondering if it is simply worth dropping the "fast path" altogether in favour of a generic dispatch via `_getitem_next`. My intuition is that the slicing logic itself shouldn't be a bottleneck that we are hugely worried about solving, whereas duplicate implementations that are not exactly identical *is* a problem we will want to avoid.

At the same time, I am also wondering if there's actually a motivating factor for keeping it e.g. a real-world legitimate (perf?) issue that I do not know about?

agoose77:
Additionally, do we have a policy on whether `ak._v2.index.IndexN` are permitted as index items? We seem to prefer arrays (hence 7b138df), and  raise errors for `Index64` into a `NumpyArray` but places like `EmptyArray` have special cases for them.

My first reaction to this is that we should allow `Index64`, and should convert `NumpyArray`s to them during the `getitem` preparation logic.

This discussion is prompted by the above fixes breaking `EmptyArray`.

jpivarski:
The motivation is performance, but it seems like a big deal because a tuple slice has to go through mountains of pre-processing code, whereas a non-tuple slice is usually pretty direct. Going from v1 to v2, non-tuple bool/int array slices were fast-tracked, and I considered that a win.

As for actual measurements, `__iter__` is implemented with simple int slices, and it's pretty slow, slow enough that I think it might be a problem. (Try comparing an iteration through an Awkward Array versus an iteration through an equivalent list, or even an equivalent NumPy array. It's an antipattern, I know, but it's so extreme that maybe something should be done.)

Eliminating the non-tuple path for slices would make the slice logic simpler and would reduce the possibility that a case gets fixed in one path and not the other. But look around at the entirety of the slicing code—it's huge—removing the non-tuple paths might reduce it by 5%, but at a cost that the very most common slices become much slower. I agree that there's simplicity to be gained, but not much and at a large cost in performance. It doesn't sound like a good deal to me.

agoose77:
OK, I was mainly curious if there was a motivating bug report that drove the change. 

If it were just the `__iter__` impl, then even handling only these kind of slices would reduce our duplicating implementation. 
That said, the difference in code is on the smaller side, so I'm happy to follow along with special casing the non-tuple case. I'll have a think about how we can ensure the two stay in sync (besides unit-testing) and/or defining the permitted index types.

jpivarski:
I just remembered something else: `array[100:200]` is a view and `array[100:200,]` is a copy (unless we recognize that the tuple has length 1 and redirect it to the non-tuple code path). In the big, complex tuple-handling getitem, we don't know if a slice in a tuple can be just a view until we get to that level of recursion, and by then, it's too late.

(Recognizing that a slice consists of only ranges, i.e. `slice` objects, and avoiding a copy would be to _add another_ code path, motivated by performance. Now that I think about it, I'm not completely sure that it's possible if any lists are variable-length.)

Still "just a performance thing" (as distinct code paths for the same effective outcome would always be), but it can be pretty significant.

agoose77:
Currently this PR regresses #1567 because we now pre-emptively convert lists into empty `Index64`s. This needs to be fixed. At first glance, it looks like I might want to tweak
```
ak._v2._slicing.getitem_next_array_wrap(
    out, head.metadata.get("shape", (head.length,))
)
```
but that's enough for one day ;)

At that, I've probably broken the test because I was copying 1.0 behaviour, which is not correct IIRC.

NB - in case it's not evident, I use "we" to refer to checked-in code written by _anyone_. I personally feel like that "we" is inclusive and makes it clear that I am not interested in who wrote the code (especially if it's not working as intended) - I would be much more interested discussing the fix :)

agoose77:
OK, these commits technically break the test suite, but bring empty list handling in-line with the rest of Awkward's advanced index handling. So, we need to address #1601 in order to fix this.

agoose77:
I'd like the reviewers to particular focus on whether the changes to the test suite are OK. I've made only small changes, but I really don't want to accidentally sneak a bug into our tests!

src/awkward/_v2/contents/recordarray.py
jpivarski: It sounds like this is resolved, right?

(It's complicated and I'm just catching up.)

agoose77: Yes, it is. I misunderstood the problem at this point.

jpivarski:
My comment about "The motivation is performance" was about the comment before this one, not addressing this one:

> Additionally, do we have a policy on whether `ak._v2.index.IndexN` are permitted as index items? We seem to prefer arrays (hence [7b138df](https://github.com/scikit-hep/awkward/commit/7b138df01b7ad01a5c87201589ac93e7dc344e5d)), and raise errors for `Index64` into a `NumpyArray` but places like `EmptyArray` have special cases for them.
> 
> My first reaction to this is that we should allow `Index64`, and should convert `NumpyArray`s to them during the `getitem` preparation logic.
> 
> This discussion is prompted by the above fixes breaking `EmptyArray`.

The distinction between

```python
ak._v2.index.Index64.empty(0, nplike)
```

and

```python
nplike.empty(0, dtype=np.int64)
```

is that the first returns an Index64 and the second returns an actual NumPy array (or CuPy, etc.). Some functions require an Index (for metadata), such as `_carry`, so there are assertions to make sure they get the right thing.

All of the tuple items are normalized to a fixed set of types before launching into `_getitem_next`. We considered inventing custom classes for all of these slice types, as v1/C++ _had_ to, but I decided against inventing types that Python already distinguishes. However, it would have had the advantage of having auspicious names, like "SliceInt", "SliceRange", "SliceEllipsis", "SliceRectArray", etc.

I looked at [ListOffsetArray._getitem_next](https://github.com/scikit-hep/awkward/blob/2b1f32404a13e91085db5752b78374a4c380935c/src/awkward/_v2/contents/listoffsetarray.py#L354-L622) as a prototypical case, to get a reminder of what the normalized types are. They are:

   * `int` (really a Python `int`, not a NumPy scalar or anything) for integer-slicing
   * `slice` for range-slicing
   * `str` for single field-slicing
   * `list` for multiple field-slicing; if we ever see a bare Python `list` after normalization, we _assume_ it's a list of strings
   * `np.newaxis`/`None` for newaxis-slicing
   * `Ellipsis` for ellipsis-slicing
   * `Index64` (not any other integer-width) for rectilinear array-slicing: so not an actual NumPy array (or CuPy, etc.); note that a single integer gets broadcasted to this when used in the same slice as a rectilinear array
   * `ListOffsetArray` (and not `ListArray`) for ragged array-slicing
   * `IndexedOptionArray` (and not any other option-type) for missing value-slicing

So rectilinear arrays need to be normalized to `Index64`, not `np.ndarray`.

I just checked all of the other `Content._getitem_next` implementations and they all require exactly these types—any one of them would raise an AssertionError if they got an `np.ndarray` instead of an `Index64` (except RecordArray, which passes some types on to whatever's next). I know that I'm acting like a human-MyPy here: explicit type annotations would be better, but that's beyond the scope of this right now.

So I'm going to keep looking into this. `prepare_tuple_item` should not return an `np.ndarray`.

agoose77:
@jpivarski it looks like we do return NumPy Arrays rather than Index64 in `prepare_tuple_item`: e.g. for NumPy arrays themselves
```python
    elif isinstance(item, ak._v2.contents.NumpyArray):
        return item.data
```
These are then converted into Index64 by the broadcast routine.

If we fix this, then we might need to fix the broadcasting, because I don't know if it handles Index64.

jpivarski:
I think the intention was to normalize them all to `Index64`, but the next step (broadcasting) ensures that they're all `Index64`, so no errors because it turned out alright in the end.

Broadcastable types are detected via

https://github.com/scikit-hep/awkward/blob/2b1f32404a13e91085db5752b78374a4c380935c/src/awkward/_v2/_slicing.py#L23-L43

So regardless of whether they're `Index64` or `np.ndarray`, they'll get into the `broadcastable` list.

When they actually get broadcasted,

https://github.com/scikit-hep/awkward/blob/2b1f32404a13e91085db5752b78374a4c380935c/src/awkward/_v2/_slicing.py#L54

any `Index64`s get turned into raw arrays (`nplike`'s array type) because `Index64.__array__` is defined:

https://github.com/scikit-hep/awkward/blob/2b1f32404a13e91085db5752b78374a4c380935c/src/awkward/_v2/index.py#L118-L119

And then these all get wrapped up as `Index64`, so it's alright in the end:

https://github.com/scikit-hep/awkward/blob/2b1f32404a13e91085db5752b78374a4c380935c/src/awkward/_v2/_slicing.py#L61-L86

It could be cleaner, but at least everything is normalized before it gets into `_getitem_next`.

jpivarski:
So 7b138df01b7ad01a5c87201589ac93e7dc344e5d shouldn't have made a difference. For cleaner code, maybe `prepare_tuple_item` should always be returning `np.ndarray` (or `nplike`'s array type) with the knowledge that they will later be normalized to `Index64` by `getitem_broadcast`.

jpivarski:
(The v1 slice-tuple preparation was worse! This is a step forward on a hard problem.)

src/awkward/_util.py
```diff
@@ -37,6 +39,10 @@ def is_file_path(x):
         return False
 
 
+def is_sized_iterable(obj):
+    return isinstance(obj, Iterable) and isinstance(obj, Sized)
```
jpivarski, line 43: Correct: too often, I assume that iterables have a `__len__` and common test examples don't check it. We're talking about `Sized` `Iterables` in anything we'd support in a slice tuple.

src/awkward/_v2/_slicing.py
```diff
@@ -474,17 +482,44 @@ def prepare_tuple_bool_to_int(item):
...
         if isinstance(size, ak._v2._typetracer.UnknownLengthType):
             size = 1
-        out = ak._v2.contents.RegularArray(out, size, length, None, None)
-    return out
+        outcontent = ak._v2.contents.RegularArray(outcontent, size, length, None, None)
+    return outcontent
+
+
+def index_is_advanced_separator(item):
+    return (item is np.newaxis) or (item is Ellipsis) or isinstance(item, slice)
```
jpivarski: Thinking about the name of this: `np.newaxis`, `Ellipsis`, and `slice` are three examples of things that NumPy would not call "advanced slices." I'm pretty sure the word is only used for arrays in a slice.

Maybe this should be named `is_nonadvanced_separator`?

src/awkward/_v2/_slicing.py
```diff
@@ -474,17 +482,44 @@ def prepare_tuple_bool_to_int(item):
...
+
+
+def ensure_supported_tuple(where):
```
jpivarski: This function does check for whether we support a tuple-slice, but it's not the only such place. This:

https://github.com/scikit-hep/awkward/blob/2b1f32404a13e91085db5752b78374a4c380935c/src/awkward/_v2/_slicing.py#L45-L50

ensures that the tuple is supported in the sense of not combining Awkward-indexing (ragged or missing values in arrays) with NumPy advanced indexing (more than one array/int in the tuple, including the Awkward one).

Your function guards against another unsupported case, interleaved advanced and basic NumPy indexing, which is described in the first bullet point and second example of [this section in the NumPy docs](https://numpy.org/doc/stable/user/basics.indexing.html#combining-advanced-and-basic-indexing).

So "`exclude_interleaved_basic_advanced_numpy_indexing`"? It's a mouthful, but the function is only called once.

And maybe the above should be refactored into a function beside this one called "`exclude_mixed_awkward_numpy_advanced_indexing`", so that it's clear that they're both excluding possible cases?

tests/v2/test_0397-arrays-as-constants-in-numba.py
```diff
@@ -100,7 +100,7 @@ def f4():
 
 
 def test_Record():
-    record = ak.Record({"x": 1, "y": [1, 2, 3]})
+    record = ak._v2.Record({"x": 1, "y": [1, 2, 3]})
```
jpivarski, line 103: Yikes! Good catch!

tests/v2/test_0024-use-regular-array.py
```diff
@@ -20,7 +20,7 @@ def test_empty_array_slice():
         a.typetracer[2, 1, np.array([], dtype=int)].form
         == a[2, 1, np.array([], dtype=int)].form
     )
-    assert to_list(a[2, np.array([1], dtype=int), np.array([], dtype=int)]) == [[]]
+    assert to_list(a[2, np.array([1], dtype=int), np.array([], dtype=int)]) == []
```
jpivarski, line 23: Your change to this test is correct. Looking at NumPy (and adjusting the definition of `a` so that it's regular),

```python
>>> a = np.array([[[], [], []], [[], [], []], [[], [], []]])
>>> a[2, np.array([1]), np.array([], dtype=int)].shape
(0,)
```

Selecting the third of three `[[], [], []]`'s, the second item within it (`[]`), and then nothing within that (which is the only thing you can do) does return an empty list, not a list of an empty list. That is,

```python
>>> a[2, np.array([1]), np.array([], dtype=int)].tolist()
[]
```

So yes, this one _should_ be changed.

tests/v2/test_1568-fix-lengths-empty-regular-slices.py
```diff
@@ -54,18 +54,17 @@ def test_lengths_empty_regular_slices():
...
+        _ = e[:, [], :, []]
+
+    with pytest.raises(ValueError):
+        _ = f[:, [], :, []]
+
+    with pytest.raises(ValueError):
+        _ = e[1:4, [], -3:, []]
+
+    with pytest.raises(ValueError):
+        _ = f[1:4, [], -3:, []]
```
jpivarski: This test looks very deliberate, interleaving advanced and basic slices, but since all the other tests pass after you put in a guard against it, this is the only instance in the whole test suite that violates the rule against interleaving.

It evidently worked for this corner case with empty arrays, but there's no guarantee that it will work in general.

I can say that when the whole getitem system was designed, I intentionally gave up on that interleaving case as something that can't be adapted to variable-length arrays. (The slicers are regular, but they're slicing an Awkward Array that might be anything.) It is the only NumPy rule I consciously gave up on, and I really should have documented that as an explicit prohibition as you just did.

When @ioanaif and I added these test cases 10 days ago, I wasn't thinking about that rule, at least not enough to have realized that these cases are not ones that are expected to work.

So yes, we can change this test to say that this is now an error case.

tests/v2/test_1568-fix-lengths-empty-regular-slices.py
```diff
@@ -92,16 +91,6 @@ def test_lengths_empty_regular_slices():
...
-    assert (
-        d[:, [], :, []].shape
-        == ak._v2.to_numpy(e[:, [], :, []]).shape
-        == ak._v2.to_numpy(f[:, [], :, []]).shape
-    )
-    assert (
-        d[1:4, [], -3:, []].shape
-        == ak._v2.to_numpy(e[1:4, [], -3:, []]).shape
-        == ak._v2.to_numpy(f[1:4, [], -3:, []]).shape
-    )
```
jpivarski, line 97: Also interleaving basic and advanced NumPy slices. It's correct to remove them from the test suite.

src/awkward/_v2/contents/content.py
```diff
@@ -522,13 +522,6 @@ def _getitem(self, where):
                 self._nplike,
             )
 
-            if any(isinstance(x, list) and len(x) == 0 for x in items) and any(
-                isinstance(x, slice) for x in items
-            ):
-                attempt = next.maybe_toNumpyArray()
-                if attempt is not None:
-                    next = attempt
```
jpivarski, line 530: It's interesting that it works without this: that RegularArrays that could be a single NumpyArray are being sliced the same way as that NumpyArray, though I know that we put in tests that check for RegularArray-vs-NumpyArray in the corner cases we were concerned about, and they're evidently passing.

(One way to say it is that the `RegularArray._getitem_next` is doing what it's supposed to do. We thought that maybe this was a problem for edge cases, but all of those edge cases are in the test suite.)

The tests that have been removed or changed in the test suite are not ones that this is addressing. So it's fine to remove this code, as you have.

jpivarski (changes requested):
I went over this with a fine-toothed comb because getitem is complicated, and I know you said you were changing some of the expected test results. But (as you can see below), I think everything here is correct.

Some things are messy, like the normalization of `Index64` vs `np.ndarray` (or CuPy, etc.) before the last step of normalization, but the normalization is correct before entering `_getitem_next`, which is the important part.

The whole getitem system can stand to have some developer-focused documentation, or better naming, internal function docstrings, comments, or (if they can be made to do anything at all, given the size of the codebase) type annotations. The getitem system had once been under heavy development, but now it's rather stable and that sort of documentation would not go out of date quickly.

I can also say that it's 1 order of magnitude better than v1's pre-processing of slices and 2 orders of magnitude better than v0's handling of slices. We're converging.

The only actual changes I'm requesting are renaming two functions for clarity (below).

src/awkward/_v2/_slicing.py
agoose77: Haha, I chose the name as these 'primitives' separate advanced array indices! Maybe that is less clear than your suggestion.

agoose77:
I've made changes to this PR since your review. I merged the NumPy indexing case with the Awkward-NumPy logic inside the broadcasting function. To my mind, this broadcasting function is really just doing "advanced indexing preparation", so I don't think we need multiple functions. Validation of the index needs to happen at different times, so although we *could* split the two cases into separate before-and-after functions, I think it's not needed for now.

I, therefore, renamed the existing "prepare" functions to reflect the distinction that they're used for normalising the arguments.

I also found that my validation logic was too naive, so I moved to a state machine to be sure. There's a functional one [here](https://gist.github.com/agoose77/8d5e451383db5da34f844c5dffc2206e) but I thought you might prefer the iterator-based approach in this PR.

> I can also say that it's 1 order of magnitude better than v1's pre-processing of slices and 2 orders of magnitude better than v0's handling of slices. We're converging.

Oh, for sure. I really did not enjoy running through the unboxing logic to fix bugs in v1. I dread to think of v0 if it's worse!

src/awkward/_util.py
agoose77: It's a slight pain that there's not an ABC for sized iterables, but it's not too hard to write our own helper, so I should not complain :P

jpivarski (approved):
These names and error messages are much better.

Your functional validator is quite nice! I don't have an explicit preference for inline iterators; it's just that that was quicker to write for not having to invent a bunch of new names. (Names are the hardest part.)

(The [Awkward 0 getitem, just for JaggedArray](https://github.com/scikit-hep/awkward-0.x/blob/dd885bef15814f588b58944d2505296df4aaae0e/awkward0/array/jagged.py#L517-L787) gets pretty deep and still can't even handle a lot of cases.)

This is ready to squash-and-merge!

PR #1602, opened 2022-08-18
chore: include Python 3.11

henryiii:
Testing this out now that NumPy supports Python 3.11 with binary wheels.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -58,11 +58,15 @@ jobs:
       with:
         submodules: true
 
-    - uses: pypa/cibuildwheel@v2.8.1
+    - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_ARCHS_MACOS: universal2
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
         CIBW_BUILD: cp39-win_amd64 cp36-manylinux_x86_64 cp38-macosx_universal2
+
+    - uses: pypa/cibuildwheel@v2.9.0
+      env:
+        CIBW_BUILD: cp311-manylinux_x86_64
 
     - name: Upload wheels
       uses: actions/upload-artifact@v3
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -62,7 +62,12 @@ jobs:
       env:
         CIBW_ARCHS_MACOS: universal2
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
-        CIBW_BUILD: cp39-win_amd64 cp36-manylinux_x86_64 cp38-macosx_universal2 cp311-manylinux_x86_64
+        CIBW_BUILD: cp39-win_amd64 cp36-manylinux_x86_64 cp38-macosx_universal2
+
+    - uses: pypa/cibuildwheel@v2.9.0
+      if: matrix.os == 'ubuntu-latest'
+      env:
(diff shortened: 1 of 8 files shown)


henryiii:
This will be easier after #1550, so will rebase after that.

henryiii:
This should probably be done instead of https://github.com/scikit-hep/awkward/pull/1583 since that one doesn't add testing for 3.11, but just the new wheel builds only.

.github/workflows/tests.yml
```diff
@@ -58,12 +58,17 @@ jobs:
...
+    - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_ARCHS_MACOS: universal2
         CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
         CIBW_BUILD: cp39-win_amd64 cp36-manylinux_x86_64 cp38-macosx_universal2
 
+    - uses: pypa/cibuildwheel@v2.9.0
+      if: matrix.os == 'ubuntu-latest'
+      env:
+        CIBW_BUILD: cp311-manylinux_x86_64
```
henryiii, line 70: We are hardcoding manylinux1 above, and manylinux1 doesn't support 3.11 (or 3.10, IIRC), so made this a separate step.

.pre-commit-config.yaml
```diff
@@ -19,6 +19,7 @@ repos:
   rev: v2.0.0
   hooks:
   - id: setup-cfg-fmt
+    args: [--include-version-classifiers, --max-py-version=3.11]
```
henryiii, line 22: This got missed recently, you now need `--include-version-classifiers` as of setup-cfg-fmt 2.

pyproject.toml
```diff
@@ -74,6 +74,7 @@ filterwarnings = [
   "ignore:the imp module is deprecated in favour of importlib:DeprecationWarning",  # move to importlib to fix (setuptools issue)
   "ignore:The NumPy module was reloaded:UserWarning",
   "ignore:.*np\\.MachAr.*:DeprecationWarning",  # until https://github.com/numba/numba/issues/7758 is fixed
+  "ignore:module 'sre_.*' is deprecated:DeprecationWarning", # From lark & Python 3.11
```
henryiii, line 77: FYI, this can also be "default", "once", or "all" if you want the warning to still show up in the tests. But they aren't really our fault so I think ignoring them is fine. But there are options.

jpivarski (approved):
Great!

Similar to the last few new tests, I'll flip the switch to start requiring "3.11-dev" in about a week. On my calendar for August 31.

.github/workflows/tests.yml
jpivarski: Is this what becomes easier when we drop Python 3.6? I'm going to open issues for dropping 3.6 in Uproot and Awkward as a reminder.

jpivarski:
To be clear: this updates pybind11 on purpose, right? (I missed that when scanning over the files that have changed, since the whole submodule is one small line.) The new pybind11 commit coincides with [release 2.10.0](https://github.com/pybind/pybind11/releases/tag/v2.10.0), which looks very intentional.

I'll make a note in tomorrow's slides to be careful about accidentally back-porting pybind11 (and I'll be on the lookout for that, too, since it happens so easily).

henryiii:
pybind11 2.10.0 is required for CPython 3.11 support, yes.

PR #1607, opened 2022-08-19
feat: prevent reducers like ak.sum on records (v2)

### LINKED ISSUE #1375: Prevent reducers like ak.sum on records (v1 and v2)
### Version of Awkward Array

HEAD

@chadfreer just brought this up, but I seem to remember @lgray and/or @nsmith- asking about it a long time ago.

The issue is this: if we have records whose fields represent non-Cartesian data, they should not be naively added (or otherwise "reduced").

```python
>>> ak.sum(ak.Array([[{"rho": 1.1, "phi": -0.1}, {"rho": 2.2, "phi": 0.1}]]), axis=1)
<Array [{rho: 3.3, phi: 0}] type='1 * {"rho": float64, "phi": float64}'>
```

Knowing that this is a generic array, we can shrug and say, "I guess it would do that. It has no way of knowing what `rho` and `phi` mean." But if it were a [VectorAwkward2D](https://github.com/scikit-hep/vector/blob/8a289725a31dae1d23564e1119bd46b8c5b15169/src/vector/_backends/awkward_.py#L655), we'd really expect it to take the meaning of `rho` and `phi` into consideration when it adds them, because that's what `+` does:

```python
>>> import vector
>>> vector.register_awkward()
>>> (ak.Array([[{"rho": 1.1, "phi": -0.1}]], with_name="Vector2D") +
...  ak.Array([[{"rho": 2.2, "phi": 0.1}]], with_name="Vector2D")
<VectorArray2D [[{rho: 3.29, phi: 0.0334}]] type='1 * var * Vector2D["rho": floa...'>
```

When either @lgray or @nsmith- brought this up, I said that we couldn't let third-party libraries override reducers like `ak.sum` because they are implemented (in Awkward 1.x) in C++, where it would be very difficult to apply overrides implemented in Python. But in Awkward v2, the relevant code is Python and it would be possible.

But there are two issues here:

   1. The fact that Coffea and Vector override `+` and not `ak.sum` is dangerously misleading: attempts to `ak.sum` records should be prevented for exactly the same reason as `+` is prevented between non-overridden records. That is, apply the rationale behind #457 to reducers, not just ufuncs.
   2. Awkward v2 should provide hooks to override chosen reducers. The syntax could be:

```python
ak._v2.behavior[ak._v2.sum, "VectorArray2D"] = special_function
```

and the `special_function` would be invoked [somewhere in here](https://github.com/scikit-hep/awkward-1.0/blob/34094aa26e498210047428b2ecde6d0b1aa1fc7f/src/awkward/_v2/contents/listoffsetarray.py#L1859-L1921), for ListOffsetArray for example. (Although I thought so while talking with @chadfreer, this can't be implemented through #1126, since this is a behavior that might be discovered deep inside an array, not at the level where a Python object exists.)

There would have to be some rules on when `special_function` is applied. Overrides for `axis <` the depth of the matched name (e.g. `"VectorArray2D"`) wouldn't make a lot of sense, since they cross list/record object boundaries:

```python
>>> ak.sum(ak.Array([[1, 2, 3], [100, 200]]), axis=0)
<Array [101, 202, 3] type='3 * int64'>
```

While thinking about expanding capabilities in v2 is fun, item (1) in the above list is the higher-priority item: we need to avoid misleading users into thinking that `ak.sum(geometric_vectors)` does what they might think it does. An error message is better than a wrong answer.

(This issue is a rare bug-feature-policy!)

ioanaif:
This PR addresses issue #1375: 

If we have records whose fields represent non-Cartesian data, they should not be naively added (or otherwise "reduced")

```python
>>> ak.sum(ak.Array([[{"rho": 1.1, "phi": -0.1}, {"rho": 2.2, "phi": 0.1}]]), axis=1)
<Array [{rho: 3.3, phi: 0}] type='1 * {"rho": float64, "phi": float64}'>
```

In V2, this will now throw a TypeError 
`Error details: no ak.sum overloads for custom types: rho, phi`


src/awkward/_v2/contents/recordarray.py
```diff
@@ -824,30 +824,29 @@ def _reduce_next(
...
-                    mask,
-                    keepdims,
+        reducer_name = reducer.__name__.split(".")[0].lower()
+
+        if behavior is None or not (
+            any(
+                x[0].__module__.split(".")[-1].split("_")[-1] == reducer_name
+                for x in behavior.keys()
+            )
+        ):
```
jpivarski: This test should be implemented in _util.py, similar to

https://github.com/scikit-hep/awkward/blob/9e8c841cd9722efb5211040b0c476710706de69d/src/awkward/_v2/_util.py#L590-L597

and the other functions surrounding it. The `reducer` object isn't guaranteed to have the same name as the `ak.*` function (though right now that happens to be true); we should do it by referential identity matching the function itself. I guess we could attach each of those functions to the reducer in question.

For instance, at the end of each file like https://github.com/scikit-hep/awkward/blob/main/src/awkward/_v2/operations/ak_argmin.py, add

```python
ak._v2._reducers.ArgMin.function = argmin
```

(and similarly for argmin, argmax, count, count_nonzero, sum, prod, any, all, min, max). Then, we can always get the high-level function from the `reducer`, and that's what we need to look up in `behavior`. The second part is `self.parameter("__record__")`, which is not allowed to be `None`.

As a helper function in [_util](https://github.com/scikit-hep/awkward/blob/main/src/awkward/_v2/_util.py), this could look like

```python
def reducer_recordclass(reducer, layout, behavior):
    behavior = Behavior(ak._v2.behavior, behavior)
    rec = layout.parameter("__record__")
    if isstr(rec):
        return behavior[reducer.function, rec]
```

If `ak._v2._util.reducer_recordclass(reducer, self, behavior)` returns `None`, then you should raise the `TypeError`. If it's not `None`, then you have the `NotImplementedError` for now, as a placeholder for future implementation.

Note that this `Behavior` is a way of merging local array `behavior` with the global `ak.behavior` that should be applied consistently for all behavior-lookups, and its `__getitem__` would return `None` if there's no match (so both code paths in `reducer_recordclass` can return `None`).

jpivarski (changes requested):
Great! I just have one requested change. You've added a behavior-lookup, and there's a common form for those in helper functions in _util. See below.

With that behavior-lookup refactored out to _util.py, this will be ready. Re-request me to review it and I'll do a quick check, then merge.

Thanks!

src/awkward/_v2/contents/recordarray.py
```diff
@@ -824,30 +824,23 @@ def _reduce_next(
...
-                    parents,
-                    outlength,
-                    mask,
-                    keepdims,
+        reducer_recordclass = ak._v2._util.reducer_recordclass(reducer, self, behavior)
+        if reducer_recordclass is None:
+            raise ak._v2._util.error(
+                TypeError(
+                    "no ak.{} overloads for custom types: {}".format(
+                        str(reducer).split(".")[-1], ", ".join(self._fields)
```
jpivarski:
```suggestion
                        reducer.name, ", ".join(self._fields)
```

jpivarski (approved):
Discussed in meeting; this is ready to go!

PR #1610, opened 2022-08-19
Public interface for layout.recursively_apply and broadcast_and_apply.

### LINKED ISSUE #516: Public broadcast wrapper and recursively_apply tool
Both functions in `awkward1._util` are used a few times in coffea.
The [wrapper function](https://github.com/scikit-hep/awkward-1.0/blob/master/src/awkward1/_connect/_numpy.py#L41) that broadcasts and flattens inputs for numpy ufuncs was copied in a simplified form to coffea lookup_tools: https://github.com/CoffeaTeam/coffea/blob/master/coffea/lookup_tools/lookup_base.py#L19-L45
The recursively_apply function [is used](https://github.com/CoffeaTeam/coffea/blob/akcache/coffea/nanoevents/methods/base.py#L65-L75) to make sure a index take operates on the flat array in NanoEvents cross-references.

jpivarski:
Fixes #516.

### DIFF
--- a/src/awkward/_v2/contents/bitmaskedarray.py
+++ b/src/awkward/_v2/contents/bitmaskedarray.py
@@ -6,7 +6,7 @@ import math
 
 import awkward as ak
 from awkward._v2.index import Index
-from awkward._v2.contents.content import Content
+from awkward._v2.contents.content import Content, unset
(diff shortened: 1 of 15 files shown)


agoose77:
I like this API a lot, nice work Jim!

PR #1614, opened 2022-08-22
fix: improve `ak.from_iter` performance for long (axis=0) arrays

### LINKED ISSUE #1588: Remove pybind11 bottleneck in `ak.from_iter` (v1 and v2)
After a few hours of reading the docs, I came up with a few simple scenarios where I struggle to achieve performance improvements with `awkward` `1.8.0` with ragged data, and for the most part see slow downs compared to simple iteration and application of NumPy ufuncs, even when I increase the number of records and degree of "jaggedness" in the data.

I'll place a reproducer below, and I think my question is:
- are there prospects for improvement here?
- if not, maybe this is something to document re: appropriate use cases?

```python
from time import perf_counter

import numpy as np
from numpy.testing import assert_allclose
import awkward as ak


def setup_nested_list(num_records):
    # as you increase num_records, you get both
    # more records and more jagged data
    nested_list = []
    for size in range(3, num_records):
        nested_list.append(np.ones(shape=(size, 3)))
    return nested_list


def avg_no_awk(nested_list):
    result_arr = np.empty((len(nested_list), 3),
                          dtype=np.float64)
    for position, subarray in enumerate(nested_list):
        result_arr[position] = np.average(subarray, axis=0)
    return result_arr


def avg_with_awk(nested_list):
    awk_arr = ak.Array(nested_list)
    result_arr = ak.mean(awk_arr, axis=1)
    return ak.to_numpy(result_arr)


if __name__ == "__main__":
    for num_records in [500,
                        5000,
                        7000]:
        print("num jagged records:", num_records)
        nested_list = setup_nested_list(num_records)
        start = perf_counter()
        result_no_awk = avg_no_awk(nested_list=nested_list)
        end = perf_counter()
        without_awkard = end - start
        print("without awkward (s):", without_awkard)
        start = perf_counter()
        result_with_awk = avg_with_awk(nested_list=nested_list)
        end = perf_counter()
        with_awkward = end - start
        print("with awkward (s):", with_awkward)
        if with_awkward < without_awkard:
            print("speedup factor with awkward:", without_awkard / with_awkward)
        else:
            print("slowdown factor with awkward:", with_awkward / without_awkard)
        assert_allclose(result_with_awk, result_no_awk)

```

```
num jagged records: 500
without awkward (s): 0.00580483900012041
with awkward (s): 0.3663976220013865
slowdown factor with awkward: 63.119342671482585
num jagged records: 5000
without awkward (s): 0.2445533330010221
with awkward (s): 36.80084666699986
slowdown factor with awkward: 150.48188554782874
num jagged records: 7000
without awkward (s): 0.45862196600137395
with awkward (s): 72.65730339400034
slowdown factor with awkward: 158.42525823061573

```

This is a simplified version of an attempt to achieve performance improvements downstream (https://github.com/MDAnalysis/mdanalysis/issues/3776). Those cases were smaller, and `ak.Array` creation dominated, but increasing the record counts here doesn't seem too helpful in switching the balance--in fact, it is the opposite?

agoose77:
Fixes #1588

### DIFF
--- a/src/awkward/operations/convert.py
+++ b/src/awkward/operations/convert.py
@@ -887,10 +887,8 @@ def from_iter(
             )
 
     out = ak.layout.ArrayBuilder(initial=initial, resize=resize)
-    for x in iterable:
-        out.fromiter(x)
-    layout = out.snapshot()
-    return ak._util.maybe_wrap(layout, behavior, highlevel)
+    out.fromiter(iterable)
+    return ak._util.maybe_wrap(out[0], behavior, highlevel)
 
 
 def to_list(array):
--- a/src/awkward/_v2/operations/ak_from_iter.py
+++ b/src/awkward/_v2/operations/ak_from_iter.py
@@ -83,12 +83,11 @@ def _impl(iterable, highlevel, behavior, allow_record, initial, resize):
             )
 
     builder = ak.layout.ArrayBuilder(initial=initial, resize=resize)
-    for x in iterable:
-        builder.fromiter(x)
+    builder.fromiter(iterable)
 
     formstr, length, buffers = builder.to_buffers()
     form = ak._v2.forms.from_json(formstr)
 
     return ak._v2.operations.from_buffers(
         form, length, buffers, highlevel=highlevel, behavior=behavior
-    )
+    )[0]


src/awkward/operations/convert.py
```diff
@@ -887,10 +887,8 @@ def from_iter(
             )
 
     out = ak.layout.ArrayBuilder(initial=initial, resize=resize)
-    for x in iterable:
-        out.fromiter(x)
-    layout = out.snapshot()
-    return ak._util.maybe_wrap(layout, behavior, highlevel)
+    out.fromiter(iterable)
+    return ak._util.maybe_wrap(out[0], behavior, highlevel)
```
jpivarski, line 891: I guess `ArrayBuilder.__getitem__` does a snapshot?

Yes, it does:

https://github.com/scikit-hep/awkward/blob/b1f7792899b65f11cd5686d3d3aa3010cecb08b4/src/python/content.cpp#L977-L982

jpivarski (approved):
Well, that was simple! (It's nice when the solution that looks like it's going to work does.) This is a deep, fundamental, radical change that affects every single test, and yet they all pass, so I guess that's it!

There doesn't seem to be anything else to do here, so I'll merge it. Thanks!!!

jpivarski:
Note: this has a user-visible consequence. See https://github.com/scikit-hep/uproot5/pull/694 and https://github.com/scikit-hep/awkward/pull/1642.

PR #1617, opened 2022-08-22
refactor: replace 3 ak.from_json* functions with one

### LINKED ISSUE #192: ak.to_json/ak.from_json should be revamped/organized for v2

jpivarski:
Unifying `ak.from_json` (in-memory), `ak.from_json_file`, and `ak.from_json_schema`.

- [x] verify that `ak.to_json` is up-to-date
- [x] verify that `ak_from_json_new` was actually finished way back in May
- [x] update tests
- [x] integrate the with-schema case into this function

All of the `from_json` cases go to RapidJSON for the parsing and are incremental in the JSON parsed (doesn't load a big string into memory, though it may result in a big ak.Array). For line-delimited JSON, a `line_delimited=True` argument is now necessary. Passing a file name now requires a `pathlib.Path`, rather than a bare `str` (which could be JSON text). All file/non-file cases are wrapped as a Python file-like object for reading in RapidJSON, which releases the GIL at all times except when reading from that file-like object.

Technically, the "refactor" commit type might not be correct because this PR is changing the API, but it's more than just adding a "feature" and it's not a "fix".

The new implementations were mostly copied from the following files, now deleted:

- [include/awkward/forth/SpecializedJSON.h](https://github.com/scikit-hep/awkward/blob/e50a2ed5fba578f3916d15da07b678f1b2a17b65/include/awkward/forth/SpecializedJSON.h)
- [src/libawkward/forth/SpecializedJSON.cpp](https://github.com/scikit-hep/awkward/blob/e50a2ed5fba578f3916d15da07b678f1b2a17b65/src/libawkward/forth/SpecializedJSON.cpp)
- [src/awkward/_v2/operations/ak_from_json_schema.py](https://github.com/scikit-hep/awkward/blob/bd897775de14878258a18d99f03db9ddcae9ca40/src/awkward/_v2/operations/ak_from_json_schema.py)

The last file, in turn, was developed before it as moved; the original location was:

- [src/awkward/_v2/operations/convert/ak_from_json_schema.py](https://github.com/scikit-hep/awkward/blob/6ff41fb299167333d2715cae7c0904d8cd5dba7e/src/awkward/_v2/operations/convert/ak_from_json_schema.py)


jpivarski:
Finally closes #192!

jpivarski:
One thing I forgot: it's still missing documentation. (Not this PR.)

PR #1620, opened 2022-08-23
refactor: from_rdataframe to use LayoutBuilder

ianna:
- [x] Extend `LayoutBuilder` API to support user-allocated `char` buffers
- [x] Replace `static_cast` with `reinterpret_cast` in `GrowableBuffer` and `LayoutBuilder`s
- [x] Cleanup some warnings
- [x] Replace `from_rdataframe` data accumulation to `std::vectors` with `LayoutBuilder`s
- [x] Add tests

### DIFF
--- a/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
@@ -67,7 +67,11 @@ def from_rdataframe(data_frame, column):
         form = ak._v2.forms.from_json(form_str)
         list_depth = form.purelist_depth
         if list_depth > 4:
-            raise ak._v2._util.error(NotImplementedError)
+            raise ak._v2._util.error(
+                NotImplementedError(
+                    "Retrieving arbitrary depth nested containers is not implemented yet."
+                )
+            )
 
         def supported(form):
             if form.purelist_depth == 1:
--- a/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
@@ -104,7 +104,7 @@ def from_rdataframe(data_frame, column):
                 )
             return buffers
 
-        dtype = form_dtype(form)
+        data_type = cpp_type_of[form_dtype(form).name]
 
         # pull in the CppBuffers (after which we can import from it)
         CppBuffers = cppyy.gbl.awkward.CppBuffers[column_type]
@@ -112,9 +112,7 @@ def from_rdataframe(data_frame, column):
 
         if isinstance(form, ak._v2.forms.NumpyForm):
 
-            NumpyBuilder = cppyy.gbl.awkward.LayoutBuilder.Numpy[
-                cpp_type_of[dtype.name]
-            ]
+            NumpyBuilder = cppyy.gbl.awkward.LayoutBuilder.Numpy[data_type]
             builder = NumpyBuilder()
             builder_type = type(builder).__cpp_name__
 
@@ -126,7 +124,7 @@ def from_rdataframe(data_frame, column):
             # NOTE: list_depth == 2 or 1 if its the list of strings
             ListOffsetBuilder = cppyy.gbl.awkward.LayoutBuilder.ListOffset[
                 "int64_t",
-                f"awkward::LayoutBuilder::Numpy<{cpp_type_of[dtype.name]}",
+                f"awkward::LayoutBuilder::Numpy<{data_type}",
             ]
             builder = ListOffsetBuilder()
             builder_type = type(builder).__cpp_name__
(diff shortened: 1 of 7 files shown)


src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -94,61 +94,70 @@ def form_dtype(form):
...
+            builder_type = type(builder).__cpp_name__
+
+            cpp_buffers_self.fill_offsets_and_flatten_2[builder_type](builder)
 
         elif list_depth == 3:
-            depths, offsets_length = CppBuffers.offsets_and_flatten_3(cpp_buffers_self)
+            ListOffsetBuilder = cppyy.gbl.awkward.LayoutBuilder.ListOffset[
+                "int64_t",
+                f"awkward::LayoutBuilder::ListOffset<int64_t, awkward::LayoutBuilder::Numpy<{data_type}>",
+            ]
```
ianna, line 142: I think, this should be generated from a Form. I'll do it as a separate PR after introducing multiple columns.

ianna (commented):
@jpivarski - I'm done with this PR. Please, review when you have time. Thanks!

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -66,7 +66,7 @@ def _wrap_as_record_array(array):
     if form_str.startswith("{"):
         form = ak._v2.forms.from_json(form_str)
         list_depth = form.purelist_depth
-        if list_depth > 3:
+        if list_depth > 4:
             raise ak._v2._util.error(NotImplementedError)
```
jpivarski: Why is there a limit on the list depth? Since these things get generated by recursive functions, I might have expected practical limits (memory use, etc.) but not an in-principle limit like this.

Anyway, the NotImplementedError would have to explain what happened, since I'd be very confused if I got this error and wouldn't think that it's the list depth that's at fault.

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -94,61 +94,70 @@ def form_dtype(form):
...
+            ListOffsetBuilder = cppyy.gbl.awkward.LayoutBuilder.ListOffset[
+                "int64_t",
+                f"awkward::LayoutBuilder::Numpy<{data_type}",
+            ]
+            builder = ListOffsetBuilder()
+            builder_type = type(builder).__cpp_name__
+
+            cpp_buffers_self.fill_offsets_and_flatten_2[builder_type](builder)
 
         elif list_depth == 3:
```
jpivarski, line 138: These things are not made with recursive functions: it's case by case. Do you have plans to generalize this?

jpivarski: These are functions defined in https://github.com/scikit-hep/awkward/blob/main/src/awkward/_v2/cpp-headers/rdataframe_jagged_builders.h, right?

I've also been wondering why this file is not in the https://github.com/scikit-hep/awkward/tree/main/src/awkward/_v2/cpp-headers/awkward directory. If it's because the `awkward` directory is our public API and this file is part of the machinery for rdataframe, then it could be two directories, side by side:

 * src/awkward/_v2/cpp-headers/awkward
 * src/awkward/_v2/cpp-headers/rdataframe

so that it says what it's used for and the tree structure is balanced. That makes it easier for a newcomer to understand what everything is.

tests/v2/test_1620-layout-builders.py
```diff
@@ -0,0 +1,168 @@
...
+    # assert data_frame.GetColumnType("x") == "ROOT::VecOps::RVec<double>"
+
+    ak_array_out = ak.from_rdataframe(
+        data_frame,
+        column="x",
+    )
+    assert ak_array_in.to_list() == ak_array_out["x"].to_list()
+
+
+def test_data_frame_vec_of_vec():
```
jpivarski, line 105: For all of these tests, is the value of `out` correct? They verify that it does not crash, but not that it returns the right value.

jpivarski (approved):
I think it looks good! I have a few questions inline about how things will evolve (will explicit functions be converted to something that works on general depth someday)?

Also, I see the `to_char_buffers` that we had talked about.

jpivarski:
Oh, I'll let you decide when to merge, since you might want to make changes based on my questions. But I approve the PR and you can squash-and-merge it when you're ready.

tests/v2/test_1620-layout-builders.py
ianna: done

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
ianna: the latter is done, the former is going to be addressed in the next PR with implementation of multiple columns.

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py

ianna:
> Oh, I'll let you decide when to merge, since you might want to make changes based on my questions. But I approve the PR and you can squash-and-merge it when you're ready.

Thanks, @jpivarski! I think, I've addressed most of your comments. The recursive method is for the next PR because it has to take multiple columns to avoid multiple event loops.

PR #1624, opened 2022-08-23
feat: add `to_raggedtensor` and `from_raggedtensor`

### LINKED ISSUE #1466: New to/from_raggedtensor and to/from_nestedtensor functions

TensorFlow's RaggedTensor has been around for a while and PyTorch's NestedTensor is new, but they should be conversion targets for Awkward Arrays. And then a tutorial should be written in the spirit of this one:

https://www.tensorflow.org/text/tutorials/text_generation

but using Awkward strings ↔ RaggedTensors instead of making `np.array(["...", "...", ...], dtype=object)`. It's a great example of why that interface would be useful, particularly when paired with `ak.str.*` functions. (We can demonstrate the difference with a scaling test on some enormous dataset. What's an enormous text dataset? I wonder...)

I believe that this is an equivalent for PyTorch's NestedTensor:

https://pytorch.org/tutorials/intermediate/char_rnn_generation_tutorial.html

I don't know if JAX has a ragged array interface, but if it does, let's get that, too. All of these converters are relatively easy, but we'd have to get TensorFlow, PyTorch, and JAX into at least one of the CI tests. TensorFlow and JAX can be hard to install.

agoose77:
Fixes #1466

### DIFF
--- a/src/awkward/_v2/__init__.py
+++ b/src/awkward/_v2/__init__.py
@@ -18,6 +18,7 @@ import awkward._v2._lookup
 # third-party connectors
 import awkward._v2._connect.numpy
 import awkward._v2._connect.numexpr
+import awkward._v2._connect.tensorflow
 import awkward._v2.numba
 
 # high-level interface
--- a/src/awkward/_v2/_connect/tensorflow.py
+++ b/src/awkward/_v2/_connect/tensorflow.py
@@ -1,0 +1,36 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import warnings
+
+import awkward as ak
+
+checked_version = False
+
+
+def import_tensorflow():
+    global checked_version
+    try:
+        import tensorflow
+    except ModuleNotFoundError as err:
+        raise ak._v2._util.error(
+            ModuleNotFoundError(
+                """install the 'tensorflow' package with:
+
+    pip install tensorflow --upgrade
+
+or
+
+    conda install tensorflow"""
+            )
+        ) from err
+    else:
+        if not checked_version and ak._v2._util.parse_version(
+            tensorflow.__version__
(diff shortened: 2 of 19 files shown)


agoose77:
@jpivarski do we want to support converting unused optiontype layouts into RaggedTensors? (where `UnmaskedArray` is a perfect example of "unused optiontype")?

I'm currently leaning into "no", or "not silently" (require a flag to allow unused options).

BTW, we can discuss this tomorrow, I'm just getting it down on GH now.

jpivarski:
I don't think RaggedTensors can allow for missing values. However, if the policy is to turn missing numbers into `np.nan` and missing lists into empty lists, then UnmaskedArrays would just go away.

I think it would be desirable to do that, considering how many missing values come out of Parquet files and Arrow. (And Awkward Array might become the go-between from a Parquet/Arrow source to machine learning.)

agoose77:
> I think it would be desirable to do that, considering how many missing values come out of Parquet files and Arrow. (And Awkward Array might become the go-between from a Parquet/Arrow source to machine learning.)

Counterpoint - we already have `fill_none` et al. to let users do this. Should we keep `to_raggedtensor` simple by requiring users to convert their data first? That way, they also are required to think about what their data should look like after the transformation.

Additionally, I'm not sure how useful `nan` is in a context of ML (for which I'm referrring to TensorFlow's main use case)?

(I'm generally uneasy about making it easier to ignore whether an array has an option, mainly because they can significantly affect the interpretation of the array, and proliferate easily between array transformations.)

I'm leaning into whatever your final take on this is, so let me know ;) We can discuss tomorrow.

agoose77:
As discussed in my weekly meeting, we should just convert `None` to `np.nan` so that the burden is placed on the user in the tensorflow space. There are routines to subsequently convert these values to meaningful results.

agoose77:
This will need rebasing, so I'll close it for now.

PR #1625, opened 2022-08-24
feat: retrieve multiple columns from RDataFrame in a single event loop

ianna:
Currently `from_rdataframe` function converts a selected column to a native Awkward Array. This function pulls one column at a time: this simplifies the interface and Awkward Arrays can be inexpensively joined with awkward `zip`.
This PR is extending this API as the users requested not to trigger RDF event loop multiple times.

- [x] Proposed change to `ak.from_rdataframe`:
from:
```python
out = ak.from_rdataframe(
    rdf,
    column="x",
)
```
to:
```python
out = ak.from_rdataframe(
    rdf,
    columns=("x", "y",),
)
```
- [x] Implement handling multiple columns in `jagged_builders`, generalise for any depth using a recursive function
- [x] Add tests

### DIFF
--- a/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
@@ -11,7 +11,7 @@ import ctypes
 import os
 from awkward._v2.types.numpytype import primitive_to_dtype
 
-_primitive_to_cpp_type_dict = {
+cpp_type_of = {
     "bool": "bool",
     "int8": "int8_t",
     "uint8": "uint8_t",
@@ -56,25 +56,17 @@ assert done is True
 
 def from_rdataframe(data_frame, columns):
     def supported(form):
-        if form.purelist_depth == 1:
-            # special case for a list of strings form
-            return isinstance(
-                form, (ak._v2.forms.ListOffsetForm, ak._v2.forms.NumpyForm)
-            )
+        if isinstance(form, ak._v2.forms.NumpyForm) and form.inner_shape == ():
+            return True
         else:
             return isinstance(form, ak._v2.forms.ListOffsetForm) and supported(
                 form.content
             )
 
     def form_dtype(form):
-        if form.purelist_depth == 1:
-            # special case for a list of strings form
-            return (
-                primitive_to_dtype(form.content.primitive)
-                if isinstance(form, ak._v2.forms.ListOffsetForm)
-                else primitive_to_dtype(form.primitive)
-            )
-        else:
+        if isinstance(form, ak._v2.forms.NumpyForm) and form.inner_shape == ():
+            return primitive_to_dtype(form.primitive)
+        elif isinstance(form, ak._v2.forms.ListOffsetForm):
             return form_dtype(form.content)
 
     def empty_buffers(cpp_buffers_self, names_nbytes):
@@ -184,7 +176,7 @@ def from_rdataframe(data_frame, columns):
 
             list_depth = form.purelist_depth
 
-            data_type = _primitive_to_cpp_type_dict[form_dtype(form).name]
+            data_type = cpp_type_of[form_dtype(form).name]
 
             # pull in the CppBuffers (after which we can import from it)
             CppBuffers = cppyy.gbl.awkward.CppBuffers[col_type]
--- a/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
@@ -11,16 +11,22 @@ import ctypes
 import os
 from awkward._v2.types.numpytype import primitive_to_dtype
 
-cpp_type_of = {
+_primitive_to_cpp_type_dict = {
+    "bool": "bool",
     "int8": "int8_t",
     "uint8": "uint8_t",
+    "int16": "int16_t",
+    "uint16": "uint16_t",
     "int32": "int32_t",
     "uint32": "uint32_t",
     "int64": "int64_t",
     "uint64": "uint64_t",
     "float32": "float",
     "float64": "double",
+    "complex64": "std::complex<float>",
     "complex128": "std::complex<double>",
+    "datetime64": "std::time_t",
+    "timedelta64": "std::difftime",
 }
 
 np = ak.nplike.NumpyMetadata.instance()
@@ -178,7 +184,7 @@ def from_rdataframe(data_frame, columns):
 
             list_depth = form.purelist_depth
 
-            data_type = cpp_type_of[form_dtype(form).name]
+            data_type = _primitive_to_cpp_type_dict[form_dtype(form).name]
 
             # pull in the CppBuffers (after which we can import from it)
             CppBuffers = cppyy.gbl.awkward.CppBuffers[col_type]
--- a/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
@@ -158,7 +158,8 @@ def from_rdataframe(data_frame, columns):
             awkward_contents[col] = layout
 
         else:  # Convert the C++ vectors to Awkward arrays
-            form = ak._v2.forms.from_json(ROOT.awkward.type_to_form[col_type](0))
+            form_str = ROOT.awkward.type_to_form[col_type](0)
+            form = ak._v2.forms.from_json(form_str)
 
             list_depth = form.purelist_depth
             form_dtype_name = form_dtype(form).name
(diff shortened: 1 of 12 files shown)


src/awkward/_v2/operations/ak_from_rdataframe.py
```diff
@@ -3,12 +3,12 @@
...
 
 
-def from_rdataframe(data_frame, column):
+def from_rdataframe(data_frame, columns):
     """
     Args:
         data_frame (`ROOT.RDataFrame`): ROOT RDataFrame to convert into an
              Awkward Array.
-         column (str): A column to be converted to Awkward Array.
+         columns (str or tuple of str): A column or multiple columns to be converted to Awkward Array.
```
ianna, line 11: @jpivarski - now that we've re-introduced `columns`, should we also re-introduce `excluded`?

ianna (commented):
@jpivarski - I think, this PR is ready to be reviewed. Thanks.

src/awkward/_v2/operations/ak_from_rdataframe.py
jpivarski: What was `excluded`?

(If we have a way of making a positive statement, the `columns` we want to return, then there shouldn't also be a way of making the negative statement, the columns to be `excluded`. Otherwise, which list wins? But maybe you meant something else by that.)

ianna: > What was `excluded`?
> 
> (If we have a way of making a positive statement, the `columns` we want to return, then there shouldn't also be a way of making the negative statement, the columns to be `excluded`. Otherwise, which list wins? But maybe you meant something else by that.)

Yes, I agree. It was the list of columns not to include: either take some `columns` or take all but `excluded`.

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -12,10 +12,15 @@
...
-    "float64": "double",
+    "int8": "int8_t",
+    "uint8": "uint8_t",
+    "int32": "int32_t",
+    "uint32": "uint32_t",
     "int64": "int64_t",
+    "uint64": "uint64_t",
+    "float32": "float",
+    "float64": "double",
     "complex128": "std::complex<double>",
```
jpivarski, line 27: Might as well sort this and make it complete. The full set of NumPy dtypes that Awkward handles is

https://github.com/scikit-hep/awkward/blob/352b0dead74846ad2a56d385be4694ec87072a08/src/awkward/_v2/types/numpytype.py#L66-L82

which includes 16-bit integers (signed and unsigned) and `std::complex<float>`. It's okay to not include the time-types, but someday you might want to include conversions to and from [std::chrono::time_point](https://cplusplus.com/reference/chrono/time_point/) and [std::chrono::duration](https://cplusplus.com/reference/chrono/duration/). (It would have to be a conversion because it doesn't look like the memory layout is compatible.)

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -43,140 +48,214 @@
...
+    def supported(form):
+        if form.purelist_depth == 1:
+            # special case for a list of strings form
+            return isinstance(
+                form, (ak._v2.forms.ListOffsetForm, ak._v2.forms.NumpyForm)
+            )
+        else:
+            return isinstance(form, ak._v2.forms.ListOffsetForm) and supported(
+                form.content
+            )
```
jpivarski: Independently checking

```
supported(T) = (is_list(T) or is_primitive(T)) or (is_list(T) and supported(T.content))
```

is completely equivalent to

```
supported(T) = is_primitive(T) or (is_list(T) and supported(T.content))
```

(including lists that are strings, since they're also list-of-primitive).

However, you're excluding fixed-length lists/RegularForms (why?). If there's a good reason for that, you also need to exclude fixed-length lists as NumpyForms with an inner shape.

```suggestion
    def supported(form):
        if isinstance(form, ak._v2.forms.NumpyForm) and form.inner_shape == ():
            return True
        else:
            return isinstance(form, ak._v2.forms.ListOffsetForm) and supported(
                form.content
            )
```

But coming back to the intention: are these the Forms you want to select? Often, if you're taking ListOffsetForm, then you also want ListForm and maybe even RegularForm (in which case, those NumpyForms with non-trivial inner shapes would be acceptable).

Is there some reason you can expect the desired list types to always be ListOffsetForm and not ListForm?

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
```diff
@@ -43,140 +48,214 @@
...
+    def form_dtype(form):
+        if form.purelist_depth == 1:
+            # special case for a list of strings form
+            return (
+                primitive_to_dtype(form.content.primitive)
+                if isinstance(form, ak._v2.forms.ListOffsetForm)
+                else primitive_to_dtype(form.primitive)
+            )
+        else:
+            return form_dtype(form.content)
```
jpivarski: This function is similar to the "`supported`" function above, except that it wouldn't complain about any Form types that have a `content`, including all the list types (ListOffsetForm, ListForm, and RegularForm), but also anything at all with a `content`, such as IndexedOptionForm, IndexedForm, BitMaskedForm, ByteMaskedForm, UnmaskedForm.

Records and unions would simply fail with an AttributeError, since they don't have a `content`. Is this function protected somehow from these possibilities? That is, is there a proof that it will never be called with these other Forms as an input argument (`form`)?

Maybe the function should raise an error on these cases, but it likely needs to be something informative for users, not "`AttributeError: UnionForm does not have an attribute 'content'`".

jpivarski (approved):
The first thing that I did was look at the tests: this is doing exactly what it's supposed to do. The only possibility is that the tests are not general enough: they might fail in strange ways if the types are not within the range of those expected.

Some of our tests cover the whole range of data types/forms, such as [this one](https://github.com/scikit-hep/awkward/blob/main/tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py) (but look elsewhere; the same array constructors were used in a lot of tests, maybe some are easier to copy-paste out than others, such as [the very next one](https://github.com/scikit-hep/awkward/blob/main/tests/v2/test_1300b-same-for-numba.py).)

Another thing that a full suite of tests should do is make sure that none of the functions are assuming that offsets are zero. That can be accomplished by constructing test arrays with a slice: `array[1:]`.

I also saw some things in the code that look insufficiently general (which is why I'm making suggestions for testing), but I don't know the context of how those functions are being used to know whether it's okay or not.

Another thing: is `type(builder).__cpp_name__` a cppyy thing? In general, we can't introduce new double-underscored names because Python might introduce them—the double-underscore means "reserved." But if it's cppyy that's introducing it, then we just have to use its interface.

I'm approving this because I'm going to leave it up to you to decide when it's done and can be merged. My comments below are feedback that might or might not lead to actions. (If the functions that looked insufficiently general to me are protected and can never be used on types they would fail on, then they're fine: nothing to change.) Bottom line: the tests demonstrate that it's doing the right thing, on these specific examples, at least.

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
agoose77: I can answer `__cpp_name__`, it is a `cppyy` built-in: https://cppyy.readthedocs.io/en/latest/misc.html#:~:text=__cpp_name__%3A%20a%20string%20that%20every%20C%2B%2B%20bound%20class%20carries%20and%20contains%20the%20actual%20C%2B%2B%20name%20(as%20opposed%20to%20__name__%20which%20has%20the%20Python%20name).%20This%20can%20be%20useful%20for%20template%20instantiations%2C%20documentation%2C%20etc.

agoose77:
> Another thing: is `type(builder).__cpp_name__` a cppyy thing? In general, we can't introduce new double-underscored names because Python might introduce them—the double-underscore means "reserved." But if it's cppyy that's introducing it, then we just have to use its interface.

I can answer `__cpp_name__`, it is a `cppyy` built-in: https://cppyy.readthedocs.io/en/latest/misc.html#:~:text=__cpp_name__%3A%20a%20string%20that%20every%20C%2B%2B%20bound%20class%20carries%20and%20contains%20the%20actual%20C%2B%2B%20name%20(as%20opposed%20to%20__name__%20which%20has%20the%20Python%20name).%20This%20can%20be%20useful%20for%20template%20instantiations%2C%20documentation%2C%20etc.

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
ianna: done, thanks!

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py
ianna: The `Form` that is checked here is actually generated from the C++ type of the RDF columns such as nested `std::vectors` or nested `RVecs`, so it's chosen to be only `ListOffsetForm`.

jpivarski: That would be a sufficient reason, and it also answers my question about other types (BitMaskedForm, etc.).

src/awkward/_v2/_connect/rdataframe/from_rdataframe.py

PR #1627, opened 2022-08-24
ci: aarch64 should also use cibuildwheel v2.9.0

jpivarski:
Missed when line 62 was updated.


jpivarski:
When this MacOS 3.6 is done, I'm going to merge without requirements.

PR #1628, opened 2022-08-25
docs: add user-guide for header-only layout builder

ManasviGoyal:


ManasviGoyal:
@agoose77 please have a look

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+    jupytext_version: 1.14.1
+kernelspec:
+  display_name: C++14
+  language: C++14
+  name: xcpp14
+---
+
+How to use the header-only LayoutBuilder in C++
+===============================================
+
```
ianna, line 16: I think, the user guide should mention how to get appropriate `-I` and the compiler flag options available from `awkward._v2.config`:
```
yana@iannas-macbook-pro-2 awkward % python -m awkward._v2.config --cflags
-std=c++17 -I/Users/yana/Projects/PR1625/awkward/awkward/src/_v2/cpp-headers
yana@iannas-macbook-pro-2 awkward % python -m awkward._v2.config --cflags-only-I
-I/Users/yana/Projects/PR1625/awkward/awkward/src/_v2/cpp-headers
yana@iannas-macbook-pro-2 awkward % python -m awkward._v2.config --incdir
/Users/yana/Projects/PR1625/awkward/awkward/src/_v2/cpp-headers
```

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+* `EmptyRecord`
+* `Record`
+* `Tuple`
+* `Regular`
+* `Indexed`
+* `IndexedOption`
+* `Unmasked`
+* `ByteMasked`
+* `BitMasked`
+* `Union`
```
ianna: This looks more like a reference guide. I think, would be better to add a link to the reference guide instead of listing them.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+How to use Layout Builders?
+-----------------------------
+
+The following cpp-headers which reside in this [directory](https://github.com/scikit-hep/awkward/tree/main/src/awkward/_v2/cpp-headers/awkward) are needed to use Layout Builders.
+
+1. BuilderOptions.h
+2. GrowableBuffer.h
+3. LayoutBuilder.h
+4. utils.h
+
```
ianna, line 50: I'd recommend to use an awkward installation (e.g. a versioned set of files  with pip install) and the config options described above. Otherwise, the users might copy a master branch at some point and we'd have difficulties tracing the bugs reports.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+```{code-cell}
+#pragma cling add_include_path("../../src/awkward/_v2/cpp-headers")
+
+#include "awkward/LayoutBuilder.h"
+
+#include <string>
+#include <fstream>
+
+#include "nlohmann/json.hpp"
```
ianna: I'd move it down when you use JSON - in general, the users will not need it.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+```
+
+```{code-cell}
+using UserDefinedMap = std::map<std::size_t, std::string>;
```
ianna, line 90: this belongs on the line 108

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+1. <b>Constructing a Layout Builder:</b> from variadic templates. (It is implicit template instantiation).
+2. <b>Filling the Layout Builder:</b> while repeatedly walking over the raw pointers within the LayoutBuilder.
+3. <b>Taking the data out to user allocated buffers:</b> Then user can pass them to Python if they want.
+
+Example
+-------
+
+Below is an example for creating RecordArray with NumpyArray and ListOffsetArray as fields.
+
+Type alias can be used for each builder class.
```
ianna: I think, it should be moved to line 83.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+
```
ianna, line 80: Please, add some clarification why there is only one header file included. You've mentioned a few before.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+```{code-cell}
+#pragma cling add_include_path("../../src/awkward/_v2/cpp-headers")
```
ianna: perhaps, move it out in a separate cell and mention that it is here for the purpose of demonstration. The user will not need to include it.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+
+#include "awkward/LayoutBuilder.h"
+
+#include <string>
+#include <fstream>
```
ianna: is this one used?

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,260 @@
...
+
+```{code-cell}
+RecordBuilder<
+  RecordField<Field::one, NumpyBuilder<double>>,
+  RecordField<Field::two, ListOffsetBuilder<int64_t,
+      NumpyBuilder<int32_t>>>
+> builder(fields_map);
+
+//  builder.set_field_names(fields_map);
+```
```
ianna, line 134: or alternative code fragment with setting the map via function and in addition to this one without commented out line?

ianna (changes requested):
Looks good! I have a few comments. Please, keep in mind that, in general, a user would copy-paste the code.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
agoose77: Yep, that's definitely a todo :)

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -0,0 +1,267 @@
...
+
+```{code-cell}
+template<class PRIMITIVE>
+using NumpyBuilder = awkward::LayoutBuilder::Numpy<PRIMITIVE>;
+
+template<class PRIMITIVE, class BUILDER>
+using ListOffsetBuilder = awkward::LayoutBuilder::ListOffset<PRIMITIVE, BUILDER>;
+```
+
+If multiple Record Builders are used in a Builder, then a user-defined `field_map` has to be provided for each Record Builder.
```
agoose77: @ManasviGoyal I haven't looked closely at `LayoutBuilder` yet: could you elaborate here to show how to use single RecordBuilders without a user-defined `field_map`, if that is possible?

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
agoose77: Oops, that snuck in there I believe.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
ManasviGoyal: We can't use the RecordBuilders without a user-defined field_map for now since we are limited to C++14 and can't send strings as template parameters in Record Fields. Although we have two methods to set the field names. Will it be okay in terms of the code cell, if I add the alternative method in a separate cell right after the first method?

Please let me know when you are done with the changes. I`ll address some of the comments by Ianna and make some other changes. Thanks.

agoose77: It is allowed from a c++ standpoint, but what would the builder do if you called both methods, would it error? If so, then you could create a new builder to demonstrate it.

agoose77: I'm finished for now, besides fixing the CI, so my changes shouldn't affect you.

ManasviGoyal: It doesn't throw any error, so I can just put it in a separate code cell.

Thanks. I will add my changes.

ManasviGoyal:
@ianna Should I also add the `dump()` method to print the output to show the buffer contents?

ianna:
> @ianna Should I also add the `dump()` method to print the output to show the buffer contents?

I don't think it is needed here. You have a link to the example in c++.

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
ManasviGoyal: Can you please elaborate? Not sure how the users will be using this, Thanks.

agoose77:
This needs a bit more work on the CI side (me!)

docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
ianna: Awkward Array can be installed [from PyPI](https://pypi.org/project/awkward) using pip:

`pip install awkward`

To get the compiler options needed to pick up the `LayoutBuilder` from this installation:
`python -m awkward._v2.config --cflags`

A user would need to pass these options to the compiler in order to use it.

agoose77:
As discussed in our meeting, we'll drop cling here as it's not hugely motivated.

jpivarski:
As discussed today, I just need a readable copy of the text to send to the ctapipe developers; this PR can be merged later. It looks to me like I can just use this file:

https://github.com/scikit-hep/awkward/blob/manasvi/layout-builder-user-guide/docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md

(They can read markdown.) Is that it? I plan to send this to them by email.

ManasviGoyal:
> As discussed today, I just need a readable copy of the text to send to the ctapipe developers; this PR can be merged later. It looks to me like I can just use this file:
> 
> https://github.com/scikit-hep/awkward/blob/manasvi/layout-builder-user-guide/docs-sphinx/user-guide/how-to-use-header-only-layoutbuilder.md
> 
> (They can read markdown.) Is that it? I plan to send this to them by email.

Yes. This can be used. If I am right, I think PR is failing due to some issues on the CI side for running the C++ code blocks but that shouldn't be any issue if we are going to send only the Markdown file.

Please let me know if anything else needs to be added in the user-guide. Should I also add instructions about Awkward installation and compiler flag options like Ianna mentioned? I wasn't sure how to add the instructions for that.

Thanks.

agoose77:
@jpivarski here's an HTML rendered variant of @ManasviGoyal's docs:
[index.html.txt](https://github.com/scikit-hep/awkward/files/9533138/index.html.txt)

The default PDF export was less appealing than the HTML generation (I used the single-document exporter which doesn't use Sphinx). I can render a PDF if needs be, though.

ManasviGoyal:
Hi, @agoose77 I have added a few changes. Please have a look in case anything needs to fixed. 

Are they any other changes pending for this PR?

agoose77:
@ManasviGoyal my force-push alias was set to `git push --force` instead of `git push --force-with-lease` ... so I managed to overwrite your changes. Sorry! I recovered a single commit (the old HEAD) [5f06f73](https://github.com/scikit-hep/awkward/commit/5f06f737392bdb6663c79ce0359b5ebce1d68f32), could you confirm that you only added a single commit, and it was [5f06f73](https://github.com/scikit-hep/awkward/commit/5f06f737392bdb6663c79ce0359b5ebce1d68f32)? I suggest not pulling this branch until you're confident it's correct! :)

ManasviGoyal:
> @ManasviGoyal my force-push alias was set to `git push --force` instead of `git push --force-with-lease` ... so I managed to overwrite your changes. Sorry! I recovered a single commit (the old HEAD) [5f06f73](https://github.com/scikit-hep/awkward/commit/5f06f737392bdb6663c79ce0359b5ebce1d68f32), could you confirm that you only added a single commit, and it was [5f06f73](https://github.com/scikit-hep/awkward/commit/5f06f737392bdb6663c79ce0359b5ebce1d68f32)? I suggest not pulling this branch until you're confident it's correct! :)

yes it is only one commit - [5f06f73](https://github.com/scikit-hep/awkward/commit/5f06f737392bdb6663c79ce0359b5ebce1d68f32)

PR #1629, opened 2022-08-25
ci: add Linux ROOT build to run ROOT-based tests

ianna:


.github/workflows/build-test.yml
```diff
@@ -219,3 +219,81 @@ jobs:
       - name: Codecov
         run: 'bash <(curl -s https://codecov.io/bash)'
         if: matrix.python-version == '3.9'
+
+  Linux-ROOT:
```
ianna, line 223: `check-yaml` fails here:
```
 Check failure on line 223 in .github/workflows/build-test.yml

 GitHub Actions
/ .github/workflows/build-test.yml

Invalid workflow file
You have an error in your yaml syntax on line 223
```

ianna (commented):
@agoose77  - I'd appreciate any pointers on how to define this. Thanks!

agoose77:
@ianna I think that should fix the YAML definition :crossed_fingers: . I find it helpful to pass it through https://jsonformatter.org/yaml-validator to find the true cause of the parse failure.

I also removed the Python setup block, which should be redundant with Conda. We might need to start rethinking how to define our development requirements, because I'm also now using conda in the CI for the docs.

ianna:
> @ianna I think that should fix the YAML definition 🤞 . I find it helpful to pass it through https://jsonformatter.org/yaml-validator to find the true cause of the parse failure.
> 
> I also removed the Python setup block, which should be redundant with Conda. We might need to start rethinking how to define our development requirements, because I'm also now using conda in the CI for the docs.

Thanks! We can discuss it at the meeting today :-)

ianna:
@jpivarski - I'm done with this PR. Please, have a look when you have time. Thanks!

.github/workflows/build-test.yml
jpivarski: To get quicker turn-around on YAML formatting issues, I try loading them in [PyYAML](https://pypi.org/project/PyYAML/) to see if there are any errors.

```python
>>> import yaml
>>> ok = yaml.safe_load(open(".github/workflows/build-test.yml"))
```

It's not usually necessary to look at the value of `ok`, though that can help diagnose a correct-syntax-but-wrong-nesting or other semantic issue.

YAML has some nasty corner-cases, such as `on`/`off` and `yes`/`no` mean `true`/`false`, not the strings `"on"`/`"off"` and `"yes"`/`"no"` (even as keys of a dict).

```python
>>> yaml.safe_load("""
... on: 1
... off: 2
... yes: 3
... no: 4
... true: 5
... false: 6
... """)
{True: 5, False: 6}
```

When in doubt, quote it.

The use of "`safe_yaml`" is because YAML, as a format, is allowed to run arbitrary code. We know that our own YAML is not malicious, but I don't even remember the command for the non-safe loader.

setup.cfg
```diff
@@ -20,6 +20,7 @@ classifiers =
     Operating System :: MacOS :: MacOS X
     Operating System :: Microsoft :: Windows
     Operating System :: POSIX :: Linux
+    Operating System :: POSIX :: Linux-ROOT
     Operating System :: Unix
```
jpivarski: setup.cfg is not involved in configuring CI. This list here is just the list of OSes that Awkward is known to run on, so "Linux-ROOT" is not instance to be listed.

```suggestion
    Operating System :: POSIX :: Linux
    Operating System :: Unix
```

jpivarski (approved):
I see that the Linux-ROOT test does not contain these lines:

```
SKIPPED [1] tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py:7: could not import 'ROOT': No module named 'ROOT'
SKIPPED [1] tests/v2/test_1374-to-rdataframe.py:10: could not import 'ROOT': No module named 'ROOT'
SKIPPED [1] tests/v2/test_1473-from-rdataframe.py:10: could not import 'ROOT': No module named 'ROOT'
SKIPPED [1] tests/v2/test_1477-generator-entry-type-as-rvec.py:10: could not import 'ROOT': No module named 'ROOT'
SKIPPED [1] tests/v2/test_1508-awkward-from-rdataframe.py:11: could not import 'ROOT': No module named 'ROOT'
SKIPPED [1] tests/v2/test_1613-generator-tolayout-records.py:10: could not import 'ROOT': No module named 'ROOT'
SKIPPED [1] tests/v2/test_1620-layout-builders.py:8: could not import 'ROOT': No module named 'ROOT'
```

and it does contain these lines:

```
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_EmptyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_NumpyArray_shape PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_RegularArray_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_RegularArray_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_RegularArray_NumpyArray_v2b PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_ListArray_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_ListArray_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_ListOffsetArray_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_ListOffsetArray_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_RecordArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_IndexedArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_IndexedOptionArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_ByteMaskedArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_BitMaskedArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_UnmaskedArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_UnionArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_NumpyArray_shape PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_RegularArray_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_RegularArray_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_ListArray_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_ListArray_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_ListOffsetArray_NumpyArray[False] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_ListOffsetArray_NumpyArray[True] PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_RecordArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_IndexedArray_NumpyArray PASSED [ 85%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_IndexedOptionArray_NumpyArray PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_ByteMaskedArray_NumpyArray PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_BitMaskedArray_NumpyArray PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_UnmaskedArray_NumpyArray PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_nested_UnionArray_NumpyArray PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_ListArray_strings PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_RegularArray_strings PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_NumpyArray_iterator PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_NumpyArray_iterator2 PASSED [ 86%]
tests/v2/test_1300-awkward-to-cpp-converter-with-cling.py::test_NumpyArray_riterator PASSED [ 86%]
```

So it really is running and testing ROOT. I also see that there are no changes to any other tests, so this is a very non-disruptive PR.

I do have to revert one line (see below), but then I'll squash-and-merge. Thanks!

jpivarski:
Oh, and this will become a required test on September 1.

.github/workflows/build-test.yml
agoose77: `yaml.full_load` is an alias for `yaml.load(..., loader=yaml.FullLoader)` IIRC

agoose77:
@jpivarski @ianna should we consider running *only* the ROOT tests for this job? Our CI is already fairly heavy. I was thinking we could use something like https://docs.pytest.org/en/7.1.x/example/markers.html to run *only* ROOT tests here

jpivarski:
I think the compilation/installation time might still exceed the testing time. (This job is not doing the kernel unit tests, for the reason that they'd be superfluous.)

Yes: 1 minute to get conda + 2 minutes to install ROOT + 6 minutes to compile Awkward = 9 minutes versus 3 minutes to run all tests.

Dropping v1 will speed up the compilation, but the compilation will still take at least 1 minute. This optimization is not fixing a bottleneck, so I'd just as soon ignore it. (My mental model of optimization is an [L-∞ metric](https://en.wikipedia.org/wiki/L-infinity).)

agoose77:
Thanks for crunching the numbers. That's not a significant gain if this is only a single job. Better to focus on bringing down e.g compilation times (bring on the big split!)

PR #1641, opened 2022-08-29
perf: `ak.concatenate` should flatten for first axis, single-array

### LINKED ISSUE #1049: ak.concatenate with a single array argument effectively flattens; it should short-circuit to ak.flatten with axis=1
### Version of Awkward Array

?

```python
ak.concatenate(some_array_with_more_than_one_dimension)
```

whether that array is an Awkward Array or NumPy (or CuPy? how to check all the libraries?), it should instead `ak.flatten` it with `axis=1`. That's because `ak.concatenate` _does_ effectively flatten, but not in a way that was designed to be remotely CPU or memory efficient. (It iterates over the items as a Python iterable, making a _lot_ of little arrays to concatenate.)

agoose77:
This PR does the following:
- move `is_XXX_buffer` to `awkward.nplike`
- add case for Jax arrays to `to_layout`
- allow caller to specify fallback `NumpyLike` if `ak.nplike.of` does not find an array

During the development of this PR, I noticed that we seem to allow records to be passed to `ak.concatenate`. Is this intentional? It doesn't seem right to me, but I've not looked at this code in a while.

We can't test that this is faster than `ak.concatenate(list(array))`, but that seems reasonable.

I might add logic to check for array dimension correctness in `ak.concatenate` rather than delegating this to `ak.flatten`. Additionally, in its present form, I chose to treat this as a fast-path rather than an exclusive case, so I'm using an early return rather than `if: else:`.

Fixes #1049

### DIFF
--- a/src/awkward/_v2/_util.py
+++ b/src/awkward/_v2/_util.py
@@ -1200,21 +1200,3 @@ def to_arraylib(module, array, allow_missing):
         raise ak._v2._util.error(
             ValueError(f"{module.__name__} is not supported by to_arraylib")
         )
-
-
-def is_numpy_buffer(array):
-    import numpy as np
-
-    return isinstance(array, np.ndarray)
-
-
-def is_cupy_buffer(array):
-    return type(array).__module__.startswith("cupy.")
-
-
-def is_jax_buffer(array):
-    return type(array).__module__.startswith("jaxlib.")
-
-
-def is_jax_tracer(tracer):
-    return type(tracer).__module__.startswith("jax.")
--- a/src/awkward/_v2/index.py
+++ b/src/awkward/_v2/index.py
@@ -160,12 +160,12 @@ class Index:
 
     def __getitem__(self, where):
         out = self._data[where]
-        from awkward.nplike import is_jax_buffer
-        from awkward.nplike import is_cupy_buffer
 
         if hasattr(out, "shape") and len(out.shape) != 0:
             return Index(out, metadata=self.metadata, nplike=self.nplike)
-        elif (is_jax_buffer(out) or is_cupy_buffer(out)) and len(out.shape) == 0:
+        elif (ak.nplike.is_jax_buffer(out) or ak.nplike.is_cupy_buffer(out)) and len(
+            out.shape
+        ) == 0:
             return out.item()
         else:
             return out
--- a/src/awkward/_v2/index.py
+++ b/src/awkward/_v2/index.py
@@ -160,7 +160,8 @@ class Index:
 
     def __getitem__(self, where):
         out = self._data[where]
-        from awkward._v2._util import is_cupy_buffer, is_jax_buffer
+        from awkward.nplike import is_jax_buffer
+        from awkward.nplike import is_cupy_buffer
 
         if hasattr(out, "shape") and len(out.shape) != 0:
             return Index(out, metadata=self.metadata, nplike=self.nplike)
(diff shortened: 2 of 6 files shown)


jpivarski:
> During the development of this PR, I noticed that we seem to allow records to be passed to `ak.concatenate`. Is this intentional? It doesn't seem right to me, but I've not looked at this code in a while.

It was intentional, though there have been other cases in which we've scaled back the use of records in functions designed for numerical arrays.

I don't see any logical problems with it, though. You might want to concatenate all of the arrays of the fields of your record.

```python
>>> array = ak._v2.Array([[{"x": [1]}, {"x": [1, 2]}], [], [{"x": [1, 2, 3]}]])
>>> array.show(type=True)
type: 3 * var * {
    x: var * int64
}
[[{x: [1]}, {x: [1, 2]}],
 [],
 [{x: [1, 2, 3]}]]
>>> ak._v2.concatenate([array, array], axis=0).show(type=True)
type: 6 * var * {
    x: var * int64
}
[[{x: [1]}, {x: [1, 2]}],
 [],
 [{x: [1, 2, 3]}],
 [{x: [1]}, {x: [1, 2]}],
 [],
 [{x: [1, 2, 3]}]]
>>> ak._v2.concatenate([array, array], axis=1).show(type=True)
type: 3 * var * {
    x: var * int64
}
[[{x: [1]}, {x: [1, 2]}, {x: [1]}, {x: [1, 2]}],
 [],
 [{x: [1, 2, 3]}, {x: [1, 2, 3]}]]
>>> ak._v2.concatenate([array, array], axis=2).show(type=True)
type: 3 * var * {
    x: var * int64
}
[[{x: [1, 1]}, {x: [1, 2, 1, 2]}],
 [],
 [{x: [1, 2, 3, 1, 2, 3]}]]
```

This was developed in the spirit of "let's generalize all NumPy functions to all data types," not "does anybody actually have a use-case for this?"

Is it too weird? Should it be inhibited?

Oh, maybe this is a use-case:

```python
>>> arrays_from_uproot = ak._v2.Array([
...     {"Muon_pt": [1.0, 1.1], "Muon_eta": [2.0, 2.2], "Muon_phi": [3.0, 3.3]},
...     {"Muon_pt": [], "Muon_eta": [], "Muon_phi": []},
...     {"Muon_pt": [1.0, 1.1, 1.2], "Muon_eta": [2.0, 2.2, 2.4], "Muon_phi": [3.0, 3.3, 3.6]},
... ])
>>> ak._v2.concatenate([arrays_from_uproot] * 2, axis=1).show(type=True)
type: 3 * {
    Muon_pt: var * float64,
    Muon_eta: var * float64,
    Muon_phi: var * float64
}
[{Muon_pt: [1, 1.1, 1, 1.1], Muon_eta: [2, ..., 2.2], Muon_phi: [...]},
 {Muon_pt: [], Muon_eta: [], Muon_phi: []},
 {Muon_pt: [1, 1.1, 1.2, 1, 1.1, 1.2], Muon_eta: [2, ...], Muon_phi: ..., ...}]
```

Although we generally like `N * var * {field1: float64, field2: float64, ...}` type arrays for physics, what we get from some sources (NanoAOD) are `N * {field1: var * float64, field2: var * float64, ...}`, and this feature of `ak.concatenate` would allow them to be used in the form they have right out of Uproot (without `how="zip"`).

jpivarski (approved):
An excellent implementation as usual! Moving nplike's helper functions from _util.py to nplike.py is fine.

We're going to be able to simplify `ak.nplike.of` when we drop v1.

PR #1642, opened 2022-08-29
fix: ak.from_iter should interpret top-level tuples as ak.Array (v1 and v2).

jpivarski:
This was accidentally the behavior before #1614, and some libraries assumed it. Now the policy is as described in test_1642-from_iter-of-tuples (both v1 and v2).


agoose77:
Huh, good catch @jpivarski. I hadn't fully appreciated that `tuple` was special cased in `ArrayBuilder`. Thanks for the fix.

PR #1645, opened 2022-08-30
fix: include `self._length` in `RegularArray.mergemany`

### LINKED ISSUE #1644: Inconsistency in ak.concatenate when dim length is zero
### Version of Awkward Array

main@8c9bf49

Sorry, it's me again with another edge case...

```python
import awkward._v2 as ak
import numpy as np

a0 = ak.to_regular(ak.Array(np.empty((5, 0))), 1)
a1 = ak.to_regular(ak.Array(np.empty((8, 0))), 1)

a0.type
# ArrayType(RegularType(NumpyType('float64'), 0), 5)

a1.type
# ArrayType(RegularType(NumpyType('float64'), 0), 8)

ak.concatenate([a0, a1]).type
# ArrayType(RegularType(NumpyType('float64'), 0), 8)
```

I'd expect this to return an array of shape `(13, 0`) instead of `(8, 0)`.

agoose77:
Fixes #1644

### DIFF
--- a/src/awkward/_v2/contents/regulararray.py
+++ b/src/awkward/_v2/contents/regulararray.py
@@ -707,7 +707,7 @@ class RegularArray(Content):
         elif all(x.is_RegularType and x.size == self.size for x in others):
             parameters = self._parameters
             tail_contents = []
-            zeros_length = 0
+            zeros_length = self._length
             for x in others:
                 parameters = ak._v2._util.merge_parameters(
                     parameters, x._parameters, True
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is absolutely the right fix. (I just tried it out.)

I think the mistake was made when I was thinking that `others` is the full set of things to merge, forgetting that `self` should also be in that set?

agoose77:
> I think the mistake was made when I was thinking that others is the full set of things to merge, forgetting that self should also be in that set?

Yep, I think that's probably how it happened :)

PR #1650, opened 2022-08-30
fix: pass a copy of `RecordArray`'s internal fields in HL API

### LINKED ISSUE #1648: ak.fields (v2) passes a RecordArray's internal fields by reference
Okay, so I hadn't noticed that Awkward v2's fields are passed by reference, which exposes them to the danger that someone might modify them downstream:

v1:

```python
>>> array = awkward.Array([{"x": 1, "y": 1.1}])
>>> fields = awkward.fields(array)
>>> array
<Array [{x: 1, y: 1.1}] type='1 * {"x": int64, "y": float64}'>
>>> fields
['x', 'y']
>>> fields[0] = "XXX"
>>> fields
['XXX', 'y']
>>> array
<Array [{x: 1, y: 1.1}] type='1 * {"x": int64, "y": float64}'>
```

v2:

```python
>>> array = awkward._v2.Array([{"x": 1, "y": 1.1}])
>>> fields = awkward._v2.fields(array)
>>> array
<Array [{x: 1, y: 1.1}] type='1 * {x: int64, y: float64}'>
>>> fields
['x', 'y']
>>> fields[0] = "XXX"
>>> fields
['XXX', 'y']
>>> array
<Array [{XXX: 1, y: 1.1}] type='1 * {XXX: int64, y: float64}'>
```

It could be fixed [here, in Awkward](https://github.com/scikit-hep/awkward/blob/352b0dead74846ad2a56d385be4694ec87072a08/src/awkward/_v2/contents/recordarray.py#L162), or maybe [here](https://github.com/scikit-hep/awkward/blob/352b0dead74846ad2a56d385be4694ec87072a08/src/awkward/_v2/operations/ak_fields.py#L30) (to only suffer the list-copy when handing it off to a user, so that internal uses can still be by reference).

I'll use this comment to open an issue in Awkward. Once `awkward.fields` is guarded, your `.copy()` can be removed, but it can also not be removed with no consequences but a little performance.

_Originally posted by @jpivarski in https://github.com/scikit-hep/vector/pull/226#discussion_r958660705_

Saransh-cpp:
Fixes #1648 

The tests pass locally by running -
```
└─$ python3 -m pytest -vv -rs tests/v2/
```


jpivarski (approved):
Perfect! This solves it.

jpivarski:
@agoose77, I think we should leave `fields` as a list (following up on https://github.com/scikit-hep/awkward/issues/1648#issuecomment-1232234075) and merge this, taking it as the first step in establishing a policy that high-level functions (`ak.*`, any public methods on ak.Array/ak.Record, including `__getitem__`, ufuncs) ensure that the data they return are not mutable references inside the Awkward Array objects. However, low-level interfaces (e.g. Content methods) do not make such a guarantee.

Do you agree?

jpivarski:
Either way, @all-contributors please add @Saransh-cpp for code

src/awkward/_v2/operations/ak_fields.py
```diff
@@ -27,4 +29,4 @@ def fields(array):
 
 def _impl(array):
     layout = ak._v2.operations.to_layout(array, allow_record=True, allow_other=False)
-    return layout.fields
+    return copy.copy(layout.fields)
```
agoose77: I'd suggest using `.copy` rather than `copy.copy` here, because we know that we have a list.

Saransh-cpp: Done!

jpivarski: Oh, I forgot that—I had been (wrongly) thinking that it could be a list or `None`, but no: leaf-nodes (NumpyForm and EmptyForm) return an empty list instead of `None`.

https://github.com/scikit-hep/awkward/blob/1a0858e20cc2af56b2fe01541721f88ac61e2f20/src/awkward/_v2/forms/numpyform.py#L198-L200

https://github.com/scikit-hep/awkward/blob/1a0858e20cc2af56b2fe01541721f88ac61e2f20/src/awkward/_v2/forms/emptyform.py#L89-L91

Well, the same is true of `parameters`: we know that we have a dict:

https://github.com/scikit-hep/awkward/blob/1a0858e20cc2af56b2fe01541721f88ac61e2f20/src/awkward/_v2/forms/form.py#L259-L263

And a dict also has a `.copy()`. I'll put that in.

agoose77: Taking a `.copy()` of a dict would be a shallow copy w.r.t the values. Is there any chance that we want a deep copy? Are there any uses of `parameters` that involve setting a mutable object e.g. dict, list etc.

jpivarski: (So would `copy.copy`: also shallow.)

In principle, `parameters` can be any JSON-able thing. In practice, they've always been strings. The way it looks like things are going, they'll always be strings. If we have some non-string lists and dicts, then in principle we'd have to protect them with `deepcopy`.

```python
>>> parameters = {
...     "__record__": "SomeLongName",
...     "__categorical__": False,
...     "units": "cm",
...     "__doc__": "This is some thing. Wow."
... }
>>> def copy1(what, how_many):
...     for _ in range(how_many):
...         tmp = what.copy()
... 
>>> def copy2(what, how_many):
...     for _ in range(how_many):
...         tmp = copy.copy(what)
... 
>>> def copy3(what, how_many):
...     for _ in range(how_many):
...         tmp = copy.deepcopy(what)
... 
>>> def copy4(what, how_many):
...     for _ in range(how_many):
...         tmp = json.loads(json.dumps(what))
... 
>>> start = time.time(); copy1(parameters, 1000000); time.time() - start
0.05509829521179199
>>> start = time.time(); copy2(parameters, 1000000); time.time() - start
0.18173909187316895
>>> start = time.time(); copy3(parameters, 1000000); time.time() - start
2.2748003005981445
>>> start = time.time(); copy4(parameters, 1000000); time.time() - start
3.071228504180908
```

I don't think the unlikely possibility of weird types justifies the expense. Or wait, maybe

```python
>>> def copy5(what, how_many):
...     for _ in range(how_many):
...         if all(isinstance(x, (str, bool, numbers.Integral, numbers.Real)) for x in what.values()):
...             tmp = what.copy()
...         else:
...             tmp = copy.deepcopy(what)
... 
>>> start = time.time(); copy5(parameters, 1000000); time.time() - start
0.5682976245880127
```

is only a factor-of-10 cost, which goes into the expensive branch on weird cases (none of ours).

agoose77: Oops, I misread this as `copy.deepcopy`.

jpivarski:
(Working through the web editor...)

PR #1651, opened 2022-08-30
feat: Drop `ak.behavior['.', 'Name'] = cls`, which isn't working/isn't tested.

jpivarski:
This feature had only one test in v1 and was untested in v2. It was also implemented incorrectly (`layout.parameter("__record__")` where `layout` is a list wouldn't see the record's parameter; you'd need a variant of `purelist_parameter` that only goes one level deep).

But it seems like this is unwanted: I've never heard of anybody using this feature, not Vector and not Coffea, which are the two heaviest users of behavioral overrides. (Early on, I was thinking of regular expressions: `.` for a single character and `*` for a sequence. However, we now have a lot more experience with how these things are used in the wild.)

I'm making this its own pull request so that it will be highly visible, and we can consider reverting it, though we'd have to do more than that: we'd have to make it not-broken.

One thing that should go with this: references to

```python
ak.behavior[".", "Name"] = Class
```

need to be removed from any forward-looking (v2) documentation.

### DIFF
--- a/src/awkward/_v2/_util.py
+++ b/src/awkward/_v2/_util.py
@@ -483,11 +483,6 @@ def arrayclass(layout, behavior):
         cls = behavior[arr]
         if isinstance(cls, type) and issubclass(cls, ak._v2.highlevel.Array):
             return cls
-    rec = layout.parameter("__record__")
-    if isstr(rec):
-        cls = behavior[".", rec]
-        if isinstance(cls, type) and issubclass(cls, ak._v2.highlevel.Array):
-            return cls
     deeprec = layout.purelist_parameter("__record__")
     if isstr(deeprec):
         cls = behavior["*", deeprec]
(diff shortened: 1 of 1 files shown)


agoose77:
We spoke about this in general before, with an eye towards extending the syntax to permit users to specify dimension-aware behaviours. I wanted this at one point to better enforce structural constraints for complex behaviour classes. 

I `rg`d my repos, and found `ak.behavior[(".", "ordered_map")] = OrderedMapArray`, which is an implementation of a key-value abstraction. So, I can attest to at least one usage ;)

jpivarski:
I remember that. If dropping the `"."` case makes it easier to extend the syntax, then that's an additional benefit. What we can't change, however, is the `"*"` case, since a lot of people know about it and use it.

> So, I can attest to at least one usage ;)

In that case, do you need it? (Can your ordered map function without it?) We can reinstate this feature. As I said in Slack, it will be easier to remove it and introduce it correctly than to leave it as-is.

agoose77:
I found the relevant issue: https://github.com/scikit-hep/awkward/issues/1038

> In that case, do you need it? (Can your ordered map function without it?) 

Not urgently - I just lose some safety. I'm not convinced it's enough to arrest this PR, though :) I'm not being contrary, though it looks like it, just setting out my own usage before we merge this :)

jpivarski:
I'm going to merge this with the understanding that reintroducing the `"."` syntax is definitely on the table. But it's something we'll proactively do, rather than let that wrong code sit there untested.

Issue #1626, opened 2022-08-24
Drop support for Python 3.6 and split git into `main` and `main-v1`


jpivarski:
Timeline:

   * [x] Make a final rc release of 1.9.0rcX with support for Python 3.6.
   * [x] Release 1.9.0.
   * [x] Drop Python 3.6 using #1661.
   * [x] Release 1.10.0 _without_ support for Python 3.6, maybe with an rc1 release just to make sure it's going to work.
   * [x] Split git into `main` (v2 series) and `main-v1` (v1 backports).
   * [x] Set VERSION_INFO to "2.0.0rc1" in `main`.
   * [x] Start dismantling v1 from `main`.

I'm hoping that `main-v1`/1.10.x won't need much maintenance. The existence of a git branch is just to make that less painful than it would otherwise be. Note that 1.10.x will always support Pythons 3.7 through 3.11 (only!) because that's what's set up in the `main-v1` CI. Hopefully, we won't have to make bug-fixes at a time in the future when Python 3.7 is hard to deploy.

**Note:** I edited the above to reflect the modifications @henryiii suggested.

jpivarski:
@henryiii's suggestion: 1.9.0 with Python 3.6 and then 1.10.x as the bug-fix series for v1.

And then not need to keep any rc releases.

agoose77:
I think we're done here!

PR #1661, opened 2022-09-01
chore: drop 3.6 support

agoose77:
- Replace text references to 3.6 with 3.7 where appropriate
- Remove CI 3.6 special-cases
- Replace manylinux1 with manylinux2014

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,8 +61,8 @@ jobs:
     - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_ARCHS_MACOS: universal2
-        CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
-        CIBW_BUILD: cp39-win_amd64 cp36-manylinux_x86_64 cp38-macosx_universal2
+        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2010
+        CIBW_BUILD: cp39-win_amd64 cp37-manylinux_x86_64 cp38-macosx_universal2
 
     - uses: pypa/cibuildwheel@v2.9.0
       if: matrix.os == 'ubuntu-latest'
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,7 +61,6 @@ jobs:
     - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_ARCHS_MACOS: universal2
-        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_BUILD: cp39-win_amd64 cp37-manylinux_x86_64 cp38-macosx_universal2
 
     - uses: pypa/cibuildwheel@v2.9.0
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -61,7 +61,7 @@ jobs:
     - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_ARCHS_MACOS: universal2
-        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2010
+        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
         CIBW_BUILD: cp39-win_amd64 cp37-manylinux_x86_64 cp38-macosx_universal2
 
     - uses: pypa/cibuildwheel@v2.9.0
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -34,12 +34,6 @@ jobs:
         build: ["cp*", "pp*"]
 
         include:
-        - os: ubuntu-latest
-          type: ManyLinux1
-          arch: auto64
-          build: "cp{36,37,38}-*"
-          CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
-
         - os: macos-latest
           type: "Universal"
           arch: universal2
@@ -51,7 +45,7 @@ jobs:
 
         - os: windows-latest
           arch: auto32
-          build: "cp{36,37,38,39}-*"
+          build: "cp{37,38,39}-*"
 
 
     steps:
@@ -76,7 +70,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python: [36, 37, 38, 39, 310, 311]
+        python: [37, 38, 39, 310, 311]
         arch: [aarch64]
     steps:
 
--- a/.github/workflows/wheels.yml
+++ b/.github/workflows/wheels.yml
@@ -57,6 +57,7 @@ jobs:
       env:
         CIBW_BUILD: ${{ matrix.build }}
         CIBW_ARCHS: ${{ matrix.arch }}
+        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
 
     - name: Upload wheels
       uses: actions/upload-artifact@v3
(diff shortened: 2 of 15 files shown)


agoose77:
@jpivarski I've taken the liberty of removing manylinux1 support at the same time, as [we only need it for 3.6 as far as I can tell](https://labs.quansight.org/blog/2020/11/manylinux1-is-obsolete-manylinux2010-is-almost-eol-what-is-next). However, I noticed that our CI was building manylinux1 wheels for 3.6, 3.7, and 3.8. Do you know why that is? It makes me hesitant to remove it.

jpivarski:
> @jpivarski I've taken the liberty of removing manylinux1 support at the same time, as [we only need it for 3.6 as far as I can tell](https://labs.quansight.org/blog/2020/11/manylinux1-is-obsolete-manylinux2010-is-almost-eol-what-is-next). However, I noticed that our CI was building manylinux1 wheels for 3.6, 3.7, and 3.8. Do you know why that is? It makes me hesitant to remove it.

We should probably get a sign-off from @henryiii before merging this (between 1.9.0 and 1.10.0, following https://github.com/scikit-hep/awkward/issues/1626#issuecomment-1227458731). The time it takes to release 1.9.0rc13 and 1.9.0 (two releases == 2 days) might correspond to the time when he's back and ready to review things.

agoose77:
@jpivarski I want to (at some point) rename the CI workflows (both filename and display names):
- `test-wheel-construction.yml` (`Wheels`)
- `test-compilation.yml` (`Build`)
- `deploy-wheels` (`Deploy`)

Maybe we want to move the linting into a new workflow?
- `test-linters.yml` (`Lint`)

E.g.
image


Is this PR as good a time as any to do this?
In any case, we already need to fiddle with the required CI checks due to this PR (incidentally, I think my changes to the CI need to be further considered - the non-3.6 Linux jobs no longer match those in the expected checks)

jpivarski:
Yes, this is as good a time as any. By doing it before the git split (`main` and `main-v1`), the names of these tests can be the same in both branches without having to manually make both branches agree.

jpivarski:
It reminds me that I'll have to make two branch protection rules. (I've already done that for Uproot; just forgot about it.) In principle, the test names can differ because the two branches can have different rules, but it will be a lot easier to keep track of if they're not different. (Otherwise, we'd be saying, "main test X corresponds to main-v1 test Y...")

henryiii:
There are a very small number of users on 3.7 & 3.8 that might have a default pip that can only access manylinux1. They would already be likely to be having some issues, and more in the future. They are also using a version of pip that does not officially support Python 3.7 and 3.8 (it predated those versions) but is forced onto those Pythons (and possibly patched) by linux distro managers.

We should also be bumping pyupgrade to 3.7+, and enabling `from __future__ import annotations` injection by isort. Static typing is much easer in Python 3.7! (wink, wink)

.github/workflows/tests.yml
```diff
@@ -61,8 +61,8 @@ jobs:
     - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_ARCHS_MACOS: universal2
-        CIBW_MANYLINUX_X86_64_IMAGE: manylinux1
-        CIBW_BUILD: cp39-win_amd64 cp36-manylinux_x86_64 cp38-macosx_universal2
+        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
```
henryiii: This can go away now, it can (and is, I think) be set in pyproject.toml.

.github/workflows/wheels.yml
```diff
@@ -62,8 +56,8 @@ jobs:
     - uses: pypa/cibuildwheel@v2.9.0
       env:
         CIBW_BUILD: ${{ matrix.build }}
-        CIBW_MANYLINUX_X86_64_IMAGE: ${{ matrix.CIBW_MANYLINUX_X86_64_IMAGE }}
         CIBW_ARCHS: ${{ matrix.arch }}
+        CIBW_MANYLINUX_X86_64_IMAGE: manylinux2014
```
henryiii: Same.

CONTRIBUTING.md
```diff
@@ -267,11 +267,11 @@ Templating is only used for integer specialization.
...
+We target Python 3.7 and above. Import statements can assume Python 3 names, string-checking can assume Python 3 meanings of `str` and `bytes`, Unicode literals don't need to be prefixed by `u`, and dict order can be assumed to be stable. Python's f-strings can now be used, but not with equals signs (e.g. `f"{something = }"` rather than `f"something = {something}"`) because that's a Python 3.8 feature (its main use is in debugging, anyway).
 
-If you see any outdated (pre-Python 3.6) code, you can safely clean them up. Some strings are not easier to read as f-strings or require some work to make them more readable; it should be a case-by-case basis.
+If you see any outdated (pre-Python 3.7) code, you can safely clean them up. Some strings are not easier to read as f-strings or require some work to make them more readable; it should be a case-by-case basis.
 
-Python 3.6 will be the standard until it becomes "painful" to keep it (like, when CI and build tools no longer support it, or if we adopt type annotations).
+Python 3.7 will be the standard until it becomes "painful" to keep it (like, when CI and build tools no longer support it, or if we adopt type annotations).
```
henryiii: Is this valid, or are we beginning to follow [official CPython EoL](https://endoflife.date/python)? (Assuming not NEP 29 / SPEC 1, as that's 3.8+ currently 😉)

README-pypi.md
```diff
@@ -2,7 +2,7 @@
 
 [![PyPI version](https://badge.fury.io/py/awkward.svg)](https://pypi.org/project/awkward)
 [![Conda-Forge](https://img.shields.io/conda/vn/conda-forge/awkward)](https://github.com/conda-forge/awkward-feedstock)
-[![Python 3.6‒3.10](https://img.shields.io/badge/python-3.6%E2%80%923.10-blue)](https://www.python.org)
+[![Python 3.7‒3.10](https://img.shields.io/badge/python-3.7%E2%80%923.10-blue)](https://www.python.org)
```
henryiii: 3.7-3.11?

jpivarski: You're right that it should go up to 3.11, thanks to the tests that you added.

CONTRIBUTING.md
jpivarski: Yeah, let's go with the CPython EoL dates. I am changing my mind from when I wrote that (from "maximum comfortable support" to "let's just follow a schedule"). I didn't have a particular schedule in mind, but the EoL dates sounds like a very defensible policy.

.github/workflows/tests.yml
agoose77: Oh yes, I forgot that we use `pyproject.toml` as a configuration source even in the CI.

CONTRIBUTING.md
agoose77: Is the new wording appropriate? I.e., are we proposing to drop Python versions alongside CPython's EoL schedule?

agoose77: > We should also be bumping pyupgrade to 3.7+, and enabling `from __future__ import annotations` injection by isort. Static typing is much easer in Python 3.7! (wink, wink)

I'm stoked about this!

jpivarski:
I'm going to make a 1.10.0rc1 out of this. The only changes I made were to CONTRIBUTING.md, so I'm not going to wait for the tests to finish.

If we need to make any more fixes, it will be 1.10.0rc2.

Thanks!!!

PR #1665, opened 2022-09-05
docs: update docstring of `ak.with_name` to reflect `None` interpretation

agoose77:
This PR adds to the documentation for `ak.with_name` to clarify that, if `name` is set to `None`, then the existing name of the record is removed.

### DIFF
--- a/src/awkward/operations/ak_with_name.py
+++ b/src/awkward/operations/ak_with_name.py
@@ -9,8 +9,8 @@ def with_name(array, name, highlevel=True, behavior=None):
     """
     Args:
         base: Data containing records or tuples.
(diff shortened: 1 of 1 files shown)


agoose77:
Ping @jpivarski, this is a small PR and wonder if it escaped your notice!

jpivarski (approved):
This is true of any parameter, though it's worthwhile emphasizing it in the docs.

The same would be true of `ak.with_parameter`. Does that documentation call it out like this one?

agoose77:
Yes, although not in the type hint. I was going to leave that, as we'll slowly remove the type information from the docstring itself as we move towards type hints.

PR #1667, opened 2022-09-05
docs: fix typos

kianmeng:
Found via `codespell -L fo,ue,nwo,subjet,unio **/*.{md,py,rst,ipynb}`

### DIFF
--- a/docs-jupyter/2020-01-22-numba-demo-EVALUATED.ipynb
+++ b/docs-jupyter/2020-01-22-numba-demo-EVALUATED.ipynb
@@ -545,7 +545,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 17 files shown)


henryiii:
FWIW, these were missed due to https://github.com/scikit-hep/awkward/blob/1bcfd70629aec4c1ef26bbf929d9183c28dbe491/.pre-commit-config.yaml#L5 - docs and these other folders are ignored by pre-commit. Why is docs* ignored???

kianmeng:
@henryiii I see. Didn't realize this repo was using pre-commit with codespell task. Closing this PR. Sorry about that.

henryiii:
I think these fixes are fine, I just want to know why they are being skipped via pre-commit. Docs should be checked, at least - maybe they are really changing or going away soon?

jpivarski:
It could have been that we didn't want to run all the tests (black, flake8) on Python code in the docs directories, such as the Sphinx configuration file. It's also the case, though, that @agoose77 is replacing a lot of that.

The spelling corrections are good; we should be sure that they're applied to the new docs.

(With the exception of "vertexes" vs "vertices", which is a matter of opinion and I prefer the "es" pluralization because it's easier to match with a single "vertex". Similarly for "indexes" vs "indices". But if codespell has the opposite opinion, I'm not going to try to fight it.)

kianmeng:
@henryiii Should I reopen this and update the pre-commit config?

PR #1671, opened 2022-09-05
fix: allow string arrays to merge without unions

### LINKED ISSUE #1646: Concatenating `RegularArray` and 2D `NumpyArray` yields a union.
### Version of Awkward Array

main

An unnecessary `UnionType` is created when the arrays have different layouts (but the same type). Especially with `merge=True`, I'd expect `ak.concatenate` to resolve these arrays to a single `RegularType`. 

```python
>>> import awkward._v2 as ak
>>> import numpy as np 
>>> a0 = ak.from_numpy(np.empty((5, 0)), regulararray=True)
>>> a1 = ak.from_numpy(np.empty((8, 0)), regulararray=False)
>>> ak.concatenate([a0, a1]).type
ArrayType(UnionType([RegularType(NumpyType('float64'), 0), RegularType(NumpyType('float64'), 0)]), 13)
```

Submitted on behalf of @grst

### LINKED ISSUE #1669: Type-unification of "string" and "option[unknown]" should return "option[string]", not a union.
Just trying this out now, and running into an issue. I can get the expected result from:

```python
awk_a = ak.Array([
    {"a": [1, 2, 3], "b": [1, 2]},
    {"a": [4, 5], "b": [3, 4]},
    {"a": [6], "b": [5]},
])


awk_b = ak.Array([
    {"a": [1, 2, 3]},
    {"a": [4, 5]},
    {"a": [6]},
])

ak.concatenate([awk_a, ak.with_field(awk_b, None, "b")]).type.show()
```

```
6 * {
    a: var * int64,
    b: option[var * int64]
}
```

However, I get a `union` from:

```python
awk_c = ak.Array([
    [{"a": 1, "b": "foo"}],
    [{"a": 2, "b": "bar"}, {"a": 3, "b": "baz"}]
])

awk_d = ak.Array([
    [{"a": 4}, {"a": 5}],
    [{"a": 6}],
    [{"a": 7}],
])

ak.concatenate([awk_c, ak.with_field(awk_d, None, "b")]).type.show()
```

```
5 * union[
    var * {
        a: int64,
        b: string
    },
    var * {
        a: int64,
        b: ?unknown
    }
]
```

Is this expected?

_Originally posted by @ivirshup in https://github.com/scikit-hep/awkward/discussions/1647#discussioncomment-3561037_

I think you found a bug: we want "string" and "option[unknown]" to unify to "option[string]", and it's not recognizing that. It should first notice that one of the two is "option" and make the result type "option", then unify "string" (or "X") with "unknown" to get "string" ("X"), since "unknown" is an identity in type-unification. I _think_ this is supposed to happen somewhere in [simplify_uniontype](https://github.com/scikit-hep/awkward/blob/1bcfd70629aec4c1ef26bbf929d9183c28dbe491/src/awkward/_v2/contents/unionarray.py#L471-L661). It's interesting that it got it right when "X" is a primitive type, int64, but not when it's a list type. This is something we should promote into an issue.

agoose77:
Fixes #1669, fixes #1646

## Changes
- Refactor `Content.mergeable` into `Content.mergeable` and `Content._mergeable`
- Add `Form.is_identity_like` to determine whether a layout should behave like an identity. 
   - If the content is an `EmptyArray`, return `True`.
   - If the content is an `IndexedOptionArray`, return `self._content.is_identity_like`.
- Add cases to convert >1D `NumpyArray`s to `RegularArray` during merging


agoose77:
@jpivarski along the way I noticed an apparent bug in `Form._parameters_equal`. Could you clarify what should happen in `ArrayType.__eq__` for parameters? How about the other reserved parameter names like `__array__` and `__record__`? My assumption is (based on v1)
- `__array__` - important to `__eq__`
- `__record__` - important to `__eq__`
- other - not important to `__eq__`

Additionally, I found that in v1 we check only the `__array__` and `__record__` parameters for contents within `mergeable`, but this is not the case for v2. Could you confirm that we should only be checking this restricted subset in v2 also?

jpivarski:
I think you want to use `only_array_record=True` in `Content.mergeable`, but not in `Form.__eq__`. To say that two Forms are unequal is to say that the array metadata differs in some not-necessarily-visible way, so even if a parameter has no consequence, a differing value should mean the Forms are different.

But you're absolutely right about `only_array_record=True` in `Content.mergeable`; [it is what v1 did](https://github.com/scikit-hep/awkward/blob/1bcfd70629aec4c1ef26bbf929d9183c28dbe491/src/libawkward/util.cpp#L630-L670). And that was on purpose, so concatenating an array with a `__doc__` and another without that parameter wouldn't become a union.

You're right about there being a bug in `ArrayType.__eq__`, but it should be

```python
    def __eq__(self, other):
        if isinstance(other, ArrayType):
            return (
                self._length == other._length
                and self._content == other._content
            )
        else:
            return False
```

The `ArrayType` is not a subclass of `Type`; it doesn't have `parameters`. But it should start a recursion into checking the equality of `content`.

agoose77:
> I think you want to use `only_array_record=True` in `Content.mergeable`, but not in `Form.__eq__`. To say that two Forms are unequal is to say that the array metadata differs in some not-necessarily-visible way, so even if a parameter has no consequence, a differing value should mean the Forms are different.

You're absolutely right. I tunnel visioned there.

agoose77:
> I think you want to use only_array_record=True in Content.mergeable, but not in Form.__eq__. To say that two Forms are unequal is to say that the array metadata differs in some not-necessarily-visible way, so even if a parameter has no consequence, a differing value should mean the Forms are different.

Ah, no - I meant that the implementation for `_parameters_equal` seemed to have a backwards case at the end of the function. This was separate to the ArrayType handling, but I must have deleted part of the comment that elucidated this!


Otherwise, yes; if I zoom out and look at the bigger picture, there are some additional bugs to fix!

jpivarski:
> Ah, no - I meant that the implementation for `_parameters_equal` seemed to have a backwards case at the end of the function. This was separate to the ArrayType handling, but I must have deleted part of the comment that elucidated this!

You're right: this is backward and should be fixed.

https://github.com/scikit-hep/awkward/blob/1bcfd70629aec4c1ef26bbf929d9183c28dbe491/src/awkward/_v2/forms/form.py#L199-L202

jpivarski:
On the subject of type equality and strings, is the type of an array of categorical strings equal to the type of an array of strings? They're supposed to be distinct.

In Forms, the categorical strings would have `__array__ = "categorical"` on the IndexedArray. However, IndexForm nodes have no equivalent in Types, so the `__array__ = "categorical"` gets turned into `__categorical__ = True` on the next level down (the Type of the strings themselves). In the _spirit_ of this "only check `__array__` and `__record__`" logic, we should identify categorical strings and strings as distinct types. However, because the indicator is not one of `__array__` and `__record__`, I think this PR would make them say they're equal types.

agoose77:
> On the subject of type equality and strings, is the type of an array of categorical strings equal to the type of an array of strings? They're supposed to be distinct.

~~Not any more (this PR fixes the type equality checks).~~

Actually, our current v2 did this correctly too, because type-strings still contain sufficient information. Only v1 didn't do this properly.

agoose77:
I've introduced a new policy for merging `IndexedOptionArray` over `EmtpyArray`. This is what the new `is_identity_like` flag is for in the Form system. In this PR so far, if `is_identity_like` is True, then we don't require parameters to match, and just merge the layouts. 

This seems restricted enough that it might be an OK exception to the rule that parameters need to match in order to be mergeable. Are you happy with this @jpivarski? This behaviour is fundamentally what's going wrong in #1669, so the answer here has a direct bearing on what we do there. The material consequences of this policy would be that user-defined arrays with `IndexedOptionArray` over `EmptyArray` would not be considered distinct types (unmergeable) from other arrays with specified behaviors (e.g. `__array__)`.

Perhaps `is_identity_like` should be restricted, and require that neither the option-type nor the `EmptyArray` have parameters. This would constrain the behaviour of `EmptyArray` in comparison to v1, which doesn't care whether the `EmptyArray` has parameters.

agoose77:
This PR breaks a few tests because previously our type handling wasn't correct. In 1bcfd70 the following code produces a union:
```python
import awkward as ak
import numpy as np


a1 = ak._v2.Array(np.array([[0.0, 1.1], [2.2, 3.3]]))
a2 = ak._v2.from_json("[[4.4, 5.5], [6.6, 7.7], [8.8, 9.9]]")
assert isinstance(a1.layout, ak._v2.contents.NumpyArray)
a2 = ak._v2.to_regular(a2, axis=1)

c = ak._v2.concatenate([a1, a2])
c.type.show()
```

and the same happens in this PR. However, in 1bcfd70 the `ArrayType.__eq__` returned True for this type against a non-union, which is not correct.

Ultimately this is #1647

agoose77:
I've added a fix for #1646 by adding logic for >1D `NumpyArray` to promote themselves to `RegularArray`, and the same in reverse (for `NumpyArray` arguments). I'm not sure if this is a brilliant way to solve this, so I'd appreciate any feedback.

src/awkward/_v2/contents/recordarray.py
```diff
@@ -585,10 +572,9 @@ def mergemany(self, others):
         for_each_field = []
         for field in self.contents:
             trimmed = field[0 : self.length]
-            for_each_field.append([field])
+            for_each_field.append([trimmed])
```
jpivarski, line 575: So we made the `trimmed` and simply forgot to pass that on instead of `field`?

Good catch!

src/awkward/_v2/contents/regulararray.py
```diff
@@ -694,6 +681,14 @@ def mergeable(self, other, mergebool):
...
             return self._content.mergeable(other.content, mergebool)
 
+        # For n-dimensional NumpyArrays, let's now convert them to RegularArray
+        # We could add a special case that tries to first convert self to NumpyArray
+        # and merge conventionally, but it's not worth it at this stage.
+        elif (
+            isinstance(other, ak._v2.contents.numpyarray.NumpyArray)
+            and other.purelist_depth > 1
+        ):
+            return self._content.mergeable(other.toRegularArray().content, mergebool)
```
jpivarski, line 691: This `toRegularArray()` is going to happen twice for every merger. It might be possible to do this at an earlier stage of processing and therefore only do it once.

But not for this PR.

tests/v2/test_0449-merge-many-arrays-in-one-pass.py
```diff
@@ -3,108 +3,84 @@
...
-                                ak._v2.contents.NumpyArray(two),
-                                ak._v2.contents.EmptyArray(),
-                                ak._v2.contents.NumpyArray(four),
-                            ]
-                        )
-                        .form
-                    )
+    for dtype1, dtype2, dtype3, dtype4 in itertools.product(
+        ("i1", "i2", "i4", "i8", "u1", "u2", "u4", "u8", "f4", "f8", "?"), repeat=4
+    ):
```
jpivarski, line 14: Fine. It's hard for me to check because the un-indent is a diff: is the content of this loop unchanged?

src/awkward/_v2/contents/numpyarray.py
```diff
@@ -405,7 +392,7 @@ def mergeable(self, other, mergebool):
         ):
             return self.mergeable(other._content, mergebool)
 
-        if isinstance(other, ak._v2.contents.numpyarray.NumpyArray):
+        elif isinstance(other, ak._v2.contents.NumpyArray):
```
jpivarski, line 395: I agree that this should be chained with the if-elif above, since both of those return values. The code flow is the same either way, but we should adopt a default, and I prefer the if-elif-elif default.

jpivarski (approved):
Moving the (old) parameters check and (new) identity-like check into `Content` is a good idea.

I lost my original review, but there's something else you're adding that is a good idea, too. (Can't remember what it was.)

I don't know that we're on the same page about this:

>> On the subject of type equality and strings, is the type of an array of categorical strings equal to the type of an array of strings? They're supposed to be distinct.
> 
> Not any more (this PR fixes the type equality checks).
> 
> Actually, our current v2 did this correctly too, because type-strings still contain sufficient information. Only v1 didn't do this properly.

An array of strings and an array of categorical strings are supposed to have different types. Toward this end, the Datashape strings are different: they look different to users and should be considered different.

It seems that the string representation isn't set up correctly in v2 yet:

```python
>>> import awkward._v2 as ak
>>> array1 = ak.Array(["one", "two", "one", "one"])
>>> array2 = ak.behaviors.categorical.to_categorical(array1)
>>> print(array1.type)
4 * string
>>> print(array2.type)
4 * string
```

but it was in v1:

```python
>>> import awkward as ak
>>> array1 = ak.Array(["one", "two", "one", "one"])
>>> array2 = ak.behaviors.categorical.to_categorical(array1)
>>> print(array1.type)
4 * string
>>> print(array2.type)
4 * categorical[type=string]
```

(But maybe my laptop's Awkward version is old, 1.9.0rc4, and I can't easily update or try this PR out personally right now.)

Can you confirm some of my questions? If we're in agreement, then I approve this and you can merge it. Thanks!

src/awkward/_v2/contents/recordarray.py
agoose77: Yes. I'm assuming that `trimmed` was in there for good reason, so I've modified the succeeding line.

tests/v2/test_0449-merge-many-arrays-in-one-pass.py
agoose77: Yes, it just moves the four nested loops into a single product loop.

src/awkward/_v2/contents/numpyarray.py
agoose77: Yes, I'm following the Awkward code style of minimal early returns :)

src/awkward/_v2/contents/regulararray.py
agoose77: Yes, this is not ideal at all. I suspect we'd replace `mergeable` interface with `prepare_merge(self, other)`, and if the two are mergeable, return potentially modified layouts.

I think that's a bigger change than we want to make in this PR, which is already "big".

agoose77:
I've confirmed that the typestrings were broken in main, and this PR fixes it.

I want to take a moment to shout out to @ioanaif for how much work you've put into the v1-v2 port. I keep digging into v1 code to compare things and I realise just how much work it must have taken! :ninja:

agoose77:
Looks like our existing type parsing logic is giving the test suite some problems. Let's see if we can fix it.

jpivarski:
Actually, this test is failing because the test was defined incorrectly, and now the code itself is right. Let me make a "suggest change" for it.

agoose77:
@jpivarski good timing! Now I'm stuck. I'm not familiar enough with the type-parser logic to understand what should be happening, in order to fix it.

The issue is that this PR makes type equality checking recursive, rather than just checking type strings. I feel that this is a good change, because we had a number of hidden bugs that we didn't catch by checking `.type` (e.g. https://github.com/scikit-hep/awkward/pull/1671#issuecomment-1237934521) 

The `repr`d type object that I'm currently looking at, from `tests/v2/test_0773-typeparser.py `
```python
ArrayType(
    ListType(
        NumpyType("uint8", parameters={"__array__": "char"}, typestr="char"),
        parameters={"__categorical__": True},
        typestr="string",
    ),
    6,
)
```

This doesn't survive a round-trip:
```python
ArrayType(
    ListType(
        NumpyType("uint8", parameters={"__array__": "char"}),
        parameters={"__array__": "string", "__categorical__": True},
    ),
    6,
)
```

The actual bug right now I think is two-fold:
- the to-type logic replaces `__array__ = "categorical"` with `__categorical__ = True` for `IndexedForm`, but we don't do the reverse.
- we don't set `typestr` in the parser

However, I'm not familiar enough with using typestring overloads to understand what we want to happen here. It makes me wonder whether checking the generated typestring _is_ intended, and we need to robust-ify our other tests that were assuming types were more concrete representations of layout than they are.

PR #1674, opened 2022-09-05
refactor: move categorical functions to operations

### LINKED ISSUE #1670: Expose `to_categorical` and `from_categorical` functions at top-level
### Version of Awkward Array

HEAD

Mentioned in https://github.com/scverse/anndata/pull/647#discussion_r963085500.

We need `ak._v2.to_categorical` and `ak._v2.from_categorical` to point to the two functions in `awkward.behaviors.categorical`. The `behaviors` and `operations` submodules are parallel. Follow the example of `src/awkward/_v2/operations.__init__.py`.

agoose77:
Fixes #1670

I opted to just move these functions. Is this too much of a breaking change in your view @jpivarski? I'm not firm on what API guarantees we've made w.r.t imports.

### DIFF
--- a/src/awkward/_v2/behaviors/categorical.py
+++ b/src/awkward/_v2/behaviors/categorical.py
@@ -14,7 +14,7 @@ class CategoricalBehavior(Array):
 class _HashableDict:
     def __init__(self, obj):
         self.keys = tuple(sorted(obj))
-        self.values = tuple(_hashable(obj[k]) for k in self.keys)
+        self.values = tuple(as_hashable(obj[k]) for k in self.keys)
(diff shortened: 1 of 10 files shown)


jpivarski (approved):
I wouldn't call it a breaking change (since we only advertised the top-level module versions: `ak.*`). In fact, it's a better idea: consolidating the number of different kinds of things.

This can be merged when it has passed checks. Oh, I recently advertised the fully qualified name (which has now changed); I'll fix that.

agoose77:
All tests passed. The CI doesn't know this, because it's expecting tests with a different name.

PR #1675, opened 2022-09-06
chore: remove `dev/build-awkward.sh`

agoose77:


agoose77:
Per the Slack channel, Jim's signed off on removing this :)

PR #1676, opened 2022-09-06
refactor: prepare to deprecate camelCase `toRegularArray` et al.

agoose77:
This PR introduces snake_case methods for `toRegularArray` et al. This means in future we can introduce a PR to deprecate these.


jpivarski:
The `to` → `to_` is fine, but the `RegularArray` names the type it's converting it into. Wouldn't the function of `to_RegularArray` be clearer than `to_regular_array`?

agoose77:
> The `to` → `to_` is fine, but the `RegularArray` names the type it's converting it into. Wouldn't the function of `to_RegularArray` be clearer than `to_regular_array`?

Personally, I don't like mixed cases, which motivated this PR in the first place. I don't find the name unclear, after all we have `to_arrow` instead of `to_Table` (or whichever type we convert to).

jpivarski:
Okay, sure.  `:)`

src/awkward/_v2/contents/listarray.py
```diff
@@ -1506,3 +1506,9 @@ def _layout_equal(self, other, index_dtype=True, numpyarray=True):
             and self.stops.layout_equal(other.stops, index_dtype, numpyarray)
             and self.content.layout_equal(other.content, index_dtype, numpyarray)
         )
+
+    def toListOffsetArray64(self):
+        return self.to_list_offset_array()
+
+    def toRegularArray(self):
+        return self.to_regular_array()
```
jpivarski, line 1514: Why is this here? Is this left over from trying to apply this change in stages?

src/awkward/_v2/contents/listoffsetarray.py
```diff
@@ -2218,3 +2218,9 @@ def _layout_equal(self, other, index_dtype=True, numpyarray=True):
         return self.offsets.layout_equal(
             other.offsets, index_dtype, numpyarray
         ) and self.content.layout_equal(other.content, index_dtype, numpyarray)
+
+    def toListOffsetArray64(self):
+        return self.to_list_offset_array()
+
+    def toRegularArray(self):
+        return self.to_regular_array()
```
jpivarski, line 2226: Same question as above.

src/awkward/_v2/contents/numpyarray.py
```diff
@@ -1382,3 +1382,6 @@ def _layout_equal(self, other, index_dtype=True, numpyarray=True):
             )
         else:
             return True
+
+    def toRegularArray(self):
+        return self.to_regular_array()
```
jpivarski, line 1387: Same question as above.

src/awkward/_v2/contents/regulararray.py
```diff
@@ -1326,3 +1326,9 @@ def _layout_equal(self, other, index_dtype=True, numpyarray=True):
         return self.size == other.size and self.content.layout_equal(
             self.content, index_dtype, numpyarray
         )
+
+    def toListOffsetArray64(self):
+        return self.to_list_offset_array()
+
+    def toRegularArray(self):
+        return self.to_regular_array()
```
jpivarski, line 1334: Same question as above.

jpivarski (changes requested):
I'm fine with this change, though I would have preferred `to_RegularArray` rather than `to_regular_array`, since the method converts the object to a `RegularArray` (camel-case because it's naming a class). But I also don't feel strongly about it.

I do have a question: it looks like the old names still exist as synonyms for the new names. The line numbers where this happens have been pointed out below.

Also, I don't think you got all of the option-type conversion methods:

- `toBitMaskedArray`
- `toByteMaskedArray`
- `toIndexedOptionArray`

Because of these questions, I'm going to hold off on this at least until after 1.10.0rc3.

src/awkward/_v2/contents/listarray.py
agoose77: I've left this here because I don't know what kind of deprecation period we want for internal APIs. If you're happy to make the switch then these can all go.

agoose77:
> Also, I don't think you got all of the option-type conversion methods:
>
>- `toBitMaskedArray`
> - `toByteMaskedArray`
> - `toIndexedOptionArray`
> Because of these questions, I'm going to hold off on this at least until after 1.10.0rc3.

Ah, that's what I was trying to remember during the meeting. I meant to add these as todo-items following discussion :face_in_clouds:

jpivarski:
No deprecation period is needed before the 2.0.0 release.

agoose77:
Oh, right - we're not guaranteeing that the layout API is even remotely backwards compatible with v1. I keep forgetting this fact, because I think of layouts as public API (albeit a low level one)

jpivarski:
That's right, the layouts are a low-level, public API, and we've advertised that we will be changing these slightly. Slightly means some methods/properties will have different names, which is what this PR is doing. It's the high-level, public API that's not changing (much? at all?).

One of the earlier changes we put into v2 was to change low-level, public API method names so that they would agree with their corresponding high-level, public API names. So this sort of thing is fair game.

agoose77:
This will need rebasing, so I'll close this for now.

Issue #1438, opened 2022-04-21
ArrayBuilder in Numba is missing some methods, such as 'string'


jpivarski:
I'm astonished: it's weird that I would just forget that. But it does explain why `append` isn't overloaded to cover it. Probably the feature was added later and the ArrayBuilder-in-Numba implementation is just behind the times?

```python
>>> import numba as nb
>>> import awkward as ak
>>> @nb.njit
... def add_a_string(builder, string):
...     builder.string(string)
...     return builder
... 
>>> builder = add_a_string(ak.ArrayBuilder(), "hello")
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jpivarski/mambaforge/lib/python3.9/site-packages/numba/core/dispatcher.py", line 468, in _compile_for_args
    error_rewrite(e, 'typing')
  File "/home/jpivarski/mambaforge/lib/python3.9/site-packages/numba/core/dispatcher.py", line 409, in error_rewrite
    raise e.with_traceback(None)
numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
Unknown attribute 'string' of type ak.ArrayBuilderType(None)

File "<stdin>", line 3:
<source missing, REPL/exec in use?>

During: typing of get attribute at <stdin> (3)

File "<stdin>", line 3:
<source missing, REPL/exec in use?>
```

Indeed: it's missing. Also, I don't see it in either the v1 or the v2 implementation.

This should be a separate issue: ArrayBuilder-in-Numba is missing `string` and `bytestring`, and possibly other new methods (`complex`? `datetime`?). Fortunately, we know what the interface is _supposed_ to be: it's supposed to be the same as outside of Numba, the ArrayBuilder Python interface.

(BTW: ArrayBuilder-in-Numba's lack of context managers `list`, `record`, and `tuple` is known because it's waiting on context managers as a feature from Numba, but it sounds like this will be implemented soon, and Awkward Array will be a first use-case.)

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward-1.0/issues/1420#issuecomment-1105476531_

ianna:
@jpivarski - it looks like it's implemented in v2:
```python
>>> import numba as nb
>>> import awkward._v2 as ak
>>> import awkward._v2._connect.numba.arrayview
>>> import awkward._v2._connect.numba.builder
>>> def add_a_string(builder, string):
...     builder.string(string)
...     return builder
... 
[...]

PR #1677, opened 2022-09-06
feat: Array Builder in Numba - add missing methods

ianna:
Address Issue #1438: ArrayBuilder in Numba is missing some methods, such as 'string' 

1. ArrayBuilder in Numba following methods have been added:

- string
- complex
- datetime
- timedelta

2. Test all ArrayBuilder in Numba methods

- [x] Restore CI configuration to default

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-numba-array-builder/> once Read the Docs has finished building :hammer:


src/awkward/_v2/_connect/numba/builder.py
```diff
@@ -396,6 +410,23 @@ def lower_real(context, builder, sig, args):
     return context.get_dummy_value()
 
 
+@numba.extending.lower_builtin("complex", ArrayBuilderType, numba.types.Complex)
+def lower_complex(context, builder, sig, args):
+    arraybuildertype, xtype = sig.args
+    arraybuilderval, xval = args
+    proxyin = context.make_helper(builder, arraybuildertype, arraybuilderval)
+    z_imag = xval.operands.pop()
+    z_real = xval.operands.pop().value
```
jpivarski:
```suggestion
    z = context.make_complex(builder, xtype, xval)
    z_real, z_imag = z.real, z.imag
```

I don't know if you need `z.real._getvalue()` or not...

ianna: done

src/awkward/_v2/_connect/numba/builder.py
```diff
@@ -396,6 +477,110 @@ def lower_real(context, builder, sig, args):
...
+
+    pyapi.gil_release(gil)
+
+    proxyin = context.make_helper(builder, arraybuildertype, arraybuilderval)
+    call(
+        context,
+        builder,
+        ak._libawkward.ArrayBuilder_string,
+        (proxyin.rawptr, out),
+    )
```
ianna: release gil here:
`pyapi.gil_release(gil)`

src/awkward/_v2/_connect/numba/builder.py
```diff
@@ -396,6 +477,110 @@ def lower_real(context, builder, sig, args):
...
+    arraybuilderval, xval = args
+
+    pyapi = context.get_python_api(builder)
+    gil = pyapi.gil_ensure()
+
+    strptr = pyapi.from_native_value(xtype, xval)
+    cstrptr = builder.bitcast(strptr, pyapi.cstring)
+    out = pyapi.string_as_string(cstrptr)
+
+    pyapi.gil_release(gil)
```
ianna: delete

src/awkward/_v2/_connect/numba/builder.py
ianna: done

src/awkward/_v2/_connect/numba/builder.py
ianna: done

ianna (commented):
as discussed - this is done

agoose77:
@ianna I've squashed your commits and rebased them onto main in https://github.com/scikit-hep/awkward/tree/agoose77/rebase-numba-array-builder if you're happy to force push them here. I should have preserved  you as the author of those commits :)

src/awkward/_connect/numba/builder.py
```diff
@@ -396,6 +464,159 @@ def lower_real(context, builder, sig, args):
...
+    strptr = pyapi.from_native_value(xtype, xval)
+    result = numba.core.cgutils.alloca_once(
+        builder, context.get_value_type(numba.int64)
+    )
+    p_buffer = numba.core.cgutils.alloca_once(builder, pyapi.cstring)
+
+    # FIXME: ok =
+    bytes_as_string_and_size(pyapi, strptr, p_buffer, result)
+    # FIXME: pyapi.if_object_ok
+    length = ak._connect.numba.layout.castint(
```
ianna, line 554: @jpivarski - I think, I'm casting the `result` to `int64` here, but **Windows (3.9, x86)** disagrees:
`Type of #3 arg mismatch: i32* != i64*`

src/awkward/_connect/numba/builder.py
```diff
@@ -396,6 +464,159 @@ def lower_real(context, builder, sig, args):
     return context.get_dummy_value()
 
 
+@numba.extending.lower_builtin("complex", ArrayBuilderType, numba.types.Integer)
+@numba.extending.lower_builtin("complex", ArrayBuilderType, numba.types.Float)
+@numba.extending.lower_builtin("complex", ArrayBuilderType, numba.types.Complex)
+def lower_complex(context, builder, sig, args):
```
ianna: @jpivarski - as discussed, I tried to implement two separate functions for different types - and add a note for pylint to ignore re-definition. That worked for pre-commit run locally, but the CI pylint would not take it...

ianna (commented):
@jpivarski - I have a couple of questions. Please, have a look at the comments. Thanks

src/awkward/_connect/numba/builder.py
jpivarski: (Good—that's why we have that x86 test!) Where is the cast to `int64`? It's not in the highlighted code above. You mean `p_buffer`, right? (Does the argument counting start at 1?)

Things to look for: are you casting the data in the buffer or the pointer to the buffer? Are you using `size_t` for the cast? That's what's going to differ between Windows 64 and x86.

src/awkward/_connect/numba/builder.py
jpivarski: Oh, if pylint is complaining about redefinition, the problem is that the function has the same name. A downside of Numba defining its interface in terms of decorators is that the function objects themselves end up being dummies that aren't ever referenced, so their names hardly matter. But it is wrong and pylint will complain if a function is defined twice in a module with the same names. I usually add more descriptive words between underscores, but if that's not possible, then `_1`, `_2`, etc.

ianna: Thanks!

src/awkward/_connect/numba/builder.py
ianna: done. I was looking at a wrong function :-)

ianna:
@agoose77 - do I understand correctly that some jobs get cancelled because the other architectures fail?

agoose77:
All "matrix-jobs" (e.g. py36, py37, ...) within the same top-level job, (e.g. Linux), will be canceled if a single matrix job fails, yes. :)

jpivarski:
If you need to turn off the fail-fast for diagnostic purposes, you can do that by changing the `.github/workflows` YAML in your branch—it will be fail-slow in just your branch. Just don't forget to revert it.

agoose77:
I added a task to this PR so that we don't forget to restore the CI configuration :)

src/awkward/_connect/numba/builder.py
```diff
@@ -396,6 +463,105 @@ def lower_real(context, builder, sig, args):
...
+    pyapi = context.get_python_api(builder)
+    gil = pyapi.gil_ensure()
+
+    out = pyapi.string_as_string(xval.value)
+    call(
+        context,
+        builder,
+        ak._libawkward.ArrayBuilder_string,
+        (proxyin.rawptr, out),
+    )
```
jpivarski: Even though the Python API function ensures that the `char*` string ends with a null terminator, the string might still have null bytes within the string. For instance,

```python
>>> b"arrow \xe2\x86\x92 zero \x00 not the end!".decode("utf-8")
'arrow → zero \x00 not the end!'
>>> b"arrow \xe2\x86\x92 zero \x00 not the end!".decode("utf-8").encode("utf-8")
b'arrow \xe2\x86\x92 zero \x00 not the end!'
```

Unless we explicitly ask for a length, we wouldn't be able to tell that the above string isn't `"arrow → zero "`. The ArrayBuilder has a assume-null-terminator method for convenience, but also a `char*`-and-length method for correctness:

https://github.com/scikit-hep/awkward/blob/38243b2f75fbf559bf31f97dfaeb02c428645c59/include/awkward/builder/ArrayBuilder.h#L304-L313

To use it, we also have to ask Python for the length. ([PythonAPI.string_as_string_and_size](https://github.com/numba/numba/blob/f4b6fe0124fa7734ba2bb28c7685db5b6306c486/numba/core/pythonapi.py#L1079-L1103), which is [PyUnicode_AsUTF8AndSize](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AsUTF8AndSize), and that size is the number of _bytes_, not the number of code-points, so that's good.)

```suggestion
    is_ok, out, length = pyapi.string_as_string_and_size(xval.value)
    length = ak._connect.numba.layout.castint(context, builder, numba.ssize_t, numba.int64, length)
    call(
        context,
        builder,
        ak._libawkward.ArrayBuilder_string_length,
        (proxyin.rawptr, out, length),
    )

```

tests/test_1677-array-builder-in-numba.py
```diff
@@ -0,0 +1,154 @@
...
+    out = builder.snapshot()
+    assert out.to_list() == [
+        True,
+        1,
+        1.1,
+        (1.1 + 0.1j),
+        np.datetime64("2020-09-04"),
+        np.timedelta64(5, "s"),
+        "hello",
+    ]
```
jpivarski, line 154: This will test the `"\x00"`-in-the-middle-of-a-string thing:

```suggestion
    builder = append(builder, "hello")
    out = builder.snapshot()
    assert out.to_list() == [
        True,
        1,
        1.1,
        (1.1 + 0.1j),
        np.datetime64("2020-09-04"),
        np.timedelta64(5, "s"),
        "hello",
    ]

    builder = append(builder, b"arrow \xe2\x86\x92 zero \x00 not the end!".decode("utf-8"))
    out = builder.snapshot()
    assert out.to_list() == [
        True,
        1,
        1.1,
        (1.1 + 0.1j),
        np.datetime64("2020-09-04"),
        np.timedelta64(5, "s"),
        "hello",
        b"arrow \xe2\x86\x92 zero \x00 not the end!".decode("utf-8"),
    ]
```

jpivarski (approved):
This is great, thanks!

I expanded on the string case to allow for `"\x00"` in the middle of a string, which is legal in Python. There's a test for it too; if you accept the change to the test first, it will fail, but if you accept both changes, it will (should!) pass. pre-commit might want to fix up the line width, too.

You didn't have to remove the commented-out strings in ArrayBuilder. When ArrayBuilder was being developed, I used those for debugging, and left them commented out in the codebase because they were something that I toggled often. But it has been a long time and we haven't needed to test ArrayBuilder much, so it's not bad to remove them now, as you've done.

You can merge this with the two changes; if you have no additional counter-edits. Thanks!

PR #1679, opened 2022-09-07
fix: carry parameters through broadcasting

### LINKED ISSUE #1672: ak.transform converts string to list of bytes if multiple arrays are passed
### Version of Awkward Array

1.9.0

Calling `ak.transform` on a single array containing strings will maintain the string type, but if multiple arrays are passed the strings are converted to lists of characters.

```python
In [1]: import awkward._v2 as ak

In [2]: a = ak.Array(["foo", "foo", "bar"])
   ...: a
Out[2]: <Array ['foo', 'foo', 'bar'] type='3 * string'>

In [3]: ak.transform(lambda *args, **kwargs: None, a)
Out[3]: <Array ['foo', 'foo', 'bar'] type='3 * string'>

In [4]: ak.transform(lambda *args, **kwargs: None, a, a)
Out[4]: 
(<Array ['foo', 'foo', 'bar'] type='3 * var * char'>,
 <Array ['foo', 'foo', 'bar'] type='3 * var * char'>)
```

agoose77:
Fixes #1672 

@jpivarski and I discussed the ramifications of this on Zoom. We have always been setting the parameters to None in 

This PR introduces a new `broadcast_parameters_rule` enumeration that selects from the following rules:
- `"intersect"`
   Take the intersection of the broadcasted input parameters.
- `"all_or_nothing"`
   Take the first content's parameters if they are equal to all of the other content's parameters, otherwise `None`.
- `"one_to_one"`
   Assert that the broadcast output matches the broadcast inputs, and if so, take the corresponding parameters (by position).
- `"none"`
   Always set parameters to `None` (previous behaviour)

### DIFF
--- a/src/awkward/_v2/_broadcasting.py
+++ b/src/awkward/_v2/_broadcasting.py
@@ -296,7 +296,7 @@ def _one_to_one_parameters_factory(
     The requested number of outputs is compared against the length of the given
     `inputs`. If the two values match, then a list of parameter objects is returned,
     where each element of the returned list corresponds to the parameters of the
-    content at the same position in the `contents` sequence. If the length of the
+    content at the same position in the `inputs` sequence. If the length of the
     given contents does not match the requested list length, a ValueError is raised.
     """
     # Find the parameters of the inputs, with None values for non-Contents
@@ -314,7 +314,7 @@ def _one_to_one_parameters_factory(
     return apply
 
 
-def _no_parameters_factory(
+def _none_parameters_factory(
     inputs: Sequence,
 ) -> BroadcastParameterFactory:
     """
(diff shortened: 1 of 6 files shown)


jpivarski (approved):
We discussed this on Zoom; I think this is a great solution!

The Sentinel will be moving to _util (:checkmark:).

I think the strings should be lower-case, at least when they are arguments to `ak.transform` (not done yet). It would be fine for the Enums to be upper-case (they're constants) while the strings that represent them are lower-case. It would also be fine to use just strings and not Enums. (This would be the first use of Enum.)

agoose77:
OK, I think this is ready to go. I want to, in a future PR, increase the coverage of these test. Right now, we have enough to consider merging this PR.

PR #1689, opened 2022-09-08
fix: don't project `categorical` in `ak._v2.packed`

### LINKED ISSUE #1688: `ak.packed` casts categorical to value type
### Version of Awkward Array

1.10.0.rc2

`ak.packed` seems to get rid of categoricals

```python
import awkward._v2 as ak

a = ak.Array(["foo", "foo", "bar"])
a_cat = ak.to_categorical(a)

assert a_cat.typestr == ak.packed(a_cat).typestr, f"'{a_cat.typestr}' != '{ak.packed(a_cat).typestr}'"
```

```pytb
      3 a = ak.Array(["foo", "foo", "bar"])
      4 a_cat = ak.to_categorical(a)
----> 6 assert a_cat.typestr == ak.packed(a_cat).typestr, f"'{a_cat.typestr}' != '{ak.packed(a_cat).typestr}'"

AssertionError: '3 * categorical[type=string]' != '3 * string'
```

This makes sense in that the indexed array is projected, but I did not expect `ak.packed` to change the type of an array.

The docstring also says this:

> Returns an array with the same type and values as the input, but with packed inner structures

agoose77:
Fixes #1688 by only packing the `Indexed[Option]Array`'s contents if it is a categorical type.

In the long run, it would be nice if we could mark a layout node as non-transient so that we can implement support for this without hard-coding a test for `__array__ == "categorical"`

### DIFF
--- a/src/awkward/_v2/contents/indexedarray.py
+++ b/src/awkward/_v2/contents/indexedarray.py
@@ -1244,7 +1244,16 @@ class IndexedArray(Content):
             raise ak._v2._util.error(AssertionError(result))
 
     def packed(self):
-        return self.project().packed()
+        if self.parameter("__array__") == "categorical":
+            return IndexedArray(
+                self._index,
+                self._content.packed(),
+                identifier=self._identifier,
+                parameters=self._parameters,
+                nplike=self._nplike,
+            )
+        else:
+            return self.project().packed()
 
     def _to_list(self, behavior, json_conversions):
         out = self._to_list_custom(behavior, json_conversions)
@@ -1261,8 +1270,8 @@ class IndexedArray(Content):
         return IndexedArray(
             index,
             content,
-            identifier=self.identifier,
-            parameters=self.parameters,
+            identifier=self._identifier,
+            parameters=self._parameters,
             nplike=nplike,
         )
 
--- a/src/awkward/_v2/contents/indexedoptionarray.py
+++ b/src/awkward/_v2/contents/indexedoptionarray.py
@@ -1684,10 +1684,20 @@ class IndexedOptionArray(Content):
 
     def packed(self):
         original_index = self._index.raw(self._nplike)
-
         is_none = original_index < 0
         num_none = self._nplike.index_nplike.count_nonzero(is_none)
-        if self._content.length > len(original_index) - num_none:
+        if self.parameter("__array__") == "categorical" or self._content.length <= (
+            len(original_index) - num_none
+        ):
+            return ak._v2.contents.IndexedOptionArray(
+                self._index,
+                self._content.packed(),
+                self._identifier,
+                self._parameters,
+                self._nplike,
+            )
+
+        else:
             new_index = self._nplike.index_nplike.empty(
                 len(original_index), dtype=original_index.dtype
             )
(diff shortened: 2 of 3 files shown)


src/awkward/_v2/contents/indexedarray.py
```diff
@@ -1244,7 +1244,16 @@ def continuation():
...
+        if self.parameter("__array__") == "categorical":
+            return IndexedArray(
+                self._index,
+                self._content.packed(),
+                identifier=self._identifier,
+                parameters=self._parameters,
+                nplike=self._nplike,
+            )
+        else:
+            return self.project().packed()
```
jpivarski, line 1256: This is good, what I was about to write before remembering that we now have a `copy` method to simplify this sort of thing. But let's not introduce the use of `copy` just now. (It can be a later refactor, very low on the priority list.)

src/awkward/_v2/contents/indexedarray.py
```diff
@@ -1261,8 +1270,8 @@ def _to_nplike(self, nplike):
         return IndexedArray(
             index,
             content,
-            identifier=self.identifier,
-            parameters=self.parameters,
+            identifier=self._identifier,
+            parameters=self._parameters,
```
jpivarski, line 1274: Good catch: we should prefer private variables within the class (just to have a rule to follow). In particular, `self.parameters` creates a dict when one is not needed.

jpivarski (approved):
Great! For the record (we talked about this on Slack), we decided to keep this simple and not remove unused elements from the `content`. Also, @ivirshup, who raised the issue, argued against it in https://github.com/scikit-hep/awkward/issues/1688#issuecomment-1240826171.

jpivarski:
Also for the sake of history, this was @agoose77's attempt to remove unused contents:

```python
        if self.parameter("__array__") == "categorical":
            index_nplike = self._nplike.index_nplike
            index = index_nplike.asarray(self._index.data)
            mask = index_nplike.zeros(len(self._content), dtype=np.bool_)
            mask[index] = True
            tag_valid, = index_nplike.nonzero(mask)
            index_valid = index_nplike.searchsorted(tag_valid, index)
            return IndexedArray(
                ak._v2.index.Index(index_valid),
                self._content._carry(ak._v2.index.Index(tag_valid), False),
                self._identifier,
                self._parameters,
                self._nplike
            )
```

and I was thinking about it along these lines:

```python
>>> index = np.array([4, 2, 6, 1, 5, 5, 2])
>>> content = np.array([0.0, 1.1, 2.2, 3.3, 4.4, 5.5, 6.6, 7.7, 8.8, 9.9])
>>> uniques, inverse = np.unique(index, return_inverse=True)
>>> uniques
array([1, 2, 4, 5, 6])
>>> inverse
array([2, 1, 4, 0, 3, 3, 1])
>>> uniques[inverse]
array([4, 2, 6, 1, 5, 5, 2])
```

Both of these are _O(n log n)_, either in the `np.searchsorted` or in the `np.unique` (because it sorts). For what we want, it wouldn't have to sort, but we might have needed to write a custom kernel to actually get _O(n)_.

jpivarski:
Since this merged PR is based on a branch named `agoose77/fix-categorical-packed-2`, would it be safe to say that [agoose77/fix-categorical-packed](https://github.com/scikit-hep/awkward/tree/agoose77/fix-categorical-packed) can be deleted? If so, please do it. (I'm cleaning up dead branches.)

agoose77:
Done!

PR #1666, opened 2022-09-05
chore: add isort

henryiii:
This adds isort, which sorts the includes, removing that as a variable and enabling things like injecting `__future__` imports later.

If there are any "this needs to be imported before that" rules, we can add those as isort rules.

PR #1690, opened 2022-09-09
reference: feat: replace v1 with v2

### LINKED ISSUE #1701: ak._v2.from_iter should recognize Python datetimes/timedeltas
### Version of Awkward Array

HEAD

https://github.com/scikit-hep/awkward/blob/77b06b3575737ec3cebc4c9b76ddceaf584d57c3/src/python/content.cpp#L952-L957

and

https://github.com/scikit-hep/awkward/blob/77b06b3575737ec3cebc4c9b76ddceaf584d57c3/src/python/content.cpp#L846-L886

handle NumPy datetime objects, but Python datetime objects are not explicitly recognized. (What happens? Does it raise a "cannot convert..." ValueError?)

Given a `datetime.datetime`, one can call `.timestamp()` to get a number of seconds since 1970 as a floating point number. To fill the `ArrayBuilder::datetime` or `ArrayBuilder::timedelta`, we have to provide a 64-bit integer and a type string. We could _choose_ that type string to be `"s"` and cast the floating point number as an integer as-is, or _choose_ the type string to be `"ms"` and multiply the number by 1000, etc.

Either we hard-code a choice or pass it down as an argument (but that means passing it down through C++, like the options in `from_json`). Note:

   * `"ns"` (nanosecond) resolution gives us a range from 1677 through 2262
   * `"us"` (microsecond) resolution gives us a range from 290307 B.C.E through 294247 C.E., though if we were _constructing_ Python `datetime` objects, these years are already out of range.
   * there's no point in going to less granular units (`"ms"`, `"s"`) if Python `datetime` can't even represent them.

So the only units that make any sense to assume are microseconds and nanoseconds. Microseconds covers the entire range that Python `datetime` is capable of (and more), and nanosecond resolution sounds rather special-purpose. (Users can make NumPy datetimes if they need that.) So I would vote to hard-code it as microseconds.

agoose77:
This is just me playing around with removing v1. Should you want to do this yourself, no problem!

- [x] Replace v1 tests with v2.
- [x] Fix up setup.cfg linting rules.
- [x] Move to `malloc` and `free` instead of Awkward allocators.
- [x] Remove preliminary set of unused kernels. Any unused specialisations are not yet removed. 
- [x] Replace use of contents in Forth with NumPy arrays or import v2 objects
- [x] Remove kernel dispatch, replace `kernel::malloc` with `malloc` or `new`
- [x] Support `datetime.datetime` in `from_iter`
- [x] Move `array_deleter` to `include/awkward/util.h`
- [ ] Replace `pyobject_deleter` with the appropriate pybind11 feature.
- [x] Move `LayoutBuilder` / `ArrayBuilder` to `_ext`
Fixes #1701


agoose77:
Removing the v1 Python layer is fairly easy. Need to think about the C++ layer; the Forth machine returns a `ContentPtr` in a number of places. Some of these are trivial Numpy arrays, but the `bytecodes` getter function returns a jagged list. The obvious answer is to return the bytecodes and their offsets, and let the Python binding layer build the Array object from C++ via pybind Python execution. I haven't thought yet whether I like the idea of the ForthMachine returning pybind11's `py::array` outside of the Python layer, so I think I'm leaning towards removing the `toNumpyArray` interface and doing that at the bindings level.

jpivarski:
> Need to think about the C++ layer; the Forth machine returns a `ContentPtr` in a number of places. Some of these are trivial Numpy arrays, but the `bytecodes` getter function returns a jagged list. The obvious answer is to return the bytecodes and their offsets

Knowing that the split is coming up soon, we've been using the ForthMachine in ways that don't depend on the return values being `ContentPtr` and `IndexPtr`: we've been immediately casting them as NumPy arrays.

The `bytecodes` method/property is not used much. I think nothing would break if it were removed entirely, but it's probably better to replace it with just a 2-tuple of `offsets` and `content`, as you describe. It exists as a debugging tool (though the `decompile` one, which prints out source code reconstructed from the bytecodes, is more useful), and was constructed as a jagged array just because it occurred to me that one could do that. Unlike imperative bytecode, Forth bytecode does not have arbitrary jumps; it happens to be a [structured language](https://en.wikipedia.org/wiki/Structured_programming) in the sense of not having GOTOs, a fact that was [emphasized by Forth enthusiasts in the 80's](https://www.forth.com/resources/forth-programming-language/#242_Structured_programming_disciplines), but nowadays nearly all languages are structured at the human-readable source code level. The only thing that stands out for Forth is that its bytecode can also be blocked out as subroutines, a jagged array.[^1]

Anyway, the point is that we've been using the `ContentPtr`/`IndexPtr`-dependent parts of ForthMachine in a minimal way so that they can be replaced by raw buffers (`py::buffer_info` in pybind11) without issues. Along the same lines, we will be removing the custom growable-buffer of ForthOutputBuffer to use Manasvi's new GrowableBuffer (and reap the benefits of non-contiguous filling), but that is an entirely separate step that can wait until the v1 pruning is entirely done. (ForthOutputBuffer doesn't depend on the old GrowableBuffer—that one can be safely deleted.)

[^1]: Although somebody might point out that this is now true of other bytecodes that I don't know about. Maybe JVM? I'm pretty sure Python bytecode has unconditional jumps.

agoose77:
Phew. 

Now v1-less Awkward compiles and passes most tests, having removed the array and type machinery. 

I've removed the low hanging fruit in the C++ side, so I've therefore left a much reduced version of `Index.cpp`. My rationale there is that older C++ standards don't have `std::span` yet, so either we start using `pybind` features inside of AwkwardForth, or we keep this small abstraction.

New bugs:

`from_iter` doesn't currently handle native Python datetime objects, so many tests fail. We should fix this.

jpivarski:
We should be able to remove `Index` from `ForthMachine`. I'll take a look at it later, but that's a separable problem from everything else.

agoose77:
> We should be able to remove `Index` from `ForthMachine`. I'll take a look at it later, but that's a separable problem from everything else.

I think it depends if we want to be passing around unbounded arrays. Actually, this is barely a point at the moment because the `Index` impl doesn't so anything safety wise (i.e. provide an iterator).

IIRC you never get an index directly; rather, you get the output and call `toIndexXX` on it, so we can just use the metadata from the output.

Otherwise, I still need to remove the unused kernels...

jpivarski:
In the way AwkwardForth is used, the output of `toNumpyArray` and `toIndex` are used, they're immediately wrapped with `np.asarray`, so we only need one way to get a buffer out. We can even change the name (breaking Uproot), since Uproot v5 and Awkward v2 releases will be coordinated. A name change won't be very disruptive.

> Otherwise, I still need to remove the unused kernels...

Perhaps removing unused kernels should be a separate PR. We might even be able to find more that could be removed and haven't yet. Some of them might be equivalent to `arange` or something.

jpivarski:
> wip: remove unused kernels

It's okay. But I think we can remove more, and fishing for them can be a different PR.

agoose77:
> > wip: remove unused kernels
> 
> It's okay. But I think we can remove more, and fishing for them can be a different PR.

Cool. I found these ones by directly looking with (Xonsh)
```xonsh
import yaml
import re

kernel_spec = yaml.safe_load(p"/home/angus/Git/awkward/kernel-specification.yml".read_text())['kernels']
in_use_specs = set($(rg r"awkward_\w+" src/awkward/contents src/awkward/_reducers.py -oNI).splitlines())
name_to_specs = {g['name']: [s['name'] for s in g['specializations']] for g in kernel_spec}
spec_to_name = {v: k for k, s in name_to_specs.items() for v in s}

# For cleaning up the spec file
in_use_names = (name_to_specs.keys() & in_use_specs)

# For removing kernels
for f in [f for f in pg`src/cpu-kernels/awkward*.cpp` if not any(s in f.read_text() for s in used_specs)]:
    f.unlink()
```

agoose77:
>  the output of toNumpyArray and toIndex are used

Yeah, now that all the removed code is gone, it's easier to see that we don't need an object that holds both length and ptr (besides the output itself). The main "issue" is internally within the Forth machine, rather than the bindings. It should be possible to just access the (shared) pointer directly within the Forth routines, and wrap it in the bindings.

agoose77:
>  We might even be able to find more that could be removed and haven't yet. Some of them might be equivalent to arange or something.

Agreed, there are several that I know of (not sure if they're still carried through in this PR, haven't checked) ;)

agoose77:
OK, this now removes `Index` as well, so it's nearly all the way there. I've rebased my commits where possible to give a clean-ish history. I can't guarantee that every commit builds a wheel that passes the test suite (they certainly don't when I start moving files), but after the large-scale refactoring commits, most revisions then compile.

agoose77:
I haven't fully pulled out the kernel dispatch mechanism yet. I'll tentatively note that I'm not intimately familiar with this code, but my current understanding is that our kernel handling has all been superseded by the nplike dispatch mechnism.

jpivarski:
Any kernel-dispatch in C++ is old. We have an entirely new kernel-dispatch system in Python, using ctypes.

Anything you find in C++ about `startup` functions, particularly for setting up awkward-cuda-kernels, can be removed. That's all Python now as well.

Another thing that can go is the dlpack git submodule. That was all for supporting CUDA arrays in C++, which is now handled by CuPy in nplike.

agoose77:
> Another thing that can go is the dlpack git submodule

It's already gone!

agoose77:
OK, now we're at a point where there's not much else (anything?) that needs to be removed. 

The main changes are now described in this PRs description.

agoose77:
I've added some rudimentary time handling. I originally tried employing `std::chrono`, but I clearly don't fully understand how C++ is handling times vs Python:

If I calculate the local epoch using:
```diff
+std::chrono::system_clock::time_point local_epoch() {
+    std::tm tm = {
+       /* .tm_sec  = */ 0,
+       /* .tm_min  = */ 0,
+       /* .tm_hour = */ 0,
+       /* .tm_mday = */ 1,
+       /* .tm_mon  = */ 1 - 1,
+       /* .tm_year = */ 1970 - 1900,
+    };
+    // Use local time zone DST
+    tm.tm_isdst = -1;
+    return std::chrono::system_clock::from_time_t(std::mktime(&tm));
+}
```
Then when I use this in these calculations, I get one-hour shifts for particular datetimes:
```diff
+    auto time = obj.cast<std::chrono::system_clock::time_point>();
+    int64_t time_since_epoch_us = std::chrono::duration_cast<std::chrono::microseconds>(
+        time - local_epoch()
+    ).count();
+    std::cout << "time since epoch     " << time_since_epoch_us << std::endl;
+    std::cout << "time since epoch cpp " << std::chrono::duration_cast<std::chrono::microseconds>(time.time_since_epoch()).count() << std::endl;
+    self.datetime(time_since_epoch_us, "datetime64[us]");
```

I am currently under the impression that this is because the timezones differ between Python and C++. I'm looking into this.

agoose77:
The above code example is wrong - the epoch should not be specified in local time (via `std::tm`). Instead, it's just `std::time_t (0)`.

Now that I'm more familiar with `mktime` vs `localtime`, some thoughts on my timezone (BST)
- NumPy's datetime objects are _naive_ and don't attempt to determine any DST offsets
- `datetime.datetime.timestamp` _does_ try to use the local timezone to figure out DST offsets
-  `(dt - epoch)` does not try to do this, so behaves like NumPy
- `std::mktime` (used by pybind11) converts from _local_ clock into UTC. The `tm_isdst` flag is used to help `mktime` figure out how to transform the input. It's set to `-1` by default, which means that `mktime` uses the local clock to figure out if the time period *should* be during DST, and corrects it if so
  | tm_isdst | expected DST? | offset | new tm_isdst |
  |----------|---------------|--------|--------------|
  | 1        | yes           | 0      | 1            |
  | 1        | no            | -1     | 0            |
  | 0        | yes           | 1      | 1            |
  | 0        | no            | 0      | 0            |
  | -1       | yes           | 0      | 1            |
  | -1       | no            | 0      | 0            |
		   
	 I'm least confident on `-1` for `tm_isdst`. I believe that `-1` just signals to `mktime` to determine from the local time whether the given time is DST, and reflect that guess in the field. It *doesn't* modify the time fields after deduction - it assumes the user entered the correct local time for the timezone. `mktime` is idempotent, as it only changes fields if `tm_isdst` is changed. 
- `time_t` literally returns the same as `time_since_epoch()`
- `mktime` considers whether `tm_isdst` changes when building the result `time_t`.

agoose77:
Actually, I didn't think about this - the timezone boundaries do not map onto distinct values: 01:00 and 02:00 (UTC) in my current timezone on 2022/03/27 are both the same local times (01:00):
- 1 AM BST = 1 AM UTC
- 2 AM BST = 1 AM UTC
so, it's not possible to losslessly recover the input (local) time from the UTC output (in `std::time_t`). Therefore, we either need to modify the timezone (set it to UTC), or do this in Python.

jpivarski:
Since NumPy does not deal with timezones (its times are timezone-naive, which is to say that knowledge of the timezone is not contained within the array, but is managed by the user in some external metadata somewhere), then we have to operate at the same level (also timezone-naive).

Python's `datetime` library can operate in both modes: https://stackoverflow.com/a/7065242/1623645

Probably the least-surprising way to turn timezone-aware `datetimes` (`datetimes` with the `tzinfo` set to non-None) into timezone-naive `datetimes` would be to remove the `tzinfo`, rather than converting to UTC, as in the StackOverflow example. The best we can do is to be consistent and document it.

This is not the worst thing to happen to people who deal with time-data...

agoose77:
We're on the same page w.r.t whether we support timezones!

In this PR, the issue I'm running into is concerned with the pybind11 bindings for `datetime.datetime` objects. If we want to bump performance by not calling into the Python interpreter several times per `datetime.datetime` object, then it's best if we convert them once into a native C++ type, e.g. `std::time_t`. However, the pybind11 converters use `mktime` to produce a `std::chrono::system_clock::time_point` object (which is convertible to `time_t`) according to the _localtime_ settings (see https://github.com/pybind/pybind11/issues/1638). This means that the conversion result depends upon the user's system.

Actually, writing this response has given me an idea for a "hack". These problems all go away if the localtime is UTC, because there are no DST periods. However, setting the `TZ` env var from requires us to add special code to handle windows `_putenv` vs linux `putenv`. Or, we could just call into Python and let Python handle this, once per `from_iter` call. I'd prefer to just fix this at the root, which would probably require pybind11 to ship a third-party header-only library, (or we could do this and forgo pybind11's conversion).

jpivarski:
Oh, I see: not specifying a timezone and converting a time-tuple into seconds since 1970 will implicitly pick up the platform-dependent locales. Well, we definitely don't want it to be platform- or locales-dependent, or to be changing OS-level variables to do the conversion. If there's a way to satisfy all constraints but `from_iter` runs a little slower, `from_iter` runs a little slower. It's expected that we'll be making a lot of Python calls, since we are converting from Python data, after all.

Actually, WWND (What Would NumPy Do)?

```python
>>> import numpy as np
>>> from datetime import datetime
>>> from dateutil import tz
>>> d = datetime(1976, 7, 2, 4, 0, 0, tzinfo=tz.tzoffset("EDT", -4*60*60))
>>> d
datetime.datetime(1976, 7, 2, 4, 0, tzinfo=tzoffset('EDT', -14400))
>>> d.isoformat()
'1976-07-02T04:00:00-04:00'
>>> np.array([d])
array([datetime.datetime(1976, 7, 2, 4, 0, tzinfo=tzoffset('EDT', -14400))],
      dtype=object)
>>> np.array([d], "datetime64[s]")
<stdin>:1: DeprecationWarning: parsing timezone aware datetimes is deprecated; this will raise an error in the future
array(['1976-07-02T08:00:00'], dtype='datetime64[s]')
```

Aha! It converts the timezone-aware `datetime` to UTC, which is platform-independent (it's not relying on the system's local timezone; the timezone is in the `datetime` object), but this is deprecated behavior and will raise an error in the future. See also [here](https://numpy.org/doc/stable/release/1.11.0-notes.html?highlight=numpy%20random%20f#datetime64-changes).

So we can just raise an error. (If it weren't for this NumPy deprecation, we could do that conversion as well.)

Thus, any `datetime` for which `d.tzinfo is not None` would stop processing as an error. It would be good to know what kind of error this would be, but I would assume `ValueError`.

agoose77:
> Oh, I see: not specifying a timezone and converting a time-tuple into seconds since 1970 will implicitly pick up the platform-dependent locales.

Yes that's it - naive datetimes are treated as local times during conversion.

> changing OS-level variables to do the conversion.

I think this would be fine if we are doing it in a contextual manner, e.g. RAII to setup/tear down, but still non ideal.

> So we can just raise an error.

Yes, separately for the case of timezone-specified datetimes, we should probably error.

agoose77:
I've fixed the remaining bug with date-time handling (I introduced this during refactoring). This means that the test suite now passes. I've done a very simple fix of the docs-building so we don't reference any v1 layouts anymore. I won't spend too much time on this because much of this work is/will be done in the `docs` branch.

In this PR, I'd mainly benefit from reviewers considering edge cases for `sed` et al, because that's how most of this work was done.

agoose77:
I've now rebased this PR on top of #1708, given that I expect that PR will be merged into 1.10 (it's a significant bug).

I've taken the time to rewrite most of these commits to join together similar changes, and remove intermediate commits that were used to keep `localbuild` happy. It's possible that some of the commits won't build / pass the test suite, although this is guaranteed for the early v1-v2 move commits (by virtue of separating `mv` from `sd` (`sed`)).

src/libawkward/layoutbuilder/ListArrayBuilder.cpp
```diff
@@ -66,18 +66,26 @@ namespace awkward {
     const ForthOutputBufferMap& outputs) const {
     auto search = outputs.find(vm_output_data());
     if (search != outputs.end()) {
-      auto offsets = search->second.get()->toIndex64();
+      auto output = search->second;
+      if (output->dtype() != util::dtype::int64) {
```
ianna: I don't think this check is needed - the output type is encoded by a generated Forth code.

src/libawkward/layoutbuilder/ListOffsetArrayBuilder.cpp
```diff
@@ -69,18 +69,26 @@ namespace awkward {
     BuffersContainer& container,
     const ForthOutputBufferMap& outputs) const {
     auto search = outputs.find(vm_output_data());
-    auto offsets = search->second.get()->toIndex64();
+    auto output = search->second;
+    if (output->dtype() != util::dtype::int64) {
```
ianna: same here

src/libawkward/layoutbuilder/UnionArrayBuilder.cpp
```diff
@@ -95,28 +96,40 @@ namespace awkward {
...
-        index.data(),
-        current.data(),
-        length,
-        tags.data(),
-        length);
-      util::handle_error(err, "UnionArray", nullptr);
+      auto output = search->second;
+      auto length = (ssize_t)output->len();
+
+      if (output->dtype() != util::dtype::int8) {
```
ianna: same here

ianna (commented):
@agoose77 - an impressive amount of work! Thanks!
I briefly skimmed through the changes. I have a couple of comments for the C++ side: I don't think, you need to have type checks in `LayoutBuilder` because it generates the Forth code with the required types any way.

Replacing awkward memory allocations is correct.

src/libawkward/layoutbuilder/ListArrayBuilder.cpp
agoose77: Hi @ianna, thanks for reviewing! Could you elaborate on this, i.e. why you think we _know_ we have an Index64? Are you saying that the Forth code that generates these buffers is producing 64 bit integers?

If so, I believe this might still be needed, as a safety mechanism (even if it's "less safe" than the previous virtual-dispatch based approach).

ianna: The constructor of the builder generates the Forth code, that defines the type of its output buffer.

jpivarski:
1.10.1 is out and I'm reviewing this now.

I'll start by pressing "update branch" so that it will test against `main`, though I'm pretty sure that the only change is VERSION_INFO or something like it. For formality's sake, since this is such a big PR.

jpivarski (changes requested):
I'm going to give feedback in small pieces, as I come across them.

The low-level ArrayBuilder should not go into `ak.contents`, since this is for `Content` subclasses exclusively.

(Similarly, the `Index` subclasses, `Identifier`, and low-level `Record` were also moved out of this directory. The idea is that we can quickly open all the files in this directory and get every `Content`, to apply some small change to all of them, typically.)

jpivarski:
It should also be possible to remove

```c++
  EXPORT_SYMBOL void* awkward_malloc(int64_t bytelength);
  EXPORT_SYMBOL void awkward_free(void const *ptr);
```

from include/awkward/kernel-utils.h and wherever they're implemented. I think you've already converted over all of their uses to `new` and `delete` or standard `malloc` and `free`, right?

jpivarski:
I believe we'll be able to remove all instances of `pyobject_deleter` in favor of pybind11 features, and I don't think there's any need for `array_deleter`, now that we're not sharing C++-allocated arrays with Python anymore. Even if this could be done with a few extra steps, those steps might be tricky and we can put that off for a new PR.

agoose77:
> The low-level ArrayBuilder should not go into ak.contents, since this is for Content subclasses exclusively.
(Similarly, the Index subclasses, Identifier, and low-level Record were also moved out of this directory. The idea is that we can quickly open all the files in this directory and get every Content, to apply some small change to all of them, typically.)

Could you elaborate on this Jim? I don't recall merging layout into contents. Although, I may have done this in an intermediary commit that I couldn't squash for whatever reason.

agoose77:
>  and I don't think there's any need for `array_deleter`

I think we do still need this; it's used internally to free `new`d arrays (before C++17): https://stackoverflow.com/questions/13061979/shared-ptr-to-an-array-should-it-be-used

jpivarski:
> Could you elaborate on this Jim? I don't recall merging layout into contents.

Aha! You didn't: it was only suggested by a documentation file.

```bash
% fgrep -r ak.contents.ArrayBuilder *
docs-sphinx/index.rst:**Layout-level ArrayBuilder:** :class:`ak.contents.ArrayBuilder` (used internally).
```

I've compiled this and get a lot of linker warnings. We'll address the compilation warnings (almost all integer size or signedness) in another PR, but the linker warnings make me uneasy; I'd like to understand those. I think it's visibility: do you see those warnings, too?

jpivarski (dismissed):

jpivarski:
`array_deleter`: I see. Then it will be eliminated in a later pass, when we upgrade to C++17 in a later PR.

We will eventually be able to get rid of `pyobject_deleter`, too. Right now, it's only being used to reference-count `ForthInputBuffer`'s shared ownership of input arrays, but that can be managed with a pybind11 feature. (Future PR.)

We've also talked about replacing pybind11 with nanobind. Doing that might be a good way to refresh and simplify the C++/Python interface.

src/awkward/_connect/numpy.py
```diff
@@ -14,75 +14,64 @@
...
+def _to_rectilinear(arg):
+    if isinstance(arg, tuple):
+        nplike = ak.nplike.of(*arg)
+        return tuple(nplike.to_rectilinear(x) for x in arg)
+    else:
+        nplike = ak.nplike.of(arg)
+        nplike.to_rectilinear(arg)
+
+
+def array_function(func, types, args, kwargs):
```
jpivarski, line 36: Nice! You enabled NEP 18. Are there any v2 tests that test it?

src/awkward/behaviors/categorical.py
```diff
@@ -89,262 +86,24 @@ def _categorical_equal(one, two):
...
         else:
             nextinputs.append(x)
 
     return getattr(ufunc, method)(*nextinputs, **kwargs)
 
 
-ak.behavior[ak.nplike.numpy.ufunc, "categorical"] = _apply_ufunc
-
-
-def is_categorical(array):
```
jpivarski, line 116: I verified that all of these functions have been moved to operations.

jpivarski:
LayoutBuilder has become a C++-only thing, now that it's defined in terms of templates. We shouldn't have two versions, one based on AwkwardForth and the other on C++ templates, but the AwkwardForth version can be removed in a future PR.

jpivarski:
`diff` could be improved by recognizing YAML syntax and not trying to interleave two structures when one whole subtree is removed.

jpivarski:
Our functions have a mixture of U.S. and U.K. spelling. That's not necessarily a bad thing; I don't think I would just guess the spelling, I'd look it up, or an IDE would tab-complete.

src/awkward/_typetracer.py
```diff
@@ -5,18 +5,23 @@
...
-import awkward._util
 import awkward as ak
 
 np = ak.nplike.NumpyMetadata.instance()
 
 
+# FIXME: deduplicate this
+# matches include/awkward/common.h
+kMaxInt64 = 9223372036854775806  # 2**63 - 2: see below
+kSliceNone = kMaxInt64 + 1  # for Slice::none()
```
jpivarski: Having constants defined once per language is, I think, not bad. The contortions you'd need to pick up common values (e.g. single file source of truth and reading the file in both languages, at compile-time in C++) would be more obscuring than the potential for error in the two copies of the constants.

Maybe they should have symmetric positions in both languages: util.h and _util.py, for instance.

The danger of duplication is that one value can be changed without the maintainer knowing about or remembering the existence of the other value, and then strange bugs occur when some functions pick up one value and other pick up another. But for this sort of constant, that can _never_ happen: we are 100% certain that we'll never change the value of `kMaxInt64`, since it is derived from the bit width of 64-bit numbers.

src/awkward/_util.py
```diff
@@ -44,124 +72,357 @@ def is_sized_iterable(obj):
...
+        return self._kwargs["arguments"]
 
-def deprecate(
-    message,
-    version,
-    date=None,
-    will_be="an error",
-    category=DeprecationWarning,
-    stacklevel=2,
-):
```
jpivarski: This `deprecate` is a useful function that we'll likely want in version 2. It didn't get copied over into the v2 _utils.py because it hasn't been needed _yet_: we've been free to change behavior at will, since 2.0.0 hasn't been released.

But eventually, we'll need it, so just copy it verbatim from v1 _util.py into v2 _util.py.

src/awkward/_util.py
```diff
@@ -44,124 +72,357 @@ def is_sized_iterable(obj):
...
 
-unknowntypes = (ak.layout.EmptyArray,)
+# def deprecate(
+#     message,
+#     version,
+#     date=None,
+#     will_be="an error",
+#     category=DeprecationWarning,
+#     stacklevel=2,
+# ):
```
jpivarski: Nevermind, here it is: it's just commented out in v2 _util.py (and GitHub is making a mess of the difference).

So it just needs to be un-commented.

src/awkward/_util.py
```diff
@@ -31,6 +29,36 @@
 kSliceNone = kMaxInt64 + 1  # for Slice::none()
 kMaxLevels = 48
```
jpivarski, line 31: Maybe you meant de-duplicating with this copy of constants? (Note the comment: `# matches include/awkward/common.h`.)

src/awkward/_util.py
```diff
@@ -536,1377 +801,389 @@ def extra(args, kwargs, defaults):
...
-    if attempt is None:
-        raise ValueError(
-            f"key {key!r} not found in record" + exception_suffix(__file__)
-        )
-    else:
-        return attempt
-
-
-key2index._pattern = re.compile(r"^[1-9][0-9]*$")
+# def key2index(keys, key):
```
jpivarski: We apparently never needed the `key2index` function. It can be deleted.

src/awkward/_util.py
```diff
@@ -536,1377 +801,389 @@ def extra(args, kwargs, defaults):
...
+#         ))
+#     else:
+#         return attempt
 
-    elif isinstance(array, virtualtypes):
-        return completely_flatten(array.array)
 
-    elif isinstance(array, unknowntypes):
-        return (ak.nplike.of(array).array([], dtype=np.bool_),)
+# key2index._pattern = re.compile(r"^[1-9][0-9]*$")
```
jpivarski: Including this helper.

src/awkward/_util.py
```diff
@@ -536,1377 +801,389 @@ def extra(args, kwargs, defaults):
...
 
-    elif isinstance(array, uniontypes):
-        out = []
-        for i in range(array.numcontents):
-            tmp = completely_flatten(array.project(i))
-            assert isinstance(tmp, tuple)
-            for x in tmp:
-                out.append(x)
-        return tuple(out)
+# def make_union(tags, index, contents, identifier, parameters):
```
jpivarski: The same is true of `make_union`: we never needed it in v2 and it can be deleted.

jpivarski:
GitHub is managing changes during review really well. It just unchecks the "viewed" files that have changed.

image

src/awkward/_util.py
```diff
@@ -1187,3 +1188,30 @@ def _impl(array):
...
+
+
+def deprecate(
+    message,
+    version,
+    date=None,
+    will_be="an error",
+    category=DeprecationWarning,
+    stacklevel=2,
+):
```
jpivarski: If you're just adding this, there might be a commented-out version elsewhere in the file.

Now that `deprecate` is added, all of the commented-out functions can be removed.

agoose77: Yes, noticed it after pushing ;)

src/awkward/_util.py
agoose77: Could you clarify, is it just my short-hand that needs elaborating here? I.e. 
> Duplicated constants from `include/awkward/common.h`

We could also export them if we needed.

src/awkward/_typetracer.py
agoose77: Ah yes, good catch. I forgot that I had to introduce this during refactoring.

agoose77:
@jpivarski  can you specifically check d703dfa85874fa8d50affeb24d22273ebe409fe5 and ensure I've made the right change? I think the original v2 conversion here wanted to reject the low level `ArrayBuilder` here, rather than the highlevel array that we already treat in the case that handles `ak.Array`.

jpivarski:
I was just going to suggest (and couldn't do the Suggest Change button, for some reason) that it be

```python
    elif isinstance(array, ak.highlevel.ArrayBuilder):
        return array.type
```

This is a case we can handle. (I just tested it.)

I'll take a look at the function as a whole to see if it's in the right order and whether it's missing the low-level ArrayBuilder case.

jpivarski:
Oh yeah: the `ak.highlevel.ArrayBuilder` case has already been handled above. From context, what was probably intended here is the low-level ArrayBuilder.

agoose77:
Oops, that was what I intended to say when I asked you to look into it. Lazy english!

agoose77:
@jpivarski oh, sorry. I had not pushed the full change in d703dfa85874fa8d50affeb24d22273ebe409fe5. I'd locally changed the case to `ak.layout.ArrayBuilder`, as you suggested.

jpivarski:
```python
    elif isinstance(array, ak.layout.ArrayBuilder):
        form = ak.forms.from_json(array.form())
        return ak.types.ArrayType(
            form.type_from_behavior(self._behavior), len(self._layout)
        )
```

Eventually, we'll want to get rid of `ak.layout` as a module. Now it only contains the low-level ArrayBuilder and two instances of LayoutBuilder-as-Python that we'll be removing. Maybe that's too much to tack onto this PR. But after that point, the low-level ArrayBuilder can be referred to simply as `ak._ext.ArrayBuilder`.

src/awkward/_connect/numpy.py
agoose77: The test suite hits the `array_function` implementation in three test-files (although I can't say how extensively).

agoose77:
@jpivarski NB I haven't touched "studies/" yet, because that should be a simple find-replace and I didn't want to grow this stage of the PR review yet.

agoose77:
> ```python
>     elif isinstance(array, ak.layout.ArrayBuilder):
>         form = ak.forms.from_json(array.form())
>         return ak.types.ArrayType(
>             form.type_from_behavior(self._behavior), len(self._layout)
>         )
> ```
> 
> Eventually, we'll want to get rid of `ak.layout` as a module. Now it only contains the low-level ArrayBuilder and two instances of LayoutBuilder-as-Python that we'll be removing. Maybe that's too much to tack onto this PR. But after that point, the low-level ArrayBuilder can be referred to simply as `ak._ext.ArrayBuilder`.

I'll make this a todo for later review

jpivarski:
> @jpivarski NB I haven't touched "studies/" yet, because that should be a simple find-replace and I didn't want to grow this stage of the PR review yet.

The studies/ directory should be left as-is. It's just a record of the temporary work that we did to figure out the algorithms, and so removing v1 doesn't change anything there.

The code in studies/ doesn't even need to be able to run. It's a historical record.

jpivarski:
There's a lot of JSON-handling code that can now be removed, too. Would you mind if I `git pull` and do it?

That is, you're not working on something that would be hard to merge if I remove some C++?

agoose77:
@jpivarski sure, which module is this? I thought I'd removed the unused parts. Perhaps some of src/libawkward/io/json.cpp is not required actually.

jpivarski:
There's more, but it's not specifically called out.

Also, I can remove the non-C++ template LayoutBuilder (i.e. the old one). I can do the JSON and the LayoutBuilder in two separate commits, for record-keeping.

jpivarski:
I just did `git pull` _now_, for the sake of synchronization.

agoose77:
Ah yes - I left the old layout builder because it still works, and it can be used right-now (whilst we don't have a jitted version of the new layout builder).

jpivarski:
Are there tests with the old LayoutBuilder?

The motivation for reimplementing LayoutBuilder was so that it could be optimized for use in compiled loops. A Python interface to such a thing doesn't make sense because you wouldn't be able to use it in compiled loops. The C++ LayoutBuilder will eventually be usable in C++ JIT-compiled code (cppyy) and another implementation is planned for Numba.

But if we have any tests now with LayoutBuilder in a Python loop, we can just remove them without replacement, since future uses of LayoutBuilder won't involve anything like Python loops.

agoose77:
@jpivarski Yes, `tests/test_0924-layout-builder.py` and `tests/test_1302-layout-builder-nested-list.py`

agoose77:
Ah good catch @jpivarski. And I hadn't realised that the ToJSON wasn't used any more. I wonder if it was used for the forms, which I had temporarily left in after removing the contents? :thinking:

jpivarski:
I don't think so; I think the forms built JSON with strings.

As for the two versions of from-JSON: there would have been no way to know. I recently revamped the JSON-handling code, and when I needed a C++ function, I wrote a new one beside the old one so that it would be easier to remove the old one (what I just did). There wasn't an obvious indication that one was old and the other new.

I'm almost done with the LayoutBuilder.

studies/awkward_cupy_perfscript.py
```diff
@@ -1,12 +1,12 @@
 import awkward as ak
```
jpivarski, line 1: This file (being in the studies/ directory) didn't need to be changed, but it's not bad to drop all the "._v2"s.

jpivarski (approved):
LGTM!

image

As you know, I made some changes. Look it over and see if you agree/have any changes to make based on these. I have created a [`main-v1`](https://github.com/scikit-hep/awkward/tree/main-v1) branch. If you think this is done, you can do the honors and merge it into `main` as a single commit.

(After that, PR #1666 will have to be closed and re-applied to `main` and possibly `main-v1`. I don't see how it can be easily merged with this, and anyway it's mostly generated by a pre-commit configuration.)

Final (?) compilation times (`pip install .`) on a Mac with 8 cores:

  * with v1 (old): 265.87s user 13.39s system 333% cpu 1:23.73 total
  * this PR (new): 86.52s user 7.11s system 371% cpu 25.186 total

Compiled binary sizes:

   * with v1 (old): 9.5M _ext.cpython-310-darwin.so, 1.3M libawkward-cpu-kernels.dylib, 6.1M libawkward.dylib
   * this PR (new): 1.1M _ext.cpython-310-darwin.so, 729K libawkward-cpu-kernels.dylib, 921K libawkward.dylib

Loading (`import awkward`) times:

   * with v1 (old): 95 ms
   * this PR (new): 83 ms (I guess it wasn't dominated by loading compiled symbols)

Anyway, awesome work! Disentangling v1 from v2 was not trivial!

src/awkward/_connect/avro.py
```diff
@@ -116,9 +116,9 @@ def __init__(self, file, limit_entries, debug_forth=False):
 
         for elem in form_keys:
             if "offsets" in elem:
-                container[elem] = machine.output_Index64(elem)
+                container[elem] = machine.output(elem)
             else:
-                container[elem] = machine.output_NumpyArray(elem)
+                container[elem] = machine.output(elem)
```
ianna: Is it intentional change? If yes, it looks like the `if` is not needed here

include/awkward/python/util.h
```diff
@@ -20,10 +20,8 @@
 /// reference count is held for all C++ references to the object.
 ///
 /// See also
-///   - kernel::array_deleter, which frees array buffers, rather
+///   - array_deleter, which frees array buffers, rather
```
ianna, line 23: Perhaps, the commented out lines on line 30 and 36 should be removed:
`    // std::cout << "pyobject INCREF of " << pyobj_ << std::endl;`

src/awkward/_connect/numexpr.py
```diff
@@ -1,6 +1,6 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
-# v2: replace with src/awkward/_v2/_connect/numexpr.py
+# v2: replace with src/awkward/_connect/numexpr.py
```
ianna: Should it be removed?

src/awkward/_connect/numexpr.py
```diff
@@ -124,42 +124,40 @@ def re_evaluate(local_dict=None):
...
-    return ak._v2._util.wrap(out[0], behavior)
+    return ak._util.wrap(out[0], behavior)
 
 
-# ak._v2._connect.numexpr = types.ModuleType("numexpr")
-# ak._v2._connect.numexpr.evaluate = evaluate
-# ak._v2._connect.numexpr.re_evaluate = re_evaluate
+# ak._connect.numexpr = types.ModuleType("numexpr")
+# ak._connect.numexpr.evaluate = evaluate
+# ak._connect.numexpr.re_evaluate = re_evaluate
```
ianna: Should it be enabled now that it is not conflicting with v1?

src/awkward/_connect/numpy.py
```diff
@@ -262,13 +247,13 @@ def unary_action(layout, **ignore):
...
     else:
-        out = ak._v2._broadcasting.broadcast_and_apply(
+        out = ak._broadcasting.broadcast_and_apply(
             inputs, action, behavior, allow_records=False, function_name=ufunc.__name__
         )
         assert isinstance(out, tuple) and len(out) == 1
         out = out[0]
 
-    return ak._v2._util.wrap(out, behavior)
+    return ak._util.wrap(out, behavior)
```
ianna, line 256: Should the lines 259-335 be enabled?

agoose77:
I'm happy to merge this. I won't update the docs, because we need to do more than just replace `ak.layout` in any documented examples of the mid-layer anyway.

There may well be some bugs / quirks in this branch, but I think at this point we pass the test suite, and its had 1/2 other pairs of eyes on it, so I am happy to iterate in main for now.

ianna (commented):
Great job! Just a few minor comments. Please, check

src/awkward/_connect/avro.py
agoose77: Haha, looks like I was tunnel visioned on `ForthMachine` when I made _that_ change! Good catch.

src/awkward/_connect/numexpr.py
agoose77: It looks like this was a shim to make a subset of the numexpr module public. Now we just have `numexpr.py` instead of `_numexpr.py`, so I *think* I can remove this.

src/awkward/_connect/numpy.py
agoose77: I've left these disabled as @ioanaif is going to be looking at these IIRC.

include/awkward/python/util.h
agoose77: @ianna I won't merge this whilst you're reviewing still! So, don't feel under any time pressure to do it *right now*

agoose77:
Closed by #1721

PR #1691, opened 2022-09-10
test: skip individual np.float16/128 tests because different systems have different combinations.

jpivarski:
My new Mac's NumPy has a `np.float16` but no `np.float128`. For whatever reason. Anyway, the old logic skipped the whole test "`if windows`", and that's not granular enough.


jpivarski:
Tests don't run on my new computer without this. It's important to include in 1.10.0rc4.

PR #1695, opened 2022-09-13
fix: ensure that `__copy__` and `__deepcopy__` are enabled.

### LINKED ISSUE #1694: Shallow copy is slow, seems to spend a lot of time in repr
### Version of Awkward Array

1.10.0rc3

During the PR adding awkward array support into AnnData, I noticed the test suite was taking much longer than usual. Trying to track down the culprit, I found the line where we make shallow copies of an awkward array. Under this largely seemed to be code related to making `repr` strings.

Checking if this is just us:

Making a minimal reproducible example:

```python
import awkward._v2 as ak
import numpy as np
from copy import copy, deepcopy

counts = np.random.randint(1, 1000, size=(10000))
data = np.random.randint(0, 100, size=counts.sum())
awk = ak.unflatten(data, counts)


%%prun -s cumtime
for _ in range(1000):
    copy(awk)
```

```
         3544003 function calls (3382003 primitive calls) in 3.619 seconds

   Ordered by: cumulative time

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000    3.619    3.619 {built-in method builtins.exec}
        1    0.002    0.002    3.619    3.619 <string>:1(<module>)
     1000    0.003    0.000    3.617    0.004 copy.py:66(copy)
     5000    0.041    0.000    3.264    0.001 _util.py:210(__init__)
     1000    0.002    0.000    3.078    0.003 {method '__reduce_ex__' of 'object' objects}
     1000    0.003    0.000    3.076    0.003 highlevel.py:1377(__getstate__)
28000/24000    0.056    0.000    2.881    0.000 _util.py:138(format_argument)
     4000    0.044    0.000    2.026    0.001 highlevel.py:1177(_repr)
     4000    0.005    0.000    1.699    0.000 _prettyprint.py:209(valuestr)
78000/26000    0.186    0.000    1.697    0.000 highlevel.py:1113(__setattr__)
40000/4000    0.182    0.000    1.692    0.000 _prettyprint.py:69(valuestr_horiz)
    12000    0.028    0.000    1.595    0.000 highlevel.py:187(__init__)
     1000    0.004    0.000    1.548    0.002 ak_to_buffers.py:9(to_buffers)
     1000    0.004    0.000    1.525    0.002 ak_packed.py:8(packed)
     2000    0.007    0.000    1.493    0.001 ak_to_layout.py:11(to_layout)
    78000    1.398    0.000    1.398    0.000 {built-in method builtins.dir}
    36000    0.051    0.000    1.318    0.000 _prettyprint.py:41(get_at)
    10000    0.017    0.000    1.070    0.000 _util.py:770(wrap)
```

I would expect making a shallow copy of an awkward array to be quite fast. I also wouldn't expect pretty printing methods to be called.

jpivarski:


agoose77:
We might want to also make proper use of the `memo` dict here. Right now, a layout which re-uses the same buffers will (I assume) end up with multiple copies of these buffers. Just trying to figure out a nice interface.

jpivarski:
Thanks for the clean-ups! I checked each one and they are definitely improvements, though d02d9279c3dcb1ee2cc2cd82d19427ac992ad2c9 is not actually related to this PR.

Immutable objects that take children in their constructor arguments (without [delayed initialization](https://stackoverflow.com/a/7486796/1623645)) can't have cycles, and we've made "layouts are trees" assumptions elsewhere, too. Nevertheless, I saw the code in the `copy` module that makes use of the `memo` argument before calling our `__deepcopy__`, so if we also do it, it would be redundant.

As you can see here:

https://github.com/python/cpython/blob/4781535a5796838fc4ce88e6e669e8907e426685/Lib/copy.py#L128-L178

Python's `copy.deepcopy` fills `memo` with `id(our_object)` (key) pointing to the copy that we return (value), unless it finds `id(our_object)` already in the `memo`, in which case it returns that. So all we need to do is make sure that the `memo` doesn't get lost (replaced with a new empty dict) during the recursive walk.

Since your last commit was 2 hours ago, I can safely merge this.

agoose77:
> As you can see here:

Yes, I was working off memory at the time of writing that comment. After checking the spec, I then went on to open #1697 when concerned about reference _cycles_ (via mutable parameters), but subsequently recalled that we don't allow these, so it's a none-starter :)

PR #1697, opened 2022-09-14
refactor: add copy helpers 

agoose77:


agoose77:
Given that we just don't allow cycles in `parameters`, this is not required.

jpivarski:
Okay, so that was the motivation for this!

This was looking to me like an abstraction of `__deepcopy__` from a computational implementation, "deepcopy does such-and-such an algorithm," to a declarative specification: "these attributes are deep and those are shallow, a single algorithm for all layouts uses this fact." I didn't think this abstraction was necessary if it was _only_ for `__deepcopy__`, but if there was some other reason for it, then maybe it would help to have this information around.

But if your goal was just for `__deepcopy__` and you don't have any other operations in mind, then I agree that this is probably overkill.

PR #1698, opened 2022-09-14
fix: pass memo to deepcopy

agoose77:
@jpivarski whoops, somehow I missed this on your PR.

### DIFF
--- a/src/awkward/_v2/highlevel.py
+++ b/src/awkward/_v2/highlevel.py
@@ -1423,7 +1423,8 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
     def __deepcopy__(self, memo):
         return Array(
-            copy.deepcopy(self._layout, memo), behavior=copy.deepcopy(self._behavior)
+            copy.deepcopy(self._layout, memo),
+            behavior=copy.deepcopy(self._behavior, memo),
         )
 
     def __bool__(self):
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Right—I can't think of a legitimate case in which an ak.Array instance is inside of the `behavior` dict, but this is formally correct and if such a situation does show up in the future, this implementation fixes a potential bug.

I'm setting up auto-merge.

agoose77:
NB - this PR is now placing more importance on `behavior` being None if the user wants `ak.behavior` - now we deepcopy this, which would be expensive if we are currently passing around `ak.behavior` explicitly.

jpivarski:
Presumably, if someone actually says

```python
my_new_array = copy.deepcopy(my_awk_array)
```

then they don't want `my_new_array` to share data with `my_awk_array`.

**But**, behaviors contain class objects and functions. We don't want them to be copied.

```python
>>> import copy
>>> def some_function(x): return x
... 
>>> class SomeClass: pass
... 
>>> behavior = {"function": some_function, "class": SomeClass, some_function: "as key"}
>>> copied_behavior = copy.deepcopy(behavior)
>>> behavior["function"] is copied_behavior["function"]
True
>>> behavior["class"] is copied_behavior["class"]
True
>>> list(behavior.keys())[2]
<function some_function at 0x1057c7e20>
>>> list(behavior.keys())[2] is list(copied_behavior.keys())[2]
True
```

Okay: they're not.

agoose77:
> Okay: they're not.

Yes, right now this would only be an issue if a user is passing an instance _method_ (or other non-atomically copied value) into `ak.behavior`, e.g.

```python
class SpecialAdder:
    def add(self, x, y):
        ...

ak.behavior[np.add, "point", "point"] = SpecialAdder().add
```

Whilst I don't anticipate this, I think it's worth discussing it up-front in case we this these kinds of issues later.

PR #1699, opened 2022-09-14
feat: raise Error for Record.__setattr__

agoose77:


jpivarski (approved):
I see; this is parallel to

https://github.com/scikit-hep/awkward/blob/b5545a8f07dfdd07dbb030a5ef9e08bb8f3ffcee/src/awkward/_v2/highlevel.py#L1138-L1151

It's the right thing to do.

PR #1700, opened 2022-09-14
perf: improve Array initialisation performance

agoose77:
- [x] follow up #1699 here

### DIFF
--- a/src/awkward/_v2/highlevel.py
+++ b/src/awkward/_v2/highlevel.py
@@ -1095,7 +1095,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         Note that while fields can be accessed as attributes, they cannot be
         *assigned* as attributes. See #ak.Array.__setitem__ for more.
         """
-        if where in dir(type(self)):
+        if hasattr(type(self), where):
             return super().__getattribute__(where)
         else:
             if where in self._layout.fields:
@@ -1135,7 +1135,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
         to add or modify a field.
         """
-        if name in dir(type(self)) or name.startswith("_"):
+        if name.startswith("_") or hasattr(type(self), name):
             super().__setattr__(name, value)
         elif name in self._layout.fields:
             raise ak._v2._util.error(
@@ -1816,7 +1816,7 @@ class Record(NDArrayOperatorsMixin):
            * the field name is not a valid Python identifier or is a Python
              keyword.
         """
-        if where in dir(type(self)):
+        if hasattr(type(self), where):
             return super().__getattribute__(where)
         else:
             if where in self._layout.fields:
--- a/src/awkward/_v2/highlevel.py
+++ b/src/awkward/_v2/highlevel.py
@@ -1792,9 +1792,9 @@ class Record(NDArrayOperatorsMixin):
         """
         Whenever possible, fields can be accessed as attributes.
 
-        For example, the fields of an `record` like
+        For example, the fields of a record like
 
-            ak.Record({"x": 1.1, "y": [2, 2], "z": "three"})
+            record = ak.Record({"x": 1.1, "y": [2, 2], "z": "three"})
 
         can be accessed as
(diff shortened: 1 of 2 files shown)


jpivarski:
Wait—are we sure that

```python
        if where in dir(type(self)):
```

is equivalent to

```python
        if hasattr(type(self), where):
```

The record fields are accessible via get-attribute (like `my_array.my_physics_column`). We want `__getattr__`, `__setattr__`, `__dir__` to agree on the set of attributes as including these record fields. If they disagree, it gets confusing.

Is this PR breaking that relationship?

agoose77:
> Wait—are we sure that
> 
> ```python
>         if where in dir(type(self)):
> ```
> 
> is equivalent to
> 
> ```python
>         if hasattr(type(self), where):
> ```

Well, this is the tricky part. `dir` is only supposed to show _relevant_ attributes rather than a comprehensive set (although, oftentimes it uses the default impl which is comprehensive). NB it's also the `type.__dir__` that we look at here, not `self.__dir__`

This code allows the user to set existing _type_-declared fields, or private attributes. Otherwise, they need to be trying to set an existing field name. This seems like the right approach; we want _only_ fields to be immutable, and we don't want fields to shadow existing Array attributes (otherwise third-party consumers of arrays might run into trouble, e.g. if the user adds a `fields` field)! 

So, the specification defined here is:
- allow users to write to the private attributes (`_XXX`), which are prohibited by convention for external use
- allow users to write to public type-level attributes (including attributes that override the `ak.Array` attributes, e.g. if a user defined their own behavior with `def fields`). Users should not do this, rather than us trying to validate things?
- prevent users from writing any other fields

So, we don't go as far as preventing records from containing private-shadowing fields (e.g. `"_layout_`) or behaviours overriding the base `ak.Array` type interface.

jpivarski:
We were careful to make sure that it doesn't overshadow:

```python
>>> import awkward as ak
>>> array = ak._v2.Array([{"fields": 1, "field": 2}])
>>> array.fields
['fields', 'field']
>>> array.field
<Array [2] type='1 * int64'>
```

This is in the order of tests in

https://github.com/scikit-hep/awkward/blob/b5545a8f07dfdd07dbb030a5ef9e08bb8f3ffcee/src/awkward/_v2/highlevel.py#L1098-L1112

(We're taking advantage of the difference between [`__getattr__`](https://docs.python.org/3/reference/datamodel.html#object.__getattr__) and [`__getattribute__`](https://docs.python.org/3/reference/datamodel.html#object.__getattribute__), cf [here](https://stackoverflow.com/q/3278077/1623645).)

agoose77:
> We were careful to make sure that it doesn't overshadow:
> 
> ```python
> >>> import awkward as ak
> >>> array = ak._v2.Array([{"fields": 1, "field": 2}])
> >>> array.fields
> ['fields', 'field']
> >>> array.field
> <Array [2] type='1 * int64'>
> ```
> 
> This is in the order of tests in

Yes, indeed. What I mean here is that we don't do any validation at any stage to prevent this, we just implement the get/set behaviour such that users can't break anything. That means in this context, setting an attribute always succeeds if it's the name of an `ak.Array`-"approved" attribute. This means that users could still confuse themselves if they create a "bad" record-array, but unless we validate early, we just need to use docs for this.

jpivarski:
Okay, we're in agreement about what it's supposed to do. When the dust settles, let's just make sure that all of your bullet points in https://github.com/scikit-hep/awkward/pull/1700#issuecomment-1246793713 are still satisfied.

jpivarski (approved):
The tests check all the cases I was worried about. Someone might want to set

```python
record.x = 10
```

expecting the `10` to be broadcasted over the array, but the error message is right in saying that they have to

```pyhthon
record["x"] = 10
```

It's also correct that Array and Record subclasses can override these.

jpivarski:
So this is an approval; go ahead and merge!

If I don't hear back, I'll "update branch" and "enable auto-merge (squash)", to do extra testing and give about 10 minutes to intercept, if you have something else to add.

PR #1708, opened 2022-09-20
fix: support non-hashable parameter values in broadcasting

### LINKED ISSUE #1707: Intersecting parameters introduced "unhashable type" bug.
### Version of Awkward Array

1.10.0

PR #1679 added parameter-carrying to broadcasting, but if the parameter values are dicts, rather something hashable, like strings, the implementation fails.

```python
>>> import awkward._v2 as ak
>>> array = ak.with_parameter(ak.Array([[1, 2], [], [3]]), "attrs", "hashable")
>>> array
<Array [[1, 2], [], [3]] type='3 * [var * int64, parameters={"attrs": "hash...'>
>>> array + array
<Array [[2, 4], [], [6]] type='3 * [var * int64, parameters={"attrs": "hash...'>
```

versus

```python
>>> array = ak.with_parameter(ak.Array([[1, 2], [], [3]]), "attrs", {"not": "hashable"})
>>> array
<Array [[1, 2], [], [3]] type='3 * [var * int64, parameters={"attrs": {"not...'>
>>> array + array
  File "/Users/jpivarski/mambaforge/lib/python3.10/site-packages/numpy/lib/mixins.py", line 21, in func
    return ufunc(self, other)
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/highlevel.py", line 1339, in __array_ufunc__
    return ak._v2._connect.numpy.array_ufunc(ufunc, method, inputs, kwargs)
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/_connect/numpy.py", line 278, in array_ufunc
    out = ak._v2._broadcasting.broadcast_and_apply(
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/_broadcasting.py", line 1019, in broadcast_and_apply
    out = apply_step(
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/_broadcasting.py", line 998, in apply_step
    return continuation()
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/_broadcasting.py", line 720, in continuation
    outcontent = apply_step(
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/_broadcasting.py", line 409, in apply_step
    parameters_factory = parameters_factory_impl(inputs)
  File "/Users/jpivarski/irishep/awkward/awkward/_v2/_broadcasting.py", line 277, in _intersection_parameters_factory
    functools.reduce(operator.and_, list(parameters_to_intersect))
TypeError: unhashable type: 'dict'
```

It fails in the `functools.reduce` of

https://github.com/scikit-hep/awkward/blob/ca1b9d7835570af8a11fe7af61a7c53d095a80e2/src/awkward/_v2/_broadcasting.py#L262-L276

in the hashable case, the `parameters_to_intersect` is

```python
[dict_items([('attrs', 'hashable')]), dict_items([('attrs', 'hashable')])]
```

and in the non-hashable case, it's

```python
[dict_items([('attrs', {'not': 'hashable'})]), dict_items([('attrs', {'not': 'hashable'})])]
```

The `operator.and_` is a bitwise `&` operator—I doubt that's what you want there. I think this can be done more simply in a for loop.

Also, possibly related, we're losing parameters in simple arrays:

```python
>>> array = ak.with_parameter(ak.Array([1, 2, 3]), "attrs", {"not": "hashable"})
>>> array
<Array [1, 2, 3] type='3 * int64[parameters={"attrs": {"not": "hashable"}}]'>
>>> array + array
<Array [2, 4, 6] type='3 * int64'>
```

In this case, it reaches the `break` and therefore skips the `reduce`. (Why would `_parameters_is_empty` be true for this array that clearly has parameters? I think the first step in broadcasting creates a RegularArray around this NumpyArray—it doesn't collapse the RegularArray + NumpyArray into a multidimensional NumpyArray, and lose the parameter, does it?)

Discovered by @philippemiron.

agoose77:
Fixes #1707

In #1707 we were using the nice set-like feature of Python dict views (keys, items, values) to determine the intersection of parameters dictionaries. However, given that it is possible for users to store JSON serialisable content in `parameters`, including dictionaries and lists, we cannot assume that the values of `parameters` are hashable. 

This PR replaces the set-intersection reduction with a dictionary-equality reduction.


agoose77:
## Parameter coalescence
In `numpy.py`, we have this call to `RegularArray.maybe_toNumpyArray()`

https://github.com/scikit-hep/awkward/blob/ca1b9d7835570af8a11fe7af61a7c53d095a80e2/src/awkward/_v2/_connect/numpy.py#L147

Ultimately, in the case that a `RegularArray` contains a 1D `NumpyArray` with parameters, `RegularArray.maybe_toNumpyArray()` is promoting the parameters from the inner 1D `NumpyArray` layout to an outer 2D `NumpyArray` node. This means that the `_getitem_nothing()` call that finalises the broadcast unpacking stage strips the parameters, and we lose them.

why do we do call `maybe_toNumpyArray()` here? Is it just a short-circuit to stop recursion once we have all (maybe ND) Numpy arrays? I wrote out an answer of how to fix this, but actually the simplest fix is to just keep recursing until we end up at the leaf 1D `NumpyArray` nodes, which maintains the tree structure and its parameters.

I've added a tentative fix in d47df3d

## Custom ufuncs
Incidentally, I also observed that we don't test for custom ufuncs until after checking for NumPy arrays, so `RegularArray`s that are strings don't behave properly:
```python
>>> array = ak.to_regular(ak.Array(["this", "that"]))
>>> print(array + array)
[[232, 208, 210, 230], [232, 208, 194, 232]]
```

I'd also want to fix this here too. Note that the fix in d47df3d also fixes this issue too.

Some additional questions that this spawned:
- why do we overload `__add__` on the `char` behaviour instead of using a NumPy ufunc overload?
- If we support adding chars to chars, why don't we support adding string arrays to string arrays via ufunc? Is it just where we draw the line w.r.t string handling?

jpivarski (approved):
This is a good resolution to part 1 of #1707, so I'll merge this and move part 2 out to a new PR: #1710.

jpivarski:
> Some additional questions that this spawned:
> 
> * why do we overload `__add__` on the `char` behaviour instead of using a NumPy ufunc overload?
> * If we support adding chars to chars, why don't we support adding string arrays to string arrays via ufunc? Is it just where we draw the line w.r.t string handling?

There needs to be a

```python
ak.behavior[np.add, "string, "string"]
ak.behavior[np.add, "bytestring, "bytestring"]
```

overload here: at least to prohibit character addition (i.e. add `1` to `"HAL"` to get `"IBM"`). That's not how we conceptualize strings anymore. Eventually, `np.add` should be overloaded for strings and bytestrings to do concatenation, since that's what people _do_ expect nowadays. Most of the others, like `np.log`, should just be prohibited. We might want to apply a blanket prohibition on ufuncs on strings, then poke holes in it by adding features like "`+` means concatenation".

agoose77:
So, I'm taking your reply to mean that we should long-term replace the behaviour class magic methods with overloads of `np.ufunc`? We can do that :)

jpivarski:
Yeah, I was referring to strings specifically, but it's also true that they have to be implemented as behaviors, rather than Python methods, so that they apply when the node in question is not the top-most.

Having `+` work for an array of strings but not an array of lists of strings would be surprising.

jpivarski:
Since this is merged, we should be able to delete the [agoose77/fix-broadcasting-non-hashable](https://github.com/scikit-hep/awkward/tree/agoose77/fix-broadcasting-non-hashable) branch, right?

agoose77:
Yes, I hadn't realised it was already merged when I pushed the new test!

PR #1712, opened 2022-09-21
test: ensure that parameters are kept for `NumpyArray`s under ufunc operations

agoose77:
This follows #1708 to extend the test suite.


jpivarski (approved):
Thanks! This is helpful, but since it only impacts tests, let's merge it after 1.10.1.

jpivarski:
1.10.1 is done, so this is ready to go!

PR #1715, opened 2022-09-22
fix: ensure `behavior` is taken from wrapped array in `ak.Array`

### LINKED ISSUE #1709: `ak.Array` does not copy the behaviour of a wrapped array
### Version of Awkward Array

1.10.0

See title!

agoose77:
Fixes #1709

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -198,6 +198,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
         elif isinstance(data, Array):
             layout = data._layout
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Perfect; because of the way that the `behavior_of` function is defined, an explicitly passed-in `behavior` overrides any behavior found on `data`, so this has the right priorities.

jpivarski:
I don't understand what went wrong in that one test (`no: command not found`?), but maybe it's intermittent. I've set this to auto-merge.

jpivarski:
And then un-set it when that didn't work. But I see now that it's #1722 that will fix this and the others.

PR #1716, opened 2022-09-22
test: remove `ak.behavior` usage in tests

### LINKED ISSUE #1687: Test suite is order-dependent
### Version of Awkward Array

main

I think nearly all of this is due to the use of `ak.behavior` instead of a local behavior context. We can try changing this and see if random orders still pass.

Then, we can add custom markers e.g. `slow` to order slow tests last.

agoose77:
Fixes #1687 (at least, should help!)


jpivarski (approved):
Out of curiosity, does this make it possible to run all the tests in parallel (pytest-xdist)?

Even if not, it's a step in the right direction.

agoose77:
> Out of curiosity, does this make it possible to run all the tests in parallel (pytest-xdist)?

I'm not sure if this was already possible without this PR, but I've been able to run `xdist` with 8 workers successfully!

jpivarski:
There was a point, not long in the past, when it wasn't possible; everything would break, quite impressively. When I first found out about pytest-xdist, I decided to try it, saw the mess, and then realized that we were making serial assumptions in the testing code.

The use of global `ak.behavior` _should_ break it. If parallel tests were possible before this PR, they would have been inadvisable. Dropping v1 tests might also have made a big difference.

But knowing that it works now is the bottom line!

agoose77:
Yes, I actually started this PR after recalling behaviours were one set of global state that I'd seen inhibiting parallelism. I never checked xdist against the original v2 only suite, but as you say, it's a win either way.

PR #1721, opened 2022-09-23
feat: replace v1 with v2 (preserve-history)

agoose77:
@ianna and @jpivarski, I didn't want to squash and merge #1690 into main, as it would destroy all of the v2 history. Whilst that's less catastrophic than in the case of v1, which has been around for much longer, I would rather keep it. That way, future bugs that are just porting errors will be easier to spot, as the history will inform us of this.

I've just refactored the commits in #1690 so that move operations are preserved. You can confirm that this PR does not change the result with
```bash
git diff 0c075e1f0d3f35a703d8a36ede0541b297dadae6 6e696b1f91030ffa6871b638d844fbcd76a9b914
```

If you're both happy with my changes in line with your reviews, then feel free to merge (via rebase and merge, or merge commit (NOT squash!))


src/awkward/contents/emptyarray.py
```diff
@@ -156,42 +156,40 @@ def _getitem_next(self, head, tail, advanced):
...
-            if ak._v2._util.isint(out):
+            if ak._util.isint(out):
                 return np.int64(out)
             else:
                 return out
         else:
-            out = ak._v2.index.Index64.empty(0, self._nplike)
-            return ak._v2.contents.numpyarray.NumpyArray(out, None, None, self._nplike)
+            out = ak.index.Index64.empty(0, self._nplike)
+            return ak.contents.numpyarray.NumpyArray(out, None, None, self._nplike)
```
ianna, line 185: I guess, eventually we want to use `ak.contents.NumpyArray` instead of `ak.contents.numpyarray.NumpyArray`. The same applies to other `*Array`s

ianna (approved):
Looks good to me! 

Note, I did not go through all deleted tests - I think we can rely on coverage to add more tests if anything is missing.

Please, see a comment inlined. I think, it would be nice to use either long or short form consistently. I think, it can be auto corrected after this PR is merged.

src/awkward/contents/emptyarray.py
jpivarski: Right, that would be a good thing to unify (later). I think we'll be able to take all of the short forms, e.g. `ak.contents.NumpyArray` instead of `ak.contents.numpyarray.NumpyArray`. The only reason I can think of for the long forms is if there's a cyclic reference and `awkward/contents/__init__.py` isn't fully evaluated yet, but that can only happen in module-level references.

I think there's something that I don't understand about Python's (or Python 3's) submodule imports because @henryiii said that it should be possible to make `awkward/__init__.py`'s imports order-independent. On another occasion, he said something about Vector being limited because Awkward wasn't using relative imports consistently. I think there's a rule I or we need to learn about that, but then it will be a mechanical or semi-mechanical process to apply it.

(This, by the way, might become its own issue; it doesn't affect PR #1721 dropping v1.)

src/awkward/__init__.py
```diff
@@ -1,56 +1,55 @@
...
-from awkward.behaviors.mixins import *
-from awkward.behaviors.string import *
-from awkward.behaviors.categorical import *
+import awkward.behaviors.categorical
+import awkward.behaviors.mixins
+import awkward.behaviors.string
+
+behavior = {}
+awkward.behaviors.string.register(behavior)  # noqa: F405 pylint: disable=E0602
+awkward.behaviors.categorical.register(behavior)  # noqa: F405 pylint: disable=E0602
```
jpivarski: PyLint says that we don't need to `disable=E0602` anymore.

jpivarski:
> I've just refactored the commits in #1690 so that move operations are preserved. You can confirm that this PR does not change the result with
> 
> ```shell
> git diff 0c075e1f0d3f35a703d8a36ede0541b297dadae6 6e696b1f91030ffa6871b638d844fbcd76a9b914
> ```

I'm unable to verify that these commits are the same, and I can't find 0c075e1f0d3f35a703d8a36ede0541b297dadae6 in this branch. (So `fatal: bad object 0c075e1f0d3f35a703d8a36ede0541b297dadae6`.)

I'm okay with closing #1690 and merging this PR into `main`, using the non-squashy method you have in mind. (I'm not quite gliterate enough to trust myself to do that, with all of these rebasings. I'd like you to press the button.)

I'd feel a little more secure if there is a https://github.com/scikit-hep/awkward/compare URL that shows the final #1690 commit is equal to one of these commits, and then I can look at the further diffs beyond it that are only here to be certain what happened after that. I just rechecked the last commit of #1690 (6e696b1f91030ffa6871b638d844fbcd76a9b914) and I'm on board up to that point.

jpivarski:
Followed your advice on Slack:

```
% git checkout agoose77/excise-v1-merge
% git pull 
remote: Enumerating objects: 5, done.
remote: Counting objects: 100% (5/5), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 5 (delta 3), reused 5 (delta 3), pack-reused 0
Unpacking objects: 100% (5/5), 510 bytes | 72.00 KiB/s, done.
From https://github.com/scikit-hep/awkward
   fc1427ab..78274500  agoose77/excise-v1-merge -> origin/agoose77/excise-v1-merge
Updating fc1427ab..78274500
Fast-forward
 src/awkward/__init__.py | 4 ++--
 1 file changed, 2 insertions(+), 2 deletions(-)
% git diff 78274500975cd337cb52c551bf39252075d70680 6e696b1f91030ffa6871b638d844fbcd76a9b914
diff --git a/src/awkward/__init__.py b/src/awkward/__init__.py
index 6c98b8af..27df449f 100644
--- a/src/awkward/__init__.py
+++ b/src/awkward/__init__.py
@@ -41,8 +41,8 @@ import awkward.behaviors.mixins
 import awkward.behaviors.string
 
 behavior = {}
-awkward.behaviors.string.register(behavior)  # noqa: F405
-awkward.behaviors.categorical.register(behavior)  # noqa: F405
+awkward.behaviors.string.register(behavior)  # noqa: F405 pylint: disable=E0602
+awkward.behaviors.categorical.register(behavior)  # noqa: F405 pylint: disable=E0602
 
 # operations
 from awkward.operations import *
```

This looks good! You have my approval to merge this into `main` using the method that you have in mind, to preserve the history but collapse the "drop v1" work into one or a few commits on the `main` branch.

Thanks again!

agoose77:
So, this PR has `HEAD` of 78274500975cd337cb52c551bf39252075d70680. The original and reviewed PR has HEAD 6e696b1f91030ffa6871b638d844fbcd76a9b914. So, the diff between two is
```bash
git diff 78274500975cd337cb52c551bf39252075d70680 6e696b1f91030ffa6871b638d844fbcd76a9b914
```
which yields
```diff
diff --git a/src/awkward/__init__.py b/src/awkward/__init__.py
index 6c98b8af..27df449f 100644
--- a/src/awkward/__init__.py
+++ b/src/awkward/__init__.py
@@ -41,8 +41,8 @@ import awkward.behaviors.mixins
 import awkward.behaviors.string

 behavior = {}
-awkward.behaviors.string.register(behavior)  # noqa: F405
-awkward.behaviors.categorical.register(behavior)  # noqa: F405
+awkward.behaviors.string.register(behavior)  # noqa: F405 pylint: disable=E0602
+awkward.behaviors.categorical.register(behavior)  # noqa: F405 pylint: disable=E0602

 # operations
 from awkward.operations import *
 ```

Therefore I'm happy to rebase-and-merge.

PR #1722, opened 2022-09-24
chore: use codecov action

agoose77:
This is the preferred means of uploading Codecov reports over the bash mechanism.


jpivarski (approved):
I guess this was the fix that was needed on the other two. (At least, it replaces a manual invocation of Codecov with an Action. Originally, this is how Codecov recommended to be invoked; I'm glad that they have an Action now.)

agoose77:
Yes, I suspected originally this is what they recommended; we've been around for a while ;)

PR #1723, opened 2022-09-24
chore: set `VERSION_INFO` contents to `2.0.0rc1`

agoose77:
@jpivarski I'm going to merge this without waiting for review, as it's a tiny change, and you've listed it in a todo list. Let me know if that's too presumptive!


jpivarski (approved):
I agree. I often do VERSION_INFO changes by directly committing to `main` or `main-v*` since it really can't break anything that the tests test for.

It's usually a step in making a release, in which I sanity-check the VERSION_INFO four or five times before pressing the button. That only works if the "update VERSION_INFO, tag release" action is single-threaded (i.e. one person doing it, not multiple). So that would need to be re-thought if we're both merging and releasing.

By the way, there was some problem I encountered in the past when setting the version number as part of a PR that did some other thing (such as the feature that motivated the release). I don't remember the reason at present, but it is a very good idea to keep the version number updates as separate commits on `main` than any other changes.

PR #1725, opened 2022-09-24
refactor!: disable iteration of records

### LINKED ISSUE #1582: Iterating over Records and `ak.from_iter`
Currently, `ak.Record.__iter__` returns an iterator over its _field names_, which may be what a user expects if a Record were a [Mapping/MutableMapping](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) like a dict, but it's not. It's probably for the best that it's not, because we don't want `keys`, `values`, `items` methods, particularly since we're trying to consolidate the nomenclature of field names as "fields", rather than "keys". And we especially don't want the non-vectorized `__eq__` and `__ne__` that Mapping has.

To throw a little more confusion in, `ak.Record.__len__` always returns 1 and not the number of fields (the length of what it iterates over). That's a weird enough choice to be considered a bug.

Since v2 is coming up and we can make little backward-incompatible changes like this, perhaps we should just ensure that v2 `ak.Record` has no `__iter__` and `__len__` methods. If they're really needed for something, we can be driven by the use-case as to whether `ak.Record` iteration should be over field names, field-value pairs, or something else. At the very least, `__len__` should return the length of the `__iter__` iterable.

Related to this is how `ak.Record` is treated by `ak.from_iter`: it naively iterates, which means it gets the field names:

```python
>>> import awkward as ak
>>> ak._v2.from_iter(ak._v2.Record({"a": 1, "b": 2}))
<Array ['a', 'b'] type='2 * string'>
>>> ak._v2.from_iter([ak._v2.Record({"a": 1, "b": 2})])
<Array [['a', 'b']] type='1 * var * string'>
```

Although iteration is a slow way to build an array and coming from a collection of `ak.Record` is suggestive that it's being used when a faster way is possible, the above is almost certainly not what the user had in mind. Changing v2 `ak.from_iter` to recognize `ak.Record` and build an equivalent thing is an improvement that would be independent of fixing the `ak.Record.__iter__` and `__len__` behavior, but both should be done.

**Edit:** also check `ArrayBuilder.append`; make sure it's consistent with `ak.from_iter` (it might be how `ak.from_iter` is implemented...)

agoose77:
Fixes #1582 by making `ak.highlevel.Record` non-iterable. Nothing else is changed, as the v2 `builder_from_iter` supports anything that implements a `to_list` or `tolist` method, such as `ak.Record` and `ak.record.Record`

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1657,8 +1657,8 @@ class Record(NDArrayOperatorsMixin):
     def _ipython_key_completions_(self):
         return self._layout.array.fields
 
-    def __iter__(self):
-        yield from self._layout.array.fields
+    # Disable iteration (and prevent old-style iteration from being used)
+    __iter__ = None
 
     @property
     def type(self):


jpivarski:
First, what's the exclamation point for? The "Lint PR" test accepts it as an okay prefix. From a Scheme context, I'd think it's allowing mutation (which all PRs do...).

jpivarski (approved):
Yes, we should prevent all iteration on Records. (I remember that conversation.)

The way you do this is new to me: I assume that `__iter__ = None` is different from not defining an `__iter__` method (Python apparently doesn't try to call `None` as a function). I suppose it prevents Python from assuming an iteration algorithm using `__getitem__`? But then, if Python is going to assume iteration using `__getitem__`, how does it know what keys/indexes it can pass to `__getitem__` are? With Record as it is currently defined and without `__iter__`, _does_ it try to iterate?

Huh—this is interesting:

```python
>>> class Something:
...     def __getitem__(self, index):
...         return index
... 
>>> for x in Something():
...     print(x)
... 
0
1
2
3
4
5
... (forever)
```

And `__iter__ = None` prevents that? Is the error message any better than "iteration on None is not allowed?" Maybe it would be better to define an `__iter__` that immediately raises a good error message.

agoose77:
The `__iter__ = None` does indeed disable the legacy iteration that Python otherwise uses. In all my time writing Python, I'd somehow never seen this before, but I suspect I nearly never implemented `__getitem__` and _not_ `__iter__`.

`None` is a documented special value to disable this behaviour:
https://docs.python.org/3/reference/datamodel.html#special-method-names:~:text=Setting%20a%20special%20method%20to%20None%20indicates%20that%20the%20corresponding%20operation%20is%20not%20available.%20For%20example%2C%20if%20a%20class%20sets%20__iter__()%20to%20None%2C%20the%20class%20is%20not%20iterable%2C%20so%20calling%20iter()%20on%20its%20instances%20will%20raise%20a%20TypeError%20(without%20falling%20back%20to%20__getitem__()).%202

>  Is the error message any better than "iteration on None is not allowed?"

Yes:
```python3
>>> iter(record)
Input In [7], in <cell line: 1>()
----> 1 iter(record)

TypeError: 'Record' object is not iterable
```

agoose77:
I'm going to auto-merge this, because you've approved the change, and I'm confident that you'd support following the Python docs in using `None` here :)

agoose77:
> First, what's the exclamation point for? The "Lint PR" test accepts it as an okay prefix. From a Scheme context, I'd think it's allowing mutation (which all PRs do...).

Sorry @jpivarski, I missed this comment. It implies that the commit / PR contains a breaking change. It's really a case where you're _removing_ a feature, rather than just refactoring internal code that won't really affect the user.

PR #1727, opened 2022-09-26
chore: test PyPy 3.10 in CI

agoose77:


agoose77:
Just wanted to see what the challenges were here, and off-the-bat, we don't want to pull in the runtime dependencies via `install .[dev]`.

PR #1731, opened 2022-09-26
refactor: replace Behavior with collections.ChainMap

agoose77:
This is a tiny refactor to replace a custom `Behavior` type with a stdlib `ChainMap`, courtesy of a helper function.


jpivarski (approved):
That's nice: I didn't know there was a standard library type for it. (And when this was written, we were still targeting Python 2.7, anyway.)

The value of this over making a new `dict` and `update`ing it to overlay the two mappings is that construction is faster, and that matters for the very large number of times this is built.

PR #183, opened 2020-03-19
Enforce a 79 character maximum on all lines, 72 on docstrings/comments.

jpivarski:
For the Python code, this is one item in the [PEP 8 style](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). But I'm going to do the same thing for C++.

It seems that most of the Python world agrees with this: [80 is a drop-off point in the distribution](https://jakevdp.github.io/blog/2017/11/09/exploring-line-lengths-in-python-packages/).

image

```
I don't find the arguments for this compelling; of course I know that we want t\
o put code side-by-side on our wide screens. Our text editors simply shouldn't \
be set up for horizontal scrolling: it should reflow with the window width. How\
ever, I don't mean that awful reflow that Atom does (and I can't seem to turn i\
t off), in which it breaks at word boundaries and tries to indent the remainder\
---that makes it even more unreadable. I mean it should reflow the way Emacs do\
es: the last character that fits in the window should turn into a "\" and the r\
est goes on the next line. Sure words will be broken, but we've all learned to \
read that from using terminals all these years, right? When lines are hard-wrap\
ped at some number of characters, they can't make full use of wider windows and\
 become completely unusable when the window is less than the fixed number of ch\
aracters, because soft reflow doesn't mix well with hard line breaks. But that'\
s why I checked the histogram above: if everyone _agrees_ with the same choice,\
 then it's not quite as terrible.
```

PR #1732, opened 2022-09-26
refactor: add type hints

agoose77:
A small PR to add partial type-hints to `awkward.contents.content`. This isn't expected to be perfect or even comprehensive; there are some types that I've left ambiguous because it's not immediately clear what they should be. The idea would be that we start tightening these hints once enough of the internals are typed that we can use `pyright` or `mypy`.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-refactor-add-type-hints/> once Read the Docs has finished building :hammer:


jpivarski (approved):
You're right: this is definitely a good start, and it demonstrates the power of progressive testing—we can keep going even though it's incomplete.

Does a type of "`int`" mean any `numbers.Integral`? Does it exclude booleans? Nearly every use of integers in Awkward Array would exclude booleans.

It should (eventually) be possible to define a type alias for the union of types that can be internal slices, after slice-normalization. That would handle a lot of these `head` and `tail` arguments (`head` is always one `InternalSlice` and `tail` is always `tuple[InternalSlice, ...]`).

agoose77:
`int` allows booleans, and we indeed probably don't want that in many places. I'll work those through before merging.

jpivarski:
The reason I asked about whether `int` includes `numbers.Integral` is because it's pretty common for a NumPy integer type, such as `np.int32` or `np.uint64`, to get passed into an argument that wants a generic integer, such as `axis`. (For instance, if it's computed from some formula that NumPy is involved in.) We don't want those integer types being excluded from these arguments, just as `MutableMapping` is better than `dict` in abstract interfaces.

I'd be happy to exclude booleans, since I think that would be an indication of a mistake, but typing has to go with the flow of the other packages that this gets typed with. If our `axis` signature insists that the argument is not a boolean but MyPy can't prove that it's not a boolean because some other Python package doesn't care, then it will fail typing for a bad reason (i.e. not help people find bugs, but be annoying). There's probably a community standard for accepting integer arguments that might be some integer type other than `int`, and we should go with that, whatever it is.

I'm going to make a wild guess that the appropriate spelling is "`numbers.Integral`", because that's the same sort of thing as "`MutableMapping`", at the same level of abstraction and visibility in the community (in the standard library).

agoose77:
@henryiii & @jpivarski do either of you have strong preferences for import styles in the context of type hints.

I've been moving towards 
```python
from package import module

def create_foo_bar():
    return module.FooBar()

def invoke_main(foo_bar):
    foo_bar.main()
```

in my personal projects, but in the context of type hints it sometimes feels a lot cleaner to read code that's annotated as
```python
from package.module import FooBar

def create_foo_bar() -> FooBar:
    return FooBar()

def invoke_main(foo_bar: FooBar):
    foo_bar.main()
```

There's no style guide on this that I can recall, and as much as I prefer the latter for readability, I often prefer the former for ease of locating where symbols come from. That said, I think oftentimes this is only a problem with poorly named symbols, so perhaps I'm making a stronger case for the latter.

henryiii:
For stuff in typing, I'm strongly in favor of importing the name:

```python
from typing import Sequence
```

This is standard practice, and it allows you to import from `typing_extensions` or `typing`, and some of the items there are actually temporary, with a more permanent home in the standard library elsewhere - like the one I showed! It really should be:

```python
from collections.abc import Sequence
```

Though you have to have `from __future__ import annotations` to use it (and you can only use it in type annotations) until you can require Python 3.9+.

If you have `typing.Stuff` everywhere this is harder to do. Typing is already verbose, I like keeping it as short as possible. `x: Sequence` is readable. `x: typing.Sequence` is too verbose, `x: _t.Sequence` is distracting and still verbose. (and, again, it's really `collections.abc.Sequence` since Python 3.9).

I've also seen `import typing as <something>`, which I like even less than just using `typing`. It only saves a few characters, adds an extra layer of mental burden (looking up the import name), and is different depending on what someone wanted to call it, making it harder to read code that is not yours. (I think I saw this in @HDembinski's IMinuit as `_typ` or something similar).

I usually leave non-types, like `typing.TYPE_CHECKING`, `typing.cast`, `typing.overload`, etc. directly accessed from the typing module and not imported from it.

henryiii:
For types not in typing, it's much harder. I still like the readability of not having the module, but you get a benefit from the module: you don't have to have that existing as a real object at this point in the import process if you only use it in type annotations (again, with the future import & Python 3.7+). They are also not standardize names like the items in `typing`. So it might depend, but full (or closer to full) names might be acceptable.

Also depends on if you are doing relative imports, which have benefits and drawbacks, but sometimes you can't due to circular imports, etc.

agoose77:
Thanks @henryiii 

We're using absolute imports here, and for the most part they're not that long. I'd temporarily ignored the performance aspect of not requiring the module body when only using typing. In our case, I suspect we'd want to fix those cases anyway (import cycles sometimes indicate a need to refactor)

My current take is to import the name from `typing`, and leave the rest to module-qualified imports. I'll probably make the change unless Jim wants to further the discussion

jpivarski:
Originally, I had been strongly in favor of fully qualified names everywhere, as in

```python
import module

def somewhere():
    module.submodule.subsubmodule.Class()
```

It's longer, but when reading the code in `somewhere`, you don't have to scroll up to the top of the file to find out where `Class` comes from. (For example, that made reading the Numba codebase hard: `SomeClass` is used, but where is it defined? There's a `from other_module import SomeClass`, but then in `other_module`, there's `from other_other_module import SomeClass`, and on and on.)

Fully qualified naming makes all instances of `Class` identifiable by a text-based search: `grep` works really well when a whole codebase adheres to this discipline.

But...

  1. That made the lines long. When I was working alone, I didn't mind because I don't have a problem with Emacs wrapping the line four or five times around the window (not at word boundaries). But other editors (Atom) wrapped lines at word boundaries, which is confusing in Python because of the semantic meaning of indentation. And then when I imposed the PEP8 line limit (#183), the many lines needed for most statements started to seriously hurt readability.
  2. Using really long `module.submodule.subsubmodule.Class` names in type annotations would further compound the issue, especially if any of those types are parameterized.
  3. Smart-enough IDEs should be able to find the definition or all references of a `from module import Class` name; it just needs some minimal knowledge of Python syntax to figure it out. It would be really nice to get that capability in a command-line tool, so that I can pipe the output to more `grep` and `awk`, but I've never looked for such a thing and perhaps I should.

The line-length and typing arguments in favor of `from module import Class` now outweigh the argument in favor of fully qualified names, especially if there's a definition-finding/reference-finding tool for Python codebases. The only constraint I'd add is that if two modules define a `Class` with the same name, it should be imported with enough context to disambiguate between the two: `subsubmodule.Class`. These classes were defined with the explicit intention of being referenced within their submodule; the submodule is part of the name. Here's an example:

```python
from awkward import nplike

nplike.of(some_array)
```

Importing a free-floating function named "`of`" would be problematic. Perhaps we should avoid defining function and class names this way in the future. And actually, with this specific example,

```python
from awkward import nplike
```

would itself be a problem because we use `nplike` as a variable and attribute all over the place. So I guess this one has to be

```python
import awkward as ak
import awkward.nplike   # we don't currently do this, but maybe should, to ensure order independence?

ak.nplike.of(some_array)
```

But avoiding definitions that force this would be good if we're moving away from fully qualified naming.

agoose77:
Well, bikeshedding renames is always fun! I'd say really `awkward.nplike` should be `awkward.nplikes`, as it contains many (just as we have `awkward.contents`). Some people prefer singular module names, so whatever people prefer.

OK. I think we're mostly on the same page then - line length / readability is the most important thing. At least, that's what I took from what you wrote.

jpivarski:
> OK. I think we're mostly on the same page then - line length / readability is the most important thing. At least, that's what I took from what you wrote.

Right: line length/readability is the most important thing, overriding what I've said in the past about fully qualified names (my view is changing).

> I'd say really `awkward.nplike` should be `awkward.nplikes`, as it contains many (just as we have `awkward.contents`).

That's a good argument, and I wouldn't mind if `awkward.nplike` became `awkward.nplikes`, in line with `awkward.contents`, `awkward.forms`, and `awkward.types` (which were all plural for the purpose of consistency).

henryiii:
>  It would be really nice to get that capability in a command-line tool

FYI, it's called semgrep: https://github.com/returntocorp/semgrep - I saw it mentioned in a presentation at PyCON.

agoose77:
> > It would be really nice to get that capability in a command-line tool
> 
> FYI, it's called semgrep: https://github.com/returntocorp/semgrep - I saw it mentioned in a presentation at PyCON.

You can retire for the rest of the year, this is fantastic!!

HDembinski:
> (I think I saw this in @HDembinski's IMinuit as _typ or something similar).

Yes, I ended up using _tp, because I want to hide the types. The type names pollute every module otherwise. I have seen people `del`eting them at the end of the module (I think I saw that in @henryiii libraries at some point), but I find that more ugly than prefixing with _tp. Henry, do you know a way to solve this?

All the types sometimes make the function signatures actually less readable. Have you seen the ArrayLike type from numpy.typing? It is huge. Things are going to get a little better when we do not have to use Union[...] and Optional[...] anymore, but still.

henryiii:
The way to solve it in Python 3.7+ is to declare your public interface with `__all__` and then add a simple `__dir__` function that returns it. This has always been how you declare your public interface, but the `__dir__` trick causes it to be respected `dir(...)` as well as tab completion the vanilla REPL as well as IPython. This is also the only way to "re-export" something from an internal module that mypy respects, as well. This also hides all your other imports, like numpy and various stdlib modules.

Yes, you've probably seen me `del` stuff, but this (in a Python 3.7+ world) is much better. I would actually be happy to get rid of `_internal` modules too with this, since we can now be explicit in what is public API. It's also nicer for readers, as you can see a list of what a module provides at the top of the module, rather than looking through the entire module.

Example (showing a bit of Python 3.7+ annotations too):

```python

from .somewhere_to_reexport import public_thing_2
from . import private_thing

__all__ = ["public_thing", "public_thing_2", "PublicType"]

def __dir__() -> list[str]:
    return __all__

def public_thing(x: bool) -> int | None:
    return 1 if x else None

class PublicType:
    pass
```

Simulated usage:
```PyCon
>>> import example_module
>>> print(*dir(example_module))
public_thing public_thing_2 PublicType
```

henryiii:
> All the types sometimes make the function signatures actually less readable. Have you seen the ArrayLike type from numpy.typing? It is huge.

But using it in a function signature isn't huge. It's just huge when MyPy prints it out on failures. Given the number of lines printed out by C++ code using long variants, I don't think this is solvable, but it doesn't impact the usage of signatures, only type errors. You can use TypeAliases to keep your signatures from ballooning. And while it does add characters, realize for someone reading your code not completely familiar with it this adds really valuable information. I think we just have to get used to the stuff after the colon being "checked documentation", which is basically what types are. You can add a trailing comma to ensure each item in on a separate line with Black. You'll notice NumPy style docstrings had one parameter per line, with the type included - and those are docstrings, intended _only_ for users to read! (though some IDEs were parsing them for types, like PyCharm - the issue was the types were sometimes wrong, because they were unchecked).

I have noticed most modern languages are following a similar pattern for types (`var: Type`) - Rust does this, Swift does this, Go doesn't put in the colon, but it's still a trailing type - even C++ has the `->` trailing type as an option. I've forgotten the name of the third "successor" language for C++ announced this year, but Carbon also uses `var: Type`, and Cpp2 also does `var: Type`.

HDembinski:
I looked at a signature with ArrayLike from numpy.typing in Python's builtin help(). It prints the whole definition that's several lines long.

That C++ templated types can look ugly, too, is not the point here.

agoose77:
> It prints the whole definition that's several lines long.

There's a lot of room for improvement with the current set of Python help tools (Python & IPython's super-help) to handle type hints. I'm hopeful that efforts like [papyri](https://github.com/jupyter/papyri) might tackle this kind of thing. It's not in immediate scope; papyri seems primarily concerned with generated docs, but in time I'd expect a `__doc__` transformer to fit into that machinery.

henryiii:
Ahh, that looks terrible:

```PyCon
>>> from numpy.typing import ArrayLike
>>> def f(x: ArrayLike) -> None:
...     pass
...
>>> help(f)
Help on function f in module __main__:

f(x: Union[numpy.__array_like._SupportsArray[numpy.dtype], numpy.__nested_sequence._NestedSequence[numpy.__array_like._SupportsArray[numpy.dtype]], bool, int, float, complex, str, bytes, numpy.__nested_sequence._NestedSequence[Union[bool, int, float, complex, str, bytes]]]) -> None
```

Unless, of course, you use `from __future__ import annotations`, like you should be doing in Python 3.7+ code:

```PyCon
>>> from __future__ import annotations
>>> from numpy.typing import ArrayLike
>>> def f(x: ArrayLike) -> None:
...     pass
...
>>> help(f)
Help on function f in module __main__:

f(x: 'ArrayLike') -> 'None'
```

:)

PR #1734, opened 2022-09-27
docs: update for v2

### LINKED ISSUE #1424: Awkward tutorials site says that Uproot 4 does not write jagged arrays to ROOT files yet
### Which documentation?

Tutorials site

https://awkward-array.org/how-to-convert-uproot.html#from-awkward-to-root-with-uproot-4

is wrong: Uproot 4 does support writing Awkward Arrays (only certain types, but that would always be true). It needs to be updated because it's giving people the wrong impression.

agoose77:
This PR updates the new docs to work in v2-only main:
- [x] Replace Markdown links to API-reference with Sphinx directives
- [x] Replace v1 code with v2
   - [x] Normalise `tolist` vs `to_list`
- [x] Fix download of WASM wheel on RTD  
- [x] Removes irrelevant v1 docs
- [x] Add PyPI icon
- [x] Change copyright to "Awkward Array development team"
- [x] Use GH releases for release notes
- [x] Make contributing guide point to CONTRIBUTING.md
- [x] Cleanup front-page cards
- [ ] Restore API reference ordering
   - [ ] Remove duplication on index by having subcategories in ToC?
- [ ] Mention records and unions in "what is an Awkward Array"
- [ ] Move "Differentiation with JAX" to specialised group?
- [ ] Add Conda (Forge) icon
- [ ] Address https://github.com/scikit-hep/awkward/pull/1734#issuecomment-1273813025

Closes #1424


agoose77:
https://awkward-array.readthedocs.io/en/agoose77-docs-update-for-v2/

.github/workflows/docs.yml
```diff
@@ -9,7 +9,7 @@ on:
     types: [released]
   workflow_dispatch:
 env:
-  PYTHON_VERSION: "3.10"
+  PYTHON_VERSION: "3.10.7"
```
agoose77, line 12: We want to use the same Python version between the wheel building and consumption

docs-sphinx/_toc.yml
```diff
@@ -28,8 +28,6 @@ subtrees:
                         title: "Python objects"
                       - file: user-guide/how-to-convert-json
                         title: "JSON"
-                      - file: user-guide/how-to-convert-uproot
```
agoose77, line 31: This should be done in the uproot docs, in my opinion.

docs-sphinx/_toc.yml
```diff
@@ -49,14 +47,10 @@ subtrees:
                         title: "Arrays of missing data"
                       - file: user-guide/how-to-create-strings
                         title: "Arrays of strings"
-                      - file: user-guide/how-to-create-lazy
```
agoose77, line 52: This is gone!

docs-sphinx/_toc.yml
```diff
@@ -49,14 +47,10 @@ subtrees:
...
-                        title: "Lazy arrays [todo]"
                       - file: user-guide/how-to-create-unflatten-group
                         title: "Unflattening and grouping [todo]"
                       - file: user-guide/how-to-create-arraybuilder
                         title: "ArrayBuilder (easy & general)"
-                      - file: user-guide/how-to-create-layoutbuilder
```
agoose77, line 58: This is also gone!

agoose77:
@jpivarski the goal with this PR is to merge into `docs`, and then a second PR (#1776) will merge `docs` into `main`.

docs-sphinx/_toc.yml
jpivarski: [It is](https://uproot.readthedocs.io/en/latest/basic.html#reading-a-tbranch-as-an-array), but we don't know where users will be coming from, which site they'll find first. We ought to at least link users to the Uproot documentation, to funnel them in from wherever they might be coming from to where the information is.

.github/workflows/docs.yml
jpivarski: Is this a number that will need to be updated repeatedly? Manually tracked between GitHub Actions and ReadTheDocs?

docs-sphinx/_toc.yml
jpivarski: @ManasviGoyal's documentation will appear somewhere, maybe in a different part of the documentation because it's for a different set of users (C++).

With the LayoutBuilder section gone, the text "(easy & general)" is no longer serving a function: it used to be the thing that distinguished ArrayBuilder from LayoutBuilder: ease/generality versus performance. Now the distinction is even larger: Python versus C++. (Note that @ianna will be working on a LayoutBuilder for Numba, so in general I should say "compiled.")

agoose77: Good point - these names should be changed.

.github/workflows/docs.yml
agoose77: This number need only be synchronised within *this* workflow so that we build on the same interpreter as consumes the wheel. RTD doesn't consume this wheel, so it can do whatever it wants :)

docs-sphinx/_toc.yml
agoose77: Maybe this page should just be a direct link to the uproot docs, then.

docs-sphinx/_toc.yml
agoose77: Hmm, I still think this applies Jim; array-builder is intuitive (just build a tree using a builder pattern) vs layouts, which are "advanced". Are we OK to leave this, and revisit it in the next wave of docs updates?

agoose77:
I've not added any ReST documents for the `ak._ext.fromjsonob[_schema]` methods, because I think they're probably an implementation detail that we don't need to expose (it's in the C++ docs).

jpivarski (approved):
I'm looking at the rendered documentation at [https://awkward-array.readthedocs.io/en/agoose77-docs-update-for-v2/index.html](https://awkward-array.readthedocs.io/en/agoose77-docs-update-for-v2/index.html). Some of my comments below might not be things that were changed in this PR.

But I approve of this PR. Everything below can be subsequent iterations.

I like the fact that all of the navigation agrees that the four main sections are "Getting started", "User guide", "API reference", and "Contributor's guide". It's both the top-bar navigation and the four panels on the front page, and the left-bar navigation is aligned with it: the left-bar only shows content from the guide you're on.

Experienced users will want a link to Awkward's GitHub repo on the front page. I've had that problem with other libraries; I sometimes end up clicking "Edit this page" to get to GitHub, even though I don't want to edit the page.

There seems to be two "first impressions" pages‒"What is an Awkward Array?" and "Getting started". It would be better if there's only one. People who just discover this without knowing anything about it will want to be told what to read first; it's better to not have options. It also irks me that these boxes don't line up:

image

All of the issues described above could be fixed by replacing the link to "What is an Awkward Array" with quick-click information ([GitHub](https://img.shields.io/badge/GitHub-scikit--hep%2Fawkward-lightgrey), [PyPI](https://img.shields.io/pypi/v/awkward), [conda-forge](https://img.shields.io/conda/vn/conda-forge/awkward), [license](https://img.shields.io/github/license/scikit-hep/awkward), and how-to-install):

image

Then the main "Getting started" page can _be_ the "What is an Awkward Array?" page:

image

(Note: those installation boxes are only fully visible if the window is _narrow enough_ to not have a right-bar. Maybe it would be better for them to not be boxes.)

Now the fact that the overview questions are by default collapsed is bringing a lot of value: it prevents the "What is an Awkward Array?" section from being hidden, where it can only be found by scrolling.

In the first, "What kind of data does Awkward Array handle?", there should be some mention of records and unions.

I really need to fix this image:



"Try it in your browser" is great! Having a section for "community tutorials" is also very good, since it gives us a place to recognize more. (Currently, it looks like all of those are ones I gave.)

As mentioned in an inline comment, there needs to be some link (at the level of the other "converting arrays") for ROOT, but it can point to the Uproot documentation.

"Differentiation using JAX" should probably go inside the "Specialized behavior" group. Or before it, as a high-level item alongside "Using arrays in Numba". (This hasn't been entirely figured out and there's some duplication: "Specialized behavior" has a subpage "In Numba".)

The copyright will need to be changed from "Jim Pivarski" to something like "IRIS-HEP", "Scikit-HEP", or "the Awkward Array development team." (I like the last best.) And the copyright dates to `f"2018‒{datetime.now().year}"`.

The "Release history" has turned into something that's not as useful. Now that the GitHub release notes list all of the PRs, they don't need to be separately generated as bullet points, and this page excludes a lot of releases. Actually, we don't need to have a release history page separate from GitHub's anyway. The top-bar link could just go to [https://github.com/scikit-hep/awkward/releases](https://github.com/scikit-hep/awkward/releases) or it doesn't even need to be in the top-bar; it can be one of the quick-link badges on the front page (and therefore nowhere else). Then the top-bar links would be exactly equal to the four panels on the front page, which further clarifies/simplifies things for visitors.

The "API reference" has lost the topical ordering of functions on the left-bar, and hidden functions have been accidentally included, which is something we talked about in the Google Meet. The separate listing on the first "API reference" page ("High-level data types", "Append-only data type", "Adding methods, overloading operators", ...) duplicates what's in the left-bar and it's getting out of date, which is the trouble with duplication. Maybe both of these problems could be solved at once by grouping functions in the left-bar using the categories defined on the first page—that is, put the left-bar items for "ak.Array" and "ak.Record" under a group called "High-level data types" and "ak.ArrayBuilder" under "Append-only data type". As long as those groups are not collapsed, like "Tutorials", "Main interface", "Detailed reference", etc. in the [Uproot documentation](https://uproot.readthedocs.io/en/latest/).

The kinds of things that would be useful to include in the body of the "API reference" front page are the "C++ documentation" (nice!), [Kernel interface and specification](https://awkward-array.readthedocs.io/en/agoose77-docs-update-for-v2/reference/generated/kernels.html), and [AwkwardForth virtual machines](https://awkward-array.readthedocs.io/en/agoose77-docs-update-for-v2/reference/awkwardforth.html). All three of these things should be called out as being low-level internals; users are most likely interested in the high-level functions in the left-bar.

The "Contributing Guide" can _be_ [CONTRIBUTING.md](https://github.com/scikit-hep/awkward/blob/main/CONTRIBUTING.md). If the other information on this (small) page is not in the Markdown, it will eventually get there. There should be a single place for onboarding new developers.

I just realized that there's an Octocat in the top-right that links to GitHub. Nevertheless, also having this much space

image

devoted to the top 5 things experts are looking for is not a waste of space on the front page.

On the whole, it's looking great! I'm excited about the new documentation!

docs-sphinx/_toc.yml
jpivarski: It's just a question of "easy & general" compared to what? If it's right next to a discussion of layouts and those are labeled "detailed & specialized", then maybe. But these two things, ArrayBuilder and layouts, are not really a pair to be contrasted the way ArrayBuilder and LayoutBuilder were.

In particular, that text was there to answer the question of why we need two 'Builders. ArrayBuilder and layouts are different enough in what they look like and how they work that the question wouldn't even come up.

When "easy & general" is not contrasting something, answering a question like why we have two 'Builders, then it sounds like an advertising slogan. Before, it was saying how these two similiar-looking things were differently good; now it just says why ArrayBuilder is good without reference to anything else, like an advert.

jpivarski:
> I've not added any ReST documents for the `ak._ext.fromjsonob[_schema]` methods, because I think they're probably an implementation detail that we don't need to expose (it's in the C++ docs).

Nothing in `ak._ext` has to be documented in the Python section; it's already documented in Doxygen. That means someone would have to dig to find it, but it's very much an implementation details, so that's okay.

agoose77:
@jpivarski will you take pushback on the badges here: 
image

We can (and should) have a PyPI icon in the topic bar next to the octocat, and we could have another for conda forge. It's becoming fairly standard, and it would keep the body of our front-page simpler.

jpivarski:
It's usually me who's pushing back on too many badges...



The information doesn't have to be expressed with badges, either. I agree that a PyPI icon next to the GitHub icon would be great. On the front page, I'd like to see some variant of the text "GitHub:scikit-hep/awkward" spelled out because that will make an impression in the visitor's brain, to help them find it again later. I had clicked through almost the entire site before I noticed the little Octocat in the corner. ("Standard" only works for visitors familiar with the standard. Others, like me, need more of a hint.)

Maybe it doesn't need to be on the front page, but the Octocat can be accompanied by the text "scikit-hep/awkward"? And for symmetry, the PyPI logo can be accompanied by the text "awkward"? (I don't know if that's technically possible.) Knowing that the source control address is "scikit-hep/awkward" and the pip/conda-forge installation address is "awkward" is something that might need to be typed into a terminal or added to a script. 

The current version number is not as important and it can be found on GitHub, so that doesn't need to be here.

Outside of academia, "Does it have a commercial-friendly license?" is often the first question anybody asks of an open-source product. But that doesn't need to be in the main text of the front page, since it's also on GitHub. It could also go next to the copyright message, if that's technically possible.

Then there's the "pip install awkward" and "conda install -c conda-forge awkward" text boxes. I really like how those text boxes have a copy-icon when the mouse hovers over them. If I'm starting up a new virtual machine and need to get it up to speed with packages, I want a very quick way to copy the install command. Putting them on the "Getting Started" page makes some sense, since installation is the first thing that you do when you're getting started, and then not again afterward (unless you're spinning up new virtual machines or environments all the time).

So I can be argued down to just

>  * GitHub: [scikit-hep/awkward](https://github.com/scikit-hep/awkward)
>  * PyPI: [pip install awkward](https://pypi.org/project/awkward/)
>  * Conda: [conda install -c conda-forge awkward](https://anaconda.org/conda-forge/awkward)

If they're links like this, there's usually a "copy link text" option in the context menu, so that's a very compact way of giving experts the convenience while not cluttering the page with stuff a beginner doesn't understand.

agoose77:
Current WIP for the index:
image

agoose77:
I'll merge this PR as-is, and then take the outstanding bullets to be a new PR on `main`.

jpivarski:
The boxes line up nicely.  `:)`  And I see that "Contributor's Guide" and "Release history" are both external links.

There's still the issue that there are two "I don't know anything yet; tell me what I need to know" pages: "What is an Awkward Array?" and "Getting started". How about if they're combined onto one page:



and the "What is an Awkward Array?" button that you have on the front page links to the section within that combined page? (I.e. link to the scroll-position, with `#` in the URL.) Then "What is an Awkward Array?" is highlighted but there also isn't any confusion between that and "Getting started", once there, the visitor will realize that they're just different scroll positions of the same page. That's a much easier realization to make than if they were separated by a click-through.

---

Also, if you're done with this branch, it can be deleted. (It wasn't auto-deleted for some reason.)

PR #1739, opened 2022-09-27
fix: Record is not in contents, but in record

ianna:
This is popped up during CUDA testing: `ak.contents.Record` is in `ak.record.Record`


agoose77:
Huh I remember grepping for these, that's odd. Nice catch!

agoose77:
It seems like only the `nplike` module had actual code that was trying to resolve `ak.contents.Record`. @ianna did you hit this at runtime today? If so, could you add a test that reproduces it? Clearly we're not covering these `nplike` codepaths at the moment.

ianna:
> It seems like only the `nplike` module had actual code that was trying to resolve `ak.contents.Record`. @ianna did you hit this at runtime today? If so, could you add a test that reproduces it? Clearly we're not covering these `nplike` codepaths at the moment.

Yes, we already have them: in tests-cuda

agoose77:
This just means that my CUDA tests weren't running locally, which is ... even worse than if we'd just missed a test case! :cry:

Thanks!

PR #1744, opened 2022-09-28
refactor: rename `ak.nplike` → `ak.nplikes` & `nplikes.of` →`nplikes.nplike_of`

### LINKED ISSUE #1741: Rename `ak.nplike` into `ak.nplikes`
>> I'd say really `awkward.nplike` should be `awkward.nplikes`, as it contains many (just as we have `awkward.contents`).
>
>That's a good argument, and I wouldn't mind if `awkward.nplike` became `awkward.nplikes`, in line with `awkward.contents`, `awkward.forms`, and `awkward.types` (which were all plural for the purpose of consistency).

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/issues/1732#issuecomment-1259979433_

agoose77:
## Changes
1. Rename `ak.nplike` to `ak.nplikes`
2. Rename `ak.nplikes.of` to `ak.nplikes.nplike_of`

## Motivation
In type-hinting situations, it would give us more flexability to be able to import `ak.nplikes` as `nplikes`. Currently, the `nplike` module name would clash with local-scoped `nplike` variables. This is easy to work around - just use `from awkward import nplike as nplikes`, however we'll be doing that a lot, so let's just clean this up now.

Furthermore, should we wish to import `nplike.of` as a free-standing function, the same applies. `nplike_for` is a more descriptive name than `nplike.of`, even though the latter is pretty snazzy.

Closes #1741

### DIFF
--- a/dev/generate-tests.py
+++ b/dev/generate-tests.py
@@ -674,7 +674,7 @@ def gencudakerneltests(specdict):
                 )
 
                 f.write(
-                    "import cupy\nimport pytest\n\nimport awkward as ak\nimport awkward._connect.cuda as ak_cu\n\ncupy_nplike = ak.nplike.Cupy.instance()\n\n"
+                    "import cupy\nimport pytest\n\nimport awkward as ak\nimport awkward._connect.cuda as ak_cu\n\ncupy_nplike = ak.nplikes.Cupy.instance()\n\n"
                 )
                 num = 1
                 if spec.tests == []:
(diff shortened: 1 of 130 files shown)


jpivarski (approved):
Great! (Although now it occurred to me that we have `behavior_of` and maybe a few other `of`'s, while this is `nplike_for`, but...

Ah, heck. I'll just change it to `nplike_of`.)

I'll let you counter-approve by not setting this to merge, but merge it if you agree about `nplike_for` → `nplike_of`.

PR #1743, opened 2022-09-28
fix: don't assume trailing `.` for module name in `is_XXX_buffer`

agoose77:
`ak.nplike.of` was not correctly identifying CuPy arrays, because it was testing `cls.__module__.startswith("cupy.")` instead of `cls.__module__.startswith("cupy")`. I've made this logic more robust by first partitioning the string by the first `.`.

Fixes #1742

PR #1746, opened 2022-09-28
fix: don't assume trailing `.` for module name in `is_XXX_buffer` (backport)

agoose77:
Backport of #1743 for main-v1

### DIFF
--- a/src/awkward/nplike.py
+++ b/src/awkward/nplike.py
@@ -870,15 +870,18 @@ def is_numpy_buffer(array):
 
 
 def is_cupy_buffer(array):
-    return type(array).__module__.startswith("cupy.")
+    module, _, suffix = type(array).__module__.partition(".")
+    return module == "cupy"
 
 
 def is_jax_buffer(array):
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Good! And I see that this backport is targeting `main-v1`, which is always something I've have to double-check when doing the same in Uproot.

We'll have to decide when it's time to do a bug-fix release for v1. If someone is explicitly asking for this (I didn't see it), we'd do it right away. If it's just "the right thing to do," then we'll do it after a time interval, to be sure to batch fixes and not use up our PyPI quota. Maybe October 14? (I'm making a reminder.)

kwamazibuko:
We'll have to decide when it's time to do a bug-fix release for v1. If someone is explicitly asking for this (I didn't see it), we'd do it right away. If it's just "the right thing to do," then we'll do it after a time interval, to be sure to batch fixes and not use up our PyPI quota. Maybe October 14? (I'm making a reminder.)

PR #1748, opened 2022-09-28
refactor: drop unused code from Form, rename serialisation methods

agoose77:
This is a preparatory PR to clean-up `ak.forms.Form`. It does two things:

- Remove dead code.
- Rename  `tolist`, `from_iter` to `to_dict`, `from_dict`

@jpivarski we still need `simplify_optiontype` as it is used in the Arrow deserialisation.

### DIFF
--- a/src/awkward/forms/__init__.py
+++ b/src/awkward/forms/__init__.py
@@ -3,7 +3,7 @@
 from awkward.forms.bitmaskedform import BitMaskedForm  # noqa: F401
 from awkward.forms.bytemaskedform import ByteMaskedForm  # noqa: F401
 from awkward.forms.emptyform import EmptyForm  # noqa: F401
-from awkward.forms.form import Form, from_iter, from_json  # noqa: F401
+from awkward.forms.form import Form, from_dict, from_json  # noqa: F401
 from awkward.forms.indexedform import IndexedForm  # noqa: F401
 from awkward.forms.indexedoptionform import IndexedOptionForm  # noqa: F401
 from awkward.forms.listform import ListForm  # noqa: F401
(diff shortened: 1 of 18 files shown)


agoose77:
@jpivarski following from our earlier discussion, does `to_string` still sit well with you? I'm coming back to the fact that this "solution" to avoid overlap is only helpful whilst we don't add a `to_json` to some other class in Awkward. I *do* think we need a `to_dict` - it's more accurate w.r.t what the form serialisation is doing, but I'm less convinced now of `to_string`. 

Particularly, `to_json` is helpful because we explicitly tell the caller that they're getting a JSON string. Whereas, `to_string` could feasibly return `repr(self)` and still be consistent with the name `to_string`.

jpivarski:
`to_json_str`/`from_json_str`? I agree about getting people exactly what they're getting.

`to_json_form`?

agoose77:
@jpivarski I know it's a bit of a pain, but are you happy to merge this latest version, which doesn't change `to/from_json`? That way we can come back to it once I tackle some of the bigger PRs - decision fatigue!

jpivarski (approved):
This definitely looks good. I keep looking at this and thinking that maybe "`to_json`" is the right name after all, despite the fact that arrays mean a different thing by the same name.

We _have_ settled on a wedge-shaped inheritance between *Array and *Form, so it's legal for *Array and *Form to have a method with the same name that do different things. It was a high priority to move them when we were considering a direct inheritance (Content is a subclass of Form), but we're not anymore.

Let's leave `to_json`/`from_json` on Forms as they are for now and only revisit it if it rankles you, me, or someone else with a strong opinion. One thing is consistent about our `to_json`/`from_json` methods, across all classes: they deal with strings, not dicts/lists in place of JSON strings.

PR #1751, opened 2022-09-29
ci: run Codecov for commits to main

### LINKED ISSUE #1745: Codecov isn't finding the correct `base` with "squash and merge"
### Version of Awkward Array

main

I've been noticing recently that Codecov estimates in changed coverage do not correspond to what's actually happening in the PR; namely, PRs that add no tests (e.g. those which add documentation or rename variables) still increase test coverage. 

It turns out, Codecov uses the commit history to find the base with uploaded coverage metadata. Because we use "squash and merge", and don't run any Codecov jobs in `main`, Codecov always finds an old commit (e.g. e6929460415d240f2e3aa5d3188a6bb02908094e) to use as the base. Our `build-test` workflow has a schedule, so Codecov is only ever at-most 1 month out of date. In the below figure, you can see that each new branch never finds a codecov commit until the first commit which has the `codecov` tag 

```mermaid
gitGraph
       commit tag: "codecov"
       commit
       branch feat-add-x
       checkout feat-add-x
       commit
       commit
       commit tag: "codecov"
       checkout main
       merge feat-add-x  tag: "no-codecov"
       branch feat-add-y
       checkout feat-add-y
       commit
       commit
       commit tag: "codecov"
       checkout main
       merge feat-add-y  tag: "no-codecov"
```

The simplest solution here is to run a special CI job upon the closure of a PR that uploads the codecov metadata: https://community.codecov.com/t/unable-to-determine-a-parent-commit-to-compare-against-in-base-branch-after-squash-and-merge/2480/20
However, my initial concern is that unless we also _require_ PRs to be up-to-date before merging, the merge would happen _after_ the CI had run Codecov, which could lead to out-of-sync coverage estimates.

A better solution might just be to run the build-test workflow for commits to main.

agoose77:
This PR enables `build-test.yml` and `tests.yml` for each commit to main. This addresses two problems:
1. main is currently only tested every ~30 days
2. Codecov doesn't find coverage data newer than every ~30 days

This is the "obvious" way to tackle both of these problems, but it is not necessarily the nicest for the environment.

So, there are other ways of tackling this:
- Run a special `codecov` workflow for each commit to main. Only run `build-test.yml` and `tests.yml` at a reduced frequency, e.g once every couple of days.
- Run a special `codecov` workflow for each commit to main. Also, run `tests.yml` for each commit to main.
- etc.

@jpivarski do you have a preference here?

Fixes #1745

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,6 +21,9 @@ on:
       - src/cuda-kernels
       - cuda-kernels
 
+  schedule:
+    - cron: 0 12 1 * *
+
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,9 +21,6 @@ on:
       - src/cuda-kernels
       - cuda-kernels
 
-  schedule:
-    - cron: 0 12 1 * *
-
   workflow_dispatch:
 
 concurrency:
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,10 +1,6 @@
 name: Build Test
 
 on:
-  push:
-    branches:
-      - main
-
   pull_request:
     paths-ignore:
       - README.md
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,6 +1,10 @@
 name: Build Test
 
 on:
+  push:
+    branches:
+      - main
+
   pull_request:
     paths-ignore:
       - README.md
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,72 @@
+name: Codecov
+
+on:
+  push:
+    branches:
+      - main
+    paths-ignore:
+      - README*.md
+      - VERSION_INFO
+      - .ci/*
+      - .readthedocs.yml
+      - include/*
+      - src/*
+      - docs-src/*
+      - docs-img/*
+      - docs-jupyter/*
+      - docs-doxygen/*
+      - docs-sphinx/*
+      - studies/*
+
+  workflow_dispatch:
+
+concurrency:
+  group: 'coverage-${{ github.head_ref || github.run_id }}'
+  cancel-in-progress: true
+
+jobs:
+  coverage:
+    runs-on: ubuntu-20.04
+
+    env:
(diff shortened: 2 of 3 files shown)


jpivarski (commented):
The motivation for the scheduled test of main and its low frequency was to keep the "tests" badge green, actually. We don't allow tests to fail on main (though that's only partially tested, due to the concurrency issues in always testing exactly what will be in the main branch after merging), so the badge would, in principle, always be green, anyway, but it was grey because of the lack of explicit tests. The 30 day frequency was to avoid spending too much of our resources on the aesthetics of a green badge.

To keep that badge non-gray, we'll need some CI tests to run on main occasionally. Doing it with every commit nearly duplicates the PR branch testing. Maybe it could be done with a smaller matrix? Maybe just one platform-Python version?

But doing both the tests and the Codecov on main is desirable.

agoose77:
Good, that settles it. 

It might actually be possible to check whether the merged PR used the same base commit as main. If we can do that, then we could, in those cases, assume that the coverage result would match the merged repo. Otherwise, we can run a simple single-matrix-element build (Linux py37) to compute the coverage for the merged commit.

If this is not possible, then we can just keep periodic full testing on main, and add a new per-commit single-matrix-element build as suggested in the previous paragraph.

The disadvantage of the single element case is that we might miss changes that are platform/Python version specific. We could choose random configurations such that we cover more matrix elements over the course of the month... but I see that being problematic for determinism reasons. At the end of the day, do we make per-platform changes more frequently than monthly? I suspect not.

agoose77:
I've settled on just running a single Codecov job for commits to main. Every month we'll run a full build, which will update our coverage baseline in cases where we made coverage-altering per-platform changes. I think these are probably rare enough tha t we don't need to go to the extra effort of testing a larger matrix, or determining whether we can reuse coverage results.

jpivarski (approved):
This new test does not run monthly, but adding 1 extra task upon PR-merger doesn't significantly increase resource usage, since we're running ~10 for every commit already. It looks like this will provide a Codecov baseline _and_ a quick test of main itself (as opposed to the last commit in the branch).

That seems like a good compromise between leaving actual-main untested (i.e. assuming that all PRs merge into something correct) and fully testing main on every PR merger, or requiring every PR to be up-to-date with main before merging _and_ requiring all of their tests to pass (which essentially serializes testing and merging).

Let's go with this!

PR #1752, opened 2022-09-29
refactor: split error-handling code from `_util`

agoose77:
`ak._util` has become something of a catch-all module. Besides making `_util` quite large, this also means that `_util` depends on many Awkward modules, which in turn depend upon `_util`. Currently, we resolve this by using the module object, which can break import cycles. However, this can be improved.

This PR splits out the error-handling code into an `_errors` private module. 

There is more work that can be done here, but to keep this PR small and mergeable, we can do that in another PR.

### DIFF
--- a/dev/flake8_awkward.py
+++ b/dev/flake8_awkward.py
@@ -19,10 +19,11 @@ class Visitor(ast.NodeVisitor):
     def visit_Raise(self, node):
         if isinstance(node.exc, ast.Call):
             if isinstance(node.exc.func, ast.Attribute):
-                if node.exc.func.attr in {"error", "indexerror"}:
+                if node.exc.func.attr in {"wrap_error", "index_error"}:
                     return
-            if node.exc.func.id in {"ImportError"}:
-                return
+            if isinstance(node.exc.func, ast.Name):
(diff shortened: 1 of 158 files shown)


jpivarski (approved):
Yes, _util is a catch-all (because naming is hard, and I wanted to have a low barrier to making helper functions whenever we needed one). But this _errors is a good, well-defined classification.

The intention was for _util to be a dependency of most submodules, but _util itself doesn't depend on them (i.e. very early in the dependency graph). It looks like _util depends on nplikes at module-level, but all the other accesses of `ak.*` are inside functions—in other words, delayed—and therefore not a dependency cycle.

Similarly, nplikes only depends on `ak.*` inside functions, no dependency cycles when the module loads.

agoose77:
Yes, to clarify; we don't have any "broken" import cycles, but that is managed by using the two-phase module initialisation feature (and, if this weren't possible, we could still import other names within functions). Ideally, we could declare our dependencies at the top of the file with explicit imports, i.e. as a DAG. It won't always be possible to perform this split without compromising the general structure and layout of Awkward, and we don't need to pursue that level of separation.

PR #1754, opened 2022-09-29
fix: kernel name in `IndexedArray`

### LINKED ISSUE #1753: `IndexedArray` looks for kernel with wrong name
### Version of Awkward Array

main

We should be asking for `"awkward_IndexedArray_fill_count"` instead of `"awkward_IndexedArray_fill_to64_count"`.

agoose77:
Fixes #1753

- Corrects kernel name
- Corrects index handling logic

### DIFF
--- a/kernel-specification.yml
+++ b/kernel-specification.yml
@@ -600,7 +600,7 @@ kernels:
     automatic-tests: true
     manual-tests: []
 
-  - name: awkward_IndexedArray_local_preparenext_64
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
Thanks for catching this. I was going to hold off on this until I got a chance to run it and verify that `IndexedArray.reverse_merge` is now executed, but I managed to get that information from Codecov.

Issue #1686, opened 2022-09-07
`UnionArray.simplify_uniontype()` loses parameters


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

```python
import awkward as ak

numpy = ak.nplike.Numpy.instance()

left = ak._v2.contents.NumpyArray(numpy.arange(10), parameters={"left": "other"})
right = ak._v2.contents.NumpyArray(numpy.arange(10), parameters={"right": "other"})

array = ak._v2.contents.UnionArray(
    ak._v2.index.Index8(numpy.array([0, 0, 0, 0, 1, 1, 1, 1], dtype="int8")),
    ak._v2.index.Index32(numpy.array([0, 1, 2, 3, 0, 1, 2, 3], dtype="int32")),
    [left, right],
    parameters={"some": "other"},
)

assert array.simplify_uniontype().parameter("some") is not None
```

jpivarski:
This one is still an issue, too, though it is now _forcing_ a simplification (thanks to #1939).

```python
>>> import awkward as ak
>>> import numpy as np
>>> left = ak.contents.NumpyArray(np.arange(10), parameters={"left": "leftie"})
>>> right = ak.contents.NumpyArray(np.arange(10), parameters={"right": "rightie"})
>>> array = ak.contents.UnionArray.simplified(
...     ak.index.Index8(np.array([0, 0, 0, 0, 1, 1, 1, 1], np.int8)),
...     ak.index.Index32(np.array([0, 1, 2, 3, 0, 1, 2, 3], np.int32)),
...     [left, right],
...     parameters={"some": "other"},
... )
>>> array.parameter("some")   # output is None, should be "other"
```

Issue #1685, opened 2022-09-07
`IndexedArray.project()` does not preserve parameters


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

See https://github.com/scikit-hep/awkward/blob/31f3afb6bd31949bb2014d3be4b86194a63a573c/src/awkward/_v2/contents/indexedarray.py#L296-L351

```python
import awkward as ak

numpy = ak.nplike.Numpy.instance()

layout = ak._v2.contents.IndexedArray(
    ak._v2.index.Index64(numpy.array([0, 1, 2, 3])),
    ak._v2.contents.NumpyArray(
        numpy.arange(10),
        parameters={"this": "that"}
    ),
    parameters={"some": "other"}
)

assert layout.project().parameter("some") is not None
```

jpivarski:
This is still an issue:

```python
>>> import awkward as ak
>>> import numpy as np
>>> layout = ak.contents.IndexedArray(
...     ak.index.Index64(np.array([0, 1, 2, 3], np.int64)),
...     ak.contents.NumpyArray(np.arange(10), parameters={"this": "that"}),
...     parameters={"some": "other"},
... )
>>> layout.project()
<NumpyArray dtype='int64' len='4'>
    <parameter name='this'>'that'</parameter>
    [0 1 2 3]
</NumpyArray>
```

I'll introduce a `merge_parameters` at the right place.

Issue #1041, opened 2021-07-30
`IndexedArray::mergeable` returns `false` incorrectly


agoose77:
Reproducer:
```python3
import awkward as ak
import numpy as np

x = ak.layout.NumpyArray(np.arange(10), None, {"__record__": "mm_pad"})

y = ak.layout.IndexedArray64(
    ak.layout.Index64(np.arange(10)),
    x
)

assert x.mergeable(y)
```

This seems to be caused by the following parameter check, which doesn't take into account the wrapped layout parameters:
https://github.com/scikit-hep/awkward-1.0/blob/585569b245d65078e0b7440de0b0baf6f2b67a4e/src/libawkward/array/IndexedArray.cpp#L1644-L1645

Is this intentional? Or do we want to take the `purelist_parameter`(s) such that one *could* make the `IndexedArrayOf` non mergeable, but bare depth-less contents would be mergeable?

C.f. `BitMaskedArray`, etc.

jpivarski:
This _was_ intentional. Loosening it from checking parameters to checking purelist_parameters seems to make sense in this case, since the IndexedArray should probably be invisible.

But in general, it shouldn't be _purelist_ parameters: the same parameters on different dimensions should count as distinct types. What would be ideal here would be to require all the parameters at this dimension of `self` to be equal to all the parameters at this dimension of `other`, and then that single set of parameters would be what is passed on to the merged output.

So in other words, a generalized `parameters_equal` method that

   * considers index nodes and option-type nodes to be invisible (IndexedArray, IndexedOptionArray, BitMaskedArray, ByteMaskedArray, UnmaskedArray); merging their parameters with their content's (parent wins in any conflicts)
   * considers UnionArrays to be invisible unless the different branches of the union have different parameters
   * considers lists (ListArray, ListOffsetArray, RegularArray) and RecordArrays to be opaque—does not descend.
   * VirtualArrays would be invisible, though it should go through Forms to ensure that VirtualArrays are not materialized early.
   * you can't descend below NumpyArrays or EmptyArrays

and that's all the types.

agoose77:
Ok, this marries with my understanding of things - purelist would otherwise consider inner dimensions too eagerly! I'll add this to the back burner!

agoose77:
I ran into this again today, with `IndexedArray`s losing their arrayclass due to `layout.parameter` returning `{}` when wrapping a `RecordArray`:
```python
>>> form = ak.forms.Form.fromjson("""
{
[...]

PR #1755, opened 2022-09-29
feat: extend parameter coalescence to ignore transparent layouts

agoose77:
This PR goes towards fixing #1041, but requires some policy decisions about parameters. Originally, it added `dimension_parameters`; a tree-visiting mechanism for merging parameter dictionaries between layouts of the same dimension. @jpivarski then pointed out that we really care more about the _type_ in most cases, as that is the user-visible aspect of layouts.

This PR should do the following:
- [ ] Add a mechanism for retrieiving attribute(s) that coalesces the intra-dimension parameter(s) according to a policy 
- [ ] Change `Content.mergeable`, `Content.layout_equal`, `arrayclass()` etc. to use this API instead of inspecting single-layout parameter dictionaries, or special cased `purelist_XXX`
- [ ] Document the policy types so that we can refer to them in future.

## Policies
| Node Type | Pure List                 | Same Type                 |
|-----------|---------------------------|---------------------------|
| Option    | Merge with content        | Only ours                 |
| Indexed   | Merge with content        | Merge with content        |
| List      | Merge with content        | Only ours                 |
| Union     | Merge contents*, and ours | Merge contents*, and ours |
| Record    | Only ours                 | Only ours                 |
| Leaf      | Only ours                 | Only ours                 |

Practically speaking, we'll likely implement this as members of the `Form` types. Therefore, the word "policy" could mean a family of methods `_parameters_same_type, _parameters_pure_list`, or an enum that is switched upon. In fact, it could also be both, e.g a high-level enum that dispatches the correct family of methods. This is an aside for now.

Related issues:
- https://github.com/scikit-hep/awkward/issues/1685
- https://github.com/scikit-hep/awkward/issues/1686

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -719,7 +719,9 @@ class Content:
         # Otherwise, do the parameters match? If not, we can't merge.
         elif not (
             ak.forms.form._parameters_equal(
-                self._parameters, other._parameters, only_array_record=True
+                self.dimension_parameters,
+                other.dimension_parameters,
+                only_array_record=True,
             )
         ):
             return False
@@ -1572,7 +1574,9 @@ class Content:
             and len(self) == len(other)
             and ak.identifier._identifiers_equal(self.identifier, other.identifier)
             and ak.forms.form._parameters_equal(
-                self.parameters, other.parameters, only_array_record=False
+                self.dimension_parameters,
+                other.dimension_parameters,
+                only_array_record=False,
             )
             and self._layout_equal(other, index_dtype, numpyarray)
         )
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -1,9 +1,11 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+from __future__ import annotations
 
 import copy
 import math
 import numbers
 from collections.abc import Iterable, Sized
+from typing import Any
 
 import awkward as ak
 import awkward._reducers
@@ -725,7 +727,7 @@ class Content:
         else:
             return self._mergeable(other, mergebool)
 
-    def _mergeable(self, other: "Content", mergebool: bool) -> bool:
+    def _mergeable(self, other: Content, mergebool: bool) -> bool:
         raise ak._util.error(NotImplementedError)
 
     def mergemany(self, others):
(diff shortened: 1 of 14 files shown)


jpivarski:
I think you might have meant

```python
>>> import awkward as ak
>>> type(ak.Array(["hi", "bye"]))
<class 'awkward.behaviors.string.StringBehavior'>
>>> type(ak.Array(["hi", "bye"]).mask[[True, False]])
<class 'awkward.highlevel.Array'>
```

But this is, in fact, what we want. Behaviors can only be applied when a node with a given parameter is at top-level: the behavior class that one writes for type `T` is not applicable to type `option[T]`. If we were to ignore the distinction (for just option-type, say), then behavior classes would always have to be written in such a way that their `self` might be option-type or might not be option-type.

Hmmm. If the top-level node is IndexedArray (which does not change type), then the behavior should be applicable. Maybe we'd want the `arrayclass` helper function to "see through" IndexedArray.

And maybe it's time to define more behavior syntax rules, like

```python
ak.behavior["?", "RecordName"] = PossiblyMissingRecords
```

to say that it's allowed to "see through" option-type nodes.

--------------

In general, `dimension_parameters` sounds like a great idea—we were inching toward this with all of the "purelist" stuff. I followed your enumeration of rules and agree that those represent a meaningful way to "see through" certain nodes that are supposed to be invisible.

agoose77:
Your framing of this issue in terms of the user-visible _type_ greatly clarifies what I _think_ is the direction that we need to take. As far as users are concerned, only the type matters, so it _sounds_ like we actually want something more like 
```python
@property
def type_parameters(...):
    ...
```

I will update the rules table to reflect the type-dependence over dimension.

This raises the question - when do we want `dimension_parameters` rather than `type_parameters`? 

I deliberately did not implement anything like `dimension_parameter` yet because it would lead to a lot of duplication.  Maybe we want a `policy` argument that determines how deep the `parameters_for_policy()` and `parameter_for_policy()` lookups go. We'd still want these two functions for performance.

agoose77:
Although we recently dropped support for behavior classes that were specified for a singular array, `".", "Point"`, it's possible that we might add more to this effect in future. To my mind, the routine that checks whether this condition applies would naturally be implemented as a special policy for `parameter` / `parameters`, i.e. descend through a single additional list node. We might want to be mindful of that when designing this interface.

agoose77:
I'm closing this for now; it's not the highest priority, and might need some rethinking in light of recent work.

PR #1757, opened 2022-09-29
refactor: reduce qualified names of Content, Form, Type subclasses

jpivarski:
Specifically,

```bash
find src/awkward -type f -name '*.py' -exec sed -i.bak 's/ak.contents.[a-z]*array.\([A-Za-z]*Array\)/ak.contents.\1/g' {} \;
rm -f **/*.py.bak
black src/awkward
```

```bash
find src/awkward -type f -name '*.py' -exec sed -i.bak 's/ak.forms.[a-z]*form.\([A-Za-z]*Form\)/ak.forms.\1/g' {} \;
rm -f **/*.py.bak
black src/awkward
```

and

```bash
find src/awkward -type f -name '*.py' -exec sed -i.bak 's/ak.types.[a-z]*type.\([A-Za-z]*Type\)/ak.types.\1/g' {} \;
rm -f **/*.py.bak
black src/awkward
```

When pip-installed, `ak.types` doesn't seem to have `listtype`, etc., but it does have `ListType`. Wait—could it be because MacOS is case-sensitive? No, if that's it, then this would have failed before.

Regardless of whether this fixes the issue I see in using Uproot with Awkward's `main`, it's a name change we've been wanting to do.

### DIFF
--- a/src/awkward/_connect/jax/__init__.py
+++ b/src/awkward/_connect/jax/__init__.py
@@ -23,17 +23,17 @@ pytrees_registered = False
 
 def register_pytrees():
     for cls in [
-        ak.contents.bitmaskedarray.BitMaskedArray,
-        ak.contents.bytemaskedarray.ByteMaskedArray,
-        ak.contents.emptyarray.EmptyArray,
-        ak.contents.indexedarray.IndexedArray,
-        ak.contents.indexedoptionarray.IndexedOptionArray,
-        ak.contents.numpyarray.NumpyArray,
-        ak.contents.listarray.ListArray,
-        ak.contents.listoffsetarray.ListOffsetArray,
-        ak.contents.recordarray.RecordArray,
-        ak.contents.unionarray.UnionArray,
-        ak.contents.unmaskedarray.UnmaskedArray,
+        ak.contents.BitMaskedArray,
+        ak.contents.ByteMaskedArray,
+        ak.contents.EmptyArray,
+        ak.contents.IndexedArray,
+        ak.contents.IndexedOptionArray,
+        ak.contents.NumpyArray,
+        ak.contents.ListArray,
+        ak.contents.ListOffsetArray,
+        ak.contents.RecordArray,
+        ak.contents.UnionArray,
+        ak.contents.UnmaskedArray,
         ak.record.Record,
     ]:
         jax.tree_util.register_pytree_node(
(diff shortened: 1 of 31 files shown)


agoose77:
This looks good to me. I've been earmarking this for a while. 

> When pip-installed, ak.types doesn't seem to have listtype, etc., but it does have ListType. Wait—could it be because MacOS is case-sensitive? No, if that's it, then this would have failed before.

Were you definitely getting v2 there? The fact that we have `listtype` in ak.types should follow from the import statements, so I would be very surprised if we had a platform-dependence there.

In any case, this PR looks good; I tested `tests` and `tests-cuda`, and scanned through the unified diff.

agoose77:
@jpivarski what do you want to do w.r.t the tests? I think we can also change those too.

jpivarski:
I don't understand the reason why it was not finding the name. I verified the versions.

I got another error for not finding `awkward.types.type.Type`, something not covered above. At least it's solving the issue, despite not knowing why there was an issue.

jpivarski:
"Cleanup" type things are not very important in tests. We'll likely have to look at the main codebase frequently, making changes frequently, but a test should ideally just sit there, unchanged. Occasionally, they'll get changed, but it's a lot less than the main codebase.

jpivarski:
I'll be merging this as-is because we want to do the name change anyway. It got past some Uproot errors, but not all, and we _shouldn't_ extend it to handle cases like

```
        elif not (
>           ak.forms.form._parameters_equal(
                self._parameters, other._parameters, only_array_record=True
            )
        ):
E       AttributeError: module 'awkward._v2.forms' has no attribute 'form'
```

I don't understand (1) why it doesn't see the submodule `form` in `forms`, because it's really in there, and (2) why it's naming the module `awkward._v2.forms`.

agoose77:
I'll just mirror my comment here for posterity;

The reason that our `awkward.XXX` modules are renamed to `awkward._v2.XXX` when importing from v2 is that Python's import machinery sets `__package__` on the module object, even if it retrieves an existing module from `sys.modules`. Off the top of my head, I can't think of any way to prevent this, as all of this happens in C AFAICR. We could wrap the importer that Python creates, and implement logic to restore the `__package__` and `__name__` attributes, but this feels very heavy-handed for what is ultimately a cosmetic detail affecting only importers of v2 AFAICT.

PR #1758, opened 2022-09-30
refactor: make _v2 facade a static module.

jpivarski:
Instead of making `awkward._v2` be referentially the same as `awkward`, inserted into `sys.modules` by hand (which does strange things to the `__package__` names of all submodules, described here: https://github.com/scikit-hep/awkward/pull/1757#issuecomment-1262759917), I made a file named src/awkward/_v2.py and filled it with what the old _v2 had.

Now `awkward._v2` is a plain old boring submodule, referentially distinct from `awkward`, and submodule `__package__` names are not changed. It also has exactly the contents that the old `awkward._v2` had (not a superset of them), and it gives us a place where we could add a deprecation warning, if we ever want to. The direct imports can be replaced with a module-level `__getitem__` that warns before it returns what was requested, like the module-level `__getitem__` in Uproot ([here](https://github.com/scikit-hep/uproot5/blob/main/src/uproot/dynamic.py)).

### DIFF
--- a/src/awkward/_v2.py
+++ b/src/awkward/_v2.py
@@ -2,33 +2,28 @@
 
 # Following https://github.com/scikit-hep/awkward/blob/main-v1/src/awkward/_v2/__init__.py
 
+# behaviors
+# high-level interface
+# third-party connectors
+# internal
 # layout classes; functionality that used to be in C++ (in Awkward 1.x)
-from awkward import index  # noqa: F401
-from awkward import identifier  # noqa: F401
+from awkward import Array  # noqa: F401
+from awkward import ArrayBuilder  # noqa: F401
+from awkward import Record  # noqa: F401
+from awkward import _broadcasting  # noqa: F401
+from awkward import _connect  # noqa: F401
+from awkward import _lookup  # noqa: F401
+from awkward import _slicing  # noqa: F401
+from awkward import _typetracer  # noqa: F401
+from awkward import _util  # noqa: F401
+from awkward import behavior  # noqa: F401
+from awkward import behaviors  # noqa: F401
 from awkward import contents  # noqa: F401
+from awkward import forms  # noqa: F401
+from awkward import identifier  # noqa: F401
+from awkward import index  # noqa: F401
 from awkward import record  # noqa: F401
 from awkward import types  # noqa: F401
-from awkward import forms  # noqa: F401
-from awkward import _slicing  # noqa: F401
-from awkward import _broadcasting  # noqa: F401
-from awkward import _typetracer  # noqa: F401
-
-# internal
-from awkward import _util  # noqa: F401
-from awkward import _lookup  # noqa: F401
-
-# third-party connectors
-from awkward import _connect  # noqa: F401
-
-# high-level interface
-from awkward import Array  # noqa: F401
-from awkward import Record  # noqa: F401
-from awkward import ArrayBuilder  # noqa: F401
-
-# behaviors
-from awkward import behaviors  # noqa: F401
 
 # operations
 from awkward.operations import *  # noqa: F401
-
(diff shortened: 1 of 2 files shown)


agoose77:
The only downside of this approach is that `from awkward._v2.contents import ListArray` will no longer work. That said, I'm not sure that we _can_ easily have both things without e.g. having _two_ versions of Awkward (e.g. `awkward._v2.contents.ListArray is not awkward.contents.ListArray`). So it comes down to what we prefer. 

I have a few more ideas on this, which I'm just trying out :)

jpivarski:
I'll wait to see your idea. (Leaving this PR open.)

agoose77:
I've spent some time digging, and I've realised that [my understanding ](https://github.com/scikit-hep/awkward/pull/1757#issuecomment-1262759917) of what's happening with respect to the renamed modules is not the full picture. It turns out that the `importlib` docs aren't one-to-one on what CPython actually does (or at least, they're vague), so one should actually look at the [bootstrap implementation](https://github.com/python/cpython/blob/57c6cb5100d19a0e0218c77d887c3c239c9ce435/Lib/importlib/_bootstrap.py#L987-L1016). Despite being fairly familiar with the internals of importlib, this next part took me by surprise.

What's *actually* happening is that Python's creating a new module object for imports of `awkward._v2.XXX`, and then re-writing the name `XXX` on the original module. This is actually worse than just mangling the names, because it means that we could have multiple copies of modules lying around with half-shared state. 

I had not realised that this would happen, because I'd thought that Python took the `__package__` of the `_v2` module (which would be `"awkward"`. This was an oversight on my part; this, of course, only applies to relative imports. 

So, we *need* to revert #1730, and I think the safest thing to do is to just have a proxy module like you've shown. That will support `from awkward import _v2 as ak` and `import awkward._v2 as ak`, but not `from awkward._v2.contents import XXX`. 
We could pre-populate `sys.modules` to fix this, but we run into the same problem; multiple instances of the same module would exist if a non-pre-populated module were imported. We could pre-populate all "importable" files, ensuring that this last condition is never met. This, however, would be fragile and a non-insignificant amount of code, so I only mention it for posterity.

So, apologies for holding this back, and indeed taking us down this path to begin with. I was hoping that #1730 would give us clean support for aliasing v2, but it rather turned out to be a liability.

src/awkward/_v2.py
```diff
@@ -0,0 +1,22 @@
...
+from awkward import _lookup  # noqa: F401
+from awkward import _slicing  # noqa: F401
+from awkward import _typetracer  # noqa: F401
+from awkward import _util  # noqa: F401
+from awkward import behavior  # noqa: F401
+from awkward import behaviors  # noqa: F401
+from awkward import contents  # noqa: F401
+from awkward import forms  # noqa: F401
+from awkward import identifier  # noqa: F401
+from awkward import index  # noqa: F401
```
agoose77, line 19: Should we also expose `numba.py`, etc.? Looking at [v2 in main-v1](https://github.com/scikit-hep/awkward/tree/main-v1/src/awkward/_v2), we have a few modules that aren't listed here:
- `numba`
- `operations`
I've determined this just by looking for `import awkward.XXX` - this will set `XXX` on the awkward object.
```suggestion
from awkward import index  # noqa: F401
from awkward import numba  # noqa: F401
from awkward import operations  # noqa: F401
```

agoose77: If you are happy to make this change, then provided it works with uproot, I'm happy to merge!

jpivarski:
Okay, this does revert #1730, and I've been testing it against Uproot, though I think my issues were local to my computer. Uproot's tests passed even with the other way of doing `_v2`.

So I'll enable auto-squash-and-merge.

PR #1760, opened 2022-09-30
fix: use raise with `ak._errors.wrap_error`

agoose77:
This PR fixes two unrelated bugs:
- Use `raise` with `ak._errors.wrap_error` - a couple of lines were never raising exceptions
- Remove uses of `exception_suffix` in `ak._errors` - the v1 → v2 transition removed this


jpivarski (approved):
Great! Thanks for catching these: they would have only come up in error cases, and that's when it's the most confusing.

PR #1761, opened 2022-09-30
refactor: move `is_XXX_buffer` into respective nplikes

agoose77:
Now that we have four `NumpyLike` implementations, I think it's reasonable to move the external `is_XXX_buffer` functions into their respective NumpyLike classes.

This PR adds an `is_own_buffer` classmethod to each `NumpyLike`, including `TypeTracer`


jpivarski (approved):
I agree! There should be as little special-casing to particular backends in the body of the codebase as possible. (That's why the type tracer is identified by `known_shape` and `known_data`, rather than explicitly asking if it's the type tracer.)

PR #1762, opened 2022-09-30
refactor: clean up `_connect`

agoose77:
The _motivation_ for this PR is that we need an `ak.jax` public interface to our Jax integration, namely explicit registration.

During the process of implementing this feature, I found other parts of `_connect` that I felt could benefit from some reworking as Awkward v2 has matured. Not all of these changes are necessary, so if people don't like them, we can revert them. Hopefully I can make a good case for the changes that I did make.

## New Features
- [x] expose a (new) `ak.jax` module (which we had in v1) to provide public access to register Jax
- [x] add `ak.jax.assert_registered` to ensure Jax is registered
- [x] implement support for restoring behaviours 
- [x] add `ak.jax.register_behavior_class` to register behavior classes with Jax
- [x] add `__subclass_hook__` to highlevel types in order to automatically invoke `ak.jax.register_behavior_class()`

## Refactoring
- [x] make `_connect.numexpr.import_numexpr` private
- [x] change `ak.jax.import_jax()` to assert that it is registered, instead of registering it
- [x] move the `_jax_flatten` Array/Record classmethods to high-level functions in `ak._connect.jax`
 - We don't need these to be classmethods on the Array, so let's keep the namespace empty
- [x] refactor `ak._connect.jax` to separate `_connect.jax.trees`
- [x] ensure Jax reducers inherits from `Reducer`
- [x] add `ak._connect.jax.get_jax_reducer` to find the Jax reducer for a given NumPy reducer

## Breaking Changes
- `ak.jax.register` needs to be called by external code before using Jax with Awkward. 

I think that this is a _good_ change, because this was already partly true; previously, calling `jax.XXX` functions on Awkward Arrays would only succeed if the registration step had been performed manually, or `to_jax()` had been invoked. 

This change makes it harder to forget to register Jax integration, because now the Jax integration in either direction won't work until it is enabled. Internal modules call `assert_registered()`, which encourages users to call `register_and_check()`. There is also `jax.import_jax()` which imports Jax after ensuring that it is registered.

Although this is _slightly_ less convenient; it ensures that users don't have to think about whether to call `ak.jax.register_and_check()` - one always must. It takes a stateful API `ak.to_jax` and makes it stateless.

## High Level Changes
I think our `_connect` modules should assume that the importer knows what they're doing, and _wants_ these modules because they are known to be available on the host system. This makes it easier to reason about state, and lets us avoid silently invoking the `register()` functions by accident.  

So, for our main connect modules:
- `jax` - needs to be registered, as it uses global state for integration. Internal APIs can directly use the Jax APIs, but must ensure that `assert_registered` is called. i.e. we should explicitly call `ak.jax.import_jax()`, `ak.jax.register_and_check()`, or `ak.jax.ensure_registered()` when we need these guarantees at the high-level, e.g. `ak.to_jax()`. Import checks can be done at registration, and internal functions therefore just need to check that things are currently registered.
- `numba` - see Jax[^numba] 
- `cupy` - does not need to be registered, so internal functions need to / can perform import-checks dynamically.
- `numexpr` - see CuPy
Instead of registering Jax when needed, we should explicitly call `ak.jax.import_jax()`, `ak.jax.register_and_check()`, or `jax.ensure_registered()` when we need these guarantees at the high-level, e.g. `ak.to_jax()`.

[^numba]: Perhaps not in the case of numba: it should always be registered due to the entry point. Checking this anyway would be more robust, however.`

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,7 +58,7 @@ exclude =
 
 [options.entry_points]
 numba_extensions =
-    init = awkward.numba:register
+    init = awkward.numba:_register
 
 [flake8]
 extend-select = C,B,B9,T,AK1
--- a/src/awkward/__init__.py
+++ b/src/awkward/__init__.py
@@ -30,6 +30,7 @@ import awkward._lookup
 import awkward._connect.numpy
 import awkward._connect.numexpr
 import awkward.numba
+import awkward.jax
 
 # high-level interface
 from awkward.highlevel import Array
--- a/src/awkward/_connect/jax/__init__.py
+++ b/src/awkward/_connect/jax/__init__.py
@@ -1,45 +1,8 @@
-# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
-
-import awkward as ak
-
-
-def find_numpyarray_nodes(layout):
-
-    data_ptrs = []
-
-    def find_nparray_ptrs(node, **kwargs):
-        if isinstance(node, ak.contents.NumpyArray):
-            data_ptrs.append(node.data)
-
-    layout.recursively_apply(action=find_nparray_ptrs, return_array=False)
-
-    return data_ptrs
-
-
-def replace_numpyarray_nodes(layout, buffers):
-    def replace_numpyarray_nodes(node, **kwargs):
-        if isinstance(node, ak.contents.NumpyArray):
-            buffer = buffers[0]
-            buffers.pop(0)
-            return ak.contents.NumpyArray(
-                buffer,
-                layout.identifier,
-                layout.parameters,
-                nplike=ak.nplikes.Jax.instance(),
-            )
-
-    return layout.recursively_apply(action=replace_numpyarray_nodes)
-
-
-class AuxData:
-    def __init__(self, layout):
-        self._layout = layout
-
-    @property
-    def layout(self):
-        return self._layout
-
-    def __eq__(self, other):
-        return self.layout.layout_equal(
-            other.layout, index_dtype=False, numpyarray=False
-        )
+import awkward._connect.jax.reducers  # noqa: F401
+from awkward._connect.jax.trees import (  # noqa: F401
+    AuxData,
+    find_numpyarray_nodes,
+    jax_flatten_highlevel,
+    jax_unflatten_highlevel,
+    replace_numpyarray_nodes,
+)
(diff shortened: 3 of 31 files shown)


agoose77:
@jpivarski this turned out to be a lot bigger than I was expecting. The motivation for this PR was providing a public `ak.jax.register`, but I noticed whilst implementing it that we had a global-state registration mechanism , so I generalised it.

I'm just finalising some work to support automatic behavior registration.

agoose77:
Commit 173f913c070933b0ac26caea3172495344aacf7a is one that I'd benefit from scrutiny. It introduces some global state (weakref) to handle the case where users define behaviors, and subsequently register Jax support. We actually don't need this - we could use `ak.Array.__subclasses__()`, but I slightly prefer this explicit mechanism.

agoose77:
It seems we have a bug in our existing Jax handling - the `recursively_apply` that replaces Jax nodes implicitly calls `_getitem_range` on the layout nodes (in order to trim excess lengths). This triggers Jax's leak detection - if you operate on a Jax tracer that's been marked as finished (here I'm guessing), it considers it "leaked". Our tests don't currently throw an exception when operating on a single-dimension Numpy Array, because the statement that performs the slice is only in the list types e.g. `RegularArray`. 

fdf40b1931fc0f6f621242c81aff26c5f08091dd introduces a new `trim` argument to the layout transformer functions. This allows the caller to request that the visitor acts upon the raw, untrimmed layouts. I think this is general enough to warrant adding as a feature.

d9d03195e143b318494613052a97ac026e9da793 sets this parameter to `False` during unflattening to ensure that we don't trigger this bug.

agoose77:
OK, doesn't look like this is the only problem here; a reference seems to still be dangling.

agoose77:
OK, we've found the leak! The flatten logic in `Content.(_)jax_flatten` was returning `buffers_of(self), AuxData(self)` (effectively). The LHS buffers are kept alive by `AuxData`, through its reference to the layout. Instead, I think we want to drop the references from `AuxData`. 

The quick solution here is to return an `AuxData` with empty nodes. The proper solution is probably to replace the `layout` node contained in `AuxData` with a `form`, `length` pair (i.e. what `to_buffers` returns). This will benefit from the `Form` refactoring, as right now, forms don't have an equality comparison implementation.

agoose77:
OK, this PR is too big to easily review. I'm going to squash these commits into logical chunks, and open a new PR for review.

jpivarski:
Okay, I've followed this so far and I'm on board. I'm answering all of the "is this okay?" questions in the affirmative: the global weakref, the `trim` argument, empty AuxData.

Also, I'm deleting branches associated with closed PRs. We can always get them back again.

PR #1763, opened 2022-10-01
refactor: cleanup `_connect` submodule

agoose77:
This PR is a re-write of #1762, so I'll copy the information here:

The _motivation_ for this PR is that we need an `ak.jax` public interface to our Jax integration, namely explicit registration.

During the process of implementing this feature, I found other parts of `_connect` that I felt could benefit from some reworking as Awkward v2 has matured. Not all of these changes are necessary, so if people don't like them, we can revert them. Hopefully I can make a good case for the changes that I did make.

## New Features
- [x] expose a (new) `ak.jax` module (which we had in v1) to provide public access to register Jax
- [x] add `ak.jax.assert_registered` to ensure Jax is registered
- [x] implement support for restoring behaviours 
- [x] add `ak.jax.register_behavior_class` to register behavior classes with Jax
- [x] add `__subclass_hook__` to highlevel types in order to automatically invoke `ak.jax.register_behavior_class()`

## Refactoring
- [x] make `_connect.numexpr.import_numexpr` private
- [x] change `ak.jax.import_jax()` to assert that it is registered, instead of registering it
- [x] move the `_jax_flatten` Array/Record classmethods to high-level functions in `ak._connect.jax`
 - We don't need these to be classmethods on the Array, so let's keep the namespace empty
- [x] refactor `ak._connect.jax` to separate `_connect.jax.trees`
- [x] ensure Jax reducers inherits from `Reducer`
- [x] add `ak._connect.jax.get_jax_reducer` to find the Jax reducer for a given NumPy reducer
- [x] use `NumpyLike` mechanism in `from_backend`.

## Bug Fixes
- [x] fixed tracer leaks in Jax flattening
- [x] use `nplike` parameter in `ak.to_buffers`

## Breaking Changes
- `ak.jax.register_and_check()` needs to be called by external code before using Jax with Awkward. 

I think that this is a _good_ change, because this was already partly true; previously, calling `jax.XXX` functions on Awkward Arrays would only succeed if the registration step had been performed manually, or `to_jax()` had been invoked. 

This change makes it harder to forget to register Jax integration, because now the Jax integration in either direction won't work until it is enabled. Internal modules call `assert_registered()`, which encourages users to call `register_and_check()`. There is also `jax.import_jax()` which imports Jax after ensuring that it is registered.

Although this is _slightly_ less convenient; it ensures that users don't have to think about whether to call `ak.jax.register_and_check()` - one always must. It takes a stateful API `ak.to_jax()` and makes it stateless.

## High Level Changes
I think our `_connect` modules should assume that the importer knows what they're doing, and _wants_ these modules because they are known to be available on the host system. This makes it easier to reason about state, and lets us avoid silently invoking the `register()` functions by accident.  

So, for our main connect modules:
- `jax` - needs to be registered, as it uses global state for integration. Internal APIs can directly use the Jax APIs, but must ensure that `assert_registered` is called. i.e. we should explicitly call `ak.jax.import_jax()`, `ak.jax.register_and_check()`, or `ak.jax.assert_registered()` when we need these guarantees at the high-level, e.g. `ak.to_jax()`. Import checks can be done at registration, and internal functions therefore just need to check that things are currently registered.
- `numba` - see Jax[^numba] 
- `cupy` - does not need to be registered, so internal functions need to / can perform import-checks dynamically.
- `numexpr` - see CuPy
Instead of registering Jax when needed, we should explicitly call `ak.jax.import_jax()`, `ak.jax.register_and_check()`, or `jax.ensure_registered()` when we need these guarantees at the high-level, e.g. `ak.to_jax()`.

[^numba]: Perhaps not in the case of numba: it should always be registered due to the entry point. Checking this anyway would be more robust, however.`

### DIFF
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -22,7 +22,7 @@ class _RegistrationState(enum.Enum):
     FAILED = enum.auto()
 
 
-_registration_lock = threading.Lock()
+_registration_lock = threading.RLock()
 _registration_state = _RegistrationState.INIT
 
 
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import types
+import weakref
 from typing import TypeVar
 
 import awkward as ak
@@ -60,6 +61,23 @@ def register_behavior_class(cls: HighLevelType) -> HighLevelType:
     return cls
 
 
+_known_highlevel_classes = weakref.WeakSet([highlevel.Array, highlevel.Record])
+
+
+def maybe_register_behavior_class(cls: HighLevelType):
+    """
+    Args:
+        cls: behavior class to register with Jax
+
+    Register the behavior class with Jax, if Jax integration is enabled. Otherwise,
+    queue the type for subsequent registration when/if Jax is registered.
+    """
+    if _is_registered:
+        register_behavior_class(cls)
+    else:
+        _known_highlevel_classes.add(cls)
+
+
 def _register(jax: types.ModuleType):
     """
     Register Awkward Array node types with Jax's tree mechanism.
@@ -93,7 +111,7 @@ def _register(jax: types.ModuleType):
                 cls.jax_unflatten,
             )
 
-        for cls in [ak.highlevel.Array, ak.highlevel.Record]:
+        for cls in _known_highlevel_classes:
             jax.tree_util.register_pytree_node(
                 cls,
                 jax_connect.jax_flatten_highlevel,
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -1,28 +1,12 @@
 from __future__ import annotations
 
 import types
-from typing import Any
+from typing import TypeVar
 
 import awkward as ak
 from awkward import _errors, highlevel, nplikes
 
 numpy = nplikes.Numpy()
-
-
-def jax_flatten_highlevel(
-    array: highlevel.Array | highlevel.Record,
-) -> tuple[list[numpy.ndarray], Any]:
-    import awkward._connect.jax as jax_connect
-
-    return jax_connect.jax_flatten_highlevel(array)
-
-
-def jax_unflatten_highlevel(
-    aux_data: Any, children: list[numpy.ndarray]
-) -> highlevel.Array | highlevel.Record:
-    import awkward._connect.jax as jax_connect
-
-    return jax_connect.jax_unflatten_highlevel(aux_data, children)
 
 
 _is_registered = False
@@ -50,6 +34,30 @@ def register_and_check():
         ) from None
 
     _register(jax)
+
+
+HighLevelType = TypeVar(
+    "HighLevelType", bound="type[highlevel.Array | highlevel.Record]"
+)
+
+
+def register_behavior_class(cls: HighLevelType) -> HighLevelType:
+    """
+    Args:
+        cls: behavior class to register with Jax
+
+    Return the behavior class, after registering it with Jax.
+
+    """
+    jax = import_jax()
+    import awkward._connect.jax as jax_connect
+
+    jax.tree_util.register_pytree_node(
+        cls,
+        jax_connect.jax_flatten_highlevel,
+        jax_connect.jax_unflatten_highlevel,
+    )
+    return cls
 
 
 def _register(jax: types.ModuleType):
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -125,7 +125,7 @@ def assert_registered():
     """Ensure that Jax integration is registered. Raise a RuntimeError if not."""
     if not _is_registered:
         raise _errors.wrap_error(
-            RuntimeError("Jax features require `ak.jax.register()`")
+            RuntimeError("Jax features require `ak.jax.register_and_check()`")
         )
 
 
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -55,25 +55,6 @@ HighLevelType = TypeVar(
 )
 
 
-def _register_behavior_class(cls: HighLevelType) -> HighLevelType:
-    """
-    Args:
-        cls: behavior class to register with JAX
-
-    Return the behavior class, after registering it with JAX.
-
-    """
-    jax = import_jax()
-    import awkward._connect.jax as jax_connect
-
-    jax.tree_util.register_pytree_node(
-        cls,
-        jax_connect.jax_flatten,
-        jax_connect.jax_unflatten,
-    )
-    return cls
-
-
 _known_highlevel_classes = weakref.WeakSet([highlevel.Array, highlevel.Record])
 
 
@@ -85,9 +66,13 @@ def register_behavior_class(cls: HighLevelType):
     Register the behavior class with JAX, if JAX integration is enabled. Otherwise,
     queue the type for subsequent registration when/if JAX is registered.
     """
+    # Acquire lock so that we know registration has completed
     with _registration_lock:
         if _registration_state == _RegistrationState.SUCCESS:
-            _register_behavior_class(cls)
+            # Safe to invoke JAX code here
+            import awkward._connect.jax as jax_connect
+
+            jax_connect.register_pytree_class(cls)
         else:
             _known_highlevel_classes.add(cls)
 
@@ -119,18 +104,11 @@ def _register(jax: types.ModuleType):
                 ak.contents.UnmaskedArray,
                 ak.record.Record,
             ]:
-                jax.tree_util.register_pytree_node(
-                    cls,
-                    jax_connect.jax_flatten,
-                    jax_connect.jax_unflatten,
-                )
+                jax_connect.register_pytree_class(cls)
 
             for cls in _known_highlevel_classes:
-                jax.tree_util.register_pytree_node(
-                    cls,
-                    jax_connect.jax_flatten,
-                    jax_connect.jax_unflatten,
-                )
+                jax_connect.register_pytree_class(cls)
+
         except Exception:
             _registration_state = _RegistrationState.FAILED
         else:
@@ -146,7 +124,11 @@ def assert_registered():
             )
         elif _registration_state == _RegistrationState.FAILED:
             raise _errors.wrap_error(
-                RuntimeError("JAX features require `ak.jax.register_and_check()`")
+                RuntimeError(
+                    "JAX features require `ak.jax.register_and_check()`, "
+                    "but the last call to `ak.jax.register_and_check()` did not succeed. "
+                    "Please look for a traceback to identify the error."
+                )
             )
         elif _registration_state == _RegistrationState.SUCCESS:
             return
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -1,0 +1,72 @@
+from __future__ import annotations
+
+import types
+
+import awkward as ak
+
+_is_registered = False
+
+
+def register():
+    """
+    Register Awkward Array node types with Jax's tree mechanism.
+    """
+    # Let's only do this once
+    global _is_registered
+    if _is_registered:
+        return
+
+    import jax
+
+    for cls in [
+        ak.contents.BitMaskedArray,
+        ak.contents.ByteMaskedArray,
+        ak.contents.EmptyArray,
+        ak.contents.IndexedArray,
+        ak.contents.IndexedOptionArray,
+        ak.contents.NumpyArray,
+        ak.contents.ListArray,
+        ak.contents.ListOffsetArray,
+        ak.contents.RecordArray,
+        ak.contents.UnionArray,
+        ak.contents.UnmaskedArray,
+        ak.record.Record,
+    ]:
+        jax.tree_util.register_pytree_node(
+            cls,
+            cls.jax_flatten,
+            cls.jax_unflatten,
+        )
+
+    for cls in [ak.highlevel.Array, ak.highlevel.Record]:
+        jax.tree_util.register_pytree_node(
+            cls,
+            cls._jax_flatten,
+            cls._jax_unflatten,
+        )
+    _is_registered = True
+
+
+def import_jax() -> types.ModuleType:
+    """
+    Import jax and return the module, or raise a helpful error message if it is not available.
+    """
+    try:
+        import jax
+
+    except ModuleNotFoundError:
+        raise ak._errors.wrap_error(
+            ModuleNotFoundError(
+                """install the 'numba' package with:
+
+        python3 -m pip install jax jaxlib
+
+    or
+
+        conda install -c conda-forge jax jaxlib
+    """
+            )
+        ) from None
+
+    register()
+    return jax
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -6,8 +6,6 @@ from typing import Any
 import awkward as ak
 from awkward import highlevel, nplikes
 
-_is_registered = False
-
 numpy = nplikes.Numpy()
 
 
@@ -25,6 +23,9 @@ def jax_unflatten_highlevel(
     import awkward._connect.jax as jax_connect
 
     return jax_connect.jax_unflatten_highlevel(aux_data, children)
+
+
+_is_registered = False
 
 
 def register():
@@ -79,7 +80,7 @@ def import_jax() -> types.ModuleType:
     except ModuleNotFoundError:
         raise ak._errors.wrap_error(
             ModuleNotFoundError(
-                """install the 'numba' package with:
+                """install the 'jax' package with:
 
         python3 -m pip install jax jaxlib
 
--- a/src/awkward/jax.py
+++ b/src/awkward/jax.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+import enum
+import threading
 import types
 import weakref
 from typing import TypeVar
@@ -10,7 +12,18 @@ from awkward import _errors, highlevel, nplikes
 numpy = nplikes.Numpy()
 
 
-_is_registered = False
+def assert_never(arg) -> None:
+    raise AssertionError(f"this should never be run: {arg}")
+
+
+class _RegistrationState(enum.Enum):
+    INIT = enum.auto()
+    SUCCESS = enum.auto()
+    FAILED = enum.auto()
+
+
+_registration_lock = threading.Lock()
+_registration_state = _RegistrationState.INIT
 
 
 def register_and_check():
@@ -42,7 +55,7 @@ HighLevelType = TypeVar(
 )
 
 
-def register_behavior_class(cls: HighLevelType) -> HighLevelType:
+def _register_behavior_class(cls: HighLevelType) -> HighLevelType:
     """
     Args:
         cls: behavior class to register with Jax
@@ -64,7 +77,7 @@ def register_behavior_class(cls: HighLevelType) -> HighLevelType:
 _known_highlevel_classes = weakref.WeakSet([highlevel.Array, highlevel.Record])
 
 
-def maybe_register_behavior_class(cls: HighLevelType):
+def register_behavior_class(cls: HighLevelType):
     """
     Args:
         cls: behavior class to register with Jax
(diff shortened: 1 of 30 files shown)


agoose77:
#1762 also implemented a new `trim` argument for `recursively_apply`. I removed this, as it was not required for the bug fix in the end, and I think we need to be careful not to add too many parameters just because they might be useful to someone.

agoose77:
Despite my efforts, it seems like we still have tracer leaks (see the Jax user-guide demo). Ultimately, this happens because we "leak" tracer references. This isn't necessarily an issue if we eventually dispose of them, except if the user uses a `jax.checking_leaks()` context. So, we do need to avoid leaking tracers. In `main`, the tracers are considered "leaked" because they have a nonzero refcount, due to the buffers being referenced inside `AuxData.layout`. Furthermore, without the leak-checker enabled, we trigger a leak detection event when we operate on "disposed" tracers. This happens in `main` when we go to replace the `NumpyArray` buffers; the node transformer trims list layouts, which triggers a `__getitem__` call on the tracer leaf arrays. These are old tracers from the `flatten` implementation, left in the layout.

There are some hacky solutions:
- Build a new layout for AuxData, by replacing the `NumpyArray` nodes with ones that hold empty _numpy_ buffers, using the Numpy `nplike`.
- Re-implement a `trim` parameter for `recursively_apply` that let's us disable this trimming.

Even if we fix the error (by disabling trimming) that is triggered by operating on a leaked tracer (when unflattening, the AuxData has the _old_ tracers in the layout), we still have tracer leaks, i.e. the layout is fragile w.r.t operations. 

So, I think changing aux-data to behave like `from_buffers` is _probably_ the best way to go here. This would require:
1. Support for comparing form equality (including ignoring index type differences)
2. Identifying the NumpyArray nodes from a layout and correlating them with their form keys. 

RE (2), the easiest approach is to just pull out any Jax types from the `to_buffers` result. This doesn't feel robust - we're assuming (albeit reasonably?) that the only Jax types in a layout are going to be NumpyArray content nodes. Although I can't see why that wouldn't be the case, it's not as safe as explicitly pulling out the NumpyArray contents. I'm also not sure how much we can rely on the string values of the form keys, e.g. `NumpyArray` buffers living under `"data"` - do we document that these are the defaults, somewhere (so that we can't just change them accidently in future)? I'll think some more on this.

agoose77:
`ak.to_buffers` is probably still the right way to do this - Forms are a content-free representation of arrays. However, as-of-now, `from_buffers` calls `nplike.frombuffer`, which we don't want to happen with tracers. Perhaps we need to introduce a new nplike method for "load a buffer OR array", which is just `frombuffer` for NumPy, and CuPy, but something else for Jax. In the latter case, we would ensure that the array dtype matches the requested dtype.

So, as a stop-gap solution, I've since revisited my concern about just building a layout with NumPy array data values, in [ccf6301](https://github.com/scikit-hep/awkward/pull/1763/commits/ccf6301c2ee23d702e38e5a8e2e68cbf664aac80). 
The `AuxData` layout is only going to be used by our Jax handling code, so it's okay if it's a bit fragile. And, in this case, it's not actually "fragile" - we convert _all_ nodes to NumPy nodes, which avoids mixing `nplikes` within the same layout.
The only drawback is that we're allocating empty memory only to then not-use it we could be smarter about this, e.g. maintaining a cache for each dtype
```python
buffer = get_buffer(content.dtype, at_least_length=content.length)
```

But I think we need to probably solve this properly, so I'm happy to leave this as a FIXME.

agoose77:
OK, tests are now failing only (I hope) because the dtypes from the arrays passed from Jax in `unflatten` have non-primitive dtypes. We might need some routine to coerce these types.

agoose77:
It's worth me stating here that I've made some decisions/thought processes in this PR that reflect my lack of high-level knowledge w.r.t our goals/expectations of Jax.

We need _some_ tracer support, because when we differentiate a function that operates on an Awkward Array, jax ultimately hands us some `JVPTracer` objects. These tracers technically have a known shape. `DynamicJaxprTracer` tracers come from jitted functions, and _don't_  necessarily have a known shape IIRC. Again, for us to see these in Awkward, it's probably that Jax has flattened an Awkward Array, and is now asking us to unflatten it with tracers. So, we should end up seeing tracers over `ShapedArray`, which do have shapes.

All of these facts mean that I think we *do* need to support tracers at the nplike level. I think high-level functions like `to_/from_`, etc. don't need to be jittable/diffable, so these can reject any non-concrete types. However, `to_`/`from_buffers`, which don't need to convert data (it can directly use the buffer nodes, if we pass in the same nplike), can be used with tracers.

agoose77:
OK, this *might* be acceptable. I've added a new `nplike.astype` to handle the difference between `np.ndarray.astype` and `jaxlib.DeviceArray.astype`. I've also added tracer support to `frombuffer`, although it's fairly rudimentary.

`AuxData` now uses `to_`/`from_buffers`, which is only partially correct as it compares the `to_dict()` representation of the forms. We actually want to allow forms to differ slightly w.r.t dtypes of indices.

This feels like the right direction, but I'll hold off for comments from @jpivarski (or Anish, if he's still working on this) :)

agoose77:
Making a note here to discuss with Jim RE `index_nplike` and when it should be used.

src/awkward/numba.py
```diff
@@ -2,37 +2,37 @@
...
                 "(you have version {})".format(numba.__version__)
             )
+        _has_checked_version = True
 
-    register()
+    _register()
 
 
-def register():
+def _register():
```
jpivarski, line 36: The motivation for this being a public function was so that users could call

```python
ak.numba.register()
```

if the entry-point didn't work for some reason. That's a backdoor, but it can be an important one because the entry-point mechanism depends on issues in the environment being set up correctly and that's distinct from code. (Sometimes, I've had valid code, incorrectly installed, and puzzled over why something that should be in the entry-point wasn't.)

I'd like to decouple the _possibility_ of getting Awkward-Numba to work from the messy world of people not following the best practices for package installation. (You know, that XKCD.) When somebody reports a Numba error message saying that the ak.Array type is not recognized, I'd like to be able to say, "Fix you environment; but for now, use `ak.numba.register()`."

So it would be good for this function to be public. (Otherwise, why have an `ak.numba` module at all?)

jpivarski (commented):
We talked about this a lot on Slack. (I'm not sure whether to call this a "comment" or "request changes.") It's a good refactoring, cleaning up JAX and Numba registration, and I'm on board with making JAX registration mandatory.

Our support of JAX only includes autodiff, not JIT, so in principle it would seem that we do not need to allow for JAXTracers at any level. As I understand it right now, JAX sometimes turns its DeviceArrays into JAXTracers for a very short-lived pass (unflatten and then immediately re-flatten). We unflatten by running the Form, structural_buffers, and data_buffer_keys (pointers to DeviceArrays in a dict) through `ak.from_buffers`. You've expanded `ak.from_buffers` to allow some of the "buffers" to be JAXTracers, but that extends the intent of this function: the buffers are supposed to be things that satisfy the Python Buffer protocol, which has a Venn diagram overlap with array-like things (nplike.arrays). Implementing `frombuffer` on the JAX nplike generalizes `ak.from_buffers` beyond the assumptions under which it was written.

As I said on Slack, I'm thinking it may be better to bypass `ak.from_buffers` when the data are not buffers (such as DeviceArrays). But I'm also wondering what that short-lived pass is and whether it's not something we're supposed to be able to deal with.

So I don't have definite action items at the moment, other than (1) really understanding why we're seeing JAXTracers at all and (2) _maybe_ circumventing `ak.from_buffers` in that case, rather than generalizing it.

agoose77:
As we discussed further on Slack, Jax gives us tracers when building an `ak.Array` from `primal` and `tangent` objects at the start of a `jvp` call (by flattening the primal, tangent objects; then combining them into a tracer; and finally unflattening that tracer into an `ak.Array`), and at the end (flattening the result; splitting the primal and tangent arrays; unflattening them into `ak.Arrays`). Operations on `ak.Array` objects inside the `jvp`d function will all be operating on `JVPTracer` objects. 

We *don't* support jitting, and what threw me off was an old demo in `user-guide` that used `jax.jit`, which gives us a `JaxprTracer`.

Most of this PR is therefore good, but I think we might want to defer the unflattening until a separate PR; the unflatten work will require forms to support equality comparisons, and we also want to touch the form interface in a separate refactoring change. 

We discussed generalising `from_buffers` to support, in addition to buffer-like objects, arrays that satisfy an `nplike`'s `is_own_buffer`. We can use the `to_backend` features to convert to the appropriate `nplike` in the case that it differs from the required.

src/awkward/numba.py
agoose77: I agree with the logic, and this change was just a refactor: `_register` doesn't call the check logic, which ensures that numba is available and that its version is supported. That's fine for numba - if numba is calling our entrypoint, then at least we know it's installed.

I thought rather that `ak.numba.register_and_check()` made the most sense to expose publicly (and should be used). Do you agree with that rationale?

agoose77:
I think this is ready for final review. The only failing tests are derivatives of booleans, which I'm not sure is trivially defined (both logically and in Jax). I think we just need to remove these tests. Maybe also these reducers.

src/awkward/numba.py
jpivarski: Oh, that's true—if an external user is calling this, we don't want it to go by without ensuring that the Numba version is acceptable. (That's the problem with soft-imports!)

If we're going to have multiple third-party registration functions, presenting them all with the same name would make it easier for users to guess—fewer round-trips to the documentation. On the other hand, if they all start with "register*", then tab-completion would resolve it, too.

jpivarski: Or... maybe they're _all_ named `register_and_check`. These two are.

tests/test_1399-from-jax.py
```diff
@@ -6,9 +6,10 @@
 import awkward as ak  # noqa: F401
 
 jax = pytest.importorskip("jax")
-
 jax.config.update("jax_enable_x64", True)
 
+ak.jax.register_and_check()
```
jpivarski, line 11: Just last week, you put in some effort to ensure that the tests are all thread-safe. If two of these `ak.jax.register_and_check` functions start at the same time, would they interfere with each other? I'll go check.

src/awkward/jax.py
```diff
@@ -0,0 +1,137 @@
...
+            )
+
+        for cls in _known_highlevel_classes:
+            jax.tree_util.register_pytree_node(
+                cls,
+                jax_connect.jax_flatten,
+                jax_connect.jax_unflatten,
+            )
+    finally:
+        _is_registered = True
```
jpivarski: Reading the above, `_is_registered` essentially means, "I tried to register, please don't try again."

The `_register_and_check` function could be called simultaneously from two threads, and this is setting global state. The easiest way to make this almost thread-safe is to move the

```python
_is_registered = True
```

line to the beginning, like

```suggestion
    global _is_registered
    if _is_registered:
        return
    _is_registered = True

    import awkward._connect.jax as jax_connect

    for cls in [
        ak.contents.BitMaskedArray,
        ak.contents.ByteMaskedArray,
        ak.contents.EmptyArray,
        ak.contents.IndexedArray,
        ak.contents.IndexedOptionArray,
        ak.contents.NumpyArray,
        ak.contents.ListArray,
        ak.contents.ListOffsetArray,
        ak.contents.RecordArray,
        ak.contents.UnionArray,
        ak.contents.UnmaskedArray,
        ak.record.Record,
    ]:
        jax.tree_util.register_pytree_node(
            cls,
            jax_connect.jax_flatten,
            jax_connect.jax_unflatten,
        )

    for cls in _known_highlevel_classes:
        jax.tree_util.register_pytree_node(
            cls,
            jax_connect.jax_flatten,
            jax_connect.jax_unflatten,
        )
```

Assigning the boolean to True at the beginning, instead of at the end with `finally`, still makes it mean "I tried to register; doesn't mean I was successful." If you also need to know that registration has been successful, add another boolean at the end (no need for `finally`; if it doesn't get there, it wasn't successful).

It's almost thread-safe because there's a chance that one thread sees `if _is_registered` as False, then another thread sets it to True before it gets to the next line. The fact that Python interprets these as naive bytecodes with a GIL makes that pretty unlikely.

To be ironclad about that, replace the boolean with a [threading.Event](https://docs.python.org/3/library/threading.html#event-objects). That's only needed for the "I tried" variable, not the "I was successful" variable.

jpivarski: I don't see any uses of `_is_registered` that need it to mean, "registration was successful." So just one variable is needed, it seems.

src/awkward/jax.py
```diff
@@ -0,0 +1,137 @@
...
+    Args:
+        cls: behavior class to register with Jax
+
+    Register the behavior class with Jax, if Jax integration is enabled. Otherwise,
+    queue the type for subsequent registration when/if Jax is registered.
+    """
+    if _is_registered:
+        register_behavior_class(cls)
+    else:
+        _known_highlevel_classes.add(cls)
```
jpivarski: Does this mean that after `register_and_check` is called, new high-level classes can't be added? A user's workflow can't consist of

   1. `register_and_check`
   2. try some things
   3. add a new behavior class
   4. try some more things

because the new behavior class doesn't get added to the registered PyTrees? If they call `register_and_check` a second time, it won't do the new behavior classes because it exits early.

Are we second-guessing how JAX's registration works? If we re-register something that has already been registered, are you worried that that's a memory leak? (If it's associated with a class object, I don't think Python ever gets rid of those, anyway.)

jpivarski (commented):
Again, these are more "questions to think about" than requested changes.

Personally, I'd prefer if JAX registration were implicit, like Numba registration, though Numba has an entrypoints and JAX doesn't. If JAX ever gets entrypoints, can this registration procedure be made automatic? (I.e. would it upgrade gracefully?)

src/awkward/jax.py
agoose77: This code is added precisely because at the moment we don't support user-defined behaviours. In main, when we flatten-unflatten, we lose behaviours. Sometimes we might want this (dy/dx is not y), but sometimes we do (e.g when jax builds the traced array for the first time). This solution ensures that we don't lose behaviors, but so that the user can decide whether to e.g. drop the `__array__` parameter.

To add support for user-defined behaviour classes, we need to handle registration (of classes) at all stages of the Jax integration lifecycle, namely _before_ `register_and_check()` and _after_ `register_and_check()`. This fragment adds the class to a set of classes to be registered _if_ we haven't registered Jax integration yet. Otherwise, we immediately register the class with Jax.

However, we might actually want to do more here because now that I think about it, jax doesn't like multiple registration of the same type. I noticed this whilst debugging yesterday.

tests/test_1399-from-jax.py
agoose77: Yes, this code is not thread-safe.

src/awkward/jax.py
agoose77: I think the use of `try`-`finally` here is probably a bad choice. Re-thinking it, we want registration to be considered failed if an `Exception` is raised. Given that vs always succeeding if `register_and_check` has been called, we just want a `Lock` here; then we can use a simple boolean. 

That said, I'm not sure how thread-safe Jax is, so I don't know to what lengths these changes need to go.

src/awkward/jax.py
agoose77: > However, we might actually want to do more here because now that I think about it, jax doesn't like multiple registration of the same type. I noticed this whilst debugging yesterday.

Wait, that will be fine - users should never re-register an existing class, only _new_ types with the same `__array__` parameter.

tests/test_1399-from-jax.py
agoose77: Actually, the work I undertook last week was more concerned with sequential out-of-order execution. I haven't tested threaded execution (I've used xdist for multi-process parallelism when required)

src/awkward/jax.py
jpivarski: At most, there are four states: (a) haven't tried to register yet, (b) registering now, all is good so far, (c) registration failed, and (d) registration succeeded.

```mermaid
graph TD
    A["(a) haven't started"] --> B["(b) registering now"]
    B --> C["(c) failed"]
    B --> D["(d) succeeded"]
```

From what I've seen, your _uses_ of `_is_registered` don't need to distinguish between (b), (c), (d); they only care about "not (a)." If I'm wrong about that, then you'll need more than one variable or a variable that can take on multiple states.

If what you want is boolean or close to it (2 bits instead of 1 bit), I would suggest Events, rather than Locks. Locks are very general; the state diagram can't loop back on itself, so a one-time-use mechanism like an Event seems more appropriate. (I don't know how lightweight they are or what OS resources they use, but I'm guessing the simpler Events use less. If they're built into the Python virtual machine, Events could be implemented in an extremely lightweight way, though I don't know that they did that.)

This may seem tangential to the main point of cleaning up the JAX code and moving on, but we don't want to break multithreaded testing and this sets a pattern of similar registrations. You've found that JAX doesn't like registering things twice, so it's our job to protect it from that.

src/awkward/jax.py
jpivarski: Okay, sounds good.

src/awkward/jax.py
agoose77: It's good to have this discussion, and a nice idea to use Mermaid, I keep forgetting how much it can improve communication. 

Right now, the code uses a `finally` clause. I think that is a symptom that I never quite decided what to do here, and ended up with something that solves neither approach.

Equally, I'm not sure that an `Event` is a good fit, regardless of which solution we want to use: in this context, my understanding is that we want to do something like
```python
if not event.is_set():
    event.set()
    ...
```
I don't think this is thread-safe between reading `.is_set()` and calling `.set()`. We're not really using the Event API in the way it is designed, i.e. waiting on the event.

When I suggest using a `Lock`, it is because I was actually considering states (c) and (d) important, *and*, crucially, I thought it was good to have this routine be re-entrant: I've sometimes seen it helpful to be able to retry a registration function and see the traceback, instead of the library just saying "this succeeded because we didn't try it again". To support that, a lock would signify "don't check if we're registered yet, someone else is doing that", and if the lock is already held, upon release the next caller can try and register - and get an exception if this fails. However, I don't think this is a useful design; Jax will error if you re-register the same classes here, so it's setting the user up to get confusing errors if the first registration failed midway through for some obscure reason.

So, think a simple "did we call register yet?" flag is sufficient, but will need a lock, i.e.
```python
_register_lock = threading.Lock()
_register_has_been_called = False

def register():
    global _register_has_been_called
    with _register_lock:
        if _register_has_been_called:
            return
        _register_has_been_called = True
    
    register_impl()
```

Does that sound reasonable to you?

agoose77: Oh, I missed something - we also need to wait until registration has been completed in `maybe_register_behavior_class`. Let me check that over.

jpivarski: > ```python
> ```python
> if not event.is_set():
>     event.set()
>     ...
> ```
>
> I don't think this is thread-safe between reading .is_set() and calling .set().
> ```

Yeah... right... I don't know what I was thinking. Maybe I was thinking of an atomic boolean, with an atomic operation like "if True, return True; if False, change to True and return False". There must be some primitive like that (which can be negated if need be).

Anyway, a lock does it. You said you want it to be reenterant—do you want an RLock?

(By the way, that was my first Mermaid diagram; I thought it was a good instance in which to learn how to do it.)

agoose77: > Yeah... right... I don't know what I was thinking. 

Actually, I think you are exactly reproducing the way I feel about threading primitives every time I go looking. I nearly always just want a locked variable, e.g.
```python
class Variable:
    def __init__(self, lock=None, default=None):
        self._lock = lock or Lock()
        self._value = default
    def set(self, value):
        with self._lock:
            self._value = value
    def get(self):
        with self._lock:
            return self._value
```

Being able to read atomically is fine ... if you don't have any global state that is being manipulated. But, it's not that hard to just spray some `with lock` everywhere, so maybe I complain too much ;)

> You said you want it to be reenterant—do you want an RLock?

I should have avoided that word given that we're talking about threading. I meant that perhaps the routine should only be considered "registered" if it succeeds. I've since decided that's a waste of time though, because Jax has its own global state that gets mangled if you register twice.

agoose77:
@jpivarski I think [d1e4a02](https://github.com/scikit-hep/awkward/pull/1763/commits/d1e4a029b836062b9581be78b17fff83acf76544) tackles the thread safety of the jax registration. Crucially, internal uses of Jax *must* call `assert_registered()` to avoid race conditions in registration - `assert_registered()` acquires the registration lock. Note that most of the routines in `jax.py` acquire the lock because we don't want to read it half-way through registration.

Note that I haven't done the same for Numba yet. I'm not sure yet whether we need these things; do we want to support awkward in multithreaded environments? I'm guessing probably yes, given that we use thread locals in our `ErrorContext` implementation.

jpivarski:
Because of Numba's entrypoint mechanism, calling `register_and_check` manually is a fallback, so it's not as important. Also, if you register something twice in Numba, I don't think it cares. (I don't know if the memory use grows, but that would be small unless you're completely abusing it.)

So that's it! I approve this PR.

There are some test failures to work through, though: `module 'awkward.jax' has no attribute 'maybe_register_behavior_class'`.

agoose77:
> maybe_register_behavior_class

Ah drat, that's an artefact of my IDE not synchronising `awkward` against `src`. Hold on. 

Also, I think there are some other test failures w.r.t differentiation of boolean functions.

agoose77:
I think with `threading.Lock` this code is hitting a deadlock somewhere. This is strange; I'd only imagine that's possible if something tries to acquire the lock during import, e.g. by creating a global `nplike.Jax.instance()`.

We can fix this with an `RLock`. but that's just a bandaid over the cause.

agoose77:
Nope, just an oversight on my part. I've left the lock as a `Lock`, but actually it can be an `RLock` so I've made it so. We don't use the re-entrant nature of the lock deliberately, but hopefully it will avoid any problems in future.

agoose77:
@jpivarski I've fixed the test failures (subject to CI succeeding). Are you happy with this design? If so, feel free to squash and merge.

jpivarski (approved):
Yes, I think this is great! I gave it another look-over.

PR #1768, opened 2022-10-04
fix: use dict representation of Form in pickling

### LINKED ISSUE #1767: Pickling of `Array`s should use `Form`'s dict representation
### Version of Awkward Array

main

...

agoose77:
Fixes #1767


jpivarski (approved):
When `from_buffers` gets a dict representation, it converts it? I guess so. Not explicitly converting it allows it to be any of the acceptable representations (JSON string, Python dict, or Form objects), and it's good to be loose when reading.

PR #1769, opened 2022-10-04
fix: ignore unknown JAX "buffer" types & fix nplike mixing

### LINKED ISSUE #1764: JAX sometimes gives us `None`s and `object`s
### Version of Awkward Array

main

We need to handle these. The JAX docs aren't very clear unfortunately :/ As per usual, I only discovered this after writing the new Jax docs (after merging #1763)

agoose77:
#1763 made some good strides in refactoring the JAX support, but there were some outstanding bugs with respect to parameter handling.

This PR just ensures that we have the same `nplike` throughout the layout (apart from `AuxData` layouts, which have `Numpy` at the leaves). It also introduces a test for later support of Jacobians. 

Finally, we also ignore non-buffers. This is probably the wrong thing to do in the general sense, but it's a start towards Jacobian support.

### DIFF
--- a/src/awkward/_connect/jax/trees.py
+++ b/src/awkward/_connect/jax/trees.py
@@ -119,6 +119,19 @@ class AuxData(Generic[T]):
         return self._is_highlevel
 
     def unflatten(self, buffers: tuple) -> T:
+        for buffer in buffers:
+            # Check that JAX isn't trying to give us float0 types
+            dtype = getattr(buffer, "dtype", None)
+            if dtype == np.dtype([("float0", "V")]):
+                raise ak._errors.wrap_error(
+                    TypeError(
+                        f"a buffer with the dtype {buffer.dtype} was encountered during unflattening. "
+                        "JAX uses this dtype for the tangents of integer/boolean outputs; these cannot "
+                        "reasonably be differentiated. Make sure that you are not computing the derivative "
+                        "of a boolean/integer (array) valued function."
+                    )
+                )
+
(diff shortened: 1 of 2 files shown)


agoose77:
Hmm, it looks like I need to take a step back and re-think this JAX work. I think we are currently fighting against JAX's pytree system. 

I asked a [somewhat obvious question in the JAX discussions](https://github.com/google/jax/discussions/12656), through which I realised that I was not thinking about this generally enough. Right now we're really only thinking in terms of JVPs in the JAX integration; we don't expect the transformations to change the shape of the array. 

a) when differentiation w.r.t multiple arguments, jax curries the non-dependent argument, and so the function being differentiated can be evaluated with mixed-nplike types. Maybe this is fine, we just require users to call `to_jax` on arrays first. Or, we could coerce things in the case of JAX. I don't know.

The Jacobian of a vector-valued function results in a pytree-of-pytrees in JAX. I need to think on this more, but I believe that our current approach has no elegant way to round-trip this, because we treat all content nodes as opaque with a set of leaves (NumpyArray). I think we need to be able to represent more levels of the tree structure in order to correctly handle this case, but don't hold me to this just yet yet.

jpivarski (approved):
(For the record) we talked in Zoom today about scope, and agreed to leave the Jacobian unimplemented until requested by a user. Getting in touch with the JAX team now about how to do it is a good idea, though, since it might require updates in JAX. We've never had a direct conversation with anyone from the JAX team about what we're trying to do here.

PR #1770, opened 2022-10-04
feat: make ak.from_json with schema ignore unspecified fields (instead of error).

jpivarski:
Requested by @martindurant on Slack.

> Quick question: if you provide a schema to ak.from_json which omits some fields in the input data, are they skipped at parse time? So, is this like a columns= opportunity to improve memory use and maybe parse speed?

The parser now needs to have additional mutable state, but it's quite minor: a single integer will do it. Every JSON node queries this integer (and possibly returns early, depending on its value), but surely the computation time of checking an integer is minimal compared to parsing tokens of JSON text. I didn't even do a performance test; adding this will be fine.

### DIFF
--- a/src/libawkward/io/json.cpp
+++ b/src/libawkward/io/json.cpp
@@ -355,6 +355,7 @@ namespace awkward {
       , nan_string_(nan_string)
       , posinf_string_(posinf_string)
       , neginf_string_(neginf_string)
+      , ignore_(0)
       , moved_(false)
       , schema_okay_(true) { }
 
@@ -376,6 +377,11 @@ namespace awkward {
     bool Null() {
       moved_ = true;
       // std::cout << "null " << specializedjson_->debug() << std::endl;
+
+      if (ignore_) {
+        // std::cout << "  ignoring!" << std::endl;
+        return true;
+      }
 
       switch (specializedjson_->instruction()) {
         case FillByteMaskedArray:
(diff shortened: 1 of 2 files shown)


agoose77:
This looks reasonable, and I agree with the performance motivations; we're not adding anything substantial to the parsing.

agoose77 (approved):

jpivarski:
Copied from Slack because this is worth remembering.

> While I was there, I noticed two things: there's no protection against an expected field appearing != 1 times (0 or more than 1), and in those cases, it will almost always fail with "buffer too small" (unless two JSON objects happen to exactly cancel out). The documentation says that we only assume that the JSON matches its schema, we don't check it, and that would be an example of the JSON not matching its schema. I could put in more mutable state to protect against this case (and an arrow pointing to the first JSON object that fails), but that's putting in more integer-twiddling at runtime.
>
> The other thing I noticed is that we search for matching key strings in each JSON object by iterating over a list of expected strings, in order, each time. That can be made faster by adding a mutable integer of the last index seen + 1 mod the number of keys, so at least if the keys of the JSON objects are always in the same order (a common case), the first one we check will be the right one. This can matter for JSON objects with a large number of expected keys. When I was first writing it, I considered making the key lookup a tree-search, so that lookup is always _O(average length of key)_ instead of only when the keys are in order, but that's a lot of complexity and is probably overkill. (Also, if the pointers of the tree don't point to nearby positions in memory, performance can be worse, not better.)
>
> So, it can be improved in protection against wrong JSON and it can be improved in performance for large numbers of JSON object keys.

martindurant:
Apparently, keys *should* (i.e., recommended, but not enforced) be unique https://stackoverflow.com/a/23195243/3821154

PR #1773, opened 2022-10-06
feat: add new `"fields"` key to `RecordForm`

### LINKED ISSUE #1766: `Form` is not guaranteed to preserve record field ordering under serialisation
### Version of Awkward Array

main

JSON does not define the object key ordering. We should modify the form for `RecordArray` to make it preserve the order, by defining a new `"recordlookup"` or `"fields"` field that stores the ordered field names. When this field is found, the `"contents"` field should be a JSON array, rather than an object.

agoose77:
This PR fixes #1766 by introducing a new `RecordForm` key: `fields`:
- If present, the `fields` key indicates that Awkward should read the `RecordForm` as a new-style (non-legacy) `Form`.
- If users mix form styles, Awkward will raise a `ValueError`.


agoose77:
This PR will prevent v2 _record_ (not tuple) forms from being read properly by v1. This is not ideal, but I think we _do_ need to fix this. If we were keen on being able to read v2 forms in v1, then we could hold off on changing form _serialisation_ until a later PR; allowing v2 to _read_ this new schema so that in future we can switch painlessly. However, I don't think we are as worried about `v2`→`v1` as the other way around.

Additionally, we don't _have_ to require that the entire-form is a single style. I just think that seems like a good way to catch accidental uses of this (safety).

src/awkward/forms/form.py
```diff
@@ -58,22 +71,53 @@ def from_dict(input: dict) -> Form:
...
+        if (
+            last_is_legacy_record is not None
+            and is_legacy_record != last_is_legacy_record
+        ):
+            raise _errors.wrap_error(
+                ValueError(
+                    "encountered an old-style RecordArray form after a new-style RecordArray form. "
+                    "Forms should not mix RecordArray styles"
+                )
+            )
```
jpivarski: Do we, though? Since we can detect the difference between old-style and new-style, why raise an error when they're mixed? It's true that a writer is much more likely to produce only old or only new, but a rule is easier to state (and implement) if it's memoryless.

For comparison, note how `from_dict` also accepts

   * an array class string as `"ListOffsetArray"`, `"ListOffsetArray32"`, `"ListOffsetArrayU32"`, or `"ListOffsetArray64"`. The last four of these are v1 class names, so that v2 will accept a Form-JSON made by v1 _or_ Form-JSON made by someone else who learned the format by looking at what v1 does. (That's why formats are hard to change: they become cultural products, and you can't know who everybody is that you need to reach to inform them of the update.)
   * an isolated string instead of a dict for a `NumpyArray` of defaults. This was to make the JSON more readable for records with lots of fields (NanoAOD), but most uses of Forms nowadays need `form_keys`, so this isn't a useful simplification.

In these other backward-compatiblity motivated features, we don't keep any memory of which format was used, and the `NumpyArray` one could oscillate back and forth based on whether some fields have non-default features, like `inner_shape`. So for consistency, let's scale back on the assertion here, since we don't make assertions in these other cases.

Without the error-checking, this is what the rule (as implemented) looks like to me:

   * if `fields` is absent (old style),
     * if `contents` is a JSON object (dict), produce a RecordArray with field names
     * if `contents` is a JSON array (list), produce a RecordArray without field names (i.e. tuple)
   * if `fields` is present (new style),
     * if `fields` is a list of strings, require `contents` to be a list with the same length and produce a RecordArray with field names
     * if `fields` is `null`, require `contents` to be a list and produce a RecordArray without field names (i.e. tuple)

This looks like a good rule to me!

Some of those assertions are implemented by the RecordForm constructor (`fields` and `contents` must have the same length), but be sure that the RecordForm constructor doesn't take a dict, assume it's a list, and iterate over its keys.

We don't get to say that "missing `fields` is equivalent to `fields` equal to `null`" because we need to disambiguate old style/new style and record/tuple. But anyway, that's only a rule for parameters; it's not a rule that's used anywhere else in Form-JSON (and this sets the standard; for consistency, we can't introduce a `null` ~ `undef` equivalence in Form-JSON in the future). That's fine.

tests/test_1766-record-form-fields.py
```diff
@@ -0,0 +1,209 @@
...
+
+    with pytest.raises(ValueError, match=".*Forms should not mix RecordArray styles.*"):
+        ak.from_buffers(
+            form,
+            1,
+            {
+                "node2-offsets": np.array([0, 1], dtype=np.int64),
+                "node3-data": np.array([0], dtype=np.int64),
+            },
+        )
```
jpivarski: This is a test of the assertion that I believe should be removed. So this test could just be converted into a test that it generates the right output, even though one is old-style and the other is new-style.

jpivarski (changes requested):
Just one requested change (described below): we don't need to raise an error if old-style and new-style RecordArray `contents` are mixed.

jpivarski:
Overall, inspired by the point @henryiii made on this morning's call, this change in the Form-JSON format could be accompanied by the writer's Awkward version number:

```python
"version": ak.__version__
```

or a protocol compatibility number, which is more granular than Awkward versions (like the ABI versions in pybind11, or Python's pickle protocol numbers):

```python
"format-compatibility": "v2"
```

(starting with `"v2"` because this is Awkward v2). Between these two, I prefer the latter because

   * this Form-JSON format should evolve very slowly, if at all
   * non-Awkward writers of Form-JSON need to know what to put there.

@henryiii mentioned the JSON format for hist, but while hist gets new features all the time, Forms intentionally don't. The second point about what non-Awkward writers should do comes up because Uproot is similarly at a loss to say what "version of ROOT" it is when it writes a file. It's semantically off-kilter.

It's hard to make an argument that a format is going to be slowly changing, if at all, at a moment in time when you're changing it. But here's the big picture:

   * June 2018 ‒ April 2020: Awkward v0 had very active JSON metadata, [with a version number](https://github.com/scikit-hep/awkward-0.x/blob/dd885bef15814f588b58944d2505296df4aaae0e/awkward0/persist.py#L291). It was active in the sense that it could run arbitrary Python code: the JSON was one-to-one with LISP S-expressions that picked out functions in the codebase and ran them, so that improvements could be made in metadata that would work in old versions of Awkward. Then I started worrying about metadata, added a whitelist, and it was pretty bad.
   * April 2020 ‒ June 2021: the Form-JSON format created for Awkward v1 was first implemented (in C++) in 1c18425339c99df9ecf9c49b0a27b6faa22d651c, and since it matches the structure of the Forms themselves (set of constructor arguments/properties defining the "dataclass"), which also hasn't changed, it's been stable. Replacing `ak.from_arrayset`/`ak.to_arrayset` with `ak.from_buffers`/`ak.to_buffers` (#592, introduced in version 1.0.1, old was removed in version 1.1.0) was a change in how pickling works because the length of the array is now recognized as necessary information to reconstruct an array, but that didn't affect the Form-JSON.
   * June 2021 ‒ now: Awkward v2 ignores the field "`has_identities`" and looks for an optional field "`has_identifier`" because the name has changed (#914). Identities/Identifier hasn't actually been implemented, so if a Form has this, it raises a NotImplementedError.
   * Now, October 2022, you're adding a new way to write RecordForm as JSON, but we can still read the old way.
   * I'm thinking of finally giving up on ever implementing Identifiers, and then "`has_identifier`" would become an ignored field.

As you can see, after v0, this Form-JSON format has not evolved very much and it's not expected to evolve in the future. We have complete forward and backward compatibility from April 2020 until now (breaking the ability of v1 to read v2 RecordForms), because the other changes were for optional fields or peripheral, not directly affecting the JSON-Form format.

An optional

```python
"format-compatibility": "v2"
```

could be prudent, in a preparing-for-unseen-disasters sense, not an expectation of ever having to use it. However, Form-JSON doesn't have a top-level; it's self-similar. If we were to add this, it would have to be added in every node.

Given that needing this is such a long shot and we do have format compatibility from v1 to v2, I would vote against adding this format-compatibility marker.

src/awkward/forms/form.py
agoose77: My motivation here was mainly to push people into using the new schema, if they were hand crafting forms. However, that in itself should not be the majority of our users. I don't have a strong reason to vote for it, so I've reverted the corresponding commit!

jpivarski (approved):

agoose77:
@jpivarski you've written a good summary of the version discussion and its pros and cons. Another option is to go for a monotonic integer that is deliberately meaningless - it just carries enough information for us to define epochs in the metadata space. As you say, we really aren't likely to change our metadata in a backwards incompatible (can't read old files) way, and although this is a forward incompatible change, it's the only one in some time to resolve a longstanding bug. Formats like protobuf have schema versions because they do change _all_ the time.

agoose77:
@jpivarski if you wouldn't mind checking over these last commits I'd be much obliged. If they're good, you're welcome to merge!

src/awkward/forms/form.py
```diff
@@ -65,12 +67,22 @@ def from_dict(input: dict) -> Form:
         )
 
     elif input["class"] == "RecordArray":
-        if isinstance(input["contents"], dict):
+        # New serialisation
+        if "fields" in input:
+            if isinstance(input["contents"], Mapping):
```
jpivarski, line 72: Note: since this is coming from `json.loads` or equivalent, we can make a hard assumption that JSON objects are dicts. Being more general doesn't hurt, but it's why the rest of this function is checking `isinstance(_, list)` and `isinstance(_, dict)` everywhere.

jpivarski:
Yes, this is good and I'll merge it now.

Regarding the `format-compatibility` marker, I'd rather not include it for the reasons given, and if there's no objection, then there's no action to take.

src/awkward/forms/form.py
agoose77: This warning is different to the rest of the function body, which (as you point out) is much stricter with types.

In reality, this is a high level functions that users can call. My temptation is to relax the constraints from `dict` to `Mapping`, and `list` to `sized_iterable`. How do you feel about that?

henryiii:
Quick, not looking at this in detail, comment:

> Another option is to go for a monotonic integer that is deliberately meaningless - it just carries enough information for us to define epochs in the metadata space.

This is what boost-histogram does. The first item in the tuple is a monotonic number and we increase it by one if we add a field to the pickle, and the `__setstate__` checks that number and knows how to read all the past versions of pickles. The original "0" version of pickles was a dict, so if it's a dict, we know it's the old style.

https://github.com/scikit-hep/boost-histogram/blob/00105a785f0a27c3718539078f9db35c92537e54/src/boost_histogram/_internal/hist.py#L612-L645

jpivarski:
The "format-compatibility" string and monotonic integer both decouple the format version from the code version, which I think is good. (I see that pybind11's ABI version and pickle's protocol version both do this.)

To fill you in, @henryiii, on what went wrong with our pickle compatibility: it had nothing to do with the format changing, since the format didn't change, so a version marker wouldn't have helped. The problem was that the state tuple contained a Python class:

https://github.com/scikit-hep/awkward/blob/f2092ee90e78c9ce00c54b3f5d9612bd218336ee/src/awkward/highlevel.py#L1455-L1462

where `form` is an instance of `ak._ext.forms.Form`, which no longer exists. When Python unpickles, it tries to create instances of the classes (compiled extension classes, actually), and that's where it fails.

What I wanted and intended to do, way back in April 2020, was to use `form.to_json()` there (or at least `form.to_dict()`), and convert back in `__setstate__`. That way, any pickles made from any version of 1.x would be readable in 2.x because the pickle state is made entirely out of Python builtins. But I somehow goofed that up and forgot.

We _could_ release 1.x patch that fixes this, but that only fixes pickles made with the patch release, not all the ones made since the format was introduced, which is the point of long-timescale compatibility.

We could also introduce dummy classes with the right names for Python to fill, then immediately convert them over to our actual classes, but this is not maintainer-friendly. I think it's a better overall decision to give up on pickle compatibility, rather than do that.

It's frustrating because just one line of code and a time machine are needed to fix it. That's all!

agoose77:
> We could also introduce dummy classes with the right names for Python to fill, then immediately convert them over to our actual classes, but this is not maintainer-friendly. I think it's a better overall decision to give up on pickle compatibility, rather than do that.

Yes, we could introduce shim types that just define `__setstate__` according to `main-v1`. It would actually not be too much code, but I'll let you be the guide on whether any code is too much for something people might not need.

jpivarski:
We can revisit it if somebody complains.  `:)`

PR #1575, opened 2022-08-01
docs: fill out missing sections

agoose77:
This PR builds upon #1563 to augment the empty user-guide chapters. It addresses some of the tasks in https://github.com/scikit-hep/awkward/projects/7#card-84465881

Flowchart of PRs in this project:
```mermaid
graph LR;
integrate[Integrate JB & Sphinx]
improve[Improve UX of integration]
content[Expand user-guide content]
integrate --> improve --> content
style content stroke:#f66,stroke-width:2px
click integrate "https://github.com/scikit-hep/awkward/pull/1552" "Integration PR"
click improve "https://github.com/scikit-hep/awkward/pull/1563" "Improvement PR"
click content "https://github.com/scikit-hep/awkward/pull/1575" "This PR"
```

https://awkward-array.readthedocs.io/en/agoose77/docs-expand-content

- [x] how-to-restructure-zip-project.md
- [x] how-to-restructure-add-fields.md
- [ ] how-to-restructure-rename-records.md
- [ ] how-to-restructure-concatenate.md
- [ ] how-to-restructure-sort.md
- [x] how-to-examine-list-fields.md
- [ ] how-to-math-numpy.md
- [ ] how-to-use-in-numba-features.md: would close #1064
- [ ] how-to-create-unflatten-group.md
- [ ] how-to-combinatorics-cartesian-combinations.md
- [ ] how-to-math-reducing.md
- [ ] how-to-examine-simple-slicing.md
- [ ] how-to-combinatorics-best-match.md
- [ ] how-to-filter-num.md
- [ ] how-to-math-gpu.md: would close #753 and #1139 (only `ak.num` and ufuncs work right now)
- [ ] how-to-filter-masked.md
- [ ] how-to-math-argminmax.md
- [ ] how-to-create-lazy.md
- [ ] how-to-examine-checking-validity.md
- [ ] how-to-math-broadcasting.md
- [ ] how-to-filter-jagged.md
- [ ] how-to-use-in-numba-arraybuilder.md: would close #528
- [ ] how-to-examine-type.md: would close #1366
- [ ] how-to-math-statistics.md
- [ ] how-to-filter-cut-mask.md
- [ ] how-to-examine-single-item.md
- [ ] how-to-specialize-subclass.md
- [ ] how-to-specialize-in-numba.md
- [ ] how-to-specialize-override-numpy.md
- [ ] how-to-specialize-lorentz.md

[...]

PR #1776, opened 2022-10-08
docs: merge new docs into main

agoose77:
This PR merges the new docs (#1552, #1563, #1590, #1591, #1592, #1594, #1630, #1734) into `main`.

As discussed elsewhere, these aren't finished v2 docs. But, they lay the groundwork for us to start iterating. #1575 will eventually target `main`, and fill out missing sections, in addition to the comments raised by @jpivarski in #1734 

By merging into `main`, however, we can avoid needing to keep these two branches in sync!
- [ ] Fix cache of WASM build
- [x] Add Conda (Forge) icon
- [x] Restore API reference ordering
   - [x] Remove duplication on index by having subcategories in ToC?
- [x] Mention records and unions in "what is an Awkward Array"
- [x] Move "Differentiation with JAX" to specialised group?
- [x] Address https://github.com/scikit-hep/awkward/pull/1734#issuecomment-1273813025


jpivarski:
> * I vaguely recall a conversation with @jpivarski in which he suggested that we can't, because `latest` is in many of the pre-shared docs links.

I have shared a lot of links with the form

```
https://awkward-array.readthedocs.io/en/latest/_auto/ak.{FUNCTION}.html
```

where `{FUNCTION}` is a top-level Awkward function (or class). That's because ReadTheDocs has, until now, defaulted to `latest` and I copied the link that was there. (It was often done quickly as part of a how-to explanation in chats, emails, GitHub, etc.)

However, I don't think that should prevent us from using the `latest` and `stable` labels in a meaningful way: with `latest` referring to any code merged into `main` (meaning v2 right now) and `stable` referring to code from the most recent non-RC release (meaning v1, version 1.10.1 right now). I had always expected that the links I shared would lead to newer versions of the documentation when they came out, and links for functions that are removed (virtual arrays, partitioned arrays) would become broken links. Introducing the `latest`/`stable` distinction does not change that.

An argument could be made for keeping versions of the documentation for every version of the code, in which `latest` and `stable` are moving symbolic links. Projects that have done this suffer from another issue: people share versioned links so much that search engines like Google return links to old versions in response to search queries. To fix _that_ those projects then add a banner at the top of old documentation, pointing to `stable`. That would mean keeping the old documentation forever but not in its original form: as soon as a version gets out of date, it needs to be modified to have the redirect banner. Unless that's automated by some Sphinx feature, I don't think we should reinvent it. Awkward is still a small enough project that we can just point to two versions of the documentation, `latest` and `stable`. (There's [an alternative](https://web.archive.org/web/20220305172522/https://awkward-array.org/quickstart.html) if someone absolutely needs an old version.)

If we're going to merge `docs` into `main` before the December release, it will have to be handled carefully. Is it your intention to deploy a v2 `latest` while `stable` remains v1? Then we'd be testing the ReadTheDocs mechanism that updates `latest` from `main`, but we wouldn't be testing the ReadTheDocs mechanism that updates `stable` from a tag until the December release.

Perhaps both could be tested if we do it independently of this PR. Suppose we configure ReadTheDocs to update `latest` from any change of `main` and `stable` from a tag and then release 1.10.2 (which has some updates to deliver, anyway). Then we should see a `stable` build get triggered, and subsequently changing `main` should trigger a `latest` build. Then we'd be confident that this mechanism is working on a low-stakes change of website. After that, merging this PR into `main` would trigger a radically new `latest` without changing `stable`, and we'd know that the `stable` update will work in December, when everything else will be in flux. It would be good to have one fewer thing to worry about.

There's also the issue of the Netlify site that `https://awkward-array.org` points at. It provides content that is needed by `stable` until December, but `latest` already has that content (in a more complete form, since you've been filling in a lot of the TODOs). The Netlify site will have to be undisturbed until the December release, including the DNS that points `awkward-array.org` at it. Redirecting the `awkward-array.org` domain name from Netlify to ReadTheDocs can only happen once.

Also, a lot of users probably have bookmarks to `latest`. They'll see new documentation and may be confused about whether the v2 transition has happened already. Furthermore, a lot of the documentation will be correct (because the high-level functions are almost entirely unchanged), but some will be wrong: a new page describes `ak.contents.Content`, but that name does not exist in Awkward 1.10.

Orrrrrr... we could just wait to merge this PR into `main` in December. It's disappointing because there's a lot of good material to see in the `docs` branch, and the `docs` branch is necessarily drifting far from `main`. But many of the above complications go away if we just release the new docs at the same time as the new code.

agoose77:
I don't love versioned documentation. I recognise that for some projects it is necessary, but I would need more fingers than I have available to count how many times I've accidentally landed on the wrong documentation and visually filtered out the warning banners.

> Is it your intention to deploy a v2 latest while stable remains v1? 

Yes, I would ideally like for us to be deploying the docs from GitHub so that we can ensure they're correct. Otherwise, I think we'd need to e.g. maintain a `docs` branch that regularly does something like `git reset --hard origin/main` solely for the purpose of providing a branch for RTD. I'm not aware of any mechanism to build one branch under another name, but I might not have looked hard enough?

> (in a more complete form, since you've been filling in a lot of the TODOs)

Well, in these PRs this content isn't here; it's been tricky to keep rebasing everything, so I've held off any newer changes until we had finalised the v2 merge.

If we've mainly released `latest` into the wild, then for now we'll need to respect that. Maybe we should move to a policy of sharing `stable` moving forward, so that in some future release we can worry about this less? In the meantime, let's just not _deploy_ docs from `main`, i.e. we just stop deploying *any* docs changes across any version until December. I think this is reasonable; v1 should not gain any new features IIRC.

jpivarski:
It will be a lot easier to share the `stable` link if that's the default. Right now, `latest` is the default. With `latest` == `main` and `stable` == last tag, the `stable` one should be the default.

jpivarski:
(We talked on a Google Meet about how to go forward: @agoose77 will be removing the `main`-update → ReadTheDocs trigger, and then we can merge the `docs` branch into `main` without disturbing what documentation the users see. The documentation update will be a big bang when Awkward 2.0.0 is released, to keep things simple.)

agoose77:
@jpivarski are you happy for this to now merge into `main`, given that I've disabled the step that triggers RTD on `main`, and disabled the RTD GitHub webhook?

jpivarski:
Yes, because the triggers have been disabled and v1 users won't see anything different.

However, I don't have a way to see the rendered documentation: https://awkward-array.readthedocs.io/en/docs/index.html hasn't been updated with #1734 yet. In general, how are we going to see the rendered documentation as it develops? We'll want to point other reviewers to it when we've finished iterating.

agoose77:
@jpivarski - yes, realised that it would make more sense to just disable docs push on `main`, and then we can still see changes in pushed workflows. There *might* be a quirk with `ref_name` not always doing what it should, so I'll keep an eye on things. I'm hoping that 1057232 appears in https://awkward-array.readthedocs.io/en/docs/index.html in the next ten minutes or so.

The plan is to have branches continue to build docs, if we enable them on RTD. The `docs` branch is already active on RTD (hidden, though), so we can use it as a "I want to make a docs PR that people can see" branch If that's easy?

agoose77:
https://awkward-array.readthedocs.io/en/docs/

agoose77:
I've removed "What is an Awkward Array" button:
image

It's been replaced with a link to the "Try Awkward Array" page. This page is now a retrolite notebook, which is effectively the old "What is an Awkward Array" notebook:
image

The original "What is an Awkward Array" page has therefore also been removed, in favour of the "Overview" subsection of "Getting Started", which includes more detail (including unions):
image

agoose77:
@jpivarski the v1 docs ordering stemmed from the fact that the `operations.py` module had pre-ordered functions (i.e., `from_buffers` follows `to_buffers`). It also ordered "convert" vs "structure", etc. We have disposed of this level of categorisation within our source tree. Could you clarify what kind of ordering you would like to see here?

jpivarski:
> the v1 docs ordering stemmed from the fact that the `operations.py` module had pre-ordered functions

I see, that is an issue. (I had been thinking there was an ordered list of functions in the `prepare_docstrings.py` script, but maybe that's Uproot.) Although I suggested a solution, if it's possible to group items in the left-bar while also not hiding them.

The "Navigation" content in the front page of API reference puts high-level functions and classes into groups. It's expressed as text (markdown) and it's already visibly out of date: functions like `ak.to_awkward0` no longer exist. If those groupings were instead expressed as left-bar groups, in YAML or Python code that generates YAML, then the functions within each group can be left as alphabetical; the order doesn't matter in a group of length 2‒5, as most of these are. If it's Python code generating the left-bar YAML, then it can make an "other" category for all of the functions that are discovered via `dir(ak.operations)` but not explicitly classified. We might add a function and forget to categorize it, but it will be visible in the docs and seeing it in the "other" category will be a reminder to add it to a category later.

(The URLs for the `ak.*` functions shouldn't depend on the grouping, so that we can move them from one group to another at any point in the future. In fact, the URLs should be the same as in v1, so that all the shared links, like `https://awkward-array.readthedocs.io/en/latest/_auto/ak.num.html`, still work.)

agoose77:
> (The URLs for the ak.* functions shouldn't depend on the grouping, so that we can move them from one group to another at any point in the future. In fact, the URLs should be the same as in v1, so that all the shared links, like https://awkward-array.readthedocs.io/en/latest/_auto/ak.num.html, still work.)

The v1 URLs are mapped onto v2 URLs by redirect (client side). This handles the clean-up in docs structure, but also things that have been renamed e.g. `/ak.layout.Record.html` -> `/reference/ak.record.Record.html`

> The "Navigation" content in the front page of API reference puts high-level functions and classes into groups. It's expressed as text (markdown) and it's already visibly out of date: functions like ak.to_awkward0 no longer exist. 

FYI, in the `docs` branch, this is now mostly up date: https://awkward-array.readthedocs.io/en/docs/reference/index.html

> Although I suggested a solution, if it's possible to group items in the left-bar while also not hiding them.

Cool, so we're going for similar categories to the current index? I'll try to keep it not-too-busy, let's see what happens ...

agoose77:
The easiest thing to do here is manually write the toctree, and warn the user if anything didn't appear there that was found by auto generation. The initial categories are a WIP, so feedback is welcome.

The new reference index / ToC will be under https://awkward-array.readthedocs.io/en/docs/reference/index.html when complete

agoose77:
image

jpivarski:
It looks really good!

We'll have to stop ignoring the docs warnings, then.  :)

agoose77:
> We'll have to stop ignoring the docs warnings, then.  :)

Yes, this is a goal of mine - once we get things in place, we start fixing the warnings!

agoose77:
@jpivarski do you prefer the badges, or the "try it now" link?
image
vs
image



I'm not really sure what's happening with `ccache` here. "it works in my environment" :cry: It might be that we merge this without being able to cache the wasm-build for now.

jpivarski:
I think the first one, with the badges, is more useful. It doesn't take more space, and the added complexity is managed via design: they're all on the same line, so if a reader doesn't know what they are, it's clear that they're all the same sort of thing and can all be ignored together.

The try-in-browser is amazing/wonderful, but it will need two improvements to be truly useful: (a) Pyodide needs to load more quickly. We can't make that happen, but that will happen with time. (b) We need to direct a user with specific things to try, not just an empty box. Ideally, the "What is an Awkward Array?" page would be executable (maybe need to wait for Thebe-Pyodide integration?) with hints about things to try beyond the thing that is actually executed by just shift-entering the cells.

We can have both, too. The link to in-browser executable Awkward Array can be a fourth badge, and to keep the width narrow, the text can just be "Try it!"

I see that the badge colors are themed, and we're lucky enough that they're all different from each other. Conda likes lime green, PyPI likes deep blue, GitHub likes grey, and there's no constraint on the color of the "Try it!" button/badge. How about shocking orange? It would draw the eye right at it (and "Try it!" is small enough that it won't burn your eyeballs).

agoose77:
@jpivarski just for clarification, are you aware that the "try in browser" is now a Jupyter notebook? It's not _quite_ thebe - it basically runs a minimal UI JupyterLab (RetroLab) in an IFrame.

image

jpivarski:
No, this is new since I last looked. Cool! It's exactly the point (b) I was describing: so that part is done. Point (a) is something we just need to wait for, so this is as ready as we can make it.

I vote for the 3 badges + 1 "Try it!" badge that goes to this "What is an Awkward Array?" page.

agoose77:
I've added a "Try It" in the form of a `shields.io` badge:
image

agoose77:
Currently Sphinx is not generating a primary sidebar for the API-reference children. Investigating.

agoose77:
@jpivarski this is ready to go in my view. I haven't solved the `ccache` issue, but I think a 10-minute build is acceptable for now (docs won't be a required check yet), and we can fix it down the road.

https://awkward-array.readthedocs.io/en/docs/index.html

jpivarski (approved):
It looks pretty good on a phone, too: [http://mobiletest.me/iphone_5_emulator/?u=https://awkward-array.readthedocs.io/en/docs/index.html](http://mobiletest.me/iphone_5_emulator/?u=https://awkward-array.readthedocs.io/en/docs/index.html)

(Though the badges/buttons could have a top or bottom margin to not be stuck together when the window gets very narrow.)

When I ran the "Try it in your browser" in the simulated phone first and then later in a regular browser tab, the `piplite.install("awkward")` line failed with

```
File /lib/python3.10/site-packages/micropip/_micropip.py:177, in WheelInfo.validate(self)
    175 sha256_actual = _generate_package_hash(self.data)
    176 if sha256_actual != sha256_expected:
--> 177     raise ValueError("Contents don't match hash")

ValueError: Contents don't match hash
```

but I'm guessing that it's related to the mobile test. Reloading fixed it. (This page might need to accumulate some hints, like trying "Run → Run All Cells" and reloading if there are any problems. It's nice to see that reloading restores the original state, but a visitor might want to know that would happen, so they don't write too much code in this notebook without saving it.

It would be nicer if the JupyterLite scaled with the available space, but I know that's asking for a lot. It probably isn't possible.

Leaving the "Try it in your browser" page makes Chrome say that work may be lost. It would be ideal to get rid of that message (we've already decided that deleting data upon reload is the policy), but, failing that, it could be one of the messages on the page.

We want to keep those messages at a minimum, but:

>   * To run the code, click on the first cell (gray box) and type shift-enter through the cells.
>   * Or select "Run All Cells" from the "Run" menu.
>   * Feel free to experiment, but if you need to restore the original code, reload this browser page. Any changes you make will be lost when you reload!
>   * When you leave the page, you might get a "Changes you made may not be saved" message, which you can ignore.

I really like how the front page of the User Guide has all of the contents expanded out. That facilitates immediately jumping to a sub-section without having to know which section it's hidden in, but the left-bar can remain collapsed by default.

I really, really like the API reference.

It's odd to have any `ak.*` _functions_ after the "Low-level layouts" section. For instance,

  * "Converting between backends": arguably low-level, may belong here
  * "Indexing and grouping": these are high-level functions that impact the high-level view (i.e. you don't need to know about layouts to have reason to use these functions)
  * "Copying and packing arrays": arguably low-level, though someone working at a high-level may really appreciate `ak.packed`.

The "Additional documentation" can just be removed; the diagram is wrong and it's not serving a purpose anymore.

Putting the AwkwardForth documentation last is good, though we also need the kernel documentation, which should be the very last. That kernel documentation (currently [not generated](https://awkward-array.readthedocs.io/en/docs/_auto/kernels.html)) has more information than can be found in the C++ API reference because it comes from the [kernel-specification.yml](https://github.com/scikit-hep/awkward/blob/main/kernel-specification.yml).

The nested sections in the AwkwardForth documentation and, to a much greater degree, the kernel documentation would flood the front page of the API reference if they were included. These two, at the end, can be an exception to the inlining of nested sections in `reference/index.html`.

I like how you have the "C++ documentation" called out in a box at the end, but at the end. With "Additional documentation" gone, it would be the last thing on `reference/index.html`.

That's it! This is all very nice. My new requests don't have to be part of this PR; you can merge `docs` into `main` and I'll bulletize them and make them an issue-request.

agoose77:
> It would be nicer if the JupyterLite scaled with the available space, but I know that's asking for a lot. It probably isn't possible

We could add a custom styling to this - the current iframe is set 600px, but we could do something more dynamic to constrain it between sensible defaults.

>  It's nice to see that reloading restores the original state, but a visitor might want to know that would happen, so they don't write too much code in this notebook without saving it.

We can actually save the notebook; I went to pains to disable this. I couldn't find a satisfactory way to clear the caches reliably (it should just be "Clear site data", so maybe I'm missing something obvious), so I disabled the storage backend.

> When I ran the "Try it in your browser" in the simulated phone first and then later in a regular browser tab, the piplite.install("awkward") line failed with

This is annoying; there's a serviceworker involved and I think it's caching the old wheel sometimes. Refreshing the page does seem to fix it.

jpivarski:
> We can actually save the notebook; I went to pains to disable this. I couldn't find a satisfactory way to clear the caches reliably (it should just be "Clear site data", so maybe I'm missing something obvious), so I disabled the storage backend.

It's appreciated! We don't want it to look like the tutorial is broken because a visitor has a non-standard version of it in their browser cache. That ought to be a common use-case for JupyterLite.

agoose77:
> We don't want it to look like the tutorial is broken because a visitor has a non-standard version of it in their browser cache. That ought to be a common use-case for JupyterLite.

Right, that was my worry - even *I* found myself confused sometimes, and I was midway through introducing it :rofl:

agoose77:
I'm keeping this branch open so that we can preview it on RTD :) The idea is now that we'd target `main` with PRs (and look at their per-branch previews), but periodically force push to `docs` to trigger an RTD build that's up-to-date

PR #1781, opened 2022-10-11
fix: use the flag for non-persistence

ianna:
Using the flag for non-persistence of the PyObject lookup pointers prevents the crash.

### DIFF
--- a/src/awkward/_connect/cling.py
+++ b/src/awkward/_connect/cling.py
@@ -149,7 +149,7 @@ namespace awkward {
     ssize_t stop_;
     ssize_t which_;
     ssize_t* ptrs_;
-    PyObject* lookup_;
+    PyObject* lookup_; //! lookup
   };
 }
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is different from what I thought you were saying (in person). So the `//!` is something you needed to _add_ to get it to work. I'm willing to accept that this is ROOT magic that handles the `PyObject*` more appropriately (correct lifetime: DECREFed once and only once when the RDataSource is out of scope).

Do the tests really test what this is changing? That is, without the `//! lookup` comments, do the tests fail?

agoose77:
I was curious about the `//!` markup, so I looked it up for my own benefit. Here's the link for anyone else finding this in the future: https://github.com/root-project/root/blob/07932d78211d0ac5fdb198ac2260c35530cb9647/io/doc/TFile/README.md#streamerinfo

PR #1782, opened 2022-10-12
feat: add study for JAX vmap integration

agoose77:


jpivarski:
As a study, this doesn't affect any other code. After 2.4.0 is done, let's merge it so that it's out of the open PRs list.

PR #1786, opened 2022-10-13
feat: add `cumsum` and `cumprod`

### LINKED ISSUE #1855: Support `axis=None` in `argsort` etc

Currently we don't support this argument value.

agoose77:
The mechanisms by which list-types compute reductions, unique values, and sorting, are all very similar (and involve many of the same kernels). This PR identifies a "transformer" class to complement reducers as operations that leave the array structure intact; the result can be trivially broadcast against the original array.
In doing so, it paves the way for the addition of `nancumsum`, `cumsum`, `nancumprod`, and `cumprod` functions. 

Unfortunately, whilst `np.unique` _looks_ like a transformer, the fact that it changes the shape of the result (keeping only unique values) means that it requires some additional code to implement. It seems cleaner therefore to treat `unique` as a separate case. In general, it's not clear how to reconstruct `None` values in the event that the transformation changes the length of the array. 

The string special-casing is now even more obviously distinct from the normal transformer handling.

- [x] Refactor to use transformers
- [ ] Add `cumsum`
- [ ] Add `nancumsum`
- [ ] Add `cumprod`
- [ ] Add `nancumprod`
- [ ] Closes #1855

### DIFF
--- a/kernel-specification.yml
+++ b/kernel-specification.yml
@@ -5513,6 +5513,109 @@ kernels:
     automatic-tests: true
     manual-tests: []
 
+  - name: awkward_transform_cumsum
+    specializations:
+      - name: awkward_transform_cumsum_int32_int8_64
+        args:
+          - {name: toptr, type: "List[int32_t]", dir: out}
+          - {name: fromptr, type: "Const[List[int8_t]]", dir: in, role: reducer-fromptr}
+          - {name: offsets, type: "Const[List[int64_t]]", dir: in, role: transformer-offsets}
+          - {name: offsetslength, type: "int64_t", dir: in, role: transformer-offsetslength}
+      - name: awkward_transform_cumsum_int32_int16_64
+        args:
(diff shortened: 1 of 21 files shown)


agoose77:
`IndexedOptionArray` handling is still not quite right. The current use of `maintain_none_position` is just injecting `None`s on top of the existing ones, rather than actually re-indexing. This mechanism might need some more thought, as `needs_position` and `maintain_none_position` seem like mutually exclusive options; if you maintain the positioning of the array, does it make sense to have a positional transformer? Perhaps. I'll give it some more thought in the week. Should be quite solvable, but I'm tackling it by converting the existing code rather than just implementing from scratch. Refactoring like this sometimes raises questions that require a rethink.

agoose77:
I'm closing this for now. The code will be useful for future reference, but this PR won't be worked on for a while (there are at least two other big PRs to merge), so it will just stagnate. In the mean time, it's filling up the tracker.

PR #1792, opened 2022-10-15
fix: produce valid offsets for reducing with empty trailing lists

### LINKED ISSUE #1791: Reduction of inner dimension in ragged array produces incorrect offsets
### Version of Awkward Array

main

This *might* not be a bug, depending upon what we define the behavior of `ListArray` to be w.r.t `starts[i] == stops[i]`. The following reproducer produces a result layout with `starts[1] == stops[1] == 3`. The leaf `NumpyArray` has a length of 1, so unless we define out-of-bounds starts to be valid iff. they are equal to stops, then we're not doing the right thing here.

```python
import numpy as np
import awkward as ak

array = ak.Array(
    ak.contents.ListOffsetArray(
        ak.index.Index64(
            np.array(
                [0, 1, 1],
                dtype=np.int64,
            )
        ),
        ak.contents.ListOffsetArray(
            ak.index.Index64(
                np.array(
                    [0, 1],
                    dtype=np.int64,
                )
            ),
            ak.contents.NumpyArray(np.arange(1)),
        )
    )
)
# [
#     [
#         [0],
#     ],
#     [],
# ]

print(ak.sum(array, axis=1).layout)

```

agoose77:
This appears to fix #1791, and follows from what I'd expect the kernel to be doing. The fact that this wasn't right beforehand _seems_ to me like a human error rather than an oversight, so it gives me more confidence that this simple fix is correct.

- [ ] Add test for v1 (though the v2 test should catch it, as we only change the kernel)

### DIFF
--- a/src/cpu-kernels/awkward_ListOffsetArray_reduce_nonlocal_outstartsstops_64.cpp
+++ b/src/cpu-kernels/awkward_ListOffsetArray_reduce_nonlocal_outstartsstops_64.cpp
@@ -43,8 +43,8 @@ ERROR awkward_ListOffsetArray_reduce_nonlocal_outstartsstops_64(
   }
 
   for (;  k < outlength;  k++) {
-    outstarts[k] = lendistincts + 1;
-    outstops[k] = lendistincts + 1;
+    outstarts[k] = maxdistinct + 1;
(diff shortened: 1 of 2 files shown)


agoose77:
Superseded by #1796

PR #1793, opened 2022-10-16
refactor: simplify reducer API

### LINKED ISSUE #1789: `ak.sum([])` raises `IndexError`
### Version of Awkward Array

main

This is the flattening code-path. `EmptyArray` has no length, so we must be assuming otherwise here.

agoose77:
This PR:
- [x] Fixes #1789 
- [x] Fixes `UnionArray` index handling (this needs another pass, I just did a quick `sed`)
- [x] Renames the general `_reduce` method to enshrine it as the entrypoint for reducers
- [x] Removes the layout reducer shims e.g. `layout.all()`
- [x] Adds a `reduce_flattened` method that performs a two-stage reduction where possible (i.e. for non-positional reducers)
- [ ] Make `Count` use an integer as its preferred dtype.

There may be some glaring oversights here; I just cooked this up quickly whilst thinking about the problem.

## Description

Our existing high-level reduction functions all duplicate the same boilerplate to handle two-stage reduction and the existence of typetracer. Additionally, there are a couple of bugs e.g. with `ak.sum([])` raising an `Exception`, or `ak.argmin()` returning per-sublist indices. Overall, it would be good (and safer) to move this into a common implementation.

It also occurs to me that for the mild convenience of having `layout.all()` etc., we have a large number of shims that are also defined in `ak.operations`. This isn't a "bad" thing, but I think the proper interface here is the reducer API, and we should expose that so that users can (should they be crazy enough / have need) implement their own reducers and pass them in. (Yes, nothings every truly private in Python, but I think we should be confident in exposing our reducers).

This PR handles the thorny problem of two-stage reduction by allowing non-positional reducers to pass in _two_ reducers; one to reduce the partitions, and another to reduce the final result. For positional reducers, we would need to special case this. Hence, this PR introduces a contentious change; merge the flattened array parts into a single buffer for positional reducers like `argmin`. 

As of writing, the test suite still needs updating to reflect the removal of `layout.all()` etc.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-reduce-none/> once Read the Docs has finished building :hammer:


jpivarski:
Is #1789 considered a part of this PR? It's probably blocked by it, at least.

agoose77:
Yes, this PR solves that issue. It's not in a good state right now - my approach changed once the full scope of the PR became apparent. This PR is related to https://github.com/scikit-hep/awkward/pull/1849, although my brain right now can't tell you exactly how strict this dependence is. When I get back to this, it might be possible to merge this before https://github.com/scikit-hep/awkward/pull/1849

agoose77:
This has been superseded by #2020

PR #1795, opened 2022-10-16
fix: allow empty sublists in ak.run_lengths

### LINKED ISSUE #1794: `ak.run_lengths` does not handle leading/trailing empty sublists
### Version of Awkward Array

main

This also exists in v1.

The following code raises an `Exception`:
```python
import awkward as ak
import numpy as np


layout = ak.contents.ListOffsetArray(
    ak.index.Index(np.array([0, 2, 4, 4], dtype=np.int64)),
    ak.contents.NumpyArray(
        np.arange(4)
    )
)
ak.run_lengths(layout)
```

Also, we should use `regularize_numpyarray` to handle `NumpyArray`, which currently just raises an Exception.

agoose77:
This fixes #1794. We could do this more optimally with a kernel rather than multiple `nplike` operations, but it's clearly still linear in the number of indices.

@jpivarski let me know if you'd prefer that I don't introduce ASCII diagrams into the source. I find it's clearer than the equivalent wording, but it does add to the line lengths (indirectly harming readability).


jpivarski (approved):
Yes, thanks for fixing this! It looks right to me (a refinement of the `1:-1` slice; the diagram is very helpful).

PR #1796, opened 2022-10-17
fix: simplify `ListOffsetArray_reduce_nonlocal_outstartsstops`

agoose77:
- [x] fixes #1784
- [x] fixeds #1791
- [x] removes unused kernel
- [x] adds test

This PR fixes #1784 by rewriting the `awkward_ListOffsetArray_reduce_nonlocal_outstartsstops_64` kernel. The actual fix for #1784 is just to wrap the entire kernel body in the `if (distincts[i] != -1)` guard. However, I took the opportunity to simplify and explain the kernel. This means that we can drop the `gaps` kernel, and we don't need to perform any float divisions. So, I'd hope that this is also a slight performance boost (but this did not motivate the rewrite).

I *hope* that I haven't overlooked any design invariants or corner-cases :crossed_fingers:; I've attempted to understand the purpose and function of the kernel in depth before writing this PR.

:warning: I did change the expected outcomes of the test suite - some of the kernel tests check the offset values. We don't care about the actual values of `starts` and `stops` as long as they're in-bounds and equal.

PR #1797, opened 2022-10-17
fix: simplify `ListOffsetArray_reduce_nonlocal_outstartsstops` (backport)

agoose77:
See #1796


jpivarski (approved):
I confirm that this backport is equivalent to the v2 version, #1796.

But I should also add my test examples to this one, to keep them from _unnecessarily_ diverging.

jpivarski:
I haven't been able to assign required tests for this branch (`main-v1`): GitHub doesn't think that any of the above have ever been run. So I can't enable auto-merge.

I'll try to remember to come back here and squash-and-merge this PR, but if I don't remember to and the tests have all passed, you can do it.

Issue #1799, opened 2022-10-17
Requests from review of #1776 (merging docs into main).


jpivarski:
### Which documentation?

Tutorials site

### What needs to be documented?

This is a follow-up of https://github.com/scikit-hep/awkward/pull/1776#pullrequestreview-1144743439, which is an approval of merging the `docs` branch into `main` as-is, but with some to-do items to be picked up later. This issue copies those into a bulleted list.

  - [ ] Vertical margin around the badges/buttons on the front page so that they are still distinct when the window is narrow or someone is looking at it on their phone.
  - [ ] Helpful messages on the "Try it in your browser" page (see below).
  - [ ] Better sizing of Retrolite on small screens
  - [x] All of the sections containing `ak.*` functions (everything defined in `src/awkward/operations/`) should be contiguous, before the section on "Low-level layouts". This is so that someone searching for an operation they can't quite remember the name of is less likely to miss it. The "Converting between backends", "Indexing and grouping", and "Copying and packing arrays" belong to this category.
  - [x] Remove the "Additional documentation" section (which is unnecessary and out of date, including the diagram).
  - [x] The kernel specification should be generated and included.
  - [x] The AwkwardForth sub-sections may be suppressed on `reference/index.html` and the kernel specification sub-sections _have to_ be suppressed. They would flood that page.

That's it!

agoose77:
Yes, I'd like to improve the buttons. Ideally, I'd prefer to use something like the `sphinx-design` buttons, but they don't support the dynamicism that the shields.io badges do. We can just add a CSS margin to the icon class.

agoose77:
~The kernel specification _is_ being generated, at least locally. I'm not sure why it's not appearing in the built docs~

Ah, it's the change I made to use `runpy`. I'll fix this in the PR that closes this issue.

agoose77:
@jpivarski could you clarify where you think RetroLite could be better on smaller screens? :)

jpivarski:
Whatever I had been thinking, nevermind: I just tried out https://awkward-array.readthedocs.io/en/main/getting-started/try-awkward-array.html in a window with varying sizes and it scales well: it drops the right bar and then the left bar as the window gets narrow, and the width of the RetroLite widget uses all the available space in its column, which is good. Considering that there's nothing interesting in the right bar, it would be nice if that was always collapsed for this page, but I doubt you can make layout decisions like that on a per-page basis.

[...]

PR #1800, opened 2022-10-17
docs: fix kernel generation

agoose77:
This PR addresses one of the issues in #1799; fixing the kernel docs.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-touchups/> once Read the Docs has finished building :hammer:

### DIFF
--- a/dev/generate-kerneldocs.py
+++ b/dev/generate-kerneldocs.py
@@ -78,29 +78,6 @@ The functions are implemented in C with templates for integer specializations (c
                         + "\n\n"
                     )
 
-    if os.path.isfile(
-        os.path.join(
-            CURRENT_DIR,
-            "..",
-            "docs-sphinx",
-            "_auto",
-            "toctree.txt",
-        )
-    ):
-        with open(
-            os.path.join(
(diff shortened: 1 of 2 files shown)


jpivarski:
Same as https://github.com/scikit-hep/awkward/pull/1810#issuecomment-1282979494. I could run Sphinx locally and view the rendered HTML locally, but that's probably what you're doing, and the real test of what it looks like is what ReadTheDocs makes of it. It might be too cumbersome to make ReadTheDocs render every documentation-related PR or branch, but we could do a two-step PR of (1) develop in a new branch, (2) pull that into the `docs` branch, and (3) request the `docs` branch to be merged into `main`. The downside of this is that there can't be concurrent PRs (or I would see them all in their already-merged state), but the documentation isn't currently being developed in parallel. (And later, when things settle down, it would be mostly changes to the text, which doesn't have to be reviewed in-situ.)

agoose77:
@jpivarski this is now ready to go!

jpivarski:
Fantastic! I just took a look, and it looks great.

PR #1803, opened 2022-10-18
docs: remove `vector_design_prototype`'s reference

Saransh-cpp:
`vector_design_prototype` no longer exists in vector - https://github.com/scikit-hep/vector/pull/272. I have commented out the reference as there are plans to bring it back - https://github.com/scikit-hep/vector/issues/273 - in the future.


agoose77:
@Saransh-cpp could you rebase onto `main`, and target `main`? :)

Saransh-cpp:
I always end up targeting the wrong branch haha. Rebased!

agoose77:
This *only* touches docs, and looks like valid ReST. As the

PR #1804, opened 2022-10-18
test: use a temp dir for writting out the test files

### LINKED ISSUE #1802: Write output files into a temporary directory during tests.
See https://github.com/scikit-hep/awkward/commit/b315e6519eb78c8083e088e3ca065cfdb3af5d31#r87085384, quoted below. It's a follow-up on #1781.

These (lines [30](https://github.com/scikit-hep/awkward/blob/580f3fad01327a6b1d8c2e1edbaf324f35fb7796/tests/test_1781-rdataframe-snapshot.py#L30) and [47](https://github.com/scikit-hep/awkward/blob/580f3fad01327a6b1d8c2e1edbaf324f35fb7796/tests/test_1781-rdataframe-snapshot.py#L47)) need to use the `tmp_path` fixture, like

https://github.com/scikit-hep/awkward/blob/580f3fad01327a6b1d8c2e1edbaf324f35fb7796/tests/test_1619-from-parquet-empty-field.py#L13-L29

so that it doesn't leave `test-integers.root` and `test-listarray.root` files in the main directory after running.

Thanks!

ianna:
fixes https://github.com/scikit-hep/awkward/issues/1802

### DIFF
--- a/src/awkward/_connect/cling.py
+++ b/src/awkward/_connect/cling.py
@@ -145,11 +145,14 @@ namespace awkward {
     }
 
   protected:
+    // ROOT streamer customization is done by giving specific instructions
+    // in the comments written after the declaration of data members: the values
+    // of `ptrs_` and `lookup_` will be ignored (//!).
     ssize_t start_;
     ssize_t stop_;
     ssize_t which_;
-    ssize_t* ptrs_;    //! data pointer
-    PyObject* lookup_; //! lookup
+    ssize_t* ptrs_;    //! transient data pointer
+    PyObject* lookup_; //! transient lookup
   };
 }
 """.strip()
@@ -183,10 +186,13 @@ namespace awkward {
     }
 
   protected:
+    // ROOT streamer customization is done by giving specific instructions
+    // in the comments written after the declaration of data members: the values
+    // of `ptrs_` and `lookup_` will be ignored (//!).
     ssize_t at_;
     ssize_t which_;
-    ssize_t* ptrs_;    //! data pointer
-    PyObject* lookup_; //! lookup
+    ssize_t* ptrs_;    //! transient data pointer
+    PyObject* lookup_; //! transient lookup
 
   };
 }
--- a/src/awkward/_connect/cling.py
+++ b/src/awkward/_connect/cling.py
@@ -190,6 +190,7 @@ namespace awkward {
     // ROOT streamer customization is done by giving specific instructions
     // in the comments written after the declaration of data members: the values
     // of `ptrs_` and `lookup_` will be ignored (//!).
+    // https://github.com/root-project/root/blob/master/io/doc/TFile/README.md#streamerinfo
     ssize_t at_;
     ssize_t which_;
     ssize_t* ptrs_;    //! transient data pointer
--- a/src/awkward/_connect/cling.py
+++ b/src/awkward/_connect/cling.py
@@ -148,7 +148,7 @@ namespace awkward {
     ssize_t start_;
     ssize_t stop_;
     ssize_t which_;
-    ssize_t* ptrs_;
+    ssize_t* ptrs_;    //! data pointer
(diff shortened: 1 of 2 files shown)


ianna:
@jpivarski - Linux build failed to fetch `numpy` with a `Network is unreachable`.

src/awkward/_connect/cling.py
```diff
@@ -148,7 +148,7 @@ class ArrayView {
     ssize_t start_;
     ssize_t stop_;
     ssize_t which_;
-    ssize_t* ptrs_;
+    ssize_t* ptrs_;    //! data pointer
     PyObject* lookup_; //! lookup
```
jpivarski: Why is this needed? Is it a `//!` that was missed before? If so, then it's not directly addressing #1802, but (in an ideal world) should have been a separate PR (but I'm not going to require it).

src/awkward/_connect/cling.py
```diff
@@ -185,7 +185,7 @@ class RecordView {
   protected:
     ssize_t at_;
     ssize_t which_;
-    ssize_t* ptrs_;
+    ssize_t* ptrs_;    //! data pointer
     PyObject* lookup_; //! lookup
```
jpivarski: Same here.

jpivarski (approved):
This solves the issue, including a third ROOT file that I didn't notice. Go ahead and merge it! The two comments below are questions, not requested changes.

src/awkward/_connect/cling.py
ianna: I was trying to find out the reason of the [SystemError](https://github.com/root-project/root/issues/11602). Running the code in `lldb` showed the following error that is fixed by adding the comment: 
`Error in <TStreamerInfo::Build>: awkward::ArrayView, discarding: long* ptrs_, no [dimension]`

jpivarski: So just all of the class's pointers? Is there any particular meaning to "`data pointer`" versus "`lookup`", or can any word go there?

Maybe it would be best to include a comment explaining what these `//!` things do; a link to the issue you pointed to above would be a step, at least. We don't want a future maintainer to look through this and say, "I don't know what this comment means, so it's not useful as a comment" and remove it.

jpivarski: @agoose77 found this link:

https://github.com/root-project/root/blob/master/io/doc/TFile/README.md#streamerinfo

(https://github.com/scikit-hep/awkward/pull/1781#issuecomment-1282628108).

ianna: done

jpivarski (approved):
Thanks! It looks like you're done, so I'll just add the ROOT documentation link and then merge this PR.

src/awkward/_connect/cling.py
```diff
@@ -145,11 +145,14 @@ class ArrayView {
     }
 
   protected:
+    // ROOT streamer customization is done by giving specific instructions
+    // in the comments written after the declaration of data members: the values
+    // of `ptrs_` and `lookup_` will be ignored (//!).
```
jpivarski, line 150:
```suggestion
    // ROOT streamer customization is done by giving specific instructions
    // in the comments written after the declaration of data members: the values
    // of `ptrs_` and `lookup_` will be ignored (//!).
    // https://github.com/root-project/root/blob/master/io/doc/TFile/README.md#streamerinfo
```

src/awkward/_connect/cling.py
```diff
@@ -183,10 +186,13 @@ class RecordView {
...
```
jpivarski, line 192:
```suggestion
    // ROOT streamer customization is done by giving specific instructions
    // in the comments written after the declaration of data members: the values
    // of `ptrs_` and `lookup_` will be ignored (//!).
    // https://github.com/root-project/root/blob/master/io/doc/TFile/README.md#streamerinfo
```

PR #1809, opened 2022-10-18
feat: Awkward Arrays in Numba's CUDA target

### LINKED ISSUE #1658: What is the state of Awkward Arrays in Numba's CUDA target?
[Way back in 2020](https://numba.discourse.group/t/making-awkward-arrays-work-in-the-cuda-target/63/3), @gmarkall said that it should be possible to add custom types to Numbafied functions in the CUDA context, and then [at an Nvidia meeting in 2022](https://www.google.com/url?q=https://reg.rainfocus.com/flow/nvidia/gtcspring2022/aplive/page/ap/session/16339878397050012ADx&sa=D&source=calendar&usd=2&usg=AOvVaw2Ug6x1pTlyvhWuiozDnaNJ), demonstrated that it is possible.

Does this work for us out of the box? What would it take to get it to work? It is definitely a goal to have Awkward Arrays work in Numba's CUDA target (and a little embarrassing that it's been on the back burner this long).

ianna:
Fixes https://github.com/scikit-hep/awkward/issues/1658


gmarkall:
Maybe relevant to this work, the high-level extension API now works for the CUDA target - this might make it quicker / easier to implement the extensions needed by Awkward. [This notebook](https://github.com/gmarkall/extending-numba-cuda/blob/main/Extending%20Numba's%20CUDA%20target%20with%20the%20High-level%20Extension%20API.ipynb) has some examples using the high-level extension API with the CUDA target.

ianna:
> Maybe relevant to this work, the high-level extension API now works for the CUDA target - this might make it quicker / easier to implement the extensions needed by Awkward. [This notebook](https://github.com/gmarkall/extending-numba-cuda/blob/main/Extending%20Numba's%20CUDA%20target%20with%20the%20High-level%20Extension%20API.ipynb) has some examples using the high-level extension API with the CUDA target.

@gmarkall - Thanks, yes, I think, I need to register `ak2.ArrayView` (and its types) extensions:

```python
self = <numba.cuda.dispatcher._Kernel object at 0x7f7c8f923a30>
ty = ak2.ArrayView(ak2.NumpyArrayType(array(int64, 1d, C), none, {}), None, ()), val = <Array [0, 1, 2, 3] type='4 * int64'>
stream = 0, retr = [], kernelargs = []

    def _prepare_args(self, ty, val, stream, retr, kernelargs):
        """
        Convert arguments to ctypes and append to kernelargs
        """
    
        # map the arguments using any extension you've registered
        for extension in reversed(self.extensions):
            ty, val = extension.prepare_args(
                ty,
                val,
                stream=stream,
                retr=retr)

<snip>

        else:
>           raise NotImplementedError(ty, val)
E           NotImplementedError: (ak2.ArrayView(ak2.NumpyArrayType(array(int64, 1d, C), none, {}), None, ()), <Array [0, 1, 2, 3] type='4 * int64'>)

kernelargs = []
retr       = []
self       = <numba.cuda.dispatcher._Kernel object at 0x7f7c8f923a30>
stream     = 0
ty         = ak2.ArrayView(ak2.NumpyArrayType(array(int64, 1d, C), none, {}), None, ())
val        = <Array [0, 1, 2, 3] type='4 * int64'>

../../../.local/lib/python3.9/site-packages/numba/cuda/dispatcher.py:417: NotImplementedError
```

ianna:
We actually register `ArrayView` and `RecordView`:
```python
@numba.extending.typeof_impl.register(ArrayView)
def typeof_ArrayView(obj, c):
    return ArrayViewType(obj.type, obj.behavior, obj.fields)
```

gmarkall:
This also needs an extension for handling them as arguments to kernels - unfortunately the interface here is a bit clunky and lacking in documentation. One example of where I implemented one is in the cuRAND example:

https://github.com/numba/numba-examples/blob/c423f5419a459f5ab8874fda6d39bb5ea05d04b2/examples/cuda/curand/numba_curand.py#L104-L118

which then has to be specified as an extension to the `@cuda.jit` decorator, like:

https://github.com/numba/numba-examples/blob/c423f5419a459f5ab8874fda6d39bb5ea05d04b2/examples/cuda/curand/example.py#L45

So a similar extension would need to be implemented for your `ArrayView`.

It's a bit inconvenient at present and I think this highlights that we really need a way to register these extensions globally rather than forcing the user to specify the extensions - in the meantime for your prototyping, can you test with the `extensions=` kwarg?

ianna:
> This also needs an extension for handling them as arguments to kernels - unfortunately the interface here is a bit clunky and lacking in documentation. One example of where I implemented one is in the cuRAND example:
> 
> https://github.com/numba/numba-examples/blob/c423f5419a459f5ab8874fda6d39bb5ea05d04b2/examples/cuda/curand/numba_curand.py#L104-L118
> 
> which then has to be specified as an extension to the `@cuda.jit` decorator, like:
> 
> https://github.com/numba/numba-examples/blob/c423f5419a459f5ab8874fda6d39bb5ea05d04b2/examples/cuda/curand/example.py#L45
> 
> So a similar extension would need to be implemented for your `ArrayView`.
> 
> It's a bit inconvenient at present and I think this highlights that we really need a way to register these extensions globally rather than forcing the user to specify the extensions - in the meantime for your prototyping, can you test with the `extensions=` kwarg?

@gmarkall - The type that is passed to `_prepare_args` is an `ak.Array`. It seems that it's boxed as an `ArrayView` at some boundary after the `extensions=` are handled. Somewhere between the `kernel.launch` and the `driver.launch_kernel`? Is this type mismatch the reason for a `CudaAPIError`? Thanks.
```python
self = <numba.cuda.cudadrv.driver.Driver object at 0x7f1f8fbb75e0>, fname = 'cuLaunchKernel', retcode = 1

    def _check_ctypes_error(self, fname, retcode):
        if retcode != enums.CUDA_SUCCESS:
            errname = ERROR_MAP.get(retcode, "UNKNOWN_CUDA_ERROR")
            msg = "Call to %s results in %s" % (fname, errname)
            _logger.error(msg)
            if retcode == enums.CUDA_ERROR_NOT_INITIALIZED:
                self._detect_fork()
>           raise CudaAPIError(retcode, msg)
E           numba.cuda.cudadrv.driver.CudaAPIError: [1] Call to cuLaunchKernel results in CUDA_ERROR_INVALID_VALUE

```

gmarkall:
**Note:** I've tried to write a quick summary to cover everything, as there's quite a few aspects to making this work - if any of the explanations are unclear / too short / confusing, please let me know and I'll elaborate / clarify.

The boxing / unboxing doesn't occur on the CUDA target - it's required that the `prepare_args` function does the equivalent work of boxing / unboxing the data. For the `ArrayView`, I see from the data model that it has five members:

```python
        members = [
            ("pos", numba.intp),
            ("start", numba.intp),
            ("stop", numba.intp),
            ("arrayptrs", numba.types.CPointer(numba.intp)),
            ("pylookup", numba.types.pyobject),
        ]
```

this corresponds with the compiled kernel expecting five parameters (for one `ArrayView`), which we can see by setting `NUMBA_DUMP_ASSEMBLY=1` when running `test_array_1d` (some inline comments added to the dump by me):

```asm
; Kernel declaration
.visible .entry _ZN6cudapy22test_1809_2dcuda_2djit7swallowB2v1B106cw51cXTLSUwHBinCqbbgUAAGBlq82ILSCEQYkgSQBFCjFSaBZJtttTo4sahbKRjoKKiDvAVKN0AuNEDVYWGDJATUC_2bQZ12oCAA_3d_3dEN2ak14ArrayView_28ak88NumpyArrayType_28array_28int64_2c_201d_2c_20C_29_2c_20_7b_7d_29_2c_20None_2c_20_28_29_29E(
        ; pos
	.param .u64 _ZN6cudapy22test_1809_2dcuda_2djit7swallowB2v1B106cw51cXTLSUwHBinCqbbgUAAGBlq82ILSCEQYkgSQBFCjFSaBZJtttTo4sahbKRjoKKiDvAVKN0AuNEDVYWGDJATUC_2bQZ12oCAA_3d_3dEN2ak14ArrayView_28ak88NumpyArrayType_28array_28int64_2c_201d_2c_20C_29_2c_20_7b_7d_29_2c_20None_2c_20_28_29_29E_param_0,
        ; start
	.param .u64 _ZN6cudapy22test_1809_2dcuda_2djit7swallowB2v1B106cw51cXTLSUwHBinCqbbgUAAGBlq82ILSCEQYkgSQBFCjFSaBZJtttTo4sahbKRjoKKiDvAVKN0AuNEDVYWGDJATUC_2bQZ12oCAA_3d_3dEN2ak14ArrayView_28ak88NumpyArrayType_28array_28int64_2c_201d_2c_20C_29_2c_20_7b_7d_29_2c_20None_2c_20_28_29_29E_param_1,
        ; stop
	.param .u64 _ZN6cudapy22test_1809_2dcuda_2djit7swallowB2v1B106cw51cXTLSUwHBinCqbbgUAAGBlq82ILSCEQYkgSQBFCjFSaBZJtttTo4sahbKRjoKKiDvAVKN0AuNEDVYWGDJATUC_2bQZ12oCAA_3d_3dEN2ak14ArrayView_28ak88NumpyArrayType_28array_28int64_2c_201d_2c_20C_29_2c_20_7b_7d_29_2c_20None_2c_20_28_29_29E_param_2,
        ; arrayptrs
	.param .u64 _ZN6cudapy22test_1809_2dcuda_2djit7swallowB2v1B106cw51cXTLSUwHBinCqbbgUAAGBlq82ILSCEQYkgSQBFCjFSaBZJtttTo4sahbKRjoKKiDvAVKN0AuNEDVYWGDJATUC_2bQZ12oCAA_3d_3dEN2ak14ArrayView_28ak88NumpyArrayType_28array_28int64_2c_201d_2c_20C_29_2c_20_7b_7d_29_2c_20None_2c_20_28_29_29E_param_3,
        ; pylookup
	.param .u64 _ZN6cudapy22test_1809_2dcuda_2djit7swallowB2v1B106cw51cXTLSUwHBinCqbbgUAAGBlq82ILSCEQYkgSQBFCjFSaBZJtttTo4sahbKRjoKKiDvAVKN0AuNEDVYWGDJATUC_2bQZ12oCAA_3d_3dEN2ak14ArrayView_28ak88NumpyArrayType_28array_28int64_2c_201d_2c_20C_29_2c_20_7b_7d_29_2c_20None_2c_20_28_29_29E_param_4
)
{
       ; Empty kernel body, just returns
	ret;
}
```

So, your `prepare_args` implementation will need to take the `ak.Array` `val`, and prepare those five arguments from it. For Numba to correctly pass those five arguments you will need to return a tuple type of five types, and a tuple of the five values. These will then be handled by the `BaseTuple` case later on in Numba's `_prepare_args` to add the five individual arguments to the call:

https://github.com/numba/numba/blob/fb952f8f5959ce31583da612a91b204bca58da81/numba/cuda/dispatcher.py#L445-L448

Another consideration is whether you need to ensure for the user that host data is automatically copied to the device before launch, and retrieved after the kernel - Numba does this for NumPy already - if the array passed in is not in the device, it creates a device array copy to use for the duration of the kernel then copies it back over the original NumPy array. You can see the implementation of this also in Numba's `_prepare_args`:

- [`wrap_arg`](https://github.com/numba/numba/blob/main/numba/cuda/dispatcher.py#L381) copies to the device if necessary,
- At [the end of `launch()`](https://github.com/numba/numba/blob/main/numba/cuda/dispatcher.py#L363-L365), any retrieval functions are run.

The retrieval functions are a function that performs the copy back - an example of this is [in `args.py`](https://github.com/numba/numba/blob/main/numba/cuda/args.py#L51).

The retrieval functions should be added to the `retr` list that is passed into the `prepare_args` function. At the moment your `prepare_args` function is declared as

```python
def prepare_args(self, ty, val, **kwargs):
```

but if you change it to

```python
def prepare_args(self, ty, val, stream, retr):
```

then you should be able to append to `retr` in there.

As a sketch of what the extension might need to look like:

```python
class ArrayViewArgHandler:
    def prepare_args(self, ty, val, stream, retr):
        if isinstance(val, ak.Array):
            # Use uint64 for start, stop, pos, the array pointers value and the pylookup value
            tys = types.UniTuple(types.uint64, 5)
  
            # ... Copy data to device if necessary...
            dev_array = # ???

            start = # Get start from val
            stop = # Get stop from val
            pos = # Get pos from val
            arrayptrs = # Get address of arrayptrs for val
            pylookup = 0 # Presumably pylookup is not used from CUDA

            # Retrieve data back to host
            def retrieve():
                # Copy dev_array back to host somehow

            # Append retrieve function if necessary
            retr.append(retrieve)

            return tys, (start, stop, pos, arrayptrs, pylookup)

        elif isinstance(val, ak._connect.numba.arrayview.ArrayView):
            # ArrayView handling... TBC later?
        else:
            return ty, val
```

ianna:
@gmarkall - I can lower an `ak.Array` as an `ArrayView` for Numba in `prepare_args`, but we need to make sure that the lifetime of the `ArrayView` is guaranteed beyond the kernel execution. The `ArrayView` is transient and defines the lifetime of `start`, `stop`, etc. How can I achieve that? I also want a user know very little about supplying an extension to a function and also be in control of the data loading to the device.

gmarkall:
> I can lower an `ak.Array` as an `ArrayView` for Numba in `prepare_args`, but we need to make sure that the lifetime of the `ArrayView` is guaranteed beyond the kernel execution. The `ArrayView` is transient and defines the lifetime of `start`, `stop`, etc. How can I achieve that?

One possible solution is to hold a reference to it in a retrieval function. If you define a retrieval function in your `prepare_args` function that refers to the transient `ArrayView`, its code object will hold a reference to the `ArrayView`. If you add it to the `retr` list, it will live beyond the kernel execution because it will only be cleared up after Numba has exited its kernel-launching logic.

Note that kernel launches are asynchronous - does the `ArrayView` need to live until the kernel has finished executing on the device, or only until you've finished preparing all the arguments to launch the kernel?

> I also want a user know very little about supplying an extension to a function

In the long term I need to add a way to register an extension globally, so that it is implicitly present for all `@cuda.jit` functions.

In the short term, you could define your own jit decorator (`@ak.cudajit`?) that wraps Numba's `@cuda.jit` decorator, adding the extension to it, and ask users to use `@ak.cudajit` instead of `@cuda.jit`... Would this be a workable solution for the time being?

> and also be in control of the data loading to the device.

When users are using Awkward with CUDA in a non-Numba context (i.e. using those functions / operations that are already built into Awkward), what do they do? Is there already a way to control data movement to / from the device?

ianna:
> > I can lower an `ak.Array` as an `ArrayView` for Numba in `prepare_args`, but we need to make sure that the lifetime of the `ArrayView` is guaranteed beyond the kernel execution. The `ArrayView` is transient and defines the lifetime of `start`, `stop`, etc. How can I achieve that?
> 
> One possible solution is to hold a reference to it in a retrieval function. If you define a retrieval function in your `prepare_args` function that refers to the transient `ArrayView`, its code object will hold a reference to the `ArrayView`. If you add it to the `retr` list, it will live beyond the kernel execution because it will only be cleared up after Numba has exited its kernel-launching logic.
> 
> Note that kernel launches are asynchronous - does the `ArrayView` need to live until the kernel has finished executing on the device, or only until you've finished preparing all the arguments to launch the kernel?

yes, it does need to live until the kernel has finished executing on the device

> 
> > I also want a user know very little about supplying an extension to a function
> 
> In the long term I need to add a way to register an extension globally, so that it is implicitly present for all `@cuda.jit` functions.
> 
> In the short term, you could define your own jit decorator (`@ak.cudajit`?) that wraps Numba's `@cuda.jit` decorator, adding the extension to it, and ask users to use `@ak.cudajit` instead of `@cuda.jit`... Would this be a workable solution for the time being?
> 

I think, we can base our development on yours :-) 
I've added an [issue](https://github.com/numba/numba/issues/8609) to keep track of it here.

> > and also be in control of the data loading to the device.
> 
> When users are using Awkward with CUDA in a non-Numba context (i.e. using those functions / operations that are already built into Awkward), what do they do? Is there already a way to control data movement to / from the device?

yes

gmarkall:
> yes, it does need to live until the kernel has finished executing on the device

OK - to ensure correctness, we may need to enqueue the cleanup function as a host callback on the stream that the kernel is executing in, so that the cleanup runs only once device execution is complete. The [`add_callback()`](https://numba.readthedocs.io/en/latest/cuda-reference/host.html#numba.cuda.cudadrv.driver.Stream.add_callback) function can be used for this, but we need to be careful to ensure this is done in a safe way, because we can't call CUDA functions from within callbacks, and the callback may execute on a different thread. It may be necessary to maintain a list of "active" `ArrayView`s, and have the callback function remove the relevant `ArrayView` from the list, with a periodic cleanup of old `ArrayView`s. This would be similar to how Numba's deferred deallocation scheme works - this is documented in [Deallocation Behavior](https://numba.readthedocs.io/en/latest/cuda/memory.html#deallocation-behavior).

> I think, we can base our development on yours :-) I've added an [issue](https://github.com/numba/numba/issues/8609) to keep track of it here.

Many thanks for adding the issue!

ianna:
@gmarkall - somehow I do not see a kernel being executed for the `ak.Array` type. The `cuda.synchronize` is called, the `mempool` shows the data have been loaded. The back end data is a CuPy array:
`GPU arrayptr <MemoryPointer 0x7fd4cf200600 device=0 mem=<cupy.cuda.memory.PooledMemory object at 0x7fd4ec9eff30>>`
 Do I need to wrap it in a `DeviceNDArray`? Or could it be the pointers address miss-match? Thanks.

BTW, the same kernel works fine on a Numpy array.

gmarkall:
How are you observing whether a kernel is executed?

ianna:
> How are you observing whether a kernel is executed?

Sorry, it does get executed! It <strike>is something else...</strike> looks like a lowered `operator.getitem` access to the `ak.ArrayView` that operates on a `ValueStructProxy`. I think, I need to check that the output (something like `%".216" = load i64, i64* %".215"`) refers to the same memory.

> I get the result that is not zeros and I can see printout from it:
> ```
> ArrayViewArgHandler::prepare_args array([10,  1,  2,  3]) <class 'numpy.ndarray'>
> ArrayViewArgHandler::prepare_args array([0, 0, 0, 0]) <class 'numpy.ndarray'>
> ArrayViewArgHandler::prepare_args 3 <class 'int'>
> 39
> 12
> 15
> 18
> Multiply numpy array [10  1  2  3] [30  3  6  9]
> 
> ```
> while for the ak.Array:
> `Multiply awkward array [0, 1, 2, 3] [0 0 0 0]
> `

ianna:
[back on this after a break]

It looks like before the `prepare_args` is called, Numba has already compiled and lowered an already initialised `ArrayView`'s `getitem`, and the data is already loaded on the GPU. The very same `ak.ArrayView` is passed to `prepare_args` so, the necessary pointers are taken from an existing data proxy, I think, but the `getitem` still does not return anything... 

@gmarkall - Should `@numba.extending.lower_builtin` `lower_getitem_at` be specialised for GPU?

```python
@cuda.jit(extensions=[array_view_arg_handler])
def multiply(array, out, n):
    tid = cuda.grid(1)
    out[tid] = array[tid] * n
```

```python
akarray = ak.Array([0, 1, 2, 3], backend="cuda")
arr = np.zeros(4, dtype=int)
d_arr =  cuda.device_array_like(arr)
multiply[1, 4](akarray, d_arr, 3)
cuda.synchronize()
result_array = d_arr.copy_to_host()
```

ianna:
@gmarkall - the following `printf` works from a `@numba.njit` function:

```python
@numba.extending.lower_builtin(len, ArrayViewType)
def lower_len(context, builder, sig, args):
    proxyin = context.make_helper(builder, sig.args[0], args[0])
    cgutils.printf(builder, "stop %d, start %d", proxyin.stop, proxyin.start)
    return builder.sub(proxyin.stop, proxyin.start)
```

The same function fails with a `numba.cuda.cudadrv.driver.LinkerError` in a `@cuda.jit`-ed function:
```
self = <numba.cuda.cudadrv.driver.CtypesLinker object at 0x7f047c2db460>

    def complete(self):
        cubin = c_void_p(0)
        size = c_size_t(0)
    
        try:
            driver.cuLinkComplete(self.handle, byref(cubin), byref(size))
        except CudaAPIError as e:
>           raise LinkerError("%s\n%s" % (e, self.error_log))
E           numba.cuda.cudadrv.driver.LinkerError: [300] Call to cuLinkComplete results in CUDA_ERROR_INVALID_SOURCE
E           error   : Undefined reference to 'printf' in '<cudapy-ptx>'

cubin      = c_void_p(None)
self       = <numba.cuda.cudadrv.driver.CtypesLinker object at 0x7f047c2db460>
size       = c_ulong(0)

../../../.local/lib/python3.9/site-packages/numba/cuda/cudadrv/driver.py:2734: LinkerError
```

gmarkall:
`cgutils.printf` is not tested on CUDA (and probably not supported) but you can use `print_varargs` from `numba.cuda.printimpl`: https://github.com/numba/numba/blob/main/numba/cuda/printimpl.py

ianna:
@gmarkall - many thanks for your help! This test works now as expected:
```python
@cuda.jit(extensions=[array_view_arg_handler])
def multiply(array, n, out):
    tid = cuda.grid(1)
    out[tid] = array[tid] * n

def test_array_multiply():
    akarray = ak.Array([0, 1, 2, 3], backend="cuda")
    results = cuda.to_device(np.zeros(4, dtype=np.int32))
    multiply[1, 4](akarray, 3, results)
    cuda.synchronize()
    host_results = results.copy_to_host()
    assert (ak.Array(host_results).tolist() == [0,3,6,9])
```

src/awkward/_connect/numba/arrayview_cuda.py
```diff
@@ -0,0 +1,41 @@
...
+                start = val._numbaview.start
+                stop = val._numbaview.stop
+                pos = val._numbaview.pos
+                arrayptrs = val._numbaview.lookup.arrayptrs.data.ptr
+                pylookup = 0
+
+                return tys, (pos, start, stop, arrayptrs, pylookup)
+            else:
+                raise ak._errors.wrap_error(
+                    NotImplementedError(
```
ianna: type error
how to transfer ak.to_backend

src/awkward/_connect/numba/arrayview_cuda.py
```diff
@@ -0,0 +1,41 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+
+from numba import types
+
+import awkward as ak
+
+########## ArrayView Arguments Handler for CUDA JIT
+
+ak.numba.register_and_check()
```
ianna: at entry point

src/awkward/_connect/numba/arrayview_cuda.py
```diff
@@ -0,0 +1,41 @@
...
+                        f"{repr(val.layout.nplike)} is not implemented for CUDA. Please transfer the array to CUDA backend to "
+                        "continue the operation."
+                    )
+                )
+
+        else:
+            return ty, val
+
+
+array_view_arg_handler = ArrayViewArgHandler()
```
ianna: entry point?

jpivarski (approved):
Fantastic! This is something we've been looking forward to for 3 years now.

(@ianna and I discussed this PR over Zoom.)

To-do items beyond this PR:

  1. The instantiation of `ArrayViewArgHandler` should go into `ak.numba._register`. I don't see any problems with creating this handler in environments that don't have CUDA, cudatoolkit, or `numba.cuda`, but if there is, it can be avoided with a try-catch.
  2. The error message for passing non-CUDA Awkward Arrays to a `numba.cuda.jit` function should be a `TypeError` and say that [ak.to_backend](https://awkward-array.org/doc/main/reference/generated/ak.to_backend.html) should be called with `backend="cuda"` to put the array on the GPU before using it. The error message should include a simple example of how to do that.
  3. It should be tested with a full suite of Awkward types, such as this: https://github.com/scikit-hep/awkward/blob/main/tests/test_1134_from_buffers_to_buffers.py

src/awkward/_connect/numba/arrayview_cuda.py
ianna: done

src/awkward/_connect/numba/arrayview_cuda.py
ianna: done

src/awkward/_connect/numba/arrayview_cuda.py
ianna: done

src/awkward/_connect/numba/arrayview_cuda.py
```diff
@@ -0,0 +1,40 @@
...
+                        "the array on the GPU before using it: "
+                        'ak.to_backend(array, backend="cuda")'
+                    )
+                )
+
+        else:
+            return ty, val
+
+
+# array_view_arg_handler = ArrayViewArgHandler()
```
ianna: @agoose77 - I'd like your opinion on this

tests-cuda/test_1809_array_cuda_jit.py
```diff
@@ -0,0 +1,98 @@
...
+config.CUDA_WARN_ON_IMPLICIT_COPY = False
+
+
+@nb_cuda.jit(extensions=[ak.numba.ArrayViewArgHandler()])
+def multiply(array, n, out):
+    tid = nb_cuda.grid(1)
+    out[tid] = array[tid] * n
+
+
+@nb_cuda.jit(extensions=[ak.numba.ArrayViewArgHandler()])
```
ianna: @agoose77 - should I just define this as `array_view_arg_handler = ArrayViewArgHandler()` in `numba.py` and use it here as `ak,numba.array_view_arg_handler`?

ianna (commented):
@agoose77 - please, have a look and let me know if you have a better suggestion? Thanks!

tests-cuda/test_1809_array_cuda_jit.py
agoose77: My attitude towards the `ak._connect.numba` vs `ak.numba` modules is that anything in `ak._connect.numba` can assume that Numba is available and compatible with Awkward, whilst `ak.numba` needs to check these things. As such, `ak.numba` can only import `ak._connect.numba` via a function call somewhere (not at module execution time).

Although we should expect Awkward to be registered by Numba in most cases, I think we still need to enforce that. Therefore, I think we want to be able to detect whether Numba has been registered in the process of obtaining an instance of `ArrayViewArgHandler`.

There are two main approaches that stand out to me:

1. A non-callable `cuda_extension` module variable 
	```python
	nb_cuda.jit(extensions=[ak.numba.cuda_extension])
	```
	If we don't have a callable `ArrayViewArgHandler`, then we would need to check that Awkward's Numba machinery is registered in `ArrayViewArgHandler.prepare_args`. 
2. A factory function in `ak.numba`:
	```python
	nb_cuda.jit(extensions=[ak.numba.create_cuda_extension()])
	```
    This function checks that Numba is registered, before returning the instance `ak._connect.numba.arrayview_cuda.ArrayViewArgHandler()`. I suppose this could be a singleton, I don't think it matters?


Although I assume that `ArrayViewArgHandler.prepare_args` is not called that often (i.e. it's only called for Python invocations), it would be better to check for registration as infrequently as possible. I am leaning towards option (2)!

agoose77 (approved):

ianna (commented):
@jpivarski - I've added more tests as you suggested. Please, feel free to merge the PR. I have some questions on better handling the Record fields and masked arrays - that we can discuss on Zoom. Thanks!

src/awkward/_connect/numba/arrayview_cuda.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+from numba import types
+
+import awkward as ak
+
+########## ArrayView Arguments Handler for CUDA JIT
+
+
+class ArrayViewArgHandler:
```
jpivarski, line 11: It's fine to import this in `ak.numba._register` because it doesn't load anything CUDA-specific (won't fail on computers with GPUs or environments without `cudatoolkit`).

I don't understand, however, how `numba.cuda.jit` knows that this `ArrayViewArgHandler` belongs to the `ArrayView` class—I don't see any registration making it so. Is it by name?

No complaint; I'm just saying that I don't know how `numba.cuda.jit` _finds_ this class, though it evidently does because the tests would fail if it didn't.

jpivarski (approved):
Very nice—this is the full suite of tests I was asking about before.

Thanks!

tests-cuda/test_1809_array_cuda_jit.py
```diff
@@ -0,0 +1,423 @@
...
+config.CUDA_WARN_ON_IMPLICIT_COPY = False
+
+
+@nb_cuda.jit(extensions=[ak.numba.array_view_arg_handler])
+def multiply(array, n, out):
+    tid = nb_cuda.grid(1)
+    out[tid] = array[tid] * n
+
+
+@nb_cuda.jit(extensions=[ak.numba.array_view_arg_handler])
```
agoose77, line 33: @jpivarski this explicit reference to `array_view_arg_handler` is how Numba executes our code (your earlier question)

jpivarski: This is something _users_ have to invoke, right? When they write a

```python
@nb.cuda.jit
def function(...)
```

they'll have to explicitly add `extensions=[ak.numba.array_view_arg_handler]`, if I understand this correctly. In that case, we should name it something more memorable (if it can't be completely automatic, as it is in the case of CPU extensions).

What about

```python
@nb.cuda.jit(extensions=[ak.numba.cuda])
def function(...)
```

?

agoose77: > This is something users have to invoke, right? 

Yes. @ianna is tracking this in https://github.com/numba/numba/issues/8609, but right now it's explicit by the user.

```python
@nb.cuda.jit(extensions=[ak.numba.cuda])
def function(...)
```

I'd perhaps go even further
```python
@nb.cuda.jit(extensions=[ak.numba.cuda_ext])
def function(...)
```
Or `.cuda_extension`

jpivarski: I'm watching that feature request, too, now.

As long as this explicit extension is needed, we want the name to be memorable enough that users don't need to look it up. Creating a JIT-compiled function is something one does at the start of a project or in a new Python prompt on the fly—when having to look up a name would cause real friction.

PR #1810, opened 2022-10-18
docs: fix redirects

### LINKED ISSUE #1808: URL mapping in new docs
### Which documentation?

Tutorials site

https://github.com/scikit-hep/uproot5/issues/760 points out a dead link in the Uproot docs:

https://awkward-array.readthedocs.io/en/latest/ak.types.Type.html

That's an indication that this link used to work and it has probably been shared beyond just the Uproot docs. It doesn't have the `_auto` in the URL because it was one of the hand-written reST files, not one made by `prepare_docstrings.py`.

Looking in the `docs` branch by swapping `latest` with `docs`, it still doesn't exist, but

https://awkward-array.readthedocs.io/en/docs/reference/generated/ak.types.Type.html

does exist, and it's where this URL is supposed to go. So we'll need a redirect.

Also, the layout URLs that are now

https://awkward-array.readthedocs.io/en/docs/reference/generated/ak.contents.Content.html

(unfortunately!) need a redirect from "`ak.layout.`" to "`ak.contents.`" (all the Content classes). Similarly for "`ak.layout.Record`" (only one Record) → "`ak.record.Record`" and "`ak.layout.Index`" → "`ak.index.Index`" (all the Index classes). Unless we want those links to fail as an indicator to users that the name no longer exists? Maybe better, a page briefly describing the name change (rather than an automated redirect)?

agoose77:
Fixes #1808


jpivarski:
Is there a way to test this in ReadTheDocs? Maybe by keeping the `docs` branch up-to-date with `main` and pulling your own requests into `docs` (or committing directly into `docs`), and then make a PR from `docs` to `main`? Since `docs` is still set up in ReadTheDocs to serve changes, we could do all of the human-eyeball tests there.

agoose77:
There's something odd going on with this CI...

PR #1812, opened 2022-10-19
ci: enable versions for built docs

agoose77:
Now we explicitly enable a version on RTD, and then build it. A separate workflow edits the PR description to show a link to the rendered docs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -117,7 +117,9 @@ jobs:
         name: jupyter-cache
         path: docs-sphinx/_build/.jupyter_cache
     - name: Trigger RTD Build
-      # NB: head_ref is only valid for PR triggers
+      env:
+          RTD_TOKEN: "${{ secrets.RTD_TOKEN }}"
       run: |
+        # NB: head_ref is only valid for PR triggers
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
This seems like the right thing, but maybe it needs to be merged before it takes effect? (No, it should be triggered by docs.yml...)

The branch name translation turns `/` into `-`, so

* [https://awkward-array.readthedocs.io/en/agoose77-docs-build-prs/index.html](https://awkward-array.readthedocs.io/en/agoose77-docs-build-prs/index.html)

ought to point to _this_ PR's documentation. I don't see it, so one of my assumptions is wrong.

If this is something that will only take effect after being merged, that's fine. I'll merge it and then see what it does when merged into the open docs PRs.

agoose77:
I think this failed due to a network issue. However, the scripts didn't assert that the HTTP response code indicated success, so I've ensured that they will now raise an exception in future. 

The feature that edits PR descriptions to include a docs link will not run until merged, as it uses the `pull_request_target` feature that allows PRs to access secrets. For security reasons, GH only runs the destination branch's workflows with `pull_request_target`

PR #1815, opened 2022-10-20
docs: fix redirects

### LINKED ISSUE #1808: URL mapping in new docs

agoose77:
This PR fixes the redirects that map old v1 URLs onto v2.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-fix-url-redirects/> once Read the Docs has finished building :hammer:


jpivarski:
Fixes scikit-hep/uproot5#760

jpivarski (approved):
The great thing about this is that I could directly test it. I tried all of the old links, substituting `latest` → `agoose77-docs-fix-url-redirects`, and they all work now.

I'll enable auto-merge.

agoose77:
I think that might already have been fixed by making `latest` point to `main-v1` on RTD? The `latest` branch is still v1.

jpivarski:
Or not. This doesn't trigger the build tests. But the docs tests have all completed successfully, and my test of the links does what an integration test would do.

agoose77:
Don't merge just yet - noticed a link misbehaving

jpivarski:
Okay, I won't.

agoose77:
OK, seems fine!

jpivarski:
For instance, this was the originally reported dead link:

* https://awkward-array.readthedocs.io/en/latest/ak.types.Type.html

(which is live now, but v1, as it should be).

What I tested was:

* https://awkward-array.readthedocs.io/en/agoose77-docs-fix-url-redirects/ak.types.Type.html

and similar for all `layout` → `index`, `record`, `contents`; even `ak.layout.ArrayBuilder` went to the right place.

I've approved this; when you're done with the change you want to make, you can merge it.

PR #1816, opened 2022-10-20
ci: use custom PR message action

agoose77:
I tried to use the RTD action in #1812, but it transpires that this leverages the RTD GitHub integration that we aren't using. We couldn't see this until it made it into a base branch, so I only discovered it after the merge.

So, this PR 
- Roll our own PR message editing machinery.
- Replace Python with bash (avoid Python step in job).
- Bundle new JS to edit commit message.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-disable-pr-link/> once Read the Docs has finished building :hammer:


agoose77:
I cancelled the CI because this doesn't touch any compiled code.

jpivarski (approved):
I would have asked if we should switch to the GitHub ReadTheDocs Integration, but you've already adapted to the interface we have.

I'll merge this and we'll see if it works. I got your message that you intentionally cancelled the build because it isn't testing anything; I'll use the red button to merge without successful tests.

(Fingers in ears...)

PR #1826, opened 2022-10-24
feat: preserve `None` in `np.ravel()`

### LINKED ISSUE #1460: `ak.ravel` doesn't mirror `np.ravel` for `None` leaf values
### Version of Awkward Array

1.8.0

## The bug :bug:
In #1459 I noticed that `ak.ravel(masked_array)` does not match `np.ravel(masked_array)` for a NumPy `masked_array`. 

An example:
```python
import awkward as ak
import numpy as np


array = np.ma.array(
    np.random.random(size=(10, 3)),
   mask=np.random.randint(0, 2, size=(10, 3)).astype(np.bool_)
)

assert np.ravel(array).tolist() == ak.ravel(array).tolist()
```

## Why is this a bug? :question: 
I think this is a problem, because we try to make NumPy functions on `NumpyArray`s behave as they do in regular NumPy. Given that we explicitly handle the `masked_array` type in our codebase (converting it to a `ByteMaskedArray`), I assert that we should also expect the same to happen for masked arrays.

## What should we do here? :mag_right: 
Whilst I think we do want to drop `None` values that are non-leaf (this level of structure is erased anyway, so keeping `None` values seems pointless), I don't think we necessarily want to lose those at the leaves unless we want to break NumPy compatibility. 

The existing behaviour in `ak.ravel` _is_ useful, though, particularly when using histograms. So, I think we might want this to be an option to `ak.ravel`, e.g. `ak.ravel(drop_nones=True)`. 

I'd be interested in hearing other people's thoughts here.

agoose77:
This PR:
- Distinguishes `ak.ravel(array)` from `ak.flatten(array, axis=None)`
   - The latter drops `None` values (and the option-type)
- Removes `x in Array` support. 
   - This was discussed offline with @jpivarski - our implementation doesn't support anything more complex than scalar values, which is probably more harmful than helpful.]
- Adds `completely_flatten` support to `ak.record.Record`
- Adds preliminary docs to explain `ak.ravel` vs `ak.flatten`.

Closes #1460

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-ravel-preserve-none/> once Read the Docs has finished building :hammer:


jpivarski (changes requested):
We talked about this on Slack and I approve. I also gave the changes here a close read-through. Although this PR touches many files, it does very similar things to each. The removal of `ak.highlevel.*.__contains__` is tangential, but fine.

This PR changes the behavior of `ak.ravel` (something that I'd call a "feat", not a "fix"). Do you have any tests that demonstrate the new behavior? That's the only thing I'd request before the PR is done.

agoose77:
I'm glad you caught that - yes, I do have tests, but they've not been pushed. Fixing...

jpivarski (approved):
That does it: the test of `ak.ravel` has the numeric `None` stay but the list `None` goes away, and the test of `ak.flatten` eliminates `None` at all levels (because `axis=None`).

I'll set it to auto-merge.

PR #1827, opened 2022-10-24
fix: `ak.prod` for booleans had been incorrectly casting output

jpivarski:
Discussed on Slack with @agoose77; we came to the same conclusion at the same time.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-fix-boolean-prod-reducer-bug/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_reducers.py
+++ b/src/awkward/_reducers.py
@@ -229,7 +229,6 @@ class Sum(Reducer):
         if array.dtype == np.bool_:
             if result.dtype in (np.int64, np.uint64):
                 assert parents.nplike is array.nplike
-                assert parents.nplike is array.nplike
                 array._handle_error(
                     array.nplike[
                         "awkward_reduce_sum_int64_bool_64",
@@ -314,11 +313,11 @@ class Prod(Reducer):
             raise ak._errors.wrap_error(
                 ValueError(f"cannot compute the product (ak.prod) of {array.dtype!r}")
             )
-        result = array.nplike.empty(
-            cls.maybe_double_length(array.dtype.type, outlength),
-            dtype=cls.return_dtype(array.dtype),
-        )
         if array.dtype == np.bool_:
+            result = array.nplike.empty(
+                cls.maybe_double_length(array.dtype.type, outlength),
+                dtype=np.dtype(np.bool_),
+            )
             assert parents.nplike is array.nplike
             array._handle_error(
                 array.nplike[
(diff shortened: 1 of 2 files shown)


src/awkward/_reducers.py
```diff
@@ -314,16 +313,16 @@ def apply(cls, array, parents, outlength):
...
+            result = array.nplike.empty(
+                cls.maybe_double_length(array.dtype.type, outlength),
+                dtype=np.dtype(np.bool_),
+            )
             assert parents.nplike is array.nplike
             array._handle_error(
                 array.nplike[
                     "awkward_reduce_prod_bool",
-                    array.dtype.type,
+                    result.dtype.type,
```
agoose77, line 325: My only suggestion, feel free to revert.

agoose77 (approved):

src/awkward/_reducers.py
jpivarski: Yes, absolutely. I don't see how it worked without this.

jpivarski: With anything but `np.bool_` as the first argument, `nplike.__getitem__` should have been unable to find the ctypes function.

```python
>>> import awkward as ak, numpy as np
>>> numpy = ak.nplikes.nplike_of(None)
>>> numpy["awkward_reduce_prod_bool", np.bool_, np.int64, np.int64]
<NumpyKernel awkward_reduce_prod_bool, bool, int64, int64>
>>> numpy["awkward_reduce_prod_bool", np.int64, np.int64, np.int64]
  File "/home/jpivarski/irishep/awkward/awkward/nplikes.py", line 495, in __getitem__
    return NumpyKernel(ak._cpu_kernels.kernel[name_and_types], name_and_types)
KeyError: ('awkward_reduce_prod_bool', <class 'numpy.int64'>, <class 'numpy.int64'>, <class 'numpy.int64'>)
```

Oh well! It's correct now and works now.

I'll merge it.

agoose77: This is actually what I was trying to refer to yesterday. We accidentally always selected a `bool` output function even when the result was actually allocated as an integer array. This led to the weird results that you observed in the original issue. Now that you've clarified all the logic, it's a lot easier to see that this was wrong, and that this PR fixes the general problem.

jpivarski: Ohhhhh. Okay! Sorry for not getting it yesterday.

I went forward with this PR because I wanted to keep the logic simple with a small amount of duplication (declaring `result` in all three branches of `if`-`elif`-`else`, the same way in only two of those). Declaring the output as `bool` and then `astype` to the actual output type is a little wasteful, but in a corner case of a corner case. But it is important to pick the correct kernel for the array types you have!

PR #1829, opened 2022-10-25
feat: to/from RDataFrame support for bool type arrays

ianna:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-rdataframe-bool/> once Read the Docs has finished building :hammer:


jpivarski (approved):
Nice!

I noticed that the test only does a round-trip, and I wasn't sure whether RDataFrame was seeing these values as being boolean, too. (Other than the `GetColumnType`, that is: I wanted to know if we could do computations on them as booleans.)

I'll merge it.

PR #1831, opened 2022-10-25
docs: add notes on "declaring `ArrayBuilder` type"

agoose77:
In my view, `ak.ArrayBuilder` serves many purposes:
- Construction of Awkward Arrays whose sizes are unknown (dynamic lengths)
- Construction of Awkward Arrays whose types are unknown (dynamic types)

`ak.LayoutBuilder()` also serves the former point (dynamic lengths), but we are removing the Python API in v2. Therefore, I think it worth improving the docs for `ak.ArrayBuilder()` (the next best thing) to handle the cases that `ak.LayoutBuilder()` previously handled.This small PR adds notes to the user guide on `ak.ArrayBuilder()` that explain how to stabilise the output type. 

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-arraybuilder-schema/> once Read the Docs has finished building :hammer:

### DIFF
--- a/docs-sphinx/user-guide/how-to-create-arraybuilder.md
+++ b/docs-sphinx/user-guide/how-to-create-arraybuilder.md
@@ -376,9 +376,87 @@ new_array
 new_array.layout
 ```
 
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
This is a good clarification. I might have called it "Setting the type of empty lists," since the semantics is not like an upfront declaration, and the empty containers might be nested lists within an array.

I'll leave it to you to decide whether to keep the title or change it, and you can merge when done.

PR #1832, opened 2022-10-25
docs: add vertical margin between badges

### LINKED ISSUE #1799: Requests from review of #1776 (merging docs into main).

agoose77:
Closes #1799

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-badges-margin/> once Read the Docs has finished building :hammer:

### DIFF
--- a/docs-sphinx/_static/css/awkward.css
+++ b/docs-sphinx/_static/css/awkward.css
@@ -42,3 +42,8 @@ html[data-theme="dark"] .sd-card {
   margin-left: auto;
   margin-right: auto;
 }
+
+.shield-badge {
+  margin-top: 0.15em;
+  margin-bottom: 0.15em;
+}
--- a/docs-sphinx/_static/css/try-awkward-array.css
+++ b/docs-sphinx/_static/css/try-awkward-array.css
@@ -1,0 +1,20 @@
+/*Styles for our try-awkward-array demo*/
+[role=main] {
+    flex-grow: 1;
+    /*Allow children to grow*/
+    display: flex;
+    flex-direction: column;
+}
+
+[role=main] > * {
+    /*Accomodate parent size*/
+    flex-grow: 1;
+    /*Allow children to grow*/
+    display: flex;
+    flex-direction: column;
+}
+
+iframe {
+    flex-grow: 1;
+    min-height: 300px;
+}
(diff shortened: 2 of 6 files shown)


jpivarski:
Yes! That works perfectly! (I'm very surprised that it was possible.) It looks great on my phone, though [this one](http://mobiletest.me/iphone_5_emulator/?u=https://awkward-array.readthedocs.io/en/agoose77-docs-badges-margin/getting-started/try-awkward-array.html) (and smaller) would benefit from minimizing the page title:

* "Try it in your browser" → "Try it!"

The fact that it runs in the browser, amazing as that is, doesn't matter to many users. Reducing the title to two words would make it fit on nearly all screens.

Phones use a significant amount of space on the bottom of the screen for "< Previous", "Next >", and the copyright message, but _this is okay_ because that space gets used up by the keyboard when you try to edit any of the code. It adjusts better between yes-keyboard and no-keyboard, which is important, since you lose the keyboard when you press ▸. So RetroLite not filling the bottom of the screen is not a problem, while RetroLite not filling the top of the screen is a problem that can be solved by minimizing the title.

jpivarski (approved):
I wrote a comment before the request for review came up. My suggestion about changing the title is a weak-suggest, not a must-do.

jpivarski:
Checkmark to 5c5f330224780b770c1c808ee078a88900304961!

PR #1836, opened 2022-10-26
docs: add rdataframe user guide

ianna:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-rdataframe-user-guide-v1/> once Read the Docs has finished building :hammer:


agoose77:
Hey @ianna, I hope to get the C++ dependency management implemented in #1837. Once that's in, you should be able to add ROOT as a dependency for the docs.

agoose77:
@jpivarski I've touched this PR, so it would be best if you were the reviewer.

I've switched the CI from `xeus-cling` to `root` as our C++ kernel provider (in addition to the PyROOT integration for this PR). I'm re-using the same environment spec for wheel building and consumption to mitigate the fact that we don't build a manylinux wheel.

docs-sphinx/environment.yml.cog
```diff
@@ -3,8 +3,8 @@ channels:
   - conda-forge
 dependencies:
   - pip
-  - xeus-cling=0.13.0
-  - gxx_linux-64=9.4.0
+  - cxx-compiler
+  - root
   - pip:
     # [[[cog
     # import cog
```
jpivarski, line 10: It's best to remove, rather than comment-out, unnecessary code. It's always available in history.

Also, if this doesn't need to be an environment.yml.cog anymore, but can be a static environment.yml, that would also be preferred. I was nervous about adding it in the first place because it's like replacing pyproject.toml with a setup.py: configurations are safer and more maintainable when they're declarative/static, and most of the work in the Python configuration world is trying to get more declarative/static. This is a backwards step.

(When you need it, you need it, so it's nice that such things exist, but we should view it as having a negative weight. Now that it's not needed—great! Let's go back to static files!)

jpivarski (approved):
Looks good! (I viewed the rendered docs as well.)

My only complaint is the commented-out code (below). Once that's gone, fee free to merge!

docs-sphinx/environment.yml.cog
agoose77: Cog is still being used here in order to install separate dependencies for our CI and for RTDs build system. This is fundamentally a problem of performance; it takes ~270s to install the `environment.yml` on RTD with mamba, vs ~50s for just the PyPI dependencies. We really want our docs to be as fast as possible to deploy, especially as "right now" there is no nice feedback that the docs are live. 

The code in the comments is actually executed by cog - it ignores common line prefixes that sit in front of the marker tags. This is a nice feature of cog; the `environment.yml.cog` is valid YAML, and the code that generates the result can be bundled inside the target.

Cog is used to embed the `requirements.txt` inside the `environment.yml`. The idea is to avoid having two files that need to be kept in-sync (requirements.txt and environment.yml) by having the latter be generated via a templating system. Conventionally, you can use cog in-place, i.e. it can update a cog-file, however I want to avoid checking-in a stale template result, so I make it clear that we have a template (`.yml.cog`). As such this is still mostly declarative; we use the static contents of `requirements.txt` and `environment.yml.cog`, and compose them into a single file. Conda does provide a native way to do this, but it's not working properly in Mamba, by the looks of things (it doesn't locate the requirements.txt file), hence cog.

ianna (commented):
@jpivarski - @agoose77 argument seems convincing. If it resolves the conversation, please, merge it.

docs-sphinx/environment.yml.cog
agoose77: To add - I'm not wedded to cog! I just recall it being a good solution :)

jpivarski: (I see: it's not commented _out_; commenting is just how cog works.)

Also, I see that this environment.yml (for docs) is derived from the global requirements.txt, and that's why you're using cog here.

I'm not objecting to cog as a particular technology to make that happen; but rather that we're doing it at all. Sometimes, two copies of some code/configuration that are clearly supposed to be the same thing is better than a clever system to remove one duplication. Actually, I start thinking about duplication when the number gets up to three or four.

What we really want here is some declarative syntax for "insert some other file here", like setup.cfg has, replacing the `open("source-file", "r").read()` that used to be in setup.py. Because that _is_ what we're talking about here with cog: going from a setup.cfg-like solution to a setup.py-like solution.

Aha! It looks like there is such a syntax: https://stackoverflow.com/a/59056234/1623645

agoose77: I'm agreed on this being non-ideal. Unfortunately we cannot use `-r requirements.txt` with Mamba due to https://github.com/mamba-org/mamba/issues/1726 
This makes the tradeoff:
- no templating logic, use slow Conda
- new templating dependency, use fast Mamba
- no templating logic, use fast Mamba & duplicate `requirements.txt` alongside `environment.yml`

My assessment here is that it's safer to have a template than two files that *both* implement a dependency specification; these nearly always get out of date :/

jpivarski: Okay, sure. Let's stick with cog, at least until Mamba is fixed.

agoose77: OK, I'll add a note!

PR #1841, opened 2022-10-26
fix: array.dtype.type will never be in ak.types.numpytype._dtype_to_primitive_dict

### LINKED ISSUE #1840: ak.type does not understand `numpy.<type>` style dtypes
### Version of Awkward Array

2.0.0rc1

numpy = 1.23.4

```python3
>>> import awkward as ak
>>> import numpy as np
>>> x = np.random.normal(size=100)
>>> ak.type(x)
```
results in:
```
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_type.py", line 99, in _impl
    out = ak.types.numpytype._dtype_to_primitive_dict[array.dtype.type]
KeyError: <class 'numpy.float64'>

  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_type.py", line 60, in type
    return _impl(array)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_type.py", line 101, in _impl
    raise ak._errors.wrap_error(
TypeError: while calling

    ak.type(
        array = numpy.ndarray([ 0.27824033 -1.483569   -0.61108357 ...
    )

Error details: numpy array type is unrecognized by awkward: <class 'numpy.float64'>
```

`np.float64` (or `np.<type>` in general) is a fairly common way for folks to denote typing, we should probably support it.

lgray:
Fixes #1840 

Original input to square brackets would never work?

### DIFF
--- a/src/awkward/operations/ak_type.py
+++ b/src/awkward/operations/ak_type.py
@@ -72,15 +72,7 @@ def _impl(array):
         primitive = ak.types.numpytype.dtype_to_primitive(np.dtype(array))
         return ak.types.NumpyType(primitive)
 
-    elif isinstance(array, builtins.type) and issubclass(
-        array,
-        tuple(x.type for x in ak.types.numpytype._dtype_to_primitive_dict),
-    ):
-        return ak.types.NumpyType(
-            ak.types.numpytype._dtype_to_primitive_dict[np.dtype(array)]
-        )
-
-    elif isinstance(array, (bool, np.bool_)):
+    elif isinstance(array, bool):
         return ak.types.NumpyType("bool")
 
     elif isinstance(array, numbers.Integral):
@@ -103,7 +95,7 @@ def _impl(array):
         if len(array.shape) == 0:
             return _impl(array.reshape((1,))[0])
         else:
-            primitive = ak.types.numpytype.dtype_to_primitive(np.dtype(array.dtype))
+            primitive = ak.types.numpytype.dtype_to_primitive(array.dtype)
             out = ak.types.NumpyType(primitive)
             for x in array.shape[-1:0:-1]:
                 out = ak.types.RegularType(out, x)
--- a/src/awkward/operations/ak_type.py
+++ b/src/awkward/operations/ak_type.py
@@ -76,7 +76,7 @@ def _impl(array):
         primitive = ak.types.numpytype.dtype_to_primitive(np.dtype(array))
         return ak.types.NumpyType(primitive)
 
-    elif isinstance(array, bool):
+    elif isinstance(array, (bool, np.bool_)):
         return ak.types.NumpyType("bool")
 
     elif isinstance(array, numbers.Integral):
--- a/src/awkward/operations/ak_type.py
+++ b/src/awkward/operations/ak_type.py
@@ -96,7 +96,7 @@ def _impl(array):
             return _impl(array.reshape((1,))[0])
         else:
             try:
-                out = ak.types.numpytype._dtype_to_primitive_dict[array.dtype.type]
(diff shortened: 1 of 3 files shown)


agoose77:
Would you be able to add a `test_1840-type-numpy-array.py` to `tests/` containing the following:
```python3

import numpy as np  # noqa: F401
import pytest  # noqa: F401

import awkward as ak  # noqa: F401


def test():
    array = np.random.random(size=512).astype(dtype=np.float64)
    assert ak.type(array) == ak.types.ArrayType(
        ak.types.NumpyType("float64"), 512
    )
```

lgray:
Sure! Just a bit.

agoose77 (approved):

jpivarski:
I'm working on fixes to this, but I'm in a meeting now, so that's paused.

jpivarski:
(Do not merge!)

agoose77:
Weirdly enough, I was actually looking at this code for another issue when I noticed that we have a `dtype_to_primitive` function that we can use here.

jpivarski:
I got the 10 minutes to do this, and now realize that I had misunderstood the original issue #1840. I thought it was about making `ak.type(np.float64)` work, but it had been more broken than that: `ak.type(np.array([1, 2, 3]))` didn't work.

Well, all of these cases work now. Since it's doubly approved, I'll reenable auto-merge.

agoose77:
I've held this from merging, as we probably want to use `numpytype.dtype_to_primitive` - but we might need to do more with respect to the loop that checks for membership in the type dictionary. We could use a `try/except`, but perhaps there's a more elegant way of writing this code.

src/awkward/operations/ak_type.py
```diff
@@ -64,15 +65,18 @@ def _impl(array):
...
+        and issubclass(array, np.generic)
     ):
-        return ak.types.NumpyType(
-            ak.types.numpytype._dtype_to_primitive_dict[array.dtype]
-        )
+        primitive = ak.types.numpytype.dtype_to_primitive(np.dtype(array))
+        return ak.types.NumpyType(primitive)
 
-    elif isinstance(array, (bool, np.bool_)):
+    elif isinstance(array, bool):
```
jpivarski: I have to revert this:

```suggestion
    elif isinstance(array, (bool, np.bool_)):
```

because a NumPy boolean is not a Python boolean:

```python
>>> [isinstance(x, bool) for x in np.array([False, True])]
[False, False]
>>> [isinstance(x, np.bool_) for x in np.array([False, True])]
[True, True]
```

jpivarski (approved):
The `dtype_to_primitive` function is much better. I think it includes some logic for getting the time units of datetime/timedelta types right. Even if it doesn't, it's the right function to use here.

Just one problem below. I'll self-accept my suggestion and enable auto-merge.

src/awkward/operations/ak_type.py
agoose77: The `isinstance(x, np.generic)` should pick up the instances of the primitive NumPy `type`s:
```pycon
>>> isinstance(np.array([True,False])[0], np.generic)
True
```

jpivarski: That's true. And I see that the `np.generic` check is above the boolean one. Well, checking for `(bool, np.bool_)` is common enough to be an idiom, and it's paired with checking numbers for `numbers.Integral` and `numbers.Real` (usually written that way to catch NumPy, though it would catch more than that).

jpivarski:
I think this is done now; I'm reenabling auto-merge.

PR #1845, opened 2022-10-28
chore: remove Identifier (once known as Identities) from codebase.

### LINKED ISSUE #1843: Remove Identifiers/identifier/has_identifiers from the codebase
It no longer makes sense to keep this "foot in the door" for a feature that we evidently never needed.

I'll do a sweep and remove it; it's important for this to happen before 2.0.0.

jpivarski:
Identifiers were motivated by [PartiQL](https://github.com/jpivarski/PartiQL), an experimental combinatorics language that uses set-like referential identity to track particles through a calculation. It would require our data to track a [surrogate index](https://en.wikipedia.org/wiki/Surrogate_key) through all of its operations, like a Pandas `Index`.

The `Identities` (Awkward v1) and `Identifier` (Awkward v2) were a "foot in the door" implementation to add such a feature. Originally, it was supposed to be implemented before the Awkward 1.0.0 release, but other features were more pressing. I left them in on the theory that it's easier to take them out than add them later, but it's been 3 years now and 2.0.0 is almost ready to ship—it's not going to happen. Moreover, 2.0.0 is almost ready to ship—removing them would mean changing the public API (albeit for the low-level layer, intended for downstream dependencies, rather than data analysts). So it must be done now.

Hopefully, this won't be too hard to merge with the other PRs. The most likely complication is that all Content subclasses and Form subclasses have 1 fewer argument: instead of "`identifier, parameters`", it's just "`parameters`". "`None`" is a likely value to be passed as `identifier`, and it's a likely value to be passed to `parameters`, so it shows up as "`None, parameters`" or "`None, None`".

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-remove-identifiers-aka-identities/> once Read the Docs has finished building :hammer:


jpivarski:
Remove 1345 lines that were never tested (Identifiers were almost always `None`) and the coverage goes up by... 0.1%.

> Merging [#1845](https://codecov.io/gh/scikit-hep/awkward/pull/1845?src=pr&el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=scikit-hep) ([10bede1](https://github.com/scikit-hep/awkward/commit/10bede146f25ffbf12de07659dacd04f146c0a1a)) into [main](https://codecov.io/gh/scikit-hep/awkward/commit/178b4e92a0ce7ab515d3ccd9b1929d9bda5ec183?el=desc&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=scikit-hep) ([178b4e9](https://github.com/scikit-hep/awkward/commit/178b4e92a0ce7ab515d3ccd9b1929d9bda5ec183)) will **increase** coverage by `0.10%`.
> The diff coverage is `93.75%`.

jpivarski:
The docs have (correctly) lost all references to `identifier`.

src/awkward/_connect/numba/layout.py
```diff
@@ -1089,23 +1029,15 @@ def from_form(cls, form):
...
+    def __init__(self, contenttypes, fields, parameters):
         super().__init__(
-            name="ak.RecordArrayType(({}{}), ({}), {}, {})".format(
-                ", ".join(x.name for x in contenttypes),
-                "," if len(contenttypes) == 1 else "",
-                "None" if fields is None else repr(tuple(fields)),
-                identifiertype.name,
-                json.dumps(parameters),
-            )
+            name=f"ak.RecordArrayType(({', '.join(x.name for x in contenttypes)}{',' if len(contenttypes) == 1 else ''}), {'None' if fields is None else repr(tuple(fields))}, {json.dumps(parameters)})"
```
agoose77: Could we keep this as a `str.format`? I find it a bit more readable than the `fstring` in this case.

src/awkward/_connect/numba/layout.py
```diff
@@ -1384,25 +1316,16 @@ def from_form(cls, form):
...
         super().__init__(
-            name="ak.UnionArrayType({}, {}, ({}{}), {}, {})".format(
-                tagstype.name,
-                indextype.name,
-                ", ".join(x.name for x in contenttypes),
-                "," if len(contenttypes) == 1 else "",
-                identifiertype.name,
-                json.dumps(parameters),
-            )
+            name=f"ak.UnionArrayType({tagstype.name}, {indextype.name}, ({', '.join(x.name for x in contenttypes)}{',' if len(contenttypes) == 1 else ''}), {json.dumps(parameters)})"
```
agoose77: Same here :)

agoose77 (commented):
This was a big PR! It will be nice to drop the identities mechanism given that we don't properly leverage it; less code to read and type in future!

I went line-by-line. I'll run a few greps offline to see if we've missed anything, but once that's done I'll approve!

agoose77:
Also, I took the liberty of removing the v1 content from the Doxygen index, as you already have to touch that file in this PR.

agoose77:
I checked `NumpyArray`, `IndexedOptionArray`, and `ListOffsetArray`, which I expected to be the most common layout types, and found no misuse.

Do you know what the `is_identifier.match` excerpt refers to in `_prettyprint.py` (used in a few places)?

jpivarski:
> Do you know what the `is_identifier.match` excerpt refers to in `_prettyprint.py` (used in a few places)?

That checks to see if a field name of a record fits the regex that most languages use as identifiers—in particular, the Datashape language (which inherits it from whatever parser they're using). That's `/[A-Za-z_][A-Za-z_0-9]*/`.

This PR was not a sed substitution. I checked each usage of strings matching `/identi/i` (because I wanted to be sure there wasn't anything left-over from the "Identity"/"Identities" days).

I'll also simplify the two f-string substitutions. It used to be all `format` method calls because we were originally allowing for Python 2. So, little by little, I've been converting them to f-strings, and it seemed to be a good time/place to do that if there were identifiers to remove.

agoose77:
> > Do you know what the `is_identifier.match` excerpt refers to in `_prettyprint.py` (used in a few places)?
> 
> That checks to see if a field name of a record fits the regex that most languages use as identifiers—in particular, the Datashape language (which inherits it from whatever parser they're using). That's `/[A-Za-z_][A-Za-z_0-9]*/`.

Fab, this was my read of things, but I never closely looked at identifiers before you removed them, so I wanted to clarify.
 
> This PR was not a sed substitution. I checked each usage of strings matching `/identi/i` (because I wanted to be sure there wasn't anything left-over from the "Identity"/"Identities" days).

You can tell; lots of tricky things to find beyond a guided exploration. Even just reading all the files took a while, nice effort! 

> I'll also simplify the two f-string substitutions. It used to be all `format` method calls because we were originally allowing for Python 2. So, little by little, I've been converting them to f-strings, and it seemed to be a good time/place to do that if there were identifiers to remove.

I'm 100% on-board with f-strings. In these two rare cases, I'd either assign the long statements to local variables, or just use a multi-line `string.format`, hence the suggestion.

agoose77 (approved):

jpivarski:
In this case, I'm going to leave these strings as f-strings, but predefine the variables to substitute. That way, all of the Numba type strings will be generated in the same way.

jpivarski:
Auto-merging.

Thanks!

PR #1852, opened 2022-10-31
ci: allow pushes to main to deploy

agoose77:
I noticed that #1836 didn't trigger an RTD build for pushes to `main`, so I've updated `docs.yml`

I find the expression `NOT (X AND NOT Y)` clearer than `(NOT X) OR Y` because in this case `X` and `Y` are related. We only avoid deploying the docs if the event was a PR event, and the PR does not originate from `scikit-hep/awkward`

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-ci-docs-deploy-on-main/> once Read the Docs has finished building :hammer:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -169,8 +169,8 @@ jobs:
   deploy-docs:
     runs-on: ubuntu-22.04
     needs: [build-docs]
-    # We can only deploy for PRs on host repo
-    if: github.event.pull_request.head.repo.full_name == github.repository
+    # We can only deploy for PRs on host repo, or pushes to main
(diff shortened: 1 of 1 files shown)


agoose77:
It's not easy to guarantee that this is working, but if the docs deploy for this PR then we'll know at least that it didn't break anything :)

jpivarski (approved):
I frequently get boolean logic backward, too. But the good thing is that it's easy to notice with a test, since it will either always work or always fail.

I verified that the docs for this branch got pushed, so I guess it worked!

agoose77:
For posterity - the docs were deploying on the branch. The problem was that I made an oversight and didn't think about the case where the docs workflow runs on `push` (not `pull_request`) to `main`. In that case, the `github.event.pull_request` object is `null`, so the condition always fails. So, let's now see if this merged PR deploys...

PR #1853, opened 2022-10-31
ci: don't run PR job on `main`

agoose77:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-ci-fix-deploy-main/> once Read the Docs has finished building :hammer:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -198,6 +198,7 @@ jobs:
               "https://readthedocs.org/api/v3/projects/awkward-array/versions/$VERSION_SLUG/builds/"
 
     - name: Edit PR description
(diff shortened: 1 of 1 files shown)


agoose77:
I missed something in the last PR (note-to-self, when you `assume`, you make an `ass` out of `u` and `me`). This fixes the last approved PR, so I'm happy to merge without a second pair of eyes.

jpivarski (approved):
Yes, that's fine. Some of these I review as, "I don't know how the docs/CI system works, but I can see that the output is right." I had thought that the qualifier for the last one was that the docs were generated.

agoose77:
Looks like it succeeded (at least the build triggers!)

image

PR #1856, opened 2022-11-01
fix: separate contiguity from dimension test in `NumpyArray._reduce_next`

### LINKED ISSUE #1847: Non-contiguous `NumpyArray` calls `toRegularArray()`
### Version of Awkward Array

main

```python
>>> array = ak.from_numpy(np.arange(24)[::3])
>>> ak.sum(array, axis=-1)
  File "/home/angus/.mambaforge/envs/awkward-1.0/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3398, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-26-8650b4044dc1>", line 1, in <cell line: 1>
    ak.sum(array)
  File "/home/angus/Git/awkward-1.0/awkward/operations/ak_sum.py", line 208, in sum
    return _impl(
  File "/home/angus/Git/awkward-1.0/awkward/operations/ak_sum.py", line 279, in _impl
    out = layout.reduce(
  File "/home/angus/Git/awkward-1.0/awkward/contents/content.py", line 855, in reduce
    return self._reduce_axis_none(
  File "/home/angus/Git/awkward-1.0/awkward/contents/content.py", line 830, in _reduce_axis_none
    layout = reducer.apply_many(parts, mask)
  File "/home/angus/Git/awkward-1.0/awkward/reducers.py", line 460, in apply_many
    name = "prod"
  File "/home/angus/Git/awkward-1.0/awkward/reducers.py", line 58, in _apply_many_trivial
    partial_reductions = [
  File "/home/angus/Git/awkward-1.0/awkward/reducers.py", line 59, in <listcomp>
    x.reduce(
  File "/home/angus/Git/awkward-1.0/awkward/contents/content.py", line 894, in reduce
    next = self._reduce_next(
  File "/home/angus/Git/awkward-1.0/awkward/contents/numpyarray.py", line 1090, in _reduce_next
    return self.toRegularArray()._reduce_next(
  File "/home/angus/Git/awkward-1.0/awkward/contents/numpyarray.py", line 1090, in _reduce_next
    return self.toRegularArray()._reduce_next(
  File "/home/angus/Git/awkward-1.0/awkward/contents/numpyarray.py", line 1090, in _reduce_next
    return self.toRegularArray()._reduce_next(
  [Previous line repeated 2970 more times]
  File "/home/angus/Git/awkward-1.0/awkward/contents/numpyarray.py", line 188, in toRegularArray
    out = NumpyArray(self._data.reshape(-1), None, None, self._nplike)
  File "/home/angus/Git/awkward-1.0/awkward/contents/numpyarray.py", line 46, in __init__
    self._data = nplike.asarray(data)
  File "/home/angus/Git/awkward-1.0/awkward/nplikes.py", line 100, in asarray
    return self._module.asarray(*args, **kwargs)
RecursionError: maximum recursion depth exceeded while calling a Python object
```

agoose77:
~Reshaping an ND NumPy with `np.reshape(-1)` will produce a contiguous array as a byproduct.~ However, for 1D arrays, this does not ensure that the result is contiguous. This PR therefore separates these two tests. Additionally, I *think* there was something awry with our contiguity test. I've just add an `nplike` function for this, as it seems like the proper solution anyway.

This PR *doesn't* tackle `unique`, which will need further changes.

Fixes #1847

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-numpy-array-non-contiguous/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -918,3 +918,6 @@ class TypeTracer(ak.nplikes.NumpyLike):
     @classmethod
     def is_own_array(cls, obj) -> bool:
         return isinstance(obj, TypeTracerArray)
+
+    def is_c_contiguous(self, array) -> bool:
+        return True
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -505,24 +505,7 @@ class NumpyArray(Content):
 
     @property
     def is_contiguous(self):
-        if isinstance(self._nplike, ak._typetracer.TypeTracer):
-            return True
-
-        # Alternatively, self._data.flags["C_CONTIGUOUS"], but the following assumes
-        # less of the nplike.
-
-        if isinstance(self.nplike, ak.nplikes.Jax):
-            return True
-
-        x = self._data.dtype.itemsize
-
-        for i in range(len(self._data.shape), 0, -1):
-            if x != self._data.strides[i - 1]:
-                return False
-            else:
-                x = x * self._data.shape[i - 1]
-
-        return True
+        return self._nplike.is_c_contiguous(self._data)
 
     def _subranges_equal(self, starts, stops, length, sorted=True):
         is_equal = ak.index.Index64.zeros(1, self._nplike)
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -1071,8 +1071,8 @@ class NumpyArray(Content):
         keepdims,
         behavior,
     ):
-        if not self.is_contiguous:
-            return self.contiguous()._reduce_next(
+        if self._data.ndim > 1:
+            return self.toRegularArray()._reduce_next(
                 reducer,
                 negaxis,
                 starts,
@@ -1083,9 +1083,8 @@ class NumpyArray(Content):
                 keepdims,
                 behavior,
             )
-
-        elif self._data.ndim > 1:
-            return self.toRegularArray()._reduce_next(
+        elif not self.is_contiguous:
+            return self.contiguous()._reduce_next(
                 reducer,
                 negaxis,
                 starts,
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -1071,7 +1071,20 @@ class NumpyArray(Content):
         keepdims,
         behavior,
     ):
-        if len(self._data.shape) != 1 or not self.is_contiguous:
+        if not self.is_contiguous:
+            return self.contiguous()._reduce_next(
+                reducer,
+                negaxis,
+                starts,
+                shifts,
+                parents,
+                outlength,
+                mask,
+                keepdims,
+                behavior,
+            )
+
+        elif self._data.ndim > 1:
             return self.toRegularArray()._reduce_next(
                 reducer,
                 negaxis,
@@ -1083,6 +1096,10 @@ class NumpyArray(Content):
                 keepdims,
                 behavior,
             )
+
+        # Yes, we've just tested these, but we need to be explicit that they are invariants
+        assert self.is_contiguous
+        assert self._data.ndim == 1
 
         if isinstance(self.nplike, ak.nplikes.Jax):
             from awkward._connect.jax.reducers import get_jax_reducer  # noqa: F401
--- a/src/awkward/nplikes.py
+++ b/src/awkward/nplikes.py
@@ -374,6 +374,9 @@ class NumpyLike(Singleton):
         """
         raise NotImplementedError
 
+    def is_c_contiguous(self, array) -> bool:
+        return array.flags["C_CONTIGUOUS"]
+
 
 class NumpyKernel:
     def __init__(self, kernel, name_and_types):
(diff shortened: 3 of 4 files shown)


jpivarski (approved):
This is a good refactoring. The C++ implementation had to handle contiguity, shapes, and strides by itself, and much of that was ported over when it could instead be checked via flags (or it's always true for TypeTracer and JAX).

The one thing that I worried about is that multidimensional NumpyArray → RegularArrays of NumpyArray require that the underlying buffer is contiguous, and now it looks like that's not being checked on (new) line 1057 of src/awkward/contents/numpyarray.py.

However, if `toRegularArray` needs the buffer to be contiguous, it should be `toRegularArray` that does that. I checked and it does.

https://github.com/scikit-hep/awkward/blob/eb879638f17370ac8e6c2ba499c0c33d4c154d75/src/awkward/contents/numpyarray.py#L175-L187

The key step is `self._data.reshape(-1)`, which will copy a non-contiguous array into a contiguous array so that it can be one-dimensional.

```python
>>> contiguous = np.arange(2*3*5).reshape(2, 3, 5)
>>> not_contiguous = contiguous[:, ::2, 1::2]
>>> contiguous[0, 0, 1] = 99
>>> contiguous
array([[[ 0, 99,  2,  3,  4],
        [ 5,  6,  7,  8,  9],
        [10, 11, 12, 13, 14]],

       [[15, 16, 17, 18, 19],
        [20, 21, 22, 23, 24],
        [25, 26, 27, 28, 29]]])
>>> not_contiguous
array([[[99,  3],
        [11, 13]],

       [[16, 18],
        [26, 28]]])
>>> contiguous_again = not_contiguous.reshape(-1)
>>> contiguous_again
array([99,  3, 11, 13, 16, 18, 26, 28])
>>> contiguous[0, 0, 1] = 1
>>> contiguous_again
array([99,  3, 11, 13, 16, 18, 26, 28])
```

So `toRegularArray` is always safe to call, and everything in this PR is good as-is. I'll merge it!

agoose77:
@jpivarski I actually realised that we've both made an assumption here that doesn't quite match what happens; I too looked at a particular instance of `reshape(-1)` and noted that the result was contiguous. I took that to mean that _all_ non contiguous arrays are copied into contiguous buffers with `reshape`. It turns out that this is not true; a counter-example:
```python
>>> non_contiguous = np.arange(4*3*2)[::2].reshape(3, 4)
>>> non_contiguous.reshape(-1).flags
  C_CONTIGUOUS : False
  F_CONTIGUOUS : False
  OWNDATA : False
  WRITEABLE : True
  ALIGNED : True
  WRITEBACKIFCOPY : False
  UPDATEIFCOPY : False
```

In my tests, I used the transpose, which exchanges the `strides` and therefore doesn't trivially map to a constant stride in the 1D case, requiring a copy. In this counterexample however, the reshape only drops the "outer" stride, recovering the original non-contiguous array.

Luckily, however, the changes in *this* PR ensure that call into `_reduce_next()` at the leaf `NumpyArray` will correctly cast the result to a contiguous buffer. 

I checked the docs, and [NumPy actually mention this](https://numpy.org/doc/stable/reference/generated/numpy.reshape.html#:~:text=Note%20there%20is%20no%20guarantee%20of%20the%20memory%20layout%20(C%2D%20or%20Fortran%2D%20contiguous)%20of%20the%20returned%20array.) So, now we know! 

I don't think there's anything to change though; I believe that we should defer enforcing contiguity as late as possible in most cases, which is what would happen if this counterexample were passed in. Maybe we should add it to the test suite, though

jpivarski:
Good point: `reshape(-1)` doesn't ensure that it's contiguous, but that it can be one-dimensional. But that's what RegularArray needs: RegularArray doesn't have a `strides` parameter, so it has to assume that every element visible in the next node down is a part of the visible 2D array that RegularArray presents (modulo any extra content past `length * size`). But it doesn't need to make any assumptions about the contiguity of the data hidden below the next node down.

So, for instance, this is legal:

```python
>>> import awkward as ak, numpy as np
>>> np_array = np.arange(20)[::2]
>>> np_array.dtype.itemsize != np_array.strides[0]   # not contiguous
True
>>> ak.Array(ak.contents.RegularArray(ak.contents.NumpyArray(np_array), 5))
<Array [[0, 2, 4, 6, 8], [10, 12, 14, 16, 18]] type='2 * 5 * int64'>
```

What the RegularArray needs is for all of the visible NumpyArray values (up to `2 * 5 = 10`) to be visible in the RegularArray. It doesn't care whether there are any hidden values inside the NumpyArray, such as all the odd numbers in that buffer.

So actually this relaxes a too-strict requirement we were imposing in the past. We can't turn a 2D NumpyArray like ("`.`" means hidden)

```
  1  2  3  .  .
  6  7  8  .  .
  .  .  .  .  .
```

into a RegularArray of a 1D NumpyArray without copying it into

```
  1  2  3  6  7  8
```

because there's no way to encode the "show 3, skip 2" in the RegularArray, but what happens inside the RegularArray's `content` is that `content`'s business.

So... "partial contiguity"? Anyway, it's exactly what `ndarray.reshape(-1)` does, so the code is safe.

PR #1857, opened 2022-11-01
docs: import awkward._v2 now raises an explanatory error message.

### LINKED ISSUE #1771: `awkward._v2` imports broken with subpackages
### Version of Awkward Array

2.0.0rc1

In dask-awkward we have some imports that break after installing from the `main` branch (but _not_ removing the `_v2` imports). For example: https://github.com/ContinuumIO/dask-awkward/blob/7a0cceeade6aa8f64f5f3c29fb59c75a807230f7/src/dask_awkward/lib/io/io.py#L9
This is because `awkward._v2.types` used to be package and now it's just a module with the updated repo structure.

Note that dask-awkward is fine if we transition to _not_ importing `_v2` things (https://github.com/ContinuumIO/dask-awkward/pull/86)

jpivarski:
This is a follow-up of https://github.com/scikit-hep/awkward/issues/1771#issuecomment-1298880407.

If you think this can be handled better, let me know here. Otherwise, this is how we'll retire the use of `awkward._v2`.

Cc: @agoose77, @douglasdavis, @martindurant, @alexander-held, @henryiii

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-v2-ModuleNotFoundError-message/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/__init__.py
+++ b/src/awkward/__init__.py
@@ -48,9 +48,6 @@ awkward.behaviors.categorical.register(behavior)  # noqa: F405
 # operations
 from awkward.operations import *
 
-# temporary shim to access v2 under _v2
-import awkward._v2
-
 # version
 __version__ = awkward._ext.__version__
 __all__ = [x for x in globals() if not x.startswith("_")]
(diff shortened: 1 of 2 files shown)


jpivarski:
Oh, this is an open call for reviewers. I'll leave this open until our regular meeting on Thursday.

alexander-held (commented):
I think this is a great solution! I gave the changes a try and everything works as expected.

douglasdavis:
+1

PR #1860, opened 2022-11-03
fix: try to fix long long to double

ianna:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-copy_int_as_complex/> once Read the Docs has finished building :hammer:


src/awkward/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -142,6 +151,23 @@ namespace awkward {
...
+    copy_as(TO_PRIMITIVE* to_ptr, size_t offset) {
+      for (size_t i = 0; i < length_; i++) {
+        double val = static_cast<double>(ptr_.get()[i]);
+        to_ptr[offset++] = TO_PRIMITIVE(val);
+      }
+      if (next_) {
+        next_->copy_as(to_ptr, offset);
+      }
+    }
+
```
ianna, line 187: @jpivarski - this template specialisation makes the Linux compiler happy. However, now I realise that the `copy_as` allows pretty much any type conversion including those that would lead to a data loss... We leave it to a user to make a sensible decision, but I'd rather make it more restrictive implementation. What do you think?

ianna (commented):
I think, we should revisit this code and add the performance tests, and test it beyond the scope of the `ArrayBuilder` usage. For example,  if the `PRIMITIVE` type is the same as the `TO_PRIMITIVE ` type we can skip the for loop.

src/awkward/cpp-headers/awkward/GrowableBuffer.h
jpivarski: Is "the user" in this case code that uses GrowableBuffer? If so, that's just us: LayoutBuilder is user-facing (for users that are downstream developers of C++ applications, or maybe users of RDataFrame), but GrowableBuffer is not. Although the header-only files expose GrowableBuffer as much as they do LayoutBuilder, we haven't documented it or encouraged its use.

If you're asking whether we can have an internal function that converts type X to type Y, possibly losing a lot of data (e.g. X = uint64 and Y = int8, or X = double), then that's fine. A converter function is going to convert the types you give it. Especially since GrowableBuffer is internal, we can do this.

jpivarski (commented):
I'm confused about why this works.

If the compiler is complaining that `long long int` can't be converted to `double` without loss of precision, and this is a more explicit way to convert that type (still with the loss of precision, but without the warning), then that's fine. (That's what the title seems to be saying that this PR is doing.)

But the two implementations of `copy_as` apply to non-`std::complex` and `std::complex`. Since `std::complex` is actually a `struct`, I can see why that one requires special care. But the implementation that applies to `std::complex`,

```c++
for (size_t i = 0; i < length_; i++) {
  double val = static_cast<double>(ptr_.get()[i]);
  to_ptr[offset++] = TO_PRIMITIVE(val);
}
```

is converting through `double`. If that `length_` is the number of complex values, it looks like this would convert complex numbers like `1.1+2.2j` into a real number, `1.1`, and then that into whatever `TO_PRIMITIVE` is; suppose it's `std::complex<float>`: `1.100000023841858+0j`. The problem is that we've lost the imaginary part.

Does this code get that case right? What would be needed to test, say, `std::complex<double` → `std::complex<float>` to be sure that the imaginary parts are not being lost?

src/awkward/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -132,7 +141,10 @@ namespace awkward {
     ///
     /// Changes the data type from `PRIMITIVE` to `TO_PRIMITIVE`/
     template <typename TO_PRIMITIVE>
-    void
+    typename std::enable_if<(!awkward::is_tt<std::complex, TO_PRIMITIVE>::value &&
+                             !awkward::is_tt<std::complex, PRIMITIVE>::value) ||
+                            (awkward::is_tt<std::complex, TO_PRIMITIVE>::value &&
+                             awkward::is_tt<std::complex, PRIMITIVE>::value)>::type
```
ianna, line 147: This template specialization handles any conversion between the arithmetic types or between the complex types. The latter is tested for all combinations of the type of the real and imaginary components: `float`, `double`, and `long double`.

src/awkward/cpp-headers/awkward/GrowableBuffer.h
```diff
@@ -142,6 +154,37 @@ namespace awkward {
...
     }
 
+    template <typename TO_PRIMITIVE>
+    typename std::enable_if<!awkward::is_tt<std::complex, TO_PRIMITIVE>::value &&
+                             awkward::is_tt<std::complex, PRIMITIVE>::value>::type
+    copy_as(TO_PRIMITIVE* to_ptr, size_t offset) {
+      for (size_t i = 0; i < length_; i++) {
+        to_ptr[offset++] = static_cast<TO_PRIMITIVE>(ptr_.get()[i].real());
+        to_ptr[offset++] = static_cast<TO_PRIMITIVE>(ptr_.get()[i].imag());
+      }
```
ianna, line 164: This template specialization expands a complex type array into an arithmetic type array by converting  the real and imaginary components into the arithmetic types. The length of a new buffer is double of the original.

src/awkward/cpp-headers/awkward/GrowableBuffer.h
ianna: ok, thanks!

ianna (commented):
I think, the tests cover (almost) all `std::complex` combinations. The real and imaginary components are copied as expected.

src/awkward/cpp-headers/awkward/GrowableBuffer.h
jpivarski: Okay! I see the tests, and that demonstrates that it's working.

src/awkward/cpp-headers/awkward/GrowableBuffer.h
jpivarski: Oh, I see: because `offset` is incremented _twice_ in this loop over `length_`.

jpivarski:
I've just looked at the updates since my last review, and I sign off on it. I don't know if you're done (your last comment was 35 minutes ago), so feel free to squash-and-merge when you're done.

PR #1865, opened 2022-11-04
test: fastjet clang-format configuration [DO NOT MERGE]

ianna:
@jpivarski suggested to look at a  fastjet/boost-histogram `.clang-format`. Here is a test for discussion. Perhaps we'd like to tune some things?

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-clang-format-config-test/> once Read the Docs has finished building :hammer:

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -189,4 +189,3 @@ WhitespaceSensitiveMacros:
   - NS_SWIFT_NAME
   - CF_SWIFT_NAME
 ...
-
--- a/.clang-format
+++ b/.clang-format
@@ -1,23 +1,191 @@
 ---
 Language:        Cpp
-BasedOnStyle:  Google
-ColumnLimit:     80
-NamespaceIndentation: All
-SortIncludes:    false
-IndentWidth:     2
+# BasedOnStyle:  LLVM
 AccessModifierOffset: -2
-PenaltyBreakComment: 30
-PenaltyExcessCharacter: 100
 AlignAfterOpenBracket: Align
+AlignArrayOfStructures: None
+AlignConsecutiveMacros: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: None
+AlignConsecutiveDeclarations: None
+AlignEscapedNewlines: Right
+AlignOperands:   Align
 AlignTrailingComments: true
 AllowAllArgumentsOnNextLine: true
 AllowAllParametersOfDeclarationOnNextLine: true
-AllowShortBlocksOnASingleLine: false
+AllowShortEnumsOnASingleLine: true
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: All
+AllowShortLambdasOnASingleLine: All
 AllowShortIfStatementsOnASingleLine: Never
-AllowShortLambdasOnASingleLine: false
 AllowShortLoopsOnASingleLine: false
-AlwaysBreakAfterReturnType: All
-BinPackParameters: false
-AlwaysBreakTemplateDeclarations: Yes
-ReflowComments: false
-BinPackArguments: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: MultiLine
+AttributeMacros:
+  - __capability
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterCaseLabel:  false
+  AfterClass:      false
+  AfterControlStatement: Never
+  AfterEnum:       false
+  AfterFunction:   false
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
+  AfterExternBlock: false
+  BeforeCatch:     false
+  BeforeElse:      false
+  BeforeLambdaBody: false
+  BeforeWhile:     false
+  IndentBraces:    false
+  SplitEmptyFunction: true
+  SplitEmptyRecord: true
+  SplitEmptyNamespace: true
+BreakBeforeBinaryOperators: None
+BreakBeforeConceptDeclarations: true
+BreakBeforeBraces: Attach
+BreakBeforeInheritanceComma: false
+BreakInheritanceList: BeforeColon
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializersBeforeComma: false
+BreakConstructorInitializers: BeforeColon
+BreakAfterJavaFieldAnnotations: false
+BreakStringLiterals: true
+ColumnLimit:     80
+CommentPragmas:  '^ IWYU pragma:'
+QualifierAlignment: Leave
+CompactNamespaces: false
+ConstructorInitializerIndentWidth: 4
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DeriveLineEnding: true
+DerivePointerAlignment: false
+DisableFormat:   false
+EmptyLineAfterAccessModifier: Never
+EmptyLineBeforeAccessModifier: LogicalBlock
+ExperimentalAutoDetectBinPacking: false
+PackConstructorInitializers: BinPack
+BasedOnStyle:    ''
+ConstructorInitializerAllOnOneLineOrOnePerLine: false
+AllowAllConstructorInitializersOnNextLine: true
+FixNamespaceComments: true
+ForEachMacros:
+  - foreach
+  - Q_FOREACH
+  - BOOST_FOREACH
+IfMacros:
+  - KJ_IF_MAYBE
+IncludeBlocks:   Preserve
+IncludeCategories:
+  - Regex:           '^"(llvm|llvm-c|clang|clang-c)/'
+    Priority:        2
+    SortPriority:    0
+    CaseSensitive:   false
+  - Regex:           '^(<|"(gtest|gmock|isl|json)/)'
+    Priority:        3
+    SortPriority:    0
+    CaseSensitive:   false
+  - Regex:           '.*'
+    Priority:        1
+    SortPriority:    0
+    CaseSensitive:   false
+IncludeIsMainRegex: '(Test)?$'
+IncludeIsMainSourceRegex: ''
+IndentAccessModifiers: false
+IndentCaseLabels: false
+IndentCaseBlocks: false
+IndentGotoLabels: true
+IndentPPDirectives: None
+IndentExternBlock: AfterExternBlock
+IndentRequires:  false
+IndentWidth:     2
+IndentWrappedFunctionNames: false
+InsertTrailingCommas: None
+JavaScriptQuotes: Leave
+JavaScriptWrapImports: true
+KeepEmptyLinesAtTheStartOfBlocks: true
+LambdaBodyIndentation: Signature
+MacroBlockBegin: ''
+MacroBlockEnd:   ''
+MaxEmptyLinesToKeep: 1
+NamespaceIndentation: None
+ObjCBinPackProtocolList: Auto
+ObjCBlockIndentWidth: 2
+ObjCBreakBeforeNestedBlockParam: true
+ObjCSpaceAfterProperty: false
+ObjCSpaceBeforeProtocolList: true
+PenaltyBreakAssignment: 2
+PenaltyBreakBeforeFirstCallParameter: 19
+PenaltyBreakComment: 300
+PenaltyBreakFirstLessLess: 120
+PenaltyBreakOpenParenthesis: 0
+PenaltyBreakString: 1000
+PenaltyBreakTemplateDeclaration: 10
+PenaltyExcessCharacter: 1000000
+PenaltyReturnTypeOnItsOwnLine: 60
+PenaltyIndentedWhitespace: 0
+PointerAlignment: Right
+PPIndentWidth:   -1
+ReferenceAlignment: Pointer
+ReflowComments:  true
+RemoveBracesLLVM: false
+SeparateDefinitionBlocks: Leave
+ShortNamespaceLines: 1
+SortIncludes:    CaseSensitive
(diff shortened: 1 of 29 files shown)


ianna:
I can see that not all our includes are `iwyu`:
```zsh
/Users/runner/work/awkward/awkward/include/awkward/datetime_util.h:27:21: error: unknown type name 'dtype'
      units_to_format(dtype dt, const std::string& units, int64_t step);
                      ^
```

include/awkward/builder/ArrayBuilder.h
```diff
@@ -1,4 +1,5 @@
-// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+// BSD 3-Clause License; see
+// https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
```
jpivarski: We should turn off whatever it is that rearranges comments. Some comments might have nice ASCII diagrams, and this would ruin that.

include/awkward/builder/ArrayBuilder.h
```diff
@@ -7,369 +8,329 @@
...
-#include "awkward/builder/Builder.h"
 #include "awkward/BuilderOptions.h"
+#include "awkward/builder/Builder.h"
+#include "awkward/common.h"
 
 namespace awkward {
-  class Builder;
-  using BuilderPtr = std::shared_ptr<Builder>;
+class Builder;
+using BuilderPtr = std::shared_ptr<Builder>;
```
jpivarski, line 16: I'm in favor of not indenting inside namespaces (like this). The fact that some functions and classes are inside of a namespace is almost an afterthought in how I look at a C++ file. I wouldn't want adding a layer of namespaceness to cause everything in it to indent and diff poorly (as this is dedenting right now).

include/awkward/builder/ArrayBuilder.h
```diff
@@ -7,369 +8,329 @@
...
+  /// @brief Creates an ArrayBuilder from a full set of parameters.
+  ///
+  /// @param options Configuration options for building an array;
+  /// these are passed to every Builder's constructor.
+  ArrayBuilder(const BuilderOptions &options);
+
+  /// @brief Copy the current snapshot into the BuffersContainer and
+  /// return a Form as a std::string (JSON).
+  const std::string to_buffers(BuffersContainer &container,
+                               int64_t &form_key_id) const;
```
jpivarski: Alignment of arguments is good, but black does this better by not having the initial alignment depend on the length of the name of the function. If we rename this function to something like `to_buffers_xyz`, then all of the arguments would get moved and it would be a bigger diff.

If there's any configuration that would change this to

```suggestion
  const std::string to_buffers(
    BuffersContainer &container,
    int64_t &form_key_id,
  ) const;
```

(i.e. black's style), then it would be _better_, but not _necessary_. Note that [trailing commas are allowed in C++ as of C++11](https://stackoverflow.com/a/6372686/1623645).

include/awkward/builder/ArrayBuilder.h
```diff
@@ -7,369 +8,329 @@
...
-  };
-}
+  /// @brief Constant equal to `nullptr`.
+  static const char *no_encoding;
+  /// @brief Constant equal to `"utf-8"`.
+  static const char *utf8_encoding;
+  /// @brief Root node of the Builder tree.
+  BuilderPtr builder_;
+};
+} // namespace awkward
```
jpivarski, line 214: That's nice!

include/awkward/builder/TupleBuilder.h
```diff
@@ -1,134 +1,110 @@
...
-    const BuilderPtr
-      endrecord() override;
+  const BuilderPtr endrecord() override;
 
-    const BuilderOptions&
-      options() const { return options_; }
+  const BuilderOptions &options() const { return options_; }
 
-    const std::vector<BuilderPtr>& contents() const { return contents_; }
+  const std::vector<BuilderPtr> &contents() const { return contents_; }
```
jpivarski, line 92: This must be that change with respect to LLVM-style that @henryiii was talking about: the `&` (and presumably also the `*`) of pointers are associated with the variable name/function name, rather than the type.

Everybody's got an opinion. _I_ think it should be associated with the type. "`int`" is one type, "`int*`" is another type, and "`int**`" is another type. If the variable names are `x`, `y`, `z`, I would write

```c++
int x;
int* y;
int** z;
```

not

```c++
int x;
int *y;
int **z;
```

The only argument I can think of for the latter has to do with how a series of variable declarations can share type descriptors:

```c++
int x, *y, **z;
```

But we shouldn't be sharing type descriptors, anyway. We should always be explicit.

So if we're getting this feature as a _deviation_ from standard LLVM style, let's go back to standard LLVM style for this one.

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -1,493 +1,411 @@
-// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+// BSD 3-Clause License; see
+// https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
-#define FILENAME(line) FILENAME_FOR_EXCEPTIONS("src/libawkward/builder/ArrayBuilder.cpp", line)
+#define FILENAME(line)                                                         \
+  FILENAME_FOR_EXCEPTIONS("src/libawkward/builder/ArrayBuilder.cpp", line)
```
jpivarski, line 4: Much like the two-line LICENSE comment, this multi-line split is really unnecessary. It might be harder to explain to clang-format that we want it off, though.

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -1,493 +1,411 @@
...
 
-  void
-  ArrayBuilder::clear() {
-    if (builder_) {
-      builder_.get()->clear();
-    }
+void ArrayBuilder::clear() {
+  if (builder_) {
+    builder_.get()->clear();
   }
```
jpivarski, line 29: I'm glad that one-line if statements still get the braces. We should use braces all the time, uniformly.

```c++
  if (builder_)
    builder_.get()->clear();
```

is dangerous.

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -1,493 +1,411 @@
...
   return 0;
 }
 
-uint8_t awkward_ArrayBuilder_null(void* arraybuilder) {
-  awkward::ArrayBuilder* obj =
-    reinterpret_cast<awkward::ArrayBuilder*>(arraybuilder);
+uint8_t awkward_ArrayBuilder_null(void *arraybuilder) {
+  awkward::ArrayBuilder *obj =
+      reinterpret_cast<awkward::ArrayBuilder *>(arraybuilder);
```
jpivarski, line 164: Yes, it's doing it with the `*` as well as the `&`...

src/libawkward/builder/ArrayBuilder.cpp
```diff
@@ -1,493 +1,411 @@
...
+uint8_t awkward_ArrayBuilder_boolean(void *arraybuilder, bool x) {
+  awkward::ArrayBuilder *obj =
+      reinterpret_cast<awkward::ArrayBuilder *>(arraybuilder);
   try {
     obj->boolean(x);
-  }
-  catch (...) {
+  } catch (...) {
     return 1;
   }
```
jpivarski, line 180: This is another thing that I would want to turn off. I'm assuming that this feature would also make

```c++
if (predicate_1) {
  consequent_1;
} else if (predicate_2) {
  consequent_2;
} else {
  alternate;
}
```

But that needs more room to breathe:

```c++
if (predicate_1) {
  consequent_1;
}
else if (predicate_2) {
  consequent_2;
}
else {
  alternate;
}
```

The latter takes up more lines, but all block-starting statements are now at the same level of depth and all closing brackets are equally visible.

src/libawkward/builder/StringBuilder.cpp
```diff
@@ -11,212 +13,181 @@
...
-    if (length < 0) {
-      for (int64_t i = 0;  x[i] != 0;  i++) {
-        content_.append((uint8_t)x[i]);
-      }
+const BuilderPtr StringBuilder::string(const char *x, int64_t length,
+                                       const char * /* encoding */) {
+  if (length < 0) {
+    for (int64_t i = 0; x[i] != 0; i++) {
+      content_.append((uint8_t)x[i]);
     }
```
jpivarski, line 129: Here's something that I don't expect it to preserve. For decades, I've been writing for loops like this:

```c++
for (int i = 0;  i < 10;  i++) {
```

rather than

```c++
for (int i = 0; i < 10; i++) {
```

The three parts of the for header are statements, separated by semicolons, and so I give them more room than comma-separated list data. But I think I'm the only person who does that—I never attempted to popularize the style.

So clang-format almost certainly doesn't have an option for it, and that's fine.

jpivarski (commented):
@henryiii had a good idea of fixing the version of clang-format and expressing the `.clang-format` file as a difference—a few options to change relative to LLVM style, rather than a dump of options from some version of LLVM style. (Now that that's possible.)

I put a few formatting opinions inline. Don't try too hard to convince clang-format to provide them, but if it's easy, they're what I think is most readable.

If there's a C++ style ("LLVM style" being just one example of a style) that is closer to black, that could be better. Black is designed to interact with diff well, and that's as much of a good feature for C++ as it is for Python.

(I know; I resisted black when it was first introduced to the Python codebase, and now I'm asking for the C++ to be more black-like. I've been converted.)

include/awkward/builder/ArrayBuilder.h
ianna:
```diff
-PenaltyBreakComment: 300
+PenaltyBreakComment: 300000000
```

include/awkward/builder/ArrayBuilder.h
ianna: I did not find a simple option for that - unless we go for a `Custom` style.

I kind of like this additional option aligning parameter names, but it will only apply if the line is longer than a limit.
```diff
-AlignConsecutiveDeclarations: None
+AlignConsecutiveDeclarations: true
```
```c++ 
  /// @brief Copy the current snapshot into the BuffersContainer and
  /// return a Form as a std::string (JSON).
  const std::string to_buffers(BuffersContainer &container,
                               int64_t          &form_key_id) const;
```
also it's not very consistent:
```c++
  void concatenate_to(
      PRIMITIVE *to_ptr,
      size_t offset) const noexcept {
    memcpy(to_ptr + offset,
           reinterpret_cast<void *>(
               ptr_.get()),
           length_ * sizeof(PRIMITIVE));
    if (next_) {
      next_->concatenate_to(
          to_ptr, offset + length_);
    }
  }

  /// @brief Fills (one panel) GrowableBuffer<TO_PRIMITIVE> with the
  /// elements of (possibly multi-panels) GrowableBuffer<PRIMITIVE>.
  ///
  /// Changes the data type from `PRIMITIVE` to `TO_PRIMITIVE`/
  template <typename TO_PRIMITIVE>
  void copy_as(TO_PRIMITIVE *to_ptr,
               size_t        offset) {
    for (size_t i = 0; i < length_;
         i++) {
      to_ptr[offset++] =
          static_cast<TO_PRIMITIVE>(
              ptr_.get()[i]);
    }
    if (next_) {
      next_->copy_as(to_ptr, offset);
    }
  }

```

include/awkward/builder/TupleBuilder.h
ianna: This is a default option.

include/awkward/builder/ArrayBuilder.h
agoose77: Hmm, it looks like it does it in the `for` loop, which is good, but I think it's too aggressive here - the line length isn't too long yet.

RE the argument parameters, maybe we want to set `InsertTrailingCommas` and `AllowAllParametersOfDeclarationOnNextLine`?

Sympathies with you @ianna, I've done this before and it's painful :(

.clang-format
```diff
@@ -189,4 +189,3 @@ WhitespaceSensitiveMacros:
   - NS_SWIFT_NAME
   - CF_SWIFT_NAME
 ...
-
```
ianna: This file was created with:
```bash
% clang-format --version
clang-format version 14.0.6
```
and
```
clang-format -style=llvm -dump-config > .clang-format
```

ianna (commented):
@jpivarski - This is the least invasive modification - there are only two options different from the default configuration. I've tried different "brake after brackets" styles - it seems that none of them would work asis: we'd need to go for a "Custom" style.

.clang-format
ianna: The recipe to reproduce and compare the changes as follows:
```bash
for filename in ./include/awkward/builder/*.h; do clang-format "${filename}" >& "${filename}-new"; done
for filename in ./src/libawkward/builder/*.cpp; do clang-format "${filename}" >& "${filename}-new"; done
```
then:
```bash
for filename in ./include/awkward/builder/*.h; do mv "${filename}-new" "${filename}"; done
for filename in ./src/libawkward/builder/*.cpp; do mv "${filename}-new" "${filename}"; done
```

src/libawkward/builder/ArrayBuilder.cpp
jpivarski: It's something people care about:

 - https://stackoverflow.com/q/65452234/1623645
 - https://stackoverflow.com/q/31976583/1623645

though neither of these are clang-format (vscode and Visual Studio).

jpivarski:
Now that the packages have been split, this .clang-format can be reapplied in a new PR (with the modifications described above). Thanks!

ianna (commented):
@jpivarski - I'm not sure what to do with this one. There is an issue open and linked to it. It looks like the branch was deleted and restored.

jpivarski:
We still need to decide on what C++ formatting to adopt. The PR is useful because it shows what settings were used, but the branch doesn't need to be a named branch within our branches list. It's always accessible from this PR page. (GitHub doesn't actually delete branches; we're just talking about removing it from the branches list.)

PR #1867, opened 2022-11-05
fix: propagate local behaviors in ak.combinations

### LINKED ISSUE #1866: ak.combinations drops types when accessing components
### Version of Awkward Array

2.0.0rc2

NanoEvents objects lose their behavior when run through combinations.

reproducer:
```
$ git clone https://github.com/CoffeaTeam/coffea.git -b awkward2_dev
$ cd coffea
$ pip install -e '.[dev]'
$ python
>>> from coffea.nanoevents import NanoEventsFactory
>>> factory = NanoEventsFactory.from_root("tests/samples/nano_dy.root")
>>> events = factory.events()
>>> events.Muon[4][0] + events.Muon[4][1] # to show it works in the first place
>>> combos = ak.combinations(events.Muon, 2)
>>> dimuons = combos["0"] + combos["1"]
```

Fails to add pairs of muon objects together.

You can add print outs for the behaviors of the objects and you will find that they are `None` while `events.Muon` is not.

lgray:
This fixes relevant tests in coffea.

This is a possible fix to #1866, and indeed makes the repro there work, but I'm sure a more correct implementation (that also covers other similar operations) is possible.

### DIFF
--- a/src/awkward/operations/ak_combinations.py
+++ b/src/awkward/operations/ak_combinations.py
@@ -177,13 +177,15 @@ def combinations(
             parameters,
             with_name,
             highlevel,
-            array.behavior if behavior is None else behavior,
+            behavior,
         )
 
 
 def _impl(
     array, n, replacement, axis, fields, parameters, with_name, highlevel, behavior
 ):
+    behavior = ak._util.behavior_of(array, behavior=behavior)
+
     if parameters is None:
         parameters = {}
     else:
--- a/src/awkward/operations/ak_combinations.py
+++ b/src/awkward/operations/ak_combinations.py
@@ -177,7 +177,7 @@ def combinations(
             parameters,
             with_name,
             highlevel,
-            behavior,
+            array.behavior if behavior is None else behavior,
         )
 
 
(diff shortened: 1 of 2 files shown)


jpivarski:
Thanks for finding it!

The only change that I've made is to keep the `_impl` call strictly pass-through and use the `ak._util.behavior_of` function, which exists for this purpose. It defines rules for which behaviors win in case of conflict (which can be changed centrally if we ever decide to do that, which I doubt).

jpivarski:
Oh, I haven't tried running your reproducer yet. I'll take a look and see if it can be turned into a test.

jpivarski:
Yes, this definitely toggles the reproducer. (I had to go hunting for nano_dy.root; I shouldn't have deleted it after the last issue!)

The reproducer can't be included as a test as-is, but I think I can make an alternative.

Or—it seems you've already done that: https://github.com/scikit-hep/awkward/issues/1866#issuecomment-1304562978.

jpivarski (approved):
Even though I added to it, I really just took @lgray's work and put it into a test file, so I think it's not off-limits for me to approve this PR. (I.e. this doesn't count as a "self-approval.")

Thanks, @lgray! I'll enable auto-merge.

PR #1868, opened 2022-11-07
test: cuda status tests

ianna:


tests-cuda/test_1809-cuda-overload.py
```diff
@@ -0,0 +1,38 @@
...
+    def sum_impl(arr):
+        res = 0 
+        for i in range(len(arr)):
+            res += arr[i]
+        return res 
+    return sum_impl
+
+@cuda.jit
+def f(arr):
+    print("Sum is", arr.sum())
```
ianna, line 32: @gmarkall - I'm having a problem with overloading this method copied from your notebook. Could you, please, check if I'm missing something? Thanks.

Here is the error:
```python
>               raise newerr.with_traceback(tb)
E               numba.core.errors.LoweringError: Failed in cuda mode pipeline (step: native lowering)
E               No definition for lowering (<class 'numba.core.types.npytypes.Array'>, 'sum')(array(int64, 1d, C),) -> int64
E               
E               File "tests-cuda/test_1809-cuda-overload.py", line 32:
E               def f(arr):
E                   print("Sum is", arr.sum())
E                   ^
E               
E               During: lowering "$10call_method.4 = call $8load_method.3(func=$8load_method.3, args=[], kws=(), vararg=None, varkwarg=None, target=None)" at /eos/home-y/yana/Projects/PR1809/awkward/tests-cuda/test_1809-cuda-overload.py (32)
```

ianna (commented):
@gmarkall - I'm using:
```
>>> numba.__version__
'0.56.3'
```
and the function overload (from the same tutorial) works as expected. Could you, please, check? Thanks.

tests-cuda/test_1809-cuda-overload.py
gmarkall: I'm getting the same failure with Numba 0.56.3 - I think that correct operation of the overload API requires a fix that only went into 0.56.4: https://github.com/numba/numba/pull/8570.

Unfortunately Numba 0.56.4 is not yet available in conda-forge (see https://github.com/conda-forge/numba-feedstock/pull/108) or the default channels, but it can be installed from the Numba channel:

```
$ mamba search -c numba numba

# ... lots of output snipped ...

numba                         0.56.4 np1.11py3.7hc13618b_g288a38bbd_0  numba               
numba                         0.56.4 np1.11py3.8hc13618b_g288a38bbd_0  numba               
numba                         0.56.4 np1.16py3.9h59fbc97_g288a38bbd_0  numba               
numba                         0.56.4 np1.21py3.10ha54aa7d_g288a38bbd_0  numba     
```

ianna: Thanks! I confirm that it works - after pip installing `numba` 0.56.4.

jpivarski:
How is this related to #1809?

PR #1869, opened 2022-11-07
fix: ensure that behaviors are propagated through `ak.XXX` operations

agoose77:
#1866 wasn't the first time we've seen a regression in behavior preservation, so I've gone through and checked all of our other operations. This PR hopefully catches some missed cases, and ensures that we test both explicit `behavior=some_dict` and `behavior=None` argument behaviour.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-propagate-behavior/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_argcombinations.py
+++ b/src/awkward/operations/ak_argcombinations.py
@@ -96,4 +96,4 @@ def _impl(
         out = layout.combinations(
             n, replacement=replacement, axis=axis, fields=fields, parameters=parameters
         )
-        return ak._util.wrap(out, behavior, highlevel)
+        return ak._util.wrap(out, behavior, highlevel, like=array)
--- a/src/awkward/operations/ak_categories.py
+++ b/src/awkward/operations/ak_categories.py
@@ -35,11 +35,7 @@ def _impl(array, highlevel):
             return None
 
     layout = ak.operations.to_layout(array, allow_record=False, allow_other=False)
-    layout.recursively_apply(action)
+    behavior = ak._util.behavior_of(array)
+    layout.recursively_apply(action, behavior=behavior)
 
-    if output[0] is None:
-        return None
-    elif highlevel:
-        return ak._util.wrap(output[0], ak._util.behavior_of(array))
-    else:
-        return output[0]
+    return ak._util.wrap(output[0], behavior, highlevel)
--- a/src/awkward/operations/ak_argsort.py
+++ b/src/awkward/operations/ak_argsort.py
@@ -59,4 +59,4 @@ def argsort(array, axis=-1, ascending=True, stable=True, highlevel=True, behavio
 def _impl(array, axis, ascending, stable, highlevel, behavior):
     layout = ak.operations.to_layout(array, allow_record=False, allow_other=False)
     out = layout.argsort(axis, ascending, stable)
-    return ak._util.wrap(out, behavior, highlevel)
+    return ak._util.wrap(out, behavior, highlevel, like=array)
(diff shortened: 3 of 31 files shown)


jpivarski (approved):
At first, I was worried about this PR because some operations must pass on parameters and other operations must not. (If the output is conceptually a different type, _parameters_ must not be passed on.) However, you're talking about _behaviors_, which should always go through.

The early experience with Coffea (cc @lgray) was more about finding all the places where we needed to pass on _parameters_, not _behaviors_. A hint to the fact that I misspoke (miswrote) in https://github.com/scikit-hep/awkward/issues/1866#issuecomment-1304581419 is that I referred to the C++ to Python translation; only parameters had been in C++ in v1, not behaviors.

So a PR ensuring that all behaviors always pass through is welcome!

This PR looks good; the only thing I would say is that `ak._util.wrap`'s `like` parameter is redundant with explicitly calling `ak._util.behavior_of`. Sometimes, we use `like`, and sometimes we use `behavior_or` (more often the latter). It's an implementation detail that can be cleaned up later, so I'm just noting it in passing.

This is good as-is, I'll merge it!

agoose77:
@jpivarski in this PR I intended on replacing `behaviorof` with `wrap(..., like=)` where possible to improve readability. Did I happen to miss some places and introduce a surplus `like=`?

src/awkward/operations/ak_argcombinations.py
```diff
@@ -96,4 +96,4 @@ def _impl(
         out = layout.combinations(
             n, replacement=replacement, axis=axis, fields=fields, parameters=parameters
         )
-        return ak._util.wrap(out, behavior, highlevel)
+        return ak._util.wrap(out, behavior, highlevel, like=array)
```
jpivarski, line 99: This is an introduction of `like` (furthering its use).

src/awkward/operations/ak_categories.py
```diff
@@ -35,11 +35,7 @@ def action(layout, **kwargs):
...
-    if output[0] is None:
-        return None
-    elif highlevel:
-        return ak._util.wrap(output[0], ak._util.behavior_of(array))
-    else:
-        return output[0]
+    behavior = ak._util.behavior_of(array)
+    layout.recursively_apply(action, behavior=behavior)
+
+    return ak._util.wrap(output[0], behavior, highlevel)
```
jpivarski, line 41: And this is an introduction of `behavior_of` (furthering its use).

The reason it was done here is so that the `behavior` can be used in `recursively_apply`. There are a few other examples like this, in which the `behavior_of` is far from the `recursively_apply` (or `broadcast_and_apply`). This is correct, by the way: it's the merged behavior that these functions want.

(For this reason, I think the right solution to the "which internal interface should we use?" is `behavior_of`, not `like`.)

agoose77: Oh, I think I see what you're saying - that we should eliminate one of these interfaces? I originally introduced the `like` mechanism to remove boilerplate in our internal interfaces where we just want to copy the behavior from the existing array. If we wanted to do this, then I'd also be in favour of removing `like` - it's not as powerful.

I did also wonder whether we should have a `to_layout` variant that returns both the layout and behavior, but I decided that it's too specialised and better to be explicit as we are now.

jpivarski: Since it turns out that the boilerplate is sometimes necessary (because `behavior` is used in `recursively_apply`, as well as the output array), `like` is not a simplification. I was thinking that we'd eventually cull one of the interfaces, and since `behavior_of` is more general than the `like` argument, it's the `like` argument that would go.

Alternatively, the `like` argument could just get renamed as a `behavior_of` _argument_. It would still be done in two ways (explicit function call vs argument), but having the same name would clue-in code readers/maintainers that they're the same thing. Each high-level function should have exactly one appearance of "`behavior_of`", whether it's the function or the argument.

agoose77: @jpivarski I like the suggestion to rename `like`. It was never a good name to begin with. :eyes:

PR #1871, opened 2022-11-09
fix: set appropriate error message for decimal types in arrow

### LINKED ISSUE #1870: Bad error message for decimal type
### Version of Awkward Array

main

If you attempt to load a parquet file containing decimal columns, you get a ValueError, which is expected, but the message is:
```
"Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
```
which looks like it's been copied from somewhere else.

https://github.com/scikit-hep/awkward/blob/main/src/awkward/_connect/pyarrow.py#L476

agoose77:
Fixes #1870 

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-from-arrow-error/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -473,7 +473,7 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
         # NumPy doesn't support decimal: https://github.com/numpy/numpy/issues/9789
         raise ak._errors.wrap_error(
             ValueError(
-                "Arrow arrays containing Decimal types can't be converted into Awkward Arrays"
+                "Arrow arrays containing pyarrow.decimal128 or pyarrow.decimal256 types can't be converted into Awkward Arrays"
             )
         )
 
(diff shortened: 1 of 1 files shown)


src/awkward/_connect/pyarrow.py
```diff
@@ -746,7 +746,7 @@ def form_popbuffers(awkwardarrow_type, storage_type):
         # NumPy doesn't support decimal: https://github.com/numpy/numpy/issues/9789
         raise ak._errors.wrap_error(
             ValueError(
-                "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
+                "Arrow arrays containing Decimal types can't be converted into Awkward Arrays"
```
jpivarski:
```suggestion
                "Arrow arrays containing pyarrow.decimal128 or pyarrow.decimal256 types can't be converted into Awkward Arrays"
```

src/awkward/_connect/pyarrow.py
```diff
@@ -473,7 +473,7 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
...
```
jpivarski:
```suggestion
                "Arrow arrays containing pyarrow.decimal128 or pyarrow.decimal256 types can't be converted into Awkward Arrays"
```

jpivarski (approved):
Just making it a little more specific, so users know what to look for.

PR #1873, opened 2022-11-10
feat: better mask_identity defaults for reducer-like functions.

### LINKED ISSUE #1320: For NumPy compliance, `ak.mean` (and possibly others) should turn None into NaN
### Version of Awkward Array

HEAD

```python
>>> np.mean(np.array([[], [], []]), axis=1)
/home/jpivarski/mambaforge/lib/python3.9/site-packages/numpy/core/fromnumeric.py:3440: RuntimeWarning: Mean of empty slice.
  return _methods._mean(a, axis=axis, dtype=dtype,
/home/jpivarski/mambaforge/lib/python3.9/site-packages/numpy/core/_methods.py:181: RuntimeWarning: invalid value encountered in true_divide
  ret = um.true_divide(
array([nan, nan, nan])
```

versus

```python
>>> ak.mean(ak._v2.Array([[], [], []]), axis=1)
<Array [None, None, None] type='3 * ?float64'>
```

Identified by @douglasdavis. We saw it in a typetracer, but the typetracer follows what the Array does.

jpivarski:
Motivated by a difference with respect to NumPy: #1320.

The new defaults have been more carefully thought through.

  * monoidal reducers like "sum" (which have identities) have `mask_identity=False`
  * semigroup reducers like "argmax" (no identity) have `mask_identity=True`
  * functions defined in terms of reducers with "count" in the denominator have `mask_identity=False` so that the `np.nan` appears for empty lists (just as it does for NumPy)
  * functions defined in terms of reducers, but are on the whole map-like, rather than reducer-like: `mask_identity=False`
  * functions defined in terms of reducers, computed from "min" and "max": `mask_identity=True`

I also ensured that we're using `np.errstate` consistently. We want to ignore `np.nan` operations and divisions by zero.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-mask_identity-defaults/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_corr.py
+++ b/src/awkward/operations/ak_corr.py
@@ -11,7 +11,7 @@ def corr(
     weight=None,
     axis=None,
     keepdims=False,
-    mask_identity=True,
+    mask_identity=False,
     flatten_records=False,
 ):
     """
@@ -82,7 +82,7 @@ def _impl(x, y, weight, axis, keepdims, mask_identity, flatten_records):
             ak.operations.to_layout(weight, allow_record=False, allow_other=False)
         )
 
-    with np.errstate(invalid="ignore"):
+    with np.errstate(invalid="ignore", divide="ignore"):
         xmean = ak.operations.ak_mean._impl(
             x, weight, axis, False, mask_identity, flatten_records
         )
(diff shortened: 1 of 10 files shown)


agoose77 (commented):
Looks good to me. I like this change!

agoose77 (approved):

PR #1876, opened 2022-11-12
chore: rename is_XXXType to is_XXX (lowercase)

### LINKED ISSUE #1681: Rename the `is_XXXType` flags in `Content` et al.
### Version of Awkward Array

main

Right now, these create confusion (as evidenced in https://github.com/scikit-hep/awkward/discussions/1654#discussioncomment-3527347), and we could choose a better name. `is_list_like` might be a better variant of `is_ListType`.

agoose77:
Fixes #1681 

This was done with `sed`.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-chore-rename-is-XXX-type/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -873,7 +873,7 @@ def to_awkwardarrow_type(
 
 
 def remove_optiontype(akarray):
-    assert type(akarray).is_OptionType
+    assert type(akarray).is_option
     if isinstance(akarray, ak.contents.IndexedOptionArray):
         return ak.contents.IndexedArray(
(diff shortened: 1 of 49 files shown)


jpivarski (approved):
I approve of the new attribute names, and after that, it's an easy find-and-replace.

We've given examples using the `is_XXXType` names to some early adopters. When they notice that it has changed, we'll need to point them here to see what the new names are. Hopefully, it's easy to figure out by `is_` <tab> in some interactive IDE.

But anyway, this change would have been a lot more difficult if it were after the 2.0.0 release!

Issue #1261, opened 2022-01-28
Matrix multiplication (@ operator) is broken and needs to be ported to kernels (out of Numba)


alexander-held:
### Version of Awkward Array

1.7.0

### Description and code to reproduce

While trying to use awkward arrays in matrix multiplication, I ran into two different kinds of unexpected behavior. The following code shows an example.

```python
import awkward as ak
import numpy as np

v = np.asarray([1,2,3])
A = np.asarray([[1,0,0],[0,1,0],[0,0,1]])
vT1 = np.asarray([[1],[2],[3]])
vT2 = np.asarray([1,2,3])

print(v @ A @ vT1)  # [14]
print(v @ A @ vT2)  # 14

v = ak.Array([1,2,3])
A = ak.Array([[1,0,0],[0,1,0],[0,0,1]])
vT1 = ak.Array([[1],[2],[3]])
vT2 = ak.Array([1,2,3])

print(v @ A @ vT1)  # output [[a], [b], [c]], different values for a, b, c each time
print(v @ A @ vT2)  # ValueError: in ListOffsetArray64 attempting to get 3, stops[i] > len(content)
```

In the first case with `vT1`, uninitialized data seems to be accessed, leading to different outputs every time. The output shape also changes when compared to the `numpy` approach. I naively expected the second case with `vT2` to also work, and to report the same result as `numpy`. I believe the different treatment for this may be on purpose and a design choice, but wanted to include it here as well to be sure this is expected.

jpivarski:
I made #1324 and mentioned @lukasheinrich in it, but that's a duplicate of this one, so I'll close that and prioritize this.

jpivarski:
This is harder than we thought and should be deferred beyond the next release.

`np.matmul` on Awkward Arrays should raise `NotImplementedError` for now.

PR #1877, opened 2022-11-14
feat: np.matmul should raise NotImplementedError until it gets implemented

### LINKED ISSUE #1874: `np.matmul` should raise `NotImplementedError` until it gets implemented.
This is harder than we thought and should be deferred beyond the next release.

`np.matmul` on Awkward Arrays should raise `NotImplementedError` for now.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/issues/1261#issuecomment-1310986350_

ioanaif:
Given that the matrix multiplication (@ operator) is broken (as discussed in #1261), this PR sets `matmul` to raise `NotImplementedError` until a correct implementation for `matmul` is developed. 

This fixes #1874 

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ioanaif-matmul-notimplemented/> once Read the Docs has finished building :hammer:


src/awkward/_connect/numpy.py
```diff
@@ -141,9 +141,10 @@ def action(inputs, **ignore):
             return _array_ufunc_adjust(custom, inputs, kwargs, behavior)
 
         if ufunc is numpy.matmul:
-            custom_matmul = action_for_matmul(inputs)
-            if custom_matmul is not None:
-                return custom_matmul()
+            raise NotImplementedError
```
agoose77: I *think* we want to return `NotImplemented` rather than raise an error here
```suggestion
            return NotImplemented
```

The docs (https://numpy.org/doc/stable/user/basics.subclassing.html?highlight=numpy%20memmap%20all#array-ufunc-for-ufuncs, https://numpy.org/doc/stable/reference/arrays.classes.html#numpy.class.__array_ufunc__) aren't super clear on this, but I think that's the guidance. 

My understanding is that `return NotImplemented` will allow other arguments to implement `matmul`, e.g. an `ak.Array(...) @ np.array()` where the `ak.Array` is regular.

ioanaif: It should raise an error because at the moment matmul is not implemented for any of the cases (not even the one you mention, ak.Array and np.array). I have always raised NotImplemented when working on awkward, but if the protocol has changed, I can change it as well. @jpivarski

ioanaif: > My understanding is that `return NotImplemented` will allow other arguments to implement `matmul`, e.g. an `ak.Array(...) @ np.array()` where the `ak.Array` is regular.

It would exit the function on `return` as it does on `raise`

agoose77: N.b. the following is here for the benefit of keeping track of my thought processes :)

The `return NotImplemented` mechanism is usually used when there is the option for third-parties to implement the method, i.e. it's an agreed protocol:
```python
class IntLike:
    def __init__(self, value):
        self._value = value
    def __radd__(self, other):
        assert isinstance(other, int)
        return type(self)(self._value + other)
    def __repr__(self):
        return f"{type(self).__name__}({self._value!r})"

1 + IntLike(2)
```
i.e. `int` doesn't know about `IntLike`, so its `__add__` returns `NotImplemented` to allow `IntLike.__radd__(..., 1)` to be invoked. I *think* NumPy does something similar here; if we tell NumPy that we don't implement `matmul`, I would imagine that it will invoke NumPy's `matmul`. But, let me check!

I seemed to remember a NumPy passthrough that would see `np.asarray` being called on the Awkward types. This was the fallback mechanism that I expected to work here. However, my understanding of how this happens is not clear; if we return `NotImplemented` from the `__array_ufunc__()`, NumPy rightly doesn't know what to do with the Awkward Array type, and errors out. What I expected to see happen was that NumPy would try coercing each argument to a NumPy Array, and then try `np.matmul`. 

I'm fairly sure that there is a mechanism like this, but I think it might only be for `__array_function__`. Maybe, then, what we should do here is to try and coerce all `ak.Content`s into NumPy `ndarray`s, and call `np.matmul`? This would clearly need to be done outside the `apply`, as we need the entirety of each array to be regular.

jpivarski: That's right, `return NotImplemented` lets other overloads take over, while `raise NotImplementedError` is a full-stop: no computation takes place. (These two cases unfortunately look very similar to one another.)

Doing a `return NotImplemented` would allow NumPy to take over, and it would try to cast the Awkward Array as a NumPy array by calling its `__array__` function. However, returning `NotImplemented` at this point in the code might not do it; we'd need to get the return value all the way back to the original `__array_ufunc__` call, if it needs to go through layers of the call stack in `broadcast_and_apply`, it might break something unrelated. To actually `return NotImplemented`, we might need to throw and catch an internal, custom exception, just so that we can skip most of the call stack and get the `NotImplemented` object directly out of `__array_ufunc__`.

Asking the question, "What _should_ we do?" I'm leaning more toward `raise NotImplementedError`, so that `@` always fails with an error message that indicates that this operator can and will be handled in the future; we might end up soliciting requests to implement it sooner (finding out just how many people need it). Doing `return NotImplemented` (properly) would make `@` work for regular arrays and not work for irregular arrays with a weird error message, most likely, "This array is not convertible to NumPy."

But in order for that to be useful, this PR should add an error message:

```python
            raise ak._errors.wrap_error(
                NotImplementedError(
                    "matrix multiplication (`@` or `np.matmul`) is not yet implemented for Awkward Arrays"
                )
            )
```

> I'm fairly sure that there is a mechanism like this, but I think it might only be for `__array_function__`.

Yes, the mechanism that returns `NotImplemented` to allow NumPy to try to cast the Awkward Array and use its own function is for NEP 18 (`__array_function__`), not NEP 13 (`__array_ufunc__`). With the exception of matrix multiplication, all ufuncs go through a common mechanism that applies to all Awkward Arrays except those with records (because we deliberately prevented it). None of the ufuncs (currently) `return NotImplemented` to let NumPy cast it.

src/awkward/_connect/numpy.py
```diff
@@ -141,9 +141,10 @@ def action(inputs, **ignore):
             return _array_ufunc_adjust(custom, inputs, kwargs, behavior)
 
         if ufunc is numpy.matmul:
-            custom_matmul = action_for_matmul(inputs)
-            if custom_matmul is not None:
-                return custom_matmul()
+            raise ak._errors.wrap_error(NotImplementedError)
+            # custom_matmul = action_for_matmul(inputs)
+            # if custom_matmul is not None:
+            #     return custom_matmul()
```
jpivarski:
```suggestion
            raise ak._errors.wrap_error(
                NotImplementedError(
                    "matrix multiplication (`@` or `np.matmul`) is not yet implemented for Awkward Arrays"
                )
            )
```

jpivarski (changes requested):
My only request is to add an error message, such as this one. If you have a better one, of course I'd take that!

jpivarski (approved):
Looks good to me! I think it can be merged.

@agoose77, is the `raise NotImplementedError`, rather than `return NotImplemented` okay with you?

src/awkward/_connect/numpy.py
ioanaif: > None of the ufuncs (currently) `return NotImplemented` to let NumPy cast it.

This was my motivation when choosing `raise`.

agoose77: @jpivarski I tried enabling `return NotImplemented`, but as we both remark, it looks like ufuncs don't implement the `np.asarray` fallback that would catch this case, so it fails just the same.
> Doing return NotImplemented (properly) would make @ work for regular arrays and not work for irregular arrays with a weird error message, most likely, "This array is not convertible to NumPy."

We could add an exception for the non-NumPy case, which is what we do for `__array_function__` cases.

@ioanaif thanks for the discussion; the question of whether to do fallbacks is a bit murky, and I can see that there are varying opinions on this one.

agoose77:
Yes; although we *could* enable regular-array support for matmul, I can be convinced that it's little-used enough in the awkward context that we can let users convert to NumPy in the interim if needed. Thanks all!

agoose77 (approved):

PR #1887, opened 2022-11-18
fix: use env var for Python version

agoose77:
The codecov workflow should use the same Python environment as the build-test coverage run, so that our baseline reporting is accurate. See #1888 for more information.


agoose77:
I'm merging this without testing; it *only* touches the Coverage workflow, which runs on main.

PR #1891, opened 2022-11-18
wip: run deploy for PR

agoose77:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-ci-fix-deploy-workflow/> once Read the Docs has finished building :hammer:

### DIFF
--- a/.github/workflows/deploy-cpp.yml
+++ b/.github/workflows/deploy-cpp.yml
@@ -70,7 +70,6 @@ jobs:
       run: |
         import subprocess
         import os
-        import pathlib
         result = subprocess.run(
           ["git", "log", "-1", '--format=%ad', '--date=unix', '--', 'pyproject.toml'],
           cwd="awkward-cpp",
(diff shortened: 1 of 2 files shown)


agoose77:
@jpivarski [this commit](https://github.com/scikit-hep/awkward/actions/runs/3497113584/jobs/5855815030) tested the deployment for Python, whilst [this commit](https://github.com/scikit-hep/awkward/actions/runs/3496389220/jobs/5854246327) tested the C++.

I installed the wheels using
```bash
cd $(mktemp-d)
python3 -m venv .venv
source .venv/bin/activate

# Populate wheels
mkdir wheels
python3 -m pip wheel packaging numpy pyparsing -w ./wheels
unzip <PATH-TO-CPP-WHEELS> -d wheels
unzip <PATH-TO-PY-WHEELS>  -d wheels

# Install awkward
python3 -m pip install --pre awkward --find-links=./wheels --no-index

# Test
python3 -c "import awkward as ak; builder=ak.ArrayBuilder();builder.integer(100);print(builder.snapshot())"
```

which succeeded. I was also able to build a C++ wheel from the C++ sdist, which passed the `ArrayBuilder` test above.

We can just merge this by skipping tests if you're happy to merge; we don't touch anything about from the deployment CI, which doesn't run in the test suite.

.github/workflows/deploy-cpp.yml
```diff
@@ -10,13 +10,18 @@ on:
 jobs:
 
   make_sdist:
-    name: "Make SDist"
+    name: Make SDist
```
jpivarski, line 13: I started indiscriminately quoting strings in YAML after being burned by some cases that looked like strings but weren't, due to some special character. It never _hurts_ to quote. I don't think it's harder to read with quotes.

This one can be unquoted, of course.

.github/workflows/deploy-cpp.yml
```diff
@@ -64,10 +69,21 @@ jobs:
       with:
         submodules: true
 
+    - name: Python 3.10
+      uses: actions/setup-python@v4
+      with:
+        python-version: '3.10'
+
     - name: Compute SOURCE_DATE_EPOCH
+      shell: python
```
jpivarski, line 78: That's new to me. I guess you opted for Python as a shell because to use `strip()`. There's probably a shell equivalent, but why bother? `:)`

jpivarski (approved):
I approve!

.github/workflows/deploy-cpp.yml
agoose77: Yes, I have the same philosophy. The Python versions *always* catch me out. I chose to make things look more uniform, but when writing new YAML I usually always quote.

.github/workflows/deploy-cpp.yml
agoose77: Actually, I'd rather not use Python! But, we run this on all OSs, and the Python implementation easily works on all platforms. We might want to make this a dev script, I'm half minded.

PR #1894, opened 2022-11-18
ci: set tokens & disable release for C++ [skip ci]

agoose77:
This PR adds an input form to enable publishing the C++ distributions to PyPI, and sets the API token.


jpivarski (approved):
Right—we'll merge it without tests because the only real test would be to try to deploy.

PR #1895, opened 2022-11-18
chore: add README.md to awkward-cpp

agoose77:
This definitely fixes the wheel (locally), but I also added tests to ensure that we check the metadata of the wheels before upload. This just ensures that if we do practice runs, we'll catch metadata issues a bit earlier.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-chore-fix-cpp-readme/> once Read the Docs has finished building :hammer:

### DIFF
--- a/awkward-cpp/README.md
+++ b/awkward-cpp/README.md
@@ -1,0 +1,3 @@
+# `awkward-cpp`
+
+`awkward-cpp` provides the `awkward` package with a suite of high-performance (compiled) kernels which facilitate computations on ragged arrays. It also implements performance-sensitive features, such as the `ak.ArrayBuilder()` class.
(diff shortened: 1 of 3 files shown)


agoose77:
@jpivarski I don't love the Python-as-shell in the workflow, but it seems like the easiest solution. Any suggestions?

awkward-cpp/README.md
```diff
@@ -0,0 +1,3 @@
+# `awkward-cpp`
+
+`awkward-cpp` provides the `awkward` package with a suite of high-performance (compiled) kernels which facilitate computations on ragged arrays. It also implements performance-sensitive features, such as the `ak.ArrayBuilder()` class.
```
jpivarski: The list of features was missing AwkwardForth and the header-only classes. Instead of making a list that could get out of date, let's just say that they probably don't want this on its own.

```suggestion
`awkward-cpp` provides precompiled routines for the `awkward` package. It is not useful on its own, only as a dependency for `awkward `.
```

jpivarski (approved):
Approved with changes. Thanks!

PR #1897, opened 2022-11-19
ci: simplify hash computation [skip ci]

agoose77:
The release workflow failed in https://github.com/scikit-hep/awkward/actions/runs/3501131090, https://github.com/scikit-hep/awkward/actions/runs/3501173113, because the endpoint URL that is used to look up the PyPI package was not properly formed; when the relative package URL was joined with the endpoint, the `pypi` component was lost. 

This PR fixes the default URL, and adds a check to make this impossible in future. It also simplifies the SOURCE_DATE_EPOCH computation.

### DIFF
--- a/.github/workflows/deploy-cpp.yml
+++ b/.github/workflows/deploy-cpp.yml
@@ -27,7 +27,7 @@ jobs:
       run: |
         import os, subprocess
         result = subprocess.run(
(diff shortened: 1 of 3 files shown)


agoose77:
I'm going to merge this without a sign-off, because I authored the code that's being fixed, and I've tested that it works locally. I haven't checked the upload step, but I assume it will work as it worked before the split. There is additional impetus to upload the non-cpp package as we have already made a GH release.

PR #1904, opened 2022-11-21
feat: add ak.drop_none()

ioanaif:
This PR adds the `drop_none` functionality.  Requested in #832 

`ak.drop_none(array, axis)` - removes missing values (None) from a given array.

For example, in the following array,

`a = ak.Array([[[0]], [[None]], [[1], None], [[2, None]]])`

The None values will be removed, resulting in

```
>>> ak.drop_none(a)
<Array [[[0]], [[]], [[1]], [[2]]] type='4 * var * var * int64'>
```

The default axis is None. However, an axis can be specified:

```
>>> ak.drop_none(a,axis=1)
<Array [[[0]], [[None]], [[1]], [[2, None]]] type='4 * var * var * ?int64'>
```


src/awkward/operations/ak_drop_none.py
```diff
@@ -0,0 +1,82 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import awkward as ak
+
+np = ak.nplikes.NumpyMetadata.instance()
+
+
+def drop_none(array, axis=None, highlevel=True, behavior=None):
```
jpivarski, line 8: I hadn't been thinking of this, but I see the argument for making `axis=None` be the default. I think you're right in assuming that most users would want to eliminate _all_ missing values, regardless of depth. Dropping missing values at a particular level of depth is more niche, but also necessary.

jpivarski (approved):
This is a good implementation and I can see that the tests are exhaustive, using the full suite of layouts.

I think it's good and ready to merge, as soon as the build-docs issue is fixed. If it's fixed in #1905 first, we'll merge that into main and then merge main into here. (One way or the other.)

src/awkward/contents/listoffsetarray.py
```diff
@@ -1975,6 +1975,35 @@ def _completely_flatten(self, nplike, options):
             flat = self._content[self._offsets[0] : self._offsets[-1]]
             return flat._completely_flatten(nplike, options)
 
+    def _drop_none(self):
+        if self._content.is_option:
```
agoose77, line 1957: I noticed that this will be skipped if the content wraps an option type, e.g.
```python
import numpy as np
import awkward as ak
layout = ak.contents.RegularArray(
    ak.contents.IndexedArray(
        ak.index.Index64(np.array([0, 1, 2], dtype=np.int64)),
        ak.contents.IndexedOptionArray(
            ak.index.Index64(np.array([-1, -1, 0], dtype=np.int64)),
            ak.contents.NumpyArray(np.arange(3)),
        ),
    ),
    3,
)
ak.drop_none(layout, axis=1)
```

agoose77 (commented):
Nice work @ioanaif!! I needed this feature *so much* during my analysis work.

I'm on holiday, but I know that this is a big feature and wanted to try and offer an additional set of eyes.

src/awkward/contents/listoffsetarray.py
jpivarski: An `IndexedOptionArray` inside of an `IndexedArray` should not appear in user arrays; these things can exist temporarily in our internal code, but before returning it to the user, `simplify_optiontype` should have turned this into a single `IndexedOptionArray`.

```
>>> layout.content
<IndexedArray len='3'>
    <index><Index dtype='int64' len='3'>[0 1 2]</Index></index>
    <content><IndexedOptionArray len='3'>
        <index><Index dtype='int64' len='3'>[-1 -1  0]</Index></index>
        <content><NumpyArray dtype='int64' len='3'>[0 1 2]</NumpyArray></content>
    </IndexedOptionArray></content>
</IndexedArray>

>>> layout.content.simplify_optiontype()
<IndexedOptionArray len='3'>
    <index><Index dtype='int64' len='3'>[-1 -1  0]</Index></index>
    <content><NumpyArray dtype='int64' len='3'>[0 1 2]</NumpyArray></content>
</IndexedOptionArray>
```

@ioanaif and I talked about this, about the implementation only needing to work for these "well formed" arrays. (We need a good name for these different levels of validity.) If we ever encounter a situation in which a user finds that `drop_none` isn't working because they have an `IndexedArray` of `IndexedOptionArray`, we'd investigate how they ended up with the `IndexedArray` of `IndexedOptionArray`, because that's already wrong.

agoose77: Ah this does actually sound familiar. Maybe we should write this up in our CONTRIBUTING.md? If in favour, I can create an issue.

I vote for "canonical representation" as the term to describe these simplified layouts? Or, just "awkward-produced arrays should always be simplified". 

How about this layout:
```python3
layout = ak.contents.RegularArray(
    ak.contents.RecordArray(
        [
            ak.contents.IndexedOptionArray(
                ak.index.Index64(np.array([-1, -1, 0], dtype=np.int64)),
                ak.contents.NumpyArray(np.arange(3)),
            )
        ],
        fields=None,
    ),
    3,
)
```

This is simplified, but the option is within the record. I assume that we don't care about what happens inside a record (i.e. we don't support it).

src/awkward/contents/content.py
```diff
@@ -1573,6 +1573,9 @@ def _completely_flatten(
     ) -> list:
         raise ak._util.error(NotImplementedError)
 
+    def drop_none(self):
+        return self._drop_none()
+
```
agoose77, line 1019: @ioanaif would you mind adding a fallback for the private method? Eventually I'll get around to more typing work and it will be useful to have it :)
```suggestion

def _drop_none(self) -> Content:
    raise ak._errors.wrap_error(NotImplementedError)
```

agoose77:
I'm done touching this PR now (hands off) - I fixed the whitespace in my suggestion that didn't merge properly :)

src/awkward/operations/ak_drop_none.py
```diff
@@ -0,0 +1,82 @@
...
+            if posaxis == depth and layout.is_option:
+                return layout.project()
+            elif posaxis == depth and layout.is_list:
+                if layout.content.is_option:
+                    return layout.drop_none()
+
+            depth_context["posaxis"] = posaxis
+
+    depth_context = {"posaxis": axis}
+    out = layout.recursively_apply(action, behavior, depth_context)
```
jpivarski: None of the `Content` subclasses have a `recursively_apply` method (without an underscore). What's going on here?

jpivarski (changes requested):
I looked at the tests too hastily: this is not wrapping its result as a high-level array, which it should by default.

There's also an error (below) that I noticed when trying the RecordArray example, which should be possible. I don't know why RecordArray reveals it but NumpyArray doesn't. I'm looking more closely at this now...

src/awkward/operations/ak_drop_none.py
jpivarski: I'm wrong: all of the `Content` subclasses have `recursively_apply` (no underscore). I don't know what I was seeing a moment ago.

```python
>>> ak.Array([1, 2, 3]).layout._recursively_apply
<bound method NumpyArray._recursively_apply of <NumpyArray dtype='int64' len='3'>[1 2 3]</NumpyArray>>

>>> ak.Array([{"x": 1}, {"x": 2}, {"x": 3}]).layout.recursively_apply
<bound method Content.recursively_apply of <RecordArray is_tuple='false' len='3'>
    <content index='0' field='x'>
        <NumpyArray dtype='int64' len='3'>[1 2 3]</NumpyArray>
    </content>
</RecordArray>>
```

ioanaif: But `Content` does, which is where this is triggered.

agoose77: @jpivarski you're not seeing things - 3fea24eb83e84f1875ef9298b657808903214bcb broke `recursively_apply` gracefully by subsuming it into the body of another function. I fixed it in e72ebaa510af6f9c3413d6a6e0a42f72b433c84f, if you pull from `origin`. Many cooks and all that!

jpivarski: I see that there's a

```python
return ak._util.wrap(out, behavior, highlevel)
```

and that `highlevel=True` is the default. So why is this happening?

```python
>>> ak.drop_none(ak.Array([1, None, 3]))
<NumpyArray dtype='int64' len='2'>[1 3]</NumpyArray>

>>> ak.drop_none(ak.Array([1, None, 3]), highlevel=True)
<NumpyArray dtype='int64' len='2'>[1 3]</NumpyArray>
```

ioanaif: Fixing it. It is because I made an assumption on NumpyArray and returned at the start of the method

jpivarski: If there's nothing to drop, it returns a high-level array, regardless of how `highlevel` is set:

```python
>>> ak.drop_none(ak.Array([1, 2, 3]))
<Array [1, 2, 3] type='3 * int64'>

>>> ak.drop_none(ak.Array([1, 2, 3]), highlevel=True)
<Array [1, 2, 3] type='3 * int64'>

>>> ak.drop_none(ak.Array([1, 2, 3]), highlevel=False)
<Array [1, 2, 3] type='3 * int64'>
```

src/awkward/contents/indexedoptionarray.py
```diff
@@ -1575,6 +1575,10 @@ def _completely_flatten(self, nplike, options):
         else:
             return [self.simplify_optiontype()]
 
+    def _drop_none(self):
+        _, _, outindex = self._nextcarry_outindex(self._nplike)
+        return (outindex, self.project())
```
jpivarski: Some implementations of `_drop_none` return a `Content`, while others return a 2-tuple of `(Index, Content)`. That's probably part of the problem.

jpivarski (changes requested):
Okay, I'm sorry, but I missed some things. The `to_list` tests in test_1904-drop-none.py are insensitive to the difference between high-level and low-level arrays, so that's why we didn't catch that issue.

I suspect that having `_drop_none` return different types for different nodes (`Content` vs `(Index, Content)`) is probably the indirect cause, though I don't know what mechanism is responsible for it. Anyway, it will be safer to have all the `_drop_none` methods return the same type; the top-level `drop_none` (no underscore) is in a good position to drop the unnecessary `outindex` at the end.

Also, this should at least pass through records: when users want to remove missing values, they'll want to remove them from within records just as much as from recordless arrays. A good guide for this could be `is_none`, which also looks inside of arrays, and `is_none` and `drop_none` should be pretty similar.

```python
>>> ak.is_none(ak.Array([[{"x": [1]}], [{"x": [None]}]]), axis=2).show()
[[{x: [False]}],
 [{x: [True]}]]

>>> ak.is_none(ak.Array([[{"x": [1], "y": [[2]]}], [{"x": [None], "y": [[None]]}]]), axis=-1).show()
[[{x: False, y: [False]}],
 [{x: False, y: [False]}]]
```

In the above, `x` and `y` have lists of different depths, but `axis=-1` counts up from the bottom. That's why most functions keep calling `wrap_axis_if_negative` until it finally becomes positive (after the recursion has passed through the record and it's seeing a single, unambiguous depth).

I think these are the only two issues, though. There are tests in test_1904-drop-none.py that involve records; I'll take a look at what happened there now.

jpivarski (changes requested):
This looks like it's almost completely done. I see that you're using the new `maybe_posaxis`.

There are some cases that the tests didn't (couldn't) cover because the tests are based on slicing with `is_none`, which isn't possible if a record introduces different depths. I've worked through some examples that could be used as direct tests.

Starting from this `array`, which has missing values at all levels, with different levels for the `x` and `y` branches of the record.

```python
array = ak.Array(
    [None,
     [None, {"x": [1], "y": [[2]]}],
     [{"x": [3], "y": [None]}, {"x": [None], "y": [[None]]}]
    ])
```

```python
>>> array.show()
[None,
 [None, {x: [1], y: [[2]]}],
 [{x: [3], y: [None]}, {x: [None], y: [[None]]}]]
```

First, with `axis=0`, the result from `is_none` and `drop_none` are **correct** because only the `None` outside of any lists is at `axis=0`.

```python
>>> ak.is_none(array, axis=0).show()
[True,
 False,
 False]
>>> ak.drop_none(array, axis=0).show()
[[None, {x: [1], y: [[2]]}],
 [{x: [3], y: [None]}, {x: [None], y: [[None]]}]]
```

Next, with `axis=1`, `is_none` and `drop_none` are still both **correct**.

```python
>>> ak.is_none(array, axis=1).show()
[None,
 [True, False],
 [False, False]]
>>> ak.drop_none(array, axis=1).show()
[None,
 [{x: [1], y: [[2]]}],
 [{x: [3], y: [None]}, {x: [None], y: [[None]]}]]
```

Next, with `axis=2`, `is_none` is correct because it is only making `x: [True]` or `y: [True]` for a `None` at this level; anything deeper is hidden inside a `False` and anything shallower is visible, so `is_none` is **correct**.

However, `drop_none` is producing the right values but putting them in the wrong places: the `x: [3]` should be in the previous record, so `drop_none` is **incorrect** in this example.

```python
>>> ak.is_none(array, axis=2).show()
[None,
 [None, {x: [False], y: [False]}],
 [{x: [False], y: [True]}, {x: [True], y: [False]}]]
>>> ak.drop_none(array, axis=2).show()
[None,
 [None, {x: [1], y: [[2]]}],
 [{x: [], y: []}, {x: [3], y: [[None]]}]]
```

Next, with `axis=-1`, the deepest `x` values have one `[`, `]` while the deepest `y` values have two `[[`, `]]`. The `is_none` function reports `x` to be `[False]` or `[True]` with one bracket, and it reports `y` to be `[[False]]` or `[[True]]` with two brackets, so `is_none` is **correct**.

The `drop_none` correctly leaves the `y: [None]` (one bracket) because it's above `axis=-1` and it keeps `x: [3]` or removes it, `x: []`, also with one bracket because that's what `axis=-1` means for `x`. However, the `x: [3]` should be on the previous record, so it's **incorrect** for this example.

```python
>>> ak.is_none(array, axis=-1).show()
[None,
 [None, {x: [False], y: [[False]]}],
 [{x: [False], y: [None]}, {x: [True], y: [[True]]}]]
>>> ak.drop_none(array, axis=-1).show()
[None,
 [None, {x: [1], y: [[2]]}],
 [{x: [], y: [None]}, {x: [3], y: [[]]}]]
```

Finally, with `axis=-2`, `is_none` should say `x: False` and (potentially) `x: True` (no instances in this `array`) with zero brackets and `y: [False]` and `y: [True]` with one bracket. It does, and `is_none` is **correct**.

Thinking about what `drop_none` should do in this case, it can't remove an `x` field without also removing a `y` field, so `axis=-2` should raise an exception for this kind of thing, and it doesn't:

```python
>>> ak.is_none(array, axis=-2).show()
[None,
 [None, {x: False, y: [False]}],
 [{x: False, y: [True]}, {x: False, y: [False]}]]
>>> ak.drop_none(array, axis=-2).show()
[None,
 [None, {x: [1], y: [[2]]}],
 [{x: [3], y: []}, {x: [], y: [[None]]}]]
```

To be concrete: the original `array` has `x: [1]`, `x: [3]`, and `x: [None]` at this `axis`, all of which are not missing, but if one was, then it would want to remove that `x` value, yet `axis=-2` for `y` does not mean for it to act at this level. You can implement this exception by adding code where the recursion gets to the RecordArray: if `posaxis == depth - 1` for some of its fields but not others, then it should raise a `np.AxisError`.

(`is_none` raises an `np.AxisError` if `axis=-3`, which is impossible for what `is_none` wants to do.)

I need a better `array2` for this case, so that I'm not talking in hypotheticals.

```python
array2 = ak.Array(
    [None,
     [None, {x: [1], y: [[2]]}],
     [{x: None, y: [None]}, {x: [None], y: [[None]]}]
    ])
```

```python
>>> array2.show()
[None,
 [None, {x: [1], y: [[2]]}],
 [{x: None, y: [None]}, {x: [None], y: [[None]]}]]
```

Now `is_none` at `axis=-2` is

```python
>>> ak.is_none(array2, axis=-2).show()
[None,
 [None, {x: False, y: [False]}],
 [{x: True, y: [True]}, {x: False, y: [False]}]]
```

which has `x: False` wherever `x` has a list and `x: True` wherever `x` has a direct `None`.

Trying to run `drop_none` on this raises a `ValueError` in the RecordArray constructor, but it should have been an `np.AxisError` before attempting to compute and construct the RecordArray.

These can be directly dropped in as new unit tests. ~~This PR should also increase the `awkward-cpp` version number to `3` in `awkward-cpp/pyproject.toml`.~~

I think the above is just one error in indexing plus needing to add one check-and-raise-exception. It's an enormous amount of work to get this far.

jpivarski:
> ~~This PR should also increase the `awkward-cpp` version number to `3` in `awkward-cpp/pyproject.toml`.~~

Actually, I'm going to stick to the habit of always changing version numbers as direct commits to `main`, so that the fact that a version has changed is highly visible. I'll do that right now.

src/awkward/operations/ak_drop_none.py
```diff
@@ -0,0 +1,109 @@
...
+            if posaxis == depth - 1 and layout.is_list and layout.content.is_option:
+                return layout.drop_none()
+
+    depth_context = {"max_branch_depth": layout.branch_depth[1]}
+    options = {}
+    out = ak._do.recursively_apply(layout, action, behavior, depth_context, options)
+    if "new_offsets" in options and axis is not None:
+        out = ak._do.recursively_apply(
+            out, recompute_offsets, behavior, depth_context, options
+        )
```
agoose77: Can we remove `depth_context` now?
```suggestion
        def recompute_offsets(layout, depth, **kwargs):
            posaxis = ak._util.maybe_posaxis(layout, axis, depth)
            if posaxis == 0 and posaxis == depth - 1 and layout.is_list:
                out = layout._rebuild_without_nones(
                    options["new_offsets"], layout.content
                )
                return out
            if posaxis == depth and layout.is_list:
                out = layout._rebuild_without_nones(
                    options["new_offsets"], layout.content
                )
                return out

        def action(layout, depth, **kwargs):
            posaxis = ak._util.maybe_posaxis(layout, axis, depth)
            if posaxis == 0:
                if not layout.is_option:
                    return layout
                else:
                    return layout.drop_none()
            if posaxis == depth - 1 and layout.is_option:
                _, _, new_offsets = layout._nextcarry_outindex(layout.backend)
                options.update(new_offsets=new_offsets)
                return layout.drop_none()
            if posaxis == depth - 1 and layout.is_list and layout.content.is_option:
                return layout.drop_none()

    options = {}
    out = ak._do.recursively_apply(layout, action, behavior, options)
    if "new_offsets" in options and axis is not None:
        out = ak._do.recursively_apply(
            out, recompute_offsets, behavior, options
        )
```

src/awkward/operations/ak_drop_none.py
```diff
@@ -0,0 +1,109 @@
...
+
+    return ak._util.wrap(out, behavior, highlevel)
```
agoose77: @jpivarski @ioanaif do we want to use `like=` here, or `behavior=ak._util.behaviorof(array, behavior=behavior)`? I can't remember!
```suggestion
    return ak._util.wrap(out, behavior, highlevel, like=behavior)
```

agoose77 (commented):
I pushed some docs fixes, but any code suggestions I've left here! Nice work @ioanaif :)

jpivarski (approved):

jpivarski (approved):
I believe this is done! (Let's see the tests pass.)

Thanks for all of the hard work on this; it was definitely a lot more involved than I had thought it was going to be when I first suggested it. But now it works in all the extreme cases and we know it's not going to come up again as a bug in somebody's analysis.

It reminds me of something... found it:



This applies a lot more often than I'd like it to.

ioanaif:
> 

Yay! Hope all corner cases have been discovered! 🥳🥳

PR #1905, opened 2022-11-21
feat: made 'very optional' arguments keyword-only

### LINKED ISSUE #1872: Some arguments should be keyword-only.

This is a reminder to sweep through and label the obviously keyword-only arguments as such. The codebase doesn't have any keyword-only arguments because it once supported Python 2, but that rationale is long gone. This change is backward-incompatible (in a small way, easy to work around), so it should be done before the release.

jpivarski:
Also, the functions that are defined in terms of reducers (`ak.ptp`, `ak.moment`, `ak.mean`, `ak.var`, `ak.std`, `ak.covar`, `ak.corr`, `ak.linear_fit`, `ak.softmax`) now pass behavior consistently. (Previously, only `ak.std` made a half-hearted attempt.)

`ak.to_categorical` had a `behavior` argument added (somehow overlooked before).

The `ak.*_like` functions had their arguments reordered so that `dtype` is earlier than `highlevel` or `behavior` (which is the freedom to fix things that this PR is intended to make permanent).

In general, any `ak.*` function arguments that are "very optional"—configuring unusual situations—are now keyword-only, especially if they're booleans. It's very easy to mix up the order of boolean arguments. But `axis` arguments were left positional, since there are instances of positional `axis` in the wild.

The `Content` and `Form` subclasses, however, were left as they are. They're part of the low-level public API, and have only two arguments that are at all optional: `parameters` and `nplike` or `form_key`. These are all different types and unlikely to get misordered, and very unlikely to change order in the future. So there's not a strong need to make these arguments keyword-only.

The `Type` and `Index` subclasses are also low-level public API. `Type`'s `typestr` argument and `Index`'s `metadata` and `nplike` deserve to be keyword-only because they are rarely used and could be confusing. Any downstream developers who _do_ use them should have a good understanding of what they are and do, at least at the level of spelling out their names.

That's everything that's going on in this PR!

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-made-very-optional-arguments-keyword-only/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/behaviors/mixins.py
+++ b/src/awkward/behaviors/mixins.py
@@ -63,7 +63,7 @@ def mixin_class(registry, name=None):
     return register
 
 
-def mixin_class_method(ufunc, rhs=None, transpose=True):
+def mixin_class_method(ufunc, rhs=None, *, transpose=True):
     """
     Args:
         ufunc (numpy.ufunc): A universal function (or NEP18 callable) that is
--- a/src/awkward/operations/ak_from_iter.py
+++ b/src/awkward/operations/ak_from_iter.py
@@ -8,18 +8,24 @@ np = ak.nplikes.NumpyMetadata.instance()
 
 
 def from_iter(
-    iterable, highlevel=True, behavior=None, allow_record=True, initial=1024, resize=1.5
+    iterable,
+    *,
+    allow_record=True,
+    highlevel=True,
+    behavior=None,
+    initial=1024,
+    resize=1.5
 ):
     """
     Args:
         iterable (Python iterable): Data to convert into an Awkward Array.
+        allow_record (bool): If True, the outermost element may be a record
+            (returning #ak.Record or #ak.record.Record type, depending on
+            `highlevel`); if False, the outermost element must be an array.
         highlevel (bool): If True, return an #ak.Array; otherwise, return
             a low-level #ak.contents.Content subclass.
         behavior (None or dict): Custom #ak.behavior for the output array, if
             high-level.
-        allow_record (bool): If True, the outermost element may be a record
-            (returning #ak.Record or #ak.record.Record type, depending on
-            `highlevel`); if False, the outermost element must be an array.
         initial (int): Initial size (in bytes) of buffers used by the
             [ak::ArrayBuilder](_static/classawkward_1_1ArrayBuilder.html).
         resize (float): Resize multiplier for buffers used by the
@@ -56,9 +62,9 @@ def from_iter(
         "ak.from_iter",
         dict(
             iterable=iterable,
+            allow_record=allow_record,
             highlevel=highlevel,
             behavior=behavior,
-            allow_record=allow_record,
             initial=initial,
             resize=resize,
         ),
--- a/docs/user-guide/how-to-create-arraybuilder.md
+++ b/docs/user-guide/how-to-create-arraybuilder.md
@@ -376,7 +376,7 @@ new_array
 new_array.layout
 ```
 
-Above, we see that `new_array` is just making references ({class}`ak.layout.IndexedArray`) of an {classY}`ak.layout.RecordArray` with `x = [1, 2, 3]` and `y = [1.1, 2.2, 3.3]`.
+Above, we see that `new_array` is just making references ({class}`ak.layout.IndexedArray`) of an {class}`ak.layout.RecordArray` with `x = [1, 2, 3]` and `y = [1.1, 2.2, 3.3]`.
 
 +++
 
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -412,7 +412,7 @@ def behavior_of(*arrays, **kwargs):
     highs = (
         ak.highlevel.Array,
         ak.highlevel.Record,
-        #        ak.highlevel.ArrayBuilder,
+        # ak.highlevel.ArrayBuilder,
     )
     for x in arrays[::-1]:
         if isinstance(x, highs) and x.behavior is not None:
@@ -429,7 +429,6 @@ def behavior_of(*arrays, **kwargs):
     return behavior
 
 
-# maybe_wrap and maybe_wrap_like go here
 def wrap(content, behavior=None, highlevel=True, like=None):
     assert content is None or isinstance(
         content, (ak.contents.Content, ak.record.Record)
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -1364,7 +1364,9 @@ class ListArray(Content):
         )
 
     def _to_numpy(self, allow_missing):
-        return ak.operations.to_numpy(self.toRegularArray(), allow_missing)
+        return ak.operations.to_numpy(
+            self.toRegularArray(), allow_missing=allow_missing
+        )
 
     def _completely_flatten(self, nplike, options):
         if (
--- a/src/awkward/contents/listoffsetarray.py
+++ b/src/awkward/contents/listoffsetarray.py
@@ -1963,7 +1963,9 @@ class ListOffsetArray(Content):
         if array_param in {"bytestring", "string"}:
             return self._nplike.array(self.to_list())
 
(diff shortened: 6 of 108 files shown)


src/awkward/behaviors/mixins.py
```diff
@@ -63,7 +63,7 @@ def register(cls):
     return register
 
 
-def mixin_class_method(ufunc, rhs=None, transpose=True):
+def mixin_class_method(ufunc, rhs=None, *, transpose=True):
```
jpivarski, line 66: This is one that I should ask @nsmith- about: is it the case that `rhs` is likely to be positional but `transpose` should always be written out as keyword-only?

src/awkward/operations/ak_from_iter.py
```diff
@@ -8,18 +8,24 @@
...
 def from_iter(
-    iterable, highlevel=True, behavior=None, allow_record=True, initial=1024, resize=1.5
+    iterable,
+    *,
+    allow_record=True,
+    highlevel=True,
+    behavior=None,
+    initial=1024,
+    resize=1.5
 ):
```
ioanaif, line 18: This is the only operation where `highlevel` and `behavior` are not the last parameters sent. Does it make sense to re-order them to have the same pattern across all operations?

src/awkward/behaviors/mixins.py
nsmith-: Correct. Any binary method would need the `rhs` specified but `transpose` is more rare.
As an aside, in retrospect, I should have made `transpose` default false but it is too late. That was the cause of https://github.com/CoffeaTeam/coffea/issues/674

src/awkward/operations/ak_from_iter.py
jpivarski: I know; I thought about that. If we had a rule like "highlevel and behavior are always last" and also a rule like "extremely internal arguments are always last," then one rule has to override the other when a function has both.

The `initial` and `resize` arguments determine how memory is allocated by `GrowableBuffer`, and I think the new `GrowableBuffer` doesn't even use `resize`. The sizes of panels in `GrowableBuffer` after the first one (which has `initial` bytes) hasn't been entirely settled. Since it's not copying into new contiguous buffers anymore, the exponential growth of new buffers is no longer needed, but making every panel the same size is probably not optimal, either, because there's probably two populations of `GrowableBuffers`: those that never exceed their first panel and those that have many panels. There should probably be two parameters: `initial` and `subsequent`, so that the first panel can be small (for the ones that don't get filled much) and subsequent panels can be large (to minimize overhead for the ones that get filled a lot), but linearly: without exponential growth.

I think it's very unlikely that any user would want to configure such details, and they might change in the future. (It will be easy to do it in a deprecation cycle: use of `resize` will raise a warning until some version, after which it gets removed.) So that's why I thought these options should be more hidden than `highlevel` and `behavior`.

But the good news is that we can now change that order at will, even after the 2.0.0 release, because users cannot depend on their order anymore.

src/awkward/behaviors/mixins.py
jpivarski: Okay, good to know about `rhs`.

About `transpose`: changing the default from `True` to `False` would be dangerous, even at the 2.0.0 boundary, unless we also change the name. As a keyword-only argument, the written-out old name would be different from the written-out new name, so old code can't pass through accidentally. After the 2.0.0 boundary, we'd have to have a period of time in which both names exist, with a warning on the old name, but we could make a clean switch to a new name if we do it before the 2.0.0 boundary.

What do you think? Is there a better word than `transpose` that would justify the change in default?

ioanaif (approved):
LGTM!

jpivarski:
It's weird that build-docs failed with

```
/home/runner/work/awkward/awkward/docs/user-guide/how-to-convert-buffers.md:54: ERROR: Unknown interpreted text role "method".
/home/runner/work/awkward/awkward/docs/user-guide/how-to-create-arraybuilder.md:379: ERROR: Unknown interpreted text role "classY".
```

> LGTM!

Okay, great! I'm looking at the build-docs error, and if I resolve that, then I'll merge this PR. My question to @nsmith- can be followed up in a new PR, if need be.

src/awkward/behaviors/mixins.py
nsmith-: I think this will just have to stay. It is not a well-trodden path anyway.

jpivarski:
There are more errors than that:

```
/home/runner/work/awkward/awkward/docs/reference/generated/ak.from_json.rst:24: ERROR: Unexpected indentation.
/home/runner/work/awkward/awkward/docs/reference/generated/ak.from_json.rst:143: ERROR: Unexpected indentation.
/home/runner/work/awkward/awkward/docs/reference/generated/ak.from_json.rst:185: ERROR: Unexpected indentation.
/home/runner/work/awkward/awkward/docs/reference/generated/ak.to_json.rst:13: ERROR: Unexpected indentation.
/home/runner/work/awkward/awkward/docs/reference/generated/ak.to_json.rst:25: ERROR: Unexpected indentation.
/home/runner/work/awkward/awkward/docs/user-guide/how-to-convert-buffers.md:54: ERROR: Unknown interpreted text role "method".
/home/runner/work/awkward/awkward/docs/user-guide/how-to-create-arraybuilder.md:379: ERROR: Unknown interpreted text role "classY".
```

The ones I've been able to trace back do seem to be legitimate errors, but I don't know why they'd stop the build now and not previously. I'll see what happens when I fix the ones I can fix.

ioanaif:
> There are more errors than that:
> 
> ```
> /home/runner/work/awkward/awkward/docs/reference/generated/ak.from_json.rst:24: ERROR: Unexpected indentation.
> /home/runner/work/awkward/awkward/docs/reference/generated/ak.from_json.rst:143: ERROR: Unexpected indentation.
> /home/runner/work/awkward/awkward/docs/reference/generated/ak.from_json.rst:185: ERROR: Unexpected indentation.
> /home/runner/work/awkward/awkward/docs/reference/generated/ak.to_json.rst:13: ERROR: Unexpected indentation.
> /home/runner/work/awkward/awkward/docs/reference/generated/ak.to_json.rst:25: ERROR: Unexpected indentation.
> /home/runner/work/awkward/awkward/docs/user-guide/how-to-convert-buffers.md:54: ERROR: Unknown interpreted text role "method".
> /home/runner/work/awkward/awkward/docs/user-guide/how-to-create-arraybuilder.md:379: ERROR: Unknown interpreted text role "classY".
> ```
> 
> The ones I've been able to trace back do seem to be legitimate errors, but I don't know why they'd stop the build now and not previously. I'll see what happens when I fix the ones I can fix.

#1904 triggers the same errors, I haven't found the cause yet.

jpivarski:
The only one that I know how to fix is

```
ERROR: Unknown interpreted text role "classY".
```

It's pretty clear that this should be "class". I'll put that fix in, but it should still fail because of those other errors.

agoose77:
The error here is actually due to `LayoutBuilder` failing in the LayoutBuilder notebook. I can take a quick look. This is actually something we should address; errors in notebooks aren't very visible in the tracebacks.

Aside - fixing the docs warnings is a long-term todo ...

agoose77:
See https://github.com/scikit-hep/awkward/pull/1907

jpivarski:
It's strange that #1904 now passes but #1905 still fails. Was 8d6abd6ee5f11ade295ec899ab934c89b87dd124 wrong? It doesn't look like that's what the problem is.

agoose77:
@jpivarski in the `docs` action log, [I can see mention](https://github.com/scikit-hep/awkward/actions/runs/3525540904/jobs/5912511860#step:13:538) of a different notebook this time. I'll take a look.

agoose77:
@jpivarski looks like this was a "good" failure - apparently our test suite doesn't check `ak.Array.mask`! Eventually mypy will catch these kinds of bugs, but it will take time to get there.

jpivarski:
To avoid letting this get stale, I'm going to merge it now. (After a merge-with-main update and re-test.)

PR #1907, opened 2022-11-22
docs: fix broken build

agoose77:
This PR:
- cleans up how we find our headers - they're now available in the project root
- exposes the tracebacks of broken builds (hopefully)
- fixes the broken dependency of ipykernel that's being pulled from conda-forge.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-header-only-docs/> once Read the Docs has finished building :hammer:

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -12,7 +12,6 @@ build:
       - pip install requests pyyaml
       # Generate required code / metadata
       - python3 dev/generate-kernel-docs.py
-      - python3 dev/copy-cpp-headers.py
       # Download built artefacts
       - python3 dev/download-github-artefact.py scikit-hep/awkward-1.0 "^jupyter-cache$" -d docs/_build/.jupyter_cache
       - python3 dev/download-github-artefact.py scikit-hep/awkward-1.0 "^awkward-cpp-wasm$" -d docs/lite/pypi
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -121,6 +121,8 @@ nb_ipywidgets_js = {
         "crossorigin": "anonymous",
     },
 }
+nb_execution_show_tb = True
+    
 # Additional stuff
 master_doc = "index"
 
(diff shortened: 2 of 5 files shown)


agoose77:
Ah, this isn't the only fix, though this is still a good change. I suspect ROOT depends upon an older version of metakernel, which we might need to pin.

agoose77:
It looks like `ipykernel` is the culprit; the latest release was yanked. I don't think this has been yanked from conda-forge, which is why we're still running into this. I'm not clear on what the conda-forge yanking mechanism is, or indeed how conda is ignoring the "yanked" status.  If conda-forge does support yanking, then this might also just be a caching issue; if `main` ran the `docs` workflow between `6.18.0` being released and subsequently yanked, then this branch would inherit that cache and fail. 

In any case, the quickest solution will be to reject 6.18.0. I think this might be better done in the conda-forge dependencies of environment.yml.cog, but let's see what happens with requirements.txt first.

agoose77:
The `pip` change worked, let me try conda now (as the preferred solution).

agoose77:
OK, looks like doing this via the conda dependencies also fixes it.

agoose77:
Ah darn, I'd forgotten that `skip-ci` just disables the CI tests rather than keeping the previous results. You can see that [e1eddd4](https://github.com/scikit-hep/awkward/pull/1907/commits/e1eddd40060345c1dab7ee98b2116e6f1cb1a6b4) passed the tests, and the subsequent commit only touches `.readthedocs.yml`, so it's safe to merge.

jpivarski (approved):
If this fixes it, then that's great! We'll merge this into main and then into #1904 and #1905.

It looks like you didn't need the "`{classY}`" → "`{class}`", but I'm pretty sure that is an improvement, so it will piggyback on #1905.

Once the currently running tests finish, you or I (whoever is first) can press the red button to merge without all tests, since the code tests won't run on a docs-only PR.

jpivarski:
All the tests that are going to run were successful. I'll merge it now.

PR #1908, opened 2022-11-22
test: add @ioana's tests of to_arraylib.

### LINKED ISSUE #1765: v2 implementation of `to_arraylib` is missing cases from the original v1 `to_numpy`
### Version of Awkward Array

HEAD

Lost in translation (and evidence that we need more tests).

v2 `to_arraylib`:

https://github.com/scikit-hep/awkward/blob/3401e2bf53b2c53b877c9d29a5eb0b36e7e01582/src/awkward/_util.py#L753-L806

v1 `to_numpy`:

https://github.com/scikit-hep/awkward/blob/f2092ee90e78c9ce00c54b3f5d9612bd218336ee/src/awkward/operations/convert.py#L208-L370

The libraries other than NumPy won't be able to use some of those other layout node types, but they general implementation should have them anyway. At least, when this is used by `to_numpy`, they need to be there.

jpivarski:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-add-ioanas-test-of-to_arraylib/> once Read the Docs has finished building :hammer:

### DIFF
--- a/tests/test_1765-add-ioanas-test-of-to_arraylib.py
+++ b/tests/test_1765-add-ioanas-test-of-to_arraylib.py
@@ -1,0 +1,19 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import numpy as np  # noqa: F401
+import pytest  # noqa: F401
+
+import awkward as ak  # noqa: F401
+
+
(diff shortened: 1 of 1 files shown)


ioanaif (approved):
LGTM!

PR #1911, opened 2022-11-22
fix: refactor '_nextcarry-outindex' to have the same signature everywhere

ioanaif:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ioanaif-refactor-nextcarry-outindex/> once Read the Docs has finished building :hammer:


jpivarski (approved):
I see; `ByteMaskedArray._nextcarry_outindex` just didn't need it before. But it's better to be consistent, even if superfluous.

Go ahead and merge, so that you can use this in #1904!

PR #1912, opened 2022-11-23
fix: ignore .nox

ianna:
I think, it should be ignored

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-git-ignore-nox/> once Read the Docs has finished building :hammer:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -99,6 +99,7 @@ pip-delete-this-directory.txt
 
 # Unit test / coverage reports
 htmlcov/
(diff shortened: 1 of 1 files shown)


agoose77:
Ah yes. This is a problem with having a global gitignore configuration - I don't always notice what isn't gitignored by *this* repository. Please feel free to add anything else that must be ignored!

agoose77 (approved):

PR #1913, opened 2022-11-24
chore: fix NumPy bound & noxfile installs

agoose77:
The NumPy lower bound for Awkward increased when we moved to `python_version>=3.7`. We made this fix in the CI, but not in the packaging metadata.

We also didn't test the `tests` or `coverage` Nox sessions, as they don't work with `scikit-build-core` out of the box. I'll follow up with @henryiii, but in the mean time we can at least ensure that they try to install the right package :facepalm: 

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-pytest/> once Read the Docs has finished building :hammer:

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -23,7 +23,7 @@ def tests(session):
     """
     Run the unit and regular tests.
     """
-    session.install("-r", "requirements-test.txt", "./awkward-cpp", "./awkward")
+    session.install("-r", "requirements-test.txt", "./awkward-cpp", ".")
     session.run("pytest", *session.posargs if session.posargs else ["tests"])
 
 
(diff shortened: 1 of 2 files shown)


noxfile.py
```diff
@@ -23,7 +23,7 @@ def tests(session):
     """
     Run the unit and regular tests.
     """
-    session.install("-r", "requirements-test.txt", "./awkward-cpp", "./awkward")
+    session.install("-r", "requirements-test.txt", "./awkward-cpp", ".")
```
henryiii, line 26: This needs to be in two lines, I think. Same below.

```suggestion
    session.install("-r", "requirements-test.txt", "./awkward-cpp")
    session.install(".")
```

Also, I think if you reuse sessions it won't rebuild if the version doesn't change. I think.

ioanaif (approved):

PR #1915, opened 2022-11-27
chore: run `yesqa`

agoose77:
This PR:
- runs the `yesqa` tool to remove unnecessary `#noqa` comments. It doesn't seem to be perfect, so I've not added it to pre-commit in the end.
- fixes the flake8 configuration that was erroneously ignoring some `src` files (my bad!)
- catches some instances of exceptions not being raised

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-chore-yesqa/> once Read the Docs has finished building :hammer:

### DIFF
--- a/dev/flake8_awkward.py
+++ b/dev/flake8_awkward.py
@@ -11,7 +11,7 @@ class Flake8ASTErrorInfo(NamedTuple):
 
 
 class Visitor(ast.NodeVisitor):
-    msg = "AK101 exception must be wrapped in ak._util.*error"
+    msg = "AK101 exception must be wrapped in ak._errors.*_error"
 
     def __init__(self):
         self.errors: list[Flake8ASTErrorInfo] = []
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -130,19 +130,19 @@ class Content:
         self,
         getkey: Callable[[Content], form.Form],
     ) -> form.Form:
-        raise ak._util.error(NotImplementedError)
+        raise ak._errors.wrap_error(NotImplementedError)
 
     @property
     def Form(self) -> type[form.Form]:
-        raise ak._util.error(NotImplementedError)
+        raise ak._errors.wrap_error(NotImplementedError)
 
     @property
     def typetracer(self) -> Self:
-        raise ak._util.error(NotImplementedError)
+        raise ak._errors.wrap_error(NotImplementedError)
 
     @property
     def length(self) -> int:
-        raise ak._util.error(NotImplementedError)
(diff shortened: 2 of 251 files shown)


dev/flake8_awkward.py
```diff
@@ -11,7 +11,7 @@ class Flake8ASTErrorInfo(NamedTuple):
 
 
 class Visitor(ast.NodeVisitor):
-    msg = "AK101 exception must be wrapped in ak._util.*error"
+    msg = "AK101 exception must be wrapped in ak._errors.*_error"
```
jpivarski, line 14: Noted: this PR also fixes documentation.

src/awkward/contents/content.py
```diff
@@ -130,19 +130,19 @@ def _form_with_key(
         self,
         getkey: Callable[[Content], form.Form],
     ) -> form.Form:
-        raise ak._util.error(NotImplementedError)
+        raise ak._errors.wrap_error(NotImplementedError)
```
jpivarski, line 133: Oops, it's actually more than documentation: there is no `error` function in _util.py. All of these would have been "errors in the error" message handling, likely confusing to the user and untested because we don't test all the error cases (out of exhaustion).

So this PR is much more than just clean-up; these are real bug-fixes.

jpivarski (approved):
I had thought that flake8 was checking for unnecessary noqa's—at least, I thought we encountered "please remove unnecessary noqa" errors before, or that the unnecessary noqa's were automatically removed (which led to problems in which one version required a particular noqa and another removed it). But this PR makes it clear that a lot more could have been removed. (Maybe that old feature of removing noqa's by default went away?)

Let's merge this! I'm also going to see if there are any PRs open for flake8 version 6.0. I hope we don't have to hold it back due to its API changes. Ideally, it will just work; second-best would be being able to make it work without much effort.

agoose77:
Yes this PR subsumed another as we needed to touch the same files. 

Initially I wanted to make `yesqa` part of our CI/pre-commit checks, but unfortunately, it seems to have some false positives that remove needed checks. At this stage, I'm not motivated enough to find out why that is over just manually applying the corrections periodically.

jpivarski:
It's a definite improvement!

Following up on

> I'm also going to see if there are any PRs open for flake8 version 6.0.

I just tested it locally—verified that I really did change flake8 from 5.x to 6.0, and that `pre-commit run -a` still runs with the new version. So we don't have any bad surprises coming due to the flake8 API changes.

However, I'm not going to set up a PR for it—I'm going to let auto-update do its job. We have enough things to do and automation for that already exists. We just know now that it won't cause extra work and can ignore it.

PR #1918, opened 2022-11-28
feat: split nplike into two nplikes

agoose77:
Our `nplike` mechanism encodes two different things; a NumPy abstraction, and a higher-level index-content abstraction through `index_nplike`. This is problematic for several reasons:
- accidental confusion of nplikes is easier, as it's hard to tell whether a function uses `nplike` or `nplike.index_nplike`
- index nplikes shouldn't also have `index_nplike`

Additionally, we don't need to pass around `nplike` everywhere now that we know we can only have a single nplike family per layout. This does have consequences for our JAX integration, which cheats and stores the *wrong* nplike in the layout. We need to tackle this anyway.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -118,9 +118,7 @@ def all_same_offsets(nplike, inputs):
                 return False
 
             elif offsets is None:
-                offsets = index_nplike.empty(
-                    starts.shape[0] + 1, dtype=starts.dtype
-                )
+                offsets = index_nplike.empty(starts.shape[0] + 1, dtype=starts.dtype)
                 if offsets.shape[0] == 1:
                     offsets[0] = 0
                 else:
@@ -600,9 +598,7 @@ def apply_step(
                 )
                 index = Index64(index)
                 if any(not isinstance(x, optiontypes) for x in inputs):
-                    nextindex = index_nplike.arange(
-                        mask.shape[0], dtype=np.int64
-                    )
+                    nextindex = index_nplike.arange(mask.shape[0], dtype=np.int64)
                     nextindex[mask] = -1
                     nextindex = Index64(nextindex)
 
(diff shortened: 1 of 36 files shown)


jpivarski:
`nplike` isn't an argument of the `Content` subclasses anymore, is it? 

I had been thinking of making it keyword-only, also, but decided to not touch the `Content` subclass constructors. Removing `nplike` entirely makes that point moot.

agoose77:
In this PR, it is a per-content thing; the leaf nodes need it, as well as any nodes that take an optional number of children (records, unions). But the list, indexed, and option types don't have this parameter any more in this PR

jpivarski:
If we're ever tempted to add more arguments to a `Content` subclass—which I find very unlikely at this stage, but in principle—we'd want to add them before `parameters`. Making `parameters` or `parameters` and `nplike` keyword-only would be somewhat disruptive.

Let me check to see what Vector assumes.

agoose77:
This PR would disrupt any existing usages of `nplike` for non-leaf / contentless nodes. On a purely ergonomic point, I'd support keyword-only for the default parameters of our content nodes. For the record, we can use keyword-only for non-default parameters, but most of the time one doesn't want to do this.

jpivarski:
Nope: Vector, Coffea, dask-awkward, and awkward-pandas don't come anywhere near using all the arguments of `Content` subclasses, and when Uproot does, it always spells out the `parameters` argument as a keyword.

It looks like it would be safe to make everything from `parameters` onward into a keyword-only argument for the `Content` subclasses. There might be quite a few constructor calls to fix in the Awkward codebase, but I don't see a reason to worry about known dependent libraries.

The same applies to Forms. The Types already have keyword-only `*`s, but they're before `typestr`; they would have to move before `parameters` for consistency.

That might be a big undertaking, and therefore a separate PR. Is it too intertwined with your `nplike`-argument removal?

agoose77:
@jpivarski I split that into a PR that should merge first; that way we get it merged ASAP.

PR #1919, opened 2022-11-28
feat: rename `toXXX` methods

agoose77:
This PR tackles a small naming inconsistency. Because these are public functions, we should decide now whether to merge this!

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-refactor-rename-to-xxx/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -346,7 +346,7 @@ def apply_step(
         isinstance(x, NumpyArray) and x.data.ndim != 1 for x in inputs
     ):
         inputs = [
-            x.toRegularArray() if isinstance(x, NumpyArray) else x for x in inputs
+            x.to_RegularArray() if isinstance(x, NumpyArray) else x for x in inputs
         ]
 
     # Rare that any function would want this, but some do.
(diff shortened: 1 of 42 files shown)


agoose77:
@jpivarski I feel like we touched on this before but I don't recall the solution and can't find the discussion :/ If you're against this, close it, if in favour, merge! :)

jpivarski (approved):
I remember it, too. I didn't like your earlier suggestions that made the whole method names lowercase (they should reflect the names of the classes they instantiate), but I was in favor of the underscore after "to," which this PR applies.

You're right that this is our last chance to fix it because it's public API. In fact, it's public API that Uproot uses, so we'll also need to coordinate releases. It's not used by dask-awkward, awkward-pandas, or Vector, which would be harder to coordinate with.

  1. I'll make a PR on Uproot that uses the new `to_` names, but I will not merge it. You'll be the reviewer of that PR.
  2. We'll make another pre-release of Awkward this week. (Probably more than one.)
  3. Then when Uproot breaks, I'll merge that Uproot PR to unbreak it.
  4. There should probably be another pre-release of Uproot, to give people something to test.

agoose77:
I ran another test with `rg "to[A-Z]" src` and found a missing case.

PR #1922, opened 2022-11-29
refactor: introduce `backend` as a higher abstraction than `nplike`

agoose77:
This supercedes #1918 as a means of hardening the nplike mechanism. It:

- [x] Changes initialisers of non-leaf `Content` classes to remove `nplike` argument[^caveat].
- [x] Introduces a "backend" object that holds the kernels, nplike, and index nplike. This replaces the `nplike` object that we pass around in most cases, although we might wish to keep that if there are compatibility needs.
- [x] Imposes constraint that layout index-types must match content backend.
- [x] Reverts previous attempt to make JAX leak-proof (see comments)

## Changes to Development Practices

With this PR, most of the time we want to work with a `backend` object. A backend object can be taken from an existing array with 
```python

backend = ak._backends.backend_of(*arrays)
```

Unlike `ak._nplikes.nplike_of`, `ak._backends.backend_of` will throw an exception if it can't detect a single array-like object amongst the arguments. This is to avoid defaulting to NumPy when we do not want it to. The `default` argument can be provided to request a default, e.g.
```python

backend = ak._backends.backend_of(
	*arrays, 
	default=ak._backends.NumpyBackend.instance()
)
```

This `backend` object has `nplike` and `index_nplike` attributes, which replace `nplike` and `nplike.index_nplike`.

:warning: you must now pass the _`index_nplike`_ to `Index`, i.e.
```python3
index = ak.index.Index64(..., nplike=self._backend.index_nplike)
```

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-refactor-introduce-backend/> once Read the Docs has finished building :hammer:

[^caveat]: Except for layouts that might need to pick one.


agoose77:
The big thorn in this PR is that by making the nplike-passing mechanism more robust (and ultimately replacing it with `backend`), we are not as easily able to trick Awkward into mixing JAX and NumPy nplikes/backends. We currently do this to replace JAX tracers with NumPy buffers so that subsequent Awkward operations don't trigger JAX's leak detection. Anything that performs a slice on a `NumpyArray` that holds a JAX tracer will trigger the leak detector if the user has it enabled. 

I can't remember whether these are *true* leaks in the sense that they persist beyond the lifetime of the JAX processing. I don't believe they do; only the `AuxData` should hold the layout, and the layout should go out of scope. Nevertheless, in future I'd like for the leak checker to work. We're planning on revisiting the JAX integration anyway, so I'm happy for JAX to work without leak checking in the near term.

NB, we could add a hack into the `NumpyArray` node that let's us store an `ndarray` inside a JAX node, but I think that's worse than just not supporting leak checking.

agoose77:
I'm moving the code over to pass a `backend` around in favour of `nplike`. The reason is that `nplike` should really only deal with providing an array interface. The kernel logic and index-nplike relationship is now defined by the backend. It might be useful for layouts to unpack the nplikes from the backend into private attributes, so that the code is less noisy (`self._nplike` vs `self._backend.nplike`), but I'd prefer to avoid more state to keep track of, so we'll see how this evolves.

src/awkward/_backends.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __getitem__(self, index: KernelKeyType) -> CupyKernel | NumpyKernel:
+        from awkward._connect import cuda
+
+        cupy = cuda.import_cupy("Awkward Arrays with CUDA")
+        _cuda_kernels = cuda.initialize_cuda_kernels(cupy)
+        func = _cuda_kernels[index]
+        if func is not None:
+            return CupyKernel(func, index)
+        return NumpyKernel(awkward_cpp.cpu_kernels.kernel[index], index)
```
agoose77: @jpivarski do you know why we have this fallback? I imagine it would be proper just to fail here for missing kernels.

jpivarski (approved):
We talked about this extensively on Zoom today. As promised, I did scan over all of the diffs. Nothing jumped out, though I can't say that nothing slipped through, considering that this is 3 kLoC of changes. It is, however, a refactoring, and we do have a lot of tests.

src/awkward/_backends.py
jpivarski: Let's remove it. It hides cases in which the CuPy kernel doesn't exist, making it look like something works in the `"cuda"` backend when really, it's just copying the array to the CPU.

Thanks for finding it.

```suggestion
        else:
            raise ak._errors.wrap_error(AssertionError(f"CuPyKernel not found: {index!r}"))
```

ianna:
@agoose77 - sorry, I'm a bit late to comment. Is this `backend` hidden from the users? Would it be confusing to keep a `backend` as follows (in `tests-cuda`):
```python
one = ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]], backend="cuda")
```

agoose77:
No worries @ianna, you were AFK IIRC. 

This PR introduces the `ak._backends.Backend` protocol as the "proper" counterpart to the high level strings that users can see. So `backend="cuda"` means getting an `ak._backends.Cupy` object (though perhaps we should rename the `CupyBackend` to `CudaBackend`). 

High-level users should _never_  see the `Backend` objects. Whereas, mid-level developers can access `ak.contents.Content.backend`. We've decided that users _can_ find this object if they poke around, but they should treat it like a "token" - i.e. not touch any of the attributes.

PR #1930, opened 2022-11-30
test: cover the failure modes of unflatten

agoose77:
`ak.unflatten` throws a secondary error if a 2D ragged array is passed for `counts`. This PR fixes that by catching the `else` case, and adds tests to try out the various improper `count` values.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-ndim-unflatten/> once Read the Docs has finished building :hammer:


jpivarski (approved):
That's right; `counts` can only be integers or option-of-integers. Anything else should raise an error.

It's ready to merge!

PR #1938, opened 2022-12-02
feat: add rich IPython display hook to array/record

agoose77:
This PR adds a basic `_ipython_display_` hook to the highlevel `Array` and `Record` types. This hook simply prints the type and contents of the array.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-feat-add-rich-display/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -2,6 +2,7 @@
 
 import copy
 import io
+import itertools
 import keyword
 import re
 import sys
@@ -1244,6 +1245,21 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         else:
             stream.write(out + "\n")
 
+    def _ipython_display_(self):
+        value_buff = io.StringIO()
+        self.show(type=False, stream=value_buff)
+        header_lines = value_buff.getvalue().splitlines()
+
+        type_buff = io.StringIO()
+        self.type.show(stream=type_buff)
+        footer_lines = type_buff.getvalue().splitlines()
+
+        if header_lines[-1] == "":
+            del header_lines[-1]
+
+        n_cols = max(len(line) for line in itertools.chain(header_lines, footer_lines))
+        print("\n".join([*header_lines, "-" * n_cols, *footer_lines]))  # noqa: T201
+
     def __array__(self, *args, **kwargs):
         """
         Intercepts attempts to convert this Array into a NumPy array and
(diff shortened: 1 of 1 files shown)


agoose77:
Here's a preview page: https://awkward-array.readthedocs.io/en/agoose77-feat-add-rich-display/user-guide/how-to-restructure-flatten.html

jpivarski (approved):
I didn't check to see that this works, but I approve!

By the way, shouldn't it return a string, rather than printing it? I'd be surprised if `_ipython_display_` actually works that way, since standard output is harder to capture than a Python return value.

agoose77:
> By the way, shouldn't it return a string, rather than printing it? I'd

That's how I can tell you checked the diff ;) You would imagine so, wouldn't you?

It's actually deliberate; `_ipython_display_` is an escape hatch that permits objects to invoke the `display` machinery, hence it relies on side-effects.

agoose77:
@jpivarski here are two examples in the docs:

image

You can see that 1D arrays are expressed as columns. I think this is an acceptable trade-off for showing the type information, which is usually what I'm most interested in when looking at arrays in the notebook. Could you confirm that you're happy with this before I merge?

jpivarski:
(For the sake of people following this on GitHub, we talked about this at length on Slack.)

It's true that the one-dimensional arrays will now be columns, and that wastes a bit of space. (It's not that different from what pyarrow does, though.)

I'm on board with this change, though I think it would be a little better if you inject "`type: `" at the beginning of the first line of the type. The values and types of ak.Records are going to look very similar—it won't be immediately obvious which is the value and which is the type. I don't think that a corresponding "`value: `" is necessary, though.

agoose77:
I had to change tack here; the IPython docs were wrong. I've moved to use `_repr_mimebundle_`, and added two MIME keys: `text/plain` and `text/html`. The latter is ignored in a conventional console, and the `text/plain` field is the normal repr. I preferred this approach to begin with, but I didn't want to craft HTML. However, crafting HTML _is_ the clean way to solve this, so it's the solution!

jpivarski (approved):
I see that there is now a "`type: `" string prepended before the type, and that you're using an interface that returns structured data, instead of writing to standard out. I'm glad that IPython has updated their interface—intercepting standard out is a messy way to do it. (What happens in parallel processing?)

So I approve this version, too. More enthusiastically.

PR #1939, opened 2022-12-02
feat: move long-range metadata checks to constructors

### LINKED ISSUE #1910: Long-range metadata checks during ak.Array creation

In https://github.com/scikit-hep/awkward/pull/1904#discussion_r1029626522, I said that a particular construction (`IndexedArray` of `IndexedOptionArray`) doesn't need to be considered because it's not a valid array, and while we try to use `simplify_optiontype` and `simplify_uniontype` to ensure that these things don't exist, erroneous cases have slipped through in the past, and they might again.

First of all, we even need a way of talking about these states. @agoose77 suggested "canonical representation," but there are multiple levels of canonicity.

- Level 0, short-range metadata constraints: `Content` node with wrong-type or out-of-range metadata, like a `RegularArray` with a `size` of `"some_string"` or `-12`. This is already prohibited by the `Content` constructors and can never exist as an object.
- Level 1, long-range metadata constraints: `Content` node with metadata that is incompatible with another `Content` node, such as its direct child `content`. The above-mentioned `IndexedArray` of `IndexedOptionArray` is an example. These states are allowed to exist in the midst of a calculation, but they need to be fixed before handing them off to the user.
- Level 2, long-range value constraints: `Content` node with array values that are incompatible with itself or another node (in practice, always a direct child), such as an `IndexedArray` with a negative value in `index`, or with a value in `index` that is greater than or equal to the length of its `content`. The `validity_error` tests (`check_valid` argument of the `ak.Array` constructor) check these, as well as some long-range metadata constraints, like the fact that a `"string"` should be a list-type array that contains a `"char"` `NumpyArray`.

The level 2 checks are _O(n)_ where _n_ is the length of the array, so they have to be invoked by user request. Adding them to the `ak.Array` constructor, for instance, would be a bad performance regression, and almost always unnecessary.

(Also note that if a kernel function needs a particular level 2 constraint to be satisfied, it performs that check as part of its operation. That's usually inexpensive because the _O(n)_ loop was happening anyway and the relevant data are already in fast memory anyway.)

Level 1 checks, however, scale with the size of the tree but not with the lengths of any arrays. Low-level arrays (layouts) are allowed to be level-1 invalid while we work on them, but if we're ever wrapping them as high-level `ak.Arrays`, they should already be simplified: i.e. the `simplify_optiontype` method should be called on an `IndexedArray` of `IndexedOptionArray` to turn it into a single-level `IndexedOptionArray`. (Similarly for `simplify_uniontype`.)

Scaling with the size of the tree might still be a problem if it happens frequently: our internal code pulls layouts out of `ak.Arrays` and re-wraps them as new `ak.Arrays` quite a lot, and some records have thousands of fields. We wouldn't want the same checks to be performed on the same layouts, over and over. Perhaps the best way is to add a `_metadata_checked: bool` attribute to every `Content` instance. If `not _metadata_checked` in the `ak.Array` or `ak.Record` constructor, a check should be performed to either set it to `True` or to raise an error. Since Awkward Arrays are immutable, a `True` value of `_metadata_checked` at any level means that `_metadata_checked` is true for all nested children.

Some node types never need to be checked:

```python
class EmptyArray(Content):
    _metadata_checked = True
```

and

```python
class NumpyArray(Content):
    _metadata_checked = True
```

can be class attributes, `True` for all instances of the class. Similarly,

```python
class Content:
    _metadata_checked = False
```

as a class attribute would set the default that is only overridden with a new reference to the `True` PyObject when a check is actually performed, so this has zero overhead for temporary layouts that never get into any `ak.Array.

The use of the word "metadata" in level 1 checks won't ever be confused with level 0 because level 0 fails at [RAII time](https://en.wikipedia.org/wiki/Resource_acquisition_is_initialization).

Adding the long-range metadata check as an "always on" consistency test means that it can (and should) be removed from the `validity_error` tests, which would then focus exclusively on long-range value checks.

Whatever we implement for normal arrays will also work for `TypeTracerArrays` because we're only accessing the metadata. Maybe a little special care would be needed for `RecordArray`, whose `length` must be less than or equal to the `length` of all of its `contents`, as a long-range metadata constraint. This might need to speculatively pass at the `TypeTracerArray` stage (i.e. Dask graph creation) and fail when real arrays are executed (Dask worker nodes).

The value of all of this is to learn very quickly about any missing calls to `simplify_optiontype` or `simplify_uniontype`. I think that this should go in _after the 2.0.0 release_, rather than adding more work to the remaining time in the schedule. Technically, this would turn situations that might happen to work into situations that raise errors, but turning a silent error into a noisy one can hardly be called an API change.

### LINKED ISSUE #1927: Forbid UnionArrays/UnionForms with fewer than 2 contents in constructor
This is part of #1910, since a UnionArray with 0 or 1 `contents` is non-canonical (level 1). However, it affects the constructor arguments: UnionArray does not need an explicit `backend`. The constructor argument set appears in `__init__` (naturally), `copy`, and `simplified`.

jpivarski:
Which would make it impossible for a "non-canonical" layout to exist.

Fixes #1927 #1910

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-long-range-metadata-checks-in-layout-constructors/> once Read the Docs has finished building :hammer:


src/awkward/_connect/pyarrow.py
```diff
@@ -282,6 +282,11 @@ def mask_parameters(awkwardarrow_type):
         return None
 
 
+def revertable(modified, original):
+    modified.__pyarrow_original = original
+    return modified
```
jpivarski, line 287: The first constraint (in BitMaskedArray) revealed that the from-pyarrow code was assuming that we could temporarily wrap something with an option-type array and then remove it later.

We actually can't because the outer layer, with `simplified`, could merge in some non-revertable way with its contents.

To fix that, we now hold onto both the (possibly `simplified`) option-type array and the original array, so that reverting it is a matter of just pulling out the original, held as a `__pyarrow_original` attribute. (Don't you love Python?)

`ak.from_arrow` is a gatekeeper for `handle_pyarrow`/`popbuffers`/`popbuffers_finalize`. It's a private, internal implementation detail (notwithstanding the fact that some tests call it directly). All users, including third-party developers, must access it through `ak.from_arrow`, and this function now removes the `__pyarrow_original` attribute.

src/awkward/_connect/pyarrow.py
```diff
@@ -873,26 +917,11 @@ def to_awkwardarrow_type(
...
 
 def remove_optiontype(akarray):
-    assert type(akarray).is_option
-    if isinstance(akarray, ak.contents.IndexedOptionArray):
-        return ak.contents.IndexedArray(
-            akarray.index, akarray.content, parameters=akarray.parameters
-        )
-    else:
-        return akarray.content
+    return akarray.__pyarrow_original
```
jpivarski, line 920: The removal code has become a lot simpler because the original was saved.

src/awkward/operations/ak_from_arrow.py
```diff
@@ -68,4 +68,10 @@ def _impl(array, generate_bitmasks, highlevel, behavior):
             if awkwardarrow_type.record_is_scalar:
                 out = out._getitem_at(0)
 
+    def remove_revertable(layout, **kwargs):
+        if hasattr(layout, "__pyarrow_original"):
+            del layout.__pyarrow_original
+
+    out.recursively_apply(remove_revertable)
```
jpivarski, line 75: We don't let an array get into the users' hands with a bunch of `__pyarrow_original` attributes in it.

src/awkward/contents/numpyarray.py
```diff
@@ -405,14 +405,14 @@ def _mergeable(self, other, mergebool):
...
+                self._data.dtype.type,
+                other._data.dtype.type,
             ):
                 return False
 
             if not matching_dtype and np.timedelta64 in (
-                self._data.dtype,
-                other._data.dtype,
+                self._data.dtype.type,
+                other._data.dtype.type,
```
jpivarski, line 426: This had always been wrong: it was considering all datetime types and timedelta types as mergeable, but they must not be. They should only be mergeable if they're exactly the the same dtype, including units.

The reason it was wrong was because `np.datetime64` and `np.timedelta64` are both type objects, like `self._data.dtype.type`, but unlike `self._data.dtype`, which is a dtype object. Sheesh!

src/awkward/contents/unionarray.py
```diff
@@ -1412,9 +1412,13 @@ def _to_arrow(self, pyarrow, mask_node, validbytes, length, options):
...
+            this_length = 0
+            if len(this_index) != 0:
+                this_length = this_index.max() + 1
+
             values.append(
                 content._to_arrow(
-                    pyarrow, mask_node, this_validbytes, this_index.shape[0], options
+                    pyarrow, mask_node, this_validbytes, this_length, options
                 )
             )
```
jpivarski, line 1421: Wow! That was wrong!

When propagating the expected `length` through `_to_arrow`, we should note that the length of a UnionArray possibility must be (at least) one more than the largest value in the possibility's index, not the length of the index itself!

Like, if the index values associated with a possibility are `[391, 440, 87]`, the content's length must be at least `441`, so that the index `440` exists. Not 3!

And if the index values associated with a possibility are `[2, 2, 2, 2, 0, 1, 1]`, then we only need the content to have length `3`, not 7!

src/awkward/operations/ak_from_buffers.py
```diff
@@ -129,7 +131,7 @@ def getkey(form, attribute):
 }
 
 
-def reconstitute(form, length, container, getkey, backend):
+def reconstitute(form, length, container, getkey, backend, simplify):
```
jpivarski, line 134: I had to add a back door to `from_buffers` (notice that this only affects the implementation; it's inaccessible from high-level; we can make it public at any point in the future, if we have to, and then we'll think about the semantics).

The problem is that ArrayBuilder creates non-canonical Forms! (So the Form elements can't have the metadata constraint that Content nodes do; we have to support the existence of non-canonical Forms.) Specifically, if there's a union with missing values, ArrayBuilder makes `IndexedOptionArray(UnionArray(...))`. Actually, having ArrayBuilder _accumulate_ a canonical layout would make it a lot more complicated than it is; we should allow this to continue.

With this back door, `ak_from_buffers._impl` simplifies the layout while it builds it. This means that the resulting array will not have the same Form as is given to `ak_from_buffers._impl`, but there isn't a need for ArrayBuilder Forms to match the resulting array—it just has to make an array of the right type.

This back door should _not_ be what `ak.from_buffers` does because users would be very surprised if it normally produced arrays whose Form differs from the user-provided Form. The only reason we can get away with it with ArrayBuilder is because a human or third-party library is not involved in the construction of that Form.

We could make `ak.ArrayBuilder.form` canonicalize its own Form before showing it to the user by calling `.simplified` on its elements (recursively, from the leaves to the root). Then the secret will be safe...

tests/test_0835-datetime-type.py
```diff
@@ -457,6 +457,29 @@ def test_min_max():
...
+                np.datetime64("2020-06-27T10:41:11"),
+            ],
+            [
+                np.datetime64("2020-02-27T10:41:11"),
+                np.datetime64("2020-03-27T10:41:11"),
+                np.datetime64("2020-01-27T10:41:11"),
+            ],
+        ]
+    ).layout
     assert to_list(array.min()) == [
```
jpivarski, line 483: Previously, this was passing even though we can't take the minimum of a union because it was illicitly unionizing the datetimes.

jpivarski (commented):
What a great way to catch bugs!

Slowly adding canonicity requirements to the constructors has been changing the layouts of some of the tests to equivalent layouts that fail, even though they shouldn't (being equivalent). These are real bugs and they're getting fixed now.

src/awkward/contents/numpyarray.py
agoose77: Yes, the dtype system is a real pain for this. I was adding an API to abstract these kind of checks in the nplike PR, but for now we just need to fix them!

src/awkward/operations/ak_from_buffers.py
agoose77: > The problem is that ArrayBuilder creates non-canonical Forms! (So the Form elements can't have the metadata constraint that Content nodes do; we have to support the existence of non-canonical Forms.) Specifically, if there's a union with missing values, ArrayBuilder makes IndexedOptionArray(UnionArray(...)). Actually, having ArrayBuilder accumulate a canonical layout would make it a lot more complicated than it is; we should allow this to continue.

Aha! This was my worry yesterday, I'm glad to see that I'm not totally bonkers! :rofl: 

Being pedantic, could we go for a name that makes it sound scarier to set `simplify=False`? Don't look at me for suggestions :eyes: !

> Then the secret will be safe...

Lol..

src/awkward/contents/unionarray.py
agoose77: That's a fantastic bug to catch. A true Purple Emperor.

src/awkward/contents/unmaskedarray.py
```diff
@@ -184,7 +192,7 @@ def _getitem_fields(self, where, only_fields=()):
         )
 
     def _carry(self, carry, allow_lazy):
-        return UnmaskedArray(
+        return UnmaskedArray.simplified(
             self._content._carry(carry, allow_lazy), parameters=self._parameters
         )
```
jpivarski, line 197: ByteMaskedArray and UnmaskedArray have the only two `_carry` operations with a `simplified` because they propagate the `_carry` to their `_content`. The other option-types and union-type don't do that, and so they shouldn't `simplify`.

Also, there are places in the code where we assume that a carried layout has an unchanged type (like in broadcasting: `if isinstance(x, UnionArray): x[mask].project(i)` where the `x[mask]` would carry and make it not a `UnionArray` anymore).

src/awkward/operations/ak_concatenate.py
```diff
@@ -18,9 +18,6 @@ def concatenate(
             outermost dimension is `0`, followed by `1`, etc., and negative
             values count backward from the innermost: `-1` is the innermost
             dimension, `-2` is the next level up, etc.
-        merge (bool): If True, combine data into the same buffers wherever
-            possible, eliminating unnecessary #ak.contents.UnionArray8_64 types
-            at the expense of materializing #ak.contents.VirtualArray nodes.
         mergebool (bool): If True, boolean and numeric data can be combined
             into the same buffer, losing information about False vs `0` and
             True vs `1`; otherwise, they are kept in separate buffers with
```
jpivarski, line 21: The `merge` argument has been removed from `ak.concatenate` because now, one cannot produce an unmerged union. `mergebool` is still a meaningful argument.

jpivarski (commented):
Applying the UnionArray constraints was mostly an exercise in finding out how many tests make use of non-canonical unions, just because it's easier to write the tests.

I could convert most of the tests to something more meaningful, usually by promoting a union of numbers and numbers to numbers and something else, where the something else is strings (put the number in quotes) or lists (put the number in brackets).

src/awkward/types/recordtype.py
```diff
@@ -107,7 +107,7 @@ def _str(self, indent, compact):
                     in (
                         "unknown",
                         "string",
-                        "bytes",
+                        "byte",
```
jpivarski: There were a lot of `s/bytes/byte/` errors.

src/awkward/contents/content.py
```diff
@@ -42,14 +42,45 @@ class Content:
...
+                    )
+                )
+            if not self.is_indexed and parameters.get("__array__") == "categorical":
+                raise ak._errors.wrap_error(
+                    TypeError(
+                        '{} is not allowed to have parameters["__array__"] = "{}"'.format(
+                            type(self).__name__, parameters["__array__"]
+                        )
+                    )
+                )
```
jpivarski: If a layout node type is _not_ supposed to have a certain parameter, then that is now enforced. Sometimes, `"string"` parameters got onto UnionArrays because parameters from various levels were over-merged.

jpivarski (commented):
Adding a constructor-level check for parameters—that `"string"` must contain `"char"`, `"bytestring"` must contain `"byte"`, etc.—found a lot of little bugs.

jpivarski:
The argument for why a UnionArray must never have zero `contents` is because then it would be impossible to instantiate a single instance. That is,

```
union[]
```

is the [bottom type](https://en.wikipedia.org/wiki/Bottom_type), which is the type of a non-halting program or a program that raises an exception (no return value). It might be conceivable to have an empty array of this type, but we already have an `EmptyArray` type (of `unknown`, which is more like a [top type](https://en.wikipedia.org/wiki/Top_type).) So that's why I'm excluding it.

By contract, records ([product types](https://en.wikipedia.org/wiki/Product_type) instead of [sum types](https://en.wikipedia.org/wiki/Tagged_union)) can have zero `contents`.

```python
>>> ak.Array([{}, {}, {}, {}, {}])
<Array [{}, {}, {}, {}, {}] type='5 * {}'>

>>> ak.Array([(), (), (), (), ()])
<Array [(), (), (), (), ()] type='5 * ()'>
```

(An instance of a product type is all of its `contents`, which can be a pack of zero things. An instance of a sum type is any one of its `contents`, which has to be a non-empty set.)

The reason a union can't have 1 type is because

```
union[X] == X
```

so we canonicalize it to just `X`. Therefore, the minimum `contents` a union can have (from now on) is 2.

jpivarski:
This is ready for review, @agoose77! I don't think it would be easy to review it via the final diff; that's why I added a lot of comments along the way, to explain what each of the commits is doing. For the most part, one commit is a new constraint on one node type, plus fixes for all of the bugs that it revealed. The work on UnionArray was done in two commits.

Don't feel compelled to go deep into this, checking each line or something. That would take a very long time. It would be enough to cross-check my motivations for things and to ask me questions about anything that doesn't seem to be covered already.

Thanks!

jpivarski:
Oh, I should write down the new rules. They are enforced on `Content` subclasses only, not `Forms`.

An option or indexed type (including `IndexedOptionArray`) must:

  * never contain another option or indexed type
  * never contain a union type

A union type must:

  * never contain another union type
  * contain at least two `contents`
  * none of those `contents` may be `mergeable` with any of the other `contents` (mergeability is an equivalence relation)

A string must:

   * be a list type (`ListArray`, `ListOffsetArray`, `RegularArray`) with `__array__ = "string"`
   * contain a 1-dimensional uint8 `NumpyArray` with `__array__ = "char"`

Same for `"bytestring"` and `"byte"`.

A categorical must:

   * be an indexed type with `__array__ = "categorical"`

There's also a `"sorted_map"` that isn't used yet.

jpivarski:
`ak.validity_error` is still needed to check the "level 2" rules: long-range value constraints. These are not checked by every constructor because that would be expensive (_O(n)_). In the implementations of the kernels, any level 2 check that is needed by a particular kernel is checked on the fly by that kernel (when it's not expensive because the data are running through the CPU anyway). So any level 2 errors that come up are _kernel_ errors.

The "level 1" rules were taken _out_ of `ak.validity_error` check, because they're now guaranteed for every existing `Content` instance.

src/awkward/_broadcasting.py
```diff
@@ -685,9 +685,9 @@ def continuation():
                         if isinstance(x, RegularArray):
                             if maxsize > 1 and x.size == 1:
                                 nextinputs.append(
-                                    IndexedArray(
-                                        tmpindex, x.content[: x.length * x.size]
-                                    ).project()
+                                    x.content[: x.length * x.size]._carry(
```
agoose77, line 688: This carries the content of the `RegularArray`, which might be a `RecordArray`. We used to introduce index types for these nodes as they have a multiplicative effect; many contents within a single content that we don't want to copy eagerly. This change would remove that optimisation. Is it needed in this case? I would assume that a lazy carry should use `simplified` internally, so don't we want to allow lazy?

src/awkward/_typetracer.py
```diff
@@ -601,6 +601,8 @@ def arange(self, *args, **kwargs):
             and ak._util.is_integer(step)
         ):
             length = max(0, (stop - start + (step - (1 if step > 0 else -1))) // step)
+        else:
+            length = UnknownLength
```
agoose77, line 605: Good, I keep fixing this in my long-term PRs, but they have all yet to merge :cry:

src/awkward/_util.py
```diff
@@ -541,11 +541,27 @@ def direct_Content_subclass_name(node):
...
 
-def merge_parameters(one, two, merge_equal=False):
+meaningful_parameters = {
+    ("__array__", "string"),
+    ("__array__", "bytestring"),
+    ("__array__", "char"),
+    ("__array__", "byte"),
+    ("__array__", "sorted_map"),
+    ("__array__", "categorical"),
+}
```
agoose77:
```suggestion
meaningful_parameters = frozenset({
    ("__array__", "string"),
    ("__array__", "bytestring"),
    ("__array__", "char"),
    ("__array__", "byte"),
    ("__array__", "sorted_map"),
    ("__array__", "categorical"),
})
```

src/awkward/behaviors/string.py
```diff
@@ -249,7 +249,7 @@ def register(behavior):
     behavior["__typestr__", "char"] = "char"
 
     behavior["bytestring"] = ByteStringBehavior
-    behavior["__typestr__", "bytestring"] = "bytes"
+    behavior["__typestr__", "bytestring"] = "byte"
```
agoose77: Is this better? I think `bytes` implies bytestring better than `byte`. The `byte` type should be for the `byte` `__array__`, no?

src/awkward/contents/recordarray.py
```diff
@@ -138,7 +111,6 @@ def __init__(
         for content in contents:
             if backend is None:
                 backend = content.backend
-                break
```
agoose77, line 141: I had changed this, but it must have been in a subsequent unmerged PR. Good catch

src/awkward/contents/recordarray.py
```diff
@@ -192,6 +181,16 @@ def simplified(
...
         return cls(contents, fields, length, parameters=parameters, backend=backend)
 
+    @property
+    def is_tuple(self):
+        return self._fields is None
+
+    @property
+    def as_tuple(self):
+        return RecordArray(
+            self._contents, None, self._length, parameters=None, backend=self._backend
```
agoose77, line 191: Do we want to drop the parameters here? My feeling is that unless we know the parameters are tied to the record-ness (vs tuple-ness), we should keep them and let the user drop them manually if needed.

I realise you didn't add this method, but we're touching it now.

src/awkward/contents/unionarray.py
```diff
@@ -988,9 +994,14 @@ def _reverse_merge(self, other):
                 AssertionError("FIXME: handle UnionArray with more than 127 contents")
             )
 
-        parameters = ak._util.merge_parameters(self._parameters, other._parameters)
-        return ak.contents.UnionArray(
-            tags, index, contents, parameters=parameters, backend=self._backend
+        parameters = ak._util.merge_parameters(
+            self._parameters,
+            other._parameters,
+            exclude=ak._util.meaningful_parameters,
```
agoose77, line 1000: Love this.

src/awkward/forms/bitmaskedform.py
```diff
@@ -10,35 +8,6 @@
...
-    def __copy__(self):
-        return self.copy()
-
-    def __deepcopy__(self, memo):
-        return self.copy(
-            mask=copy.deepcopy(self._mask, memo),
-            content=copy.deepcopy(self._content, memo),
-            parameters=copy.deepcopy(self._parameters, memo),
-        )
-
```
agoose77, line 41: Why are we dropping `copy.[deep]copy` support for `Form`s?

src/awkward/forms/form.py
```diff
@@ -422,4 +422,19 @@ def length_zero_array(
...
+            form=self,
+            length=1,
+            container={
+                "": b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+            },
+            buffer_key="",
+            backend=backend,
+            highlevel=highlevel,
+            behavior=behavior,
+            simplify=False,
```
agoose77:
```suggestion
        # A length-1 array will need at least N bytes, where N is the largest dtype (e.g. 256 bit complex)
        # Similarly, a length-1 array will need no more than 2*N bytes, as all contents need at most two
        # index-types e.g. `ListOffsetArray.offsets` for their various index metadata. Therefore, we 
        # create a buffer of this length (2N) and instruct all contents to use it (via `buffer_key=""`).
        # At the same time, with all index-like metadata set to 0, the list types will have zero lengths
        # whilst unions, indexed, and option types will contain a single value.
        return ak.operations.ak_from_buffers._impl(
            form=self,
            length=1,
            container={
                "": b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
            },
            buffer_key="",
            backend=backend,
            highlevel=highlevel,
            behavior=behavior,
            simplify=False,
```

agoose77: Do we support the 128-bit complex types? If so, we would need to increase the container size here.

src/awkward/operations/ak_from_arrow.py
agoose77: I'd like to find a way to avoid having to manually catch this, but I think it's probably not worth the effort given how these methods are called.

src/awkward/operations/ak_from_avro_file.py
```diff
@@ -67,10 +67,13 @@ def from_avro_file(
...
-    return ak.from_buffers(
+    return ak.operations.ak_from_buffers._impl(
         form=form,
         length=length,
         container=container,
+        buffer_key="{form_key}-{attribute}",
+        backend="cpu",
         highlevel=highlevel,
         behavior=behavior,
+        simplify=True,
```
agoose77, line 78: Maybe we shouldn't pass this explicitly - I think of `simplify=False` as the meaningful value, rather than `simplify=True` which is really "don't do anything dangerous".

src/awkward/operations/ak_from_buffers.py
```diff
@@ -73,10 +73,12 @@ def from_buffers(
             behavior=behavior,
         ),
     ):
-        return _impl(form, length, container, buffer_key, backend, highlevel, behavior)
+        return _impl(
+            form, length, container, buffer_key, backend, highlevel, behavior, False
```
agoose77, line 77: Shouldn't this be `True` for the user-facing function?
```suggestion
            form, length, container, buffer_key, backend, highlevel, behavior, True
```

src/awkward/operations/ak_from_iter.py
```diff
@@ -99,6 +99,13 @@ def _impl(iterable, highlevel, behavior, allow_record, initial, resize):
...
-        form, length, buffers, highlevel=highlevel, behavior=behavior
+    return ak.operations.ak_from_buffers._impl(
+        form,
+        length,
+        buffers,
+        buffer_key="{form_key}-{attribute}",
+        backend="cpu",
+        highlevel=highlevel,
+        behavior=behavior,
+        simplify=True,
```
agoose77, line 110: Same here (don't pass `simplify`?)

src/awkward/types/listtype.py
```diff
@@ -47,7 +47,7 @@ def _str(self, indent, compact):
             out = ["string"]
 
         elif self.parameter("__array__") == "bytestring":
-            out = ["bytes"]
+            out = ["byte"]
```
agoose77: Same discussion as for typestring

src/awkward/types/recordtype.py
agoose77: Same discussion as for typestring

tests/test_0032-replace-dressedtype.py
```diff
@@ -135,8 +135,8 @@ def test_builder_string():
     assert (
         ak.operations.to_json(a, convert_bytes=bytes.decode) == '["one","two","three"]'
     )
-    assert repr(a) == "<Array [b'one', b'two', b'three'] type='3 * bytes'>"
-    assert str(ak.operations.type(a)) == "3 * bytes"
+    assert repr(a) == "<Array [b'one', b'two', b'three'] type='3 * byte'>"
+    assert str(ak.operations.type(a)) == "3 * byte"
```
agoose77: Same discussion as for typestring

agoose77:
I've left comments. I didn't read most of the tests, because they are very long and given that we directly execute them, I'm comfortable with the fact that they pass being enough evidence. Obviously, we could design a bad test that succeeds all the time, but I think that's the time-correctness tradeoff for doing a code review?

src/awkward/operations/ak_from_arrow.py
jpivarski: It's a hack for this particular function, `ak.from_arrow`, and it doesn't extend beyond it. This manual removal is like garbage collection: we could try something slicker but then we might remove too much or too little—a final run that clears the private attribute from the whole tree is the simplest, cleanest way to be sure.

src/awkward/contents/numpyarray.py
jpivarski: This is the only place where datetime mergeability is considered.

src/awkward/operations/ak_from_buffers.py
jpivarski: I'm thinking we might eventually expose it publicly, since it's possible to make non-canonical Forms. The thing to warn users about is that you'll get a different Form out than what you put in. Maybe `reform=False` (as the default)?

src/awkward/_broadcasting.py
jpivarski: I can't see the arguments of this `_carry` above (on my phone), but if the second argument is `allow_lazy=True`, then a RecordArray will take that opportunity to put itself inside an IndexedArray, rather than actually carrying. When applying the optimization, we didn't replace all these `_carry` calls, we added the option. There are a few places where `allow_lazy=False` because without it, there would be an infinite loop. So we allow lazy as much as possible—this might be replacing code that always carried.

src/awkward/behaviors/string.py
jpivarski: This is the convention that was fixed for a while, it was just wrong in a few places in the codebase. We might even find that some of the "lost strings" issues are fixed now.

The reason for singular is the same as the reason why "int" or "float" (singular) is used for an array of integers or floats. Or "string" for an array of strings. This is a NumpyArray in which each element is a byte, not a ListArray in which every element is a bytestring.

Similarly for "char" in "string" arrays.

src/awkward/types/recordtype.py
jpivarski: I answered below.

src/awkward/contents/recordarray.py
jpivarski: We need to drop the parameters because the type has changed. Only operations that return the same type can keep the parameters.

In fact, when turning a record into a tuple, overloads would almost certainly break. Most overloads are defined in terms of record fields, which are now gone.

src/awkward/forms/bitmaskedform.py
jpivarski: `deepcopy` might automatically work. I haven't tested it, but if it doesn't, then we'll need to put the `__deepcopy__` back. It's just that the Forms were inconsistent same I made them consistent by removal, rather than adding. I know that Contents need explicit `__deepcopy__`, but I don't know that for Forms, which don't have any associated arrays.

src/awkward/_broadcasting.py
agoose77: OK, here we were actually _project_ing the `IndexedArray`, so you've not actually removed an optimisation. My question is then whether we should be projecting this vs leaving it in place. As it's in broadcasting, I suspect we project for good reason. :)

src/awkward/forms/form.py
jpivarski: You're right! That's true; it needs to be at least as large as the largest primitive itemsize, and it also needs to be at least as large as two int64s for an `offsets` array of length 2 (for a ListOffsetArray of length 1). That's what I was thinking of here.

In awkward/types/numpytype.py, we have the complete list of supported dtypes, and the code only provisionally adds some types because NumPy doesn't have them on all platforms. The above would support any 128-bit types, do we have any 256-bit types? (`complex256`?)

src/awkward/behaviors/string.py
agoose77: I'm missing something here:
```python
In [2]: ak.Array([b"hi", b"hello"])
Out[2]: <Array [b'hi', b'hello'] type='2 * byte'>
```
In my understanding of our string/char distinction, this is wrong — this should be "bytes" as we have multiple "byte", in the same way that "string"s are multiple "chars". I assume that users mostly should not be aware of the single-byte / single-char leaf nodes; these are an implementation detail of strings/bytestrings in Awkward.

Am I off-piste here? :)

src/awkward/contents/recordarray.py
agoose77: OK, I'll make a mental note :)

src/awkward/forms/bitmaskedform.py
agoose77: OK, I've tested it and it *seems* to work. I'm not used to seeing a class just work with `copy`, that's a new one on me.

src/awkward/operations/ak_from_avro_file.py
jpivarski: This is internal code, so the name can be changed. (I suggested "reform" above.) The reason for "simplify" is because it calls "simplified" and a difference between these names would confuse developers. If `simplify=True`, then `ak_from_buffers._impl` internally calls `simplified`; if not, it does not.

The Avro implementation also makes non-canonical Forms, for the same reason as ArrayBuilder: putting an IndexedOptionArray around a UnionArray is the simplest and most symmetric thing to do, but it violates Arrow convention. Nobody cares what the Avro reader's internal Form was (same for ArrayBuilder), so changing it on the fly won't bother anybody.

Don't worry—this argument is absolutely hidden from users and we can redefine it at any time.

src/awkward/forms/form.py
agoose77: >  That's true; it needs to be at least as large as the largest primitive itemsize, and it also needs to be at least as large as two int64s for an offsets array of length 2 

Oops! Yes, `ListOffsetArray` uses at least two integers. I assumed we supported a length-1 case... Not sure why.

We do allow NumPy's `complex256`, though as you say, it's only available on some platforms. I don't immediately see any code using it directly, so I think we'd only see it is users on compatible platforms passed such a type in.

I *think* this code would need to expand the array for `complex256` — indices must be integer types, which are at-most 64 bits, so the largest index array needs to be 128 bits. Meanwhile, contents would need to be 256 bits to support a single `complex256`?

src/awkward/operations/ak_from_arrow.py
agoose77: Perfect is the enemy of good enough ;)

src/awkward/operations/ak_from_iter.py
jpivarski: Needed here because of ArrayBuilder making non-canonical Forms.

Halfway through implementing this, I was wondering if maybe were being too aggressive with defining option-of-union as non-canonical, when we use it everywhere. However, this PR demonstrates that it's possible to remove all cases of it, it is redundant with union-of-options, and Arrow has set a convention that we might as well use.

src/awkward/types/listtype.py
jpivarski: Same: it has to be `"byte"` to work, and the convention has already been out in the community (in principle). Even with a major release, it would be too messy to change.

Also, the singular is technically right, because the type name describes an element of the array (singular), not the array (plural).

tests/test_0032-replace-dressedtype.py
jpivarski: Same.

src/awkward/behaviors/string.py
agoose77: Just in case it _is_ crossed wires, we have
- `byte` (single byte)
- `char` (signle char)
- `byte` (many byte)
- `string` (many char)

I'm confused by the lack of symmetry.

src/awkward/operations/ak_from_buffers.py
jpivarski: No, it's important that the user-facing function does not change the user's Forms. Users would be very surprised if they passed a Form to `ak.from_buffers` and they got an array with a different Form.

Originally, this back door was called `from_ArrayBuilder`, but non-canonical Forms can also come from AwkwardForth.

The name could even be `backdoor`, but we don't have to decide that now.

src/awkward/operations/ak_from_iter.py
agoose77: I think the most important thing is that users can still express these layouts, but they won't survive in the result. That doesn't harm the user authoring a layout; it only harms users that wish to use these layouts in some way. 

In the long run, it means we might write code that handles fewer cases (e.g. only handle union of option). This does mean that users won't be able to use Awkward to operate upon index-of-union layouts, but the trade-off is code that's easier to reason about, and fewer places that we need the equivalent of `ensure_canonical` if we can't express certain logic nicely. Of course, it should be noted that most Awkward operations change the layout _anyway_ and we don't provide any guarantees that this won't happen, so I think this is more than reasonable.

jpivarski:
I've been aggressive with clicking "resolve conversation" because it would otherwise be hard for me to tell what I've responded to or not on this phone. Reopen any that you think needs more discussion.

(I'm using "resolve" like "over" in radio.)

src/awkward/behaviors/string.py
jpivarski: The second to last (many byte strings being "byte") is wrong and you found a mistake that I added.

```python
<Array [b'hi', b'hello'] type='2 * byte'>
```

is indeed wrong. I know what mistake I made: our parameter names are `"byte"` and `"bytestring"`, but the typestr for bytestring needs to be `"bytes"` to satisfy the Datashape language.

I think you found a case that should have remained `"bytes"`, because it's not a parameter. It's a typestr.

src/awkward/forms/form.py
jpivarski: Yes, it's okay for that `bytes` object to be _longer_ than it needs to be. It needs to be at least 256 bits (32 bytes) so that complex256 doesn't break.

src/awkward/operations/ak_from_buffers.py
agoose77: That's a good point. So, what we're doing here is allowing someone to user `from_buffers` if they have a non-canonical form, but we ensure that the result _is_ canonical. 

This is a bad default for users, who likely don't want the form to be coerced. I think we should expose this to the users, then - it never results in non-canonical layouts. It just let's them use non-canonical representations, which I think is reasonable; it's perfectly sound to express such a tree, we just don't support any operations on it.

Maybe we can call this parameter `strict=True` at the high-level, so users can say "I don't care if the array type changes" as an opt-in?

src/awkward/behaviors/string.py
agoose77: And, for posterity, my confusion was that Jim was unifying our type strings with Datashape, so `bytestring →bytes`. I saw `bytestring → byte` and thought we were dropping the user-visible byte-bytes distinction!

agoose77:
```suggestion
    behavior["__typestr__", "bytestring"] = "bytes"
```

agoose77:
Let's try that again — I ran the failed tests with `pytest ... --lf`, but it somehow missed this one.

PR #1940, opened 2022-12-02
fix: backends should have defaults for user-facing operations

agoose77:
The safe default for `ak._backends.backend_of` is to fail if the inputs don't have a well defined backend e.g. for `lists`. However, for user-facing operations, this is not correct; we want to assume NumPy for these types. A `grep` shows that only a few `operations` need this fix.

I haven't added any tests because we should really do this as part of a coverage pass for all of our operations.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-backend-default/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -21,6 +21,7 @@ from awkward.nplikes import (
 from awkward.typing import (
     Any,
     Callable,
+    Final,
     Protocol,
     Self,
     Tuple,
@@ -40,6 +41,8 @@ KernelType: TypeAlias = Callable[..., None]
 
 @runtime_checkable
 class Backend(Protocol[T]):
+    name: str
+
     @property
     @abstractmethod
     def nplike(self) -> NumpyLike:
@@ -60,6 +63,8 @@ class Backend(Protocol[T]):
 
 
 class NumpyBackend(Singleton, Backend[Any]):
+    name: Final[str] = "numpy"
+
     _numpy: Numpy
 
     @property
@@ -78,6 +83,8 @@ class NumpyBackend(Singleton, Backend[Any]):
 
 
 class CupyBackend(Singleton, Backend[Any]):
+    name: Final[str] = "cupy"
+
     _cupy: Cupy
 
     @property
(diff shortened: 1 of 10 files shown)


agoose77:
@jpivarski  would you be happy for me to change `ak.backend` to throw an exception if the input arrays don't have a common backend? Right now it returns `"mixed"`, which is not a valid backend name, and has less utility now that single arrays cannot be mixed.

jpivarski (approved):
So `backend_of` would normally raise an exception if it can't infer a backend without an explicit `default=cpu`?

https://github.com/scikit-hep/awkward/blob/31ce6578891862be1cab999378db7b5f7aaca945/src/awkward/_backends.py#L176-L182

Okay, yeah, it does. So you need to provide `default=cpu` for any user-facing functions.

Looks good; go ahead and merge it!

agoose77:
Yes. My call was that we really care about the backend, and due to interop it's very hard to tell if we get it wrong (JAX can adopt NumPy, CuPy can adopt NumPy). Better to fail loudly.

agoose77:
@jpivarski would you mind confirming on https://github.com/scikit-hep/awkward/pull/1940#issuecomment-1335852226 ? I know you're busy with lots of things at the moment, but don't want to merge this without an OK!

jpivarski:
Yeah, `"mixed"` no longer means anything and shouldn't be returned by anything. I'm in favor of doing things this way, though we've got to be careful to always provide a default (and it would always be `cpu`) in user-facing functions now.

It is better to fail loudly.

I hadn't noticed that this is a draft PR. If it's a draft because you still have a lot of user-facing functions to update, you don't need re-approval to merge it. I'm in favor of the principle.

agoose77:
Thanks! I'm going to make two changes; removing `"mixed"` from our high level API, and to associate the backend name on the backend object so this code becomes more trivial.

jpivarski:
Raising an error in mixed cases is right. We should not have any array with different backends in different nodes.

agoose77:
OK, this is done. I've added a check to ensure that we throw a `ValueError` if users somehow encounter a typetracer

PR #1942, opened 2022-12-02
feat: replace nplike with backend in `to_buffers`

agoose77:
This follows up #1922 to remove any reference of `nplike` in the high level API, and renames `nplikes` to `_nplikes`

I also fixed the CUDA test generation, which I failed to fix in #1922

Note-to-self, write a git hook to run those.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-feat-to-buffers-backend/> once Read the Docs has finished building :hammer:

### DIFF
--- a/dev/generate-tests.py
+++ b/dev/generate-tests.py
@@ -679,7 +679,7 @@ def gencudakerneltests(specdict):
                 )
 
                 f.write(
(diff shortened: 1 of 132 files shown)


jpivarski (approved):
Good: `nplikes` is now hidden and users deal with `backends` everywhere instead. I've scanned over the differences twice and they look like straightforward changes.

I'd like to merge this right away, since it will be easier to merge into my big PR now, rather than later.

Issue #434, opened 2020-09-04
argmin/max with keepdims=True returns regulararray


nsmith-:
If I run
```
In [3]: a = ak.Array([[0, 1, 2], [3, 4], [5]])

In [5]: ak.argmin(a, axis=1, keepdims=True)
Out[5]: <Array [[0], [0], [0]] type='3 * 1 * ?int64'>
```
the output is of fixed size in the last axis, and as such it cannot be used in a slice:
```
In [6]: a[ak.argmin(a, axis=1, keepdims=True)]
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-6-83ba94cb4af4> in <module>
----> 1 a[ak.argmin(a, axis=1, keepdims=True)]

~/src/awkward-1.0/awkward1/highlevel.py in __getitem__(self, where)
    939         """
    940         return awkward1._util.wrap(
--> 941             self._layout[where], self._behavior, cache=self._cache
    942         )
    943

ValueError: slice items can have all fixed-size dimensions (to follow NumPy's slice rules) or they can have all var-sized dimensions (for jagged indexing), but not both in the same slice item

(https://github.com/scikit-hep/awkward-1.0/blob/0.2.36/src/libawkward/array/RegularArray.cpp#L846)
```

A workaround is to use `ak.singletons`:
```
In [7]: ak.singletons(ak.argmin(a, axis=1))
Out[7]: <Array [[0], [0], [0]] type='3 * var * int64'>

In [8]: a[_7]
Out[8]: <Array [[0], [3], [5]] type='3 * var * int64'>
```

jpivarski:
Actually, "singletons" was created (along with "firsts") because I hadn't thought of using `keepdims=True`!

The original rationale for having `keepdims=True` return regular-typed lists was because I knew the length was always equal to 1. However, it makes more sense to keep it the same as whatever the dimension used to be.

Additionally, regular-typed lists should remain regular-typed when reducing (with or without `keepdims=True`) unless there are any None values (because they can change the lengths of the outputs). That all has to be decided based on the types, not whether the lists happen to have the same number of items in each. For example, UnmaskedArray counts the same as ByteMaskedArray or any other option-type, even though it doesn't have any Nones.

Everything has been implemented and tested in PR #447 (exhaustively tested: all combinations of variable and regular-typed lists three levels deep, with `axis=-1`, `-2`, `-3`...)

[...]

PR #1943, opened 2022-12-03
fix: consolidate regular indexing

### LINKED ISSUE #1358: Regular indexing behaves like jagged index for RegularArray (v2)
### Version of Awkward Array

c2ddda2f67abf2d7b3112e5752c7c34c3467052d

```pycon
>>> array = ak.from_numpy(np.arange(4 * 4).reshape(4, 4))
>>> array[array > 4].ndim
2
>>> (array > 4).layout.purelist_isregular
True
>>> array[ak.to_numpy(array > 4)].ndim
1
```

### Cause

The NumPy-like indexing checks for `NumpyArray`, not `RegularArray`. We should use something like `maybe_toNumpyArray` beforehand to fulfil this.

agoose77:
Fixes #1358 by using `maybe_to_NumpyArray`. `RegularArray`s that succeed with `maybe_to_NumpyArray()` follow _NumPy_ indexing. Previously, they followed Awkward Indexing.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-regular-indexing/> once Read the Docs has finished building :hammer:


agoose77:
@jpivarski this PR touches the indexing logic, and needs a careful eye.

It makes two big changes:
1. fields are now only selected if a _list_ of strings is passed, rather than *any* iterable. This prevents arrays of strings selecting fields. This also means that an empty list behaves like an empty _array_, which was already the case.
2. `RegularArray`s that succeed with `maybe_to_NumpyArray()` follow _NumPy_ indexing. Previously, they followed Awkward Indexing.

(2) is the tricky point. First, a short history recap:
- v1 prevented mixed regular-ragged layouts, leading to #434, so `keepdims=True` was changed to return the same type as the layout.
- v2 removed this restriction.
- #1811 made `keepdims=True` always return a `1` dimension (I.e. reverting the fix to #434)

This PR addresses #1358, which exposes the lack of symmetry between NumpyArray and RegularArray for indexing. 
The fix for this issue is simple - make RegularArray agree with NumpyArray. However, this means users need to be careful. Consider the following:
```python
>>> a = ak.Array([[0, 1, 2], [3, 4], [5]])
>>> a[ak.argmin(a, axis=1, keepdims=True)]
<Array [[0], [3], [5]] type='3 * var * ?int64'>
>>> a[ak.argmin(a, axis=1, keepdims=True, mask_identity=False)]
<Array [[[0, 1, 2]], [[0, ...]], [[0, 1, 2]]] type='3 * 1 * var * int64'>
```
The mask from `mask_identity=False` (case 1, implicit) means that the Awkward indexing is followed. Whereas in case 2 the mask is omitted and an identity used instead. Due to this, the array can trivially be converted to a NumPy array, and we use NumPy indexing.

I think the question here is not "is this correct?" because if
- we want to support multiple indexing types, and
- we decide upon which one to use according to the layout type

then the observed behaviour is consistent with this. It's only that, from a UX perspective, it's only the difference between `?int64` and `int64` that means you get one or the other. 

Are you comfortable with this policy? (And indeed, anyone else on the team!)

jpivarski (changes requested):
Actually, I intended for any Iterable of strings to count as a fields selection, including Awkward Arrays of strings. Empty Iterables are only ambiguous (field-selection or row-selection by integers?) if untyped, and Awkward Arrays are one way of providing a runtime type. NumPy arrays are another.

I'd prefer to keep that feature. I guess there weren't any tests preventing you from changing it, but I did check that it worked while developing it.

On the other example with the `?int64` vs `int64` toggling Awkward and NumPy slicing, I can see why that happens and I agree that it's confusing. We might say some point have to deprecate that behavior (warning on NumPy-style slicing and then phase it out—forcing users to explicitly wrap as NumPy at some point...?), but not now—it's too close to release time and that would be a major, major change.

So let's leave the confusing but "correct" argmax behavior. I'd like to switch back to letting any Iterable of strings select fields, though, if it's not too much trouble.

agoose77:
> On the other example with the ?int64 vs int64 toggling Awkward and NumPy slicing, I can see why that happens and I agree that it's confusing. We might say some point have to deprecate that behavior (warning on NumPy-style slicing and then phase it out—forcing users to explicitly wrap as NumPy at some point...?), but not now—it's too close to release time and that would be a major, major change.

As a precursor statement, I would not advocate changing slicing dramatically at this point. So, agreed! Also, I don't have a proposal here - this seems to me to be a fundamental constraint with our indexing; we support many indexing features, and they are not mutually exclusive, so we have to choose according to some scheme.

My long-running feeling has generally been that it's better to have useful type information (e.g. "I reduced this axis, and so I have a length-1 dimension") over pandering to the shortfalls of our indexing mechanism (e.g. "I reduced a var dimension, so it stays var"). And, to be clear, I don't think any of this is "wrong" or "right", it just has its pros and cons.

> We might say some point have to deprecate that behavior (warning on NumPy-style slicing and then phase it out—forcing users to explicitly wrap as NumPy at some point...?), but not now—it's too close to release time and that would be a major, major change.

Actually, this seems like the "best" solution to me in the long-long term. Introducing a new accessor like `.at` would make a lot of code less readable long-term, and I don't think most people are likely to _want_ NumPy indexing. Awkward's indexing is more powerful for the kinds of data we work with. We can discuss this after the release; I'll open an issue.

I raised this just to make sure I'm not doing anything daft ­— this is very fundamental code I'm changing (fixing), and I wanted to make sure that we're all on the same page.

agoose77:
I've reverted b4456fca08ff346ef201bd8df2c091480962909c and added new tests that enshrine this behavior :)

src/awkward/_slicing.py
```diff
@@ -151,9 +151,26 @@ def normalise_item(item, backend: ak._backends.Backend):
...
+                return out.data
+            else:
+                return out
+        else:
+            return as_numpy.data
+
     elif isinstance(item, ak.contents.Content):
         out = normalise_item_bool_to_int(normalise_item_nested(item))
-        if isinstance(out, ak.contents.NumpyArray):
+        if out.is_numpy:
```
jpivarski: Be careful: I think EmptyArray also has `is_numpy` (because it's coerceable). Is this what you want here?

jpivarski (approved):
Thanks for reverting the Awkward Array of strings behavior.

I approve the intention of this PR, and I have a question about only one line of code. When you've answered it for yourself, you can merge.

I can't check the code deeply, but that one line of code was the only one that looks suspicious to me.

PR #1947, opened 2022-12-05
feat: ensure user overloads are high-level interfaces

### LINKED ISSUE #1631: Make all user-written behavior overloads consistent in what arguments they accept and what they are expected to return
### Version of Awkward Array

HEAD

Probably this: arguments passed to the user-written behavior are `ak.Array`s (i.e. high-level), and it doesn't matter whether they return layouts (low-level) or `ak.Array`s.

This includes `ak.transform`.

Probably does not include Numba overloads.

agoose77:
Fixes #1631 

Changes:
- `ak.behavior["__broadcast__", ...]` promotes any `Content` arguments to `ak.Array`
- `ak.transform` promotes the `Content`, `Record` first-parameter value(s) to `ak.Array`
- `ufunc` custom cast promotes and `Content` arguments

Not Changed:
- `ufunc` application machinery; this already promotes arguments to `Content`.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-feat-ak-transform-array/> once Read the Docs has finished building :hammer:


jpivarski (approved):
Okay, so when a downstream developer writes a function for

- [ ] `ak.transform`
- [ ] `layout.recursively_apply` (which is public, maybe shouldn't be, now that `ak.transform` exists?)
- [ ] `ak.behavior["__cast__", type] = function` for custom casting
- [ ] `ak.behavior["__broadcast__", "parameter_name"] = function` for custom broadcasting
- [ ] `ak.behavior[ufunc, "parameter_name"] = function` for ufunc overloading (Vector probably uses this)
- [ ] `ak.behavior["__numba_typer__", "parameter_name"] = function` (including `"*", "parameter_name"`) for Numba
- [ ] `ak.behavior["__numba_lower__", "parameter_name"] = function` (including `"*", "parameter_name"`) for Numba
- [ ] `ak.behavior["__numba_typer__", "parameter_name", "attribute"] = function` for Numba
- [ ] `ak.behavior["__numba_lower__", "parameter_name", "attribute"] = function` for Numba
- [ ] `ak.behavior["__numba_typer__", unaryop, "parameter_name"] = function` for unary operators
- [ ] `ak.behavior["__numba_lower__", unaryop, "parameter_name"] = function` for unary operators
- [ ] `ak.behavior["__numba_typer__", "parameter_name1", binaryop, "parameter_name2"] = function` for binary operators
- [ ] `ak.behavior["__numba_lower__", "parameter_name1", binaryop, "parameter_name2"] = function` for binary operators
- [ ] `ak.behavior[reducer, "parameter_name"] = function` for overloaded reducers
- [ ] `ak.behavior[function, type, type, type...] = function` for generic overloads

the arguments that get passed to their function are always high-level ak.Arrays? Do those ak.Arrays always have a relevant `behavior`?

I'm not completely sure if always high-level (ak.Array) is correct or if always low-level (layout) is correct. Users of these features are downstream developers who are well aware of layouts, and most of what they need to do is modifying the internal tree structure of the array. For instance, they're usually written as "`if layout.is_optiontype: ... elif layout.is_uniontype: ...`". Hiding layouts inside of ak.Arrays is not doing these developers any favors.

Another side of this is that the developer might write a function that returns an ak.Array or returns a layout. If the arguments are ak.Arrays, we need to be able to accept either ak.Arrays or layouts as return values. If the arguments are always layouts, it would be easier to argue that developers should always return layouts.

In fact, that's sounding better to me: the idea that this level of interface—functions that modify the internal tree structure of an array—should be at the layout level throughout, input arguments and output return values.

Especially since there are so many ways to use `ak.behavior`. The smaller delta will be to make them all take layouts, rather than ak.Arrays. My original issue, #1631, was just to make them all consistent, not necessarily to make them all high-level.

agoose77:
Closed in favour of #1951

PR #1950, opened 2022-12-05
fix: UnionArray.simplified preserves parameters.

### LINKED ISSUE #1686: `UnionArray.simplify_uniontype()` loses parameters

jpivarski:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-unionarray-simplified-preserve-parameters/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -239,7 +239,16 @@ class UnionArray(Content):
                                     contents[k].length,
                                 )
                             )
-                            contents[k] = contents[k].merge(inner_cont)
+                            old_parameters = contents[k]._parameters
+                            contents[k] = (
+                                contents[k]
+                                .merge(inner_cont)
+                                .copy(
+                                    parameters=ak._util.merge_parameters(
+                                        old_parameters, inner_cont._parameters
+                                    )
+                                )
+                            )
                             unmerged = False
                             break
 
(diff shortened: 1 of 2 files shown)


src/awkward/contents/unionarray.py
```diff
@@ -239,7 +239,16 @@ def simplified(
                                     contents[k].length,
                                 )
                             )
-                            contents[k] = contents[k].merge(inner_cont)
+                            old_parameters = contents[k]._parameters
+                            contents[k] = (
+                                contents[k]
+                                .merge(inner_cont)
+                                .copy(
+                                    parameters=ak._util.merge_parameters(
```
agoose77, line 247: I would imagine that we want to lose categorical here, too?

jpivarski: `"__array__": "categorical"` can't be passed into `UnionArray.simplified` as a `parameters` argument. It will fail at some point (with a nice error message) because the output of `UnionArray.simplified` can't be an `IndexedArray` or `IndexedOptionArray`, and these are the only two node types that are allowed to have `"__array__": "categorical"` (now that wrong combinations are excluded in constructors).

agoose77: Aha! I had assumed as much, yet I couldn't see where we checked this. I had forgotten that we now test this in `Content._init`, not the per-layout constructors. Good.

PR #1951, opened 2022-12-05
refactor: clarify input types for user overloads

### LINKED ISSUE #1631: Make all user-written behavior overloads consistent in what arguments they accept and what they are expected to return

agoose77:
Closes #1631 

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-ufunc-as-layout/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_to_dataframe.py
+++ b/src/awkward/operations/ak_to_dataframe.py
@@ -12,7 +12,7 @@ def to_dataframe(
 ):
     """
     Args:
-        array: Data to convert into one or more Pandas DataFrames.
+        array: Array-like data (anything #ak.to_layout recognizes).
         how (None or str): Passed to
             [pd.merge](https://pandas.pydata.org/pandas-docs/version/1.0.3/reference/api/pandas.merge.html)
             to combine DataFrames for each multiplicity into one DataFrame. If
--- a/src/awkward/operations/ak_to_dataframe.py
+++ b/src/awkward/operations/ak_to_dataframe.py
@@ -121,7 +121,7 @@ def to_dataframe(
               3         4.0  NaN
     """
     with ak._errors.OperationErrorContext(
-        "ak.to_layout",
+        "ak.to_dataframe",
         dict(
             array=array,
             how=how,
--- a/src/awkward/operations/ak_transform.py
+++ b/src/awkward/operations/ak_transform.py
@@ -1,7 +1,6 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
 import copy
-from collections.abc import Iterable
 
 import awkward as ak
 
@@ -168,9 +167,7 @@ def transform(
        * options (dict): Options provided to #ak.transform.
 
     If there is only one array, the `transformation` function must either return
-    None or return an array: #ak.contents.Content, #ak.Array, or a NumPy,
-    CuPy, etc. array. (The preferred type is #ak.contents.Content; all others
-    are converted to a layout.)
+    None or return an #ak.contents.Content.
 
     If there are multiple arrays (`more_arrays`), then the transformation function
     may return one array or a tuple of arrays. (The preferred type is a tuple, even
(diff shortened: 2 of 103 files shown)


agoose77:
@jpivarski  could you reflect on where this sits with you? Still to decide on `__broadcast__`. That feels more low level than `ufunc` etc.

jpivarski:
> @jpivarski could you reflect on where this sits with you? Still to decide on `__broadcast__`. That feels more low level than `ufunc` etc.

The `__broadcast__` overload is low-level; it's used inside `broadcast_and_apply`, so that one can write the string equality operation and have it not broadcast the strings to have the same lengths in the two arrays. That's very much a tree-structure manipulation thing.

src/awkward/operations/ak_to_dataframe.py
```diff
@@ -121,7 +121,7 @@ def to_dataframe(
               3         4.0  NaN
     """
     with ak._errors.OperationErrorContext(
-        "ak.to_layout",
+        "ak.to_dataframe",
```
jpivarski, line 124: Good catch!

jpivarski (commented):
I take it that this part of the documentation is no longer correct:

https://github.com/scikit-hep/awkward/blob/6c5d24b19f354ad8a0806822c2bba1fc6844a52a/src/awkward/operations/ak_transform.py#L120-L135

I'll fix the documentation.

src/awkward/operations/ak_to_dataframe.py
agoose77: Oh, I missed that during my pass over. Good spot.

jpivarski: No, actually, you caught that!

agoose77: Ah yes, I meant the documentation! Always the thing one misses :/

agoose77:
@jpivarski happy with this? I didn't touch the numba code, because numba deals with types.

jpivarski (approved):
My understanding is that this PR changes `ak.transform` to now require Content instances (or None) as a return value from the function that the downstream developer writes. I don't see any other changes in how developer-provided functions get called or expect as return values—I guess we decided that all other cases should stay as they are.

I ran the Vector tests; they still work with this PR.

Also, I see that `numpytype` has been removed as a `to_layout` argument.

Based on my understanding, I approve this PR and it can be merged!

agoose77:
> My understanding is that this PR changes `ak.transform` to now require `Content` instances (or None) as a return value from the function that the downstream developer writes. I don't see any other changes in how developer-provided functions get called or expect as return values—I guess we decided that all other cases should stay as they are.

Yes, it makes `ak.transform` less permissive in what the user can return.

It *also*:
- ensures that all `_array_ufunc_custom_cast` casts its arguments to `ak.Array` (in practice I think they always are)
- removes `numpytype` from `ak.to_layout`
- ensures that `__array_function__` overload results are converted to `ak.Array`.

src/awkward/operations/ak_transform.py
```diff
@@ -499,46 +501,33 @@ def action(layout, **kwargs):
...
 
-            if isinstance(out, ak.highlevel.Array):
-                return (out.layout,)
+            elif isinstance(out, tuple):
+                for x in out:
+                    if not isinstance(x, ak.contents.Content):
+                        raise ak._errors.wrap_error(
+                            TypeError(
+                                f"transformation must return a Content, tuple of Contents, or None, not {type(x)}\n\n{x!r}"
+                            )
```
agoose77:
```suggestion
                            TypeError(
                                f"transformation must return a Content, tuple of Contents, or None, not a tuple containing {type(x)}\n\n{x!r}"
                            )
```

jpivarski:
These are all good changes; it can be merged. ("Ensure" is not a change in behavior, but important for API!)

jpivarski:
This one would close #1631, wouldn't it?

PR #1955, opened 2022-12-06
refactor: removed the 'ak._util.extra' argument helper.

jpivarski:
It was a vestige of Python 2: as you can see, we can express the keyword-only arguments after a `*arrays` without any trouble in Python 3. This is relevant because the documentation will be generated with more information now—these extra arguments and their defaults will appear in the HTML rendering of the function signatures.

Also, `ak.broadcast_arrays` was missing information about its `broadcast_parameters_rule` argument. Presumably, that was an oversight? (And it demonstrates the problem with `ak._util.extra`.)

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-remove-extra-argument-helper/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -70,9 +70,12 @@ def broadcast_arrays(
             ak.broadcast_arrays(
                 arrays = (array([1, 2]), array([[ 0.1,  0.2,  0.3],
                [10. , 20....
-                depth_limit = None,
-                highlevel = True,
-                behavior = None,
+                depth_limit = None
+                broadcast_parameters_rule = 'one_to_one'
+                left_broadcast = True
+                right_broadcast = True
+                highlevel = True
+                behavior = None
             )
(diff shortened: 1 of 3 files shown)


src/awkward/operations/ak_broadcast_arrays.py
```diff
@@ -6,18 +6,31 @@
...
             for no broadcasting.
+        broadcast_parameters_rule (str): Rule for broadcasting parameters, one of:
+            - `"intersect"`
+            - `"all_or_nothing"`
+            - `"one_to_one"`
+            - `"none"`
+        left_broadcast (bool): If True, follow rules for implicit
+            left-broadcasting, as described below.
+        right_broadcast (bool): If True, follow rules for implicit
+            right-broadcasting, as described below.
```
agoose77, line 33: I *swear* we had this at one point. Maybe it never made it to the final PR :frowning:

agoose77 (approved):

PR #1956, opened 2022-12-06
feat: ak.from_rdataframe should accept a single string 'columns'.

### LINKED ISSUE #1931: `from_rdataframe` API should handle case that users pass single string
### Version of Awkward Array

main

A user ran into a cryptic error message when they passed in a string instead of a tuple (it's easy to think you've written a tuple if you forget the comma). We should either handle this (`columns: str | tuple[str, ...]`) or raise an exception for a bare string.

In general, I think our current API design is to check for a "sized iterable that is not a string/bytestring", though we could also check for a sized iterable with at least one string.

jpivarski:
This adds a guard, but it also introduces a way of calling `ak.from_rdataframe` (which was already partially in the documentation anyway).

It's to be expected that users will sometimes pass a single string into `columns`. When this happens, we could

  * raise an error to let them know that's wrong (it's already doing that, but the error message could be more clear)
  * turn a single string `x` into `(x,)`, squashing the distinction between a string and a length-1 tuple of strings
  * use the single string interface to mean "I want an Awkward Array _of_ a column" (what this function did before #1625) and the collection of strings (even if there's only one of them) to mean "I want an Awkward Array of multiple columns." The latter will always be a record array, and if the collection has only one string in it, the record array will have only one field in it.

In this PR, I opted for the third case. It seems to me to be the most useful one.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-from_rdataframe-should-accept-a-single-string/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_from_rdataframe.py
+++ b/src/awkward/operations/ak_from_rdataframe.py
@@ -8,10 +8,17 @@ def from_rdataframe(rdf, columns):
     Args:
         rdf (`ROOT.RDataFrame`): ROOT RDataFrame to convert into an
             Awkward Array.
-        columns (str or tuple of str): A column or multiple columns to be
+        columns (str or iterable of str): A column or multiple columns to be
             converted to Awkward Array.
 
-    Converts ROOT Data Frame columns into an Awkward Array.
+    Converts ROOT RDataFrame columns into an Awkward Array.
+
+    If `columns` is a string, the return value represents a single RDataFrame column.
+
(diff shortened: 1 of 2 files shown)


ianna (approved):
Looks great! Thanks.

agoose77 (approved):

agoose77:
That's all reviewers! I'll merge.

PR #1960, opened 2022-12-06
fix: unknown type column

### LINKED ISSUE #1932: `type_to_form` returns non-JSON string in case of error
### Version of Awkward Array

main

When using the `RDataFrame` integration, users attempting to work with unsupported column types see a cryptic error message. This occurs because `type_to_form` returns the string `"unsupported type"` for unsupported types, which then triggers an exception when Awkward attempts to call `json.loads` on the generated form.

Could we use `std::variant` here to catch the error case?

ianna:
----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/ianna-1932-type_to_form-returns-non-json-string-in-case-of-error/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_connect/rdataframe/from_rdataframe.py
@@ -166,6 +166,12 @@ def from_rdataframe(data_frame, columns):
 
         else:  # Convert the C++ vectors to Awkward arrays
             form_str = ROOT.awkward.type_to_form[col_type](0)
+
+            if form_str == "unsupported type":
+                raise ak._errors.wrap_error(
+                    TypeError(f'"{col}" column\'s type "{col_type}" is not supported.')
+                )
+
             form = ak.forms.from_json(form_str)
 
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is good: it fixes the observed error. Is there any other non-valid value that `form_str` can take? Is it true that any other value it can take is JSON and represents a real Form that `from_buffers` can either use or will fail with an appropriate error message there?

(This is just a question for follow-up work. This PR is an improvement in itself and I'll be merging it now. Thanks!!!)

ianna:
> This is good: it fixes the observed error. Is there any other non-valid value that `form_str` can take? Is it true that any other value it can take is JSON and represents a real Form that `from_buffers` can either use or will fail with an appropriate error message there?
> 
> (This is just a question for follow-up work. This PR is an improvement in itself and I'll be merging it now. Thanks!!!)

Yes, any other value returned by `type_to_form` and passed here is a valid JSON and represents a Form.

PR #1962, opened 2022-12-06
fix!: always broadcast `with_field` assignments against existing array

### LINKED ISSUE #1936: `array['x'] = 1` raises an `AttributeError`
### Version of Awkward Array

main

#### Replacement with scalar
Replacing with an array with a scalar throws an `AttributeError`

```python
array = ak.Array({
    "x": [1, 2, 3]
})
array['x'] = 4
```

It seems that this also fails in `1.10.0`.

This should be an easy fix, and I think it should be next release as it is fairly fundamental.

#### Replacement with array
Replacing with an array doesn't broadcast, but adding a new field with an array does.

```python
array = ak.Array({
    "x": [1, 2, 3]
})
array['x'] = [1]
array["x"]
```

I'm currently working on docs, so I'm filing the issues I find here in the mean time.

agoose77:
This PR fixes #1936 by removing the branch in which no broadcasting takes place.

It also narrows the acceptable type of `where` from an `iterable` (of strings) to a `sequence`, 

### Note
This is a small change, but I want to justify it in the wider context of Awkward.

We currently do iterable→sequence conversion internally, and I don't think there's any benefit to accepting an iterable. My view is that unless you can naturally express an algorithm without consuming the iterable, then it's better just to accept sequences. 

If we wanted to retain the iterable type without converting the iterable to a sequence, then we'd need to do something like
```python
if ak._util.is_sized_iterable(where):
    tail = iter(where)
    head = next(tail)
    if not isinstance(head, str):
        raise ...    
```
There's no performance benefit for the user, or readability benefit for us.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-fix-item-assignment/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_with_field.py
+++ b/src/awkward/operations/ak_with_field.py
@@ -1,7 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
 import copy
-from collections.abc import Iterable
+from collections.abc import Sequence
 
 import awkward as ak
 
@@ -13,9 +13,9 @@ def with_field(array, what, where=None, *, highlevel=True, behavior=None):
     Args:
         array: Array-like data (anything #ak.to_layout recognizes).
         what: Array-like data (anything #ak.to_layout recognizes) to add as a new field.
-        where (None or str or non-empy iterable of str): If None, the new field
+        where (None or str or non-empy sequence of str): If None, the new field
             has no name (can be accessed as an integer slot number in a
-            string); If str, the name of the new field. If iterable, it is
+            string); If str, the name of the new field. If a sequence, it is
             interpreted as a path where to add the field in a nested record.
         highlevel (bool): If True, return an #ak.Array; otherwise, return
             a low-level #ak.contents.Content subclass.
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
There's a Python culture of being willing to accept Iterables instead of concrete lists, but we've run into trouble a few times with this policy because we often want to call `len` on the collection we're given.

So it needs to be a Sized Iterable: that would be exactly the level of abstraction we want, since this is something that's guaranteed to have `__len__` and `__iter__`. Usually with our input collections, we don't use anything more than that.

Sequence is more: in addition to the above, it also has `__getitem__`, and if you're emulating a class (as opposed to accepting arguments, as in this case), it will also make a bunch of methods you don't want: `__contains__`, `__reversed__`, `index`, `count`. But here, we're just accepting arguments and having too many methods doesn't matter.

Sized Iterable is not a combination that people care about much in the Python world, so we can jump over that Goldilocks solution to requiring Sequences.

Actually, I'm probably even overthinking it. This is fine.  `:)`

agoose77:
Plus, we can always relax this if it becomes important.

PR #1963, opened 2022-12-06
feat: add `ak.without_field`

### LINKED ISSUE #1961: `ak.without_field`

It is not possible to remove a field from an `RecordArray` without rebuilding the layout manually e.g. with `ak.zip`. 

I think a natural extension of `with_field` and `__setitem__` would be `without_field` and `__delitem__`. Unlike `with_parameter`, where `None` has special meaning, `None` values are supported as array values, so we need a distinct function

agoose77:
This fixes #1961 by implementing `ak.without_field`.

This is not urgent to merge; it's simple and not a next-release item, but I added it whilst I was thinking about it.

### DIFF
--- a/src/awkward/operations/ak_without_field.py
+++ b/src/awkward/operations/ak_without_field.py
@@ -1,0 +1,92 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+from collections.abc import Sequence
+
+import awkward as ak
+
+np = ak._nplikes.NumpyMetadata.instance()
+
+
+def without_field(array, where, *, highlevel=True, behavior=None):
+    """
+    Args:
+        array: Array-like data (anything #ak.to_layout recognizes).
+        where (str or non-empy sequence of str): If str, the name of the field
+            to be removed. If a sequence, it is interpreted as a path where to
+            remove the field in a nested record.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+
+    Returns an #ak.Array or #ak.Record (or low-level equivalent, if
+    `highlevel=False`) with an existing field removed. This function does not
+    change the array in-place.
+
+    See #ak.Array.__delitem__ and #ak.Record.__delitem__ for a variant that
+    changes the high-level object in-place. (These methods internally use
+    #ak.without_field, so performance is not a factor in choosing one over the
+    other.)
+    """
+    with ak._errors.OperationErrorContext(
+        "ak.without_field",
+        dict(array=array, where=where, highlevel=highlevel, behavior=behavior),
+    ):
+        return _impl(array, where, highlevel, behavior)
+
+
+def _impl(base, where, highlevel, behavior):
+    if isinstance(where, str):
+        where = [where]
+    elif not (
+        ak._util.is_non_string_iterable(where)
+        and isinstance(where, Sequence)
+        and all(isinstance(x, str) for x in where)
+    ):
+        raise ak._errors.wrap_error(
+            TypeError(
+                "Field names must be given as a single string, or a sequence of strings"
+            )
+        )
+
+    behavior = ak._util.behavior_of(base, behavior=behavior)
+    base = ak.operations.to_layout(base, allow_record=True, allow_other=False)
+
+    def action(layout, lateral_context, **kwargs):
+        if isinstance(layout, ak.contents.RecordArray):
+            field, *next_where = lateral_context["where"]
+
+            # For parent record arrays, we don't change the fields, just
+            # modify the contents
+            i_field = layout.field_to_index(field)
+            if len(next_where):
+                next_contents = []
+                for i, content in enumerate(layout.contents):
+                    if i == i_field:
+                        # Visit this content to remove the next item in `where`
+                        next_content = content.recursively_apply(
+                            action, behavior, lateral_context={"where": next_where}
+                        )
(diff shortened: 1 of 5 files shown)


src/awkward/operations/ak_without_field.py
```diff
@@ -0,0 +1,93 @@
...
+    behavior = ak._util.behavior_of(base, behavior=behavior)
+    base = ak.operations.to_layout(base, allow_record=True, allow_other=False)
+
+    def action(layout, lateral_context, **kwargs):
+        if isinstance(layout, ak.contents.RecordArray):
+            field, *next_where = lateral_context["where"]
+
+            # For parent record arrays, we don't change the fields, just
+            # modify the contents
+            i_field = layout.field_to_index(field)
```
jpivarski, line 57: Good: users get a helpful error message if they pass in a field name that does not exist.

```python
>>> ak.Array([{"x": 1}]).layout.field_to_index("y")
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/recordarray.py", line 265, in field_to_index
    return self.form_cls.field_to_index(self, field)
  File "/home/jpivarski/irishep/awkward/src/awkward/forms/recordform.py", line 128, in field_to_index
    raise ak._errors.wrap_error(
IndexError: no field 'y' in record with 1 fields
```

src/awkward/operations/ak_without_field.py
```diff
@@ -0,0 +1,93 @@
...
+                "Field names must be given as a single string, or a sequence of strings"
+            )
+        )
+
+    behavior = ak._util.behavior_of(base, behavior=behavior)
+    base = ak.operations.to_layout(base, allow_record=True, allow_other=False)
+
+    def action(layout, lateral_context, **kwargs):
+        if isinstance(layout, ak.contents.RecordArray):
+            field, *next_where = lateral_context["where"]
```
jpivarski: I think this needs to be a `depth_context`. Let's see: I can construct an array that branches without having to consume field names by making it a union:

```python
>>> first = ak.Array([{"x": {"one": 1, "two": 2}, "y": {"three": 3, "four": 4}}])
>>> second = ak.Array([{"x": {"five": 5, "six": 6}, "y": {"seven": 7, "eight": 8}}])
>>> union = ak.concatenate([first, second], axis=0)
>>> union.type.show()
2 * union[
    {
        x: {
            one: int64,
            two: int64
        },
        y: {
            three: int64,
            four: int64
        }
    },
    {
        x: {
            five: int64,
            six: int64
        },
        y: {
            seven: int64,
            eight: int64
        }
    }
]
>>> union.tolist()
[{'x': {'one': 1, 'two': 2}, 'y': {'three': 3, 'four': 4}}, {'x': {'five': 5, 'six': 6}, 'y': {'seven': 7, 'eight': 8}}]
```

Though I haven't managed to break your implementation through some use of unions and

```python
del union[(field names)]
```

Still, is there a problem with using a `depth_context` here? I think it's conceptually what you're intending: you intend to consume a string when you descend through each record field, but separately for different subtrees.

jpivarski (approved):
It's good that the `del` syntax requires a tuple, not a list or any other type of Sequence, because

```python
array[["list", "of", "field", "names"]]
```

(in "load" contexts, _reading_ data) already means `"list", "of", "field", "names"` are all fields of the same record structure, while

```python
array["list", "of", "field", "names"]
```

(implicitly a tuple) already means `"names"` is a member of the record under the `"field"` field, which is a member of the record under the `"of"` field, which is a member of the record under the `"list"` field.

Also, the latter works in "store" contexts, _writing_ data:

```python
>>> array = ak.Array([{"x": {"one": 1, "two": 2}, "y": {"three": 3, "four": 4}}])
>>> array["x", "three"] = 3
>>> array.type.show()
1 * {
    y: {
        three: int64,
        four: int64
    },
    x: {
        one: int64,
        two: int64,
        three: int64
    }
}
```

You have added the tuple syntax for "del" contexts, _deleting_ data. I don't think the list/Sequence syntax can be meaningful for anything but the "load" context.

The only potential issue that I see is that you're using a `lateral_context` in the implementation. I haven't been able to prove it, but I think this is wrong, and you actually want a `depth_context`. See below.

(I'm clicking "approve" because I otherwise approve, and if you disagree after considering it, then you have some good reason for it.)

src/awkward/operations/ak_without_field.py
agoose77: > Still, is there a problem with using a depth_context here?

Right, I think this is fine. I recall making this decision because we don't need to mutate the context-value; a new dictionary is created whenever the code re-enters into `action` via the `if isinstance(layout, ak.contents.RecordArray):` branch. Using `lateral_context` was a deliberate don't-copy-this choice, rather than an oversight. 

I'm happy to use the `depth_context` if we strongly prefer it.

jpivarski: The conceptual difference is the pattern of updates:

https://awkward-array.org/doc/main/reference/generated/ak.transform.html#contexts

If a subtree is not supposed to communicate with any other subtree, as is the case here, `depth_context` is semantically correct. `lateral_context` is for if you _want_ it to accumulate as in the example linked above.

I guess it works for either context type here, since you're copying dicts in the implementation. I can see that if you switch to `depth_context`, there would be an additional (unnecessary) copy, but it's a dict of one field.

So yeah, please do switch to `depth_context`.

agoose77: > I guess it works for either context type here, since you're copying dicts in the implementation. I can see that if you switch to depth_context, there would be an additional (unnecessary) copy, but it's a dict of one field.

Yes, that's entirely it.

> So yeah, please do switch to depth_context.

Fab, I'll push a new commit.

PR #1964, opened 2022-12-06
docs: fix syntax highlighting

agoose77:
MyST-NB uses the notebook metadata to determine the language lexer, falling back upon the kernelspec if the required fields are not set. I believe that this leads to MyST-nb failing to highlight our code cells, which have a `{code-cell} ipython3` tag despite the notebook using a `python3` lexer. #1946 tried to solve this by setting the `python3` language for each code cell, but jupytext does not round trip this.

This PR sets and preserves the `pygments_lexer` explicitly. 

I ran the following Xonsh code to produce this PR:
```xonsh
metadata="""---
jupytext:
  notebook_metadata_filter: language_info.pygments_lexer
  text_representation:
    extension: .md
    format_name: myst
    format_version: 0.13
    jupytext_version: 1.14.0
kernelspec:
  display_name: Python 3 (ipykernel)
  language: python
  name: python3
language_info:
  name: python
  pygments_lexer: ipython3
---"""

cd docs/user-guide

for f in pg`how*.md`:
    content=f.read_text()
    i = content.find("---")
    j = content.find("---",i+3)
    f.write_text(metadata+content[j+3:])

sd r"cell\} python3" r"cell} ipython3" *.md
```

Plus some additional manual corrections for the C++ LayoutBuilder notebook.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-fix-langauge/> once Read the Docs has finished building :hammer:

PR #1966, opened 2022-12-06
fix: add IPython as Sphinx extension

agoose77:
This PR fixes the root cause that motivated #1964. In fact, the problem was exclusively on RTD, but when I checked my original fix, I checked the CI build result (which has not shown this highlighting problem). 

This PR adds IPython as a Sphinx extension, and reverts the changes to the `code-cell` language.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-docs-fix-syntax-highlighting/> once Read the Docs has finished building :hammer:

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -14,8 +14,6 @@ import os
 import json
 import datetime
 import runpy
-import sys
-import subprocess
 import pathlib
 
 # -- Project information -----------------------------------------------------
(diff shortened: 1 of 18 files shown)


jpivarski (approved):
I don't see a difference in syntax coloring between

https://awkward-array.readthedocs.io/en/main/reference/generated/ak.cartesian.html

and

https://awkward-array.readthedocs.io/en/agoose77-docs-fix-syntax-highlighting/reference/generated/ak.cartesian.html

but I guess the distinction would be in the `%magic` commands (IPython vs Python).

But this has my approval and can be merged when you're ready.

jpivarski:
The last message was an hour ago, so I guess you're done with this PR. Half the tests didn't run because it's a documentation-only change. I'll press the red button and merge it.

agoose77:
@jpivarski this PR is fixing the user guide, whose code cells are unstyled in the RTD main.

PR #1968, opened 2022-12-06
fix: add return_value='simplified' to ak.transform and revamp ak.firsts/ak.singletons

### LINKED ISSUE #1882: Check ak.firsts/ak.singletons semantics before the 2.0.0 release
Inspired by #983 and https://github.com/CoffeaTeam/coffea/issues/750, we need to take another look at `ak.firsts` and `ak.singletons` to make sure that they behave sensibly for all array types (e.g. that high-coverage suite of layouts copy-pasted from one v2 test to another) and add an `axis` parameter if it does not exist.

What `firsts` is supposed to do is: turn an array of `var * X` into an array of `option[X]` (of the same length) by returning the first item of each non-empty list and `None` for empty lists. `ak.firsts(array)` is like `array[:, 0]` except that it does not raise an exception if there are any empty lists. This has a natural extension to any `axis` and should work for lists buried within anything.

What `singletons` is supposed to do is: turn an array of `option[X]` into an array of `var * X` where each `None` maps to `[]` and each instance `x` of `X` maps to `[x]`. This can also be done at any `axis`.

The two functions are inverses of each other, exact inverses if `firsts` acts on lists of lengths 0 and 1 exclusively. These are the two ways of representing missing data: with a missingness token (`None`, which Awkward Array favors in all operations except this one) and with lists of length 0 and 1. (Option type is a monoid; I think only functional languages take this seriously.)

These two functions are pretty high in the count of number of uses across GitHub:

```
233 awkward.firsts(#)            <------
244 awkward.prod(#, axis=)
251 awkward.fromiter(#)
255 awkward.zip(#, with_name=)
255 awkward.singletons(#)        <------
264 awkward.all(#, axis=)
283 awkward.concatenate(#, axis=)
334 awkward.unflatten(#, #)
349 awkward.any(#, axis=)
368 awkward.fill_none(#, #)
388 awkward.sum(#)
422 awkward.count_nonzero(#)
446 awkward.Array(#)
473 awkward.where(#, #, #)
560 awkward.from_iter(#)
747 awkward.sum(#, axis=)
952 awkward.to_numpy(#)
1526 awkward.flatten(#)
1561 awkward.num(#)
```

and in the count of number of repos in which they were seen across GitHub:

```
13 awkward.singletons(#)        <------
13 awkward.local_index(#)
13 awkward.max(#)
13 awkward.count(#)
14 awkward.zip(#, depth_limit=)
14 awkward.from_parquet(#)
14 awkward.argsort(#, ascending=)
15 awkward.prod(#, axis=)
15 awkward.firsts(#)            <------
16 awkward.to_list(#)
17 awkward.max(#, axis=)
18 awkward.ones_like(#)
21 awkward.min(#, axis=)
22 awkward.combinations(#, #)
24 awkward.unzip(#)
24 awkward.all(#, axis=)
24 awkward.zeros_like(#)
24 awkward.all(#)
25 awkward.values_astype(#, #)
25 awkward.to_pandas(#)
25 awkward.broadcast_arrays(#, #)
26 awkward.count(#, axis=)
26 awkward.flatten(#, axis=)
28 awkward.concatenate(#)
28 awkward.behavior.update(#)
28 awkward.num(#, axis=)
34 awkward.concatenate(#, axis=)
35 awkward.zip(#, with_name=)
36 awkward.zip(#)
37 awkward.any(#, axis=)
38 awkward.unflatten(#, #)
38 awkward.sum(#)
38 awkward.where(#, #, #)
44 awkward.fill_none(#, #)
55 awkward.sum(#, axis=)
59 awkward.to_numpy(#)
63 awkward.num(#)
70 awkward.Array(#)
70 awkward.flatten(#)
```

jpivarski:
The first commit touches a lot of files because it propagates `Content.simplified` through `recursively_apply`, which was only partially done by #1939.

It also _changes_ the `ak.transform` API to take a `return_value = "none" | "original" | "simplified"` argument instead of `return_array = False | True`. The `ak.transform` function hasn't gotten out into the wild (much) yet, so this is a chance to make that parameter an n-way enum, rather than grow into a bunch of booleans that are only sometimes meaningful (like `return_simplified` and `return_array`; the first means nothing if the second is False).

This should have been done earlier, but `Content._recursively_apply` is now a hidden implementation detail. The `ak.transform` API is the public one that we'll support. Having two ways of doing the same thing is opening the door to inconsistency.

All of that was a stepping-stone to revamping `ak.firsts`. I didn't change its behavior within testing scope, but implementing it in terms of `Content._recursively_apply` is more robust. Branching layouts should now work automatically; they were `NotImplementedError` before.

After revamping `ak.singletons`, I'll mark this ready for review.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-revamp-ak.firsts-and-ak.singletons/> once Read the Docs has finished building :hammer:

### DIFF
--- a/src/awkward/operations/ak_firsts.py
+++ b/src/awkward/operations/ak_firsts.py
@@ -44,24 +44,55 @@ def firsts(array, axis=1, *, highlevel=True, behavior=None):
 
 
 def _impl(array, axis, highlevel, behavior):
-    layout = ak.operations.to_layout(array, allow_record=False, allow_other=False)
+    layout = ak.operations.to_layout(array)
+    behavior = ak._util.behavior_of(array, behavior=behavior)
     posaxis = layout.axis_wrap_if_negative(axis)
 
-    if posaxis == 0:
-        if len(layout) == 0:
-            out = None
-        else:
-            out = layout[0]
-    else:
-        if posaxis < 0:
-            raise ak._errors.wrap_error(
-                NotImplementedError("ak.firsts with ambiguous negative axis")
-            )
-        toslice = (slice(None, None, None),) * posaxis + (0,)
-        out = ak.operations.mask(
-            layout,
-            ak.operations.num(layout, axis=posaxis) > 0,
-            highlevel=False,
-        )[toslice]
+    if not ak._util.is_integer(axis):
+        raise ak._errors.wrap_error(
+            TypeError(f"'axis' must be an integer, not {axis!r}")
+        )
 
-    return ak._util.wrap(out, behavior, highlevel, like=array)
+    if posaxis == 0:
+        # specialized logic; it's tested in test_0582-propagate-context-in-broadcast_and_apply.py
+        slicer = ak.from_iter([None, 0])
+        if layout.length == 0:
+            out = layout[slicer[[0]]][0]
+        else:
+            out = layout[slicer[[1]]][0]
+
+    else:
+
+        def action(layout, depth, depth_context, **kwargs):
+            posaxis = layout.axis_wrap_if_negative(depth_context["posaxis"])
+            if posaxis == depth and layout.is_list:
+                nplike = layout._backend.index_nplike
+
+                # this is a copy of the raw array
+                index = starts = nplike.array(layout.starts.raw(nplike), dtype=np.int64)
+
+                # this might be a view
+                stops = layout.stops.raw(nplike)
+
+                empties = starts == stops
+                index[empties] = -1
+
+                return ak.contents.IndexedOptionArray.simplified(
+                    ak.index.Index64(index), layout._content
+                )
+
+            elif posaxis == 0:
+                raise ak._errors.wrap_error(
+                    np.AxisError(
+                        f"axis={axis} exceeds the depth of this array ({depth})"
+                    )
+                )
+
+            depth_context["posaxis"] = posaxis
+
+        depth_context = {"posaxis": posaxis}
+        out = layout._recursively_apply(
+            action, behavior, depth_context, numpy_to_regular=True
+        )
+
+    return ak._util.wrap(out, behavior, highlevel)
--- a/src/awkward/operations/ak_firsts.py
+++ b/src/awkward/operations/ak_firsts.py
@@ -94,7 +94,7 @@ def _impl(array, axis, highlevel, behavior):
             depth_context["posaxis"] = posaxis
 
         depth_context = {"posaxis": posaxis}
-        out = layout._recursively_apply(
+        out = layout.recursively_apply(
             action, behavior, depth_context, numpy_to_regular=True
         )
 
--- a/src/awkward/operations/ak_firsts.py
+++ b/src/awkward/operations/ak_firsts.py
@@ -81,7 +81,7 @@ def _impl(array, axis, highlevel, behavior):
                     ak.index.Index64(index), layout._content
                 )
 
-            elif posaxis == 0:
+            elif layout.is_leaf:
                 raise ak._errors.wrap_error(
                     np.AxisError(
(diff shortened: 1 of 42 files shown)


jpivarski:
The main thing that will look different to an ordinary user is that `ak.singletons` no longer descends to the first option-type level and works there; it needs an explicit `axis` argument. There were a few other functions that had to be upgraded that way once—I think that happened to `ak.fill_none` rather early—and `ak.singletons` is probably the last to need that correction. It's a user-facing API change.

`ak.firsts` already had an `axis` argument and it seemed to be working fine, apart from the fact that it didn't handle negative values.

So for `ak.singletons` and `ak.firsts` in particular, this PR served as a look-over to ensure that they're in good shape, plus the `axis` for `ak.singletons`. I had vaguely remembered these functions misbehaving for anything but the simplest jagged arrays, but that must have been fixed some time ago.

However, by reimplementing them with modern machinery (`_recursively_apply`), it reminded me about some things that needed to be fixed up in that machinery. I've been wanting to hide `_recursively_apply` for a while, and the issue with it needing to pass down a `simplify` option was just an oversight. Actually, a partial oversight, because `UnionArray._recursively_apply` was always applying the `simplified` constructor. It needs to be configurable, though.

So that's it from me: if you approve of these changes, let me know!

src/awkward/operations/ak_firsts.py
```diff
@@ -44,24 +44,55 @@ def firsts(array, axis=1, *, highlevel=True, behavior=None):
 
 
 def _impl(array, axis, highlevel, behavior):
-    layout = ak.operations.to_layout(array, allow_record=False, allow_other=False)
+    layout = ak.operations.to_layout(array)
+    behavior = ak._util.behavior_of(array, behavior=behavior)
```
agoose77, line 48: These keep getting missed, good change. Maybe we should have a utility function like `ak._util.decompose_array` that returns a `(layout, behavior)` tuple. That would make it easier to spot this.

src/awkward/operations/ak_firsts.py
```diff
@@ -44,24 +44,55 @@ def firsts(array, axis=1, *, highlevel=True, behavior=None):
...
+        raise ak._errors.wrap_error(
+            TypeError(f"'axis' must be an integer, not {axis!r}")
+        )
+
     if posaxis == 0:
-        if len(layout) == 0:
-            out = None
+        # specialized logic; it's tested in test_0582-propagate-context-in-broadcast_and_apply.py
+        slicer = ak.from_iter([None, 0])
+        if layout.length == 0:
```
agoose77, line 62: This is clever.

Perhaps a comment like this (but not this):
```suggestion
        # Build an integer-typed slice array, so that we can
        # ensure we have advanced indexing for both length==0
        # and length > 0 cases.
        slicer = ak.from_iter([None, 0])
        if layout.length == 0:
```

agoose77 (commented):
In general, I like this PR. I like the cleanups particularly. 

I want to talk about the choice to move `recursively_apply` to a private method. I don't want this conversation to distract from the review, so I'll put this as a [separate discussion](https://github.com/scikit-hep/awkward/discussions/1969)

I'd be _most_ happy if we split the `recursively_apply` rename into a separate PR to discuss separately. In that case, this PR gets a green light from me already.

However, I know that there is a time pressure here, so if you feel so strongly about https://github.com/scikit-hep/awkward/discussions/1969 and this particular part of the refactor, then it's your call - I do support merging this as-is.

jpivarski:
@agoose77 Now this PR does not hide `recursively_apply` (that will be the next PR), but it does everything the title says it does.

agoose77 (approved):

PR #1983, opened 2022-12-08
refactor: hide L3 API

agoose77:
- Adds `highlevel`, `behavior` arguments to the `ak.XXX` reducers
- Removes `layout.XXX` reducers
- Moves `layout._reduce` to `ak._do.reduce`
- Rename `layout.layout_equal` to `Content.is_equal_to`
- Removes unused methods of `NumpyArray`
- Replaces `NumpyArray.maybe_to_array()` with `NumpyArray.maybe_to_NumpyArray()`
- Renames `Content` / `Content` subclass public method/properties
- Remove `Content.tolist()` (but keep high-level implementation)
- Move `validity_error` to `ak._do.validity_error`
- Rename `Content.typetracer` with `Content.to_typetracer(...)`
- Replace `Content.forget_length()` with argument to `Content.to_typetracer()`
- Rename `ak.packed()` to `ak.to_packed()` (same for `Content.packed()`)

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/agoose77-refactor-l3-l4-split/> once Read the Docs has finished building :hammer:


jpivarski (approved):
We talked about this on Slack while it was in development, and I am completely in favor of these changes.

agoose77:
I made the last fixes that follow from the list of changes in #1972. In a couple of places, I think I've made the code worse - `NumpyArray._raw` is used externally but is private. However, the main priority here is hiding functions; later on we can clean this up.

docs/prepare_docstrings.py
```diff
@@ -336,169 +336,6 @@ def dofunction(link, linelink, shortname, name, astfcn):
             dofunction(link, linelink, shortname, toplevel.name, toplevel)
 
 
-categories = [
```
jpivarski, line 339: I see: `categories` is completely gone, now.

src/awkward/contents/numpyarray.py
```diff
@@ -660,7 +642,7 @@ def _is_unique(self, negaxis, starts, parents, outlength):
             return True
 
         elif len(self.shape) != 1 or not self.is_contiguous:
-            contiguous_self = self if self.is_contiguous else self.contiguous()
+            contiguous_self = self.to_contiguous()
```
jpivarski, line 645: I guess that was superfluous, considering that `to_contiguous` would return `self` if already contiguous.

jpivarski (approved):
I checked it over again; all is good. Now I'll merge it.

PR #1986, opened 2022-12-09
fix: replace `axis_wrap_if_negative` with `maybe_posaxis`, simpler and more correct

### LINKED ISSUE #1914: ak.is_none is off-by-one with negative axis and doesn't handle an edge case of positive axis
### Version of Awkward Array

HEAD

Here's an array with missing values at many levels, including unequal branches through a RecordArray:

```python
>>> array = ak.Array([
...     None,
...     [None],
...     [{"x": None, "y": None}],
...     [{"x": [None], "y": [None]}],
...     [{"x": [1], "y": [[None]]}],
...     [{"x": [2], "y": [[1, 2, 3]]}],
... ])
```

For `axis=0`, we expect item `0` to be `True` and the rest to be `False`, which is correct in v1 and v2 (only v2 is shown):

```python
>>> ak.is_none(array, axis=0).show(type=True)   # v2
type: 6 * bool
[True,
 False,
 False,
 False,
 False,
 False]
```

For `axis=1`, we expect item `0` to be `None`, item 1 to be a singleton list containing `True`, and all the rest to be singleton lists containing `False`, which is correct in v1 and v2:

```python
>>> ak.is_none(array, axis=1).show(type=True)   # v2
type: 6 * option[var * bool]
[None,
 [True],
 [False],
 [False],
 [False],
 [False]]
```

For `axis=2`, we expect similar pass-throughs on items `0` and `1`, and item `2` also passes through because record fields are at the same level of `axis` depth as the record itself. (That's a design decision across all of Awkward Array; we can't change it.) Item `3` contains singleton `[True]` for both fields because they're both missing at this level. v1 and v2 are both correct:

```python
>>> ak.is_none(array, axis=2).show(type=True)   # v2
type: 6 * option[var * ?{
    x: option[var * bool],
    y: option[var * bool]
}]
[None,
 [None],
 [{x: None, y: None}],
 [{x: [True], y: [True]}],
 [{x: [False], y: [False]}],
 [{x: [False], y: [False]}]]
```

For `axis=3`, we expect pass-throughs on items `0`, `1`, `2`, and `3`, but item `4` should pass-through on field `x` (`axis=3` is deeper than that field) while field `y` is `[[True]]`. Only v1 gets it right (showing v1):

```python
>>> ak.is_none(array, axis=3).tolist()   # v1
[None, [None], [{'x': None, 'y': None}], [{'x': [None], 'y': [None]}], [{'x': [1], 'y': [[True]]}], [{'x': [2], 'y': [[False, False, False]]}]]
```

v2 raises an AxisError:

```python
>>> ak.is_none(array, axis=3).show(type=True)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_is_none.py", line 28, in is_none
    return _impl(array, axis, highlevel, behavior)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_is_none.py", line 72, in _impl
    raise ak._errors.wrap_error(
numpy.AxisError: while calling

    ak.is_none(
        array = <Array [None, ..., [{x: [2], ...}]] type='6 * option[var * ...'>
        axis = 3
        highlevel = True
        behavior = None
    )

Error details: axis=3 exceeds the depth (2) of this array
```

It looks like there's an "and" when there ought to be an "or," or something like that.

--------------------------

Now for negative `axis`. For `axis=-1`, I would expect

```
type: 6 * option[var * ?{
    x: var * bool,
    y: option[var * bool]
}]
[None,
 [None],
 [{x: None, y: None}],
 [{x: [True], y: [None]}],
 [{x: [False], y: [[True]]}],
 [{x: [False], y: [[False, False, False]]}]]
```

because `axis=-1` goes all the way down to the option-type integers in field `y` and all the way down to the option-type integers in field `x`, which are different depths.

Instead, `axis=-1` in v1 and v2 implements what I would expect for `axis=-2`:

```python
>>> ak.is_none(array, axis=-1).show(type=True)   # v2
type: 6 * option[var * ?{
    x: bool,
    y: option[var * bool]
}]
[None,
 [None],
 [{x: True, y: None}],
 [{x: False, y: [True]}],
 [{x: False, y: [False]}],
 [{x: False, y: [False]}]]
```

```python
>>> ak.is_none(array, axis=-1).tolist()   # v1
[None, [None], [{'x': True, 'y': None}], [{'x': False, 'y': [True]}], [{'x': False, 'y': [False]}], [{'x': False, 'y': [False]}]]
```

There seems to have always been an off-by-one error in this accounting. It only needs to be fixed in v2.

Incidentally, if we make an array just like above but without `x`, it's still wrong in the same way. This second bug is not due to confusion about the record-branching, it's a constant off-by-one error.

```python
>>> array = ak.Array([None, [None], [{"y": None}], [{"y": [None]}], [{"y": [[None]]}], [{"y": [[1, 2, 3]]}]])

>>> ak.is_none(array, axis=-1).show(type=True)   # v2
type: 6 * option[var * ?{
    y: option[var * bool]
}]
[None,
 [None],
 [{y: None}],
 [{y: [True]}],
 [{y: [False]}],
 [{y: [False]}]]
```

```python
>>> array = ak.Array([None, [None], [{"y": None}], [{"y": [None]}], [{"y": [[None]]}], [{"y": [[1, 2, 3]]}]])

>>> ak.is_none(array, axis=-1).tolist()   # v1
[None, [None], [{'y': None}], [{'y': [True]}], [{'y': [False]}], [{'y': [False]}]]
```

jpivarski:
I can't believe it: `axis_wrap_if_negative` has always been wrong. This function dates back to the early days of Awkward 1.x. We've been using it for so long, in so many things, that I would have thought that it's been through every possible test.

To convert a negative `axis` into its non-negative equivalent, you have to know the absolute depth of the array. With `depth == 1` for a 1-dimensional array and increasing for each nested list, the non-negative axis is `axis + depth` where `axis` is negative. For instance, if `axis == -1` (deepest) and `depth == 2` (array of lists of primitives), `axis + depth == 1`: the equivalent non-negative `axis == 1`.

If there's branching (a union or a record), then there isn't a single depth; we can't evaluate `axis + depth`. However, as we recurse down, we'll eventually recurse through the branching structure. Inside of a record of `{x: var * float64, y: var * var * int32}`, for instance, we can evaluate one `depth` inside the `x` branch and a different `depth` inside the `y` branch. We want to do that so that `axis=-1` means "deepest for all branches," i.e. the `float64` for `x` and the `int32` for `y`.

That's why `axis_wrap_if_negative` passes `axis` through unchanged if (a) it's already non-negative or (b) the `layout` it's looking at is branched. The idea is that we'll repeatedly call `axis_wrap_if_negative` while the value we get back is negative. As soon as it's non-negative, we keep that value.

However, if we're recursing, the `layout` that we have at some level is not the whole layout. If we only have the `x` branch of a record, for instance, we'll see its `var * float64` as "`layout`", and if `x`'s record were embedded within other lists, we wouldn't see them. The depth we compute form _that_ "`layout`" is therefore relative to the depth of last branching (the record with `x` and `y`, in this example).

Since `axis_wrap_if_negative` takes this partial "`layout`" and the negative `axis` as arguments, it can _never_ determine the absolute depth, and therefore cannot correctly convert the `axis` into a non-negative `axis`... in general.

There's only one case that works anyway: if the branching happens to be at the top level, the relative depth is equal to the absolute depth and we get the right answer. I've looked around, and I've only ever seen tests of negative `axis` with non-branching arrays (no records or unions) or the branching is a RecordArray at top level (this special case). There are tests in which different fields have different depths, like

```
# * {x: var * float64, y: var * var * int32}
```

but never within another list, like

```
# * var * {x: var * float64, y: var * var * int32}
```

The examples in #1914 were apparently the first general test of negative to non-negative `axis` conversion.

So I started this PR by introducing a new function (which will eventually live in `ak._util`, but right now it's in `ak._do` to be next to `axis_wrap_if_negative` for juxtaposition), which does the conversion correctly. It needs the same arguments as `axis_wrap_if_negative` _and also_ the current `depth`, to correct for the fact that the `layout` it's seeing is not the whole array.

I've implemented `ak.is_none` with it. Now I'm going to look for other functions that were using `axis_wrap_if_negative` incorrectly. I don't think we'll fix all of these bugs before the 2.0.0 release tomorrow, but we'll get the pattern established.

They're bugs and we can change them without a deprecation cycle.

----
:books: The documentation for this PR will be available at <https://awkward-array.readthedocs.io/en/jpivarski-improved-axis-to-posaxis/> once Read the Docs has finished building :hammer:

### DIFF
--- a/tests/test_1137-num.py
+++ b/tests/test_1137-num.py
@@ -56,7 +56,7 @@ def test_numpyarray():
     ]
     with pytest.raises(ValueError) as err:
         ak._do.num(array, 4)
-    assert str(err.value).startswith("axis=4 exceeds the depth of this array (3)")
+    assert str(err.value).startswith("axis=4 exceeds the depth of this array")
 
 
 def test_regulararray():
@@ -73,7 +73,7 @@ def test_regulararray():
     ]
     with pytest.raises(ValueError) as err:
         ak._do.num(array, 4)
-    assert str(err.value).startswith("axis=4 exceeds the depth of this array (3)")
+    assert str(err.value).startswith("axis=4 exceeds the depth of this array")
 
     empty = ak.contents.RegularArray(
         ak.highlevel.Array([[1, 2, 3], [], [4, 5]]).layout, 0, zeros_length=0
@@ -110,7 +110,7 @@ def test_listarray():
     ]
     with pytest.raises(ValueError) as err:
         ak._do.num(array, 4)
-    assert str(err.value).startswith("axis=4 exceeds the depth of this array (3)")
+    assert str(err.value).startswith("axis=4 exceeds the depth of this array")
 
 
 def test_listoffsetarray():
@@ -138,7 +138,7 @@ def test_listoffsetarray():
     ]
     with pytest.raises(ValueError) as err:
         ak._do.num(array, 4)
-    assert str(err.value).startswith("axis=4 exceeds the depth of this array (3)")
+    assert str(err.value).startswith("axis=4 exceeds the depth of this array")
 
 
 def test_indexedarray():
@@ -171,7 +171,7 @@ def test_indexedarray():
 
     with pytest.raises(ValueError) as err:
         ak._do.num(array, 4)
-    assert str(err.value).startswith("axis=4 exceeds the depth of this array (3)")
+    assert str(err.value).startswith("axis=4 exceeds the depth of this array")
 
 
 def test_indexedoptionarray():
@@ -208,7 +208,7 @@ def test_indexedoptionarray():
 
     with pytest.raises(ValueError) as err:
         ak._do.num(array, 4)
-    assert str(err.value).startswith("axis=4 exceeds the depth of this array (3)")
+    assert str(err.value).startswith("axis=4 exceeds the depth of this array")
 
 
 def test_recordarray():
@@ -311,7 +311,7 @@ def test_array_3d():
     ]
     with pytest.raises(ValueError) as err:
         assert ak._do.num(array, axis=3)
-        assert str(err.value).startswith("axis=3 exceeds the depth of this array (2)")
+        assert str(err.value).startswith("axis=3 exceeds the depth of this array")
 
     assert to_list(ak._do.num(array, axis=-1)) == [
         [2, 2, 2, 2, 2],
@@ -323,4 +323,4 @@ def test_array_3d():
 
     with pytest.raises(ValueError) as err:
         assert ak._do.num(array, axis=-4)
-        assert str(err.value).startswith("axis=-4 exceeds the depth of this array (3)")
+        assert str(err.value).startswith("axis=-4 exceeds the depth of this array")
--- a/tests/test_1137-num.py
+++ b/tests/test_1137-num.py
@@ -29,34 +29,35 @@ def test_bytemaskedarray_num():
         None,
         [[], [10.0, 11.1, 12.2]],
     ]
-    assert ak._do.num(array, axis=0) == 5
-    assert ak._do.num(array, axis=-3) == 5
-    assert to_list(ak._do.num(array, axis=1)) == [3, 0, None, None, 2]
-    assert to_list(ak._do.num(array, axis=-2)) == [3, 0, None, None, 2]
-    assert to_list(ak._do.num(array, axis=2)) == [[3, 0, 2], [], None, None, [0, 3]]
-    assert to_list(ak._do.num(array, axis=-1)) == [[3, 0, 2], [], None, None, [0, 3]]
+    assert ak.num(array, axis=0) == 5
+    assert ak.num(array, axis=-3) == 5
+    assert to_list(ak.num(array, axis=1)) == [3, 0, None, None, 2]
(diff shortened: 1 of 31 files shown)


tests/test_1137-num.py
```diff
@@ -29,59 +29,60 @@ def test_bytemaskedarray_num():
...
+    assert to_list(ak.num(array, axis=2)) == [[3, 0, 2], [], None, None, [0, 3]]
+    assert to_list(ak.num(array, axis=-1)) == [[3, 0, 2], [], None, None, [0, 3]]
 
 
 def test_emptyarray():
     array = ak.contents.EmptyArray()
-    assert to_list(ak._do.num(array, 0)) == 0
-    assert to_list(ak._do.num(array, 1)) == []
-    assert to_list(ak._do.num(array, 2)) == []
+    assert to_list(ak.num(array, 0)) == 0
```
agoose77, line 42:
```suggestion
    assert to_list(ak.num(array, 0)) == 0
    assert to_list(ak.num(array, -1)) == 0
```

agoose77 (approved):

PR #1987, opened 2022-12-09
ci: output linkcheck information

agoose77:
This PR invokes the Sphinx `linkcheck` builder to ensure our links in the docs are up to date. It does _not_ test GitHub links, because many of these are generated by our documentation generator, and `linkcheck` doesn't play nicely with the SPA anchors. So, for now, it's only external links and internal documents that are checked.

This runs in series with the actual build. I'm not sure whether it would be better to have this run periodically, instead. Broken links are shown in the `Files changed` tab as "checks" overlayed onto the files.

I also removed references to Doxygen within the `ak.operations` docstrings. This is not something that high-level users need to be aware of, and it makes their docstrings less readable.


jpivarski (approved):
Seeing the "`en/latest`" → "`doc/main`" fix, I did a grep of the state of the repo in this branch and found a few more places to fix it:

* `README.md`: more instances than the one your found
* `.github/ISSUE_TEMPLATE/feature-request.yml`
* `.github/ISSUE_TEMPLATE/documentation.yml`

Those templates can be reconsidered, too. I prefer a very minimal use of issue templates because if someone is faced with 20 boxes to fill in, they'll be more likely to give up.

The `feature-request.yml` template has a list of where to find help, to minimize cases of users submitting feature requests when the feature exists in a different form than they expected. This lists three documentation sites (tutorials, Python docstrings, C++ doxygen) because the first two weren't as well integrated as they are now and the third was more important in the past. These three bullets can be reduced to one.

The `documentation.yml` template makes people actually say what form of documentation they're asking to have improved, in part so that they are aware of all the places to look, but also so that we understand what they're asking to change. Nowadays, this is the documentation website (singular), GitHub README.md, CONTRIBUTING.md (that's documentation, too!), Python docstrings, and well, maybe also C++ doxygen comments.

I went ahead and did the above edits, since I was being so prescriptive, anyway.

I also checked for references to the website that include "`scikit-hep/awkward-1.0`" and there are a lot of them, mostly because they're in LICENSE comments at the top of each file and embedded within SVG images. There's too many of them to add to this PR, and it's super-low priority because GitHub redirects.

There's nothing that needs to be done here; this PR can be merged. I'll let you do it, to counter-review my changes.

agoose77:
> Those templates can be reconsidered, too. I prefer a very minimal use of issue templates because if someone is faced with 20 boxes to fill in, they'll be more likely to give up.

I agree, actually, despite pushing for more granular templates early on.

PR #1990, opened 2022-12-09
ci: move to Pages for docs

agoose77:


jpivarski (approved):
We've been talking about this at length on Slack, and I heartily approve!

In fact, since we can invent our own version tags, let's keep the latest as "`main`", since that's the name of the git branch. The other version tags would be "`2.0`", "`2.1`", ... and those can be literal git branch names, too. Then the explanation of what each documentation set represents can be "the state of the code at the head of branch X". Some people do git-clone Awkward to test things against `main` (a higher proportion than NumPy) and it would be good to clear up the ambiguity of whether "latest" means the latest release or the state of the code in GitHub.

As we discussed, this goes under "`doc`" (NumPy does not have an "s"), so

  * `awkward-array.org/` → non-permanent redirect to `awkward-array.org/doc/main`
  * `awkward-array.org/doc`: space for future version list; we don't need to populate it now (can be 404)
  * `awkward-array.org/doc/main`: generated from the `main` branch, triggered by pushes to `main`
  * `awkward-array.org/doc/2.0`: space for a copy of the documentation that is somehow generated from the `2.0` branch; whether that involves any stored tarballs is an open question. We don't need to populate it now (can be 404)
  * `awkward-array.org/doc/2.1`: etc.

PR #1996, opened 2022-12-12
fix: add `Record.copy`

### LINKED ISSUE #1995: ak.record.Record does not have a `copy` method
It should have one for consistency with the ak.contents.Content subclasses, even though it can't be duck-typed with them (ak.record.Record doesn't have any of the arguments that ak.contents.Content have, and vice-versa).

agoose77:
I haven't added any tests, because we probably need to revisit our `copy` tests anyway (this code has been touched a reasonable amount)

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -1223,3 +1223,6 @@ class Content:
 
     def _fill_none(self, value: Content) -> Content:
         raise ak._errors.wrap_error(NotImplementedError)
+
+    def copy(self) -> Self:
+        raise ak._errors.wrap_error(NotImplementedError)
--- a/src/awkward/record.py
+++ b/src/awkward/record.py
@@ -5,6 +5,7 @@ import copy
 from collections.abc import Iterable
 
 import awkward as ak
+from awkward._util import unset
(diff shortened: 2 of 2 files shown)


jpivarski:
I guess I should have been clearer in the issue. (I wrote it quickly, while trying to not lose track of something else.)

The `Content` subclass's `copy` methods take all the same arguments as the corresponding `__init__`, but with `unset` as the default. They construct a new (shallow copy) instance with only the not-`unset` arguments replaced.

Maybe a better name would have been "copy except for..." but it's already past 2.0.0 and now [it's being used in fastjet](https://github.com/scikit-hep/fastjet/pull/156).

The `Content.copy` doesn't have a top-level protocol because each subclass has a different set of arguments.

agoose77:
@jpivarski yes, right - this wasn't supposed to be up for review yet ... too many PRs open at once :laughing:

ioanaif (approved):
This follows the implementation in `contents` so it looks correct.  However, I am not aware of the `issue` Jim mentions.

agoose77:
Ah good point, I should link that here.

PR #2007, opened 2022-12-14
docs: redirect paths for user-guide

agoose77:
Google et al. still cache links to `awkward-array.org` from its Netlify days. We need to redirect these on GitHub Pages.

This isn't done for AWS, but we can just generate those redirects once and PUT them into the root.


jpivarski (approved):
Approved! (We talked about the motivation for this earlier.)

PR #2011, opened 2022-12-15
docs: remove mention of numexpr

agoose77:
Numexpr support is not public yet, so we hide it for now.

Also, cleanup JAX documentation string.


jpivarski (approved):
Yes, this is good. By removing it from visibility, we'll find out if anyone is actually using it, too (if they're not shy and ask where it went). If there is an outcry to put it back, we can do so right away. If not, we can try to find a way to get Awkward Arrays natively supported by NumExpr or through a mechanism like NEP13.

PR #2012, opened 2022-12-15
fix: support empty record arrays in `ak.to_numpy`

### LINKED ISSUE #1944: `ak.to_numpy()` fails cryptically for `ak.Array([{}])`
### Version of Awkward Array

main

This raises
```pytb
Cell In [10], line 1
----> 1 ak.to_numpy(x)

File ~/Git/awkward/src/awkward/operations/ak_to_numpy.py:39, in to_numpy(array, allow_missing)
      9 """
     10 Converts `array` (many types supported, including all Awkward Arrays and
     11 Records) into a NumPy array, if possible.
   (...)
     33 See also #ak.from_numpy and #ak.to_cupy.
     34 """
     35 with ak._errors.OperationErrorContext(
     36     "ak.to_numpy",
     37     dict(array=array, allow_missing=allow_missing),
     38 ):
---> 39     return ak._util.to_arraylib(numpy, array, allow_missing)

File ~/Git/awkward/src/awkward/_util.py:806, in to_arraylib(module, array, allow_missing)
    803 layout = ak.operations.to_layout(array, allow_record=True, allow_other=True)
    805 if isinstance(layout, (ak.contents.Content, ak.record.Record)):
--> 806     return layout.to_numpy(allow_missing=allow_missing)
    807 else:
    808     return module.asarray(array)

File ~/Git/awkward/src/awkward/contents/content.py:1561, in Content.to_numpy(self, allow_missing)
   1560 def to_numpy(self, allow_missing: bool):
-> 1561     return self._to_numpy(allow_missing)

File ~/Git/awkward/src/awkward/contents/recordarray.py:922, in RecordArray._to_numpy(self, allow_missing)
    917 if any(len(x.shape) != 1 for x in contents):
    918     raise ak._errors.wrap_error(
    919         ValueError(f"cannot convert {self} into np.ndarray")
    920     )
    921 out = self._backend.nplike.empty(
--> 922     contents[0].shape[0],
    923     dtype=[(str(n), x.dtype) for n, x in zip(self.fields, contents)],
    924 )
    925 mask = None
    926 for n, x in zip(self.fields, contents):

IndexError: list index out of range
```

Clearly we're assuming that a `RecordArray` always has contents. This should be a simple fix.

agoose77:
Fixes #1944

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -868,7 +868,7 @@ class RecordArray(Content):
                 ValueError(f"cannot convert {self} into np.ndarray")
             )
         out = self._backend.nplike.empty(
-            contents[0].shape[0],
+            self.length,
             dtype=[(str(n), x.dtype) for n, x in zip(self.fields, contents)],
         )
         mask = None
--- a/tests/test_1944-to-numpy-empty-record-array.py
+++ b/tests/test_1944-to-numpy-empty-record-array.py
@@ -1,0 +1,7 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import awkward as ak
+
+
+def test():
+    assert ak.to_numpy(ak.Array([{}, {}])).tolist() == [(), ()]


jpivarski (approved):
Absolutely! (RecordArray didn't originally have a `length` independent of its `contents`. It was added so that records with zero fields would at least be representable, but there are vestiges like this one in which `contents`, rather than the `length`, is used.)

PR #2019, opened 2022-12-19
fix: ensure Exception if branch evaluates for Awkward type

agoose77:
Our `type_to_form` should never be called on an Awkward type in `from_rdataframe`, but we should probably handle that case anyway to make the code more readable.


agoose77:
Hmm, let me retry this.

src/awkward/_connect/rdataframe/from_rdataframe.py
```diff
@@ -171,6 +171,10 @@ def cpp_fill_function(depth):
                 raise ak._errors.wrap_error(
                     TypeError(f'"{col}" column\'s type "{col_type}" is not supported.')
                 )
+            elif form_str == "awkward type":
+                raise ak._errors.wrap_error(
+                    RuntimeError("this code should not be reached.")
```
jpivarski: One thing: we use `AssertionError` to denote unreachable code.

```suggestion
                    AssertionError("this code should not be reached.")
```

jpivarski (approved):
This looks like a good thing to do. I'd say "let's merge it" with one change (below). I'll commit that and then enable auto-merge.

PR #2020, opened 2022-12-19
refactor!: use exclusively `axis=-1` reduction for `axis=None`

### LINKED ISSUE #1789: `ak.sum([])` raises `IndexError`

agoose77:
This PR reworks #1793. 
- [x] Fixes #1789 
- [x] Removes the `flatten_records` argument to reducers
- [x] General cleanups of reducers

I plan to go back into `_reducers` after working on the nplike mechanism to cleanup the dtype handling. For now, it works, but I'd prefer us to deal exclusively with `np.dtype` objects rather than NumPy's generic types.


agoose77:
So, something I hadn't thought about when designing this is that we effectively can't build a union of a JAX array if it's a tracer — we can, but the tangents are lost, because we perform the copy with a kernel rather than a `jax.numpy.concatenate`-like copy. This means that `axis=None` wouldn't work for JAX functions, whereas this does currently work with an explicit high-level loop. N.b. that the explicit high-level loop is not without its own problems, particularly the amount of boilerplate if we need to handle `mask`, `initial`, etc.

This needs some more thought.

agoose77:
Following a discussion with Jim, it was noted that only the `flatten_records` feature of the reduction mechanism requires us to handle multiple branches. Removing that argument both makes this PR far, far easier, but also brings `axis=None` reduction in line with the `axis != None` cases, for which reducers can't move past records  (which may be overloaded in future).

agoose77:
This needs some tests and a check for `keepdims` ...

src/awkward/_do.py
```diff
@@ -304,61 +304,92 @@ def mergemany(contents: list[Content]) -> Content:
 def reduce(
     layout: Content,
     reducer: ak._reducers.Reducer,
-    axis: int = -1,
+    axis: int | None = -1,
```
jpivarski: Is there a type for `AxisMaybeNone` that could be used here? (And what happens if an `axis` is passed an `np.int64`? We should accept it.)

src/awkward/contents/numpyarray.py
```diff
@@ -1044,6 +1044,17 @@ def _reduce_next(
...
+        # NOTE: this really needs to stay here for now, i.e. can't go in
+        # ak._do (because internal code instantiates Reducer instances)
+        # Ideally we would use the kernel mechanism for this, but JAX arrays are immutable whilst out kernel API assumes
+        # that arrays are mutable and that the inputs/outputs have a common nplike. For JAX we would probably want
+        # both numpy/cupy and jax.numpy which are not directly compatible, so this "hack" will likely remain.
+        if isinstance(self.backend, ak._backends.JaxBackend):
+            from awkward._connect.jax.reducers import get_jax_reducer
+
+            reducer = get_jax_reducer(reducer)
+
```
jpivarski: JAX is very special with regard to reducers. Having it make this much noise here in the codebase doesn't bother me.

(It's only a hack in the sense that Content code should not be asking questions about specific backends, to separate those concerns, but JAX reducers stand out enough as a thing that should get special attention that I wouldn't call this a hack.[^1])

[^1]: Well, maybe in a technical sense. I've been converging on the idea that the word "to hack" means to break levels of abstraction.[^2] For instance, if you use the fact that CPython's `id` function returns C function pointers to `PyObject*` as a number, you are committing a hack because the Python language is supposed to be a different level of abstraction than its implementation. So if we declare that Content implementations are at a layer of abstraction above backend implementations, then doing anything in Content that acts differently for specific backends would technically be a hack. Content and backends/nplikes weren't originally conceived of as different layers of abstraction, but they're growing toward that direction.

[^2]: And so puns are language hacks. The meanings of words is a different level of abstraction from the way that they sound, so if a similar-sounding word is used in place of another to change the meaning of a sentence, it's mixing levels of abstraction. Some things in type theory are even called "puns" when they take advantage of unintended equivalences.

src/awkward/operations/ak_all.py
```diff
@@ -57,60 +56,35 @@ def all(
...
-            array, axis, keepdims, mask_identity, flatten_records, highlevel, behavior
-        )
+        if flatten_records is not unset:
+            raise ak._errors.wrap_error(
+                ValueError(
+                    "`flatten_records` is no longer a supported argument for reducers. "
+                    "Instead, use `ak.ravel(array)` first to remove the record structure "
+                    "and flatten the array."
+                )
+            )
```
jpivarski, line 70: We are hereby committing to a hard-deprecation. The function argument becomes unusable immediately. I think this is okay, considering how obscure the function argument is and how recently 2.0 was released.

tests/test_1490-jax-reducers-combinations.py
```diff
@@ -170,455 +30,85 @@ def func_numpyarray_1(x):
...
-    assert ak.to_list(value_vjp) == value_vjp_jax.tolist()
-    assert ak.to_list(jvp_grad) == jvp_grad_jax.tolist()
-    assert (
-        ak.to_list(vjp_func(value_vjp)[0]) == (vjp_func_jax(value_vjp_jax)[0]).tolist()
-    )
-
+@pytest.mark.parametrize("axis", [0, 1, None])
+@pytest.mark.parametrize("func_ak", [ak.sum, ak.prod, ak.min, ak.max, ak.sum])
+def test_reducer(func_ak, axis):
+    func_jax = getattr(jax.numpy, func_ak.__name__)
```
jpivarski: You seem to be removing a large number of tests from this file, but are you actually just consolidating it with `@pytest.mark.parameterize`? I have not verified that all of the old tests still run as parameters of this one test. Is that the case?

tests/test_0835-datetime-type.py
```diff
@@ -498,12 +498,12 @@ def test_highlevel_min_max_axis_None():
             [
                 np.datetime64("2020-03-27T10:41:11"),
                 np.datetime64("2020-01-27T10:41:11"),
-                np.datetime64("2020-05"),
+                np.datetime64("2020-05-27T10:41:11"),
```
jpivarski, line 501: This had been a test of mixed datetime units; now all the datetimes in this array have the same units and it's no longer a union array.

Okay: reductions of unions with `axis=None` worked before this PR (not any other `axis`, though!) and this PR is removing that capability, but it's good because we're bringing everything in line with the same set of capabilities.

jpivarski (approved):
Great! This has been a long time coming, and I think it's in good shape now. I have comments throughout, but nothing preventing this PR from being merged.

src/awkward/_do.py
agoose77: Oh yes, there is! 

As for `int64`, I've left that out of the types for now. I *think* the way to support that is by literally enumerating an `IntLike` type, but at this stage I just want to get partial coverage down. When we start worrying more about the accuracy of these types, we can look to what the NumPy typing module offers, too.

src/awkward/contents/numpyarray.py
agoose77: i'm not uncomfortable with "HACK" because in my view it is; a proper abstraction would be to define a ...

OK, I'm just going to write the proper abstraction :laughing:

jpivarski: I hope I haven't started you on a large project! We can get there incrementally.

tests/test_1490-jax-reducers-combinations.py
```diff
@@ -170,455 +30,85 @@ def func_numpyarray_1(x):
...
-
-    assert ak.to_list(value_jvp) == value_jvp_jax.tolist()
-    assert ak.to_list(value_vjp) == value_vjp_jax.tolist()
-    assert ak.to_list(jvp_grad) == jvp_grad_jax.tolist()
-    assert (
-        ak.to_list(vjp_func(value_vjp)[0]) == (vjp_func_jax(value_vjp_jax)[0]).tolist()
-    )
-
+@pytest.mark.parametrize("axis", [0, 1, None])
+@pytest.mark.parametrize("func_ak", [ak.sum, ak.prod, ak.min, ak.max, ak.sum])
```
agoose77:
```suggestion
@pytest.mark.parametrize("func_ak", [ak.sum, ak.prod, ak.min, ak.max])
```

agoose77:
@jpivarski could you pass your eye over `Backend` and how I've changed the way that JAX overrides the ufuncs and reducers?

src/awkward/_util.py
```diff
@@ -908,7 +914,7 @@ def visitor(left, right) -> bool:
         elif left.is_option:
             return numpy.array_equal(
                 left.index.data < 0, right.index.data < 0
-            ) and visitor(left.content, right.content)
+            ) and visitor(left.project(), right.project())
```
jpivarski, line 917: On this line, it looks like you've fixed a potentially subtle bug. Or maybe not that subtle; any test with missing values in it at this level would be sensitive to this difference.

jpivarski (approved):
I see: the explicit check for a JAX nplike in the Content code has been moved up into the Reducer code, which is more backendy. I like this separation—better than I was comfortable with the explicit check in the Content code. I felt that that was okay, but this is better.

I also see the typing for `axis` as `Integral`, rather than `int`, and what looks like a good bug-fix below.

All in all, these are improvements and I approve!

src/awkward/_util.py
agoose77: Yes, I actually ran into this, as it was an oversight on my part!

PR #2021, opened 2022-12-19
fix: extend TypeTracerArray with __eq__, __ne__, and __array_ufunc__.

jpivarski:
https://github.com/dask-contrib/dask-awkward/pull/133#issuecomment-1356884239 happened because TypeTracerArray didn't have an `__eq__` method. It was easy to add, along with `__ne__`.

In addressing this issue, I first added an `__array_ufunc__` ([NEP-13](https://numpy.org/neps/nep-0013-ufunc-overrides.html)) because I thought that's what the issue was and I was surprised that it didn't have one. I'm still surprised it doesn't have one. We _could_ implement all of the `__eq__`, `__ne__`, `__gt__`, `__ge__`, `__lt__`, `__le__` methods in one swoop by making TypeTracerArray inherit from `ak._connect.numpy.NDArrayOperatorsMixin`.

What do you think, @agoose77? Actually, I've pretty much talked myself into it. The next commit will do that. It's git, everything's reversible, of course!

### DIFF
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -447,6 +447,18 @@ class TypeTracerArray:
         else:
             raise ak._errors.wrap_error(NotImplementedError(repr(where)))
 
+    def __eq__(self, other):
+        if isinstance(other, numbers.Real):
+            return TypeTracerArray(np.bool_, self._shape)
+        else:
+            return NotImplemented
+
+    def __ne__(self, other):
+        if isinstance(other, numbers.Real):
+            return TypeTracerArray(np.bool_, self._shape)
+        else:
+            return NotImplemented
+
     def __lt__(self, other):
         if isinstance(other, numbers.Real):
             return TypeTracerArray(np.bool_, self._shape)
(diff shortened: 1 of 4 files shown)


agoose77:
This is definitely needed - the question for me is whether I'll be done with #1849 before this is 'needed' (it's already needed, but that's hard to define!)

Ths likelihood is that I'll be replacing this code in the near future. If you think it's needed now, merge! (I'll review anyway)

jpivarski:
"Why not make TypeTracerArray inherit from NDArrayOperatorsMixin?"

  1. It would cause a circular import unless NDArrayOperatorsMixin is moved somewhere more central, such as `ak._util`.
  2. https://github.com/scikit-hep/awkward/blob/9ab248c2dc0ce87a0836aed5d444a7529552d3eb/src/awkward/_typetracer.py#L352 starts failing because somewhere, `len` of a TypeTracerArray is now being called.
  3. Replacing that `where.index` with a loop over `where` to look for the `Ellipsis` results in yet more errors.

At step 3, I stopped looking into it. The refactor made the TypeTracerArray code a little nicer (dropped 6 very short methods), but not a lot, and it seems to be getting into some deep waters that I don't have time for right now. I.e. the cost-benefit is not good enough to keep following this line.

jpivarski:
Yes, it's needed in the short term, and it's fine if it gets refactored. The `ak.where` function triggers it, here:

https://github.com/scikit-hep/awkward/blob/9ab248c2dc0ce87a0836aed5d444a7529552d3eb/src/awkward/operations/ak_where.py#L111

(`(npcondition == 0)` needs to return a TypeTracerArray, not a Python `bool`.) So I'll do a more useful thing and add a test of type-tracer through `ak.where`, so that when you refactor, it will not get lost.

jpivarski:
This is ready to merge (I see your informal assent, above). The test failures are unrelated, but they need to be fixed.

As soon as tests pass, we'll merge.

PR #2025, opened 2022-12-21
ci: drop pages deployment

agoose77:
This PR should remove the GitHub Pages documentation deployment from the CI, and disable version selectors for previews.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -197,9 +197,13 @@ jobs:
     - name: Copy C++ documentation
       run: cp -r awkward-cpp/docs/html/ docs/_static/doxygen
 
+    - name: Enable analytics & version selector
+      if: github.event_name == 'push' || github.event_name == 'release'
+      run: |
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Yes, we should only be generating analytics for the released documentation (`main`, `stable`, `2.0`, ...) and not PR-branch test-deployments. Removing this will make the data analysis easier later, because we won't have to filter on URLs (if that's even possible for the time-series data).

After this goes through, we should spot-check generated documentation to be sure that

```html
<script data-domain="awkward-array.org" defer="defer" src="https://views.scientific-python.org/js/plausible.js"></script>
```

is in the released documentation and not in the PR-branch test-deployments (whenever there's a new one to check, just take a look to be sure).

jpivarski:
I suppose this PR is the first thing that can be tested: the line of HTML is not in http://preview.awkward-array.org.s3-website.us-east-1.amazonaws.com/agoose77/docs-switch-to-aws/, so that's good.

PR #2026, opened 2022-12-21
docs: add sitemap

agoose77:


jpivarski (approved):
I assume that this is not something that the user looks at directly (it's a big XML, right?), but it helps search engines like Google, right?

Anyway, I approve! It's pretty clear that this PR is just turning on a feature that Sphinx provides for us.

agoose77:
Yes, the plan is to use this for Google's search indexing to speed things up

PR #2027, opened 2022-12-21
feat: add data-touch reporting to the type-tracer.

jpivarski:
This came out of conversations with @lgray and @nsmith-: perhaps the type-tracer can determine which input columns have been touched, where

```python
restructured = ak.zip({
    "muons": ak.zip({
        "pt": original["Muon_pt"],
        "phi": original["Muon_phi"],
        "eta": original["Muon_eta"],
        "mass": original["Muon_mass"],
    })
})
```

wouldn't count as "touching" `"Muon_pt"`, `"Muon_phi"`, `"Muon_eta"`, and `"Muon_mass"`, but

```python
restructured.muons.pt
```

would count as touching `"Muon_pt"`. That kind of influence would be hard to track on the high road that dask-awkward takes, but type-tracers take the low road and can pass a label from the `"Muon_pt"` field of `original` to the `"pt"` field of `muons`.

The next commit will add tests, in which I'll see if this can truly be made to work. If it does work, then it could be a single-pass solution to identifying columns for `project_columns` in dask-awkward.

Cc: @douglasdavis and @martindurant


martindurant:
Well this would definitely make things a lot simpler for us. There would still be cases where a function is opaque and we cannot tell (e.g., numba), but not so many. From our point of view, any columns that are still available at the output are all "touched" (meaning we cannot prune them).

jpivarski:
It might be safe to say that anything that's passed as an input to Numba is touched. The first step in preparing an Awkward Array for Numba is to make a NumPy array of pointers to all the NumPy arrays in the Awkward Array, so they'd get instantiated in any lazy system.

Although I suppose if they were projected out by Dask, a different Numba function would get compiled, one without the other columns. Hmmm.

jpivarski:
I _think_ this is everything: type-tracers take some specialized code paths throughout the codebase, to avoid asking them questions they can't answer. In these code paths, we at least have to inform them that their data would be touched in an execution. (We're using type-tracers for more than was originally envisioned, so the information hadn't been tracked until now.)

If we encounter a bug—some column that shouldn't be identified as touched is, or some column that should be isn't—it's likely due to

  * a missing `touch_data`/`touch_shape` call
  * a spurious `touch_data`/`touch_shape` call
  * `touch_data`/`touch_shape` with `recursive=True` that should be `False`
  * or vice-versa

On the plus side, this gives us many knobs to turn to fit the behavior to expectations. Whether something like

```python
array[some:range]
```

should be considered "touching data" or not is a matter of taste. By setting a particular pattern of `touch_data`/`touch_shape` calls, we can make it do exactly what we want.

src/awkward/_typetracer.py
```diff
@@ -198,6 +201,104 @@ def __str__(self):
...
+
+    elif isinstance(layout, ak.contents.UnionArray):
+        assert isinstance(form, ak.forms.UnionForm)
+        layout.tags.data.form_key = form.form_key
+        layout.tags.data.report = report
+        layout.index.data.form_key = form.form_key
+        layout.index.data.report = report
+        for x, y in zip(layout.contents, form.contents):
+            _attach_report(x, y, report)
+
```
agoose77, line 292:
```suggestion
        for x, y in zip(layout.contents, form.contents):
            _attach_report(x, y, report)

    else:
        raise ak._errors.wrap_error(TypeError("this code should never be reached"))
```

agoose77:
For the benefits of public discourse — I've had a limited number of conversations with Doug, Martin, Jim etc. about why this is needed, so I'm starting somewhat from a position of ignorance.

My reading of the purpose of this PR is to identify at the buffer level whether inputs are propagated to outputs. This is done by tagging the buffer with the form key, such that any report found in the final result with the appropriate form key indicates that the data were used.

@jpivarski on the specifics of this PR, does this mean that
```python
x = dak.from_json(...) # has columns foo, bar

foo = x["foo"]
bar = x["bar"]

total = dak.sum(bar)
result = total.compute()
```

would load both "foo" and "bar", despite only "bar" actually being computed? It looks like the reports on each buffer are mutable, rather than operations which touch data producing a new array. And, that would seem like a fairly impassable obstacle; many operations that don't return new arrays e.g. `bar[0]` need to indicate that if they are computed, they touch data.

jpivarski:
Here's a big overview to bring everybody up to speed on various conversations.

----------------

> My reading of the purpose of this PR is to identify at the buffer level whether inputs are propagated to outputs.

That's right: this comes back to the "high-level" and "low-level" tracking that we were talking about at the beginning of the Dask work, which I'll call the "high road" and "low road"[^1] to avoid confusion with Dask's own concept of high-level graphs. There are two ways to describe a computation on Awkward Arrays, one that considers what `ak.*`/slice/ufunc operations happen to ak.Arrays (each of those operations is a node in the graph) and one that considers what kernels/nplike functions happen to one-dimensional buffers (each of those operations is a node in the graph).

The _ability_ to walk this low road is one of the motivations of the Awkward 2 project as a whole. I had thought at the start of the Dask project that the Dask graph would be at this level. From an early prototype, however, we concluded that doing so would make the Dask graph too big. A simple computation made a graph with about 100 nodes; typical data analyses would bring it into the thousands. @martindurant informed me that this would be a performance issue in the Dask machinery, and people do visualize the Dask graphs to understand performance issues: a graph full of kernel calls would not be easy to understand. And besides, most of those sub-graphs would look like most of the other sub-graphs: if you do a lot of some particular kind of slice, there would be a subgraph with, say, dozens of nodes that look like every other subgraph for the same kind of slice. At that level, all that detail would be unnecessary. So the Dask graph goes the high road and that's a good thing.

But there is some information that can only be collected on the low road. Predicting the output data types was one of these. Type-tracers (immutably) collect these on the zeroth pass, while the Dask graph is being built.

This PR adds another ability. Augmented type-tracers can now be run through the graph again as one of the first steps after `.compute()` to collect information about which buffers are touched. It does not build a new graph (with hundreds or thousands of nodes). It starts by getting labels for each input buffer, which are provided by an external process: dask-awkward or possibly `uproot.dask`/`dak.from_parquet`. Then the external process (dask-awkward) runs the labeled type-tracer through the graph _once_ and it mutably reports which labeled buffers are involved in any operations that require access to its data: requests for a `TypeTracerArray`'s `dtype` or `shape` do not require its data, nor do views or view-like slices or reshaping, but computations and element access do require its data.

This is effectively walking through the thousands-of-nodes graph, but it's not constructing such a graph as an object, just virtually walking through it, collecting only the information that we care about.

So for your example, suppose that a dak.Array has a `meta` like

```python
>>> meta = ak.Array(
...     ak.Array([{"foo": 1.1, "bar": [2, 3, 4]}, {"foo": 5.5, "bar": []}]).layout.to_typetracer()
... )
```

As a first step after `.compute()`, something (`uproot.dask`? `dak.from_parquet`?) would label the Form with buffer names derived from the input source (TBranch names or Parquet columns):

```python
>>> column_projection_phase = copy.deepcopy(meta.layout.form)
>>> column_projection_phase.content("foo").form_key = "track-foo"
>>> column_projection_phase.content("bar").form_key = "track-bar-offsets"
>>> column_projection_phase.content("bar").content.form_key = "track-bar-data"
>>> print(column_projection_phase)
{
    "class": "RecordArray",
    "fields": [
        "foo",
        "bar"
    ],
    "contents": [
        {
            "class": "NumpyArray",
            "primitive": "float64",
            "form_key": "track-foo"
        },
        {
            "class": "ListOffsetArray",
            "offsets": "i64",
            "content": {
                "class": "NumpyArray",
                "primitive": "int64",
                "form_key": "track-bar-data"
            },
            "form_key": "track-bar-offsets"
        }
    ]
}
```

And then create a labeled type-tracer and `report`,

```python
>>> take_me_through_the_graph, report = ak._typetracer.typetracer_with_report(
...     column_projection_phase
... )
>>> x = ak.Array(take_me_through_the_graph)
```

and walk the labeled type-tracer through the graph:

```python
>>> foo = x["foo"]
>>> bar = x["bar"]
>>> total = ak.sum(bar)
```

Now the `report` says which of the buffers have been touched:

```python
>>> report.data_touched
['track-bar-offsets', 'track-bar-data']
```

So, for instance, "foo" is not touched because even though it was assigned to a variable, nothing happened to it that required its data. Both the list offsets and the data were needed for "bar" because the reducer needed both as part of its operation.

The reason that this is another pass, rather than the zeroth pass during which the Dask graph is constructed, is because the report-collection is mutable. Dask objects might be involved in several different `.compute()` operations, as well as graph-building operations that don't end in `.compute()`. If we were building this report along with a Dask graph and a user decided to

```python
>>> x
<Array-typetracer type='?? * {foo: float64, bar: var * int64}'>
```

but this operation _is not included_ in the graph that computes `total`, the report for `total.compute()` should not include everything that is touched by the `x.__repr__()` call.

```python
>>> report.data_touched
['track-bar-offsets', 'track-bar-data', 'track-foo']
```

So the report-collection has to happen in a controlled environment.

Alternatively, we could have collected the report immutably, to have every type-tracer instance carry a separate report with all of the input buffers that have been touched to produce that buffer. So, for instance, a type-tracer representing an array that was produced from some inputs but does not contain those inputs would have a tuple of those inputs that it needs to pass on to any new type-tracers that are produced from it. That's a possible implementation, but annotating our codebase with all of the data-touching operations would get more complicated: we'd have to be replacing nodes everywhere we want to declare that they've been touched, rather than calling a less-obtrusive `x.touch_data()` on them.

The reason why we need those annotations at all is because sometimes type-tracers take different code paths than nplikes with eager data. We've tried to keep that at a minimum, but sometimes it was necessary because of the assumptions we want to make about the buffers in the calculation. So on the type-tracer code path, identified by either `not nplike.known_data` or `not nplike.known_shape` (I searched for all such cases and addressed each of them that did not immediately lead to raising an exception), we have to say whether the non-type-tracer code path touches data. Saying so with

```python
x.touch_data()
```

is less obtrusive than

```python
x = x.touched_data()
```

where we could get lost in replacing symbols with the same name. (This operation _wants_ to be implemented mutably, and we can either take advantage of the mutability of a new type that we made up, `TypeTracerReport`, or the mutability of the Python namespace scope in a function.)

[^1]: Everybody gets to Scotand eventually.

src/awkward/_typetracer.py
```diff
@@ -241,9 +342,19 @@ def from_array(cls: TTypeTracerArray, array, dtype=None) -> TTypeTracerArray:
...
+        return cls(dtype, array.shape, form_key, report)
+
+    def __init__(self, dtype, shape=None, form_key=None, report=None):
+        self.form_key = form_key
+        self.report = report
 
-    def __init__(self, dtype, shape=None):
         self._dtype = np.dtype(dtype)
         self.shape = shape
```
agoose77, line 363: This would mean that creating a `TypeTracerArray` always involves touching its shape do we want this?

agoose77: Also, should `TypeTracerArray.__repr__` not also trigger this touching? Or are we only considering "programmatic" APIs?

src/awkward/contents/listoffsetarray.py
```diff
@@ -153,6 +165,10 @@ def _repr(self, indent, pre, post):
         return "".join(out)
 
     def to_ListOffsetArray64(self, start_at_zero=False):
+        if not self._backend.nplike.known_data:
```
agoose77: Does passing our `offsets` untouched to another layout counts as "touch"ing it?

I'm a bit confused on that point; I think the answer should be no; we don't need to know the values, so it doesn't require touching data. 

If this reasoning holds, then we should move these `touch_` calls to the early return on L178,  and the access to offsets on L182. L187 doesn't require us to know the values of these arrays, so no touching is required.

src/awkward/contents/regulararray.py
```diff
@@ -856,6 +865,8 @@ def _combinations(self, n, replacement, recordlookup, parameters, axis, depth):
                 tocarry.append(ptr)
                 if self._backend.nplike.known_data:
                     tocarryraw[i] = ptr.ptr
+                else:
```
agoose77: I don't think this line should be here - `ptr` is not referring to any external buffers.

Also, shouldn't the various `Content._carry` be touching data?

src/awkward/contents/bytemaskedarray.py
```diff
@@ -290,6 +304,7 @@ def _getitem_at(self, where):
 
     def _getitem_range(self, where):
         if not self._backend.nplike.known_shape:
+            self._touch_shape(recursive=False)
```
agoose77, line 307: My understanding is that this `touch_shape` is used because we need to know `self.length` (which tests the length of the mask array) to know our length. Is this so? If so, why not put this in the getter for `self.length`? 

My guess is for performance reasons /  because we don't well-distinguish between L3 and L4?

src/awkward/contents/unmaskedarray.py
```diff
@@ -157,12 +165,14 @@ def _getitem_nothing(self):
 
     def _getitem_at(self, where):
         if not self._backend.nplike.known_data:
+            self._touch_data(recursive=False)
```
agoose77, line 168: Again,  is this here just for posterity?

agoose77 (approved):

agoose77:
I'll fix the merge conflicts now.

src/awkward/_typetracer.py
jpivarski: Right! But this kind of error (our programming error, rather than a user's/downstream developer's usage error) is `AssertionError`. I implemented this in d66e7ccf6f792d0316d2c243e69704778c633dce.

src/awkward/_typetracer.py
jpivarski: Good point. `shape_touched` isn't very useful if it's always on. If you're assigning a shape to an array (which I think can only happen during construction—any other time it would be a `reshape` call...), then you already know it; you don't need to learn it.

The `touch_shape` call is still in the `shape` property, but not in the setter anymore: e8d39df2e3e51b02544b76b24209b7f311256078.

jpivarski: I don't think `__repr__` should count as a touch. It's not something that any ak.Array or even any layout interaction does, and it can't be represented by any Dask nodes. Whether it counts or not, it wouldn't have an impact on any foreseeable applications.

src/awkward/contents/listoffsetarray.py
jpivarski: Just passing along the `offsets` doesn't count as "touching" them. This kind of subtlety is what motivated this low-level implementation, so that we can make free choices at this level that give us the desired effects.

`ListOffsetArray.to_ListOffsetArray64` makes some value-dependent choices, and the decision to touch or not touch can't depend on values. (Note: the return type does not depend on values—that much is okay.)

I guess there's one case that we can determine from types and metadata only in which the `offsets` could go untouched:

  * if the `ListOffsetArray` already has 64-bit `offsets`
  * if `start_at_zero=False`

The case in which `start_at_zero=True` and `offsets[0] == 0` can't go untouched because we can't know that with a type-tracer.

I'm not going to try threading the touch-logic through the implementation, but I'll put this special case in the touch-logic block on lines 168‒170. It's in 98d0cce5cdc7740912d67ce94e872e55aee75ab2.

src/awkward/_typetracer.py
agoose77: Hmm, I think I prefer to keep `AssertionError` exclusively for the `assert` statements, which can be optimised out of the bytecode. In any other case, it's not an assertion but rather an invalid value / type / condition. 

The message that I wrote doesn't help elucidate that understanding though :laughing: I should have written 
> Only the ... layouts are supported 

or something like that.

src/awkward/contents/listoffsetarray.py
agoose77: > Just passing along the `offsets` doesn't count as "touching" them.

This is what's giving me a little confusion. Imagine that we create a `ListOffsetArray`, and pass it to `ak.with_parameter`. Ignoring the actual implementation, it's easy to imagine that this routine doesn't need to look at any buffers, it just touches the parameter metadata. We still want this array to consider those leaf arrays important, right? Equally, we don't want to touch anything before it's used because then this mechanism becomes meaningless.

> I'll put this special case in the touch-logic block on lines 168‒170. It's in https://github.com/scikit-hep/awkward/commit/98d0cce5cdc7740912d67ce94e872e55aee75ab2.

Fab, that is what I was thinking

src/awkward/contents/regulararray.py
jpivarski: > I don't think this line should be here - `ptr` is not referring to any external buffers.

You're right: I fixed all three cases in c9ba811f1c0cd47579c9ab3074360386f8ec9f1e.

> Also, shouldn't the various `Content._carry` be touching data?

It does, but it does so without special annotation because there's no `if nplike.known_data` hiding it from the type-tracer. For instance, in

https://github.com/scikit-hep/awkward/blob/b695e2fe78fbdb0a810a06a92e9fdd3aa3931d12/src/awkward/contents/listarray.py#L272-L283

the `self._starts[carry.data]` calls `__getitem__` on the `TypeTracerArray`, which finds its way into this case:

https://github.com/scikit-hep/awkward/blob/b695e2fe78fbdb0a810a06a92e9fdd3aa3931d12/src/awkward/_typetracer.py#L533-L540

which touches it.

The special `touch_data`/`touch_shape` annotations are only needed if the code that _would have called_ `TypeTracerArray`'s methods or `TypeTracer`'s functions doesn't get called because type-tracers are handled on a special code path. In this PR, I searched through all the places where `known_data` or `known_shape` are invoked and added annotations to the type-tracer code paths only.

src/awkward/contents/bytemaskedarray.py
jpivarski: No (it's not a performance thing), it's because it's already touched, since this is not a specialized type-tracer code path. In the case of `ByteMaskedArray.length` (and any other `Content.length`, but taking this as an example),

https://github.com/scikit-hep/awkward/blob/b695e2fe78fbdb0a810a06a92e9fdd3aa3931d12/src/awkward/contents/bytemaskedarray.py#L178-L180

calls

https://github.com/scikit-hep/awkward/blob/b695e2fe78fbdb0a810a06a92e9fdd3aa3931d12/src/awkward/index.py#L123-L125

which calls

https://github.com/scikit-hep/awkward/blob/b695e2fe78fbdb0a810a06a92e9fdd3aa3931d12/src/awkward/_typetracer.py#L372-L375

so the shape gets touched.

src/awkward/contents/regulararray.py
agoose77: Right, you're absolutely right.

src/awkward/contents/bytemaskedarray.py
agoose77: Yes, I clean-forgot about the type-tracer overloads whilst writing this comment

src/awkward/contents/unmaskedarray.py
jpivarski: It was fulfilling a common pattern. If `UnmaskedArray` had any `Indexes` or other buffers directly attached to it (it doesn't), then this would be necessary to touch them before descending into its `content` (and touching whatever's there).

It's a formality.

https://github.com/scikit-hep/awkward/blob/b695e2fe78fbdb0a810a06a92e9fdd3aa3931d12/src/awkward/contents/unmaskedarray.py#L92-L94

with `recursive=False` does nothing—you're right—but it's good to treat all `Content` subclasses the same way, even if it means calling do-nothing methods from time to time.

jpivarski: I think the answer to your question is "yes." It's "for posterity," in the (highly unlikely) possibility that `UnmaskedArray` might acquire an `Index` someday. If it did and we didn't treat the code in a symmetric way like this, it would be hard to find all the missing calls.

src/awkward/_typetracer.py
jpivarski: Hmmm. This is the pattern we use everywhere: the else block is essentially an assertion, and I want an easy message for users about which errors are their fault and which are our fault.[^1] What I usually say is that `ValueErrors` and `TypeErrors` are complaining about user-specified values and types—change the input and try again—while `AssertionErrors` and `RuntimeErrors` are internal inconsistencies in our codebase—file a bug report.

The `assert` statement can be optimized out: you can run Python in a way that skips over these statements. It would be an antipattern to write

```python
if some_bad_condition():
    raise AssertionError("helpful message")
```

because this can't be optimized out. However,

```python
else:
    raise AssertionError("helpful message")
```

isn't an opportunity for optimization because there's no extra statement in the way of normal code to skip. (I'm assuming that adding an `else` clause to an `elif` chain does not change its performance in any significant way: the `else` comes at the _end_. There's going to be a byte-code jump one way or another. Most importantly, there's no `some_bad_condition()` to evaluate, which could be arbitrarily expensive.)

I'd like to keep using `AssertionError` at the end of these `elif` chains, in the block that you're never supposed to reach. But if we decide differently, there would be a lot of them to change.

[^1]: Actually, I'd like that kind of "blame" to become a language feature. When programming-in-the-large, there's an important distinction to be made: which package is responsible for which errors. Some errors are sanity checks on the user's input (not just things a type-checker would catch, but also values), others are "can't compute for some detailed reason," others are "can't get what I need from some other library/network connection," while others are "I don't know how we got here, so it must be a bug in this package." All programming languages that I know of have only one kind of "`raise`" and the exception classes are (somewhat) informative about what the error means but not informative about what to do about it: change inputs and try again? file a bug report for this library? file a bug report for a dependent library? This kind of feature would only be useful if all libraries had access to the same mechanism, and therefore it would need to be a language feature, not a library.

src/awkward/contents/listoffsetarray.py
jpivarski: The bottom line is whether Dask needs to go fetch data to populate the buffer at execution-time. Users (Coffea in particular) want that to happen as little as possible. To put your example in code,

```python
original_data = some_data_source.download_stuff("htts://source.cern")

a_new_array = original_data["list_offset_array_3"]

array_with_parameter = ak.with_parameter(a_new_array, "some_parameter", "some_value")

actually_compute_something = function(original_data["another_field"])

result = ak.zip({"one": array_with_parameter, "two": actually_compute_something})
```

If this is the entire Dask graph (`result.compute()`), my thinking was that we don't _want_ to download the data with field `"list_offset_array_3"` because we never do anything with it. But you bring up a good point that if `"list_offset_array_3"` is removed from the Dask graph, this code will fail in the executor. I hadn't considered that.

One way out is to say that the set of touched input buffers are not removed from the Dask graph: the Dask graph is untouched (_not_ what we were planning!) and instead, the unused buffers are filled with dummy data. NumPy can make very lightweight dummy data:

```python
>>> dummy = np.broadcast_to(np.float64(0), 100)
>>> dummy
array([0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.,
       0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
>>> dummy.shape, dummy.strides
((100,), (0,))
```

The buffer associated with the above only has 8 bytes, not 800.

_On the other hand,_ maybe if somebody makes a dask-awkward array like `result`

```python
result = ak.zip({"one": array_with_parameter, "two": actually_compute_something})
```

and then ask for

```python
materialized = result.compute()
```

then they want `materialized["one"]` to contain downloaded data, not a dummy placeholder. They wrote the above code, deliberately, to get `"list_offset_array_3"` as `"one"` with a parameter. In this interpretation (which I'm thinking is the right one, now that I'm going through it), `"list_offset_array_3"` _should_ be included in the Dask graph and _should_ be downloaded.

The way that dask-awkward can get that is by doing this in the last step of report-collecting:

```python
result.meta.layout._touch_data(recursive=True)
```

where `result.meta.layout` is the `Content` containing `TypeTracerArrays`. Everything that makes its way into the final `result` should be touched because the user is explicitly asking for it. If there's another `"list_offset_array_2"` field in the `original_data`, that would be excluded because it doesn't get to the final `result`.

If we do things that way, `"list_offset_array_3"` does not get touched in the `ak.with_parameter` call, but does get touched because it's in the output. If there was another line in the code,

```python
unused = ak.with_parameter(original_data["list_offset_array_2"], "x", "y")
```

this `unused` wouldn't be part of the Dask graph if it's not added to the `result` that gets computed, and so it's important that `ak.with_parameter` does not, by itself, touch data.

Coffea users would have to be careful:

```python
nanoaod = uproot.dask("root://source.cern")
nanoevents = ak.zip({
    "muons": ak.zip({
        "pt": nanoaod["Muon_pt"],
        "phi": nanoaod["Muon_phi"],
        "eta": nanoaod["Muon_eta"],
        "charge": nanoaod["Muon_charge"],
    }),
})

materialized = nanoevents.muons.compute()
```

would cause all muon fields to be downloaded while

```python
materialized = nanoevents.muons[nanoevents.pt > 10].eta
```

would only cause `"Muon_pt"` and `"Muon_eta"` to be downloaded. _Not_ being specific about what you want causes _more_ to be downloaded.

Cc: @douglasdavis, @martindurant, @lgray, and @nsmith-

jpivarski: The above doesn't imply any changes in how this PR is implemented, just how it is used. I think the PR is mergeable in its current state. Do you agree, @agoose77?

jpivarski: Well, @agoose77 brought up another few cases that make it clear we'll need to do both of the things described above: we'll need dummy buffers _and_ dask-awkward will need to recursively touch the final `result` (the array that gets `.compute()`d).

**Here's case 1:**

```python
original = go_get_data()

zipped = ak.zip({"x": original["field1"], "y": original["field2"], "z": original["field3"]})

result = zipped.x + zipped.y

materialized = result.compute()
```

When dask-awkward gathers a report, it would find that only `"field1"` and `"field2"` have been touched, and that's good because these are the only buffers needed to produce the final `result`. However, if `"field3"` is excluded from the Dask graph, executors would fail to compute the `ak.zip`. The `"field3"` data is not needed, but the metadata is.

**Here's case 2:**

```python
original = go_get_data()   # type: list(X)

result = ak.num(original)

materialized = result.compute()
```

When dask-awkward gathers a report, it would find that only the `offsets` buffer has been touched, not the `content` of the top-level list node, and that's good because the `content` can include anything, such as a large record array, and none of that is in the final `result`. However, Awkward Array can't build a list array node without some kind of `content`.

Both of the above problems can be solved by using the set of touched buffers to determine which buffers need to be downloaded/read from the source, with the others filled in with a dummy. As I said above, NumPy is capable of making very lightweight dummies with arbitrary `shape` (to satisfy metadata constraints). Using the report this way, however, is entirely different from the dask-awkward implementation that we've been pursuing so far: this leaves the Dask graph untouched (i.e. no `project_columns`).

This needs to be done _in addition_ to the recursive-touch described above, because **there's still case 3:**

```python
original = go_get_data()

result = ak.zip({"xy": original["field1"] + original["field2"], "z": original["field3"]})

materialized = result.compute()
```

Unless dask-awkward explicitly calls `result.meta.layout._touch_data(recursive=True)`, the `materialized` result would not include `"field3"`, and it needs to include `"field3"` because a user might subsequently access

```python
materialized.z
```

and expect to not get a dummy buffer.

So unfortunately, this approach would be a big left-turn: different from what you've been developing so far. However, I think it solves all of the problems.

Note that dummy buffers aren't going to cause any performance issues, calculating lots of zeros for no reason. **For example, in case 4:**

```python
original = go_get_data()

zipped = ak.zip({"x": original["field1"], "y": original["field2"], "z": original["field3"]})

unused = ackermann(zipped.z)

result = zipped.x + zipped.y

materialized = result.compute()
```

The `"field3"` buffer would be a dummy in the Dask executor, but the Ackermann function would _not_ be executed on it because it's not part of the Dask graph represented by `result`. When you do the report-generation, you're walking over the operations defined in the Dask graph only.

------------------

This will likely affect all users of dask-awkward, such as `uproot.dask` (cc @kkothari2001), since they'll need to do something different from implementing `project_columns` to reduce the number of array-reads. They'll need to reduce the number of array-reads and replace them with dummy buffers. Somebody (Awkward? dask-awkward?) should provide a standard function for this so that it doesn't get reinvented multiple times.

lgray: @jpivarski thanks for the detail here and I think the way you've laid out where `events.Muon.compute()` would get the whole muon object and `events.Muon[events.Muon.pt > 10].eta` would only get `Muon_pt` and `Muon_eta` in the underlying nanoaod file is largely expected behavior among our users (as I understand from seeing usage patterns). It (largely) preserves the lazy interface to reading data that was there before using uproot4/awkward1.

A problem arises when we start thinking about users using histograms and tracing programs.
```python3
def processor_proxy(dx):
    h = hist.Hist(hist.axis.Regular(bins=10, start=0, stop=1, name="x"))
    h.fill(x = dx.compute())
    return h

x = ak.Array([[1.0, 2.0], [], [3.0, 4.0, 5.0]]) 
dx = dak.from_awkward(x)
# these two lines get replaced with uproot.dask in something realistic, but you get the picture...
h = processor(proxy(dx))
```

Boiled down to its essentials, this will be the most common initial pattern that users encounter with coffea 0.8 and dask awkward. Most users will have some histograms interspersed with data access, and I'm not sure from the discussion above what will happen in "tracing mode" (when, say, we are only trying to figure out what columns are desired rather than doing any real processing) when a concrete array is required and no concrete data is actually rendered (like what we would do as a preprocessing step for using spark, since you must define the columns you want to read before running if you want to be efficient in reading).

In the fullness of time `h` becomes a `dask_histogram` and the dask task graph becomes much smaller and less branchy (as the above would need to be replicated over each chunk of a dataset as we do now, largely to preserve interface). Once nothing is materialized by the user this this becomes less of an issue, I think. Dask histograms and spark are a different conversation entirely, though. In dask or parsl this will all make much more sense.

jpivarski: Yeah, the `.compute()`s will have to be added carefully.

With all of the above implemented, the following would download everything, but only histogram one column:

```python
nanoaod = uproot.dask("root://source.cern")

h = hist.Hist(hist.axis.Regular(bins=10, start=0, stop=1, name="pt"))

h.fill(pt = nanoaod.compute()["Muon_pt"])
```

but the following would only download one column and histogram it:

```python
h.fill(pt = nanoaod["Muon_pt"].compute())
```

Awkward 1 VirtualArrays tried to deduce what needed to be read, sometimes incorrectly, but now that `.compute()` is an explicit step, a user can use it wrong.

The same is true if it went through a restructuring step:

```python
nanoaod = uproot.dask("root://source.cern")

nanoevents = ak.zip({
    "muons": zk.zip({
        "pt": nanoaod["Muon_pt"],
        "phi": nanoaod["Muon_phi"],
        "eta": nanoaod["Muon_eta"],
        "charge": nanoaod["Muon_charge"],
    }),
    "electrons": ...,
})

h = hist.Hist(hist.axis.Regular(bins=10, start=0, stop=1, name="pt"))

h.fill(pt = nanoevents.compute().muons.pt)
```

downloads everything but only uses `"Muon_pt"`,

```python
h.fill(pt = nanoevents.muons.compute().pt)
```

downloads all of the muon data but only uses `"Muon_pt"`, and

```python
h.fill(pt = nanoevents.muons.pt.compute())
```

downloads only the `"Muon_pt"`. Scripts that upgrade to Coffea 0.8 will need to have "`.compute()`" added in various places anyway, but it will be important to add them in the right places!

Maybe dask-histogram would further simplify this:

```python
h = dask_hist.Hist(hist.axis.Regular(bins=10, start=0, stop=1, name="pt"))

h.fill(pt = nanoevents.muons.pt)

h.compute()
```

The histogram-filling would be part of the Dask graph; following that back, only the `nanoevents.muons.pt` is involved in histogram-filling, and the other buffers are untouched, so this would do the right thing with the advantage that the "`.compute()`" is only added to the final histogram, which may be easier to reason about.

lgray: Hi Jim, that's not quite the concern!

It is very easy to tell users to put any `compute()` they may want at the end of a data-access statement as a matter of best practice, or to make a migration script (but this has a lot of possible misinterpretations of user code, so maybe it is best to tell people rather than to generate code that functions strangely compared to user's intent).

Mostly I would like the following hypothetical code to complete without error and tell me what data was touched:
```python3
proc = SomeProcessor(args)

events = NanoEventsFactory.from_root("/some/file.root", schemaclass=SomeSchema) 
# the above will use uproot.dask or dak.from_awkward, or dak.from_parquet
tracer = dak.typetracer_from_form(events._form)
proc.process(tracer)
necessary_columns  = tracer.touched_data()
```

Regardless of what histograms or other data products are filled within that Processor's `process()`.

jpivarski: For debugging what was touched? You could use the same `ak._typetracer.typetracer_with_report` function that dask-awkward will be using. We'll be making that interface public at some point, but first we want to find out how much of it is going to be needed, so that we can make the public API as small as possible.

After the dust settled, the final way to use it is:

```python
result = proc.process(events)  # Coffea, ending with a dak.Array

labeled_form = copy.deepcopy(result._meta.layout.form)

labeled_form.content.form_key = ...   # label each of the nodes

new_typetracer, report = ak._typetracer.typetracer_with_report(labeled_form)

final = result._graph.run_through_graph_with(ak.Array(new_typetracer))

final._meta.layout._touch_data(recursive=True)  # important: all outputs should be labeled

print(report.data_touched)
```

  * You have to label the nodes in `labeled_form` because the buffers are the values of a dict and don't know their keys. But here, you get to specify what's important to you.
  * I don't actually know the Dask command to `run_through_graph_with` a new type-tracer, but I know that there's a mechanism for this because it's what's being done right now.
  * The step of recursively touching all the data in the final computed array is important.

But then the `report.data_touched` list contains the "`necessary_columns`" that you're asking about, a subset of the labels you assigned in the `labeled_form` step.

Is that what you're asking?

lgray: Not only debugging what was touched, in particular for those using spark we have to pre-determine the arrays to load since spark requires a static schema when talking to root. Therefore we will need to trace the Processor once per dataset to determine the dataset schema to-be read to define the spark schema when reading data from uproot.

Here's an example, fully exploded spark workflow using spark3 and coffea 0.8 PR where you can why I need to define the dataset's schema ahead of time. It's nice since it uses mapInArrow everywhere but that does not rid us of the schema issue. I'm happy to field ideas for removing it though!:
```python3
import pyspark
from pyarrow.util import guid
import os
import os.path as osp
from coffea.processor.spark.detail import (
    _spark_initialize,
    _spark_stop,
)

spark_config = (
    pyspark.sql.SparkSession.builder.appName("spark-executor-test-%s" % guid())
    .master("local[*]")
    .config("spark.sql.execution.arrow.enabled", "true")
    .config("spark.driver.host", "127.0.0.1")
    .config("spark.driver.bindAddress", "127.0.0.1")
    .config("spark.executor.x509proxyname", "x509_u12409")
    .config("spark.sql.execution.arrow.maxRecordsPerBatch", 200000)
)

spark = _spark_initialize(
    config=spark_config, log_level="ERROR", spark_progress=False
)

filelist = {
    "ZJets": {
        "files": ["file:" + osp.join(os.getcwd(), "tests/samples/nano_dy.root")],
        "treename": "Events",
    },
    "Data": {
        "files": [
            "file:" + osp.join(os.getcwd(), "tests/samples/nano_dimuon.root")
        ],
        "treename": "Events",
    },
}

columns = ["dataset", "files", "treename"]
reshaped = []
for key, item in filelist.items():
    reshaped.append((key, *(val for val in item.values())))

df = spark.createDataFrame(reshaped, columns)

def explode_dataset(iterator):
    import pyarrow as pa
    import awkward as ak
    for batch in iterator:
        array = ak.from_arrow(batch)
        expander = ak.zeros_like(array.files[0], dtype=int)
        array = ak.zip({"dataset": array.dataset[expander], "file": array.files[0], "treename": array.treename[expander]}, depth_limit=1)
        fields = ak.fields(array)
        yield pa.RecordBatch.from_arrays([array[field] for field in fields], fields)

from pyspark.sql.types import StructType, StructField, StringType, BinaryType, IntegerType, LongType, FloatType, ArrayType

outSchema = StructType([       
    StructField('dataset', StringType(), True),
    StructField('file', StringType(), True),
    StructField('treename', StringType(), True)
])

df = df.mapInArrow(explode_dataset, outSchema)

columns = ["Muon_pt", "Muon_eta", "Muon_phi", "Muon_mass", "Muon_charge", "nMuon"]

def datasetdef_to_columns(iterator):    
    import pyarrow as pa
    import awkward as ak
    import numpy as np
    import uproot
    for batch in iterator:
        dsinfo = ak.from_arrow(batch)
        for i in range(len(batch)):
            with uproot.open({dsinfo.file[i]: dsinfo.treename[i]}) as t_in:
                ak_arrays = t_in.arrays(columns)
                ak_arrays["nMuon"] = ak.values_astype(ak_arrays["nMuon"], np.int64)
                expander = ak.zeros_like(len(ak_arrays)*[0], dtype=int)
                ak_arrays["dataset"] = dsinfo.dataset[expander]
                pa_arrays = ak.to_arrow_table(ak_arrays, list_to32=True, string_to32=True, extensionarray=False).to_batches()
                for out_batch in pa_arrays:
                    yield out_batch

columnTypes = []
for column in columns:
    if column == "nMuon":
        columnTypes.append(StructField(column, LongType(), False))
    elif column == "Muon_charge":
        columnTypes.append(StructField(column, ArrayType(IntegerType()), False))
    else:
        columnTypes.append(StructField(column, ArrayType(FloatType()), False))
    
columnsSchema = StructType(columnTypes + [StructField('dataset', StringType(), False)])

df = df.mapInArrow(datasetdef_to_columns, columnsSchema)

from coffea.nanoevents import NanoEventsFactory, NanoAODSchema
from coffea.processor.test_items import NanoEventsProcessor
from coffea.nanoevents.mapping import SimplePreloadedColumnSource

import lz4.frame
import pickle

nano_schema = NanoAODSchema
lz4_clevel = 1
processor_instance = NanoEventsProcessor()

def columns_to_histograms(iterator):
    global processor_instance, lz4_clevel, nano_schema

    import awkward as ak
    import pyarrow as pa

    for batch in iterator:
        arrays = ak.from_arrow(batch)

        fields_no_meta = ak.fields(arrays)
        fields_no_meta.remove("dataset")
        arrays_no_meta_no_none = {field: ak.fill_none(arrays[field],0) for field in fields_no_meta}
        
        src = SimplePreloadedColumnSource(arrays_no_meta_no_none, None, len(arrays), object_path='/Events')

        events = NanoEventsFactory \
            .from_preloaded(src, metadata={'dataset': arrays.dataset[0]}, schemaclass=nano_schema) \
            .events()  

        vals = processor_instance.process(events)

        valsblob = lz4.frame.compress(pickle.dumps(vals), compression_level=lz4_clevel)
        
        out_batches = ak.to_arrow_table(valsblob, bytestring_to32=True, extensionarray=False).to_batches()        
        for out_batch in out_batches:
            yield out_batch

histoSchema = StructType([
    StructField('histos',BinaryType(),False)
])

df = df.mapInArrow(columns_to_histograms, histoSchema)

df.show()
```

jpivarski: Okay, so you do need to use this outside of dask-awkward: there will be two direct users of it, dask-awkward and Coffea. That will influence the design of the public API, after we see how this works in practice.

Assuming that you have a Processor function named `process`, which takes an input `nanoevents` with known Form `nanoevents_form` and produces results through histogram-filling side-effects, this should do it:

```python
labeled_form = copy.deepcopy(nanoevents_form)

labeled_form.content.form_key = ...   # label each of the nodes

nanoevents_tt, report = ak._typetracer.typetracer_with_report(labeled_form)

process(nanoevents_tt)

print(report.data_touched)
```

Unlike the Dask case, there's no final `materialized` array that users might do subsequent operations on and be surprised that it lacks or has dummy values for some field that they asked for but didn't touch in the asking. So there's no need to call `_touch_data(recursive=True)` on a "final result."

If the histogram functions try to get data from the `nanoevents_tt` by converting the arrays into NumPy arrays (e.g. what hist does), the `TypeTracerArray` would complain with

```
AssertionError: bug in Awkward Array: attempt to convert TypeTracerArray into a concrete array
```

because we were assuming that only we would be using `TypeTracerArrays` directly; dask-awkward doesn't try to coerce it into a NumPy array, it just puts the `ak.to_numpy` operation in the graph.

If you have control over the histogram `fill` function (Coffea histograms, wrapped boost-histogram/hist, or some Spark histogram), you could replace the

```python
ak.to_numpy(awkward_data_to_histogram)  # or
np.asarray(awkward_data_to_histogram)
```

with

```python
awkward_data_to_histogram.layout._touch_data(recursive=True)
```

Otherwise, we'll need to make `TypeTracerArray.__array__` not raise an error, call `_touch_data` on its own, and return a dummy array. Not having this assertion removes one of the guardrails in our own development, but most of that development is done now, so that would be okay.

PR #2029, opened 2022-12-22
fix: replace protocol with direct subclass

### LINKED ISSUE #2028: Checking isinstance of a Protocol is slow
### Version of Awkward Array

2.0.2

In a profiling run of a coffea processor with awkward2 (eager mode) I found that 20% of the time was spent in the following line:
https://github.com/scikit-hep/awkward/blob/7e6f504c3cb0310cdbe0be7b5d662722ee73aaa7/src/awkward/contents/content.py#L94
image
This instance check would normally be very fast but I suspect because the type is a `@runtime_checkable` protocol, it is doing more work.
https://github.com/scikit-hep/awkward/blob/7e6f504c3cb0310cdbe0be7b5d662722ee73aaa7/src/awkward/_backends.py#L42-L45

Perhaps there is a way to have it first check the class `__mro__` and then fall back to the protocol?

If this time is removed from the profile, the remaining time is in line with what I get running the same processor in awkward 1.

agoose77:
Python typing protocols make it possible to enforce duck-typing at type-check time, i.e. ensure that an object has the appropriate methods and attributes rather than whether it inherits from another class. To partially check this at runtime there is a `runtime_checkable` decorator that simply test whether a given object implements the same methods (ignoring their types). 

However, this apparently is much slower than I'd anticipated #2028. This issue is only a problem because we need to sanitise the user inputs to our layouts. Otherwise, once `mypy` is fully covered through the code-base, we could just remove our runtime input validation in most cases. Given that we want to provide helpful user errors, we need runtime checks to work. We could fast-path this with 
```python
def isinstance_(obj, type_) -> bool:
    return type.__instancecheck__(type_, obj) or type_.__instancecheck__(obj)  
```
but in this case, we don't truly need to duck-type backends; they're a limited abstraction that we can forgo the flexibility of protocols.

Fixes #2028

### DIFF
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 
 import awkward_cpp
 
@@ -18,18 +18,7 @@ from awkward._nplikes import (
     nplike_of,
 )
 from awkward._typetracer import NoKernel, TypeTracer
-from awkward.typing import (
-    Any,
-    Callable,
-    Final,
-    Protocol,
-    Self,
-    Tuple,
-    TypeAlias,
-    TypeVar,
-    Unpack,
-    runtime_checkable,
-)
+from awkward.typing import Callable, Final, Tuple, TypeAlias, TypeVar, Unpack
 
 np = NumpyMetadata.instance()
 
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
I appreciate what the typing is expressing about constraints here, but as @nsmith- showed, it's adding too much time.

We talked on Zoom about this. Since we are requiring that the `backend` is one of (our own) four backend objects, the test could be a function in `ak._util`:

```python
def is_backend(backend):
    return (
        backend is ak._backends.NumpyBackend
        or backend is ak._backends.CupyBackend
        or backend is ak._backends.JaxBackend
        or backend is ak._backends.TypeTracerBackend
    )
```

which is a centralized place to update if we ever add a fifth backend. That could be even faster than a standard `isinstance` check. (I'm describing an `is` check on each possibility, rather than a `backend in (...)` to even avoid `__eq__` checks.)

But that's probably not necessary; this PR solves the immediate performance issue as-is.

PR #2030, opened 2022-12-22
ci: fix flake8 warning

agoose77:
I'm not sure why this is needed; I can't reproduce locally, and it should be disabled by default. pre-commit.ci seems to complain, though.

As this is breaking PR workflows, and touches .flake8 only, I'm going to merge when it's successful.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,14 +1,13 @@
-
 [flake8]
 extend-select = C,B,B9,T,AK1
-extend-ignore = E203,E501,B950,E266
+# B905 should be removed once python>=3.10
+extend-ignore = E203,E501,B950,E266,B905
 max-complexity = 100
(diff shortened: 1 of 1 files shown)


agoose77:
This only touches flake8 configuration, and those tests have passed. Merging.

.flake8
```diff
@@ -1,14 +1,13 @@
-
 [flake8]
 extend-select = C,B,B9,T,AK1
-extend-ignore = E203,E501,B950,E266
+# B905 should be removed once python>=3.10
+extend-ignore = E203,E501,B950,E266,B905
 max-complexity = 100
-exclude = studies, pybind11, rapidjson, docs-*, src/awkward/_typeparser/generated_parser.py, awkward/_typeparser/generated_parser.py
+exclude = studies, pybind11, rapidjson, docs-*, src/awkward/_typeparser/generated_parser.py
```
jpivarski, line 6: You're right: it's clearly a copy-paste error. Surely excluding the same file twice can't have a useful meaning!

PR #2031, opened 2022-12-22
fix: support `UnknownLength` in `ak.types.ArrayType`

### LINKED ISSUE #1937: Typetracers with unknown length can't print their type
### Version of Awkward Array

HEAD

```python
>>> import awkward as ak
>>> typetracer = ak.Array([[1, 2, 3], [], [4, 5]]).layout.typetracer
>>> typetracer.offsets.data.shape = (ak._typetracer.UnknownLength,)
>>> ak.Array(typetracer).type
  File "/home/jpivarski/irishep/awkward/src/awkward/highlevel.py", line 463, in type
    self._layout.form.type_from_behavior(self._behavior), len(self._layout)
TypeError: 'UnknownLengthType' object cannot be interpreted as an integer
```

This should be easy: replacing `len(self._layout)` with `self._layout.length` and using the string `"??"` in case of `ak._typetracer.UnknownLength`.

Cc: @douglasdavis

agoose77:
Fixes #1937

### DIFF
--- a/src/awkward/types/arraytype.py
+++ b/src/awkward/types/arraytype.py
@@ -16,12 +16,12 @@ class ArrayType:
                 )
             )
         if not (
-            (ak._util.is_integer(length) and length > 0)
+            (ak._util.is_integer(length) and length >= 0)
             or length is ak._typetracer.UnknownLength
         ):
             raise ak._errors.wrap_error(
                 ValueError(
-                    "{} 'length' must be a positive integer or unknown length, not {}".format(
+                    "{} 'length' must be a non-negative integer or unknown length, not {}".format(
                         type(self).__name__, repr(length)
                     )
                 )
--- a/src/awkward/types/arraytype.py
+++ b/src/awkward/types/arraytype.py
@@ -3,7 +3,6 @@
 import sys
 
 import awkward as ak
-import awkward.types.type
 
 
 class ArrayType:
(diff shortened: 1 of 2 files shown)


src/awkward/types/arraytype.py
```diff
@@ -16,10 +15,13 @@ def __init__(self, content, length):
...
             )
-        if not ak._util.is_integer(length) or length < 0:
+        if not (
+            (ak._util.is_integer(length) and length >= 0)
+            or length is ak._typetracer.UnknownLength
+        ):
             raise ak._errors.wrap_error(
                 ValueError(
-                    "{} 'length' must be of a positive integer, not {}".format(
+                    "{} 'length' must be a non-negative integer or unknown length, not {}".format(
```
jpivarski, line 24: Good catch!

jpivarski (approved):
Earlier today, @ioanaif said she was planning on doing this one, but it wasn't assigned at the time that you started it, and it is a pretty small fix.

Fortunately, there are many more issues to fix.  `:)`

jpivarski:
I've been putting a little time between approval and merging by collecting these into a browser window, reviewing them one after another, and then going back to the beginning to merge them. Now I'm back around to the end of the list, so I'll merge this.

agoose77:
Ah, sorry @ioanaif! I didn't assign the issue to myself because I noticed that it was fairly small. I'll make a note to be better at always assigning things I'm working on.

PR #2033, opened 2022-12-23
refactor: add `@final` to contents, types, and forms

agoose77:
This type hint makes it clear (and, in future, enforces) that contents, types, and forms are all final non-inheritable types.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
   rev: "v5.11.3"
   hooks:
     - id: isort
-      exclude: ^src/(awkward|awkward/_v2)/__init__\.py$
+      exclude: ^src/awkward/__init__\.py$
 
 - repo: https://github.com/asottile/setup-cfg-fmt
   rev: v2.2.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -218,6 +218,7 @@ exclude = ["^src/awkward/[^/]+\\.py$"]
 plugins = [
     "numpy.typing.mypy_plugin"
 ]
+python_version = "3.11"
 
 [[tool.mypy.overrides]]
 module = [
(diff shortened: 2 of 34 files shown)


jpivarski:
A warning from PyLint (they easily get buried):

> typing.final is not supported by all versions included in the py-version setting

Is there a way to import it as backports for Python versions as early as 3.7?

jpivarski (approved):
That's right: the `Content`, `Form`, and `Type` subclasses should not be subclassed; they're final.

So are `ak.types.ArrayType`, `ak.record.Record`, and the `ak.index.Index` subclasses (e.g. `ak.index.Index64`). These are not subclasses of the above, but are closely related.

I suppose you can't annotate `Content`, `Form`, and `Type` (the superclasses) to say that they can't have subclasses _beyond_ those that are defined in our codebase. (I think Scala has a way to say that: they have a lot of module-level/namespace-level type qualifiers.)

Also, this would only prevent people from defining subclasses and then type-checking them, right? We're not preventing people from hacking things as a test—just formally declaring something as "okay for production," right?

agoose77:
This is done, it just needs a fix for mypy on older Python versions. I think the issue is how we import `typing_extensions` and `typing` in `awkward.typing`. It seems like there's no easy way to say "import everything from typing and typing_extensions (if appropriate), where typing_extensions takes priority". I think I might have to explicitly import the appropriate types from `typing_extensions`, but I'll properly dig into this next week.

> I suppose you can't annotate Content, Form, and Type (the superclasses) to say that they can't have subclasses beyond those that are defined in our codebase. (I think Scala has a way to say that: they have a lot of module-level/namespace-level type qualifiers.)

Unfortunately not, but I think we'll be OK at least with constraining the immediate subclasses.

> Also, this would only prevent people from defining subclasses and then type-checking them, right? We're not preventing people from hacking things as a test—just formally declaring something as "okay for production," right?

Right, this is not runtime-enforced at all.

PR #2036, opened 2022-12-23
fix: soft-deprecate `flatten_records` instead of hard-deprecation

agoose77:


jpivarski (approved):
Yes, this should fix https://github.com/dask-contrib/dask-awkward/pull/136#issuecomment-1364181164, assuming that dask-awkward doesn't have "warnings as errors" turned on!

I'll push this through to a new release.

PR #2040, opened 2022-12-29
docs: correct canonical URL

agoose77:
This PR ensures that we only build the sitemap for the `main` version, and that the `link rel="canonical"` links point to `main`. This ensures that Google does not index versioned links.


agoose77:
I believe that I'm free to merge this; my recollection is that Jim is happy for me to fix docs-related bugs that are semi-urgent (in this case, propagating correct unversioned links & adding Sphinx upper cap).

jpivarski:
That's true. Thanks for merging this!

PR #2044, opened 2022-12-30
test: rename tests to use identifiers

agoose77:
This PR fixes a minor annoyance that both @lobis and I have run in to. PyCharm, our preferred editor, requires that tests are importable in order to correctly detect the PyTest configuration. 

If no-one has a strong reason to keep our non-identifier naming convention, this makes it possible to run individual unit-tests from PyCharm's test runner
image


jpivarski (approved):
So this means no more hyphens in test file names? That's okay, and I'd be surprised if anyone else on the team has an opinion on that.

I started using hyphens because it allows for some readability distinctions in names: e.g. "`test_9999-change-from_buffers-arguments.py`" where "`from_buffers`" is an identifier in a sentence or phrase, with hyphens taking the role of spaces (and sometimes dots: "`change-ak-from_buffers`").

Emacs Lisp has trained me to see hyphens as spaces, with function names like [tool-bar-add-item-from-menu](https://www.gnu.org/software/emacs/manual/html_mono/elisp.html#index-tool_002dbar_002dadd_002ditem_002dfrom_002dmenu) and variable names like [switch-to-buffer-preserve-window-point](https://www.gnu.org/software/emacs/manual/html_mono/elisp.html#index-switch_002dto_002dbuffer_002dpreserve_002dwindow_002dpoint).

Hyphens in file names can also be a good way to ensure that a script is a script, not importable as a module. But you want to do the opposite here. So yes, let's merge this!

agoose77:
In general, I like the fact that we have hyphenated names (for the reasons that you give). But, as you note, I want to change that in this case to solve a tooling problem :)

PR #2046, opened 2022-12-30
fix: remove unused keyword arg

agoose77:
#1968 added the redundant `return_simplified` argument to `ak.transform`. I missed it in my initial review, but it doesn't do anything and should be removed. I think we can do this immediately; the option isn't documented, and doesn't do anything.

### DIFF
--- a/src/awkward/operations/ak_transform.py
+++ b/src/awkward/operations/ak_transform.py
@@ -19,7 +19,6 @@ def transform(
     right_broadcast=True,
     numpy_to_regular=False,
     regular_to_jagged=False,
-    return_simplified=True,
     return_value="simplified",
     highlevel=True,
     behavior=None,


jpivarski (approved):
Oops!

You're right: that's a relic of a partial implementation. At first, `return_array` and `return_simplified` were to be two booleans, but two booleans have 4 possible values and only 3 of them would have been meaningful, so I combined them into a single enumerated argument, `return_value`.

The `return_simplified` is left over from having implemented it the one way, changing my mind, and implementing it the other way without completely cleaning up.

It _is_ possible for someone to depend on it the same way as the last change in arguments: someone could be uselessly passing an argument through a wrapper of `transform`, which would then fail if an argument of this name disappears. However, this is much less likely than in the previous case, since `transform` is a new function and it's too low-level to be a dask-awkward function. So I agree that we can try silently removing it.

I approve this PR for merging.

PR #2048, opened 2022-12-30
fix: support `regular_to_jagged` in `Content._recursively_apply` / `ak.transform`

### LINKED ISSUE #2047: `ak.transform` doesn't respect `regular_to_jagged` for single argument
### Version of Awkward Array

main

This simple demo does not produce a ragged array:
```python
import awkward as ak
import numpy as np

x = ak.transform(
    lambda layout, depth, continuation, **kwargs: continuation(),
    ak.from_numpy(np.arange(2 * 3 * 4 * 5).reshape(2, 3, 4, 5)),
    regular_to_jagged=True,
    numpy_to_regular=True
)
x.show(type=True)
```

Yet, I would expect it to, because of `regular_to_jagged` and `numpy_to_regular`. It is not documented that `regular_to_jagged` is only implemented for multiple array arguments, and we also implement `numpy_to_regular` for the single-array variant. Therefore, I suspect this is an omission rather than a policy choice.

agoose77:
Fixes #2047

### DIFF
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -27,6 +27,7 @@ def recursively_apply(
     return_simplified: bool = True,
     return_array: bool = True,
     function_name: str | None = None,
+    regular_to_jagged=False,
 ) -> Content | Record | None:
 
     if isinstance(layout, Content):
@@ -40,6 +41,7 @@ def recursively_apply(
                 "allow_records": allow_records,
                 "keep_parameters": keep_parameters,
                 "numpy_to_regular": numpy_to_regular,
+                "regular_to_jagged": regular_to_jagged,
                 "return_simplified": return_simplified,
                 "return_array": return_array,
                 "function_name": function_name,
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
This does seem like a good idea. Had it been a `broadcast_and_apply` argument that wasn't implemented in `recursively_apply`? If so, this is a good step toward further unification. `regular_to_jagged` is something one might want to do or not do, independent of whether there's only one array to recurse over or there are several arrays to broadcast together.

jpivarski:
And yes, it can be merged.

agoose77:
> Had it been a `broadcast_and_apply` argument that wasn't implemented in `recursively_apply`? If so, this is a good step toward further unification. 

I *think* so, at least, that's what it looked like. I didn't dig too deeply into the `git blame`, but it seems as though the broadcasting logic came directly from the C++, and we know that `ak.transform` is younger than that.

jpivarski:
I do remember a `regular_to_jagged` boolean argument, and in this PR, it's pretty obvious that such an argument wasn't in `recursively_apply` until now, so I guess it must have been in `broadcast_and_apply`.

PR #2049, opened 2022-12-30
chore: fix poor rename

agoose77:
Somehow a single test was incorrectly renamed in #2044. I wrote a simple Xonsh script to perform that translation, so I wonder if a local untracked file threw something off.


jpivarski (approved):
Okay!

PR #2050, opened 2022-12-31
docs: correct docstring for `ak.metadata_from_parquet`

agoose77:


jpivarski (approved):
Thanks for making this consistent! At one point, the function really did return the Parquet metadata as-is, but I can see in the `out` variable that it's no longer true.

This can be merged when you're done with it.

agoose77:
@jpivarski yes I recall that too! I think this is correct now - I'm literally reading the implementation.

PR #2052, opened 2022-12-31
fix: read behavior from highlevel `ak.ArrayBuilder`

### LINKED ISSUE #2051: `ak.ArrayBuilder` loses behaviour through `ak._util.behaviorof`
### Version of Awkward Array

main

```python
import awkward as ak

class Point:
    def length(self):
        raise NotImplementedError

behavior = {
    ('*', 'point'): Point
}
builder = ak.ArrayBuilder(behavior=behavior)
with builder.record("Point"):
	builder.field("x").real(1.0)
	builder.field("y").real(2.0)
	builder.field("z").real(3.0)

assert ak._util.arrays_approx_equal(builder, builder.snapshot())

agoose77:
`ak._util.behavior_of` currently ignores the `behavior` of an `ak.ArrayBuilder`. This PR fixes #2051, and also fixes some bugs in the `arrays_approx_equal` function w.r.t record/array handling.

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -892,7 +892,7 @@ def arrays_approx_equal(
 
         # Require that the arrays have the same evaluated types
         if not (
-            (arrayclass(left, left_behavior) is arrayclass(right, right_behavior))
+            arrayclass(left, left_behavior) is arrayclass(right, right_behavior)
             or not check_parameters
         ):
             return False
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -859,8 +859,8 @@ def arrays_approx_equal(
 
     import awkward.forms.form
 
-    left_behavior = ak._util.behavior_of(left, behavior=ak.behavior)
-    right_behavior = ak._util.behavior_of(right, behavior=ak.behavior)
+    left_behavior = ak._util.behavior_of(left)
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is not checking all of the `parameters`, it's just ensuring that the `arrayclass`/`recordclass` is the same (which might be a recursive search through the layout). Is that desirable? The `arrays_approx_equal` will be insensitive to made-up parameters, which are used in some tests to verify that parameters can go round-trip through another format (such as in the Arrow/Parquet tests). The `arrays_approx_equal` function isn't used for those tests, but if you're planning on expanding the use of this function, you'll want to consider that. After you use it in a million tests, assuming that it only checks `arrayclass`/`recordclass`, it would be harder to change.

Feel free to merge this PR after you've made a decision about how much of the `parameters` you want this function to test.

agoose77:
@jpivarski this function was intended to check parameters already with https://github.com/scikit-hep/awkward/blob/f00a54bda4ccdced2fe876928697b390ed0b33ba/src/awkward/_util.py#L888-L891, which should be working. I *think* you might not have seen this if you just saw the diff, which is what I imagine happened, but let me know if I've made an oversight!

jpivarski:
That's right! (I only looked at the diff.) So this checks everything, then.

Looks good!

PR #2055, opened 2023-01-01
fix: rebuild invalid `check` pointers in `ArrayBuilder`

### LINKED ISSUE #2053: `ak.ArrayBuilder` loses data when initalising builder outside of Numba
### Version of Awkward Array

main

The following reproducer does not preserve values set outside of Numba:

```python3
import awkward as ak
from numba import njit


@njit
def func(builder):
    builder.begin_record()
    builder.field("time").real(2.0)
    builder.end_record()
    return builder


builder = ak.ArrayBuilder()
builder.begin_record()
builder.field("time").real(0.0)
builder.end_record()

func(builder)

builder.begin_record()
builder.field("time").real(5.0)
builder.end_record()

print(builder.snapshot())

```

### Observations
This only seems to happen if the record is not given a name AND if the Python `ak.ArrayBuilder` operations occur before the njitted function.

### LINKED ISSUE #2054: `ak.ArrayBuilder.type` contains non-canonical union
### Version of Awkward Array

main

In the following example, the `ak.ArrayBuilder()._layout.form()` contains a union that is ultimately simplified. However, the `type` attribute looks at this form, and so includes the union.

```python
import awkward as ak
from numba import njit


@njit
def func(builder):
    builder.begin_record("x")
    builder.field("time").real(2.0)
    builder.end_record()
    return builder


builder = ak.ArrayBuilder()
builder.begin_record("x")
builder.field("time").real(0.0)
builder.end_record()

func(builder)

print(builder.type)
print(builder.snapshot().type)

```

### Observations
This seems to require the same conditions as #2053

agoose77:
This PR fixes #2054, fixes #2053, which occur when a user mixes "fast" and "slow" modes of the record / field builders. The string pointers that are used to implement the fast mechanism are assumed to come from an intern pool, i.e. the same string has the same pointer. This fails when Python strings are used, as they come from different pools, and aren't guaranteed to be stable.

To correct this, we can ensure that the slow-mode invalidates any pointers, and that the fast-mode treats `nullptr` as a cue to rebuild the pointer.

### DIFF
--- a/awkward-cpp/src/libawkward/builder/RecordBuilder.cpp
+++ b/awkward-cpp/src/libawkward/builder/RecordBuilder.cpp
@@ -399,15 +399,18 @@ namespace awkward {
     }
 
     if (!begun_  &&
-        ((check  &&  name_ == name)  ||  (!check  &&  nameptr_ == name) || (!check && nameptr_ == nullptr && name_ == name))) {
+        ((check  &&  name_ == name)  ||  (!check  &&  nameptr_ == name))) {
+      begun_ = true;
+      nextindex_ = -1;
+      nexttotry_ = 0;
+    } else if (!begun_  &&
+        ((!check && nameptr_ == nullptr && name_ == name))) {
       begun_ = true;
       nextindex_ = -1;
       nexttotry_ = 0;
 
       // Rebuild pointer for this name
-      if (!check && nameptr_ == nullptr) {
-          nameptr_ = name;
-      }
+      nameptr_ = name;
     }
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is definitely a different solution than what I had in mind, but it makes a lot of sense. Now `_check` and `_fast` are safe to mix (though a `_check` will slow down `_fast`).

I see that you've implemented this for both `beginrecord_*` and `field_*`.

This will require an increase in the awkward-cpp version.

PR #2056, opened 2023-01-01
chore: increase awkward-cpp version for #2055.

jpivarski:
I could do this one pretty quickly, since I'm at my laptop for the moment!


jpivarski:
Looking at the labels, I'm reminded that I'll need to remove the `next-release` from Awkward and Uproot. I'll do that on Monday.

The next time we use those labels, it will be for 2.1.0.

agoose77 (approved):

PR #2063, opened 2023-01-04
fix: support merging of `RegularArray` and `NumpyArray`

### LINKED ISSUE #2058: TypeError in ak.to_packed with UnionType
### Version of Awkward Array

2.0.5

I'm getting a TypeError when calling `ak.to_packed` on a UnionType of two RegularTypes with different dimensions. 

Context: this is a unit test case from https://github.com/scverse/anndata/pull/647 that failed after upgrading to v2.x. 

```python
arr = ak.concatenate([ak.Array(np.ones((2, 2))), ak.Array(np.ones((2, 3)))])
ak.to_packed(arr)
```

```pytb
  File "<string>", line 1, in <module>
  File "/home/sturm/anaconda3/envs/anndata_dev/lib/python3.9/site-packages/awkward/operations/ak_to_packed.py", line 70, in to_packed
    return _impl(array, highlevel, behavior)
  File "/home/sturm/anaconda3/envs/anndata_dev/lib/python3.9/site-packages/awkward/operations/ak_to_packed.py", line 75, in _impl
    out = layout.to_packed()
  File "/home/sturm/anaconda3/envs/anndata_dev/lib/python3.9/site-packages/awkward/contents/unionarray.py", line 1550, in to_packed
    return UnionArray(
  File "/home/sturm/anaconda3/envs/anndata_dev/lib/python3.9/site-packages/awkward/contents/unionarray.py", line 81, in __init__
    raise ak._errors.wrap_error(
TypeError: while calling

    ak.to_packed(
        array = <Array [[1, 1], [...], ..., [1, 1, 1]] type='4 * union[2 * ...'>
        highlevel = True
        behavior = None
    )

Error details: UnionArray cannot contain mergeable 'contents' (RegularArray of '2 * float64' and RegularArray of '3 * float64'); try UnionArray.simplified instead
```

agoose77:
This is a PoC to fix #2058. Jim and I also discussed doing this properly at the form level, which I might even explore in another draft PR.

- [ ] Restore `main` casting semantics

### DIFF
--- a/src/awkward/_nplikes.py
+++ b/src/awkward/_nplikes.py
@@ -64,7 +64,10 @@ class NumpyMetadata(Singleton):
 
     nat = numpy.datetime64("NaT")
     datetime_data = numpy.datetime_data
-    issubdtype = numpy.issubdtype
+
+    @property
+    def issubdtype(self):
+        return numpy.issubdtype
 
     AxisError = numpy.AxisError
 
@@ -365,6 +368,9 @@ class NumpyLike(Singleton):
 
     def datetime_as_string(self, *args, **kwargs):
         return self._module.datetime_as_string(*args, **kwargs)
+
+    def can_cast(self, *args, **kwargs):
+        return self._module.can_cast(*args, **kwargs)
 
     @classmethod
     def is_own_array(cls, obj) -> bool:
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -989,7 +989,7 @@ class TypeTracer(ak._nplikes.NumpyLike):
             try_touch_data(x)
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def concatenate(self, arrays):
+    def concatenate(self, arrays, casting="same_kind"):
         for x in arrays:
             try_touch_data(x)
 
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -1236,6 +1236,9 @@ class TypeTracer(ak._nplikes.NumpyLike):
         # array, max_line_width, precision=None, suppress_small=None
         return "[?? ... ??]"
 
+    def can_cast(self, *args, **kwargs):
+        return numpy.can_cast(*args, **kwargs)
+
     def datetime_as_string(self, *args, **kwargs):
         for x in args:
             try_touch_data(x)
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -931,7 +931,7 @@ class ListArray(Content):
         ):
             return self._mergeable(other.content, mergebool)
 
-        if isinstance(
+        elif isinstance(
             other,
             (
                 ak.contents.RegularArray,
@@ -940,6 +940,9 @@ class ListArray(Content):
             ),
         ):
             return self._content._mergeable(other.content, mergebool)
+
+        elif isinstance(other, ak.contents.NumpyArray) and len(other.shape) > 1:
+            return self._mergeable(other._to_regular_primitive(), mergebool)
 
         else:
             return False
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -348,10 +348,11 @@ class NumpyArray(Content):
             if self._data.ndim != other._data.ndim:
                 return False
 
+            # Obvious fast-path
             if self.dtype == other.dtype:
                 return True
 
-            # Merge booleans i.e. {bool, number}
+            # Special-case booleans i.e. {bool, number}
             elif mergebool and (
                 np.issubdtype(self.dtype, np.bool_)
                 and np.issubdtype(other.dtype, np.number)
@@ -360,13 +361,7 @@ class NumpyArray(Content):
             ):
                 return True
 
-            # Merge number e.g. {complex, float} etc.
-            elif np.issubdtype(self.dtype, np.number) and np.issubdtype(
-                other._data.dtype, np.number
-            ):
-                return True
-            # We should only be here with datetime dtypes, failed boolean
-            # merges, or obscure datetime
+            # Default merging (can we cast one to the other)
             else:
                 return self.backend.nplike.can_cast(
                     self.dtype, other.dtype, casting="same_kind"
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -1332,9 +1332,12 @@ class NumpyArray(Content):
         else:
             return True
 
-    def _to_regular_primitive(self):
+    def _to_regular_primitive(self) -> ak.contents.RegularArray:
+        # A length-1 slice in each dimension
         index = tuple([slice(None, 1)] * len(self.shape))
+        # Broadcast this trivial slice to the true dimensions (zero-copy)
         new_data = self.backend.nplike.broadcast_to(self._data[index], self.shape)
+        # Convert contiguous array to `RegularArray`
         return NumpyArray(
             new_data, backend=self.backend, parameters=self.parameters
         ).to_RegularArray()
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -340,17 +340,9 @@ class NumpyArray(Content):
             )
 
     def _mergeable_next(self, other, mergebool):
-        if isinstance(
-            other,
-            (
-                ak.contents.IndexedArray,
-                ak.contents.IndexedOptionArray,
-                ak.contents.ByteMaskedArray,
-                ak.contents.BitMaskedArray,
-                ak.contents.UnmaskedArray,
-            ),
-        ):
-            return self._mergeable(other._content, mergebool)
+
+        if len(self.shape) > 1:
+            return self._to_regular_primitive()._mergeable(other, mergebool)
 
         elif isinstance(other, ak.contents.NumpyArray):
             if self._data.ndim != other._data.ndim:
@@ -401,8 +393,7 @@ class NumpyArray(Content):
         if len(others) == 0:
             return self
 
-        # Resolve merging against regular types by
-        if any(isinstance(o, ak.contents.RegularArray) for o in others):
+        if len(self.shape) > 1:
             return self.to_RegularArray()._mergemany(others)
 
         head, tail = self._merging_strategy(others)
@@ -1355,3 +1346,10 @@ class NumpyArray(Content):
             )
         else:
             return True
+
+    def _to_regular_primitive(self):
+        index = tuple([slice(None, 1)] * len(self.shape))
+        new_data = self.backend.nplike.broadcast_to(self._data[index], self.shape)
+        return NumpyArray(
+            new_data, backend=self.backend, parameters=self.parameters
+        ).to_RegularArray()
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -361,6 +361,15 @@ class NumpyArray(Content):
             ):
                 return True
 
+            # Currently we're less permissive than NumPy on merging datetimes / timedeltas
+            elif (
+                np.issubdtype(self.dtype, np.datetime64)
+                or np.issubdtype(self.dtype, np.timedelta64)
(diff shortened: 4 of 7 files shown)


src/awkward/contents/numpyarray.py
```diff
@@ -340,59 +340,34 @@ def _offsets_and_flattened(self, axis, depth):
...
-            as_regular_array = self.to_RegularArray()
-            assert isinstance(as_regular_array, ak.contents.RegularArray)
-            return as_regular_array._content._mergeable(other._content, mergebool)
+            # Default merging (can we cast one to the other)
+            else:
+                return self.backend.nplike.can_cast(
+                    self.dtype, other.dtype, casting="same_kind"
+                ) or self.backend.nplike.can_cast(
+                    other.dtype, self.dtype, casting="same_kind"
+                )
```
agoose77, line 379: The idea here is that `np.concatenate` uses something like `can_cast` to test mergeability. I *think* this should predict `np.concatenate`, at least (I'm not certain).

src/awkward/contents/numpyarray.py
```diff
@@ -1355,3 +1331,10 @@ def _is_equal_to(self, other, index_dtype, numpyarray):
             )
         else:
             return True
+
+    def _to_regular_primitive(self):
+        index = tuple([slice(None, 1)] * len(self.shape))
+        new_data = self.backend.nplike.broadcast_to(self._data[index], self.shape)
+        return NumpyArray(
+            new_data, backend=self.backend, parameters=self.parameters
+        ).to_RegularArray()
```
agoose77: Yes, this is a terrible name; this PR is a PoC and I'd appreciate a better suggestion. Maybe `_to_mergeable_hack` ;)

src/awkward/_nplikes.py
```diff
@@ -64,7 +64,10 @@ class NumpyMetadata(Singleton):
 
     nat = numpy.datetime64("NaT")
     datetime_data = numpy.datetime_data
-    issubdtype = numpy.issubdtype
+
+    @property
+    def issubdtype(self):
+        return numpy.issubdtype
```
agoose77, line 70: `issubdtype` behaves like a method descriptor.

agoose77:
N.B that the tests will fail, because this changes merging behavior for date-times (in line with NumPy AFAICT).

src/awkward/_typetracer.py
```diff
@@ -989,7 +989,7 @@ def unique(self, *args, **kwargs):
             try_touch_data(x)
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def concatenate(self, arrays):
+    def concatenate(self, arrays, casting="same_kind"):
```
agoose77, line 992: This is not right - we should actually consider this variable in `concatenate`. However, this would fall under the typetracer / nplike refactoring that I'm also working on separately, so I'll handle it there.

agoose77:
TODO: restore existing datetime mergeability test (stricter)

src/awkward/_nplikes.py
jpivarski: Is this just formal, for code symmetry? Whether it's an attribute or a property on the `NumpyMetadata.instance()` singleton, `issubdtype` acts like a function (it doesn't consume a `self` argument).

src/awkward/_typetracer.py
jpivarski: Okay.

src/awkward/contents/listarray.py
```diff
@@ -931,7 +931,7 @@ def _mergeable_next(self, other, mergebool):
         ):
             return self._mergeable(other.content, mergebool)
 
-        if isinstance(
+        elif isinstance(
```
jpivarski, line 934: The original idea of having the option-type/index check be a separate `if` statement from the subsequent chain is that it's conceptually distinct. Most node types, across the `Content` class hierarchy, start by asking this same question before getting into questions that are specific to each node type.

But since the option-type/index `if` block ends by returning, `if` and `elif` are functionally equivalent here, and looking at any one node's `_mergeable_next` implementation in isolation, a single `if` chain would be simpler.

src/awkward/contents/numpyarray.py
```diff
@@ -340,59 +340,34 @@ def _offsets_and_flattened(self, axis, depth):
...
-                ak.contents.IndexedOptionArray,
-                ak.contents.ByteMaskedArray,
-                ak.contents.BitMaskedArray,
-                ak.contents.UnmaskedArray,
-            ),
-        ):
-            return self._mergeable(other._content, mergebool)
+
+        if len(self.shape) > 1:
+            return self._to_regular_primitive()._mergeable(other, mergebool)
```
jpivarski, line 345: This is the main part of the PR.

In our conversation, I was advocating for

```python
        assert len(self.shape) == 1
```

here, with an upstream conversion of the whole layout to replace `NumpyArray` with `RegularArray` before it enters into `mergeable` _and_ `mergemany`.

I'm assuming that the `NumpyArray` would have to be converted (often a copy) into `RegularArray` again if `mergeable` returns `True` and the `mergemany` step happens. I was just hoping that could be made a single conversion. (It's not a big deal if it isn't.)

Part of that reasoning is that `mergeable`/`mergemany` is an internal API that we have to be careful with, anyway. (Calling `mergemany` on data that would have `mergeable == False` is undefined; I suspect a segfault would result.)

src/awkward/contents/numpyarray.py
jpivarski: It's specifically this part of the code (line 350 to here) that was tuned to give the right merging behavior for datetime and timedelta types. It was originally wrong; I fixed it here:

https://github.com/scikit-hep/awkward/pull/1183/commits/0734ab26a47e65548f84f750225fa0eb4adb435f#diff-cdc2b042962b124ea98cc8bacd6aea030a50dc73f6502f9686475554c8913320

("Wrong" was mixing datetimes or timedeltas of different units without conversion.)

As I said in today's meeting, I'd like to continue being stricter than NumPy, not allowing datetimes or timedeltas to be mixed with non-datetimes or non-timedeltas. When the Array API Consortium eventually gets around to defining time-unit behavior, I suspect that they'll come to the same conclusion: promoting non-time data to act as time-data means _guessing_ a unit, which usually isn't safe. If I'm right and they _do_ pick this behavior, we won't have to change it. (Since it's a behavior without an API marker, like a function argument, it would be hard to set up a deprecation cycle.)

So let's revert the use of `nplike.can_cast`, at least for time units or any conversions that are already in violation with Array API decisions. We can diverge from current-NumPy if we think doing so means making smaller or no changes later, when NumPy is compliant with Array API.

jpivarski: I'm surprised, though, that the above didn't break any tests. I remember there being some tests that were sensitive to the time-unit fix, but I can't find them now.

jpivarski (commented):
Just a few comments, because this is in the draft stage.

src/awkward/contents/numpyarray.py
agoose77: This conversion is low cost, as only for JAX do we actually need to copy any data. Is your point about wanting an `assert` here still necessary in spite of that?

agoose77:
> I'm surprised, though, that the above didn't break any tests. I remember there being some tests that were sensitive to the time-unit fix, but I can't find them now.

It does, it breaks many tests! That was why I raised it today, because we clearly care about this.

> Is this just formal, for code symmetry? Whether it's an attribute or a property on the NumpyMetadata.instance() singleton, issubdtype acts like a function (it doesn't consume a self argument).

The Python attribute resolution machinery will always invoke `__get__` on a function that's resolved from the class dictionary. So, without this descriptor, Python runs `func.__get__` to return a bound function to our nplike object, which is not desirable.

src/awkward/contents/numpyarray.py
jpivarski: We concluded from some experiments (and hindsight) that when `reshape` reduces the number of dimensions, it _does_ have to copy, for any backend.

```python
>>> a = np.arange(3*5*7).reshape(3, 5, 7)
>>> b = a[1:, 1:, 1:]
>>> c = a.T

>>> a[-1, -1, -1] = 123

>>> b
array([[[ 43,  44,  45,  46,  47,  48],
        [ 50,  51,  52,  53,  54,  55],
        [ 57,  58,  59,  60,  61,  62],
        [ 64,  65,  66,  67,  68,  69]],

       [[ 78,  79,  80,  81,  82,  83],
        [ 85,  86,  87,  88,  89,  90],
        [ 92,  93,  94,  95,  96,  97],
        [ 99, 100, 101, 102, 103, 123]]])
>>> c  # also has a "123" in the last element

>>> b_flat = b.reshape((-1,) + b.shape[2:])
>>> c_flat = c.reshape((-1,) + c.shape[2:])

>>> a[-1, -1, -1] = 321

>>> b_flat
array([[ 43,  44,  45,  46,  47,  48],
       [ 50,  51,  52,  53,  54,  55],
       [ 57,  58,  59,  60,  61,  62],
       [ 64,  65,  66,  67,  68,  69],
       [ 78,  79,  80,  81,  82,  83],
       [ 85,  86,  87,  88,  89,  90],
       [ 92,  93,  94,  95,  96,  97],
       [ 99, 100, 101, 102, 103, 123]])
>>> c_flat  # also does not have a "321" in the last element
```

It's because when a dimension is removed, there's nowhere for NumPy to insert the information that there should be a different byte-skip between the affected numbers.

So this `NumpyArray` → `RegularArray` conversion is an $\mathcal{O}(n)$ performance cost for all backends, including NumPy.

JAX had some different performance cost, which NumPy and CuPy didn't have, which I can't remember right now.

agoose77: Right, if we use `to_RegularArray()` we would see this penalty (for non-contiguous arrays).

Hence, I wrote [this function](https://github.com/scikit-hep/awkward/pull/2063/files#diff-513bcd10ffc2a21a0f20e0bab73d62a00044beaea16ea6245642c31c74521e11R1335-R1343), which provides a hacky solution given that we don't care about the buffer contents; scale the array to a single scalar, and then broadcast it to the appropriate dimensions. This creates an array with `strides = (0, 0, ..., 0)`, which consumes no additional memory. Broadcasting like this doesn't work for JAX, because JAX doesn't have views over buffers in the same way that NumPy, JAX do.

agoose77:
@jpivarski I've restored the date-time handling that corresponds to the previous behaviour. Due to the changes in this PR, `_mergeable_next` should match the old behaviour (besides the change to shape dependence), but the logic is inverted to be clearer, and is now potentially _less_ permissive than the previous implementation. Due to the exhaustiveness of the various test cases, I *don't* think that is likely ­— for the types we expect, at this point in the function we should have non-equal numeric types AFAICT.

Previously, we just returned `True` if we didn't early-exit. This routine is more explicit about why we're returning `True`, by ending on a `can_cast` call.

src/awkward/contents/numpyarray.py
jpivarski: Okay, yeah, that's cheap-as-free! Now I remember that this, specifically, is the case that JAX doesn't handle as well as NumPy and CuPy. And it's fine to ignore this performance issue in JAX (it's a corner case of a corner case).

jpivarski:
Okay, I'm on board with this now. My one performance question was addressed, the dates are as strict as they used to be, and calling `can_cast` at the `else` at the end of the `elif` chain is safer than saying things are mergeable if they pass our gauntlet.

> potentially _less_ permissive than the previous implementation

If it's actually less permissive, then in principle that would be backward incompatible (something that used to work now won't). Do you know of cases that we used to merge but would now be considered not mergeable?

agoose77:
> If it's actually less permissive, then in principle that would be backward incompatible (something that used to work now won't). Do you know of cases that we used to merge but would now be considered not mergeable?

Yes, that was why I expounded upon this to assert that although it cannot be more permissive, I don't think it's less permissive either. It doesn't fail any tests, and I can't think of any types that would not be supported.

Ooh, actually, let me try one combination.

jpivarski:
Okay, so if the one new test you're trying passes, then I think this PR is ready to be merged.

agoose77:
Now that I think about it, `np.concatenate` defaults to `"same_kind"`, so the merge would fail if we had previously permitted an invalid operation.

PR #2065, opened 2023-01-05
fix: support option-in-record for `fill_none`

### LINKED ISSUE #2064: `ak.fill_none` doesn't support option-in-record
### Version of Awkward Array

main

The following example is left unchanged by `ak.fill_none`:
```python
import awkward as ak

record = ak.zip({
    'x': [1, None],
    'y': [2, 3]
})

result = ak.fill_none(record, 0, axis=-1)
result.show()
```

I'm fairly sure that this is a bug, because `ak.is_none` walks successfully through records. My understanding of records is that it's only reducers that cannot peer through them. 

Of course, this example would imply
```python
record = ak.zip({
    "x": [1, 4], 
	"y": [2, 3]
}).mask[[True, False]]
assert ak.fill_none(record, 0, axis=-1).to_list() == [
    {"x": 1, "y": 2},
    0
]
```

I think this is reasonable; again, `is_none` behaves identically to this. Maybe we _do_ want a parameter in future, e.g. `visit_records` or some such.

agoose77:
Fixes #2064 (see issue for details). I feel like we *might* have touched upon this before, so keen for comments.

### DIFF
--- a/src/awkward/operations/ak_fill_none.py
+++ b/src/awkward/operations/ak_fill_none.py
@@ -63,7 +63,7 @@ def fill_none(array, value, axis=-1, *, highlevel=True, behavior=None):
 
 def _impl(array, value, axis, highlevel, behavior):
     arraylayout = ak.operations.to_layout(array, allow_record=True, allow_other=False)
-    behavior = ak._util.behavior_of(array, behavior=behavior)
+    behavior = ak._util.behavior_of(array, value, behavior=behavior)
     backend = ak._backends.backend_of(arraylayout, default=cpu)
 
     # Convert value type to appropriate layout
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Some operations that used to act on data within records have been stopped at the RecordArray border, but I see no reason why this one should be. As a user, I would _expect_ to be able to say `ak.fill_none` on an array containing records and have the missing values in the records get filled in.

It has to respect the `axis` parameter, but it looks like it does.

I'd say this is ready to merge.

agoose77:
@jpivarski let me just test one additional case! I'm sorry for pinging you and then pausing the merge; I noticed something peculiar.

agoose77:
OK, the peculiarity I noticed was if the record contains options, and the record itself is wrapped in an option. This PR does what `is_none` does; the first wins.

Issue #2061, opened 2023-01-03
`awkward.backend` should handle typetracers instead of throwing exception


lgray:
### Description of new feature

Presently if you pass a typetracer to `awkward.backend` it presents an exception as follows:
```
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/lgray/coffea-dev/coffea/tests/test_jetmet_tools.py", line 74, in test_factorized_jet_corrector
    corrs_jag = corrector.getCorrection(
  File "/Users/lgray/coffea-dev/coffea/coffea/jetmet_tools/FactorizedJetCorrector.py", line 176, in getCorrection
    corrs = self.getSubCorrections(**kwargs)
  File "/Users/lgray/coffea-dev/coffea/coffea/jetmet_tools/FactorizedJetCorrector.py", line 230, in getSubCorrections
    corrections.append(func(*fargs))  # raw awkward is eager
  File "/Users/lgray/coffea-dev/coffea/coffea/lookup_tools/lookup_base.py", line 42, in __call__
    print([awkward.backend(dak.typetracer_from_form(arg.layout.form)) for arg in args])
  File "/Users/lgray/coffea-dev/coffea/coffea/lookup_tools/lookup_base.py", line 42, in <listcomp>
    print([awkward.backend(dak.typetracer_from_form(arg.layout.form)) for arg in args])
  File "/Users/lgray/miniforge3/lib/python3.9/site-packages/awkward/operations/ak_backend.py", line 25, in backend
    return _impl(arrays)
  File "/Users/lgray/miniforge3/lib/python3.9/site-packages/awkward/operations/ak_backend.py", line 31, in _impl
    raise ak._errors.wrap_error(
ValueError: while calling

    ak.backend(
        *arrays = (<Array-typetracer type='?? * var * float64'>,)
    )

Error details: at least one of the given arrays was a typetracer array. This is an internal backend that you should not have encountered. Please file a bug report at https://github.com/scikit-hep/awkward/issues/
```

After discussing with @jpivarski he recommends that this instead return `"tt"` as the backend type for typetracers since notionally that is a type of processing "backend" it's just purely metadata.

agoose77:
This actually is planned, although it's not in main yet! Good to know it is needed :)

PR #2066, opened 2023-01-05
feat: expose typetracer in public backend API

agoose77:
This PR exposes typetracer in `ak.backend` and `ak.to_backend`. It does not yet expose any callable typetracer APIs.


jpivarski (approved):
Looks good! We had to pick something for the string represesntation of a type-tracer (which I've been spelling many different ways: with and without capitalization, with and without the hyphen, sometimes specifically as the `TypeTracerArray` class, sometimes not). The name you picked, `"typetracer"` is good—we'll go with that!

agoose77:
Closes #2061

PR #2069, opened 2023-01-05
fix: listarray - slicing expects scalars

### LINKED ISSUE #2068: cuda ak.num tests fail
### Version of Awkward Array

'2.0.5'

`python -m pytest tests-cuda/test_1276-cuda-num.py`
- [x] ListArray `ak.num` test fails in `axis=1` with `TypeError: slice indices must be integers or None or have an __index__ method`
CuPy returns 0-D arrays instead of scalars.

- [x] RecordArray ak.num test fails in `axis=0 `with `AttributeError: 'int' object has no attribute 'tolist'`

```
$ python
Python 3.10.8 (main, Nov 24 2022, 14:13:03) [GCC 11.2.0] on linux
>>> import cupy as cp
>>> cp.__version__
'11.4.0'
>>> import numpy as np
>>> np.__version__
'1.23.4'
>>> 

```

ianna:
fixes issue #2068

### DIFF
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -1394,14 +1394,15 @@ class ListArray(Content):
             and self._backend.nplike.known_data
             and self._starts.length != 0
         ):
-            startsmin = self._starts.data.min()
+            # TypeError: slice indices must be integers or None or have an __index__ method
+            startsmin = self._starts.data.min().item()
             starts = ak.index.Index(
                 self._starts.data - startsmin, nplike=self._backend.index_nplike
             )
             stops = ak.index.Index(
                 self._stops.data - startsmin, nplike=self._backend.index_nplike
             )
-            content = self._content[startsmin : self._stops.data.max()]
+            content = self._content[startsmin : self._stops.data.max().item()]
         else:
             self._touch_data(recursive=False)
             starts, stops, content = self._starts, self._stops, self._content
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -1394,15 +1394,14 @@ class ListArray(Content):
             and self._backend.nplike.known_data
             and self._starts.length != 0
         ):
-            # TypeError: slice indices must be integers or None or have an __index__ method
-            startsmin = self._starts.data.min().item()
+            startsmin = int(self._starts.data.min())
             starts = ak.index.Index(
                 self._starts.data - startsmin, nplike=self._backend.index_nplike
             )
             stops = ak.index.Index(
                 self._stops.data - startsmin, nplike=self._backend.index_nplike
             )
-            content = self._content[startsmin : self._stops.data.max().item()]
+            content = self._content[startsmin : int(self._stops.data.max())]
         else:
             self._touch_data(recursive=False)
             starts, stops, content = self._starts, self._stops, self._content
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -1397,14 +1397,14 @@ class ListArray(Content):
             and self._backend.nplike.known_data
             and self._starts.length != 0
         ):
-            startsmin = int(self._starts.data.min())
+            startsmin = self._starts.data.min().item()
             starts = ak.index.Index(
                 self._starts.data - startsmin, nplike=self._backend.index_nplike
             )
             stops = ak.index.Index(
                 self._stops.data - startsmin, nplike=self._backend.index_nplike
             )
-            content = self._content[startsmin : int(self._stops.data.max())]
+            content = self._content[startsmin : self._stops.data.max().item()]
         else:
             self._touch_data(recursive=False)
             starts, stops, content = self._starts, self._stops, self._content
--- a/tests-cuda/test_1276-cuda-num.py
+++ b/tests-cuda/test_1276-cuda-num.py
@@ -127,8 +127,7 @@ def test_num_8():
     )
 
     cuda_recordarray = ak.to_backend(recordarray, "cuda")
-    
-    # AttributeError: 'int' object has no attribute 'tolist'
+
     assert ak.num(cuda_recordarray, 0) == ak.num(recordarray, 0)
 
     content0 = ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]).layout
--- a/tests-cuda/test_1276-cuda-num.py
+++ b/tests-cuda/test_1276-cuda-num.py
@@ -127,7 +127,9 @@ def test_num_8():
     )
 
     cuda_recordarray = ak.to_backend(recordarray, "cuda")
-    assert ak.num(cuda_recordarray, 0).tolist() == ak.num(recordarray, 0).tolist()
+    
+    # AttributeError: 'int' object has no attribute 'tolist'
+    assert ak.num(cuda_recordarray, 0) == ak.num(recordarray, 0)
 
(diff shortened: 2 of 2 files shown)


ianna:
@agoose77 - I haven't checked if you might have fixed it already?

agoose77:
Nope, I haven't touched this yet!

agoose77 (changes requested):
Ah I see, so this bug occurs because CuPy returns 0-D arrays instead of scalars. This is annoying! I think your fix looks like the best option available; we want to convert these to primitive types as soon as possible.

One request: could we change `x.item()` for `int(x)`? It would be one-less thing that typetracer needs to implement (even though this is a `known_data` branch).

ianna:
> Ah I see, so this bug occurs because CuPy returns 0-D arrays instead of scalars. This is annoying! I think your fix looks like the best option available; we want to convert these to primitive types as soon as possible.
> 
> One request: could we change `x.item()` for `int(x)`? It would be one-less thing that typetracer needs to implement (even though this is a `known_data` branch).

done

tests-cuda/test_1276-cuda-num.py
```diff
@@ -127,7 +127,8 @@ def test_num_8():
     )
 
     cuda_recordarray = ak.to_backend(recordarray, "cuda")
-    assert ak.num(cuda_recordarray, 0).tolist() == ak.num(recordarray, 0).tolist()
+
```
ianna, line 130: @agoose77 - this is unrelated to CuPy. It looks like the `ak.num` returns a scalar, not a 0-D array now.

agoose77: Right, that's probably the right behaviour - we would want the `ak` functions to return Python scalars always (or `UnknownScalar` for typetracers, I assume

agoose77 (approved):

tests-cuda/test_1276-cuda-num.py
jpivarski: The current policy is that Awkward's scalar outputs have NumPy scalar types (or more specifically, whatever the scalar output of the underlying array library is; for CuPy, that's zero-dimensional arrays). In other words, we do as little as possible to the array library's output—just pass it through.

Wrapping CuPy scalars (zero-dimensional arrays) with `int` is a synchronization point. In general, CUDA kernels run asynchronously and Python only waits for a kernel output if it needs it for something, such as printing it to the screen. Converting a ([promised](https://en.wikipedia.org/wiki/Futures_and_promises)) CuPy zero-dimensional array into a Python `int` will make Python sit and wait for all of the CUDA processes that array depends on to finish.

Is it not the case that

```python
self._content[startsmin : self._stops.data.max()]
```

can accept any `numbers.Integral` number, not necessarily `int`? That's why we try to keep our integer-checking generic like that. I think there's an `ak._util.is_int` helper function to check for "any integer type but not boolean," so that it will accept NumPy's `np.int64` or CuPy's `cp.ndarray` with `shape=()`.

I see in the original issue, the error message was

```
TypeError: slice indices must be integers or None or have an __index__ method
```

I don't see that error message in our codebase. Is it Awkward complaining or CuPy? Does it pass our `is_int` check but fail one of CuPy's? (Is it really the case that CuPy returns values from `max` that can't be used as slice endpoints? It seems to be accepting `array[:array.max()]` where `array` is a CuPy array.)

I'm wondering if this is the right place/way to fix this bug.

agoose77: Keeping this terse for readability, not tone!

> I'm wondering if this is the right place/way to fix this bug.

Sure! Here's my reasoning:

> Wrapping CuPy scalars (zero-dimensional arrays) with int is a synchronization point. 

Right, both `.item()` and `int()` should synchronize the device. 

> Is it Awkward complaining or CuPy? Does it pass our is_int check but fail one of CuPy's?

Awkward (well, Python)

> The current policy is that Awkward's scalar outputs have NumPy scalar types (or more specifically, whatever the scalar output of the underlying array library is; for CuPy, that's zero-dimensional arrays). In other words, we do as little as possible to the array library's output—just pass it through.

I think practically this layout method _does_ need to know a concrete length - we're not in TypeTracer, and we're reading data to determine a slice. The original error

> TypeError: slice indices must be integers or None or have an __index__ method

is raised by `slice.indices`, because CuPy's 0-D array does not provide `__index__`
```
>>> x.min().__index__
AttributeError: 'ndarray' object has no attribute '__index__'
```

NumPy, on the other hand, does expose this (yay for array libraries ;))

jpivarski: I see; CuPy wants us to know that we're slowing down the GPU. But in this case, we have to, so this is the best solution to the problem.

Both `x.item()` and `int(x)` work, but I prefer `x.item()`, too.

ianna: I wonder if we should introduce `ak.Index.max()` and `ak.Index.min()` so that instead of `int(self._stops.data.max())` we'd have `self._stops.max()`

ianna: @agoose77 and @jpivarski - Please, let me know if I should revert it to `x.item()`. Otherwise, please merge it. Thanks!

jpivarski: I'd prefer to not have `Index.max` and `Index.min`, since the same argument could be made for all of an array's methods, and that's a lot. Also, @agoose77 is planning to remove method-calls in favor of nplike functions, so that would be going away, anyway.

I have only a slight preference for `item`. But I can do it in the web interface, accept my own changes, and turn on auto-merge if you're done.

Thanks!

src/awkward/contents/listarray.py
```diff
@@ -1397,14 +1397,14 @@ def _recursively_apply(
             and self._backend.nplike.known_data
             and self._starts.length != 0
         ):
-            startsmin = self._starts.data.min()
+            startsmin = int(self._starts.data.min())
```
jpivarski:
```suggestion
            startsmin = self._starts.data.min().item()
```

src/awkward/contents/listarray.py
```diff
@@ -1397,14 +1397,14 @@ def _recursively_apply(
...
             starts = ak.index.Index(
                 self._starts.data - startsmin, nplike=self._backend.index_nplike
             )
             stops = ak.index.Index(
                 self._stops.data - startsmin, nplike=self._backend.index_nplike
             )
-            content = self._content[startsmin : self._stops.data.max()]
+            content = self._content[startsmin : int(self._stops.data.max())]
```
jpivarski:
```suggestion
            content = self._content[startsmin : self._stops.data.max().item()]
```

jpivarski (approved):
Just changing `int` to `item`.

jpivarski:
> One request: could we change `x.item()` for `int(x)`?

Oops—I just noticed that. I thought your opinion was the other way around. Is it because you'll be swapping method calls for function calls (although `int` isn't an nplike function)?

As I said, I only have a slight preference for `item`; it's because `int` looks like a mistake and I'd have to remind myself of this issue with CuPy to know why it's there. I was assuming that this method call (`item`) would be replaced by an nplike function (either `item` or a name that array libraries are unlikely to take in the future).

tests-cuda/test_1276-cuda-num.py
ianna: Thanks!

agoose77:
I made the case to @ianna to change this, but it was only a mild preference. Given that we both have mild preferences, let's just accept this as-is!

agoose77:
(@ianna feel free to merge!)

PR #2074, opened 2023-01-05
fix: render keyword and varargs

### LINKED ISSUE #2073: Documentation is missing keyword-only arguments
(And if we had any functions with position-only arguments, they would be missing, too.)

These three lines:

https://github.com/scikit-hep/awkward/blob/8d7a2c31f1ea24cef8a909fb12bb14eb812477ac/docs/prepare_docstrings.py#L146-L148

should become

```python
        argnames = [x.arg for x in node.args.posonlyargs + node.args.args + node.args.kwonlyargs]
        defaults = ["=" + tostr(x) for x in node.args.defaults + node.args.kw_defaults]
        defaults = [""] * (len(argnames) - len(defaults)) + defaults
```

As seen in this AST dump, that's where the other arguments are hiding:

```python
>>> print(ast.dump(ast.parse("def f(a, /, b, c=1, *, d=2): pass"), indent=4))
Module(
    body=[
        FunctionDef(
            name='f',
            args=arguments(
                posonlyargs=[
                    arg(arg='a')],
                args=[
                    arg(arg='b'),
                    arg(arg='c')],
                kwonlyargs=[
                    arg(arg='d')],
                kw_defaults=[
                    Constant(value=2)],
                defaults=[
                    Constant(value=1)]),
            body=[
                Pass()],
            decorator_list=[])],
    type_ignores=[])
```

----------------------

Is the above block the _only_ code that needs to be changed? The handling of lambda expressions might get the same treatment:

https://github.com/scikit-hep/awkward/blob/8d7a2c31f1ea24cef8a909fb12bb14eb812477ac/docs/prepare_docstrings.py#L86-L89

but there wouldn't be any lambda expressions in our documentation. That's only there for completeness.

Similarly, we don't actually have expressions in our documentation, but if we did, call would be missing keywords:

https://github.com/scikit-hep/awkward/blob/8d7a2c31f1ea24cef8a909fb12bb14eb812477ac/docs/prepare_docstrings.py#L61-L64

For reference, here is how the `Lambda` and `Call` AST nodes are formatted:

```python
>>> print(ast.dump(ast.parse("lambda a, /, b, c=1, *, d=2: None"), indent=4))
Module(
    body=[
        Expr(
            value=Lambda(
                args=arguments(
                    posonlyargs=[
                        arg(arg='a')],
                    args=[
                        arg(arg='b'),
                        arg(arg='c')],
                    kwonlyargs=[
                        arg(arg='d')],
                    kw_defaults=[
                        Constant(value=2)],
                    defaults=[
                        Constant(value=1)]),
                body=Constant(value=None)))],
    type_ignores=[])
```

and

```python
>>> print(ast.dump(ast.parse("f(a, b=1, *args, **kwargs)"), indent=4))
Module(
    body=[
        Expr(
            value=Call(
                func=Name(id='f', ctx=Load()),
                args=[
                    Name(id='a', ctx=Load()),
                    Starred(
                        value=Name(id='args', ctx=Load()),
                        ctx=Load())],
                keywords=[
                    keyword(
                        arg='b',
                        value=Constant(value=1)),
                    keyword(
                        value=Name(id='kwargs', ctx=Load()))]))],
    type_ignores=[])
```

But I think only the function signatures (in the `dosig` function) need to be addressed.

(This, by the way, is a byproduct of the project starting in Python 2—no position-only or keyword-only arguments—and ReadTheDocs, which couldn't compile the C++ part and therefore couldn't actually `import awkward`, but it could compile the Python and therefore examine its AST. It could be resolved, in the long run, by replacing this now-unnecessary AST pass, but the above is a very quick fix to the current problem.)

agoose77:
Fixes #2073

### DIFF
--- a/docs/prepare_docstrings.py
+++ b/docs/prepare_docstrings.py
@@ -143,8 +143,8 @@ def dosig(node):
     if node is None:
         return "self"
     else:
-        non_keword_args = node.args.posonlyargs + node.args.args
-        argnames = [x.arg for x in non_keword_args + node.args.kwonlyargs]
+        non_keyword_args = node.args.posonlyargs + node.args.args
+        argnames = [x.arg for x in non_keyword_args + node.args.kwonlyargs]
         defaults = [
             "=" + tostr(x)
             for x in node.args.defaults + node.args.kw_defaults
@@ -153,9 +153,9 @@ def dosig(node):
         defaults = [""] * (len(argnames) - len(defaults)) + defaults
         rendered = [x + y for x, y in zip(argnames, defaults)]
         if node.args.vararg is not None:
-            rendered.insert(len(non_keword_args), f"*{node.args.vararg.arg}")
+            rendered.insert(len(non_keyword_args), f"*{node.args.vararg.arg}")
         elif node.args.kwonlyargs:
-            rendered.insert(len(non_keword_args), "*")
+            rendered.insert(len(non_keyword_args), "*")
         return ", ".join(rendered)
 
 
--- a/docs/prepare_docstrings.py
+++ b/docs/prepare_docstrings.py
@@ -143,12 +143,19 @@ def dosig(node):
     if node is None:
         return "self"
     else:
-        argnames = [x.arg for x in node.args.posonlyargs + node.args.args + node.args.kwonlyargs]
-        defaults = ["=" + tostr(x) for x in node.args.defaults + node.args.kw_defaults if x is not None]
+        non_keword_args = node.args.posonlyargs + node.args.args
+        argnames = [x.arg for x in non_keword_args + node.args.kwonlyargs]
+        defaults = [
+            "=" + tostr(x)
+            for x in node.args.defaults + node.args.kw_defaults
+            if x is not None
+        ]
         defaults = [""] * (len(argnames) - len(defaults)) + defaults
         rendered = [x + y for x, y in zip(argnames, defaults)]
         if node.args.vararg is not None:
-            rendered.insert(len(node.args.posonlyargs + node.args.args), f"*{node.args.vararg.arg}")
+            rendered.insert(len(non_keword_args), f"*{node.args.vararg.arg}")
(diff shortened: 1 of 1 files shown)


docs/prepare_docstrings.py
```diff
@@ -143,10 +143,13 @@ def dosig(node):
...
     else:
-        argnames = [x.arg for x in node.args.args]
-        defaults = ["=" + tostr(x) for x in node.args.defaults]
+        argnames = [x.arg for x in node.args.posonlyargs + node.args.args + node.args.kwonlyargs]
+        defaults = ["=" + tostr(x) for x in node.args.defaults + node.args.kw_defaults if x is not None]
         defaults = [""] * (len(argnames) - len(defaults)) + defaults
-        return ", ".join(x + y for x, y in zip(argnames, defaults))
+        rendered = [x + y for x, y in zip(argnames, defaults)]
+        if node.args.vararg is not None:
+            rendered.insert(len(node.args.posonlyargs + node.args.args), f"*{node.args.vararg.arg}")
```
agoose77: This should handle the `args` in
```python
def func(self, x, *args, y):
    ...
```

jpivarski (approved):
It does:

  * http://preview.awkward-array.org.s3-website.us-east-1.amazonaws.com/agoose77/docs-fix-keyword-only-rendering/reference/generated/ak.to_parquet.html
  * http://preview.awkward-array.org.s3-website.us-east-1.amazonaws.com/agoose77/docs-fix-keyword-only-rendering/reference/generated/ak.forms.RegularForm.html

jpivarski:
I'm enabling auto-merge on this one because it seems very straightforward and done. Many of the tests haven't started, but I suspect that's because we're using the maximum number that we're allowed to use. If it's held up by the link-checker, then this branch will need to be updated when that goes into `main`.

agoose77:
This won't run all the tests because we don't run build tests if the code isn't touched. If the non-queued tests all pass, I'll merge.

PR #2075, opened 2023-01-05
ci: remove link checker

agoose77:
Discussed with Jim: checking links is useful, but the benefits do not outweigh the drawbacks RE flukiness & build time. In general, links break; we just need to run these tests locally on a schedule to ensure that nothing breaks due to mistyping.


jpivarski (approved):
I see that this removes the link-checker (`sphinx-build -M linkcheck`) that we talked about on Slack.

What's the new `prepare-release.yml`? Is it something that we should run manually? It doesn't have the link-checker in it.

Oh! It sets version numbers. That's useful. Was it intended as part of this PR—is it ready?

agoose77:
> What's the new prepare-release.yml? Is it something that we should run manually? It doesn't have the link-checker in it.

@jpivarski :flushed: glad you caught that. I don't know how that snuck in there; I always use `git add --update` or manual staging to avoid accidental commits!

PR #2077, opened 2023-01-06
fix: support unzipping `ak.Record`

### LINKED ISSUE #2076: `ak.unzip(record)` fails with exception
### Version of Awkward Array

main

```python3
record = ak.Record({'x': 1, 'y': 2})
print(ak.unzip(record))
```

agoose77:
Fixes #2076 by adding an `allow_other=False` argument to `ak._util.unzip`. Most of the time we want to only be wrapping layouts, but it might occasionally be the case that an operation can legitimately produce a scalar. 

`allow_other=True` is set only for `ak.unzip` with this PR.

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -427,9 +427,11 @@ def behavior_of(*arrays, **kwargs):
     return behavior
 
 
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
Great! I'll merge it.

PR #2079, opened 2023-01-06
fix: don't try to re-wrap `array_function` overload results

### LINKED ISSUE #2078: `np.broadcast_arrays` array function overload has unknown result type
### Version of Awkward Array

main

Our generic dispatch for array function always tries to wrap the result as a layout. For overloads that awkward defines, we should trust that the result is already a layout.

agoose77:
Our integration with NumPy's array function mechanism means that users can in some cases invoke NumPy-only functions on Awkward Arrays (if they're rectilinear). However, our existing support for this mechanism is breaking Awkward overloads such as `ak.broadcast_arrays`, because it assumes that the result should be a layout.

In the long run, we could apply an inverse `_to_rectilinear` transform to recover high-level `ak.Array`/`ak.Record`s, but in this instance, I think the solution is even more clear cut.

If Awkward has an overload for an array function, then it should already be returning high-level objects. Therefore, we don't need to re-wrap the result. This PR applies such a fix, which means that any NumPy-only overloads that return non-arrays will still need a fix. However, this is a simple first step.

Fixes #2078

### DIFF
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -50,15 +50,16 @@ def array_function(func, types, args, kwargs, behavior):
         rectilinear_args = tuple(_to_rectilinear(x) for x in args)
         rectilinear_kwargs = {k: _to_rectilinear(v) for k, v in kwargs.items()}
         result = func(*rectilinear_args, **rectilinear_kwargs)
+        # We want the result to be a layout (this will fail for functions returning non-array convertibles)
+        out = ak.operations.ak_to_layout._impl(
+            result, allow_record=True, allow_other=True
+        )
+        if isinstance(out, (ak.contents.Content, ak.record.Record)):
+            return ak._util.wrap(out, behavior=behavior)
+        else:
+            return out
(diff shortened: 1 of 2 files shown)


jpivarski (commented):
~~Okay, got it. This is looking good so far. I'll take another look when it's out of draft mode (remind me).~~

Nope, it is out of draft mode! (I was thinking of a different PR.) I did take a good look at this and it makes sense. It is mergeable as-is, with the understanding that it's only for a simple case and that more is needed to solve the problem overall.

agoose77:
Right. It doesn't try to solve the NumPy side of things, e.g. if we didn't define `ak.broadcast_arrays`, the result of calling `np.broadcast_arrays` would fail in the same manner. 

We could be smarter there, but I also think there comes a point where we _need_ to write an overload to handle the dynamicism of the NumPy API. That said, exploring returned tuples / lists and interpreting `np.ndarray` is definitely possible. Thankfully, we would still want this PR if we do decide to go that extra step in the future.

PR #2083, opened 2023-01-06
fix: correct NumPy zero-size broadcasting

### LINKED ISSUE #2082: Broadcasting a zero-length dimensions differs from NumPy
### Version of Awkward Array

main

`np.broadcast_shapes((4, 0, 3), (1, 1, 3))` produces the shape `(4, 0, 3)`. The equivalent operation with `ak.broadcast_arrays` does not succeed.

We don't have to make this work for broadcasting ragged arrays against regular arrays, in my view.

agoose77:
Fixes #2082, i.e. zero-size broadcasting:
```python
import numpy as np
import awkward as ak


x = np.arange(4 * 5 * 9).reshape(4, 5, 9)
result = np.broadcast_arrays(x[:, :0, ...], x[:, :1, ...])

y = ak.from_numpy(x)
result_ak = ak.broadcast_arrays(y[:, :0, ...], y[:, :1, ...])

assert ak._util.arrays_approx_equal(
    result, result_ak
)
```

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -670,18 +670,34 @@ def apply_step(
                 isinstance(x, RegularArray) or not isinstance(x, listtypes)
                 for x in inputs
             ):
-                maxsize = max(x.size for x in inputs if isinstance(x, RegularArray))
+                # Ensure all layouts have same length
+                length = None
+                for x in inputs:
+                    if isinstance(x, Content):
+                        if length is None:
+                            length = x.length
+                        elif backend.nplike.known_shape:
+                            assert length == x.length
+                assert length is not None
 
                 if backend.nplike.known_data:
+                    if any(x.size == 0 for x in inputs if isinstance(x, RegularArray)):
+                        dimsize = 0
+                    else:
+                        dimsize = max(
+                            x.size for x in inputs if isinstance(x, RegularArray)
+                        )
+
                     for x in inputs:
                         if isinstance(x, RegularArray):
-                            if maxsize > 1 and x.size == 1:
+                            if dimsize > 1 and x.size == 1:
+                                # For any (N, 1) array, we know we'll broadcast to (N, M) where M is maxsize
                                 tmpindex = Index64(
                                     backend.index_nplike.repeat(
                                         backend.index_nplike.arange(
                                             x.length, dtype=np.int64
                                         ),
-                                        maxsize,
+                                        dimsize,
                                     ),
                                     nplike=backend.index_nplike,
                                 )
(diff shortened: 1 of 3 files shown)


src/awkward/_util.py
```diff
@@ -931,7 +931,10 @@ def visitor(left, right) -> bool:
                 and (left.is_tuple == right.is_tuple)
                 and all([visitor(x, y) for x, y in zip(left.contents, right.contents)])
             )
-        elif left.is_empty:
+        elif left.is_unknown:
```
agoose77, line 934: Unrelated to this PR, but this PR exposed the bug.

src/awkward/_broadcasting.py
```diff
@@ -670,39 +670,55 @@ def continuation():
...
-                maxsize = max(x.size for x in inputs if isinstance(x, RegularArray))
+                # Ensure all layouts have same length
+                length = None
+                for x in inputs:
+                    if isinstance(x, Content):
+                        if length is None:
+                            length = x.length
+                        elif backend.nplike.known_shape:
+                            assert length == x.length
+                assert length is not None
```
agoose77, line 681: I moved this for readability

src/awkward/_broadcasting.py
```diff
@@ -670,39 +670,55 @@ def continuation():
...
-                            if maxsize > 1 and x.size == 1:
+                            if dimsize > 1 and x.size == 1:
                                 nextinputs.append(
                                     x.content[: x.length * x.size]._carry(
                                         tmpindex, allow_lazy=False
                                     )
                                 )
-                            elif x.size == maxsize:
+                            elif x.size == dimsize:
                                 nextinputs.append(x.content[: x.length * x.size])
```
agoose77, line 713: If `dimsize == 0`, then this is a zero-length slice (which is what we want), so the code does not need to change

src/awkward/_broadcasting.py
```diff
@@ -670,39 +670,55 @@ def continuation():
...
+                            elif dimsize == 0:
+                                nextinputs.append(x.content[:0])
```
agoose77, line 715: `size != 0` arrays when `dimsize == 0` will encounter this case. We also want these to be zero-length.

src/awkward/_broadcasting.py
```diff
@@ -670,39 +670,55 @@ def continuation():
...
+                for x in inputs:
+                    if isinstance(x, Content):
+                        if length is None:
+                            length = x.length
+                        elif backend.nplike.known_shape:
+                            assert length == x.length
+                assert length is not None
+
+                if any(x.size == 0 for x in inputs if isinstance(x, RegularArray)):
+                    dimsize = 0
```
agoose77, line 684: Any zero means all zeros!

jpivarski: Right, the original motivation for the `max` was because the target size is "the one that isn't `1`." But if there's any `0`, then it broadcasts to `0`.

jpivarski (approved):
Got it. Thanks for catching this!

In my opinion, it can be merged.

PR #2084, opened 2023-01-06
fix: remove Long64_t from common header-only

### LINKED ISSUE #2032: C++ for JIT-compilation (RDataFrame and cppyy) shouldn't be in `awkward/header-only`.
...because changes to what gets JIT-compiled (like #2023) shouldn't require a new `awkward-cpp` package to get precompiled. In fact, it's a little dangerous:

https://github.com/scikit-hep/awkward/blob/36d5a5600da35f34b436ae6ef7325839279a9b71/header-only/awkward/utils.h#L67-L71

is in the header-only code, and therefore what third party projects might pick up and use in their compilations, but `Long64_t` is not a symbol that exists outside of ROOT, as far as I can tell. Somehow, we manage to compile GrowableBuffer and ArrayBuilder without errors, despite the fact that we don't include ROOT headers in our compilation. I actually don't know why that works (C++ template bindings are very late?).

The code in `awkward/header-only` should only include

  * GrowableBuffer
  * LayoutBuilder
  * BuilderOptions

If the RDataFrame/cppyy code _uses_ LayoutBuilder in the future, that's fine, but there can't be anything in this directory that is not used by at least one of the above or anything that is defined in ROOT headers.

There's a directory for C++ utilities for RDataFrame/cppyy:

https://github.com/scikit-hep/awkward/tree/main/src/awkward/_connect/rdataframe/include/rdataframe

Splitting the two could even involve copying some code—that would be fine, if needed.

ianna:
- fixes issue #2032 
- enable boolean type test


header-only/awkward/utils.h
```diff
@@ -12,14 +12,17 @@
...
 namespace awkward {
 
   /// @brief Returns the name of a primitive type as a string.
   template <typename T>
   const std::string
   type_to_name() {
-    std::cout << "Type " << typeid(T).name() << " is not recognized." << std::endl;
-    return typeid(T).name();
+    if(std::is_same_v<T, long long>) {
+    	return "int64";
```
ianna: We may want to consider extending this to cover other ROOT type aliases if users request it: https://root.cern/doc/master/RtypesCore_8h_source.html

ianna (commented):
@jpivarski - I think, using a fundamental type here is ok for either ROOT users or those who don't have it installed. The `Long64_t` is a type alias for a `long` type (that is also an `int64_t`) on Mac, but it's a `long long` on Ubuntu.

header-only/awkward/utils.h
```diff
@@ -12,14 +12,17 @@
...
+    }
+    return "unsupported_primitive_type";
```
jpivarski: According to C99 (I don't know about C++), `long long` has _at least_ 64 bits, so it wouldn't be safe to map from `long long` to 64 bits. (Not that there are many platforms that would interpret any built-in type in C as 128 bits!)

At least in principle, then, it should be safer to say

```python
    if (std::is_same_v<T, int64_t>) {
    	return "int64";
    }
```

Would that work? I'm assuming that the `std::is_same_v<T1, T2>` tells you whether types `T1` and `T2` are functionally equivalent, rather than the same name, right? (Also, doesn't it need a `()` at the end, since `std::is_same_v<T1, T2>` by itself is just the name of a type, not a value for `if` to recognize as boolean, and does it have the `_v` at the end?).

I noticed that you already have a way to map `int64_t` to `"int64`":

https://github.com/scikit-hep/awkward/blob/dbefdb638a9ea782c518d574e5baf3e95a2fc4a0/header-only/awkward/utils.h#L60-L66

but it's not triggering, possibly because `long long` and `int64_t` are only functionally equivalent on this platform, but don't have the same name. Presumably, that's how the template instantiation is resolving.

Is the `std::is_same_v<T1, T2>` working because it's a runtime check? If so, maybe all of the integer types (by size/functional equivalence, not all the different names and aliases in RtypesCore) should have fallbacks:

```suggestion
    if (std::is_same_v<T, bool>) {
    	return "bool";
    }
    else if (std::is_same_v<T, int8_t>) {
    	return "int8";
    }
    else if (std::is_same_v<T, int16_t>) {
    	return "int16";
    }
    else if (std::is_same_v<T, int32_t>) {
    	return "int32";
    }
    else if (std::is_same_v<T, int64_t>) {
    	return "int64";
    }
    else if (std::is_same_v<T, uint8_t>) {
    	return "uint8";
    }
    else if (std::is_same_v<T, uint16_t>) {
    	return "uint16";
    }
    else if (std::is_same_v<T, uint32_t>) {
    	return "uint32";
    }
    else if (std::is_same_v<T, uint64_t>) {
    	return "uint64";
    }
    else if (std::is_same_v<T, float>) {
    	return "float";
    }
    else if (std::is_same_v<T, double>) {
    	return "double";
    }
    else if (std::is_same_v<T, bool>) {
    	return "bool";
    }
    else if (std::is_same_v<T, char>) {
    	return "char";
    }
    else if (std::is_same_v<T, std::complex<float>>) {
    	return "complex64";
    }
    else if (std::is_same_v<T, std::complex<double>>) {
    	return "complex128";
    }
    else {
        return "unsupported_primitive_type";
    }
```

And if the runtime type-check works, is there really a need for a compile-time type check? This `type_to_name` function is likely only called once per RDataFrame by ROOT; it's not like it needs to be fast. (Similarly for `type_to_numpy_like`.) I'll leave that decision up to you. My preference is to reduce the duplication if `std::is_same_v` works better than template instantiation ("better" in the sense of recognizing `Long64_t` as `long long` and therefore `int64_t`, on platforms for which that is true).

header-only/awkward/utils.h
```diff
@@ -62,14 +65,6 @@ namespace awkward {
     return "int64";
   }
 
-  /// @brief Returns `int64` string when the primitive type
-  /// is a 64-bit signed integer.
-  template <>
-  const std::string
-  type_to_name<Long64_t>() {
-    return "int64";
-  }
```
jpivarski, line 71: I see: the `std::is_same_v` _replaced_ this template instantiation. I'm just suggesting doing it for all the types. Running down the list of integer widths is a finite job: we don't need to worry about any integer widths that don't have 8, 16, 32, or 64 bits, signed or unsigned. Including all the possible _renamings_ of these types is not a finite job.

jpivarski (approved):
Looks good! I'd accept it as it is, but I made a suggestion to consider before merging it. Whether you decide to accept the suggestion or not, you can merge this PR.

If you merge it today (i.e. in the next few hours), then it will be included in the next `awkward-cpp` release. Otherwise, we'll do another `awkward-cpp` release later.

header-only/awkward/utils.h
ianna: done.I think, it covers almost all types.

ianna (commented):
@jpivarski - please, check. I'm done with this PR for today. Thanks!

header-only/awkward/utils.h
```diff
@@ -12,135 +12,108 @@
...
+      /// is a 64-bit signed integer.
+      return "int64";
+    } else if(std::is_same_v<T, unsigned long long int>) {
+      /// @brief Returns `uint64` string when the primitive type
+      /// is a 64-bit unsigned integer.
+      return "uint64";
+    } else {
+      return "unsupported_primitive_type";
+    }
   }
```
jpivarski: To avoid platform dependence, we have to _not_ specify these in terms of C builtin types. For instance, `signed char` is defined as "at least 8 bits" and its `if` statement comes before the larger types. So if this code runs on a platform in which `signed char` has 16 bits, `type_to_name` would return the string `"int8"` and something would break (possibly without an explicit error message, unless the length is also constrained).

And since `std::is_same_v<T, bool>` checks compatibility/equivalence, rather than exact alias names, testing for `bool` in an `if` statement before `int8_t` would result in any 8-bit data being presented as booleans. That might include ASCII character strings or union tags, which would be very wrong. So for the types that have to be distinguished by exact alias names, I'll reinstate the template-instantiation method.

```suggestion
    if (std::is_same_v<T, int8_t>) {
      return "int8";
    }
    else if (std::is_same_v<T, uint8_t>) {
      return "uint8";
    }
    else if (std::is_same_v<T, int16_t>) {
      return "int16";
    }
    else if (std::is_same_v<T, uint16_t>) {
      return "uint16";
    }
    else if (std::is_same_v<T, int32_t>) {
      return "int32";
    }
    else if (std::is_same_v<T, uint32_t>) {
      return "uint32";
    }
    else if (std::is_same_v<T, int64_t>) {
      return "int64";
    }
    else if (std::is_same_v<T, uint64_t>) {
      return "uint64";
    }
    else if (std::is_same_v<T, float>) {
      return "float32";
    }
    else if (std::is_same_v<T, double>) {
      return "float64";
    }
    else if (std::is_same_v<T, std::complex<float>>) {
      return "complex64";
    }
    else if (std::is_same_v<T, std::complex<double>>) {
      return "complex128";
    }
    else {
      return "unsupported_primitive_type";
    }
  }

  template <>
  const std::string
  type_to_name<bool>() {
    // This takes precedence over the unspecialized template, and therefore any
    // 8-bit data that is not named bool will be mapped to "int8" or "uint8".
    return "bool";
  }
```

jpivarski (approved):
See below. I'll try committing this change and auto-merging.

jpivarski:
```
E               TypeError: while calling
E               
E                   ak.from_rdataframe(
E                       rdf = RInterface<ROOT::Detail::RDF::RLoopManager,void>-instance
E                       columns = 'x'
E                   )
E               
E               Error details: unrecognized primitive: 'unsupported_primitive_type'. Must be one of
E               
E                   bool, int8, uint8, int16, uint16, int32, uint32, int64, uint64, float32, float64, complex64, complex128, datetime64, timedelta64, float16, float128, complex256
E               
E               or a datetime64/timedelta64 with units (e.g. 'datetime64[15us]')

out        = None
primitive  = 'unsupported_primitive_type'
```

Okay, I'm going to check it out locally and see what's happening.

jpivarski:
> And since `std::is_same_v<T, bool>` checks compatibility/equivalence, rather than exact alias names...

`std::is_same_v<T, bool>` does not check for compatibility/equivalence. It checks for exact equality in some sense (possibly depends on how `stdint.h` was written):

```
root [0] std::is_same_v<Long64_t, int64_t>
(const bool) false
```

But this should do it:

```
root [1] std::is_integral_v<Long64_t> && std::is_signed_v<Long64_t> && sizeof(Long64_t) == 8
(bool) true
```

I'm updating the code.

jpivarski:
This is working for me, locally:

https://github.com/scikit-hep/awkward/blob/7929f0b740378668dedd5e42ebe62a6fa1f35bad/header-only/awkward/utils.h#L19-L77

header-only/awkward/utils.h
```diff
@@ -11,136 +11,71 @@
...
-    return "float64";
-  }
-
-  /// @brief Returns `char` string when the primitive type
-  /// is a character.
-  template <>
-  const std::string
-  type_to_name<char>() {
-    return "char";
-  }
```
ianna, line 127: @jpivarski - <strike>I think, we might need this one as well.</strike>
Perfect! There is no problem extracting the columns with an `std::string` type:
```python
ak_array_in_str = ak.Array(["True", "False", "True", "True", "True"])
>>> ak_array_in_str
<Array ['True', 'False', 'True', 'True', 'True'] type='5 * string'>
>>> data_frame_str = ak.to_rdataframe({"x": ak_array_in_str})
>>> data_frame_str.GetColumnType("x")
'std::string'
>>> data_frame_str.Display().Print()
+-----+----------------+---------+
| Row | awkward_index_ | x       | 
+-----+----------------+---------+
| 0   | 0              | "True"  | 
+-----+----------------+---------+
| 1   | 1              | "False" | 
+-----+----------------+---------+
| 2   | 2              | "True"  | 
+-----+----------------+---------+
| 3   | 3              | "True"  | 
+-----+----------------+---------+
| 4   | 4              | "True"  | 
+-----+----------------+---------+
>>> ak_array_out = ak.from_rdataframe(data_frame_str, "x")
>>> ak_array_out
<Array ['True', 'False', 'True', 'True', 'True'] type='5 * string'>
>>> ak_array_out.layout.content
<NumpyArray dtype='uint8' len='21'>
    [ 84 114 117 101  70  97 108 115 101  84 114 117 101  84 114 117 101  84
     114 117 101]
</NumpyArray>
>>> ak_array_in_str.layout.content
<NumpyArray dtype='uint8' len='21'>
    [ 84 114 117 101  70  97 108 115 101  84 114 117 101  84 114 117 101  84
     114 117 101]
</NumpyArray>

```

ianna (commented):
@jpivarski - thanks! It looks great. I will try to run more extensive tests to check the containers of strings/chars.

ianna:
> > And since `std::is_same_v<T, bool>` checks compatibility/equivalence, rather than exact alias names...
> 
> `std::is_same_v<T, bool>` does not check for compatibility/equivalence. It checks for exact equality in some sense (possibly depends on how `stdint.h` was written):
> 
> ```
> root [0] std::is_same_v<Long64_t, int64_t>
> (const bool) false
> ```
> 
yes, because in this case:
```
root [0] std::is_same_v<Long64_t, int64_t>
(const bool) false
root [1] std::is_same_v<Long64_t, long long>
(const bool) true
root [2] std::is_same_v<int64_t, long>
(const bool) true
```
> But this should do it:
> 
> ```
> root [1] std::is_integral_v<Long64_t> && std::is_signed_v<Long64_t> && sizeof(Long64_t) == 8
> (bool) true
> ```
> 
> I'm updating the code.

PR #2087, opened 2023-01-07
fix: broken links due to cpp split

### LINKED ISSUE #2081: `awkward-cpp` debug URLs are broken
### Version of Awkward Array

main

This bug is caused by two separate things:
1. We're not reading the correct variable into the `VERSION_INFO` compile definition
2. We don't generate a tag that corresponds to the `awkward-cpp` version

I know that we have discussed this previously, but I think this provides a good-enough reason to tag the repository upon release of awkward-cpp; it is helpful to be able to look at the repo as-it-was for a particular C++ release. In order to keep the C++ package and Python package decoupled, it would be better to also tag the C++ releases rather than use the Python version (which we know is also released at the same time as a C++ release, or near enough).

agoose77:
Fixes #2081 

We will also need to update our Git workflow such that C++ releases are assigned tags.

### DIFF
--- a/awkward-cpp/include/awkward/common.h
+++ b/awkward-cpp/include/awkward/common.h
@@ -30,7 +30,7 @@
 
 #define QUOTE(x) #x
 
-#define FILENAME_FOR_EXCEPTIONS_C(filename, line) "\n\n(https://github.com/scikit-hep/awkward-1.0/blob/awkward-cpp-" VERSION_INFO "/awkward-cpp/" filename "#L" #line ")"
+#define FILENAME_FOR_EXCEPTIONS_C(filename, line) "\n\n(https://github.com/scikit-hep/awkward/blob/awkward-cpp-" VERSION_INFO "/awkward-cpp/" filename "#L" #line ")"
 #define FILENAME_FOR_EXCEPTIONS(filename, line) std::string(FILENAME_FOR_EXCEPTIONS_C(filename, line))
 
 #ifdef __GNUC__
--- a/awkward-cpp/include/awkward/common.h
+++ b/awkward-cpp/include/awkward/common.h
@@ -30,9 +30,8 @@
 
 #define QUOTE(x) #x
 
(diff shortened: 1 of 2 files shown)


agoose77:
The branch preview `try-it` link can be used to verify that the following links to the correct URL:
```python
ak.to_regular([
    [1,2,3],
    [4,5]
])

File: awkward-cpp/include/awkward/common.h [hunk 1]
```diff
@@ -30,9 +30,8 @@
 
 #define QUOTE(x) #x
 
-#define FILENAME_FOR_EXCEPTIONS(filename, line) std::string("\n\n(https://github.com/scikit-hep/awkward-1.0/blob/" VERSION_INFO "/" filename "#L" #line ")")
-#define FILENAME_FOR_EXCEPTIONS_C(filename, line) ("\n\n(https://github.com/scikit-hep/awkward-1.0/blob/" VERSION_INFO "/" filename "#L" #line ")")
-#define FILENAME_FOR_EXCEPTIONS_CUDA(filename, line) ("\n\n(https://github.com/scikit-hep/awkward-1.0/blob/" QUOTE(VERSION_INFO) "/" filename "#L" #line ")")
+#define FILENAME_FOR_EXCEPTIONS_C(filename, line) "\n\n(https://github.com/scikit-hep/awkward-1.0/blob/awkward-cpp-" VERSION_INFO "/awkward-cpp/" filename "#L" #line ")"
```
jpivarski: We haven't done a general update of URLs within the codebase, but while we're touching this one, we might as well fix it. (This one, in particular, is visible to users.)

```suggestion
#define FILENAME_FOR_EXCEPTIONS_C(filename, line) "\n\n(https://github.com/scikit-hep/awkward/blob/awkward-cpp-" VERSION_INFO "/awkward-cpp/" filename "#L" #line ")"
```

jpivarski (approved):
Looks great! With this one small fix:

awkward-cpp/include/awkward/common.h
agoose77: oop, yep!

PR #2089, opened 2023-01-08
fix: implement explicit translation for NEP-18

### LINKED ISSUE #2088: NEP 18 overloaded functions don't have same function signature
### Version of Awkward Array

main

Our `ak.*` NEP-18 overloaded functions do not directly copy the NumPy function signatures, e.g.:

```python
@ak._connect.numpy.implements("std")
def std(
    x,
    weight=None,
    ddof=0,
    axis=None,
    *,
    keepdims=False,
    mask_identity=False,
    flatten_records=unset,
):
```
vs NumPy's
```python
@array_function_dispatch(_std_dispatcher)
def std(a, axis=None, dtype=None, out=None, ddof=0, keepdims=np._NoValue, *,
        where=np._NoValue):
```

Understandably, we don't want to have NumPy-only parameters in our signatures, e.g. `out=`. However, this means that positional arguments are not guaranteed to match-up with Awkward. For example in this case, passing `axis` by position to `np.std` would not work, i.e.
```pycon
>>> np.std(ak.to_numpy(x), -1)
0.816496580927726
>>> np.std(x, -1)
  File "/home/angus/.mambaforge/envs/awkward/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3442, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-9-0b53e1f1071f>", line 1, in <module>
    np.std(x, -1)
  File "<__array_function__ internals>", line 200, in std
  File "/home/angus/Git/awkward/src/awkward/highlevel.py", line 1411, in __array_function__
  File "/home/angus/Git/awkward/src/awkward/_connect/numpy.py", line 59, in array_function
    return function(*args, **kwargs)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_std.py", line 84, in std
    return _impl(x, weight, ddof, axis, keepdims, mask_identity)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_std.py", line 168, in _impl
    ak.operations.to_layout(weight, allow_record=False, allow_other=False),
  File "/home/angus/Git/awkward/src/awkward/operations/ak_to_layout.py", line 39, in to_layout
    return _impl(array, allow_record, allow_other)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_to_layout.py", line 124, in _impl
    raise _errors.wrap_error(
TypeError: while calling
    ak.std(
        x = <Array [[1, 2, 3]] type='1 * var * int64'>
        weight = -1
        ddof = 0
        axis = None
        keepdims = False
        mask_identity = False
    )
Error details: -1 cannot be converted into an Awkward Array
```

Most of the time this isn't a problem; in practice, the NumPy `out` parameter appears early in the argument list, which immediately breaks any subsequent positional correspondence.

I think, in general, we can improve this. If we bind the parameters to the signature of the _NumPy_ function, we can then pass them into the appropriate Awkward order. If we assert that the Awkward function has no position-only parameters, then this would be trivial; pass in the bound parameters as keyword arguments.

agoose77:
This PR closes #2088 by defining explicit NEP-18 translations ("implementations"). These map the NumPy argument spec onto the Awkward function, and translate incompatible arguments (e.g. NumPy's `kind` vs Awkward's `stable` for `ak.sort`).

The TLDR of the motivation for this PR is that some NumPy functions like `np.std` have incompatible differences with our implementations e.g. `ak.std`. By explicitly defining a translation, we can decouple the two APIs.

This adds a slight runtime cost; now a call to `np.sort()` jumps through two additional functions. However, this is not an area of performance that we should care about (and CPython 3.11 speeds function calls up slightly, so it's a solved problem /s).

We could also use this mechanism to define the translations for non-Awkward overloaded NumPy functions, which just need better translations that our default heuristic-based approach takes.

### DIFF
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -1,4 +1,6 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+import functools
+import inspect
 
 import numpy
 
@@ -53,24 +55,45 @@ def _to_rectilinear(arg):
         return arg
 
 
+def _array_function_no_impl(func, types, args, kwargs, behavior):
+    rectilinear_args = tuple(_to_rectilinear(x) for x in args)
+    rectilinear_kwargs = {k: _to_rectilinear(v) for k, v in kwargs.items()}
+    result = func(*rectilinear_args, **rectilinear_kwargs)
+    # We want the result to be a layout (this will fail for functions returning non-array convertibles)
+    out = ak.operations.ak_to_layout._impl(result, allow_record=True, allow_other=True)
+    return ak._util.wrap(out, behavior=behavior, allow_other=True)
+
+
 def array_function(func, types, args, kwargs, behavior):
     function = implemented.get(func)
+    # Use NumPy's implementation
     if function is None:
-        rectilinear_args = tuple(_to_rectilinear(x) for x in args)
-        rectilinear_kwargs = {k: _to_rectilinear(v) for k, v in kwargs.items()}
-        result = func(*rectilinear_args, **rectilinear_kwargs)
-        # We want the result to be a layout (this will fail for functions returning non-array convertibles)
-        out = ak.operations.ak_to_layout._impl(
-            result, allow_record=True, allow_other=True
-        )
-        return ak._util.wrap(out, behavior=behavior, allow_other=True)
+        return _array_function_no_impl(func, types, args, kwargs, behavior)
     else:
         return function(*args, **kwargs)
 
 
 def implements(numpy_function):
     def decorator(function):
-        implemented[getattr(numpy, numpy_function)] = function
+        signature = inspect.signature(function)
+        unsupported_names = {
+            p.name for p in signature.parameters.values() if p.default is unsupported
+        }
+
+        @functools.wraps(function)
+        def ensure_valid_args(*args, **kwargs):
+            parameters = signature.bind(*args, **kwargs)
+            provided_invalid_names = parameters.arguments.keys() & unsupported_names
+            if provided_invalid_names:
+                names = ", ".join(provided_invalid_names)
+                raise ak._errors.wrap_error(
+                    TypeError(
+                        f"Awkward NEP-18 overload was provided with unsupported argument(s): {names}"
+                    )
+                )
+            return function(*args, **kwargs)
+
+        implemented[getattr(numpy, numpy_function)] = ensure_valid_args
         return function
 
     return decorator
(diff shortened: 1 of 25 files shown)


jpivarski (approved):
Thanks for defining this translation layer. You mention performance from indirection, but that ought to be pretty minimal, while the benefits of a translation layer are pretty significant.

We'd like to synchronize the Awkward arguments and the NumPy arguments as much as possible, though, because we want to minimize user surprise. For instance,

> NumPy's `kind` vs Awkward's `stable` for `ak.sort`

should eventually match NumPy's arguments better, though for sorting, I was under the impression that their arguments are in flux. (They don't want to explicitly say what algorithm they're using, such as `"heapsort"`.)

In this PR, it looks like you added a translation layer for every NEP-18 overload, not just the ones that have different arguments. No, I don't see any examples that have exactly the same arguments (and maybe we'd always have `highlevel` and `behavior`, while NumPy would never have these two arguments).

agoose77:
> You mention performance from indirection, but that ought to be pretty minimal, while the benefits of a translation layer are pretty significant.

Yes, I like to make it clear that this does have an impact, but only if users are doing things in a loop. It's more of a "take note" rather than "this is a regression".

> We'd like to synchronize the Awkward arguments and the NumPy arguments as much as possible, though, because we want to minimize user surprise. 

Definitely (in the long run). It should be a goal that we avoid un-motivated divergences. Sorting might be an aberration here, as you point out.

> In this PR, it looks like you added a translation layer for every NEP-18 overload, not just the ones that have different arguments.

The idea with these translation functions is that we're not directly coupled to NumPy's interface, which may have legacy arguments or other features that we don't support. If we do this for all functions, we'll keep some symmetry on our end. I think you're right, that we should try and use this predominantly for compatibility than for needlessly deviating from the NumPy API.

> No, I don't see any examples that have exactly the same arguments (and maybe we'd always have highlevel and behavior, while NumPy would never have these two arguments).

Yes, and the main culprit here is `out`, which is nearly always present in NumPy operations, and we _never_ define. This argument is guaranteed to break positional correspondence of subsequent arguments, so these custom NEP-18 translators (?) ensure that we can use the signature of the NumPy function exactly.

> should eventually match NumPy's arguments better, though for sorting, I was under the impression that their arguments are in flux. (They don't want to explicitly say what algorithm they're using, such as "heapsort".)

Should we make a change here (in deprecation cycles)?

jpivarski:
> Should we make a change here (in deprecation cycles)?

It doesn't have to be done now. Eventually, I'd like to replace the sorting kernels—which currently defer to C++ sorting algorithms—with something that would be more like what we'll need to do in CUDA, but CUDA sorting is something that will require extra thought in itself.

PR #2090, opened 2023-01-08
refactor:  remove `kind` and `order` args to sorting protocols

agoose77:
The `sort` and `argsort` internal Content protocols don't actually use the `kind` or `order` arguments. Awkward instead uses `stable` to determine the sorting algorithm used. This PR removes those unused arguments.


agoose77:
I've requested from @jpivarski too, because his comments on #2089 suggest that we might want to redefine our sorting API arguments.

jpivarski (approved):
This is related to, but independent of, #2112. This argument list is a relic of when `ak.sort`'s arguments were more different from `np.sort` than they are now.

jpivarski:
I'll be merging it without an explicit approval from @ianna. We were talking about the sort arguments today, and this is a detail of what we had agreed to.

PR #2093, opened 2023-01-09
fix: off-by-one error in `run_lengths`

agoose77:
`ak.run_lengths([[3, 3, 3, 3], [3]])` does not correctly consider the inner list boundary. This PR fixes the causative off-by-one error.

The cause of this bug is that the logic that forcibly splits runs across list boundaries was ignoring one-length sublists at the end of the list. This off-by-one error meant that runs were incorrectly computed across a boundary in such cases.

### DIFF
--- a/src/awkward/operations/ak_run_lengths.py
+++ b/src/awkward/operations/ak_run_lengths.py
@@ -115,7 +115,7 @@ def _impl(array, highlevel, behavior):
                 # To consider only the interior boundaries, we ignore the start and end
                 # offset values. These can be repeated with empty sublists, so we mask them out.
                 is_interior = backend.nplike.logical_and(
-                    0 < offsets, offsets < len(data) - 1
+                    0 < offsets, offsets < len(data)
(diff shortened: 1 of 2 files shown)


agoose77:
Hmm, this test isn't failing on `main`...

jpivarski (approved):
I was thinking that if this was something v1 got right and v2 (without this PR) gets wrong—which I just confirmed, using your test example—then we should be able to look at the v1 code and understand why.

Here's the relevant section of the v1 code:

https://github.com/scikit-hep/awkward/blob/80bbef0738a6b7928333d7c705ee1b359991de5b/src/awkward/operations/structure.py#L338-L342

The code here is pretty different now because of `is_interior`, introduced in #1795. I guess that means this PR is a bug-fix on that PR, right?

Anyway, it looks right now and I think it should be merged!

agoose77:
@jpivarski yes, don't run `git blame` ;)

PR #2097, opened 2023-01-09
fix: unflatten should accept non-packed `counts`

### LINKED ISSUE #2059: `ak.unflatten` error message doesn't indicate axis out of boudns
### Version of Awkward Array

main

Given an N-dimensional array for which `ak.unflatten(array, counts, axis=-1)` succeeds, passing `axis>N` does not inform the user that the axis is out of bounds, but rather:
```
Error details: structure imposed by 'counts' does not fit in the array or partition at axis=N
```

agoose77:
- Fixes #2059 
- Replaces use of `np.ma` with our own option handling
- Fixes loss of recursion

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -900,8 +900,10 @@ def arrays_approx_equal(
             return False
 
         if left.is_list:
-            return numpy.array_equal(left.offsets, right.offsets) and visitor(
-                left.content, right.content
+            return (
+                numpy.array_equal(left.starts, right.starts)
+                and numpy.array_equal(left.stops, right.stops)
+                and visitor(left.content, right.content)
             )
         elif left.is_regular:
             return (left.size == right.size) and visitor(left.content, right.content)
--- a/src/awkward/operations/ak_unflatten.py
+++ b/src/awkward/operations/ak_unflatten.py
@@ -98,10 +98,11 @@ def _impl(array, counts, axis, highlevel, behavior):
         ).to_packed()
         if counts.is_option and (counts.content.is_numpy or counts.content.is_unknown):
             mask = counts.mask_as_bool(valid_when=False)
-            counts = counts.to_numpy(allow_missing=True)
-            counts = ak._nplikes.numpy.ma.filled(counts, 0)
+            counts = backend.nplike.to_rectilinear(
+                ak.operations.fill_none(counts, 0, axis=-1, highlevel=False)
+            )
         elif counts.is_numpy or counts.is_unknown:
-            counts = counts.to_numpy(allow_missing=False)
+            counts = backend.nplike.to_rectilinear(counts)
             mask = False
         else:
             raise ak._errors.wrap_error(
@@ -120,7 +121,7 @@ def _impl(array, counts, axis, highlevel, behavior):
         current_offsets[0] = 0
         backend.index_nplike.cumsum(counts, out=current_offsets[1:])
 
-    def doit(layout):
+    def unflatten_this_layout(layout):
         nonlocal current_offsets
 
         if isinstance(counts, (numbers.Integral, np.integer)):
@@ -162,7 +163,7 @@ def _impl(array, counts, axis, highlevel, behavior):
         return out
 
     if axis == 0 or ak._util.maybe_posaxis(layout, axis, 1) == 0:
-        out = doit(layout)
+        out = unflatten_this_layout(layout)
 
     else:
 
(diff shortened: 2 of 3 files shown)


src/awkward/_util.py
```diff
@@ -900,8 +900,10 @@ def visitor(left, right) -> bool:
             return False
 
         if left.is_list:
-            return numpy.array_equal(left.offsets, right.offsets) and visitor(
-                left.content, right.content
+            return (
+                numpy.array_equal(left.starts, right.starts)
+                and numpy.array_equal(left.stops, right.stops)
+                and visitor(left.content, right.content)
```
agoose77, line 906: I'd thought that `offsets` was the shared interface (and it would not work well for `ListArray`), but of course that's nonsense; it's `starts`/`stops`. This change to the test helper fixes it.

src/awkward/operations/ak_unflatten.py
```diff
@@ -84,21 +84,25 @@ def unflatten(array, counts, axis=0, *, highlevel=True, behavior=None):
 
 
 def _impl(array, counts, axis, highlevel, behavior):
-    layout = ak.operations.to_layout(array, allow_record=False, allow_other=False)
+    layout = ak.operations.to_layout(
+        array, allow_record=False, allow_other=False
+    ).to_packed()
```
agoose77, line 89: Ultimately we need to know that at any `depth`, the `counts` computed from our externally-visible structure aligns with the current content. Therefore, we need most of the guarantees of `to_packed()`. This could be done in the recursion if we wanted better performance for the non `axis=-1` cases, because we don't need the leaf to be contiguous, or indeed care about any nodes after the unflatten depth.

src/awkward/operations/ak_unflatten.py
```diff
@@ -84,21 +84,25 @@ def unflatten(array, counts, axis=0, *, highlevel=True, behavior=None):
...
 
     if isinstance(counts, (numbers.Integral, np.integer)):
         current_offsets = None
     else:
-        counts = ak.operations.to_layout(counts, allow_record=False, allow_other=False)
-
-        if counts.is_option:
+        counts = ak.operations.to_layout(
+            counts, allow_record=False, allow_other=False
+        ).to_packed()
```
agoose77, line 98: `counts` should be a trivial layout, so this is reasonable to pack (we don't need contiguousness, but we do need a non-indexed layout).

src/awkward/operations/ak_unflatten.py
```diff
@@ -84,21 +84,25 @@ def unflatten(array, counts, axis=0, *, highlevel=True, behavior=None):
...
+        if counts.is_option and (counts.content.is_numpy or counts.content.is_unknown):
```
agoose77, line 99: @jpivarski's fix

src/awkward/operations/ak_unflatten.py
```diff
@@ -84,21 +84,25 @@ def unflatten(array, counts, axis=0, *, highlevel=True, behavior=None):
...
             mask = counts.mask_as_bool(valid_when=False)
-            counts = counts.to_numpy(allow_missing=True)
-            counts = ak._nplikes.numpy.ma.filled(counts, 0)
+            counts = backend.nplike.to_rectilinear(
+                ak.operations.fill_none(counts, 0, axis=-1, highlevel=False)
+            )
```
agoose77, line 103: This replaces the use of `to_numpy`, meaning we won't incur a needless device transfer under CUDA (IIRC).

src/awkward/operations/ak_unflatten.py
```diff
@@ -113,11 +117,13 @@ def _impl(array, counts, axis, highlevel, behavior):
         if not issubclass(counts.dtype.type, np.integer):
             raise ak._errors.wrap_error(ValueError("counts must be integers"))
 
-        current_offsets = [backend.index_nplike.empty(len(counts) + 1, np.int64)]
-        current_offsets[0][0] = 0
-        backend.index_nplike.cumsum(counts, out=current_offsets[0][1:])
+        current_offsets = backend.index_nplike.empty(len(counts) + 1, np.int64)
+        current_offsets[0] = 0
+        backend.index_nplike.cumsum(counts, out=current_offsets[1:])
```
agoose77, line 122: Use `nonlocal` for readability (nonlocal appears later)

src/awkward/operations/ak_unflatten.py
```diff
@@ -202,13 +205,10 @@ def transform(layout, depth, axis):
 
                 return ak.contents.ListOffsetArray(ak.index.Index64(positions), content)
 
-            else:
-                return layout
-
-        out = transform(layout, depth=1, axis=axis)
+        out = ak._do.recursively_apply(apply, layout)
```
agoose77: Fix recursion. In v1, we had a `transform_child_layouts` that basically does what `continuation` does now (apart from the fact that it accepted a new transform function). Here, we restore the recursive behaviour

jpivarski (approved):
This is interesting: using `to_rectilinear` instead of `project`. Since the `counts` needs to be rectilinear (and not necessarily NumPy), that makes sense.

Nice work!

Issue #2070, opened 2023-01-05
`ak.to_layout(str)` does not throw exception


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

If `string` is a primitive type, then `to_layout` should not succeed; we don't support `to_layout(10)`, but `to_layout("10")` returns a length-one array of strings. This is seen in `ak.type("string")`, which returns `1 * string` instead of erroring, whilst `ak.type(10)` returns `int64`. 

It's not immediately obvious what we should do here. I am leaning towards suggesting that `ak.type()` should not support the leaf types (int64, string, etc), because it should operate on arrays.

jpivarski:
The main purpose of `ak.to_layout` is to prepare array-like arguments for other functions, so it should be designed for that. For that purpose, `ak.to_layout("10")` returning a length-one array of strings sounds right, since we'd want

```python
ak.something_or_other("one")
```

to work like

```python
ak.something_or_other(ak.Array(["one"]))
```

By that argument, `ak.to_layout(10)` should return a length-one array of integers, though.

What `ak.type` does can be brought into line with `ak.to_layout`, though it may just be better to deprecate `ak.type`. It's now redundant with `ak.Array.type`, `ak.Record.type`, and `Content.form.type`.

Well, no, we still need `ak.type` in case someone overloads that method in an `ak.Array`/`ak.Record` subclass. The word "type" is pretty general. But anyway, after getting `ak.to_layout` to do the right thing for function arguments, `ak.type` can be brought into agreement with it; it can return the type of array that `ak.to_layout` would have created.

Keep in mind, though, that `ak.to_layout` creates layouts, whose types differ from the corresponding `ak.Array` by not including the `ak.types.ArrayType` wrapper (which has the array length):

```python
>>> array = ak.Array([[1, 2, 3], [], [4, 5]])

>>> ak.type(array)
ArrayType(ListType(NumpyType('int64')), 3)
>>> print(ak.type(array))
3 * var * int64

>>> ak.type(array.layout)
ListType(NumpyType('int64'))
>>> print(ak.type(array.layout))
var * int64

```

agoose77:
@jpivarski I'd be strongly in favour of simplifying `ak.type` to return
```python
return ak.types.ArrayType(
            self._layout.form.type_from_behavior(self._behavior), self._layout.length
        )
```
with the suitable case for non-layout arguments?

[...]

PR #2098, opened 2023-01-10
fix: remove string casting from `ak.to_layout`

agoose77:
As per #2070, this PR changes `ak.to_layout("string")` to return the string (if `allow_other=True`). 

To support the primary use case of the old behaviour, a new `__cast__` behavior has been added for [byte]strings so that ufunc operations can accept Python strings.

### Changes
- Add new `is_non_string_like_sequence`, rename `is_non_string_like_iterable` helper
- Treat bytes/str as leaf types in `to_layout`
- Add `__cast__` support to string-likes
- Replaces usages of `is_non_string_like_iterable` with `is_non_string_like_sequence` where appropriate

### DIFF
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -44,7 +44,7 @@ def _to_rectilinear(arg):
         return tuple(_to_rectilinear(x) for x in arg)
     elif isinstance(arg, list):
         return [_to_rectilinear(x) for x in arg]
-    elif ak._util.is_non_string_iterable(arg):
+    elif ak._util.is_non_string_like_iterable(arg):
         raise ak._errors.wrap_error(
             TypeError(
                 f"encountered an unsupported iterable value {arg!r} whilst converting arguments to NumPy-friendly "
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -7,7 +7,7 @@ import numbers
 import os
 import re
 import sys
-from collections.abc import Iterable, Mapping, Sized
+from collections.abc import Iterable, Mapping, Sequence, Sized
 
 import packaging.version
 from awkward_cpp.lib import _ext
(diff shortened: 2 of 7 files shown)


agoose77:
Just adding support for bytestrings into `is_non_string_iterable`.

jpivarski (approved):
Right, so if `to_layout` is passed a string (or bytestring) and `allow_other=True`, ak_to_layout.py line 122 will return the string (or bytestring) as-is.

It's telling that no tests depended on this. That helpsto justify calling it a bug-fix.

It's good to merge!

agoose77:
> It's telling that no tests depended on this. That helpsto justify calling it a bug-fix.

Technically a couple of tests did implicitly ­— requiring that ufuncs support bare strings. I think the solution is actually closer to what we want to happen now, using `__cast__` (and our first usage of `__cast__` internally that I can tell).

agoose77:
@jpivarski before this makes it into a release ..., after our discussion about strings decaying to character arrays, do we want to revert this and replace it with a character-array codepath?

jpivarski:
I've lost track of what this refers to. What are you saying would be the consequences of cutting a release with this in it?

agoose77:
This PR does two things:
1. changes `ak.to_layout("string")` from returning `1 * string` to `str` iff. `allow_other=True`. 
2. converts `str` to `1 * string` via a `__cast__` behavior for ufunc operations.

We then spoke about single strings simply decaying to 	`char `arrays, so really this PR is not right either — `ak.to_layout("string") should return `N * char`. 

I think we still need the `__cast__`: we don't want ufunc operations treating Python strings as char arrays, I don't think.

Making this release would mean that users could start relying on `ak.to_layout("string", allow_other) == "string"`, which would break in the future.[^break]

[^break]: Well, `CharBehavior` means that `array_of_char == "string"` should work after we make the suggested "decay" change, but it's not _actually_  a `str`.

jpivarski:
`allow_other=True` is for passing through non-array arguments, some of which may be (Python) string-valued options. (In particular, do any ufuncs have arguments of string type, e.g. `direction="up"`? It would be hard to search for such things. We also accept ufuncs from libraries other than NumPy.)

Perhaps it should not be used so loosely; functions that call it with `allow_other=True` should pick which arguments are supposed to be arrays, make them arrays, and which arguments are not supposed to be arrays, and leave them as they are.

The reimplementation to make this return an array of characters ("decay") would be pretty easy:

```python
>>> ak.to_layout(["hello"])[0]
<NumpyArray dtype='uint8' len='5'>
    [104 101 108 108 111]
</NumpyArray>
```

Since the old bug (`to_layout` returned an array of one string) didn't trip over any tests, we're evidently not testing this case (other than tests/test_2070_to_layout_string.py) and the change should go through quickly.

So yes, it would be good to update `to_layout` to return an array of characters before the release.

PR #2099, opened 2023-01-10
fix: support categorical counts in `ak.unflatten`

### LINKED ISSUE #2071: unflatten raises counts must be an integer or a one-dimensional array of integers on a 1D awkward-array (worked in 1.x)
### Version of Awkward Array

2.0.5

I'm trying to transition from awkward 1.x -> 2.x, and ran into an issue with `ak.unflatten` where the call in 1.x works, but fails in 2.x. In my code, I call:

```python
    num_constituents = ak.num(jets.constituents, axis=2)
    # Convert to offsets
    offsets_constituents = np.cumsum(np.asarray(ak.flatten(num_constituents, axis=1)))
    offsets_constituents = np.insert(offsets_constituents, 0, 0)
    # Then into starts and stops
    starts_constituents = ak.unflatten(offsets_constituents[:-1], ak.num(num_constituents, axis=1))
    stops_constituents = ak.unflatten(offsets_constituents[1:], ak.num(num_constituents, axis=1))
```

For a bit of context, I'm trying to convert some info out from c++ back to python, and have found it convenient to return flattened np arrays and then unflatten on return. In any case, `ak.num(num_constituents, axis=1)` is 1D, as far as I can tell / recall, but in awkward 2.x, I get the following ValueError:

```python
E               ValueError: while calling
E
E                   ak.unflatten(
E                       array = numpy.ndarray([  0   6  37 ... 397 422 442])
E                       counts = <Array [1, 1, 1, 1, 1, ..., 1, 1, 1, 2, 1] type='17 * int64'>
E                       axis = 0
E                       highlevel = True
E                       behavior = None
E                   )
E
E               Error details: counts must be an integer or a one-dimensional array of integers
```

With details on the variables here:

```
array      = array([  0,   6,  37,  50,  81, 118, 146, 173, 193, 208, 242, 275, 318,
       336, 364, 397, 422, 442])
axis       = 0
backend    = <awkward._backends.NumpyBackend object at 0x1109822f0>
behavior   = None
counts     = <IndexedArray len='17'>
    <index><Index dtype='int64' len='17'>
        [ 0  1  3  4  5  6  7  8  9 10 11 12 13 15 16 18 19]
    </Index></index>
    <content><NumpyArray dtype='int64' len='20'>
        [1 1 0 1 1 1 1 1 1 1 1 1 1 1 0 1 1 0 2 1]
    </NumpyArray></content>
</IndexedArray>
highlevel  = True
layout     = <NumpyArray dtype='int64' len='18'>
    [  0   6  37  50  81 118 146 173 193 208 242 275 318 336 364 397 422 442]
</NumpyArray>
```

If I wrap the `ak.num(num_constituents, axis=1)` call with `np.asarray`, my integration tests pass, which at least suggests that it works (although the tests definitely aren't comprehensive), so it may be somewhat trivial. If it just needs to be cast, I think it would be a nicer user experience if `unflatten` did this on it's own. Thanks!

agoose77:
Fully closes #2071 by generalising support for indexed types to any `indexed` layout type (not just non-categoricals).

### DIFF
--- a/src/awkward/operations/ak_unflatten.py
+++ b/src/awkward/operations/ak_unflatten.py
@@ -93,9 +93,10 @@ def _impl(array, counts, axis, highlevel, behavior):
     if isinstance(counts, (numbers.Integral, np.integer)):
         current_offsets = None
     else:
-        counts = ak.operations.to_layout(
-            counts, allow_record=False, allow_other=False
-        ).to_packed()
+        counts = ak.operations.to_layout(counts, allow_record=False, allow_other=False)
+        if counts.is_indexed and not counts.is_option:
+            counts = counts.project()
+
         if counts.is_option and (counts.content.is_numpy or counts.content.is_unknown):
             mask = counts.mask_as_bool(valid_when=False)
             counts = backend.nplike.to_rectilinear(
--- a/tests/test_2071_unflatten_non_packed_counts.py
+++ b/tests/test_2071_unflatten_non_packed_counts.py
@@ -33,3 +33,10 @@ def test_option_counts():
         ak.unflatten([1.1, 2.2, 3.3, 4.4, 5.5], [None, 3, None, 0, 2]),
         [None, [1.1, 2.2, 3.3], None, [], [4.4, 5.5]],
     )
+
+
+def test_categorical_counts():
+    assert ak._util.arrays_approx_equal(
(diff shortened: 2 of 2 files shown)


jpivarski (approved):
That makes sense! `to_packed` wouldn't project out categoricals—that would be undesirable in most applications of packing. But here we want to do it.

This fixes it; merge whenever you're ready.

PR #2105, opened 2023-01-12
fix: merging 1D `NumpyArray` with option

### LINKED ISSUE #2104: `NumpyArray` merging is broken with option types
### Version of Awkward Array

main

This was probably caused by #2063

```python
>>> import awkward as ak
>> ak.concatenate([
...    ak.Array([1,2,3]),
...    ak.Array([1,2,None])
... ])
[1,
 2,
 3,
 1,
 2,
 None]
----------------
type: 6 * union[
    int64,
    ?int64
]

agoose77:
This used to work; I broke this in #2063 by removing this code block. I'm not sure how I managed to do so, I must have been tired and confused the ND `NumpyArray` duplicated code with the 1D case (that we still need).

Fixes #2104

### DIFF
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -342,9 +342,20 @@ class NumpyArray(Content):
             )
 
     def _mergeable_next(self, other, mergebool):
-
         if len(self.shape) > 1:
             return self._to_regular_primitive()._mergeable(other, mergebool)
+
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
And I didn't catch it!

Yes, each `_mergeable_next` (except possibly indexed/optional types?) should start by asking if the `other` is indexed/optional.

PR #2111, opened 2023-01-12
fix: support `is_indexed` types in `ak.fill_none`

### LINKED ISSUE #2107: `fill_none` doesn't move through indexed types
### Version of Awkward Array

main

The failure mode here is where we have a lazy carryd `RecordArray`, which then contains an option. Our existing logic returns early if it doesn't see an option, union, or record.

agoose77:
Fixes #2107


jpivarski (approved):
Old logic: "if union or record, continue recursing (by returning `None`); if option-type, do `fill_none`; otherwise, stop here (by returning `layout`)."

New logic: "if option-type, do `fill_none`; if union, record, or indexed (_not_ `IndexedOptionType` because this is after the check for option-type), continue recursing; otherwise, stop here."

That makes sense. It could be an `IndexedArray` of `RecordArray` (as in your test). In fact, that's pretty common, since any `carry` of a `RecordArray` results in an `IndexedArray` of `RecordArray`.

PR #2112, opened 2023-01-12
fix: be more permissive with sort translation

### LINKED ISSUE #2110: NEP-18 sort kind translation is incorrect
### Version of Awkward Array

main

We only support "stable", so we should just coerce the algorithms into a `stable` flag.

agoose77:
Fixes #2110 and removes the `awkward_quick_sort` kernel.

NumPy is moving away from naming sort algorithms. We're interpreting a specific algorithm here as "stable" or "unstable", and using that flag instead.


src/awkward/operations/ak_argsort.py
```diff
@@ -70,9 +70,9 @@ def _impl(array, axis, ascending, stable, highlevel, behavior):
 def _nep_18_impl(a, axis=-1, kind=None, order=unsupported):
     if kind is None:
         stable = False
-    elif kind == "stable":
+    elif kind == "stable" or kind == "mergesort":
```
jpivarski:
```suggestion
    elif kind in ("stable", "mergesort"):
```

src/awkward/operations/ak_argsort.py
```diff
@@ -70,9 +70,9 @@ def _impl(array, axis, ascending, stable, highlevel, behavior):
...
         stable = True
-    elif kind == "heapsort":
+    elif kind == "heapsort" or kind == "quicksort":
```
jpivarski:
```suggestion
    elif kind in ("heapsort", "quicksort"):
```

src/awkward/operations/ak_sort.py
```diff
@@ -57,9 +57,9 @@ def _impl(array, axis, ascending, stable, highlevel, behavior):
 def _nep_18_impl(a, axis=-1, kind=None, order=unsupported):
     if kind is None:
         stable = False
-    elif kind == "stable":
+    elif kind == "stable" or kind == "mergesort":
```
jpivarski:
```suggestion
    elif kind in ("stable", "mergesort"):
```

src/awkward/operations/ak_sort.py
```diff
@@ -57,9 +57,9 @@ def _impl(array, axis, ascending, stable, highlevel, behavior):
...
         stable = True
-    elif kind == "heapsort":
+    elif kind == "heapsort" or kind == "quicksort":
```
jpivarski:
```suggestion
    elif kind in ("heapsort", "quicksort"):
```

jpivarski (approved):
This is good. PyLint wants you to use `kind in (A, B)` instead of `kind == A or kind == B`. I'll do that and turn on auto-merge.

agoose77:
> This is good. PyLint wants you to use `kind in (A, B)` instead of `kind == A or kind == B`. I'll do that and turn on auto-merge.

Actually, I prefer that ­— my code-style is moving around in-line with Awkward, and sometimes gets a bit confused ;)

PR #2114, opened 2023-01-12
docs: remove reference to sorting implementation

agoose77:
We don't want users to worry about the specific sorting algorithm being used, so here we hide the implementation details. I'm also fairly sure these are old docs - we use `std::stable_sort` and `std::sort` on CPU as of writing, which is `mergesort` and `introsort` respectively


jpivarski (approved):
Good to say exactly what we mean!

I'll set auto-merge.

PR #2116, opened 2023-01-12
fix: add `ScalarType` and treat bare strings as char arrays

### LINKED ISSUE #2070: `ak.to_layout(str)` does not throw exception

### LINKED ISSUE #2096: Simplify & unify `ak.type()`
As discussed in #2070, `ak.type` would benefit from accepting a `highlevel` and` behavior` argument to control whether a composable `ak.types.Type` is returned in favour of the outer `ak.types.ArrayType`.

I also think we should simplify `ak.type` to deal exclusively with array-like objects. #1840 added support for some more scalars, but now that I think about it, we don't have any way of expressing these at the Awkward level; you can't have a bare scalar in an Awkward type, only a length-1 array. I think `ak.type` should just deal with arrays / array-like objects, and conventional `isinstance()` etc can be used for other types. 

We can make changes to `ak.type` non-breaking using `ak._util.deprecate` and requiring users to explicitly pass `highlevel=True/False`. Or, we can just declare it in the wiki, and write it in the changelog so that users don't need to write `ak.type(..., highlevel=True)`.

agoose77:
Fixes #2096, closes #2070, and adjusts #2098 to treat strings as character arrays.

The new `ScalarType` object let's us determine whether we have `ak.Record` or an `ak.Array` of records. It is the natural counterpart to the "high-level" `ArrayType`. Importantly, Awkward type objects _lose information_ if they're considered at the low-level, i.e. not wrapped by `ScalarType` or `ArrayType`; it is not possible to determine whether one has an Array or a Record. 

Note that this PR does not remove the `__cast__` implementation that converts string `ufunc` `*args` arguments to character arrays. This seems desirable in the main, unless there are well-known ufuncs that take string arguments?

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1707,7 +1707,9 @@ class Record(NDArrayOperatorsMixin):
         Note that the outermost element of a Record's type is always a
         #ak.types.RecordType.
         """
-        return self._layout.array.form.type_from_behavior(self._behavior)
+        return ak.types.ScalarType(
+            self._layout.array.form.type_from_behavior(self._behavior)
+        )
 
     @property
     def typestr(self):
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -449,12 +449,6 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
     def typestr(self):
         """
         The high-level type of this Array, presented as a string.
-
-        Note that the outermost element of an Array's type is always an
-        #ak.types.ArrayType, which specifies the number of elements in the array.
-
-        The type of a #ak.contents.Content (from #ak.Array.layout) is not
-        wrapped by an #ak.types.ArrayType.
         """
         return str(self.type)
 
@@ -1704,8 +1698,11 @@ class Record(NDArrayOperatorsMixin):
         """
         The high-level type of this Record; same as #ak.type.
 
-        Note that the outermost element of a Record's type is always a
-        #ak.types.RecordType.
+        Note that the outermost element of a Record's type is always an
+        #ak.types.ScalarType, which .
+
+        The type of a #ak.record.Record (from #ak.Array.layout) is not
+        wrapped by an #ak.types.ScalarType.
         """
         return ak.types.ScalarType(
             self._layout.array.form.type_from_behavior(self._behavior)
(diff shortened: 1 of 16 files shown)


agoose77:
Ah yes, test are now broken because we don't infer high-level / low-level from the array / layout type.

jpivarski (approved):
I've reviewed it, and it's good to merge once the tests have been updated.

agoose77:
@jpivarski in the test suite, a few places have become high-level tests (i.e., I added `ak.types.ArrayType` around the existing comparison operand(s). This means that sometimes we're now testing whether an operation produces `ArrayType(OtherType)` instead of `OtherType`, which doesn't explicitly test `OtherType`. However, I think this is fine.

I added a test to cover some of the `ak.type`  `Content` cases, to ensure that `Content.form.type` matches `ak.type(...).content`

agoose77:
@jpivarski btw, I'm maybe yielding on `highlevel=True`. It's just a thing users would need to read the docs to understand. If you think we should have the arg, I'll restore it after all!

jpivarski:
> This means that sometimes we're now testing whether an operation produces `ArrayType(OtherType)` instead of `OtherType`, which doesn't explicitly test `OtherType`.

It does test `OtherType`, since that's a nested constituent of `ArrayType`. Any equality test would recurse down the expected and observed `Type`-trees. So I think this is more than fine.

jpivarski:
> @jpivarski btw, I'm maybe yielding on `highlevel=True`. It's just a thing users would need to read the docs to understand. If you think we should have the arg, I'll restore it after all!

I never had a strong reason for it—just symmetry with the other `ak.*` functions. And it doesn't mean quite the same thing as it does in the other `ak.*` functions, so that's not a strong argument at all. Just, if a `highlevel` argument were to exist, this is what it should probably mean.

agoose77:
> It does test OtherType, since that's a nested constituent of ArrayType. Any equality test would recurse down the expected and observed Type-trees. So I think this is more than fine.

Sure, I mean simply that if `ak.type` is a black-box, we aren't technically testing the low-level, which should most carefully be done using `layout.form.type`. However, it's fine as you say.

PR #2117, opened 2023-01-13
test: fix on win32

agoose77:
For some reason, this test was failing on Windows 32-bit. I suspect it relates to the integer check of int64 against ULONGLONG, but I can't find any detailed information. It's easier just to try a different test.

### DIFF
--- a/tests/test_1318_array_function_types.py
+++ b/tests/test_1318_array_function_types.py
@@ -14,17 +14,26 @@ def test_tuple():
 
 
 def test_list():
-    data = np.array([1, 2, 3, 4, 3, 2, 1, 2], dtype=np.int64)
-    result = np.partition(ak.from_numpy(data), [4, 6])
+    A = np.eye(2) * 2
+    B = np.eye(3) * 3
+    result = np.block(
+        [
+            [ak.from_numpy(A), ak.from_numpy(np.zeros((2, 3)))],
+            [ak.from_numpy(np.ones((3, 2))), ak.from_numpy(B)],
+        ]
+    )
     assert isinstance(result, ak.Array)
-    assert ak._util.arrays_approx_equal(result, np.partition(data, [4, 6]))
+    assert ak._util.arrays_approx_equal(
+        result, np.block([[A, np.zeros((2, 3))], [np.ones((3, 2)), B]])
+    )
 
 
 def test_array():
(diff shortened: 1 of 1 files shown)


tests/test_1318_array_function_types.py
```diff
@@ -14,17 +14,26 @@ def test_tuple():
...
+        [
+            [ak.from_numpy(A), ak.from_numpy(np.zeros((2, 3)))],
+            [ak.from_numpy(np.ones((3, 2))), ak.from_numpy(B)],
+        ]
+    )
     assert isinstance(result, ak.Array)
-    assert ak._util.arrays_approx_equal(result, np.partition(data, [4, 6]))
+    assert ak._util.arrays_approx_equal(
+        result, np.block([[A, np.zeros((2, 3))], [np.ones((3, 2)), B]])
+    )
```
agoose77, line 28: I just wanted a test that involves a list argument. `np.block` errors if one passes a tuple instead of a list, and it supports _nested_ lists!

tests/test_1318_array_function_types.py
```diff
@@ -14,17 +14,26 @@ def test_tuple():
...
 
 def test_array():
-    data = np.array([1, 2, 3, 4, 3, 2, 1, 2], dtype=np.int64)
-    result = np.partition(ak.from_numpy(data), ak.Array([4, 6]))
+    haystack = np.array([1, 2, 3, 4, 4, 5, 6, 7], dtype=np.int64)
+    needle = np.array([5, 0, 2], dtype=np.int64)
+    result = np.searchsorted(ak.from_numpy(haystack), ak.from_numpy(needle))
     assert isinstance(result, ak.Array)
-    assert ak._util.arrays_approx_equal(result, np.partition(data, np.array([4, 6])))
+    assert ak._util.arrays_approx_equal(result, np.searchsorted(haystack, needle))
```
agoose77, line 36: `np.searchsorted` dodges the need to call `np.partition` with non-scalar arguments

agoose77:
Merging as I'm confident we want to start the release, and this is a test failure through NumPy usage.

jpivarski:
Thanks. I'm sorry that I was out of touch for a bit. (Daughter's violin practice.)

agoose77:
No problem! It's "out of work hours" right now, I just made a judgement call.

PR #2120, opened 2023-01-13
feat: add exception for missing field

### LINKED ISSUE #2086: Add a "field does not exist" `IndexError` subclass

This will enable framework authors to identify problems with typetracer usage leading to loss of important fields.

agoose77:
Fixes #2086 by introducing a new error type.

### DIFF
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -352,3 +352,7 @@ Issue: {}.""".format(
         version, date, will_be, message
     )
     warnings.warn(warning, category, stacklevel=stacklevel + 1)
+
+
+class FieldNotFoundError(IndexError):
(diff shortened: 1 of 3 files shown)


jpivarski:
I think @lgray wanted to know about this, since it's a way to catch Spark jobs that either had column names incorrectly optimized away or are missing for some other reason. The new exception type allows you to distinguish between spelling a column name wrong and other slicing errors (which are all `IndexError`, including this new `FieldNotFoundError`, through a subclass).

It will be going into the 2.0.6 release today.

jpivarski (approved):
Looks good! I'll merge it if the tests pass, and fix it if they don't. It will go into 2.0.6 today.

PR #2125, opened 2023-01-16
fix: re-order cases to handle NumPy scalar types properly

agoose77:
#2096 improved our type-handling, but had a sneaky bug in the ordering of the NumPy scalar type (`np.int16()`) handling: the generic `Integral` test would always catch these cases first, meaning that `ak.type(np.int16)` always produced `ScalarType(NumpyType("int64"))` etc.

This PR simplifies this logic by re-ordering the tests.

### DIFF
--- a/tests/test_2125_type_of_scalar.py
+++ b/tests/test_2125_type_of_scalar.py
@@ -12,3 +12,19 @@ def test():
     assert ak.type(np.dtype("complex128")) == ak.types.ScalarType(
         ak.types.NumpyType("complex128")
     )
+    assert ak.type("hello") == ak.types.ArrayType(
+        ak.types.NumpyType("uint8", parameters={"__array__": "char"}), 5
+    )
+    assert ak.type("int16") == ak.types.ArrayType(
+        ak.types.NumpyType("uint8", parameters={"__array__": "char"}), 5
+    )
+    assert ak.type(["int16"]) == ak.types.ArrayType(
+        ak.types.ListType(
+            ak.types.NumpyType(
+                "uint8", parameters={"__array__": "char"}, typestr="char"
+            ),
+            parameters={"__array__": "string"},
+            typestr="string",
+        ),
+        1,
+    )
--- a/tests/test_2125_type_of_scalar.py
+++ b/tests/test_2125_type_of_scalar.py
@@ -8,7 +8,9 @@ import awkward as ak
 
 def test():
     assert ak.type(np.int16()) == ak.types.ScalarType(ak.types.NumpyType("int16"))
-    assert ak.type(np.uint32) == ak.types.ScalarType(ak.types.NumpyType("uint32"))
+    with pytest.raises(TypeError):
+        assert ak.type(np.uint32) == ak.types.ScalarType(ak.types.UnknownType())
+    assert ak.type(None) == ak.types.ScalarType(ak.types.UnknownType())
     assert ak.type(np.dtype("complex128")) == ak.types.ScalarType(
         ak.types.NumpyType("complex128")
     )
--- a/tests/test_2125_type_of_scalar.py
+++ b/tests/test_2125_type_of_scalar.py
@@ -1,0 +1,14 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import numpy as np
+import pytest  # noqa: F401
+
+import awkward as ak
+
+
+def test():
+    assert ak.type(np.int16()) == ak.types.ScalarType(ak.types.NumpyType("int16"))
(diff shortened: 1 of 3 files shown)


tests/test_2125_type_of_scalar.py
```diff
@@ -0,0 +1,30 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import numpy as np
+import pytest  # noqa: F401
+
+import awkward as ak
+
+
+def test():
+    assert ak.type(np.int16()) == ak.types.ScalarType(ak.types.NumpyType("int16"))
```
jpivarski, line 10: Interesting: I didn't know that type constructors like `np.int16` could take zero arguments. Since all of these types are numbers of some sort, there's a natural return value—zero—but still, I'm surprised.

tests/test_2125_type_of_scalar.py
```diff
@@ -0,0 +1,30 @@
...
+    assert ak.type(np.uint32) == ak.types.ScalarType(ak.types.NumpyType("uint32"))
```
jpivarski: I don't know if you want to get into this. Technically, `ak.type(np.uint32)` is asking for the type of the type, which is a level of reflection that Awkward's type system doesn't have. (Python's does! `type(np.uint32)` is `<class 'type'>`)

Sometimes, Python ignores these "little" distinctions, like allowing `raise NotImplementedError` where you really mean `raise NotImplementedError()`. That's a slippery slope that Python has managed to not slide too far down.

But here, I don't see any value in allowing it. If the argument passed to `ak.type` is a type (either Python's `type` or our `ak.types.Type`), I think it would be fine for `ak.type` to raise an error.

tests/test_2125_type_of_scalar.py
```diff
@@ -0,0 +1,30 @@
...
+    assert ak.type(np.dtype("complex128")) == ak.types.ScalarType(
+        ak.types.NumpyType("complex128")
+    )
```
jpivarski, line 16: Hmmmm. The same argument I made for `ak.type(np.uint32)` could be made for `ak.type(np.dtype(...))`, although the latter is necessary to express units on datetimedeltas.

~~Actually, I think the same argument holds. Even though `np.dtype(...)` is a Python instance, it represents a NumPy type, just as `ak.types.Type` instances represent Awkward types. I don't see how it benefits anyone for the `ak.type` function to accept NumPy or Awkward _types_.~~

Actually actually, I changed my mind. Even though it is a category error (like `raise NotImplementedError` without the `()`), it's a useful one (again, like `raise NotImplementedError`). This usage of the `ak.type` function is to find out how we would name a type in the Awkward system: "`np.dtype("complex128")`" is NumPy's name for it, and "`ak.types.NumpyType("complex128")`" is ours. Similarly, one can use `ak.type` to find out how Python `int` and `str` are mapped to Awkward types through `ak.from_iter`, right?

```python
>>> ak.type(int)
...
TypeError: while calling

    ak.type(
        array = type-instance
        behavior = None
    )

Error details: unrecognized array type: <class 'int'>
>>> ak.type(str)
...
TypeError: while calling

    ak.type(
        array = type-instance
        behavior = None
    )

Error details: unrecognized array type: <class 'str'>
```

Well, okay, it can't. But also in trying it out, I found

```python
>>> ak.type("int32")
ArrayType(NumpyType('uint8', parameters={'__array__': 'char'}, typestr='char'), 5)
```

If users are expecting to be able to pass any dtype-like as an argument to `ak.type`, they'll be surprised by this one.

So, actually actually actually, I think `ak.type` should not accept a type as its argument, even if that's useful as a way of regularizing types "in the wild" into Awkward-style types. It's conflating too many different things, and we get into territory where two mutually incompatible expectations are relevant: arguments that accept a dtype often accept a dtype-like, which includes strings, and a string is an array-like with the above type. We wouldn't be able to make this work as expected for two users, each of whom has one or the other of these expectations.

So I think `ak.type` should only accept array-like arguments, not type-like arguments. That's where I've finally landed after flopping back and forth.

jpivarski (changes requested):
See below.

tests/test_2125_type_of_scalar.py
agoose77: > I don't know if you want to get into this. 

Personally, I really don't! ;) I was enforcing a test for a PR that we merged, but I'm happy to say this is out of bounds. I'm trying to move us away from scalar types altogether, but that's really only possible at the internal boundary, not publicly. 

>  I think it would be fine for ak.type to raise an error.

Excellent news, Jim!

tests/test_2125_type_of_scalar.py
agoose77: > So I think ak.type should only accept array-like arguments, not type-like arguments. That's where I've finally landed after flopping back and forth.

Yes, fantastic!

PR #2126, opened 2023-01-16
docs: disable IPyParallel & load dependencies in reverse order

agoose77:
I've noticed intermittent failures in the "Try It" demo; sometimes we get a symbol definition error. However, it transpires that re-running the import fixes this. I added an explicit import for `awkward-cpp` that appears to fix the issue.

This PR also disables IPyParallel, which is a nuisance.

### DIFF
--- a/docs/getting-started/demo/what-is-an-awkward-array.ipynb
+++ b/docs/getting-started/demo/what-is-an-awkward-array.ipynb
@@ -2,17 +2,17 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "id": "53a4d322",
+   "metadata": {
+    "deletable": false,
+    "editable": false
+   },
    "source": [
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
It's better for the "Try it!" experience to be stable and robust. It's probably fine for it to be a little slower; maybe most users (right now) understand that running Python in a browser is a different, less performant platform than regular CPython.

Also, I'm surprised that a parallel-processing extension would have any kind of speedup in a browser, which is async but not parallel. It's still the case that the installations use the `await` keyword.

agoose77:
I think IPyParallel is just a default extension that gets installed with IPython(?) I've not dug into where it comes from in great depth. In any case, we weren't using it, but it was causing a noisy popup that the server extension wasn't loaded!

PR #2127, opened 2023-01-16
feat: add `empty_if_typetracer` & expose more of typetracer to L2

agoose77:
This PR:
- adds `ak.typetracer.empty_if_typetracer`
- adds `ak.typetracer.typetracer_with_report` that builds a typetracer array and associated report from a `Form`.


jpivarski (approved):
This moves a typetracer-dependent code path out of the NumPy ufunc handling and into a generic function. I think `empty_if_typetracer` is also something that @lgray needed.

I approve this and I think it can be merged at any point, but it would be good if @lgray can tell us whether `empty_if_typetracer` is useful for his use-case or if some tweak could make it better. We're ignoring, for now, the fact that it's in a private API; everything that you use will become public API once things settle.

agoose77:
@jpivarski would you like me to remove `empty_if_typetracer` from this PR? :)

lgray:
Hey,

I'll be able to answer this better on Wednesday. Presently away.

jpivarski:
No, I thought that `empty_if_typetracer` is a good addition. Good enough, actually, for @lgray's use-case as well. We can revisit this on Wednesday. (I don't see anything this PR might be blocking.)

agoose77:
I've moved the cleanups this PR to https://github.com/scikit-hep/awkward/pull/2150. Now this is almost entirely `empty_if_typetracer`.

agoose77:
@jpivarski I think it's time to make `typetracer_with_report` public as it's now used by Dask, and `empty_if_typetracer` should be useful for @lgray (?)

lgray:
Can you give me a short example of how you expect it to be used?

agoose77:
The code in coffea's lookup tools:
https://github.com/CoffeaTeam/coffea/blob/38f0fcbe96efaea75d2ddfb70bd3e0b49ffd8ddf/coffea/lookup_tools/lookup_base.py#L33-L40

can be simplified to

```python
if backend == "cpu":
    if thelookup_wref is not None:
        thelookup = thelookup_wref()
    else:
        thelookup = thelookup_dask.compute()
elif backend == "typetracer":
    thelookup = thelookup_wref()
else:
    raise AssertionError
    
zlargs = [ak.typetracer.empty_if_typetracer(x) for x in args]
result = thelookup._evaluate(
    *(list(__pre_args__) + [awkward.to_numpy(zlarg) for zlarg in zlargs]),
    **kwargs,
)
```

You can simplify this further if you clarify the `thelookup` logic.

lgray:
I'd also like to update that to use `inspect.bind` to pack arguments instead of the clunky way I do it now. Is there a better way?

lgray:
`thelookup` is a way to consistently and quickly get a handle to the underlying correction being applied. We made a choice that corrections always take raw numpy arrays, and we unwrap awkward for that. The switch between weakref and dask is to avoid serialization overhead in dask since otherwise dask will pickle the correction metadata, which can be very large and expensive.

i.e. it's just a pointer to a function, not an awkward array or anything like that.

jpivarski (approved):
Yes, these can be public now, but if the private method access breaks, we should be careful about aligning versions so that we don't have a broken Awkward + Coffea pair in the crucial period around CHEP.

agoose77:
@lgray does this Gist help towards rewriting your transformer?

https://gist.github.com/agoose77/9fc1938bf15d1dedb0bd72a4cd501ed1

lgray:
@agoose77 Yes! This looks very much like the correct way. I'll give it a try during one of the breaks today! TYVM!

agoose77:
>  but if the private method access breaks, we should be careful about aligning versions so that we don't have a broken Awkward + Coffea pair in the crucial period around CHEP.

Agreed. I'll just export the symbol from `awkward.typetracer`, and we'll ensure there's a transition period.

lgray:
Sweet - I'll try to address this in coffea before your next release since it creates incompatibility (I think?).

agoose77:
This now doesn't rename any symbols, we only export `typetracer_with_report`, `TypeTracerReport`, and `empty_if_typetracer` functions from the public `ak.typetracer` module. So, you should make a PR, but we won't break you :)

PR #2128, opened 2023-01-16
refactor: remove `to_arraylib`

agoose77:
This PR
- Removes `ak._util.to_arraylib` and
- Deprecates `NumpyArray.to_numpy(allow_missing)` in favour of
- Adds `NumpyArray.to_backend_array(allow_missing, backend)`
- Deprecates support of `UnionArray` in `to_backend_array()`.

`to_arraylib()` duplicates much of `to_numpy()`'s logic, and we also don't need to enshrine NumPy as more special than the general backend-agnos

### DIFF
--- a/src/awkward/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_connect/rdataframe/from_rdataframe.py
@@ -239,7 +239,9 @@ def from_rdataframe(data_frame, columns):
 
             if col == "awkward_index_":
                 contents_index = ak.index.Index64(
-                    array.layout.to_numpy(allow_missing=True)
+                    array.layout.to_backend_array(
+                        allow_missing=True, backend=ak._backends.NumpyBackend.instance()
+                    )
                 )
             else:
                 contents[col] = array.layout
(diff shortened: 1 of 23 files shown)


jpivarski (approved):
~~It's great that the implementations can be unified, but there needs to be a high-level public `ak.to_numpy`, if only in name. There will be a lot of users who don't even know (or care) that Awkward runs on non-NumPy backends, and this will make them unlikely to find a function named `ak.to_backend_array`.~~

Nevermind: I see that you're leaving the L1 interface `ak.to_numpy` and just removing the L2 `Content.to_numpy` interface.

Currently, only NumPy has a way to use `allow_missing=True`; CuPy, etc. don't have a `ma` submodule or a `MaskedArray` equivalent. But in principle, they might someday get such an interface, and we can rename it through nplike if it comes in with a different name.

It's good to have the deprecation message specify version 2.2.0 like this, to always target two minor versions ahead. When this is merged, please add it directly to

https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20

to make sure that this list doesn't get out of date. The warning message is for developers who stumble onto this deprecation by calling `Content.to_numpy`, but the wiki is for a more global view: "What are all of the things that will be deprecated in version X?"

PR #2130, opened 2023-01-16
chore: apply changes from flake8

agoose77:
Recent updates to flake8-bugbear introduced B028 which warns for `f'"{str}"'` vs `f"{str!r}"`, and B906 which warns for non-exhaustive AST visitors.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -12,7 +12,6 @@ per-file-ignores =
     src/awkward/_connect/numba/*: AK1
     src/awkward/_errors.py: AK1
     src/awkward/typing.py: F405
-    src/awkward/typing.py: F405
     src/awkward/_connect/rdataframe/to_rdataframe.py: B028
 
 [flake8:local-plugins]
--- a/.flake8
+++ b/.flake8
@@ -12,6 +12,8 @@ per-file-ignores =
     src/awkward/_connect/numba/*: AK1
     src/awkward/_errors.py: AK1
     src/awkward/typing.py: F405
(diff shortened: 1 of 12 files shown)


.flake8
```diff
@@ -12,6 +12,7 @@ per-file-ignores =
     src/awkward/_connect/numba/*: AK1
     src/awkward/_errors.py: AK1
     src/awkward/typing.py: F405
+    src/awkward/_connect/rdataframe/to_rdataframe.py: B028
```
agoose77, line 15: We need to exactly produce quoted strings for C++, so don't let Python decide upon the quote char.

jpivarski (approved):
The strings that are quoted here are type names, and therefore guaranteed to not contain any characters like "`"`" that would get backslash-quoted. But using `!r` to quote things is good practice (so a future maintainer wouldn't need to know what characters these strings have, from other constraints).

I don't think the outer quoting of an f-string affects the way `!r` works in substitutions, does it? But if it does, you're right that it needs to be controlled.

agoose77:
> I don't think the outer quoting of an f-string affects the way !r works in substitutions, does it? But if it does, you're right that it needs to be controlled.

The main thing I'm worried about in this PR is `"` vs `'`. I don't believe this is affected by the choice of the outer quotes; I believe REPR quoting is exclusively a property of the object. We _definitely_ want to control these for C++, so I explicitly disabled this check for the `to_rdataframe` module. This just means we can easily read what's being produced (they'll be single-quotes by default otherwise).

PR #2131, opened 2023-01-17
refactor: move `Singleton` to its own module

agoose77:
This tiny PR is part of the process of solidifying our typetracer / nplike mechanism. A long-term refactoring goal is to make imports more explicit, i.e.
```python
from awkward._util import some_func
```
instead of
```python
import awkward as ak

def use_some_func():
	return some_func()
```
so by reducing the number of independent concepts in `_util`, we can avoid the import problems that stem from this.

### DIFF
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -14,9 +14,9 @@ from awkward._nplikes import (
     NumpyKernel,
     NumpyLike,
     NumpyMetadata,
-    Singleton,
     nplike_of,
 )
+from awkward._singleton import Singleton
 from awkward._typetracer import NoKernel, TypeTracer
 from awkward.typing import Callable, Final, Tuple, TypeAlias, TypeVar, Unpack
 
--- a/src/awkward/_nplikes.py
+++ b/src/awkward/_nplikes.py
@@ -6,21 +6,8 @@ import ctypes
 import numpy
 
 import awkward as ak
+from awkward._singleton import Singleton
 from awkward.typing import TypeVar
-
-
-class Singleton:
-    _instance = None
-
-    def __new__(cls, *args, **kwargs):
-        assert cls._instance is None
-        return super().__new__(cls, *args, **kwargs)
-
-    @classmethod
-    def instance(cls):
-        if cls._instance is None:
-            cls._instance = cls()
-        return cls._instance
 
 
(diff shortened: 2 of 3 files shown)


jpivarski (approved):
"Singleton" is general enough that it might almost be a good language feature. I think it could have gone into `_util`, but it's also fine as an independent module.

agoose77:
> "Singleton" is general enough that it might almost be a good language feature. I think it could have gone into `_util`, but it's also fine as an independent module.

That's true. For posterity, we did have this in `_util`, but I've moved it. The rationale is `_utils` is always a big bag of functions and types. It couples lots of code together by virtue of it being "useful" :laughing: I am also sure that it's a rule of programming that _every_ project has one, because it reflects the need to group things falling into contention with a lack of simple categories.

My longer-term plan is to split `_util` into only the true "utilty" functions, e.g. `regularize_path`, `is_non_string_iterable` etc; functions like `arrayclass` or `numba_array_typer` probably could live somewhere more appropriate.

To be clear, this is all on-the-go refactoring, not bikeshedding ;)

jpivarski:
I find it acceptable when categorizing things to have an "other" or "miscellaneous" category. It's a bigger problem when every codebase stashes these things in a different place, under a different name; I'd like to standardize "util" as the common name for this thing.

The junk drawer of programming:

a6e3d5e7f802c3d69a61b98de8dcd0240fb0ce3d_xxxxx_organizejunkdrawer_gettingstarted_600x600

agoose77:
> I find it acceptable when categorizing things to have an "other" or "miscellaneous" category. It's a bigger problem when every codebase stashes these things in a different place, under a different name; I'd like to standardize "util" as the common name for this thing.
> 
> The junk drawer of programming:
> 
> a6e3d5e7f802c3d69a61b98de8dcd0240fb0ce3d_xxxxx_organizejunkdrawer_gettingstarted_600x600

For sure. I think `_util` is fine, and we'll keep it around. They always grow in size, and the trick I think is to pick-out that which does not belong. This is easier in retrospect, once a design is established. In our case, we have Numba code, and layout code alongside traditional utility functions. I think the numba functions could move to `ak._connect`, whilst the layout code e.g. `behavior_of`, `arrayclass` might all belong to a new submodule for "layout-utils".

Issue #2108, opened 2023-01-12
Move existing `Content` recursions into `recursively_apply` implementations


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

Many of our `Content` methods are recursions that only have meaningful implementations at the final node. We could eliminate a class of bugs by moving these to `recursively_apply`.

There are several variants of this:
1. naive recursion, category dependent (i.e. `is_list` instead of `isinstance(layout, ListOffsetArray)`)
2. naive recursion, layout dependent (i.e. `isinstance(layout, ListOffsetArray)`)
3. complex recursion, layout dependent

Only (1) and (2) can be ported to `recursively_apply`. For (2), we likely want a category-protocol e.g. `_fill_none` for all `is_option`. This likely should be implemented by:
1. Moving `ak._do` to `ak.contents._do`.
2. Adding a set of category mixins, and an associated `_do` implementation.
3. Defining "category protocols" on these mixins (abstract methods)
4. `recursively_apply` in `ak._do` recurses down to e.g. `is_option`, and then calls `ak.contents.categories.do.fill_none`?

This means that the L3 distinction is now "is there a `do`?

PR #2132, opened 2023-01-17
refactor: move kernel logic to _kernels

agoose77:
This tiny PR  is part of the on-going nplike work. It moves the kernel logic into its own module, as kernels are associated with backends (but are still distinct concepts)

### DIFF
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -5,13 +5,11 @@ from abc import ABC, abstractmethod
 import awkward_cpp
 
 import awkward as ak
+from awkward._kernels import CupyKernel, JaxKernel, NumpyKernel
 from awkward._nplikes import (
     Cupy,
-    CupyKernel,
     Jax,
-    JaxKernel,
     Numpy,
-    NumpyKernel,
     NumpyLike,
     NumpyMetadata,
     Singleton,
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -6,14 +6,7 @@ import awkward_cpp
 
 import awkward as ak
 from awkward._kernels import CupyKernel, JaxKernel, NumpyKernel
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
The top level module is getting a lot of hidden submodules. Maybe these (including `_singleton`) should be in a directory named `_util`? Just to pack them away out of view when we're looking for the main constituents of the codebase? Or into directories that organize the submodules by purpose?

agoose77:
> Maybe these (including _singleton) should be in a directory named _util? Just to pack them away out of view when we're looking for the main constituents of the codebase? Or into directories that organize the submodules by purpose?

Yes, I think so. I was originally going to come up with a name once more of these PRs are merged, but actually I think a `_internal` submodule might be appropriate, because we're replacing `_` with a prefix name. We could call this package `ak._`, but that's not good naming in my view.

In #2108 I mentioned our thought process about organising L3/4. We discussed making L3 APIs for categories (i.e. `is_option`, `is_list`). I'm thinking that `ak._do` might become `ak._internal.content.XXX` (`ak._do` is content-wide protocols), whilst e.g. the new `fill_none` L3 entry-point could live in `ak._internal.category.option.XXX` (`ak._internal.category.option` is only L3 protocol entrypoints on options, e.g. `fill_none`, `is_none`).

agoose77:
i.e.
```
src/
├─ my-app/
│  ├─ _internal/
│  │  ├─ contents.py
│  │  ├─ categories/
│  │  │  ├─ option.py
│  ├─ contents/
│  │  ├─ numpyarray.py
│  │  ├─ listoffsetarray.py
│  │  ├─ ...
│  ├─ types/
│  │  ├─ numpytype.py
│  │  ├─ listtype.py
│  │  ├─ ...
```

PR #2134, opened 2023-01-17
refactor: use `nplike.asarray`

agoose77:
The work to move to an array-API inspired UX requires that we drop `array` and `asarray` distinctions; now `asarray` takes a `copy` parameter, and drops the `order` parameter.

I'm gradually dropping `nplike` usage in our tests, where appropriate; we should probably focus on testing L1 and L2, not L3.

### DIFF
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -5,10 +5,10 @@ from abc import ABC, abstractmethod
 import awkward_cpp
 
 import awkward as ak
-from awkward._kernels import CupyKernel, JaxKernel, NumpyKernel
+from awkward._kernels import CupyKernel, JaxKernel, NumpyKernel, TypeTracerKernel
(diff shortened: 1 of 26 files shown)


src/awkward/_backends.py
```diff
@@ -5,10 +5,10 @@
 import awkward_cpp
 
 import awkward as ak
-from awkward._kernels import CupyKernel, JaxKernel, NumpyKernel
+from awkward._kernels import CupyKernel, JaxKernel, NumpyKernel, TypeTracerKernel
```
agoose77, line 8: I forgot to cherry pick this in the last PR (oops).

jpivarski (approved):
Looks good! (For posterity: we talked about this in our weekly Zoom meeting. The functionality of `asarray` and `array` is being combined in Array-API, so we're following the same pattern in our nplikes for future-proofing.)

PR #2136, opened 2023-01-19
chore: isolate nox dependencies in `noxfile`

agoose77:
I'm not 100% happy with our current monorepo design. I'd like to better isolate the "monorepo" parts from the "Python package" parts. Right now, that boundary is mainly the `noxfile.py` and `dev/` scripts; nearly everything else remains in the sdist for awkward/awkward-cpp.
Further to this, I'd like to unify our usage of nox and our CI configuration more closely; right now, they're both somewhat separate. However, both of these bugbears are not big enough to warrant doing anything at the moment. 

This PR tackles the smallest issue, which is to remove files that are only needed by specific tools. In this case, we don't need `requirements-dev.txt` anywhere except our noxfile, so that's where the contents now live. 

Additionally, I set `reuse_venv=True`. This only means that the venvs get re-used by default, though without the `-R` flag, the dependencies are re-installed each time, so this is only a small change.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -10,6 +10,16 @@ ALL_PYTHONS = ["3.7", "3.8", "3.9", "3.10", "3.11"]
 nox.options.sessions = ["lint", "tests"]
 
 
+requirements_dev = [
+    "build",
+    "numpy",
+    "packaging",
+    "PyYAML",
+    "requests",
+    "toml",
+]
+
+
 def remove_if_found(*paths):
     for path in paths:
         if os.path.isdir(path):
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This looks good!

And related to this is the point @ianna raised at our meeting today, that it's easy to miss a `nox -s prepare` call (or to call it with nox not installed and not see the error message). It would be sufficient to check for the existence of `awkward-cpp/include/awkward/kernels.h` during wheel-building (either `awkward-cpp` or `awkward`, since the recommended way to make both is with `pip`).

PR #2137, opened 2023-01-19
refactor: make `nplike.zeros` et al. use Array API type signatures

agoose77:
This is another PR to slowly move our nplikes over towards the array API, in order to improve the safety & maintainability.

### DIFF
--- a/src/awkward/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_connect/rdataframe/from_rdataframe.py
@@ -71,7 +71,7 @@ def from_rdataframe(data_frame, columns):
     def empty_buffers(cpp_buffers_self, names_nbytes):
         buffers = {}
         for item in names_nbytes:
-            buffers[item.first] = ak._nplikes.numpy.empty(item.second, dtype=np.uint8)
+            buffers[item.first] = ak._nplikes.numpy.empty(item.second)
             cpp_buffers_self.append(
                 item.first,
                 buffers[item.first].ctypes.data_as(ctypes.POINTER(ctypes.c_ubyte)),
--- a/src/awkward/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_connect/rdataframe/from_rdataframe.py
@@ -71,7 +71,7 @@ def from_rdataframe(data_frame, columns):
     def empty_buffers(cpp_buffers_self, names_nbytes):
         buffers = {}
         for item in names_nbytes:
-            buffers[item.first] = ak._nplikes.numpy.empty(item.second)
+            buffers[item.first] = ak._nplikes.numpy.empty(item.second, dtype=np.uint8)
             cpp_buffers_self.append(
                 item.first,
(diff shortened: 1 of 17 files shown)


src/awkward/_connect/rdataframe/from_rdataframe.py
```diff
@@ -71,7 +71,7 @@ def form_dtype(form):
     def empty_buffers(cpp_buffers_self, names_nbytes):
         buffers = {}
         for item in names_nbytes:
-            buffers[item.first] = ak._nplikes.numpy.empty(item.second)
```
agoose77, line 74: @ianna Is this "fix" correct? It's my understanding that we want to allocate `item.second` bytes here, so I'm setting a dtype of `uint8`.

src/awkward/contents/emptyarray.py
```diff
@@ -85,11 +85,13 @@ def _repr(self, indent, pre, post):
...
         return ak.contents.NumpyArray(
-            backend.nplike.empty(0, dtype), parameters=self._parameters, backend=backend
+            backend.nplike.empty(0, dtype=dtype),
+            parameters=self._parameters,
+            backend=backend,
         )
 
     def __array__(self, **kwargs):
-        return numpy.empty((0,))
+        return numpy.empty(0, dtype=np.float64)
```
agoose77, line 94: @jpivarski this is correct, yes?

src/awkward/_connect/rdataframe/from_rdataframe.py
```diff
@@ -71,7 +71,7 @@ def form_dtype(form):
...
+            buffers[item.first] = ak._nplikes.numpy.empty(item.second, dtype=np.uint8)
```
agoose77: @ianna Is this "fix" correct? It's my understanding that we want to allocate `item.second` bytes here, so I'm setting a dtype of `uint8`.

src/awkward/contents/emptyarray.py
jpivarski: This is correct. The reason I was left the dtype implicit (which is, in practice, `np.float64`) is because when Awkward's `UnknownType` has to become a concrete type, we want it to become whatever NumPy's default is when a dtype is not specified.

However, we _know_ that NumPy's default is `np.float64`, and I think that fact is used elsewhere in the code (`to_NumpyArray`?). NumPy can't change their default, so I think it's fine to hard-code it like this, this that's more informative to readers of this codebase.

jpivarski (approved):
These are all good changes and I approve, but please do get @ianna's confirmation on the array that you're assuming is `np.uint8`.

Unlike the change that I approved below, which adds an explicit `dtype=np.float64` when that's what the implicit default would have been, your assignment of `dtype=np.uint8` in `from_rdataframe.py` _changes the behavior of this code_. Let's be sure that's a correct change.

agoose77:
I've reverted the change that requires Ianna's approval; that's much smaller, so I can move it to a new PR that is less time-sensitive  (this one is only sensitive because I want to keep working upon main :))

PR #2138, opened 2023-01-19
chore: drop unused nplike functions

agoose77:
Our nplikes include several functions that are not used or tested. This means that modifications to this code are untested and add to the maintenance budget. This PR removes such functions, and in the process adds some explicit signatures to any touched functions.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -605,7 +605,7 @@ def apply_step(
                         if mask is None:
                             mask = m
                         else:
-                            mask = backend.index_nplike.bitwise_or(mask, m, out=mask)
+                            mask = backend.index_nplike.logical_or(mask, m, out=mask)
 
                 nextmask = Index8(mask.view(np.int8))
                 index = backend.index_nplike.full(mask.shape[0], -1, dtype=np.int64)
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -397,7 +397,7 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
         if not isinstance(masked_index, ak.contents.UnmaskedArray):
             mask = masked_index.mask_as_bool(valid_when=False)
             if mask.any():
-                index = numpy.array(index, copy=True)
+                index = numpy.asarray(index, copy=True)
                 index[mask] = -1
 
         content = handle_arrow(paarray.dictionary, generate_bitmasks)
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -196,38 +196,6 @@ if pyarrow is not None:
         pyarrow.duration("ns"): (False, np.dtype("m8[ns]")),
     }
 
-if not ak._util.numpy_at_least("1.17.0"):
-
-    def packbits(bytearray, lsb_order=True):
-        if lsb_order:
-            if len(bytearray) % 8 == 0:
-                ready_to_pack = bytearray
-            else:
-                ready_to_pack = numpy.empty(
-                    int(numpy.ceil(len(bytearray) / 8.0)) * 8,
-                    dtype=bytearray.dtype,
-                )
-                ready_to_pack[: len(bytearray)] = bytearray
-                ready_to_pack[len(bytearray) :] = 0
-            return numpy.packbits(ready_to_pack.reshape(-1, 8)[:, ::-1].reshape(-1))
-        else:
-            return numpy.packbits(bytearray)
-
-    def unpackbits(bitarray, lsb_order=True):
-        ready_to_bitswap = numpy.unpackbits(bitarray)
-        if lsb_order:
-            return ready_to_bitswap.reshape(-1, 8)[:, ::-1].reshape(-1)
-        else:
-            return ready_to_bitswap
-
-else:
-
-    def packbits(bytearray, lsb_order=True):
-        return numpy.packbits(bytearray, bitorder=("little" if lsb_order else "big"))
-
-    def unpackbits(bitarray, lsb_order=True):
-        return numpy.unpackbits(bitarray, bitorder=("little" if lsb_order else "big"))
-
 
 def and_validbytes(validbytes1, validbytes2):
     if validbytes1 is None:
@@ -242,7 +210,7 @@ def to_validbits(validbytes):
     if validbytes is None:
         return None
     else:
-        return pyarrow.py_buffer(packbits(validbytes))
+        return pyarrow.py_buffer(numpy.packbits(validbytes, bitorder="little"))
 
 
 def to_length(nparray, length):
@@ -659,7 +627,9 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
         validbits = buffers.pop(0)
         bitdata = buffers.pop(0)
 
-        bytedata = unpackbits(numpy.frombuffer(bitdata, dtype=np.uint8))
+        bytedata = numpy.unpackbits(
+            numpy.frombuffer(bitdata, dtype=np.uint8), bitorder="little"
+        )
 
         out = ak.contents.NumpyArray(
             bytedata.view(np.bool_),
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -3,11 +3,10 @@
 import json
 from collections.abc import Iterable, Sized
 
-import numpy
-
 import awkward as ak
 
 np = ak._nplikes.NumpyMetadata.instance()
+numpy = ak._nplikes.Numpy.instance()
 
 try:
     import pyarrow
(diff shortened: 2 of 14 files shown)


src/awkward/_connect/pyarrow.py
```diff
@@ -196,38 +196,6 @@ def num_fields(self):
...
-        else:
-            return ready_to_bitswap
-
-else:
-
-    def packbits(bytearray, lsb_order=True):
-        return numpy.packbits(bytearray, bitorder=("little" if lsb_order else "big"))
-
-    def unpackbits(bitarray, lsb_order=True):
-        return numpy.unpackbits(bitarray, bitorder=("little" if lsb_order else "big"))
```
agoose77, line 229: I couldn't find the cause of the `numpy_at_least` here using `git blame`. @jpivarski is there a reason for this switch that I need to consider? Or can we just use NumPy's built-in version

agoose77:
~@jpivarski I noticed during refactoring this that we use the NumPy `out=` parameter in a few places. JAX doesn't support this, yet JAX is also somewhat of a special case. At the same time, the Array API drops the `out` parameter entirely. 
I think we should introduce a different API than `out` e.g. `maybe_out`. That way, it's not _required_ that `out` is modified if the underlying nplike does not support it. Instead, e.g. JAX would just return a new array and ignore the `maybe_out` array  entirely.~

Actually, this should just be a PR that you can review.

src/awkward/_broadcasting.py
```diff
@@ -605,7 +605,7 @@ def continuation():
                         if mask is None:
                             mask = m
                         else:
-                            mask = backend.index_nplike.bitwise_or(mask, m, out=mask)
+                            mask = backend.index_nplike.logical_or(mask, m, out=mask)
```
jpivarski, line 608: I understand: `logical_or` is more precisely what we mean.

We can switch to `logical_or` for maintenance reasons. The reason it was `bitwise_or` is because it's faster (`logical_or` would combine `0x80` and `0x0f` as `0x01`, which is a few more instructions). I would only use `bitwise_or` where I mean `logical_or` if I know for a fact that all of the inputs are either `0x00` or `0x01`, as is the case here, since `m` comes from `mask_as_bool`.

(This change is fine: I'm just explaining myself.)

src/awkward/_connect/pyarrow.py
jpivarski: Yes, it's because the `bitorder` argument was only introduced in NumPy 1.17.0 (see [np.packbits](https://numpy.org/doc/stable/reference/generated/numpy.packbits.html)).

Without NumPy 1.17.0, we have to use a cute work-around for swapping the bit order with a slice. We're still supporting NumPy from 1.13.1 onward, so that's in principle needed.

~~This whole code block is red. Where is its replacement?~~ I see: you're using `numpy.packbits` directly with an explicit `bitorder="little"`. We can't do that unless we formally give up on NumPys 1.13 through 1.16 inclusive.

jpivarski: As an alternative, you could put the cute slice work-around in an nplike implementation of `packbits`/`unpackbits`. Then we can use the future-proof interface everywhere (a `bitorder` argument, rather than `lsb_order`, which is designed around the `BitMaskedArray` class) and remove it from one spot—the Numpy nplike—when we increase our minimum NumPy.

To summarize, these are the important version thresholds:

  * 1.13.1: NEP 13 introduced
  * 1.17.0: NEP 18 became a default
  * 1.17.0: `packbits`/`unpackbits` got the `bitorder` argument

Hmmm. Maybe 1.17.0 could be a good threshold. It was July 26, 2019. (1.13.1 was July 6, 2017, so we'd be moving up 2 years. Awkward 0 started in June 2018, requiring NumPy 1.13.1 for ufuncs to work, so moving up to 1.17.0 now would still be a bigger gap now than it was at the start.)

In fact, NumPy 1.13.1 isn't even listed on https://endoflife.date/numpy or https://numpy.org/neps/nep-0029-deprecation_policy.html. By setting the minimum to NumPy 1.17.0, we're still picking a minimum version that isn't supported by the NumPy team itself (i.e. we accept all versions that NumPy supports, and then some), but at least it's on the table with a date next to it. Also, NumPy 1.13.1 technically supports Python 3.6, which we don't anymore.

New plan: we set the minimum NumPy to 1.17.0. Does that sound good to you? I can also put out a question on Slack #awkward-uproot.

jpivarski: Ah, but doing something like that, we should do it on a minor version number, so that these things are easier to track by hand (if need be; I know, pip would do this one automatically because it's a strict dependency).

So we should put "minimum NumPy version >= 1.17.0" on the table for Awkward 2.1.0 and for now have `packbits`/`unpackbits` go through nplike with the backward-compatible workaround in one place, slated for removal in 2.1.0.

jpivarski (changes requested):
I had to decide what to do about `packbits`/`unpackbits` and NumPy versions older than 1.17.0, but my personal opinion has come around to thinking we should drop support for older NumPys.

Where I ended up with this was that we should schedule a new minimum dependency, `numpy>=1.17.0`, for Awkward 2.1.0. This doesn't have to be a deprecation cycle (there's no public Awkward function call to deprecate), but I'll put it on [the table](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20) where anyone can look it up, and I'll post it on #awkward-uproot, maybe also a GitHub Discussion.

But for now, this PR needs to hide the pre-1.17.0 work-around in `nplike.packbits`/`nplike.unpackbits`, since we can't change a dependency version on a bug-fix release.

2.1.0 can come out any time now, since it's been a month since 2.0.0, but I'd like to do it closer to the beginning of a month, February 1, so that it has that cadence. We can also be vocal about it, so that it doesn't scare anyone.

src/awkward/_connect/pyarrow.py
agoose77: > Yes, it's because the bitorder argument was only introduced in NumPy 1.17.0

Huh that didn't render on mobile when I was looking, but that was my assumption... wierd!

> Also, NumPy 1.13.1 technically supports Python 3.6, which we don't anymore.

Yep, I am enthusiastic in my support for dropping 1.13 in favour of 1.17.

> packbits/unpackbits go through nplike with the backward-compatible workaround in one place,

Yep, this is an easy fix for now.

src/awkward/_broadcasting.py
agoose77: This might have been a slightly premature change on my part; I had forgotten about the cycle cost for the two instructions. Shall I revert? It's not a huge burden to restore one function.

jpivarski: Nevermind: it's not enough to worry about. Here's a test showing that it's less than a factor of 2. Three arrays of 84 kB will fit in my L1d cache (256 kB):

```python
>>> one = np.random.randint(0, 2, 84*1024, np.int8)
>>> two = np.random.randint(0, 2, 84*1024, np.int8)
>>> out = np.empty(84*1024, np.int8)
>>> timeit.timeit("np.logical_or(one, two, out=out)", number=100000, globals=globals())
0.47637764099999913
>>> timeit.timeit("np.bitwise_or(one, two, out=out)", number=100000, globals=globals())
0.267168271002447
```

Either I used to care about factors of 2 or the gap used to be bigger than that.

agoose77: I had a feeling it was about 2-4x, so this is also mildly lower than I'd thought.

agoose77:
@jpivarski this PR now moves the compatibility shim to Numpy's nplike, and has the new `python==3.7, numpy==1.14.5` CI test passing, are you happy for me to merge?

jpivarski (approved):

jpivarski:
Yes, thank you!

jpivarski:
The uses of `numpy.packbits` in pyarrow.py are not using the nplike function with the pre-1.17 work-around:

https://github.com/scikit-hep/awkward/blob/2645632e1084b73b501a33c6aa42c6cbc1a31f4c/src/awkward/_connect/pyarrow.py#L6

but I don't think you can import a sufficiently recent version of pyarrow without a recent NumPy, so the point is probably moot.

agoose77:
> The uses of `numpy.packbits` in pyarrow.py are not using the nplike function with the pre-1.17 work-around:
> 
> https://github.com/scikit-hep/awkward/blob/2645632e1084b73b501a33c6aa42c6cbc1a31f4c/src/awkward/_connect/pyarrow.py#L6
> 
> but I don't think you can import a sufficiently recent version of pyarrow without a recent NumPy, so the point is probably moot.

Oops, this fell out of sync with the rest of the PR. Hmm, I feel that we might want to start specifying lower bounds for JAX, CuPy, and PyArrow, because right now we don't have any information about which versions we need to support internally. I think PyArrow supports `>= 1.16.x` even in 10.0.0, which means it's fairly possible that someone might have NumPy < 1.1.17

agoose77:
I've returned to using our `NumpyLike` here, which has the patch. I was OK with our `_connect` code directly using `NumPy` or `CuPy`, because they're different value statements; we use `nplikes` so that we have a common interface, whilst we use the modules directly so that we don't require the JAX nplike to implement CUDA specific methods. 

In this case, I think it's fine to use the nplike, as long as we never try to bend the nplike to do too much single-library specific work.

PR #2141, opened 2023-01-19
feat: add ability to forget length of typetracer created with `typetracer_from_report`

douglasdavis:
One issue we've run into with the new typetracer-report based optimization in dask-awkward is slicing a typetracer array with "known" length created from a length zero array raises (see below)

This PR adds a new kwarg to `typetracer_with_report` to make it easy to create an unknown-length typetracer array with report.

This also raises a question: Should forgetting the length be the default?

```python
In [11]: a = ak.Array([{"x": [1, 2, 3]}, {"x": [2, 3]}])

In [12]: ak.Array(a.layout.form.length_zero_array().layout.to_typetracer())[0]
Cell In[12], line 1
----> 1 ak.Array(a.layout.form.length_zero_array().layout.to_typetracer())[0]

File ~/.pyenv/versions/3.11.1/envs/dev/lib/python3.11/site-packages/awkward/highlevel.py:956, in Array.__getitem__(self, where)
    527 """
    528 Args:
    529     where (many types supported; see below): Index of positions to
   (...)
    953 have the same dimension as the array being indexed.
    954 """
    955 with ak._errors.SlicingErrorContext(self, where):
--> 956     out = self._layout[where]
    957     if isinstance(out, ak.contents.NumpyArray):
    958         array_param = out.parameter("__array__")

File ~/.pyenv/versions/3.11.1/envs/dev/lib/python3.11/site-packages/awkward/contents/content.py:509, in Content.__getitem__(self, where)
    508 def __getitem__(self, where):
--> 509     return self._getitem(where)

File ~/.pyenv/versions/3.11.1/envs/dev/lib/python3.11/site-packages/awkward/contents/content.py:513, in Content._getitem(self, where)
    511 def _getitem(self, where):
    512     if ak._util.is_integer(where):
--> 513         return self._getitem_at(where)
    515     elif isinstance(where, slice) and where.step is None:
    516         return self._getitem_range(where)

File ~/.pyenv/versions/3.11.1/envs/dev/lib/python3.11/site-packages/awkward/contents/recordarray.py:300, in RecordArray._getitem_at(self, where)
    297     where += self.length
    299 if where < 0 or where >= self.length:
--> 300     raise ak._errors.index_error(self, where)
    301 return Record(self, where)

IndexError: while attempting to slice

    <Array-typetracer [...] type='?? * {x: var * int64}'>

with

    0

at inner RecordArray of length 0, using sub-slice 0.
```


agoose77:
If we were inventing this function for the first time, I think forgetting the length might be a sane default. That said, we have already shipped this in a release, so I think leaving the default as `False` is sensible here.

agoose77:
Ah, I'm mistaken; this is L3, so it doesn't hugely matter if we break it. I don't think anyone should be using this as released in the wild. When dask-awkward relies on this, it will always need to set awkward as a lower bound of 2.0.7 anyway, so I think change the default to forget the length by default; in this case, the length is entirely arbitrary anyway.

src/awkward/_typetracer.py
```diff
@@ -272,8 +272,10 @@ def _attach_report(layout, form, report):
         )
 
 
-def typetracer_with_report(form):
-    layout = form.length_zero_array(highlevel=False).to_typetracer()
+def typetracer_with_report(form, forget_length=False):
```
agoose77:
```suggestion
def typetracer_with_report(form, forget_length=True):
```

agoose77 (approved):

agoose77 (changes requested):
I agree with your default suggestion

agoose77:
@douglasdavis are you ready for this to merge :)

douglasdavis:
Yep I think this is good to go!

agoose77 (approved):
Merging!

PR #2142, opened 2023-01-19
ci: test NumPy < 1.17

agoose77:
My recollection of this test special case is that we want to ensure our oldest NumPy supported version is building & testing properly. I've set this to 1.14.5 for now, which is what `oldest-supported-numpy` reports for 3.7. This CI configuration was seemingly backwards.


jpivarski (approved):
This can be merged in version 2.1.0. It would be nice to have an alarm clock or something remind us about that, but I don't think there's any such feature. I could put it in my GTasks for late January.

agoose77:
@jpivarski I think this can be merged now, no? It tests our current lower bound, and we can raise it to 1.17 when we make our next minor release.

jpivarski:
Oh, I was reading it backward! Yes, we should be testing our current minimum bound. This PR can be merged now (and its inverse has to wait until 2.1.0).

PR #2145, opened 2023-01-20
fix: specify `dtype` for buffers in `from_rdataframe`.

agoose77:
I believe that we're over-allocating here, because the default dtype is usually a 64-bit float on _most_ systems. In any case, it's larger than a single byte, whilst I believe that we want to be allocating exactly `item.second` bytes.

@ianna are you able to check this logic for me? :)


agoose77:
Sorry for the review ping Ioana :facepalm:; muscle memory as you've reviewed a couple of my PRs recently!

src/awkward/_connect/rdataframe/from_rdataframe.py
```diff
@@ -71,7 +71,7 @@ def form_dtype(form):
     def empty_buffers(cpp_buffers_self, names_nbytes):
         buffers = {}
         for item in names_nbytes:
-            buffers[item.first] = numpy.empty(item.second)
+            buffers[item.first] = numpy.empty(item.second, dtype=np.uint8)
```
ianna, line 74: looks good!

ianna (approved):
Thanks for spotting it!

PR #2146, opened 2023-01-20
chore: add helpful message if cpp install is not prepared

### LINKED ISSUE #2140: Don't build sdist or wheels if nox hasn't been run
Related to #2136 is the point @ianna raised at our meeting today, that it's easy to miss a `nox -s prepare` call (or to call it with nox not installed and not see the error message). It would be sufficient to check for the existence of `awkward-cpp/include/awkward/kernels.h` during wheel-building (either `awkward-cpp` or `awkward`, since the recommended way to make both is with `pip`).

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/pull/2136#pullrequestreview-1262492186_

agoose77:
Fixes #2140 by informing the user during the build of `awkward-cpp` if the header-only directory or `include/awkward/kernels.h` are missing.

### DIFF
--- a/awkward-cpp/CMakeLists.txt
+++ b/awkward-cpp/CMakeLists.txt
@@ -16,9 +16,11 @@ if(NOT (EXISTS "header-only" AND EXISTS "include/awkward/kernels.h"))
   message(
     FATAL_ERROR
       "\
-awkward-cpp relies upon generated and copied artefacts such as the header-only libraries and generated kernel headers.\
-These could not be found, which indicates that the preparation step for building this package was skipped.\
-Please check CONTRIBUTING.md to learn more about this process.\
+awkward-cpp relies upon generated and copied artefacts such as the header-only libraries and generated kernel headers. \
+These could not be found, which indicates that\n\n\
+    nox -s prepare\
+\n\nwas skipped or failed. \
+Please check https://github.com/scikit-hep/awkward#installation-for-developers to learn more about this process.\
 ")
 endif()
 
--- a/awkward-cpp/CMakeLists.txt
+++ b/awkward-cpp/CMakeLists.txt
@@ -11,6 +11,17 @@ project(
 message(STATUS "CMake version ${CMAKE_VERSION}")
 message(STATUS "CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE}")
 
+# Check for header-only libraries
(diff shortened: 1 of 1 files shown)


awkward-cpp/CMakeLists.txt
```diff
@@ -11,6 +11,17 @@ project(
...
 message(STATUS "CMAKE_BUILD_TYPE = ${CMAKE_BUILD_TYPE}")
 
+# Check for header-only libraries
+if(NOT (EXISTS "header-only" AND EXISTS "include/awkward/kernels.h"))
+  message(
+    FATAL_ERROR
+      "\
+awkward-cpp relies upon generated and copied artefacts such as the header-only libraries and generated kernel headers.\
+These could not be found, which indicates that the preparation step for building this package was skipped.\
+Please check CONTRIBUTING.md to learn more about this process.\
```
jpivarski: This clickable URL (some terminals make URLs clickable) takes a user to the installation instructions, which in turn links to CONTRIBUTING.md for "more details."

But also, it includes the missing line to triple-click and middle-click to copy and paste on the spot. Sometimes I run commands that I know will fail because they'll give me the line I was supposed to type first, and and if they're easy enough to select and copy-paste, then I don't even need to type them.

```suggestion
awkward-cpp relies upon generated and copied artefacts such as the header-only libraries and generated kernel headers. \
These could not be found, which indicates that\n\n\
    nox -s prepare\
\n\nwas skipped or failed. \
Please check https://github.com/scikit-hep/awkward#installation-for-developers to learn more about this process.\
```

jpivarski (approved):
Below is a suggestion to further streamline the process for lazy devs.

awkward-cpp/CMakeLists.txt
jpivarski: I'm hoping that it gets rendered as

```
awkward-cpp relies upon generated and copied artefacts such as the header-only libraries and generated kernel headers. These could not be found, which indicates that

    nox -s prepare

was skipped or failed. Please check https://github.com/scikit-hep/awkward#installation-for-developers to learn more about this process.
```

to make the copy-pasteable line selectable with a triple-click.

agoose77:
image

jpivarski:
There are more newlines than I was expecting, and I'm surprised the CMake word-wraps the message, but it's definitely readable and triple-clickable.

PR #2148, opened 2023-01-20
feat: start hardening nplike signatures

agoose77:
This PR hardens[^1] the NumpyLike interface to accept concrete parameters, and require keyword-only usage of some arguments e.g. `dtype`.

It is the first in a series of PRs that will gradually split `_nplikes` into a package containing a file-per-nplike, and a subsequent full implementation of typetracer that removes `UnknownScalar`

I chose to introduce `maybe_reuse` which is similar to the existing `out` parameter that we use. The purpose of the rename is to make it clear that the nplike can ignore this, and the return result therefore may be a new array. I think we need this optimisation; it's used in broadcasting.

[^1]: "hardening" just means replacing `*args` with concrete types and names.

### DIFF
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -1048,7 +1048,7 @@ class TypeTracer(NumpyLike):
         # array
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def true_divide(
+    def divide(
         self,
         x1: TypeTracerArray,
         x2: TypeTracerArray,
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -917,7 +917,7 @@ class TypeTracer(NumpyLike):
         x1: TypeTracerArray,
         x2: TypeTracerArray,
         *,
-        maybe_reuse: TypeTracerArray | None = None,
+        maybe_out: TypeTracerArray | None = None,
     ) -> TypeTracerArray:
         try_touch_data(x1)
         try_touch_data(x2)
@@ -940,7 +940,7 @@ class TypeTracer(NumpyLike):
         x: TypeTracerArray,
         *,
         axis: int | None = None,
-        maybe_reuse: TypeTracerArray | None = None,
+        maybe_out: TypeTracerArray | None = None,
     ) -> TypeTracerArray:
         try_touch_data(x)
         raise ak._errors.wrap_error(NotImplementedError)
@@ -1035,14 +1035,14 @@ class TypeTracer(NumpyLike):
     ############################ ufuncs
 
     def sqrt(
-        self, x: TypeTracerArray, maybe_reuse: TypeTracerArray | None = None
+        self, x: TypeTracerArray, maybe_out: TypeTracerArray | None = None
     ) -> TypeTracerArray:
         try_touch_data(x)
         # array
         raise ak._errors.wrap_error(NotImplementedError)
 
     def exp(
-        self, x: TypeTracerArray, maybe_reuse: TypeTracerArray | None = None
+        self, x: TypeTracerArray, maybe_out: TypeTracerArray | None = None
     ) -> TypeTracerArray:
         try_touch_data(x)
         if np.issubdtype(x.dtype, np.integer):
@@ -1053,7 +1053,7 @@ class TypeTracer(NumpyLike):
         self,
         x1: TypeTracerArray,
         x2: TypeTracerArray,
-        maybe_reuse: TypeTracerArray | None = None,
+        maybe_out: TypeTracerArray | None = None,
     ) -> TypeTracerArray:
         try_touch_data(x1)
         try_touch_data(x2)
@@ -1064,7 +1064,7 @@ class TypeTracer(NumpyLike):
         self,
         x1: TypeTracerArray,
         x2: TypeTracerArray,
-        maybe_reuse: TypeTracerArray | None = None,
+        maybe_out: TypeTracerArray | None = None,
     ) -> TypeTracerArray:
         try_touch_data(x1)
         try_touch_data(x2)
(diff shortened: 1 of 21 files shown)


src/awkward/_typetracer.py
```diff
@@ -935,102 +988,119 @@ def concatenate(self, arrays, casting="same_kind"):
...
-        # array
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def unpackbits(self, *args, **kwargs):
-        for x in args:
-            try_touch_data(x)
-        # array
+    def packbits(
+        self, array: TypeTracerArray, *, axis=None, bitorder="big"
+    ) -> TypeTracerArray:
```
jpivarski: PyLint error. (I'm calling it out because they can be hard to find.)

jpivarski (commented):
How is `maybe_reuse` different from `out`? I see that it would no longer trouble `TypeTracerArray` because the `TypeTracer` functions simulate the situation by ignoring that argument. (It would be a side-effect whose success/failure status can only be determined at runtime.) This looks like a good _implementation_ of `out`, rather than an alternative that needs a new name.

On the name "maybe reuse," I got hung up on the "re," thinking that it referred to overwriting one of the input arguments. You mean "re" use the last argument, but as far as I can tell, that's exactly what NumPy's `out` does.

src/awkward/_typetracer.py
agoose77: > How is maybe_reuse different from out? I see that it would no longer trouble TypeTracerArray because the TypeTracer functions simulate the situation by ignoring that argument. (It would be a side-effect whose success/failure status can only be determined at runtime.) This looks like a good implementation of out, rather than an alternative that needs a new name.

Precisely that. The hint is `maybe_` which means you cannot rely on it as a side-effect. I don't mind changing names from NumPy; this API layer is already doing that, and the Array API doesn't define anything for reusing arrays like this. 

Would you prefer `maybe_out`?

jpivarski: Does this mean that we must always use it as

```python
outarray = nplike.function(inarray1, inarray2, maybe_out=outarray)
```

because if we did this:

```python
nplike.function(inarray1, inarray2, maybe_out=outarray)
```

then `outarray` would only get filled under some circumstances and not others?

(If that's the case, then definitely it needs a new name. `maybe_out` is nice because it links this with the already-known behavior of `out`.)

src/awkward/operations/ak_corr.py
```diff
@@ -150,5 +152,4 @@ def _impl(x, y, weight, axis, keepdims, mask_identity):
                 highlevel=True,
                 behavior=behavior,
             )
-        nplike = ak._nplikes.nplike_of(sumwxy, sumwxx, sumwyy)
-        return nplike.true_divide(sumwxy, nplike.sqrt(sumwxx * sumwyy))
```
agoose77, line 154: @jpivarski I dropped the use of the nplike here, because we're operating on high level `ak.Array`. In this PR, it's being made more explicit that nplikes only operate upon their low level arrays.

src/awkward/operations/ak_linear_fit.py
```diff
@@ -185,10 +187,10 @@ def _impl(x, y, weight, axis, keepdims, mask_identity):
...
             )
         delta = (sumw * sumwxx) - (sumwx * sumwx)
-        intercept = nplike.true_divide(((sumwxx * sumwy) - (sumwx * sumwxy)), delta)
-        slope = nplike.true_divide(((sumw * sumwxy) - (sumwx * sumwy)), delta)
-        intercept_error = nplike.sqrt(nplike.true_divide(sumwxx, delta))
-        slope_error = nplike.sqrt(nplike.true_divide(sumw, delta))
+        intercept = ((sumwxx * sumwy) - (sumwx * sumwxy)) / delta
+        slope = ((sumw * sumwxy) - (sumwx * sumwy)) / delta
+        intercept_error = _numpy_ufuncs.sqrt(sumwxx / delta)
+        slope_error = _numpy_ufuncs.sqrt(sumw / delta)
```
agoose77, line 193: I was originally thinking that we wanted to drop the use of `import numpy` within Awkward so that we can be confident no code is sneakily using it by accident.  However, the problem is that we need to invoke a ufunc via Awkward's ufunc machinery. Elsewhere, we also need the `np.ufunc` symbols in the modules that register awkward behaviours.

A forthcoming PR will replace the `NumpyMetadata` class with a `metadata` module, but that's for dtypes and constants rather than ufuncs. 

So, my solution here is to explicitly import numpy as `_numpy_ufuncs`, so that we could hypothetically implement a plugin to only allow imports of numpy that reflect it's intention to be used for e.g ufuncs only. It seems like a least-worst option.

@jpivarski do you have any feelings here?

src/awkward/_typetracer.py
agoose77: Yes, the idea is that you *must* assume that a copy occurs each time from the perspective of using the result.

Fab, let's use `maybe_out`.

agoose77:
@jpivarski this is ready to merge I think.

src/awkward/operations/ak_linear_fit.py
jpivarski: As long as that's how it's done consistently. The

```python
import numpy as _numpy_ufuncs   # be explicit about where NumPy is directly used
```

should probably have a comment everywhere it appears, so that someone else doesn't look at it, scratch their head, and change it back.

jpivarski (approved):
I'm in favor of merging this, but we need to do it one at a time with #2149 to make sure that they're both mergeable in the git sense (no line conflicts) and the tests run with both commits in.

I made a choice and paused the auto-merge on #2149, will merge this immediately, then update that and see what happens.

PR #2149, opened 2023-01-20
feat: implement all ufuncs on TypeTracer.

jpivarski:
This won't go in until #2148 is done, and then it will be on top of that (i.e. update this branch).

The first commit only adds code, and I'm 99% sure that it will merge easily with #2148.

What remains is to remove the `NotImplementedError` stubs of all the _ufuncs_ (only ufuncs) on `TypeTracer`.

Some of them have been implemented already, and if there are tests for the existing implementations, they would be a good check on the correctness of the general implementation.

Oops, I forgot to add broadcasting. Next commit.

### DIFF
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -665,20 +665,7 @@ class TypeTracerArray(NDArrayOperatorsMixin):
         def __getattr__(self, function_name):
             ufunc = getattr(numpy, function_name, None)
             if isinstance(ufunc, numpy.ufunc):
-
-                def generic_ufunc(*inputs, **kwargs):
-                    inputs = (self.array,) + inputs
-                    for x in inputs:
-                        getattr(x, "touch_data", lambda: None)()
-                    replacements = [
-                        numpy.empty(0, x.dtype) if hasattr(x, "dtype") else x
-                        for x in inputs
-                    ]
-                    result = ufunc(*replacements, **kwargs)
-                    return TypeTracerArray(result.dtype, shape=self.array._shape)
-
-                return generic_ufunc
-
+                return self._generic_ufunc_func(ufunc)
             else:
                 raise ak._errors.wrap_error(
                     AttributeError(
@@ -686,20 +673,30 @@ class TypeTracerArray(NDArrayOperatorsMixin):
                     )
                 )
 
+        def _generic_ufunc_func(self, ufunc):
+            def generic_ufunc(*inputs, **kwargs):
+                inputs = (self.array,) + inputs
+                for x in inputs:
+                    getattr(x, "touch_data", lambda: None)()
+                replacements = [
+                    numpy.empty(0, x.dtype) if hasattr(x, "dtype") else x
+                    for x in inputs
+                ]
+                result = ufunc(*replacements, **kwargs)
+                return TypeTracerArray(result.dtype, shape=self.array._shape)
+
+            return generic_ufunc
+
     @property
     def _module(self):
         # This is how ak._nplikes.NumpyLike calls functions that haven't been overloaded
         return TypeTracerArray._FakeModule(self)
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
-        self.touch_data()
-        for x in inputs:
-            getattr(x, "touch_data", lambda: None)()
-        replacements = [
-            numpy.empty(0, x.dtype) if hasattr(x, "dtype") else x for x in inputs
-        ]
-        result = getattr(ufunc, method)(*replacements, **kwargs)
-        return TypeTracerArray(result.dtype, shape=self._shape)
+        if method != "__call__" or len(inputs) == 0 or "out" in kwargs:
+            raise ak._errors.wrap_error(NotImplementedError)
+
+        return inputs[0]._module._generic_ufunc_func(ufunc)(*inputs[1:], **kwargs)
 
 
 def try_touch_data(array):
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -662,7 +662,7 @@ class TypeTracerArray(NDArrayOperatorsMixin):
         if method != "__call__" or len(inputs) == 0 or "out" in kwargs:
             raise ak._errors.wrap_error(NotImplementedError)
 
-        return TypeTracer._module._generic_ufunc_func(ufunc)(*inputs, **kwargs)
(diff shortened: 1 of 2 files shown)


agoose77:
BTW I've implemented some of the missing functions in the old PR: https://github.com/scikit-hep/awkward/blob/agoose77/refactor-harden-typetracer-next/src/awkward/_nplikes/typetracer.py#L480

I won't be copying over chunks of code directly, but the trivial logic is the same.

I'm currently in-the-process of porting this logic over, but not actively; there are steps between adding this implementation and where I'm at now.

jpivarski:
What I've implemented here is the equivalent of `broadcast_shapes` and `result_type`, using `broadcast_arrays` and operations on empty NumPy arrays, respectively.

Your implementation has no `true_divide`, whose result type can't be determined without knowing that it's a division operation (integer divided by integer is floating point).

If some `TypeTracerArrays` are going to get loose into the wild, even as `UnknownScalar` (zero-dimensional `TypeTracerArrays`, then they'll need all ufuncs, not just the ones we use in our codebase.

But maybe this is superfluous, after you make your changes and if we can keep a tighter rein on `TypeTracerArrays` in the wild (because Coffea will be doing Spark column-checking through Dask now).

That's why it's a perfect PR for you to review. You can decide if we need this or not; the answer may well be that we don't.

This PR will only pass tests once the explicit `TypeTracer.true_divide` overload is removed. I had intended for that to be a separate commit after merging with your PRs because removing lines of code is likely to be a merge conflict. The PR as-is mostly adds lines.

I'll make a remove-ufuncs commit after this comment, with the understanding that `main` should be merged in before that commit (so I may be reverting it, merging with `main`, and reapplying it).

src/awkward/_typetracer.py
```diff
@@ -659,12 +659,10 @@ def copy(self):
...
-        self.touch_data()
-        replacements = [
-            numpy.empty(0, x.dtype) if hasattr(x, "dtype") else x for x in inputs
-        ]
-        result = getattr(ufunc, method)(*replacements, **kwargs)
-        return TypeTracerArray(result.dtype, shape=self._shape)
+        if method != "__call__" or len(inputs) == 0 or "out" in kwargs:
+            raise ak._errors.wrap_error(NotImplementedError)
+
+        return TypeTracer._module._generic_ufunc_func(ufunc)(*inputs, **kwargs)
```
agoose77: I am just looking at the diff here, in theory it's possible to have a `TypeTracerArray` without a `TypeTracer` instance being initialised. That would be wrong, but let's be defensive?

```suggestion
        return TypeTracer.instance()._module._generic_ufunc_func(ufunc)(*inputs, **kwargs)
```

agoose77:
This PR is a quick way of fixing some of the bugs outstanding in typetracer. I think it will be temporary; in the future `TypeTracer` should inherit from a mostly empty protocol, but for now this will let dask-awkward keep moving!

My plan for our type handling is to use the existing NumPy rules with special cases for mixed scalars. Hopefully this means we won't need to carry around empty arrays for every operation on typetracers, but I suspect we'll want to use the empty-array implementation for ufunc support.

So, I think this PR can merge, and I'll end up touching this code again in the next month, but dask-awkward can progress in the meantime.

agoose77 (approved):

src/awkward/_typetracer.py
jpivarski: Good idea.

jpivarski:
Okay, I'll merge it. This can be seen as a short-term solution, with some of this undone by a later PR. However, it fixes @douglasdavis's immediate problem.

src/awkward/_typetracer.py
```diff
@@ -704,6 +704,42 @@ class TypeTracer(NumpyLike):
...
+                    if hasattr(x, "dtype") and hasattr(x, "shape"):
+                        replacements.append(numpy.empty(0, x.dtype))
+                        to_broadcast.append(x)
+                    else:
+                        replacements.append(x)
+
+                broadcasted = TypeTracer.instance().broadcast_arrays(*to_broadcast)
+
+                result = ufunc(*replacements, **kwargs)
+                return TypeTracerArray(result.dtype, shape=broadcasted[0].shape)
```
douglasdavis, line 737: This is raising an `AttributeError` in e.g. `ak.std`:
```python
>>> import awkward as ak
>>> a = ak.Array([[1, 2, 3], [4, 5]])
>>> tt = a.layout.to_typetracer(forget_length=True)
>>> tt = ak.Array(tt)
>>> tt
<Array-typetracer [...] type='?? * var * int64'>
>>> ak.std(tt, axis=1)
  File "/Users/ddavis/software/repos/awkward/src/awkward/operations/ak_std.py", line 84, in std
    return _impl(x, weight, ddof, axis, keepdims, mask_identity)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/awkward/src/awkward/operations/ak_std.py", line 172, in _impl
    return ak._nplikes.nplike_of(x, weight).sqrt(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/awkward/src/awkward/_nplikes.py", line 332, in sqrt
    return self._module.sqrt(x, out=maybe_out)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/awkward/src/awkward/_typetracer.py", line 737, in generic_ufunc
    return TypeTracerArray(result.dtype, shape=broadcasted[0].shape)
                           ^^^^^^^^^^^^
  File "/Users/ddavis/software/repos/awkward/src/awkward/highlevel.py", line 1125, in __getattr__
    raise ak._errors.wrap_error(AttributeError(f"no field named {where!r}"))
AttributeError: while calling

    ak.std(
        x = <Array-typetracer [...] type='?? * var * int64'>
        weight = None
        ddof = 0
        axis = 1
        keepdims = False
        mask_identity = False
    )

Error details: no field named 'dtype'. Did you mean: 'type'?
```

agoose77: That's quick! I'll take a look

agoose77: OK, I'm fairly confident this is because we're using Awkward's ufunc dispatching, as you're operating on the high level object. The two are distinct, and I'm touching this code in another PR. Let me spit that up so we can fix this bug.

PR #2150, opened 2023-01-20
refactor: move typetracer ufunc handling to backend [1 of 2]

agoose77:
This PR splits the `TypeTracer` ufunc logic handling from #2127, using the `Backend.apply_ufunc` machinery to implement `ufunc` operations on `ak.Array`s backed by `TypeTracer.

It implements "coercion" from `UnknownScalar`, `UnknownLength`, and `MaybeNone` values to the appropriate NumPy empty arrays, before applying the ufunc.

### DIFF
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -175,13 +175,18 @@ class TypeTracerBackend(Backend):
             if isinstance(x, ak._typetracer.TypeTracerArray):
                 x.touch_data()
                 shape = x.shape
-                numpy_args.append(numpy.empty((0,) + x.shape[1:], x.dtype))
+                numpy_args.append(numpy.empty((0,) + x.shape[1:], dtype=x.dtype))
+            # Convert scalars to 0-d arrays
+            elif isinstance(x, ak._typetracer.UnknownScalar):
+                numpy_args.append(numpy.empty((0,), dtype=x.dtype))
+            elif x is ak._typetracer.UnknownLength:
+                numpy_args.append(numpy.empty((0,), dtype=np.int64))
             else:
                 numpy_args.append(x)
 
         assert shape is not None
         tmp = getattr(ufunc, method)(*numpy_args, **kwargs)
-        return self._typetracer.empty((shape[0],) + tmp.shape[1:], tmp.dtype)
+        return self._typetracer.empty((shape[0],) + tmp.shape[1:], dtype=tmp.dtype)
 
 
 def _backend_for_nplike(nplike: ak._nplikes.NumpyLike) -> Backend:
--- a/src/awkward/_backends.py
+++ b/src/awkward/_backends.py
@@ -20,7 +20,10 @@ KernelType: TypeAlias = Callable[..., None]
 
 
 class Backend(Singleton, ABC):
-    name: str
+    @property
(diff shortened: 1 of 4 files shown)


src/awkward/_backends.py
```diff
@@ -163,6 +167,34 @@ def __init__(self):
...
+    def _coerce_ufunc_argument(self, x):
+        if isinstance(x, ak._typetracer.TypeTracerArray):
+            return numpy.empty((0,) + x.shape[1:], dtype=x.dtype)
+        # Convert scalars to 0-d arrays
+        elif isinstance(x, ak._typetracer.UnknownScalar):
+            return numpy.empty((0,), dtype=x.dtype)
+        elif x is ak._typetracer.UnknownLength:
+            return numpy.empty((0,), dtype=np.int64)
+        elif isinstance(x, ak._typetracer.MaybeNone):
+            return self._coerce_ufunc_argument(x.content)
```
agoose77, line 179: @jpivarski this is one part I'm hesitant on. We're throwing away type information here because this *might* not fail at runtime, i.e. if the value is not actually none.

src/awkward/_backends.py
```diff
@@ -20,7 +21,10 @@
 
 
 class Backend(Singleton, ABC):
-    name: str
+    @property
+    @abstractmethod
+    def name(self) -> str:
+        raise ak._errors.wrap_error(NotImplementedError)
```
agoose77, line 27: Misc cleanup.

src/awkward/_backends.py
agoose77: If we *do* want this, should `MaybeNone` also support operators (unrelated to this PR, but a more general point), e.g.. `MaybeNone(x) + y`? I'm inclined to say yes - `MaybeNone` should be fairly transparent, and serve as the outermost type like Haskell's `Maybe`.

agoose77: And, what should we do about `OneOf`? The *proper* approach would be to try all branches and re-wrap the result, but if one branch fails, we would need to handle that.

jpivarski: Our strategy is for typetracers to give code the benefit of the doubt and let it fail at runtime. The typetracer is not a type-check.

So if a typetracer (including `MaybeNone` or `OneOf`) represents $x$ or $y$ and $f(x)$ would yield some result but $f(y)$ would raise an error—great! We say that the result is $f(x)$.

If a typetracer represents $x$ or $y$ and both $f(x)$ and $f(y)$ would yield results, and they're different, that's a harder situation to deal with. I hope we don't have to because there might not be a good solution. We're not prepared to propagate non-deterministic type information. (If we had to, maybe we could go with the information-losing process of finding a common superclass of both results, like `np.numerical`.)

jpivarski (approved):
This looks good. I like it. Let's merge it!

src/awkward/_backends.py
agoose77: OK, that sounds like the best path to me, too. Let's go with the safe, minimal approach and see what happens. I haven't implemented `OneOf` here, but we can do that in another PR, as it's probably better to keep momentum of the changes that are already in the pipeline.

PR #2151, opened 2023-01-21
fix: support non-array types in `TypeTracerArray` ufunc operations

agoose77:
This PR is very similar to #2150. Whilst that PR deals with `np.ufunc()` operations on `ak.Array` objects, this PR is concerned with `np.ufunc()` operations on the _low level_ `TypeTracerArray` object. Like #2150, this PR adds support for treating scalar types (`UnknownScalar`, `MaybeNone`) as 0-D arrays.

### DIFF
--- a/src/awkward/_typetracer.py
+++ b/src/awkward/_typetracer.py
@@ -718,6 +718,18 @@ class TypeTracer(NumpyLike):
 
         @staticmethod
         def _generic_ufunc_func(ufunc):
+            def coerce_argument(x):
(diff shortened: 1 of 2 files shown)


agoose77:
In fact, I don't think we want to support arbitrary ufuncs on `TypeTracerArray`. The bigger vision here is to restrict the operations on nplike arrays to predominantly those on `nplike`. As such, we should remove this interface in favour of that approach.

PR #2152, opened 2023-01-21
refactor: split `_nplikes.py` into `_nplikes/*.py` [2 of 2]

agoose77:
This PR splits our large nplikes module into multiple modules within the `_nplikes` package. This makes it easier to scope changes to each nplike. Additionally:
- Implement `ArrayModuleNumpyLike` and `NumpyLike` distinction
- Remove naive ufunc support from `TypeTracerArray` (use ufuncs under the hood for e.g. `nplike.sqrt`).
- Replace `nplike.raw` with `to_nplike` so that each nplike is responsible for itself only.
- Fix accidental usage of NumPy aliases
- Add `ak._nplikes.ufuncs` for nice internal alias to NumPy ufuncs
   - We want to introduce a new `dtypes` module instead of `NumpyMetadata`, so that we have a canonical reference for the supported `dtype` object instances. Having the same concept for `ufuncs` is important too. It means that we can easily catch rogue `import numpy` usage.
- Adds new `flake8` check for imports of JAX, NumPy, or CuPy


agoose77:
I'm not sure on the type hints yet. Fundamentally, we want to use `NumpyLike` as a generic protocol, but that's not too useful practically as our contents will just be storing `ArrayType` implementers; we'd need to cast the array and nplikes to the appropriate types. I'll put that on the backburner for now.

Instead, I've written the nplikes to be invariant in the parameters, so according to the types, `Numpy` could accept `Cupy`'s arrays.

jpivarski (approved):
Okay, so now there's a distinction between `NumpyLike`, which means the same thing that it did before, and `ArrayModuleNumpyLike`, which applies to NumPy, CuPy, and JAX, but not to TypeTracer. These are different because TypeTracer has to define its own array class, `TypeTracerArray`, while the other three use already-existing array classes.

There's a lot of diff here, but this seems to be the main thing. Ufuncs are either already disabled from `TypeTracerArray` or soon will be, and there's more organization around getting general functions from an array library and getting ufuncs, and for metadata like `np.int32`.

The `raw` function is also gone. That had been used for casting an array in any nplike as a specific nplike. There must be some other way to do that now. Aha: `to_nplike`. That's a more informative name, too.

I approve! This can be merged when you're ready.

agoose77:
Yes, the distinction between `ArrayModuleNumpyLike` (yay for names) and `NumpyLike` is that `NumpyLike` is just a protocol (well, also a Singleton class). For now this is not hugely useful in the type hints, but we can revisit those down the road when we know better what we want to be doing here.

`TypeTracerArray` retains ufunc support in case we still need it down the road, but we'll switch that off internally during testing (a future PR).

The `awkward._nplikes.ufuncs` module is just an alias for the ufuncs from NumPy; due to NumPy's dispatch mechanism (`__array_ufunc__`), these are more like meaningful tokens than numpy-exclusive objects. I also will add a `_metadata` module for the `np.dtype` objects too, but I want to keep these PRs at least slightly similar to the code that they remove/replace.

Yes, with `raw` I felt that it would be better to implement this logic in a table form rather than per each nplike class. In this instance that seems particular feasible because we don't need N(N-1) conversion pathways, as most of the logic is relying upon the arrays having an `ArrayLike` interface only. 

Things that will happen later:
- [ ] Move typetracer to `awkward._nplikes`
- [ ] Add `awkward._nplikes.metadata`
- [ ] Validate dtypes in `nplike` functions to handle scalar-array behavior in e.g. NumPy
- [ ] Unify `UnknownScalar`, `UnknownLength`, and `TypeTracerArray`.

agoose77:
Thanks for the review, Jim!

PR #2153, opened 2023-01-21
chore: move to Ruff

henryiii:
For me, this reduces the time spent in pre-commit from about 12 seconds to 6.6 seconds. Ruff takes <1 second to process the entire codebase, and can do the work of several tools. I think "check YAML" is now the slowest check. :)

It has several checks that look very useful, and autofix is amazing, but this is a very minimal conversion to start with.

This should also stop the failures caused by plugins updating, since everything is one versioned binary.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -262,9 +262,7 @@ exclude = [
     "docs-*",
     "src/awkward/_typeparser/generated_parser.py",
 ]
-ignore = [
-    "E501", "C408", "E713", "E714",
-]
+ignore = ["E501", "C408", "E713", "E714", "UP030"]
 select = [
     "B0",
     "C",
@@ -272,6 +270,8 @@ select = [
     "F",
     "T20",
     "W",
+    "UP",
+    "I",
     "TID251"
 ]
 target-version = "py37"
@@ -293,6 +293,7 @@ mccabe.max-complexity = 100
     "F403",
     "F401",
     "F403",
+    "I001",
 ]
 "src/awkward/_connect/*" = [
     "TID251"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -253,3 +253,61 @@ module = [
 ]
 ignore_errors = true
 ignore_missing_imports = true
+
+[tool.ruff]
+exclude = [
+    "studies",
+    "pybind11",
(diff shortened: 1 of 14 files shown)


src/awkward/contents/content.py
```diff
@@ -13,6 +13,7 @@
 
 np = ak._nplikes.NumpyMetadata.instance()
 numpy = ak._nplikes.Numpy.instance()
+NumpyLike = ak._nplikes.NumpyLike
```
henryiii: This looked like a clear bug, so I fixed it instead of ignoring it. The other minor changes were auto fixes.

pyproject.toml
```diff
@@ -249,3 +249,45 @@ module = [
...
+
+[tool.ruff]
+exclude = [
+    "studies",
+    "pybind11",
+    "rapidjson",
+    "docs-*",
+    "src/awkward/_typeparser/generated_parser.py",
+]
+ignore = ["E501", "C408", "E713", "E714", "UP030"]
```
henryiii, line 265: Several of these eventually should be not skipped.

src/awkward/contents/content.py
agoose77: Fab! Our type hints are 100% human-only for now, but I will start getting mypy properly consuming them soon.

agoose77 (approved):

agoose77:
@jpivarski I'm happy with this, merge if you are too :)

jpivarski (approved):
I just tried it out, both with pre-commit and directly. The direct invocation found a not-real issue (UP003) because it doesn't know about `some_llvmite_type` in expressions like `some_llvmlite_type.type(0)`; it's thinking we're doing the free function `type(0)`, which for a constant like `0`, would be just `int`.

I added `noqa` in case someone else runs into this.

Okay, I guess I should report this to the ruff developers. Nope! It's already done: https://github.com/charliermarsh/ruff/issues/1958

There's a commit test that will remove the `noqa` when it's no longer needed. I'm going to remove the text of the comment other than the bare `noqa` so that it will be removed cleanly. (`noqa` removal tries to preserve the rest of the commit message, which will become nonsensical when the `noqa` is gone...)

Issue #2135, opened 2023-01-17
Typetracer arrays: support for more operations based on dask-awkward's necessary columns optimization


douglasdavis:
The new typetracer report optimization PR in dask-awkward (https://github.com/dask-contrib/dask-awkward/pull/141) is surfacing some features that are missing for typetracers. Creating this issue to track some findings from the dask-awkward test suite that have surfaced after implementing an optimization using the new typetracer data-touch feature.

It's definitely possible some of these may be issues in dask-awkward itself, so please let us know if you think that's the case.

- [ ] A slicing error:
	```
	  File "/Users/ddavis/.pyenv/versions/dev/lib/python3.11/site-packages/awkward/contents/recordarray.py", line 300, in _getitem_at
	    raise ak._errors.index_error(self, where)
	IndexError: while attempting to slice
	
	    <Array-typetracer [...] type='?? * {run: int64, luminosityBlock: int64,...'>
	
	with
	
	    0
	
	at inner RecordArray of length 0, using sub-slice 0.
	```
- [x] A truedivide `NotImplementedError`
- [x] `nplikes.Numpy.asarray` not compatible with `UnknownScalar`
	```
	    def asarray(self, *args, **kwargs):
	        # array[, dtype=][, order=]
	>       return self._module.asarray(*args, **kwargs)
	E       TypeError: int() argument must be a string, a bytes-like object or a real number, not 'UnknownScalar'
	```
- [ ] Error with building an array of UnknownScalars. In `ak_from_iter`:
	```
	        builder = _ext.ArrayBuilder(initial=initial, resize=resize)
	>       builder.fromiter(iterable)
	E       ValueError: cannot convert MaybeNone(UnknownScalar(dtype('int64'))) (type MaybeNone) to an array element
	E       
	E       (https://github.com/scikit-hep/awkward/blob/awkward-cpp-7/awkward-cpp/src/python/content.cpp#L191)
	
	allow_record = True
	behavior   = None
	builder    = <awkward_cpp.lib._ext.ArrayBuilder object at 0x148e4ff30>
	highlevel  = True
	initial    = 1024
	iterable   = [MaybeNone(UnknownScalar(dtype('int64'))), MaybeNone(UnknownScalar(dtype('int64'))), MaybeNone(UnknownScalar(dtype('int64')))]
	resize     = 1.5
	```

To actually run the test suite one can check out PR https://github.com/dask-contrib/dask-awkward/pull/141 

Cc: @martindurant

[...]

PR #2154, opened 2023-01-23
refactor: drop `UnknownScalar`, harden unknown scalar behavior

### LINKED ISSUE #2085: Expose typetracer to public API L1 and L2

Typetracer is going to become a public API beyond the `ak.backend` and `ak.to_backend` L1 functions. We will need to design an API that caters to all known parties, including dask-awkward and the work in Coffea (with PySpark).

TypeTracer now serves several purposes:
1. Predicting the forms of complex Awkward operations without requiring real data
2. Tracking usage of data through Awkward operations at the granularity of the individual buffers

Whilst (1) is already being used by dask-awkward, we would like to make this L2 instead of L3 if we're going to have additional users like coffea.

Coffea's use case for (2) is similar to dask-awkward, but, crucially, dask-awkward introduces a new API abstraction that does not attempt to behave like the usual `ak.Array` object. Meanwhile, coffea wants to take user-defined functions and support data tracking transparently, which is a much harder proposition.

In fact, the mechanism by which typetracer works is "tracing" — evaluating some bytecode and keeping track of what took place. Coffea would require that the traced execution matches the real-data execution, so that it knows precisely which tree branches are required by the program. This is not always possible; any time that the user writes a conditional statement, only a single branch can be taken by the interpreter, and the branch is data dependent. 

The main scenario that _we_ anticipate for coffea is that users wish to histogram data. `hist` and other histogramming packages expect a concrete array, i.e. something that `np.array(...)` returns a result for. Typetracer deliberately does not support this; typetracers have no data. Our suggestion to resolve this problem is twofold:

1. Add support for typetracers to the existing scikit-hep tools of interest, e.g. hist.
2. Add a mechanism by which users can explicitly declare data dependencies.

Both approaches may likely be solved by the introduction of a function `ak.empty_if_typetracer()`, which does two things:
1. touches typetracer array data to mark usage
2. returns an empty concrete array that can be used in histogramming, calculations etc.

For _users_, this function ensures that code expecting concrete arrays will JustWork™ with typetracers, whilst also enabling framework authors to understand that an array is used. For _libraries_, this function can easily be used to add support for typetracer data-tracking. 

@jpivarski suggested the following modification to our existing typetracer-array usage errors:

> Your code is being used in a preparatory phase without concrete data, but you're trying to access concrete data.
> Wrap this array with `ak.empty_if_typetracer(...)` to get an empty array with the right type and avoid this error.

```python
def empty_if_typetracer(array):
    layout = ak.to_layout(array, allow_other=False)
    behavior = ak._util.behavior_of(array)
    if layout.backend is ak._typetracer.TypeTracer.instance():
        layout._touch_data(True)
        return layout.form.length_zero_array(behavior=behavior)
    else:
        return ak._util.wrap(layout, behavior=behavior)
```

### LINKED ISSUE #2124: ak.isclose raises a NotImplemented error for typetracers
### Version of Awkward Array

2.0.6

In `awkward/_typetracer.py` line 1173, encountered running a dak meta through this function.

```python3
    def isclose(self, *args, **kwargs):
        for x in args:
            try_touch_data(x)
        # a, b, rtol=1e-05, atol=1e-08, equal_nan=False                                                                                                                                                                     
```

agoose77:
### Goals

At a high level, my long-term goal with typetracer is to make it look and feel mostly like any other nplike. Of course, we need to have special typetracer logic in routines that pull values out of arrays, etc, but for the most part this should be possible.

As such, this PR is full of a lot of line noise, but ultimately has the following goals:
- Unify `UnknownScalar` with `TypeTracerArray` (to reduce logic)
- Make concrete operations fail on unknown objects
- Unify typetracer under nplikes
- Fix some `NotImplementedError`s.

Out of scope for this PR is adding the mixed scalar-array guard logic, or creating a `metadata` / `dtypes` submodule for the counterpart to `ufuncs`.

### Changes
- [x] Closes #2124, closes #2085
- [x] Partly addresses https://github.com/scikit-hep/awkward/issues/2135
- [x] Replaces `UnknownScalar` with a runtime form of `TypeTracerArray`.
- [x] Treats any concrete operation like `bool()`, `int()`, etc. on an unknown scalar as an error, which means guarding our logic in the `Content` classes to protect from this. Whilst it was convenient to be able to write code that didn't need to do this, it required carrying around cognitive overhead to ensure that the code was written properly.
- [x] Moves towards `from X import Y` imports to make import chains more explicit (and reduce line length)
- [x] Removes the `__init__` imports for nplikes, as this encourages coupling.
- [x] Disables isort for certain modules
- [x] Fixes typestrings following behaviour change from scalar-array unification 

As it stands, this PR is probably too big to review. I might split it into smaller parts.


src/awkward/highlevel.py
```diff
@@ -181,18 +183,6 @@ def __init__(
             layout = data._layout
             behavior = ak._util.behavior_of(data, behavior=behavior)
 
-        elif numpy.is_own_array(data) and data.dtype != np.dtype("O"):
```
agoose77, line 184: Now this logic goes through `to_layout` after the addition of a new `regulararray` argument.

src/awkward/operations/ak_to_layout.py
```diff
@@ -94,6 +108,9 @@ def _impl(array, allow_record, allow_other):
 
         return ak.contents.NumpyArray(array, parameters=None, backend=backend)
 
+    elif ak._util.in_module(array, "pyarrow"):
+        return ak.operations.from_arrow(array, highlevel=False)
+
```
agoose77, line 113: Support arrow in `to_layout`

src/awkward/typetracer.py
```diff
@@ -0,0 +1,14 @@
...
+from awkward._nplikes import typetracer
+from awkward.typing import Any
+
+
+def is_unknown_scalar(array: Any) -> bool:
+    return typetracer.is_unknown_scalar(array)
+
+
+def is_unknown_array(array: Any) -> bool:
+    return typetracer.is_unknown_array(array)
```
agoose77: Tentative public API

dev/flake8_awkward.py
```diff
@@ -21,8 +21,8 @@ def visit_Raise(self, node):
             if isinstance(node.exc.func, ast.Attribute):
                 if node.exc.func.attr in {"wrap_error", "index_error"}:
                     return self.generic_visit(node)
-            if isinstance(node.exc.func, ast.Name):
-                if node.exc.func.id in {"ImportError"}:
+            elif isinstance(node.exc.func, ast.Name):
+                if node.exc.func.id in {"wrap_error", "index_error", "ImportError"}:
```
agoose77, line 25: Support bare `wrap_error`

src/awkward/_typetracer.py
```diff
@@ -1,1184 +1,22 @@
...
 
 
-def try_touch_shape(array):
-    if isinstance(array, TypeTracerArray):
-        array.touch_shape()
+class _UnknownScalarMeta(type):
+    def __instancecheck__(cls, instance):
+        from awkward._nplikes.typetracer import is_unknown_scalar
 
+        return is_unknown_scalar(instance)
```
agoose77, line 16: Simple metaclass to handle `isinstance(obj, UnknownScalar)`

src/awkward/_typetracer.py
```diff
@@ -1,1184 +1,22 @@
...
-
-    @classmethod
-    def is_own_array(cls, obj) -> bool:
-        return isinstance(obj, TypeTracerArray)
+    def __new__(cls, dtype):
+        from awkward._nplikes.typetracer import TypeTracer
 
-    def is_c_contiguous(self, x: ArrayLike) -> bool:
-        return True
+        return TypeTracer.instance().empty((), dtype=dtype)
```
agoose77, line 24: Simple constructor to handle `UnknownScalar(dtype)`

src/awkward/_nplikes/typetracer.py
```diff
@@ -0,0 +1,1112 @@
...
+        self.shape = shape
+        return self
+
+    def __repr__(self):
+        dtype = repr(self._dtype)
+        if self.ndim == 0:
+            shape = "??"
+        else:
+            shape = ", shape=" + repr(self._shape)
+        return f"TypeTracerArray({dtype}{shape})"
```
agoose77: Reprs are now different; >0D arrays always show length information

src/awkward/_nplikes/typetracer.py
```diff
@@ -294,60 +209,34 @@ def _length_after_slice(slice, original_length):
...
-        if array.dtype == np.dtype("O"):
-            raise ak._errors.wrap_error(
-                ValueError(
-                    f"bug in Awkward Array: attempt to construct `TypeTracerArray` "
-                    f"from a sequence of non-primitive types: {array}"
-                )
+    def __new__(cls, *args, **kwargs):
+        raise wrap_error(
+            TypeError(
+                "internal_error: the `TypeTracer` nplike's `TypeTracerArray` object should never be directly instantiated"
```
agoose77, line 217: This protects from people accidentally importing `TypeTracerArray` and operating upon it naively.

src/awkward/_nplikes/typetracer.py
```diff
@@ -678,7 +557,9 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         #         "TypeTracerArray objects should not be used directly with ufuncs"
         #     )
         # )
-        if method != "__call__" or len(inputs) == 0 or "out" in kwargs:
+        kwargs.pop("out", None)
+
+        if method != "__call__" or len(inputs) == 0:
```
agoose77, line 575: Remove error if `out` is passed.

I think this is harmless, and we need it for in-place operations upon scalars (though we could also disable that in the mixin)

agoose77:
@jpivarski sorry this is so big! Much of it is linenoise from imports.

The tests won't pass because there are changes to the `repr` of a typetracer high-level array. I need to check those tests out.

src/awkward/highlevel.py
```diff
@@ -181,18 +183,6 @@ def __init__(
...
-        elif ak._nplikes.Cupy.is_own_array(data):
-            layout = ak.operations.from_cupy(data, highlevel=False)
-
-        elif ak._nplikes.Jax.is_own_array(data):
-            layout = ak.operations.from_jax(data, highlevel=False)
-
-        elif ak._util.in_module(data, "pyarrow"):
-            layout = ak.operations.from_arrow(data, highlevel=False)
-
         elif isinstance(data, dict):
```
agoose77, line 186: Now this logic goes through `to_layout` after the addition of a new `regulararray` argument.

jpivarski (approved):
These are all good changes, many of which we've discussed for a while. It looks like this is all or most of the old PR, updated for modern times.

`ak.typetracer` is now a public interface, so we've got to be ready to support all of the non-underscored functions in it (as L2). It looks like a maintainable interface to me.

At this time of writing, it doesn't pass tests and needs to be updated to `main`, but it has my approval for when those things are fixed.

agoose77:
@jpivarski I have changed the typestring tests, because this PR changes typetracer reprs.

I've also made the `ArrayType.__eq__` more permissive; if any length is unknown, assume the lengths match.

jpivarski:
I'm not 100% happy with `__repr__` like `unknown-float64`, but anyway.

It still looks good! The only commit that `main` has ahead of this is tiny (https://github.com/scikit-hep/awkward/compare/agoose77/refactor-drop-unknown-scalar...main), so I'd say it's ready to be merged as-is.

agoose77:
@jpivarski this now passes all my local tests. Notable changes since your review:
- More places where we assume known scalars (dask test suite)
- All unknown scalars have the same typestring `##`. 
- `MaybeNone` has typestring `?{content}`
- `TypeTracerArray` can now be pickled (it has a special `__new__` to prevent a user every accidentally instantiating it).

agoose77:
Spoke too soon, a windows test revealed an assumption.

I'm now assuming that _any_ integral NumPy type is an `unknown_length`. This seems reasonable to me; it allows for any operation producing an integer to end up as a length, which is where I originally saw this going.

jpivarski:
In Slack, I backspaced over "`UnknownScalar(int64)`" since lengths might be 32-bit. I'm assuming that was the Windows error, right?

agoose77:
> In Slack, I backspaced over "`UnknownScalar(int64)`" since lengths might be 32-bit. I'm assuming that was the Windows error, right?

Bang on. I've relaxed this to "any" int type, which seems reasonable to me, as the concrete values that `UnknownLength` represent are in-practice Python scalars (i.e. unsized). I don't think we explicitly state anywhere that our `RegularArray.size` etc. values have to fit within `int64`/`int32`.

PR #2165, opened 2023-01-26
fix: update `type_to_name` for Layout builder `cxx_14` target

ianna:
- [x] temporary replacement helper variable templates for these that are part of C++17
- [x] `char` specialisation of `type_to_name` needed for Layout builder to generate a correct form

### DIFF
--- a/header-only/awkward/utils.h
+++ b/header-only/awkward/utils.h
@@ -84,7 +84,7 @@ namespace awkward {
   const std::string
   type_to_name<bool>() {
     // This takes precedence over the unspecialized template, and therefore any
-    // 8-bit data that is not named char will be mapped to "int8" or "uint8".
+    // 8-bit data that is not named bool will be mapped to "int8" or "uint8".
     return "bool";
   }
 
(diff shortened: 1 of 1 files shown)


header-only/awkward/utils.h
```diff
@@ -72,10 +84,18 @@ namespace awkward {
   const std::string
   type_to_name<bool>() {
     // This takes precedence over the unspecialized template, and therefore any
-    // 8-bit data that is not named bool will be mapped to "int8" or "uint8".
+    // 8-bit data that is not named char will be mapped to "int8" or "uint8".
     return "bool";
   }
```
jpivarski: I know that this is just a comment, but this function specifically catches the types that the compiler recognizes as `bool` at compile-time. Maybe that includes `char`; I actually don't know why the compiler considers some equivalent types "the same" and others "not the same" when matching a template instantiation. (In fact, I just hope it's not compiler-dependent!)

Did you miss the word "not" here? "8-bit data that is not named bool" is not handled here (returning the string `"bool"`); it's handled in the generic version of this function.

jpivarski (approved):
I'm in favor of this. Our current standard is C++14, so if `std::is_xyz_v<T>` is a C++17 thing, then we should define our own.

(Our C++ standard is constrained by the Docker images that pip uses to build wheels for many-Linux targets: they're tied to particular versions of gcc. I don't want to get into adding "download a new gcc" as a step in the wheel-building procedure; dependence on external downloads makes a build procedure more brittle.)

header-only/awkward/utils.h
ianna: oh, wrong comment :-)

ianna: thanks for spotting it - I've corrected the typo. Thanks

PR #2167, opened 2023-01-26
feat: coerce backends to same zero-copy type (2 of 2)

agoose77:
## TLDR
This PR
- coerces slice arrays to the same backend
- moves `reshape` and `astype` to `NumpyLike`
- overhauls `broadcast_arrays` and `reshape` for typetracer

## DR
Our high-level `ak.operations.*` mostly guard against mixed backends through `ak._backends.backend_of()`, which will complain if backends aren't compatible. If the user passes in arrays with mixed backends, an error will be raised. However, it doesn't then coerce them to the same type, meaning that for `TypeTracer` (which wins out over any other single backend), different backends can end up at different code paths. 

This PR ensures that an array and the item it is sliced with have the same backend, converting them where necessary.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -12,7 +12,8 @@ import awkward as ak
 from awkward._nplikes import nplike_of
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
-from awkward._nplikes.typetracer import TypeTracerArray, is_unknown_length
+from awkward._nplikes.typetracer import TypeTracerArray
+from awkward._util import unset
 from awkward.contents.bitmaskedarray import BitMaskedArray
 from awkward.contents.bytemaskedarray import ByteMaskedArray
 from awkward.contents.content import Content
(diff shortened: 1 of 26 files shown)


src/awkward/_do.py
```diff
@@ -247,7 +247,7 @@ def completely_flatten(
         return tuple(arrays)
 
 
-def flatten(layout: Content, axis: Integral = 1) -> Content:
+def flatten(layout: Content, axis: int = 1) -> Content:
```
jpivarski: Why make this less general? A user might pass in an `axis` value derived by NumPy, such that its type might be `np.uint32` or something, rather than `int`. That should be expected of all `axis` types. I think you had a type alias for `axis` types, with or without accepting `None` as a possibility.

jpivarski (approved):
Quite a large project! I see the things you were talking about: replacing binary operators like `+` with explicit functions like `nplike.add`, un-qualifying long names to make type hints more readable, and using `np.issubdtype` instead of `issubclass`.

It looks good to me (to the level of detail with which I can review a PR this large) and I'm in favor of merging it when you think it's ready.

src/awkward/_do.py
agoose77: Mypy doesn't support the `numbers.` type relations as type hints, so we have to be slightly less general. Right now, these hints are all internal, and we don't have any type hints (or many, if we do) at the L1 user-facing boundary in `ak.`, so we're only type checking internal-internal calls. 

My feelings here are that we'll sprinkle through some `cast` if we need to appease the type checker.

PR #2168, opened 2023-01-27
feat: use `None` for unknown lengths (1 of 2)

agoose77:
## TLDR
- This PR makes `None` the token for unknown lengths. Only `TypeTracer` actually produces a shape with `Nones`.
- Remove `UnknownLength = UnknownScalar(np.int64)` alias
- **Breaking change**: using `length=None` for unknowns means that `RecordArray(..., length=None)` needs to change behavior.[^change]
- Extend `nplike` to understand how to manipulate known and unknown lengths
- `NumpyLike.known_shape` now has concrete meaning: "does this NumpyLike support unknown dimension sizes?"

[^change]: This PR requires that `length` is unset in order to ignore it. We could use `if length is None and backend.nplike.known_shape` to warn users of this existing usage (as you can't have `None` lengths for known-shape nplikes)

## DR
This adds some verbosity to our length handling code in the `Content` methods, but this comes at a benefit in terms of explicitness. I chose to make shape-item-handling an nplike detail, rather than a a cross-nplike standard. This means that nplike `X` should not work with shapes from nplike `Y`. Instead, the nplike conversion routines should handle this. 

The previous approach set `UnknownLength = UnknownScalar(np.int64)`. Benefits include
- no need to coerce lengths to valid/invalid values (unknown scalars behave like integers)
-  single family of types (external unknown values are the same types as those returned by internal kernel operations and `Content.length`

The downsides of this are:
- form serialisation needs conversion to/from unknown scalars
- differs from Array API choice of `None` for unknown dimensions
- couples unknown shape-ness to unknown data-ness[^2]

This PR uses `UnknownLength = None`, making unknown sizes more obvious, and further increasing the "safety"	aspect by using a non-numeric type. `None` is easily serialised, and can be included in forms without serialisation.

The use of `None` means that external data needs to be treated carefully if it is used to compute lengths.

[^2]: I can't actually think of a need to create a new nplike that has one but not the other, but I digress.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -12,7 +12,7 @@ import awkward as ak
 from awkward._nplikes import nplike_of
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
-from awkward._nplikes.typetracer import TypeTracerArray, is_unknown_length
+from awkward._nplikes.typetracer import TypeTracerArray
 from awkward.contents.bitmaskedarray import BitMaskedArray
 from awkward.contents.bytemaskedarray import ByteMaskedArray
 from awkward.contents.content import Content
@@ -46,7 +46,7 @@ def length_of_broadcast(inputs: Sequence) -> int | TypeTracerArray:
 
     for x in inputs:
         if isinstance(x, Content):
-            if is_unknown_length(x.length):
+            if x.length is None:
                 return x.length
 
             maxlen = max(maxlen, x.length)
@@ -688,14 +688,14 @@ def apply_step(
                 for x in inputs
             ):
                 # Ensure all layouts have same length
-                length = None
+                length = ak._util.unset
                 for x in inputs:
                     if isinstance(x, Content):
-                        if length is None:
+                        if length is ak._util.unset:
                             length = x.length
                         elif backend.nplike.known_shape:
                             assert length == x.length
-                assert length is not None
+                # assert length is not None  TODO: validate this removal
 
                 if any(x.size == 0 for x in inputs if isinstance(x, RegularArray)):
                     dimsize = 0
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -13,6 +13,7 @@ from awkward._nplikes import nplike_of
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._nplikes.typetracer import TypeTracerArray
+from awkward._util import unset
 from awkward.contents.bitmaskedarray import BitMaskedArray
 from awkward.contents.bytemaskedarray import ByteMaskedArray
 from awkward.contents.content import Content
@@ -688,14 +689,14 @@ def apply_step(
                 for x in inputs
             ):
                 # Ensure all layouts have same length
-                length = ak._util.unset
+                length = unset
                 for x in inputs:
                     if isinstance(x, Content):
-                        if length is ak._util.unset:
+                        if length is unset:
                             length = x.length
                         elif backend.nplike.known_shape:
                             assert length == x.length
-                # assert length is not None  TODO: validate this removal
+                assert length is not unset
 
                 if any(x.size == 0 for x in inputs if isinstance(x, RegularArray)):
                     dimsize = 0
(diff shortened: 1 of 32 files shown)


src/awkward/_broadcasting.py
```diff
@@ -688,14 +689,14 @@ def continuation():
...
+                length = unset
                 for x in inputs:
                     if isinstance(x, Content):
-                        if length is None:
+                        if length is unset:
                             length = x.length
                         elif backend.nplike.known_shape:
                             assert length == x.length
-                assert length is not None
+                assert length is not unset
```
agoose77, line 699: I'm uneasy about how we handle unknown lengths here. There's a lot to consider in broadcasting, and I've not spent a lot of time in my mind on it. I'm hopeful that this PR mostly just renames `UnknownLength`, and therefore we can review this logic down the road.

src/awkward/contents/bytemaskedarray.py
```diff
@@ -353,31 +351,33 @@ def _carry(self, carry, allow_lazy):
...
                 self._mask.dtype.type,
             ](
-                numnull.data,
+                _numnull.data,
                 self._mask.data,
                 self._mask.length,
                 self._valid_when,
             )
         )
+        numnull = self._backend.index_nplike.as_shape_item(_numnull[0])
```
agoose77: This is a new "pattern". Any kernels that are used to determine a scalar (invariably a length) are prefixed with `_`, before conversion to a shape item with `as_shape_item`. This ensures we have only integers or `None`.

jpivarski (approved):
I'm glad that this is relative to the state in which `UnknownLength` still exists, so that it can be a renaming of that token, rather than `UnknownLength` → `UnknownScalar` → `None`.

So far it looks good, and we discussed it on Slack. I see there are a few test errors still, but you have my approval.

It's good to see more typing, too.

agoose77:
@jpivarski note that this PR doesn't yet address usage of `size=None` of `RegularArray`; it just sets the ground work for it.

agoose77:
With this PR, the distinction between indices (data) and lengths (metadata) is much stronger. Aligning closer with @jpivarski's original separation of length metadata from "data". 

They are now explicit groups, (`ShapeItem` and `ArrayLike`-scalars) with one-to-many mapping between the shape information and the offsets (shape items are non-negative, or None, with no fixed dtype)

Merging this PR will raise questions about `_getitem`. At present we implicitly assume `_getitem` only handles Python scalars or NumPy generic values. We need to broaden this both for all of the other nplikes, that use 0-D arrays as their scalar type.

agoose77:
@jpivarski what do you think about the future-breaking change to `RecordArray`? It only breaks existing users who pass in `length=None` under the typetracer backend; for `nplike.known_shape=True` we currently convert `length=None` to `length=unset`, because the `length=None` has no meaning for `known_shape=True`.

- [ ] Update deprecation schedule.

jpivarski:
We have a policy of not making backward-breaking changes without a deprecation period because it disrupts users and makes a tangle of which versions of interrelated libraries can work together. It is possible to be 100% strict about it, since we have defined what "public" means in a mechanically verifiable way (the underscores), but I'd rather be pragmatic.

This change would be hard to put through a deprecation cycle (waiting two minor versions before this PR can be merged, and all of the other work that depends on it), and the case that it affects is obscure: developers using typetracers (that's probably only Coffea and dask-awkward; it's a new interface) who explicitly pass `length=None` to the `RecordArray` constructor, rather than letting it be the default. If somebody's actually interested in a `RecordArray`'s `length`, it's because it might have zero fields (an unusual situation, but it's common to want to be defensive about such a possibility), and then one would want to pass an _integer_ length to not let it be derived from the possibly-nonexistent `contents`.

Even if we don't actually see a path to breakage through an interface change, we should still be careful because it could break in a way that we don't foresee. But this case is _so_ obscure that it's a risk I think we can take.

Here's a way that an unforseen issue could come in from something like this: if dask-awkward were mirroring the `RecordArray` interface with a `length=None` default, and they don't update to a `length=unset` default along with Awkward. Then any code that uses _their_ `RecordArray` without setting `length` would be saying that the length is unknown, rather than deriving it from `contents`. But dask-awkward doesn't wrap layouts, and it's hard to imagine anything that does in this way (and is also sensitive to typetracer).

We should call this kind of situation "reasonable risk" and not let it slide to such an extent that we're doing it all the time. In other words, we can only cheat if we think we can get away with it.  `:)`

PR #2171, opened 2023-01-30
fix: Layout builders clean and length bug fixes

ianna:
- [x] automate the tests checks

### DIFF
--- a/header-only/include/awkward/GrowableBuffer.h
+++ b/header-only/include/awkward/GrowableBuffer.h
@@ -54,6 +54,8 @@ namespace awkward {
           reserved_(reserved),
           next_(nullptr) {}
 
+    ~Panel() = default;
+
     /// @brief Overloads [] operator to access elements like an array.
     PRIMITIVE& operator[](size_t i) { return ptr_.get()[i]; }
 
--- a/header-only/include/awkward/LayoutBuilder.h
+++ b/header-only/include/awkward/LayoutBuilder.h
@@ -1067,14 +1067,14 @@ namespace awkward {
       clear() noexcept {
         for (size_t i = 0; i < fields_count_; i++)
           visit_at(contents, i, [](auto& content) {
-            content.builder.clear();
+            content.clear();
           });
       }
 
       /// @brief Current number of records in the first index of the tuple.
       size_t
       length() const noexcept {
-        return (std::get<0>(contents).builder.length());
+        return (std::get<0>(contents).length());
       }
 
       /// @brief Checks for validity and consistency.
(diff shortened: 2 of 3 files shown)


agoose77:
N.B. linting will be fixed by https://github.com/scikit-hep/awkward/pull/2153 or by upgrading isort. :)

header-only/awkward/LayoutBuilder.h
```diff
@@ -2247,12 +2247,16 @@ namespace awkward {
...
+        current_byte_ref_ = mask_.append_and_get_ref(current_byte_);
+        current_index_ = 0;
       }
 
       /// @brief Current length of the `mask` buffer.
       size_t
       length() const noexcept {
-        return (mask_.length() - 1) * 8 + current_index_;
+        return mask_.length() > 0 ?
+          (mask_.length() - 1) * 8 + current_index_ : current_index_;
```
jpivarski: Good catch!

I think `length()` is signed, so `length() - 1` doesn't roll over to `MAX_INT` (which would be even worse), but it still return a nonsensical (negative) result if `length()` is zero.

Oh—it's not! (In the ArrayBuilder code, I avoided the platform-dependent `size_t` and used `int64_t` instead. An argument could be made for using all `size_t` or all `int64_t`. The worst situation is using some here, some there. Compilers on 64-bit platforms don't catch it and may or may not warn about the signed-unsigned comparisons.)

So this really is fixing a worst case scenario: `length() - 1` is a crazy huge number when `length()` is zero.

jpivarski (approved):
Were the changes in GrowableBuffer.h and LayoutBuilder.h needed because the tests hadn't been running and the code became out-of-date?

Anyway, it's much better now that they are. Go ahead and merge this PR!

header-only/awkward/LayoutBuilder.h
ianna: > Were the changes in GrowableBuffer.h and LayoutBuilder.h needed because the tests hadn't been running and the code became out-of-date?
> 
No, it's because we did not test `clear()` and `length()` methods consistently. Now the test has lots of `asserts` :-)

> Anyway, it's much better now that they are. Go ahead and merge this PR!

Thanks!

jpivarski:
This update-to-`main` brings in #2153, which ought to make the pre-commit test pass. Then auto-merge will work.

jpivarski:
pre-commit passed; this is going to auto-merge.

agoose77:
This conflicts with my PR #2169, as I didn't see that this was auto-merging. I'll fix the conflicts and re-enable.

PR #2172, opened 2023-01-30
fix: support `mask_identity=True` for `axis=None` in `ptp`, `std`, etc.

agoose77:
In #2020, these tests were `xfail`ed. I must have done so whilst working locally, and committed the change. 

This PR restores support for `mask_identity=True` for the `axis=None` case of these composite reductions by pulling out the result as the final operation. As such, we choose what the identity value is for these reductions (`np.nan`, except for `ptp` which returns `0`)


agoose77:
Linting will be fixed by https://github.com/scikit-hep/awkward/pull/2153 or by upgrading `isort`

jpivarski (approved):
I see (unfinished work). It looks good to me; merge when you're ready.

PR #2177, opened 2023-01-30
fix: preserve dimensions for `keepdims=True`, `axis=None` reductions

### LINKED ISSUE #2159: Reducers with `axis=None`, `keepdims=True` differ from NumPy result
### Version of Awkward Array

main

The way that NumPy handles this is to preserve _all_ dimensions. However, we flatten all arrays to `1D` arrays before reduction, so we differ, and `keepdims` always returns a 1D array in this case.

```python
>>> np.max([[1,2,3],[4,5,6]], axis=None, keepdims=True)
array([[6]])
>>> ax.max([[1,2,3],[4,5,6]], axis=None, keepdims=True)
<Array [6] type='1 * ?int64'>
```

The most obvious solution is to extend `completely_flatten` with an argument that instructs lists to reconstruct themselves around the flattened result. Need to figure out a better name.

agoose77:
Fixes #2159 by adding a new `keepdims` to the renamed `ak._do.remove_structure` function. This just moves the list boundaries to the start and end elements of each list, so that the final result is a nested series of length-1 lists (besides the innermost dimension of length `N`).

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -771,6 +771,7 @@ def arrays_approx_equal(
     atol: float = 1e-8,
     dtype_exact: bool = True,
     check_parameters=True,
+    check_regular=True,
 ) -> bool:
     # TODO: this should not be needed after refactoring nplike mechanism
     import awkward.forms.form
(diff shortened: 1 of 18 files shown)


src/awkward/_util.py
```diff
@@ -798,7 +799,13 @@ def visitor(left, right) -> bool:
...
 
         if type(left) is not type(right):
-            return False
+            if not check_regular and (
+                left.is_list and right.is_regular or left.is_regular and right.is_list
+            ):
+                left = left.to_ListOffsetArray64()
+                right = right.to_ListOffsetArray64()
+            else:
+                return False
```
agoose77, line 808: We don't use this here, but I wrote it, and then removed the usage. I think it's good to keep for future tests.

jpivarski (approved):
This is mostly just a renaming, but it implements `keep_dims` for lists. :+1:

PR #2178, opened 2023-01-31
feat: growable buffer move_to method

ianna:
- [x] growable buffer `move_to` method implementation and test
- [x] growable buffer `clear` method bug fix
- [x] more thorough checks in the tests

### DIFF
--- a/header-only/include/awkward/GrowableBuffer.h
+++ b/header-only/include/awkward/GrowableBuffer.h
@@ -91,6 +91,12 @@ namespace awkward {
       return reserved_;
     }
 
+    /// @brief Unique pointer to the panel data.
+    std::unique_ptr<PRIMITIVE[]>&
+    data() {
+      return ptr_;
+    }
+
     /// @brief Copies the data from a panel to one contiguously allocated `to_ptr`.
     ///
     /// @param to_ptr One contiguously allocated panel.
@@ -408,6 +414,7 @@ namespace awkward {
       panel_ = std::move(std::unique_ptr<Panel<PRIMITIVE>>(
           new Panel<PRIMITIVE>((size_t)options_.initial())));
       ptr_ = panel_.get();
+      length_ = 0;
     }
 
     /// @brief Last element in last panel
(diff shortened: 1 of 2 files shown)


header-only/include/awkward/GrowableBuffer.h
```diff
@@ -408,6 +414,7 @@ namespace awkward {
       panel_ = std::move(std::unique_ptr<Panel<PRIMITIVE>>(
           new Panel<PRIMITIVE>((size_t)options_.initial())));
       ptr_ = panel_.get();
+      length_ = 0;
     }
```
jpivarski, line 418: This looks like a bug-fix. Is it?

jpivarski (approved):
I've always been a little confused about what `std::move` does. It seems pretty clear (in this example and others) that it copies data.

What's the difference between this new `move_to` function and the old `concatenate_to` function? Don't they both copy the panel data into an externally provided buffer and delete the panels?

I'm fine with making these changes (you can merge it when you're happy with it). I'm just asking.

header-only/include/awkward/GrowableBuffer.h
agoose77: > I've always been a little confused about what std::move does. It seems pretty clear (in this example and others) that it copies data.

I can chime in here! `std::move` just indicates to the compiler that the argument can be moved from, i.e. a consumer of the `std::move(...)` could avoid a copy by moving from the argument. So, it doesn't always eliminate a copy, depending upon how it is consumed. It's just a signal of lifetime intent IIRC.

ianna: > This looks like a bug-fix. Is it?

Yes, it is.

ianna:
> I've always been a little confused about what `std::move` does. It seems pretty clear (in this example and others) that it copies data.
> 
> What's the difference between this new `move_to` function and the old `concatenate_to` function? Don't they both copy the panel data into an externally provided buffer and delete the panels?
> 

The `concatenate_to` does not delete the panels. One has to call `clean` to do that. We can change it, I think. Then there will be just one method.

> I'm fine with making these changes (you can merge it when you're happy with it). I'm just asking.

jpivarski:
It would be nice if there were only one method. That way, developers don't have to guess which one to use.

Separating the functions of "copy the data out into a contiguous buffer" and "remove the panels" would keep more use-cases open. Maybe there's reason to clear the panels without copying out the data (user initiates a reset), and maybe there's a reason to copy the data out and keep filling (the original "snapshot" idea).

If `clear` is already a method, maybe `move_to` could take a boolean argument, specifying whether to clear after copying? Then, with `move_to` reproducing `concatenate_to`'s functionality, it should then _become_ `concatenate_to`, since we've already given out instructions on how to copy out the data, and they use the `concatenate_to` spelling.

PR #2181, opened 2023-01-31
fix: some usages of `len(layout)` under typetracer

agoose77:
This PR fixes a few (but certainly not all) places where we're using `len(layout)`, or otherwise relying on a layout to be non-typetracer


agoose77:
@lgray this would fix `dak.with_name()` without the meta bypass.

lgray:
Can this one be merged to main or waiting for something?

jpivarski (approved):
It was waiting for my review.

It's all good! Replacing `len(layout)` with `layout.length` is something we should be doing everywhere, ever since the `length` property was introduced (since it does not require the result to be an integer).

PR #2183, opened 2023-01-31
chore: ruff rewrite dicts

henryiii:
The built-in syntax is preferable to passing keywords to the dict constructor, which wastes a function call. Besides being more consistent and recognisable, it is also faster:

```pycon
Python 3.11.1 (main, Dec 23 2022, 09:39:26) [Clang 14.0.0 (clang-1400.0.29.202)] on darwin
>>> import timeit
>>> timeit.timeit("dict(a=1, b=2)")
0.09944732999429107
>>> timeit.timeit("{'a': 1, 'b': 2}")
0.07909351703710854
```

This removes the ignoring of `"C408"`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -263,8 +263,7 @@ exclude = [
     "src/awkward/_typeparser/generated_parser.py",
 ]
 ignore = [
-    "E501",
-    "C408",
+    "E501",   # line too long
     "UP030",  # {0} -> {}
 ]
 select = [
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1016,7 +1016,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         """
         with ak._errors.OperationErrorContext(
             "ak.Array.__setitem__",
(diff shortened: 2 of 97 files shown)


agoose77 (approved):

jpivarski (approved):
The dict constructor was only ever used in the entry-point from operations to their implementation functions. (See where the diffs are!)

The choice of dict constructor over built-in syntax was so that the function argument names would show up in the same color as they are in the entry-point function and the implementation function (as Python identifiers, not as strings). So there _was_ a logic to it.

But activating C408 means that other instances of dict constructors can't accidentally slip in, in the future, and there's value in that.

I'm okay with making this change.

henryiii:
We can deactivate it for a specific subset of files (assuming it can be matched on) if you'd prefer.

Actually, could `_errors.OperationErrorContext` take this as `**kwargs` instead? Then you remove the nesting, too.

agoose77:
In this case, I think we'd sooner replace the need to setup the context manager entirely with a decorator. We just haven't gotten around to it yet :)

PR #2184, opened 2023-02-01
chore: more Ruff

henryiii:
- chore: some Ruff additions
- chore: minor adjustment to Ruff
- chore: ruff iterable merging
- chore: RUF codes enabled


dev/generate-kernel-signatures.py
```diff
@@ -93,7 +93,8 @@
     "awkward_reduce_max",
     "awkward_reduce_min",
     "awkward_reduce_sum",
-    "awkward_reduce_sum_int32_bool_64" "awkward_reduce_sum_int64_bool_64",
+    "awkward_reduce_sum_int32_bool_64",
+    "awkward_reduce_sum_int64_bool_64",
```
agoose77, line 97: Yikes! Nice catch.

henryiii: That's what linters are for. :) (though Black helps here too by making it easy to see by eye, if someone was looking)

My hand rewrite of the template code is still slightly off, will have to find where things changed.

src/awkward/_connect/rdataframe/from_rdataframe.py
```diff
@@ -84,53 +81,52 @@ def cpp_builder_type(depth, data_type):
...
-                + "\n"
-                + f"  builder{depth}.end_list();\n"
-                + "}\n"
-                + "}\n"
+            return textwrap.dedent(
+                f"""
+                template<class BUILDER, typename PRIMITIVE>
+                void fill_offsets_and_flatten{depth}(BUILDER& builder{depth}, ROOT::RDF::RResultPtr<std::vector<PRIMITIVE>>& result) {{
+                    for (auto const& vec{depth - 1} : result) {{
+                      auto& builder{depth - 1} = builder{depth}.begin_list();\n"
```
henryiii:
```suggestion
                      auto& builder{depth - 1} = builder{depth}.begin_list();
```

Issue #2182, opened 2023-01-31
Simplifying UnionType during concatenation


grst:
### Description of new feature

In https://github.com/scikit-hep/awkward/issues/2058, you concluded that it is more useful if `ak.concatenate` simplifies types where possible instead of creating a Union. Here I have another example where I think this would be useful: 

**Current behavior**
```python
>>> a1 = ak.Array([{'a': 'aaa', 'b': 'bbb'}])
>>> a2 = ak.Array([{'b': 'bbb', 'c': 'ccc'}])

>>> res = ak.concatenate([a1, a2])
>>> res
[{a: 'aaa', b: 'bbb'},
 {b: 'bbb', c: 'ccc'}]
----------------------
type: 2 * union[
    {
        a: string,
        b: string
    },
    {
        b: string,
        c: string
    }
]
```

Like this, I can't even access fields anymore after concatenation: 

```python
>>> res['c']
FieldNotFoundError                        Traceback (most recent call last)
```

**Desired behavior**
Create an option type 
```
type: 2 * var * {
    a: ?string,
    b: string,
    c: ?string
}
```

FWIW, awkward infers exactly this option type when passing the concatenated array through a python object: 
```python
ak.Array(ak.to_list(res))
```

CC @ivirshup

agoose77:
I *think* this would be a useful feature, though probably hidden behind `merge_record` because there are times when this will lose information (if your fields are already option types).

jpivarski:
Yeah, about this kind of simplification—it's one that has come up before.

The option-type and union-type simplifications that we have now (`@classmethod simplified`) are simplifying the layouts/forms while maintaining the same high-level types. Here, I'm assuming type equivalences like

  * `option[option[X]]` is the same type as `option[X]` (each value can either be an instance of `X` or a missing value; the nested `option` does nothing visible at the level of values)
  * `option[union[X, Y]]` is the same type as `union[option[X], option[Y]]` and it's the same type as `union[option[X], Y]` (each value can either be an instance of `X`, an instance of `Y`, or a missing value).

Merging the types of `a1` and `a2` in

```
a1 = ak.Array([{'a': 'aaa', 'b': 'bbb'}])
a2 = ak.Array([{'b': 'bbb', 'c': 'ccc'}])
```

as

```
2 * var * {
    a: ?string,
    b: string,
    c: ?string
}
```

[...]

PR #2185, opened 2023-02-01
feat: add `ak.merge_union_of_records`

agoose77:
This PR addresses #2182. We already implement some of this machinery in `UnionArray`, which has a `_union_of_optionarrays` method to convert from `option[union[X, Y, Z]]` to `union[option[X], Y, ..., Z]`. It doesn't seem like that is particularly useful in this case. 

Interestingly there is some overlap with `ak._util.union_to_record`.

### Added functions
- `ak.merge_option_of_records`
- `ak.merge_union_of_records`

### DIFF
--- a/src/awkward/operations/ak_merge_option_of_records.py
+++ b/src/awkward/operations/ak_merge_option_of_records.py
@@ -1,0 +1,76 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+
+import awkward as ak
+from awkward._nplikes.numpylike import NumpyMetadata
+
+np = NumpyMetadata.instance()
+cpu = ak._backends.NumpyBackend.instance()
+
+
+def merge_option_of_records(array, axis=-1, *, highlevel=True, behavior=None):
+    """
+    Args:
+        array: Array-like data (anything #ak.to_layout recognizes).
+        axis (int): The dimension at which this operation is applied.
+            The outermost dimension is `0`, followed by `1`, etc., and negative
+            values count backward from the  innermost: `-1` is the innermost
+            dimension, `-2` is the next level up, etc.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+
+    Simplifies options of records, e.g.
+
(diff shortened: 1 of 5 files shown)


src/awkward/operations/ak_merge_union_of_records.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def _impl(array, axis, highlevel, behavior):
+    behavior = ak._util.behavior_of(array, behavior=behavior)
+    layout = ak.to_layout(array, allow_record=False)
+
+    def apply_displace_index(layout, backend, **kwargs):
+        if layout.is_record:
+            return layout
+        elif (layout.is_option or layout.is_indexed) and layout.content.is_record:
```
agoose77: This is doing something different for `is_option`. `merge_option_of_records`

agoose77:
@jpivarski what I'm not certain on here is what should happen for esoteric layouts that have multiple "option-of-record", and "union-of-record" in a single dimension. What should we do here, first wins? It seems like we do that in `ak.drop_none()`, but I don't recall us having a rule.

jpivarski (approved):
I added some tests which include the original problem in #2182 and a variation of it that has already-optional fields (to verify that the `simplified` classmethod is being used).

I also added tests for the `ak.merge_option_of_records`, and the result is wrong; it _drops_ `Nones`, rather than merging them into the record fields. I'm labeling this review as "Approve" because once that's fixed and the tests pass, it will be ready to merge.

Oh, I didn't add any tests for different `axis` values. Was that straightforward/formulaic? If it follows a pattern set by the many other functions that use `axis`, then I don't think we need to explicitly test it.

Darn it, I just convinced myself that I really do need to test it. One moment...

Okay, tested it.

src/awkward/operations/ak_merge_option_of_records.py
```diff
@@ -0,0 +1,88 @@
...
+            high-level.
+
+    Simplifies options of records, e.g.
+
+        >>> array = ak.Array([None, {"a": 1}, {"a": 2}])
+
+    into records of options, i.e.
+
+        >>> ak.merge_option_of_records(array)
+        <Array [{a: None}, {a: 1}] type='2 * {a: ?int64}'>
```
agoose77:
```suggestion
        <Array [{a: None}, {a: 1}, {a: 2}] type='3 * {a: ?int64}'>
```

PR #2190, opened 2023-02-02
feat: add support for histogram module

agoose77:
Add support for Hist's proposed `fill_flattened`: https://github.com/scikit-hep/hist/pull/474

### DIFF
--- a/src/awkward/_connect/hist.py
+++ b/src/awkward/_connect/hist.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+import numpy as np
+
+import awkward as ak
+from awkward._backends import NumpyBackend
+from awkward.typing import Any, Sequence
+
+numpy = NumpyBackend.instance()
+
+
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This adds functions in anticipation of https://github.com/scikit-hep/hist/pull/474 and does not break anything, so I think we can merge it in right away.

Usually, the `_connect` submodules are not imported at Awkward-initialization time so that they can freely `import` the third-party libraries at top-level in the submodule, but I can see why this one doesn't do that. Unless we have tests in environments without those third-party libraries, we'll have to be careful not to make Awkward accidentally depend on them at initialization time.

Following different rules about which `_connect` submodules can be imported immediately and which can only be imported if we know the user asked to work with a particular third-party module is complicated, but it's also an entirely internal matter that we can address at any time.

Since everything this PR does is perfectly safe, I'm going to merge it now.

PR #2194, opened 2023-02-02
fix: don't raise `NotImplementedError` when reading empty array from Parquet

### LINKED ISSUE #2193: Schema is not preserved when reading an empty `RecordArray` from parquet
### Version of Awkward Array

2.0.6

Hi!

I ran into what looks like a bug (or rather a non-implemented feature) when reading an `ak.RecordArray` stored in a Parquet file. The schema/type is not preserved when converting back to awkward if the stored array has no content (all columns have length zero).

This happens occasionally in a distributed workflow I'm running that involves merging a large number of arrays with the same structure (some of which could be empty) stored across multiple Parquet files. Preserving the schema even for the empty ones would make this easier to automate without having to filter out the empty arrays beforehand.

Here's a minimal example with an assertion on the array type to check schema preservation:
```python
import awkward as ak

# create a RecordArray with some content
a = ak.zip({"field_1": [1, 2], "field_2": [3, 4]})

# get an empty array (preserve schema)
a_empty = a[a['field_1'] < 0]

# write and read parquet
ak.to_parquet(a_empty, "test_empty.parquet")
a_empty_from_parquet = ak.from_parquet("test_empty.parquet")  # raises NotImplementedError

# check schema preserved
assert (a_empty_from_parquet.type == a_empty.type)
```

The `from_parquet` line raises a `NotImplemented` error:
```
  File "/.../awkward/test_parquet_empty.py", line 11, in <module>
    a_empty_from_parquet = ak.from_parquet("test_empty.parquet")  # gives NotImplementedError
  File "/.../awkward/src/awkward/operations/ak_from_parquet.py", line 75, in from_parquet
    return _load(
  File "/.../awkward/src/awkward/operations/ak_from_parquet.py", line 224, in _load
    _read_parquet_file(
  File "/.../awkward/src/awkward/operations/ak_from_parquet.py", line 304, in _read_parquet_file
    return ak.operations.ak_from_arrow._impl(
  File "/.../awkward/src/awkward/operations/ak_from_arrow.py", line 51, in _impl
    out = awkward._connect.pyarrow.handle_arrow(
  File "/.../awkward/src/awkward/_connect/pyarrow.py", line 1005, in handle_arrow
NotImplementedError:
```

This doesn't happen when round-tripping the array through Arrow, so it seems to be Parquet-related:
```python
a_empty_via_arrow = ak.from_arrow(ak.to_arrow(a_empty))  # works as expected
assert (a_empty_via_arrow.type == a_empty.type)
```

dsavoiu:
This PR proposes an implementation for reading an zero-length `ak.RecordArray` / Arrow `Table` from Parquet.

Since for zero-length there are no batches to use for conversion, an empty `RecordBatch` is created matching the input schema, and is propagated through the `handle_arrow` function. 

There might be better ways to do this (maybe avoiding the list-of-empty lists?), so let me know.

Closes #2193.

### DIFF
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -1001,8 +1001,13 @@ def handle_arrow(obj, generate_bitmasks=False, pass_empty_field=False):
     elif isinstance(obj, pyarrow.lib.Table):
         batches = obj.combine_chunks().to_batches()
         if len(batches) == 0:
-            # FIXME: create a zero-length array with the right type
-            raise ak._errors.wrap_error(NotImplementedError)
+            # create a RecordBatch of empty arrays following the schema
+            return handle_arrow(
+                pyarrow.lib.RecordBatch.from_arrays(
+                    [[] for _ in obj.schema.names],
+                    schema=obj.schema,
+                )
+            )
         elif len(batches) == 1:
             return handle_arrow(batches[0], generate_bitmasks, pass_empty_field)
         else:
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -1001,13 +1001,11 @@ def handle_arrow(obj, generate_bitmasks=False, pass_empty_field=False):
     elif isinstance(obj, pyarrow.lib.Table):
         batches = obj.combine_chunks().to_batches()
         if len(batches) == 0:
-            # create a RecordBatch of empty arrays following the schema
-            return handle_arrow(
-                pyarrow.lib.RecordBatch.from_arrays(
-                    [[] for _ in obj.schema.names],
-                    schema=obj.schema,
-                )
-            )
+            # create an empty array following the input schema
+            return form_handle_arrow(
+                obj.schema,
+                pass_empty_field=pass_empty_field,
+            ).length_zero_array(highlevel=False)
         elif len(batches) == 1:
             return handle_arrow(batches[0], generate_bitmasks, pass_empty_field)
         else:
--- a/tests/test_1125_to_arrow_from_arrow.py
+++ b/tests/test_1125_to_arrow_from_arrow.py
@@ -695,3 +695,11 @@ def test_unionarray(tmp_path, extensionarray):
     )
     paarray = akarray.to_arrow(extensionarray=extensionarray)
     arrow_round_trip(akarray, paarray, extensionarray)
+
+
+@pytest.mark.parametrize("extensionarray", [False, True])
+def test_empty_arrow(tmp_path, extensionarray):
+    akarray = ak.Array([{"x": 1, "y": 2.2}])[0:0]
+    paarray = ak.to_arrow_table(akarray, extensionarray=extensionarray)
+    arrow_round_trip(akarray, paarray, extensionarray)
+    parquet_round_trip(akarray, paarray, extensionarray, tmp_path)


jpivarski (changes requested):
This is a different solution from the one I had been thinking of, and if it were not for the fact that the `extensionarray=True` case leads to

```
  File "pyarrow/table.pxi", line 2460, in pyarrow.lib.RecordBatch.from_arrays
  File "pyarrow/table.pxi", line 1406, in pyarrow.lib._sanitize_arrays
  File "pyarrow/array.pxi", line 343, in pyarrow.lib.asarray
  File "pyarrow/array.pxi", line 317, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 39, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 144, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 121, in pyarrow.lib.check_status
pyarrow.lib.ArrowNotImplementedError: extension
```

I would have gone with it (with the minor correction of passing the arguments `generate_bitmasks` and `pass_empty_field` through the recursive call to `handle_arrow`). It's a good idea.

The other method, https://github.com/scikit-hep/awkward/issues/2193#issuecomment-1414271929, doesn't involve Arrow, so at least we won't run into any `ArrowNotImplementedError` that way.

Also, the PR needs tests. Other Arrow and Parquet conversion tests are parameterized for `extensionarray in (False, True)` so this would have been caught.

jpivarski:
It might be enough to just return this:

```python
form_handle_arrow(obj.schema, pass_empty_field=pass_empty_field).length_zero_array()
```

jpivarski:
Here is a test that triggers the error:

```python
ak.from_arrow(ak.to_arrow_table(ak.Array([{"x": 1, "y": 2.2}])[0:0], extensionarray=extensionarray))
```

(not actually Parquet-related). Tests would either go in a file named

```
tests/test_2194_fix_read_empty_parquet.py
```

or into the existing

```
tests/test_1125_to_arrow_from_arrow.py
```

Your choice, although we usually just make new files for new issues, and you can use the above as a template for parameterizing by `extensionarray`. The 1125 file has more bells and whistles than you'll probably want.

dsavoiu:
> It might be enough to just return this:
> 
> ```python
> form_handle_arrow(obj.schema, pass_empty_field=pass_empty_field).length_zero_array()
> ```

That does seems to do the trick! Looks like it needs to be `length_zero_array(highlevel=False)` though, otherwise the wrapping fails at the end.

> Also, the PR needs tests. [...] Tests would either go in a file named `tests/test_2194_fix_read_empty_parquet.py` or into the existing `tests/test_1125_to_arrow_from_arrow.py`. Your choice, although we usually just make new files for new issues, and you can use the above as a template for parameterizing by `extensionarray`. The 1125 file has more bells and whistles than you'll probably want.

Right, but putting the test in `1125` could make use of the `arrow_round_trip` and `parquet_round_trip` helper functions without needing to reimplement them or import them across tests. I've added a short test case there using the array you suggested.

jpivarski:
You're right about both points! Since I think you're at an inflection point, I'm going to do an "update branch" to make sure that it will work when up-to-date on `main`. Be sure to `git pull` before making any more changes.

jpivarski (approved):

jpivarski:
I've approved the changes, so if it passes tests, it's ready to be merged!

jpivarski:
@all-contributors please add @dsavoiu for code

PR #2196, opened 2023-02-02
fix: ak.values_astype now turns 'unknown' type into the requested type.

### LINKED ISSUE #2188: `ak.values_astype` should turn `EmptyArray` into `NumpyArray`
### Version of Awkward Array

HEAD

Follow-up from https://github.com/scikit-hep/uproot5/issues/822.

```python
>>> ak.values_astype(ak.Array([]), np.float64)
<Array [] type='0 * unknown'>
```

jpivarski:
...but I forgot that we wanted to parameterize it: https://github.com/scikit-hep/awkward/issues/2188#issuecomment-1412158519.

I'll do that when I can get back to this.


src/awkward/contents/content.py
```diff
@@ -1296,7 +1296,7 @@ def _is_equal_to(self, other: Self, index_dtype: bool, numpyarray: bool) -> bool
     def _repr(self, indent: str, pre: str, post: str) -> str:
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def _numbers_to_type(self, name: str) -> Self:
+    def _numbers_to_type(self, name: str, including_unknown: bool) -> Self:
```
agoose77: Now the return type changes! 😎 
```suggestion
    def _numbers_to_type(self, name: str, including_unknown: bool) -> Content:
```

agoose77 (approved):

PR #2201, opened 2023-02-03
docs: improve error message

### LINKED ISSUE #2200: ak.any Error message gives suggestion which will also error
### Version of Awkward Array

2.0.6

```python
import awkward as ak

ak.any(ak.Array([{"a": False}]))
```

Throws an error that says:

```pytb
TypeError: while calling

    ak.any(
        array = <Array [{a: False}] type='1 * {a: bool}'>
        axis = None
        keepdims = False
        mask_identity = False
        highlevel = True
        behavior = None
    )

Error details: cannot combine record fields unless flatten_records=True
```

But the suggested command also errors, telling me to do something else:

```python
ak.any(ak.Array([{"a": False}]), flatten_records=True)
```

```pytb
ValueError: while calling

    ak.any(
        array = <Array [{a: False}] type='1 * {a: bool}'>
        axis = None
        keepdims = False
        mask_identity = False
        highlevel = True
        behavior = None
    )

Error details: `flatten_records` is no longer a supported argument for reducers. Instead, use `ak.ravel(array)` first to remove the record structure and flatten the array.
```

`ak.ravel` should probably be suggested at the first point.

agoose77:
Fixes #2200 

@jpivarski oops, I should have left this for `one-hour-fix`. Sorry!

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -937,8 +937,7 @@ class RecordArray(Content):
                     (
                         "encountered a record whilst removing array structure{}, "
                         "but this operation does not support erasing records. "
-                        "Try calling `ak.ravel` or `ak.flatten(..., axis=None) first to "
-                        "remove the record"
+                        "Try first calling `ak.ravel` or `ak.flatten(..., axis=None)."
                     ).format(in_function)
                 )
             )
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
There is no need to avoid doing the `one-hour-fix` issues. (There will be plenty of them.) We shouldn't take that label as an "assigned to," since there's another button for that, but literally as "if you're looking for a quick thing to fix, do this one."

PR #2206, opened 2023-02-03
fix: ignore object arrays

agoose77:


jpivarski (approved):
I know it was rushed so I scanned over this PR very carefully. The notebook no longer tries to do something that shouldn't work (`ak.Array(np_array_dtype_O`) in code cells that aren't labeled by `raises-exception`.

Furthermore, `from_arraylib` short-circuits any `array` with `dtype=object`. All that assumes is that the object passed into this function has a `dtype` attribute, which is something that would be required eventually. (It's not a function that promotes array-like Sequences or anything like that.)

So this looks perfectly safe. Let's see if it solves the issue.

PR #2210, opened 2023-02-06
fix: support length-zero outer arrays in almost_equal

### LINKED ISSUE #2209: `ak.almost_equal` throws `IndexError` when passed datashape `0 * var * int64`
### Version of Awkward Array

2.0.7

```python
import awkward as ak

a = ak.Array([[1]])[0:0]
b = ak.Array([[1]])[0:0]

ak.almost_equal(a, b)
```

```pytb
Cell In[9], line 1
----> 1 ak.almost_equal(a, b)

File /usr/local/lib/python3.9/site-packages/awkward/operations/ak_almost_equal.py:134, in almost_equal(left, right, rtol, atol, dtype_exact, check_parameters, check_regular)
    131     else:
    132         raise wrap_error(AssertionError)
--> 134 return visitor(left, right)

File /usr/local/lib/python3.9/site-packages/awkward/operations/ak_almost_equal.py:95, in almost_equal.<locals>.visitor(left, right)
     88     return False
     90 if left.is_list:
     91     return (
     92         backend.index_nplike.array_equal(left.starts, right.starts)
     93         and backend.index_nplike.array_equal(left.stops, right.stops)
     94         and visitor(
---> 95             left.content[: left.stops[-1]], right.content[: right.stops[-1]]
     96         )
     97     )
     98 elif left.is_regular:
     99     return (left.size == right.size) and visitor(left.content, right.content)

File /usr/local/lib/python3.9/site-packages/awkward/index.py:196, in Index.__getitem__(self, where)
    195 def __getitem__(self, where):
--> 196     out = self._data[where]
    198     if hasattr(out, "shape") and len(out.shape) != 0:
    199         return Index(out, metadata=self.metadata, nplike=self._nplike)

IndexError: index -1 is out of bounds for axis 0 with size 0
```

On 2.0.6, `ak._util.arrays_approx_equal` would not throw an error here.

Found this from some tests starting to error when upgrading to 2.0.7 and swapping out the internal function for the new implementation.

agoose77:
- Fixes #2209 
- Adds `shapes_equal` to nplike with `None` return result for unknown shape items.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -165,11 +165,6 @@ class ArrayModuleNumpyLike(NumpyLike):
         slice_length = math.ceil((stop - start) / step)
         return start, stop, step, slice_length
 
-    def shape_equals(
-        self, shape1: tuple[ShapeItem, ...], shape2: tuple[ShapeItem, ...]
-    ) -> bool:
-        return shape1 == shape2
-
     def nonzero(self, x: ArrayLike) -> tuple[ArrayLike, ...]:
         return self._module.nonzero(x)
 
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -99,8 +99,10 @@ class ArrayModuleNumpyLike(NumpyLike):
 
     def array_equal(
         self, x1: ArrayLike, x2: ArrayLike, *, equal_nan: bool = False
-    ) -> bool:
-        return self._module.array_equal(x1, x2, equal_nan=equal_nan)
+    ) -> ArrayLike:
+        return self._module.asarray(
+            self._module.array_equal(x1, x2, equal_nan=equal_nan)
+        )
 
     def searchsorted(
         self,
@@ -118,7 +120,7 @@ class ArrayModuleNumpyLike(NumpyLike):
         return self._module.broadcast_arrays(*arrays)
 
     def reshape(
-        self, x: ArrayLike, shape: tuple[int, ...], *, copy: bool | None = None
+        self, x: ArrayLike, shape: tuple[ShapeItem, ...], *, copy: bool | None = None
     ) -> ArrayLike:
         if copy is False:
             raise ak._errors.wrap_error(
@@ -162,6 +164,11 @@ class ArrayModuleNumpyLike(NumpyLike):
         start, stop, step = slice_.indices(length)
         slice_length = math.ceil((stop - start) / step)
         return start, stop, step, slice_length
+
+    def shape_equals(
+        self, shape1: tuple[ShapeItem, ...], shape2: tuple[ShapeItem, ...]
+    ) -> bool:
+        return shape1 == shape2
 
     def nonzero(self, x: ArrayLike) -> tuple[ArrayLike, ...]:
         return self._module.nonzero(x)
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -165,6 +165,11 @@ class ArrayModuleNumpyLike(NumpyLike):
         slice_length = math.ceil((stop - start) / step)
         return start, stop, step, slice_length
 
+    def shape_equals(
+        self, shape1: tuple[ShapeItem, ...], shape2: tuple[ShapeItem, ...]
+    ) -> bool:
+        return shape1 == shape2
+
     def nonzero(self, x: ArrayLike) -> tuple[ArrayLike, ...]:
         return self._module.nonzero(x)
 
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -15,7 +15,6 @@ from awkward.typing import (
     Final,
     Literal,
     Self,
-    SupportsBool,
     SupportsIndex,
     TypeVar,
 )
@@ -887,25 +886,6 @@ class TypeTracer(NumpyLike):
                 slice_length = max(0, slice_length)
 
             return start, stop, step, self.index_as_shape_item(slice_length)
-
-    def shape_equals(
-        self, shape1: tuple[ShapeItem, ...], shape2: tuple[ShapeItem, ...]
-    ) -> SupportsBool:
-        if len(shape1) != len(shape2):
-            return False
-
-        result_is_known = True
-        for x, y in zip(shape1, shape2):
-            if x is unknown_length or y is unknown_length:
-                result_is_known = False
-                continue
-            elif x != y:
-                return False
-
-        if result_is_known:
-            return True
-        else:
-            return TypeTracerArray._new(np.bool_, shape=())
 
     def broadcast_shapes(self, *shapes: tuple[ShapeItem, ...]) -> tuple[ShapeItem, ...]:
         ndim = max([len(s) for s in shapes], default=0)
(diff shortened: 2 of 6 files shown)


src/awkward/_nplikes/array_module.py
```diff
@@ -97,8 +97,10 @@ def meshgrid(
 
     def array_equal(
         self, x1: ArrayLike, x2: ArrayLike, *, equal_nan: bool = False
-    ) -> bool:
-        return self._module.array_equal(x1, x2, equal_nan=equal_nan)
+    ) -> ArrayLike:
+        return self._module.asarray(
+            self._module.array_equal(x1, x2, equal_nan=equal_nan)
+        )
```
agoose77, line 105: We should not return concrete values in nplike functions that typetracer needs to implement.

src/awkward/_nplikes/array_module.py
```diff
@@ -116,7 +118,7 @@ def broadcast_arrays(self, *arrays: ArrayLike) -> list[ArrayLike]:
         return self._module.broadcast_arrays(*arrays)
 
     def reshape(
-        self, x: ArrayLike, shape: tuple[int, ...], *, copy: bool | None = None
+        self, x: ArrayLike, shape: tuple[ShapeItem, ...], *, copy: bool | None = None
```
agoose77, line 123: More out of date typehints

src/awkward/_nplikes/array_module.py
```diff
@@ -172,6 +174,11 @@ def div_shape_item(self, x1: ShapeItem, x2: ShapeItem) -> ShapeItem:
         assert result * x2 == x1
         return result
 
+    def shape_equals(
+        self, shape1: tuple[ShapeItem, ...], shape2: tuple[ShapeItem, ...]
+    ) -> bool:
+        return shape1 == shape2
+
```
agoose77, line 181: New shape comparison function. We use `None` to indicate "unknown". We already have 0-D `TypeTracerArray`s for unknown scalars, but I think this is distinct from arrays as it is a metadata operation. As such, we repurpose `None` (it now means no-length OR unknown-shape/length equality)

src/awkward/_nplikes/typetracer.py
```diff
@@ -869,6 +869,21 @@ def div_shape_item(self, x1: ShapeItem, x2: ShapeItem) -> ShapeItem:
...
+            return False
+
+        result_unknown = False
+        for x, y in zip(shape1, shape2):
+            if x is None or y is None:
+                result_unknown = True
+            elif x != y:
+                return False
+
+        return None if result_unknown else True
```
agoose77, line 885: Only `TypeTracer` actually returns `None`; when it encounters any unknown components the result is either `False` or `None` depending upon the remaining components.

src/awkward/operations/ak_almost_equal.py
```diff
@@ -48,6 +51,12 @@ def almost_equal(
     right = to_layout(right, allow_record=False).to_packed()
 
     backend = backend_of(left, right)
+    if not backend.nplike.known_data:
+        raise wrap_error(
+            TypeError(
+                "Awkward Arrays with typetracer backends cannot be compared with `ak.almost_equal`."
+            )
+        )
```
agoose77: We would not be able to produce very meaningful results here. We could return a 0-D scalar for not-known, but I chose just to make this an error. In general, we already know that boolean-like operations on arrays are not likely to be well supported in typetracer due to the need for concrete values.

That's why we have a proposal for `empty_if_typetracer`.

Not totally convinced here, though - I could be swayed.

src/awkward/operations/ak_almost_equal.py
```diff
@@ -88,21 +97,24 @@ def visitor(left, right) -> bool:
...
-            return (
-                backend.index_nplike.array_equal(left.starts, right.starts)
-                and backend.index_nplike.array_equal(left.stops, right.stops)
-                and visitor(
-                    left.content[: left.stops[-1]], right.content[: right.stops[-1]]
-                )
+            return backend.index_nplike.array_equal(
+                left.offsets, right.offsets
+            ) and visitor(
+                left.content[: left.offsets[-1]], right.content[: right.offsets[-1]]
```
agoose77, line 103: I didn't realise that all the list types (besides `ListArray`) now implement `offsets` too. This simplifies the code slightly.

src/awkward/operations/ak_almost_equal.py
```diff
@@ -88,21 +97,24 @@ def visitor(left, right) -> bool:
...
-                backend.nplike.isclose(
-                    left.data, right.data, rtol=rtol, atol=atol, equal_nan=False
+            return (
+                is_approx_dtype(left.dtype, right.dtype)
+                and backend.nplike.all(
+                    backend.nplike.isclose(
+                        left.data, right.data, rtol=rtol, atol=atol, equal_nan=False
+                    )
                 )
+                and backend.nplike.shape_equals(left.shape, right.shape)
```
agoose77: Fixed this logic, and also test shape!

src/awkward/operations/ak_almost_equal.py
agoose77: To add - if we support `known_data = False` here, then we need additional flags for what should happen if `left` has a known length but `right` has `unknown_length`. I think the right answer is probably to add typetracer support here, but I'm leaving this as a discussion point.

agoose77:
@jpivarski you tend to get to things within a couple of days, so I'm making an assumption that this has fallen off your radar. If you're just prioritising other things, no need to respond — I know that you'll see this.

src/awkward/operations/ak_almost_equal.py
jpivarski: `ListOffsetArray` and `RegularArray` are the only ones that have (guaranteed) contiguous `content`, so they're the only ones that can have an `offsets` property.

Note, however, that `RegularArray.offsets` has to create a buffer that didn't previously exist, so there's a performance cost.

src/awkward/operations/ak_almost_equal.py
jpivarski: The only meaningful thing that `ak.almost_equal` could do for typetracers would be to verify that they have the same high-level types.

If that short-circuits at the beginning of the function (if either array is typetracer, only check types), that would be another way of handling it, but then this error would become dead code.

It's also fine to raise errors if one is a typetracer array, as this one has it. That also allows us to put off a decision, since an error can be replaced by non-error behavior.

jpivarski (approved):
You're right to ping me in cases like this, where it seems that I've forgotten an open PR. (That is in fact what happened.) My workflow generally consists of responding to emails, and GitHub activity goes to my email.

Every now and then, I go to the PRs list and filter by "awaiting review from me."

agoose77:
Thanks @jpivarski. How do you feel about not supporting typetracer here? I suppose this could return an unknown boolean, so maybe I should change this.

jpivarski:
If this is primarily used for testing, not supporting typetracer would be fine: no one would miss it. If it's used as part of a regular analysis, it could easily come up when someone passes that into Dask. It could be NotImplementedError, and we can get back to it. (The implementation would be to return true if the types match, false otherwise. If the values don't match, they'll find out at runtime.)

src/awkward/operations/ak_almost_equal.py
agoose77: Well, typetracer can also check buffer lengths; in some cases these are known.

src/awkward/operations/ak_almost_equal.py
agoose77: Indeed. I'm assuming that this function is not a performance-critical path (if we're using this function in testing). Given that we already use `ak.to_packed`, deviating from this assumption might warrant a more layout-aware implementation of this function.

agoose77:
I want to add typetracer support here (it won't be too hard), but first I'd like to re-think how scalars behave in typetracer. So, this is a first-pass.

PR #2218, opened 2023-02-07
chore: include cuda kernel tests in sdist

agoose77:
Hatchling ignores anything that is ignored by the VCS by default. We can use `artefacts` to explicitly include e.g. the generated CUDA kernel tests.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -64,7 +64,7 @@ init = "awkward.numba:_register"
 
 [tool.hatch.build]
 artifacts = [
-    "src/awkward/_connect/header-only"
+    "/src/awkward/_connect/header-only"
 ]
 
 [tool.hatch.build.targets.wheel]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -64,7 +64,7 @@ init = "awkward.numba:_register"
 
 [tool.hatch.build]
 artifacts = [
-    "src/awkward/_connect/header-only"
+    "/src/awkward/_connect/header-only"
```
jpivarski, line 67: `src` is not in the root of the filesystem, but I suppose this filename path is interpreted by hatchling to be relative to the base of the package or source repo.

jpivarski (approved):
I'll take your word for it!

(Sounds like a terrible review, doesn't it? But in this PR, it's clear that you just needed to move `tests-cuda-kernels` from one category to another for some technical reason in hatchling. The evidence that it was the right thing to do is that it got those header files into the package, and not doing this didn't. Current tests wouldn't have been sensitive to that change, since we're not routinely testing CUDA kernels yet.)

pyproject.toml
agoose77: Yep, Hatchling uses [Git-style glob pattern](https://git-scm.com/docs/gitignore#_pattern_format) to match these entries. This doesn't "fix" any bugs, but it's good to be explicit that we're talking about one `src` directory only :)

agoose77:
> Sounds like a terrible review, doesn't it?

Haha, sometimes less is more!

PR #2219, opened 2023-02-07
fix: keep `EmptyArray` in `remove_structure`

### LINKED ISSUE #2207: Different behavior for awkward 2.x on ak.flatten with unknown type
### Version of Awkward Array

2.0.6

In some of my code, I frequently check that there are entries left in the array before trying to proceed to the next step. My use case is calling out to some c++ code to do jet finding event by event, so if there are no jets found, it eventually returns a full array of events, all of which are empty.

In awkward 1.x, I used `ak.flatten(array["data"].px, axis=None)` to flatten out a record (px is one of the fields, not calculated by vector) to check for entries (the particular record didn't really matter - I just needed one). In awkward 2.x, if I do the same, I get an assertion error:

```
In [13]: ak.flatten(jets["data"].px, axis=None)
AssertionError                            Traceback (most recent call last)
Cell In [13], line 1
----> 1 ak.flatten(jets["data"].px, axis=None)

File ~/software/dev/mammoth/.venv/lib/python3.10/site-packages/awkward/operations/ak_flatten.py:161, in flatten(array, axis, highlevel, behavior)
      9 """
     10 Args:
     11     array: Array-like data (anything #ak.to_layout recognizes).
   (...)
    155      999]
    156 """
    157 with ak._errors.OperationErrorContext(
    158     "ak.flatten",
    159     dict(array=array, axis=axis, highlevel=highlevel, behavior=behavior),
    160 ):
--> 161     return _impl(array, axis, highlevel, behavior)

File ~/software/dev/mammoth/.venv/lib/python3.10/site-packages/awkward/operations/ak_flatten.py:173, in _impl(array, axis, highlevel, behavior)
    168     out = ak._do.completely_flatten(layout, function_name="ak.flatten")
    169     assert isinstance(out, tuple) and all(
    170         isinstance(x, ak.contents.NumpyArray) for x in out
    171     )
--> 173     result = ak._do.mergemany(out)
    175     return ak._util.wrap(result, behavior, highlevel)
    177 elif axis == 0 or ak._util.maybe_posaxis(layout, axis, 1) == 0:

File ~/software/dev/mammoth/.venv/lib/python3.10/site-packages/awkward/_do.py:303, in mergemany(contents)
    302 def mergemany(contents: list[Content]) -> Content:
--> 303     assert len(contents) != 0
    304     return contents[0]._mergemany(contents[1:])

AssertionError:
```

Some more info on the array:

```python
In [12]: jets.type.show()
81 * {
    data: var * Momentum4D[
        px: unknown,
        py: unknown,
        pz: unknown,
        E: unknown,
        area: unknown,
        rho_value: float64,
        constituents: var * Momentum4D[
            px: float64,
            py: float64,
            pz: float64,
            E: float64,
            index: int64
        ],
        unsubtracted_leading_track_pt: unknown
    ]
}
```

It seems to be about the "unknown" nature of the type, since if I do the same with `pt`, vector calculates that field, which then types it with `float64`, and the call works. Also note that if I flatten without axis=None, it works fine. A pickle with the jets array is attached [regression_jets.pkl.zip](https://github.com/scikit-hep/awkward/files/10593610/regression_jets.pkl.zip) (with .zip as the extension to allow it to be uploaded). I can in principle switch over to `pt`, but it would be nicer if this worked as in awkward 1.x . Thanks!

edit: I know for sure that the behavior changed, but now I'm a bit less confident it is truly a regression as opposed to the possibility that it's an intended change in behavior. If so, I suppose please let me know, and I'll have to dig through my codebase to change it

agoose77:
Fixes #2207 by preserving `EmptyArray` in `ak._do.remove_structure`. This means that code that previously did not handle `EmptyArray` would now be expected to. I don't think there are many places where this is actually a problem; `EmptyArray` already handles reduction, and flattening almost immediately returns the result.

### DIFF
--- a/src/awkward/contents/emptyarray.py
+++ b/src/awkward/contents/emptyarray.py
@@ -318,7 +318,7 @@ class EmptyArray(Content):
         return backend.nplike.empty(0, dtype=np.float64)
 
     def _remove_structure(self, backend, options):
-        return []
+        return [self]
 
     def _recursively_apply(
         self, action, behavior, depth, depth_context, lateral_context, options
--- a/src/awkward/operations/ak_flatten.py
+++ b/src/awkward/operations/ak_flatten.py
@@ -168,7 +168,7 @@ def _impl(array, axis, highlevel, behavior):
     if axis is None:
         out = ak._do.remove_structure(layout, function_name="ak.flatten")
         assert isinstance(out, tuple) and all(
-            isinstance(x, ak.contents.NumpyArray) for x in out
+            isinstance(x, ak.contents.Content) for x in out
         )
 
         result = ak._do.mergemany(out)
--- a/tests/test_2219_flatten_empty.py
+++ b/tests/test_2219_flatten_empty.py
@@ -1,0 +1,12 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import pytest  # noqa: F401
+
+import awkward as ak
+
+
+def test():
+    array = ak.Array([[[], [], []], []])
+    flattened = ak.flatten(array, axis=None)
+    assert isinstance(flattened.layout, ak.contents.EmptyArray)
+    assert len(flattened) == 0


jpivarski (approved):
> This means that code that previously did not handle `EmptyArray` would now be expected to.

The thing I would do in this situation is to grep for all the calls to `remove_structure` and check that all of those call sites would be able to handle an `EmptyArray`. The fact that the test suite passes is half as good as that.

jpivarski:
It seems to be _entirely_ in `ak.flatten` and `ak.ravel`:

```
% fgrep -r '._remove_structure(' src/awkward
src/awkward/contents/bitmaskedarray.py:            return self.project()._remove_structure(backend, options)
src/awkward/contents/listarray.py:        return self.to_ListOffsetArray64(False)._remove_structure(backend, options)
src/awkward/contents/unmaskedarray.py:            return self.project()._remove_structure(backend, options)
src/awkward/contents/indexedarray.py:        return self.project()._remove_structure(backend, options)
src/awkward/contents/unionarray.py:                ._remove_structure(backend, options)
src/awkward/contents/listoffsetarray.py:            contents = content._remove_structure(backend, options)
src/awkward/contents/indexedoptionarray.py:            return self.project()._remove_structure(backend, options)
src/awkward/contents/bytemaskedarray.py:            return self.project()._remove_structure(backend, options)
src/awkward/contents/regulararray.py:            contents = content._remove_structure(backend, options)
src/awkward/contents/recordarray.py:                out.extend(content[: self._length]._remove_structure(backend, options))
src/awkward/_do.py:        arrays = layout._remove_structure(
```

```
% fgrep -r 'do.remove_structure(' src/awkward
src/awkward/operations/ak_flatten.py:        out = ak._do.remove_structure(layout, function_name="ak.flatten")
src/awkward/operations/ak_ravel.py:    out = ak._do.remove_structure(layout, function_name="ak.ravel", drop_nones=False)
```

jpivarski:
Both of them send the result to `mergemany` and nothing else. `mergemany` can handle `EmptyArray`, so there's no issue there.

agoose77:
I think my wording could have been better here; I am generally hesitant to change anything that we likely are making assumptions about. I did the same checks you did to be as confident as possible. I am glad you found the same results! (occasionally I have missed something like this despite thinking otherwise!)

PR #2220, opened 2023-02-07
refactor: change `Content._getitem_range` to use explicit indices

agoose77:
## TL;DR
This PR
- Changes `Content._getitem_range(slice(start, stop, None))` to `Content._getitem_range(start, stop)`
- Changes `nplike.shape_item_as_scalar` to `shape_item_as_index` etc. to support mixed known-unknown indices

## TL
The motivation for this PR is that we're constructing slice objects to immediately unpack them in many places. The `slice.indices` method that we use to unpack these slices does not know about `length = None`. Therefore, it makes sense just to forgo all of this by passing around fixed index values.

It became apparent in the work on typetracer that `shape_item_as_scalar` could not just return the 0D array of the nplike. This is because known-index values are useful; it is possible to slice a known-length content with known start & stop values such that the result shape is known. Therefore, we need to support known values alongside unknown values. 

Therefore, I've renamed `shape_item_as_scalar` so that we have a new type concept "index", which is _either_ a pure-Python integer, or a 0D array of the index type. I think this is sufficiently straightforward that it isn't opening too big a can of worms.

This is how we move between types, now:

```mermaid
flowchart LR;
shape_item[/Shape Item/]
index[/Index/]
offset[/Offset/]
shape_item -->|shape_item_as_index| index
index -->|index_as_shape_item| shape_item
index --> |nplike.asarray| offset
offset --> |index_as_shape_item| shape_item
```

Clearly, offsets are just unknown indices.

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -10,12 +10,8 @@ import awkward as ak
 from awkward._backends import Backend
 from awkward._nplikes import to_nplike
 from awkward._nplikes.numpy import Numpy
-from awkward._nplikes.numpylike import (
-    NumpyLike,
-    NumpyMetadata,
-    ShapeItem,
-    unknown_length,
-)
+from awkward._nplikes.numpylike import NumpyLike, NumpyMetadata
+from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._nplikes.typetracer import TypeTracer
 from awkward._util import unset
 from awkward.forms.form import Form, JSONMapping, _type_parameters_equal
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -13,7 +13,7 @@ from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyLike, NumpyMetadata
 from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._nplikes.typetracer import TypeTracer
-from awkward._slicing import regularise_slice
+from awkward._slicing import regularize_slice
 from awkward._util import unset
 from awkward.forms.form import Form, JSONMapping, _type_parameters_equal
 from awkward.index import Index, Index64
@@ -538,7 +538,7 @@ class Content:
 
         elif isinstance(where, slice) and where.step is None:
             # Ensure that start, stop are non-negative!
-            start, stop, _ = regularise_slice(where, self.length, backend=self._backend)
+            start, stop, _ = regularize_slice(where, self.length, backend=self._backend)
             return self._getitem_range(start, stop)
 
         elif isinstance(where, slice):
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -13,6 +13,7 @@ from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyLike, NumpyMetadata
 from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._nplikes.typetracer import TypeTracer
+from awkward._slicing import regularise_slice
 from awkward._util import unset
 from awkward.forms.form import Form, JSONMapping, _type_parameters_equal
 from awkward.index import Index, Index64
@@ -536,7 +537,9 @@ class Content:
             return self._getitem_at(where)
 
         elif isinstance(where, slice) and where.step is None:
-            return self._getitem_range(where)
+            # Ensure that start, stop are non-negative!
+            start, stop, _ = regularise_slice(where, self.length, backend=self._backend)
+            return self._getitem_range(start, stop)
 
         elif isinstance(where, slice):
             return self._getitem((where,))
@@ -703,7 +706,7 @@ class Content:
     def _getitem_at(self, where: SupportsIndex):
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def _getitem_range(self, where: slice) -> Content:
+    def _getitem_range(self, start: SupportsIndex, stop: SupportsIndex) -> Content:
         raise ak._errors.wrap_error(NotImplementedError)
 
     def _getitem_field(
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -10,7 +10,7 @@ import awkward as ak
 from awkward._backends import Backend
 from awkward._nplikes import to_nplike
 from awkward._nplikes.numpy import Numpy
-from awkward._nplikes.numpylike import NumpyLike, NumpyMetadata
+from awkward._nplikes.numpylike import IndexType, NumpyLike, NumpyMetadata
 from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._nplikes.typetracer import TypeTracer
 from awkward._slicing import normalize_slice
@@ -705,10 +705,10 @@ class Content:
                 )
             )
 
-    def _getitem_at(self, where: SupportsIndex):
+    def _getitem_at(self, where: IndexType):
         raise ak._errors.wrap_error(NotImplementedError)
 
-    def _getitem_range(self, start: SupportsIndex, stop: SupportsIndex) -> Content:
+    def _getitem_range(self, start: SupportsIndex, stop: IndexType) -> Content:
         raise ak._errors.wrap_error(NotImplementedError)
 
     def _getitem_field(
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -10,7 +10,12 @@ import awkward as ak
 from awkward._backends import Backend
 from awkward._nplikes import to_nplike
 from awkward._nplikes.numpy import Numpy
-from awkward._nplikes.numpylike import NumpyLike, NumpyMetadata, ShapeItem
+from awkward._nplikes.numpylike import (
+    NumpyLike,
+    NumpyMetadata,
+    ShapeItem,
+    unknown_length,
+)
 from awkward._nplikes.typetracer import TypeTracer
 from awkward._util import unset
 from awkward.forms.form import Form, JSONMapping, _type_parameters_equal
@@ -370,12 +375,12 @@ class Content:
         length: int,
     ):
         # if this is in a tuple-slice and really should be 0, it will be trimmed later
-        length = 1 if length is not None and length == 0 else length
+        length = 1 if length is not unknown_length and length == 0 else length
         index = Index64(head.index, nplike=self._backend.index_nplike)
         indexlength = index.length
         index = index.to_nplike(self._backend.index_nplike)
         outindex = Index64.empty(
-            self._backend.index_nplike.mul_shape_item(index.length, length),
+            index.length * length,
             self._backend.index_nplike,
         )
 
@@ -567,7 +572,7 @@ class Content:
 
             out = next._getitem_next(nextwhere[0], nextwhere[1:], None)
 
-            if out.length is not None and out.length == 0:
+            if out.length is not unknown_length and out.length == 0:
                 return out._getitem_nothing()
             else:
                 return out._getitem_at(0)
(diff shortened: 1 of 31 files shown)


src/awkward/contents/content.py
```diff
@@ -682,53 +691,35 @@ def _getitem(self, where):
...
-        )
-        if result[0]:
-            if carry.length == self.length:
-                return self
-            elif carry.length < self.length:
-                return self._getitem_range(slice(0, carry.length))
-            else:
-                raise ak._errors.wrap_error(IndexError)
-        else:
-            return None
```
agoose77: @jpivarski I am removing this only because I see no internal uses of it.

src/awkward/contents/recordarray.py
```diff
@@ -362,42 +366,22 @@ def _getitem_range(self, where):
...
-                ]
-            else:
-                contents = [
-                    self.content(i)._getitem_fields(nexthead, nexttail) for i in indexes
-                ]
-
+        contents = [self.content(i) for i in indexes]
         return RecordArray(
             contents, fields, self._length, parameters=None, backend=self._backend
         )
```
agoose77: I removed the `onlyfields` argument because it was easy to, and functionally the behaviour should match. However, I am assuming that this would be a regression. Though I can't see any mention of it in v1, I am somewhat confident that this is likely an optimisation to avoid slicing each field of an array if the fields are ultimately dropped? If so, I'd likely want to revert this change in [65ee2e6](https://github.com/scikit-hep/awkward/pull/2220/commits/65ee2e62d83ffee58540b54cb05fda31c4b81442).

jpivarski: The `only_fields` argument was not in v1; it was added in v2 as a way to avoid code duplication. It's relevant for selecting a field or fields from a call stack that goes through `_getitem_next`, rather than a direct field-selection. (That is, the `only_fields` aregument is relevant for `__getitem__` calls in which a tuple item is a string, possibly after some non-string tuple items.)

That code path hasn't gone away—it can't be dead code—I'd strongly expect some tests to fail if you just remove `only_fields`.

But for all that, I can't remember exactly what it _means_. The name, "only fields" is not ringing a bell. I'd have to examine it more closely to remind myself.

agoose77: The code path is still in use. From my understanding, it's an optimisation that means we lookup additional field slices in order to avoid intermediate slices that need to copy data (mostly indices, I think).

The optimisation is, I think, here: 9ae35faf. 

https://github.com/scikit-hep/awkward/blob/db76e5fcf38560afd4f134908cc32d3ee1811e89/src/awkward/contents/content.py#L308-L318

It rewrites the slice tuple to perform all field slices early. Maybe we should do that anyway as part of our normalisation.

The tests pass (meaning that my understanding is hopefully not incorrect) if one removes the mechanism and instead does something like

```python
nexthead, nexttail = split_head_tail(tail)
record._getitem_fields(head)
```

I'll look into this.

agoose77:
An outstanding challenge with typetracer is the handling of mixed types, e.g. non-typetracer arrays. Whilst it's clear to me that we shouldn't support this, scalars are another matter. Whilst there is strong symmetry with scalars and arrays, there is a notable case where we need to support concrete scalars; indexing. Slicing an array with two known integers should produce a known-length slice, unless the length is not itself known.

Hitherto, scalars converted from shape-items are exclusively supported as 0-D arrays. This PR starts to open that gap; returned scalars can also be Python integers. These aren't that useful with nplikes directly; nplikes are supposed to accept only arraylike types, but internally it will be useful to make this distinction.

agoose77:
The work on `_getitem_range` is bigger than this small PR. I'm going to split out the type hint-only work in order to make this easier.

agoose77:
@lgray a fix for your `RecordArray.content()` `AssertionError` is included here.

jpivarski (approved):
We talked about this on Zoom. As I said, you should sit back and take a broad look at where this is going: is the complexity added by all of these distinct integer types/kinds/sorts (index vs length vs...) justified by the gain of agreement with Array API's having `None` in `shape` tuples? If our codebase would be simpler by diverging from Array API and introducing a specialized "unknown" token instead of using Python's built-in `None`, that's something to consider. But once you do give that a good, hard consideration, if you come to the conclusion that it _is_ justified, then that's okay.

Also, be careful about slices with negative `step`. They're really different, and they need `stop=None` to be distinct from an unknown `stop`.

agoose77:
We didn't have enough time to properly cover this earlier, so let's have that discussion async here?

I'll start by summarising where we are / why we're here.

## Use of `None` to mean `unknown-length`
In #2154 I started removing `UnknownLength` and `UnknownScalar`. This was premature; we _need_ a distinction between  "Unknown length" and "Unknown Array"
  - Don't care about type information for unknown length
  - Don't want recursive "array's size is an array" phenomenon; all known NumPy-like modules return `tuple[int, ...]` for array shapes

We chose `None` because:
- It natively serialises to JSON
- It's a natural singleton
- It is used by the Array API for "unknown length" meaning

None of these are hard constraints, for various reasons.

The current approach in Awkward is to use `nplike` functions to manipulate array shapes. If we didn't use `None`, we could move some of these functions (`add_shape_item`, etc.) into magic methods on the `unknown-length` singleton.


## Disinguishing `index` from `unknown-scalar`
Indices in Awkward can come from two places:
- constants, e.g. `array[1:4]`. These are Python integers, or objects that implement `__index__`.
- arrays, e.g. `array[x.argmin()]`. Kernels might return these as 0-D arrays.

Let's imagine that arrays only ever return 0D arrays for constants.[^not-true]

In typetracer, the 0D-scalar array can _never_ have known content, so we *must* accept some kind of concrete type for constant indices. I assert that we should use Python integers for this. So, that's what this PR does; it introduces the "concept" of an index, which is practically just `int | Arraylike(shape=())`. It is a superset of the array-value indices, which are just 0D arrays.

We need to be more explicit about when we're talking about lengths and when we're talking about indices in my opinion:
- We don't want `_getitem_XXX` to need to handle three different types `(int, nplike-scalar, unknown-length)`. We should only be passing around the index-types of `(int, nplike-scalar)` for slicing etc.
- We don't want to pass in `nplike-scalar` to an nplike shape, e.g. `nplike.zeros(numnull)` where `numnull` is computed by a kernel

By this rationale, I think we want to be converting lengths to indexes eagerly and explicitly. This would be done via `nplike.shape_item_as_index`, and `nplike.index_as_shape_item`. i.e., indexing routines like `_getitem_at` should expect `int` or `nplike-scalar`.

## Conclusion

At this stage, we've discussed two orthogonal things:
1. should `unknown-length` be none-`None`, and implement operators to replace `nplike.add_shape_item(x, y)` with `x + y`?
2. should we be explicit about what is a length and what is an index?

It's my view that we need (2). (1) is less set in stone.

[^not-true]: this is not true for NumPy.

jpivarski:
Regularizing the different input types for indexes, so that our internal (L3) functions each have fewer types to consider is a good thing. I'm on board with doing that.

Requiring L2 users to have to think more about this than they otherwise would have to is a problem. We really want to make any public interfaces "as simple as possible but no simpler." If having to pass indexes through regularization functions is required _because of_ unknown `shape` items being `None`, then maybe we should rethink that decision (below).

What happens if an L2 user does not pass an input index through regularization? Does it always raise an exception, only for certain types, or is it a silent error? At least, if we're going to make L2 users do something, the warning bells should always trigger, so that it's not something they learn about in previously working code that hits a new test case. If they don't read all of the documentation, does the library itself indicate when and how they're doing it wrong?

> We chose `None` because:
> 
> * It natively serialises to JSON
> * It's a natural singleton
> * It is used by the Array API for "unknown length" meaning

None of the above are strong reasons. We could make an "unknown length" token that is distinct from `None` and have it serialize to JSON in Forms, possibly to and from JSON `null`. (`Form.from_json` is context-sensitive enough to do this.)

So that's why I'm expressing caution. You're more aware of the detailed constraints than I am, so you'll know what the full trade-offs would be, but if introducing a non-`None` unknown length token would eliminate the need for a lot of regularization, particularly at L2, then that would be a better thing to do than to coincide with the Array API.

We're not presenting our TypeTracerArrays to the world as a new xarray/cupy/pint/etc. to mix into data analysis (the motivation for Array API). If our TypeTracerArrays, which are very much an instrumental means-to-an-end (even if exposed as L2), differ from the Array API in saying that a shape is

```python
(unknown_length, 5, 7)
```

instead of

```python
(None, 5, 7)
```

it would still serve our purposes and be pretty obvious to an L2 user, likely more obvious than "you need to use these regularization functions before passing it into these other functions").

But finally, this is just a word of caution, not an imperative—if introducing a non-`None` unknown length token wouldn't _actually_ reduce the amount of regularization work that needs to be done, and you conclude that after a hard look at the holistic picture—then go ahead.

jpivarski:
Oh, and I meant to mention (remind? it was on another PR) the slice-object case: `None` already has a meaning as an endpoint that can't be fulfilled by any integer, and that meaning is distinct from "unknown." It's only really necessary for slices with `step < 0`,

```python
>>> s = slice(None, None, -1)
>>> s
slice(None, None, -1)
>>> list(range(10))[s]
[9, 8, 7, 6, 5, 4, 3, 2, 1, 0]
```

But that does happen in the codebase (through `_getitem_next`, not `_getitem_range`).

Our decision to pass around this Python built-in type, `slice`, might have been a mistake. It might have been a better idea to make our own internal `GetItemRange` class, so that such a thing could carry `unknown_length` instead of `None`, or some other way to distinguish "no endpoint" from "unknown endpoint."

agoose77:
The issues discussed above are mostly settled with #2229. The issue of proper slice normalisation is open, and I will fix shortly.

jpivarski (approved):
Yes, `_getitem_range` always takes 2-item slices (no `step`) and that's not going to change, so this is good for being more explicit (including types).

agoose77:
@jpivarski are you happy with the slice normalisation / expansion logic? Now there's a helper in `_slicing` that ensures we have only index values (`int`, `ArrayLike(shape=())`), whilst the nplike itself  handles the equivalent of `slice.indices`.

jpivarski:
I was thinking, "We had slice normalization before..." Well, we had normalization for the things that get passed to `__getitem__`, but now this is additional normalization for actual `slice` objects.

Yes, this looks good. Only one comment, which is not related to the slicing itself: for internal functions (L3 and below), I think it's best to avoid keyword defaults, and then there would be no need for keyword-only arguments. Whereas I think that keywords and keyword-only arguments are great interfaces, non-keyword functions force all the calls to be explicit. (You wouldn't want any functions that use a custom `Backend` to not pass it.)

agoose77:
@jpivarski which function has a keyword default? Off the top of my head, I don't remember adding any for the same reasons you give.

agoose77:
I just want to change `SupportsIndex` usage with `IndexType` now that we have this specced out.

jpivarski:
> @jpivarski which function has a keyword default? Off the top of my head, I don't remember adding any for the same reasons you give.

https://github.com/scikit-hep/awkward/blob/cc97554e297f7dc1a8b508faa23da4e9f610af81/src/awkward/_slicing.py#L26

and

https://github.com/scikit-hep/awkward/blob/cc97554e297f7dc1a8b508faa23da4e9f610af81/src/awkward/_slicing.py#L40

These are L3 because the `_slicing` module is hidden.

agoose77:
The reason for making _these_ keyword-only is for readability. I'm not strongly opinionated - in this case the benefit is more marginal, but for me it's not the interaction with default arguments that's as useful as being able to understand the meaning of arguments without invoking the IDE hinting / autocomplete. 

If you don't like this here, let's revert it.

jpivarski:
Oh! They're keyword-only but don't have defaults. It's the defaults that I considered a problem for internal code.

These are fine, though how much can it improve readability? Expressions that get passed into this argument are probably `self._backend`, so how much does `backend=self._backend` help? (I'm going to take another look at the code...) Yeah, it's all `backend=self._backend`.

agoose77:
Yes, less strong an individual case, more a general habit. This is internal and only used in a few places. Simply for ease, are you happy to merge as is?

PR #2221, opened 2023-02-07
fix: add `is_c_contiguous` method to Cupy

ianna:
- [x] implement an abstract `is_c_contiguous` method for Cupy class
- [x] update cuda tests

### DIFF
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -1395,7 +1395,7 @@ class ListArray(Content):
             and self._backend.nplike.known_data
             and self._starts.length != 0
         ):
-            startsmin = self._backend.index_nplike.min(self._starts.data).item()
+            startsmin = self._backend.index_nplike.min(self._starts.data)
             starts = ak.index.Index(
                 self._starts.data - startsmin, nplike=self._backend.index_nplike
             )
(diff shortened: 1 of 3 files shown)


agoose77:
Huh, this did exist in my local branch at some point 😦 Thanks for catching it!

tests-cuda/test_1276-cupy-interop.py
```diff
@@ -8,8 +8,8 @@
 
 
 def test_cupy_interop():
-    cupy = ak._nplikes.Cupy.instance()
-    numpy = ak._nplikes.Numpy.instance()
+    cupy = ak._nplikes.cupy.Cupy.instance()
+    numpy = ak._nplikes.numpy.Numpy.instance()
```
agoose77: In order to avoid breaking import cycles, the `_nplikes` package doesn't import `cupy` or `numpy`. Are you happy to import these at the top level, e.g. `from awkward._nplikes.cupy import Cupy`?

ianna: Thanks for suggestion!

src/awkward/contents/listarray.py
```diff
@@ -1395,15 +1395,15 @@ def _recursively_apply(
             and self._backend.nplike.known_data
             and self._starts.length != 0
         ):
-            startsmin = self._backend.index_nplike.min(self._starts.data).item()
+            startsmin = self._backend.index_nplike.min(self._starts.data)
```
agoose77, line 1398: I assume we don't need `.item()` here because of the subsequent `self._starts.data - startsmin`, which CuPy understands as a `1D` array and a `0D` array being subtracted from one another

agoose77 (approved):

PR #2223, opened 2023-02-07
refactor: add type hints for `Content._getitem_XXX`

agoose77:
## TL;DR
This PR
- Removes `Content._carry_asrange` and its kernel `awkward_Index_iscontinguous`, as it appears unused.
- Adds type annotations to the `_getitem_` system


jpivarski (approved):
Type hints! It all looks good.

Also, I see that the kernel associated with the dead code from a previous PR is removed. :checkmark:

PR #2225, opened 2023-02-07
feat: ByteMaskedArray numnull Cuda kernel

ianna:
- [x] implement `awkward_ByteMaskedArray_numnull` Cuda kernel
- [x] enable tests

### DIFF
--- a/src/awkward/_connect/cuda/cuda_kernels/awkward_ByteMaskedArray_numnull.cu
+++ b/src/awkward/_connect/cuda/cuda_kernels/awkward_ByteMaskedArray_numnull.cu
@@ -1,0 +1,19 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+__global__ void
+awkward_ByteMaskedArray_numnull(int64_t* numnull,
+  				const int8_t* mask,
+  				int64_t length,
+  				bool validwhen,
+                            	uint64_t invocation_index,
+                            	uint64_t* err_code) {
+  *numnull = 0;
+  if (err_code[0] == NO_ERROR) {
+    int64_t thread_id = blockIdx.x * blockDim.x + threadIdx.x;
+    if (thread_id < length) {
+      if ((mask[thread_id] != 0) != validwhen) {
+      	*numnull = *numnull + 1;
+      }
+    }
+  }
+}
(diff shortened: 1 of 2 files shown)


src/awkward/_connect/cuda/cuda_kernels/awkward_ByteMaskedArray_numnull.cu
```diff
@@ -0,0 +1,19 @@
...
+  				int64_t length,
+  				bool validwhen,
+                            	uint64_t invocation_index,
+                            	uint64_t* err_code) {
+  *numnull = 0;
+  if (err_code[0] == NO_ERROR) {
+    int64_t thread_id = blockIdx.x * blockDim.x + threadIdx.x;
+    if (thread_id < length) {
+      if ((mask[thread_id] != 0) != validwhen) {
+      	*numnull = *numnull + 1;
```
jpivarski, line 15: I doubt this is atomic. I think it would give the wrong output if the array were large enough (larger than certain memory structures within the GPU).

jpivarski (commented):
This is exactly the kind of update that @kkothari2001 would be doing for all of the kernels in the proposed GPU project. It's fine for the rest of us to fill them in as we go—it's not like we're going to run out of kernels at this rate—but I think this implementation isn't right for concurrency reasons.

This kernel counts missing values in an array's mask (instances of `0 <= thread_id < length` in which `(mask[thread_id] != 0) != validwhen`). I've highlighted the line in which

```c++
*numnull = *numnull + 1;
```

That operation might not be atomic, as-is. Threads in a GPU see a highly NUMA-like view of the memory, so they might update different copies of `*numnull` and one clobers the others when they write back to RAM memory.

Maybe what needs to be done here is to use a CUDA atomic counter. (They definitely have integer addition as an atomic operation.) But maybe, for large-array performance, filling thread-local ("block"-local?) copies and tree-reducing them is the right way to go. Maybe that's what the CUDA atomic counter does and it would be worse for us to implement it ourselves.

These are the kinds of open questions that we have about CUDA programming. @kkothari2001 has a lot of time before his project would start (and we have to remember that it's still in the planning stages), but he can start learning about these things in the context of specific cases like this.

So I'm going to leave this PR open and call it a learning exercise. It's a good place to have a back-and-forth conversation as we're all learning about CUDA. It's nice that this PR only implements one kernel, so that we have something to focus on.

When the issue is resolved, we'll merge it, of course. But there isn't strong time pressure. (The timescale is something like September 2023.)

ianna:
@kkothari2001 - please, feel free to take over this :-)

ianna (commented):
@jpivarski - I think, it's ok to close this PR and delete the branch.

PR #2226, opened 2023-02-07
fix: use of size in `to_RegularArray`

agoose77:
Anywhere that we compute lengths in kernels, we should be casting the results to lengths using `index_nplike.scalar_as_shape_item(scalar)`, This allows conversion from unknown typetracer scalars to `None` lengths.

### DIFF
--- a/src/awkward/contents/listoffsetarray.py
+++ b/src/awkward/contents/listoffsetarray.py
@@ -207,25 +207,26 @@ class ListOffsetArray(Content):
     def to_RegularArray(self):
         start, stop = self._offsets[0], self._offsets[self._offsets.length - 1]
         content = self._content._getitem_range(slice(start, stop))
-        size = ak.index.Index64.empty(1, self._backend.index_nplike)
+        _size = ak.index.Index64.empty(1, self._backend.index_nplike)
         assert (
-            size.nplike is self._backend.index_nplike
+            _size.nplike is self._backend.index_nplike
             and self._offsets.nplike is self._backend.index_nplike
         )
         self._handle_error(
             self._backend[
                 "awkward_ListOffsetArray_toRegularArray",
-                size.dtype.type,
+                _size.dtype.type,
                 self._offsets.dtype.type,
             ](
-                size.data,
+                _size.data,
                 self._offsets.data,
                 self._offsets.length,
             )
         )
-
+        size = self._backend.index_nplike.scalar_as_shape_item(_size[0])
+        length = self._backend.index_nplike.sub_shape_item(self._offsets.length, 1)
         return ak.contents.RegularArray(
-            content, size[0], self._offsets.length - 1, parameters=self._parameters
+            content, size, length, parameters=self._parameters
         )
 
     def _getitem_nothing(self):
--- a/src/awkward/contents/regulararray.py
+++ b/src/awkward/contents/regulararray.py
@@ -34,7 +34,16 @@ class RegularArray(Content):
                     )
                 )
             )
-        if size is not None:
+        if size is None:
+            if content.backend.index_nplike.known_shape:
+                raise ak._errors.wrap_error(
+                    TypeError(
+                        "{} 'size' must be a non-negative integer for backends with known shapes, not None".format(
+                            type(self).__name__
+                        )
+                    )
+                )
+        else:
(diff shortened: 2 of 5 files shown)


src/awkward/contents/listoffsetarray.py
```diff
@@ -207,25 +207,26 @@ def to_ListOffsetArray64(self, start_at_zero=False):
...
         )
         self._handle_error(
             self._backend[
                 "awkward_ListOffsetArray_toRegularArray",
-                size.dtype.type,
+                _size.dtype.type,
                 self._offsets.dtype.type,
             ](
-                size.data,
+                _size.data,
```
jpivarski, line 221: It's not very common to prepend _local_ variables hidden in a function definition with underscores. Is that just to ensure that this variable is distinct from the one on (green) line 226, also named `size`?

src/awkward/contents/listoffsetarray.py
```diff
@@ -207,25 +207,26 @@ def to_ListOffsetArray64(self, start_at_zero=False):
...
                 self._offsets.data,
                 self._offsets.length,
             )
         )
-
+        size = self._backend.index_nplike.scalar_as_shape_item(_size[0])
+        length = self._backend.index_nplike.sub_shape_item(self._offsets.length, 1)
         return ak.contents.RegularArray(
-            content, size[0], self._offsets.length - 1, parameters=self._parameters
+            content, size, length, parameters=self._parameters
```
jpivarski, line 229: The use of a length-1 Index made it possible to get a size value out of a kernel. But a kernel is statically typed: it can't return a size of `None`, ever. How does this allow the RegularArray's `size` to be `None` when it's a typetracer? Doesn't this need a specialized path for typetracers?

src/awkward/contents/listoffsetarray.py
agoose77: Yes. Now that lengths and offsets are fundamentally different types, we want to be very careful not to mix them unknowingly. This is becoming a pattern; wherever a kernel computes a length, we want the ability to realise the unknown-ness in the standard `None | int` type space.

src/awkward/contents/listoffsetarray.py
agoose77: There's a discussion to be had about how explicit typetracer integration needs to be, agreed.

In this case, this works because
- The `_size` index is a typetracer array, therefore it has no concrete values
- kernels are no-ops for typetracer
- `scalar_as_shape_item` interprets unknown-scalars as `None`

jpivarski: Aha! Okay, that makes sense.

Even though I didn't get it, and that's a sign that this is non-obvious code, I think this is a fine way to solve it. It'll become a common pattern that we'll recognize.

jpivarski (approved):

src/awkward/contents/listoffsetarray.py
agoose77: OK, fab. I'm a little torn on how explicit to be with kernels. It feels like our existing approach of treating them as no-ops is fine because we only ever touch data in kernels, rather than allocating. There is merit to just if-elseing the entire thing in terms of explicitness, but the downside is that we have to duplicate logic. As you can probably tell, I'm currently leaning towards treating typetracer just like any other nplike.

jpivarski: > There is merit to just if-elseing the entire thing in terms of explicitness, but the downside is that we have to duplicate logic.

That explicitness would be everywhere—it would force us to be thinking about typetracers even in situations where we don't have to. The existing code has to think about five or six different things, everywhere, and if we can keep it usually-implicit without bad consequences, that will make it easier to think about.

> As you can probably tell, I'm currently leaning towards treating typetracer just like any other nplike.

As much as possible and no farther.

agoose77: We're on the same page, I think. In my mind, `known_shape` / `known_data` are mainly escape hatches e.g. if some logic is more easily expressed with concrete knowledge that we don't have a typetracer primitive. The main harm with typetracer is the `add_shape_item` usage, but I think that's not terrible as at least it's explicit in its verbosity.

> As much as possible and no farther.

I think this is pretty far; the shape handling abstractions deal with much of the problems w.r.t lengths.

PR #2227, opened 2023-02-08
fix: indexing into `RegularArray` with typetracer

### LINKED ISSUE #2222: Indexing operations on regular array typetracer fails
### Version of Awkward Array

awkward@main + dask_awkward@refactor-column-optimization

```python3
import awkward as ak
import dask_awkward as dak

x = ak.to_regular(ak.Array([[[-1, 1.], [-2, 2], [-3, 3]], [[-4, 4]], []]), axis=-1)
dx = dak.from_awkward(x, 1)
dx._meta[:, -1]
```

results in:
```
  File "regular_array_problem.py", line 6, in <module>
    dx._meta[:, -1]
  File "/Users/lgray/coffea-dev/awkward/src/awkward/highlevel.py", line 951, in __getitem__
    out = self._layout[where]
  File "/Users/lgray/coffea-dev/awkward/src/awkward/contents/content.py", line 514, in __getitem__
    return self._getitem(where)
  File "/Users/lgray/coffea-dev/awkward/src/awkward/contents/content.py", line 551, in _getitem
  File "/Users/lgray/coffea-dev/awkward/src/awkward/contents/regulararray.py", line 393, in _getitem_next
    start, stop, step = head.indices(self._size)
```

agoose77:
## TL;DR
This PR 
- fixes #2222 by handling unknown lengths in `RegularArray` for `_getitem_at`. 
- introduces helper to regularize an index given an array length

## DR
This latter point (regularisation of indices) is infectious; if a length is not known, or the index is unknown, the result is unknown.

This PR is taking a step towards allowing `int` as indices alongside the scalar arrays belong to `backend.index_nplike`. We need this if we want concrete slices (`array[2:4]`) to preserve more information than unknown slices (`array[x:y]`).

### DIFF
--- a/src/awkward/contents/regulararray.py
+++ b/src/awkward/contents/regulararray.py
@@ -222,14 +222,11 @@ class RegularArray(Content):
 
     def _getitem_at(self, where: IndexType):
         index_nplike = self._backend.index_nplike
-        if index_nplike.known_data and where < 0:
-            where += self._length
-
-        if not (self._length is unknown_length or 0 <= where < self._length):
-            raise ak._errors.index_error(self, where)
-        start, stop = where * index_nplike.shape_item_as_index(self._size), (
+        where = ak._slicing.regularize_index(where, self._length, backend=self._backend)
+        size_scalar = index_nplike.shape_item_as_index(self._size)
+        start, stop = where * size_scalar, (
             where + 1
-        ) * index_nplike.shape_item_as_index(self._size)
+        ) * size_scalar
         return self._content._getitem_range(start, stop)
 
     def _getitem_range(self, start: SupportsIndex, stop: IndexType) -> Content:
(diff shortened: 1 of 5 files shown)


agoose77:
@lgray would you mind confirming that this fixes your workflow?

lgray:
Yes this fixes the repro and the more complex actual use case. Thanks!

src/awkward/contents/regulararray.py
```diff
@@ -420,19 +417,25 @@ def _getitem_next(
...
-            elif step < 0 and stop - start < 0:
-                diff = start - stop
-                nextsize = diff // (step * -1)
-                if diff % step != 0:
-                    nextsize += 1
+            # Handle unknown sizes
+            if self._size is None:
+                nextsize = None
+                start = 0
+                step = 1 if head.step is None else head.step
```
jpivarski: If `step` is negative, should `start` be `0`?

Slices are completely different when the `step` is negative; be sure there is a test case for it.

jpivarski (approved):
More loosening of assumptions about lengths is a good thing; just be careful about slices with negative `step`, as noted below.

jpivarski (approved):

PR #2228, opened 2023-02-08
refactor: drop `NumpyLike.known_shape`

agoose77:
We've discussed this before; there is a distinction to be made between known shapes and known lengths, but there is currently no use for any variant besides `known_shape = known_data = True/False`. Given that we have no way to test the isolation of this logic, it's better that we remove it. 

Note that `known_shape` just means "this shape cannot contain unknown components`. `known_data`, meanwhile, used to mean "these data are totally unknown". 

Now `known_data` means "this shape might contain unknown components, and the data aren't known".

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1193,7 +1193,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
         typestr = repr(str(self.type))[1:-1]
         if (
-            self._layout.backend.nplike.known_shape
+            self._layout.backend.nplike.known_data
             and self._layout.backend.nplike.known_data
         ):
             valuestr = ""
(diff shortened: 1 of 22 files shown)


src/awkward/highlevel.py
```diff
@@ -1193,7 +1193,7 @@ def _repr(self, limit_cols):
 
         typestr = repr(str(self.type))[1:-1]
         if (
-            self._layout.backend.nplike.known_shape
+            self._layout.backend.nplike.known_data
             and self._layout.backend.nplike.known_data
```
jpivarski: For example.

jpivarski (approved):
Thanks—I'm glad you're doing this: it was a distinction without a difference. Replacing `known_shape` → `known_data` is a good start, but there are some instances of checking for `known_data and/or known_data`.

agoose77:
@jpivarski haha, you're quicker than I can grep.

PR #2229, opened 2023-02-09
feat!: re-introduce `unknown-length`

agoose77:
## TL;DR
- Replace use of `None` as "unknown length" token with `unknown_length` singleton.
- Replace use of `nplike.add_shape_item(length, x)` with `length + x`

## TL
The rationale for using `None` as a token for an unknown length was weak, but was ultimately due to the following factors:
- Array API uses `None` for unknown dimension lengths
- `None` easily serialises to JSON
- `None` is a natural singleton

Some of this was discussed in https://github.com/scikit-hep/awkward/pull/2220#issuecomment-1422902819

However, a significant problem with using `None` is that it is also a meaningful index value, and we would be unable to disambiguate between `array[None]` and `array[array.layout.length]` for typetracer arrays. Whilst this would only affect L2 typetracer users, it's enough to change the balance.

Thereafter, there are arguments to be made in favour of using `nplike.XXX_shape_item` functions to operate upon lengths vs a rich object that implements magic methods. I preferred the former, due to explicitness and the ability to do e.g. validation on known lengths (such as ensuring that they are positive). However, code is less easy to read from an expression perspective. @jpivarski and I discussed this, and we settled on re-introducing the object-oriented API. 

We will maintain a strong distinction between an `index: int | ArrayLike` and a `length: int | type[unknown_length]`, to ensure that unknown values map correctly between spaces.

This reverts the inability to set `None` as length values in `RecordArray`.

### DIFF
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -7,6 +7,7 @@ from collections.abc import Collection, Mapping
 import awkward as ak
 from awkward import _errors
 from awkward._nplikes.numpylike import NumpyMetadata
+from awkward._nplikes.shape import unknown_length
 from awkward.typing import Final, TypeAlias
 
 np = NumpyMetadata.instance()
@@ -31,9 +32,7 @@ reserved_nominal_parameters: Final = frozenset(
 
 
 def from_dict(input: dict) -> Form:
-    if input is None:
-        return None
-
+    assert input is not None
     if isinstance(input, str):
         return ak.forms.NumpyForm(primitive=input)
 
@@ -54,7 +53,7 @@ def from_dict(input: dict) -> Form:
     elif input["class"] == "RegularArray":
         return ak.forms.RegularForm(
             content=from_dict(input["content"]),
-            size=input["size"],
+            size=unknown_length if input["size"] is None else input["size"],
             parameters=parameters,
             form_key=form_key,
         )
--- a/src/awkward/forms/regularform.py
+++ b/src/awkward/forms/regularform.py
@@ -1,7 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
 import awkward as ak
-from awkward._nplikes.numpylike import unknown_length
+from awkward._nplikes.shape import unknown_length
 from awkward._util import unset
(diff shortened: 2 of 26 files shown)


src/awkward/forms/regularform.py
```diff
@@ -61,7 +62,7 @@ def _to_dict_part(self, verbose, toplevel):
         return self._to_dict_extra(
             {
                 "class": "RegularArray",
-                "size": self._size,
+                "size": None if self._size is unknown_length else self._size,
```
agoose77, line 65: Here we encode `unknown_length` as `None`

src/awkward/forms/form.py
```diff
@@ -54,7 +53,7 @@ def from_dict(input: dict) -> Form:
     elif input["class"] == "RegularArray":
         return ak.forms.RegularForm(
             content=from_dict(input["content"]),
-            size=input["size"],
+            size=unknown_length if input["size"] is None else input["size"],
```
agoose77, line 56: Here we decode `None` to mean `unknown_length`

jpivarski: BitMaskedArray has length metadata that can now be `unknown_length`, but I guess it's not a part of the BitMaskedForm the way that `size` is for RegularForm. I had thought that there might be another place to do this, but apparently not.

tests/test_0111_jagged_and_masked_getitem.py
```diff
@@ -485,20 +486,11 @@ def test_bool_missing():
...
-                        mask = [x1, x2, x3, x4, x5]
-                        expected = [
-                            m if m is None else x
-                            for x, m in zip(data, mask)
-                            if m is not False
-                        ]
-                        array2 = ak.highlevel.Array(mask, check_valid=True).layout
-                        assert to_list(array[array2]) == expected
-                        assert array.to_typetracer()[array2].form == array[array2].form
+    for mask in itertools.repeat([True, False, None], 5):
```
jpivarski, line 489: `:)`

jpivarski (approved):
This is great—mostly replacements of `None` → `unknown_length`, and it's usually easier to go the other way (`None` appears in other contexts than unknown lengths). It looks good to me, including the serialization to and from JSON.

PR #2231, opened 2023-02-09
docs: reintroduce Content documentation from v1 reST files.

### LINKED ISSUE #2230: Documentation on Contents
### Which documentation?

Documentation website

I think it would be quite useful if the structure of each content class was documented somewhere.

I keep forgetting the specifics of how each content class works. To figure this out, I have been instantiating the class, then calling `to_buffers`. It would be great if the docs for either the Forms or Conents classes would say things like:

> ListOffsets consist of an integer valued offset array and a data array. The data array is any other ak.contents.Content subclass. The buffer array has length (len(array) + 1) and ...

This would be very helpful for low level construction and manipulation of awkward arrays, and also for potential contributors the library.

jpivarski:
Specifically,

  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.Content.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.BitMaskedArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.ByteMaskedArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.EmptyArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.IndexedArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.IndexedOptionArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.ListArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.ListOffsetArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.NumpyArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.RecordArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.RegularArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.UnionArray.rst
  * https://raw.githubusercontent.com/scikit-hep/awkward/main-v1/docs-sphinx/ak.layout.UnmaskedArray.rst

### DIFF
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -74,6 +74,8 @@ class BitMaskedArray(Content):
 
             def __getitem__(self, where):
                 if isinstance(where, int):
+                    if where < 0:
+                        where += len(self)
                     assert 0 <= where < len(self)
                     if self.lsb_order:
                         bit = bool(self.mask[where // 8] & (1 << (where % 8)))
@@ -83,6 +85,7 @@ class BitMaskedArray(Content):
                         return self.content[where]
                     else:
                         return None
+
                 elif isinstance(where, slice) and where.step is None:
                     # In general, slices must convert BitMaskedArray to ByteMaskedArray.
                     bytemask = np.unpackbits(
@@ -93,6 +96,7 @@ class BitMaskedArray(Content):
                         self.content[where.start : where.stop],
                         valid_when=self.valid_when,
                     )
+
                 elif isinstance(where, str):
                     return BitMaskedArray(
                         self.mask,
@@ -101,6 +105,7 @@ class BitMaskedArray(Content):
                         length=self.length,
                         lsb_order=self.lsb_order,
                     )
+
                 else:
                     raise AssertionError(where)
     """
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -27,6 +27,79 @@ numpy = Numpy.instance()
 
 @final
 class BitMaskedArray(Content):
+    """
+    Like #ak.contents.ByteMaskedArray, BitMaskedArray implements an
+    #ak.types.OptionType with two buffers, `mask` and `content`.
+    However, the boolean `mask` values are packed into a bitmap.
+
+    BitMaskedArray has an additional parameter, `lsb_order`; if True,
+    the position of each bit is in
+    [Least-Significant Bit order](https://en.wikipedia.org/wiki/Bit_numbering)
+    (LSB):
+
+        is_valid[j] = bool(mask[j // 8] & (1 << (j % 8))) == valid_when
+
+    If False, the position of each bit is in Most-Significant Bit order
+    (MSB):
+
+        is_valid[j] = bool(mask[j // 8] & (128 >> (j % 8))) == valid_when
+
+    If the logical size of the buffer is not a multiple of 8, the `mask`
+    has to be padded. Thus, an explicit `length` is also part of the
+    class's definition.
+
+    This is equivalent to *all* of Apache Arrow's array types because they all
+    [use bitmaps](https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps)
+    to mask their data, with `valid_when=True` and `lsb_order=True`.
+
+    To illustrate how the constructor arguments are interpreted, the following is a
+    simplified implementation of `__init__`, `__len__`, and `__getitem__`:
+
+        class BitMaskedArray(Content):
+            def __init__(self, mask, content, valid_when, length, lsb_order):
+                assert isinstance(mask, IndexU8)
+                assert isinstance(content, Content)
+                assert isinstance(valid_when, bool)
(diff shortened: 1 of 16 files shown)


src/awkward/contents/bitmaskedarray.py
```diff
@@ -27,6 +27,84 @@
...
+                elif isinstance(where, str):
+                    return BitMaskedArray(
+                        self.mask,
+                        self.content[where],
+                        valid_when=self.valid_when,
+                        length=self.length,
+                        lsb_order=self.lsb_order,
+                    )
+                else:
+                    raise AssertionError(where)
```
agoose77: When looking through the v1 docs, I (personally) didn't find these code excerpts that helpful; they're not an exact copy of the source, after all. I'd be tempted to remove them in favour of the prose that we already have (perhaps addition descriptions). Do you think these are more help than hindrance, i.e. we should keep them?

src/awkward/contents/content.py
```diff
@@ -962,10 +984,23 @@ def is_identity_like(self) -> bool:
 
     @property
     def purelist_isregular(self) -> bool:
+        """
+        Returns True if all dimensions down to the first record or tuple layer have
+        #ak.types.RegularType; False otherwise.
```
agoose77, line 994: It's good that we're talking about user-visible type objects here. 🦸

src/awkward/contents/emptyarray.py
```diff
@@ -22,6 +22,34 @@
 
 @final
 class EmptyArray(Content):
+    """
+    An EmptyArray is used whenever an array's type is not known because it is empty
+    (such as data from #ak.ArrayBuilder without enough sample points to resolve the
+    type).
+
+    EmptyArray has no equivalent in Apache Arrow.
```
agoose77, line 33:
```suggestion
    EmptyArray has no equivalent in Apache Arrow. It also cannot contain any parameters.
```

src/awkward/contents/listoffsetarray.py
```diff
@@ -24,6 +24,64 @@
...
+    "jagged" or "ragged" array). Like #ak.contents.RegularArray, the
+    underlying data for all lists are in a contiguous `content`. It is
+    subdivided into lists according to an `offsets` buffer, which specifies
+    the starting and stopping index of each list.
+
+    The `offsets` must have at least length 1 (corresponding to an empty array),
+    but it need not start with `0` or include all of the `content`. Just as
+    #ak.contents.RegularArray can have unreachable `content` if it is not
+    an integer multiple of `size`, a ListOffsetArray can have unreachable
+    content before the first list and after the last list.
```
agoose77:
```suggestion
    content before the start of the first list and after the end of the last list.
```

src/awkward/contents/recordarray.py
```diff
@@ -26,6 +26,86 @@
 
 @final
 class RecordArray(Content):
+    """
+    RecordArray represents an array of tuples or records, all with the
+    same type. Its `contents` is an ordered list of arrays.
```
agoose77, line 31: What do we mean by "all with the same type" here?

agoose77 (approved):

src/awkward/contents/recordarray.py
jpivarski: Every record in a RecordArray has the same type as each other.

agoose77:
Thanks for copying this over Jim. It looks exactly like what was asked for in the motivating issue :)

src/awkward/contents/bitmaskedarray.py
jpivarski: I was on the fence. At first, I wasn't going to include them, but then concluded that they're useful because the constraints on the constructor arguments are not documented elsewhere (types help with that, but there are also value constraints), and the way getitem computes an item is not documented elsewhere (which is the main thing).

At minimum, saying what `__getitem__(self, where: int)` returns is a large part of saying what a given layout node _is_.

In these listings, I removed the less relevant parts.

Maybe we could ask @ivirshup if they're useful?

agoose77: Hmm, it did feel like these are now smaller. If we removed the code, we would want to verbally explain some of these constraints. 

I'm going to be pragmatic: we don't have the time and energy to split these kind of hairs. Code is good for expressing constraints, and you've slimmed these down to pretty much just that. 

Unless @ivirshup feels strongly, I support your reasoning to keep them.

jpivarski: Okay, I'll set this to auto-merge, and we can always remove them if prompted. (And then we'd have to find another way to express the constraints and `__getitem__`.)

PR #2233, opened 2023-02-13
refactor: growable buffer nullptr check

ianna:


ianna (commented):
@jpivarski - it's ok to delete the branch, I think.

PR #2234, opened 2023-02-13
feat: keep column order after filtering flag

ianna:


src/awkward/_connect/rdataframe/from_rdataframe.py
```diff
@@ -246,7 +246,16 @@ def form_dtype(form):
         else:
             contents[key] = value
 
+    out = ak.zip(contents, depth_limit=1)
+
+    if keep_order:
+        sorted = ak.index.Index64(contents["rdfentry_"].data.argsort())
```
ianna, line 252: I don't think we need to specify a sorting algorithm here because the `rdfentry_` values are always unique.

ianna (commented):
@jpivarski - please, have a look. I think, I'm done with this PR. Thanks!

src/awkward/_connect/rdataframe/from_rdataframe.py
jpivarski: Right, that makes sense.

jpivarski (approved):
Great! I see that `out` is the result of an `ak.zip` with `depth_limit=1`, so it's a `RecordArray`. Reordering this with `IndexedArray` is what we typically do.

I just noticed that `from_rdataframe` doesn't have `highlevel` and `behavior` arguments, though almost all of the `ak.*` operations do. As it is, this assumes `highlevel=True, behavior=None`. As possible, low-priority future work (independently of this PR), it would be nice to add these arguments.

jpivarski:
Codecov didn't run, but I'm going to ignore that and merge it anyway.

Issue #1668, opened 2022-09-05
transform, but not caring about broadcasting regular dimensions


ivirshup:
## The request

I would like to do recursive iteration over multiple arrays without having to have aligned regular dimensions (especially the first).

## An example

I would like to take multiple arrays, and subset their records to a common set of keys. Here's a quick example of how I would expect to do this:

```python
from operator import and_
from functools import reduce

import awkward._v2 as ak

def intersect(layouts: list[ak.contents.Content], **kwargs):
    if all(l.is_RecordType for l in layouts):
        common_keys = list(reduce(and_, (set(l.fields) for l in layouts)))
        return [l[common_keys] for l in layouts]
```

This works since the arrays are of similar length:

```python
awk_a = ak.Array([
    {"a": [1, 2, 3], "b": [1, 2]},
    {"a": [4, 5], "b": [3, 4]},
    {"a": [6], "b": [5]},
])

awk_b = ak.Array([
    {"a": [1, 2, 3]}, 
    {"a": [4, 5]},
    {"a": [6]},
])

# works with these arrays
ak.transform(intersect, awk_a, awk_b)
# (<Array [{a: [1, 2, 3]}, {...}, {a: [6]}] type='3 * {a: var * int64}'>,
#  <Array [{a: [1, 2, 3]}, {...}, {a: [6]}] type='3 * {a: var * int64}'>)
```

This does not work if the first dimension for the arrays is not the same size.

```python
awk_c = ak.Array([
    {"a": [1, 2, 3], "b": [1, 2]},
    {"a": [4, 5], "b": [3, 4]},
    {"a": [6], "b": [5]},
])

awk_d = ak.Array([
    {"a": [4, 5]},
    {"a": [6]},
])

ak.transform(intersect, awk_c, awk_d)
```

```pytb
ValueError: while calling (from /var/folders/bd/43q20k0n6z15tdfzxvd22r7c0000gn/T/ipykernel_76685/4019137656.py, line 13)
...
Error details: cannot broadcast RegularArray of size 2 with RegularArray of size 3 in ak._v2.transform
```

jpivarski:
[...]

PR #2235, opened 2023-02-13
feat: add `ak.merge_union_of_lists`

### LINKED ISSUE #2211: `ak.merge_union_of_records` not merging `Union[var * {a: int, b:int}, var * {a: int}]`
### Version of Awkward Array

2.0.7

I've found what I think is a bug in the recently added `merge_union_of_records`. With the following input it seems to be a no-op, when I would have expected it to result in `5 * var * {a: int: b: ?int}`

```python
>>> a = ak.Array([[{"a": 1, "b": 4}], [{"a": 2, "b": 5}, {"a": 3, "b": 6}]])
>>> b = ak.Array(
...     [
...         [{"a": 4}, {"a": 5}],
...         [{"a": 6}],
...         [{"a": 7}],
...     ]
... )
>>> a.type.show()
2 * var * {
    a: int64,
    b: int64
}
>>> b.type.show()
3 * var * {
    a: int64
}
>>> ak.concatenate([a, b]).type.show()
5 * union[
    var * {
        a: int64,
        b: int64
    },
    var * {
        a: int64
    }
]
>>> ak.merge_union_of_records(ak.concatenate([a, b])).type.show()
5 * union[
    var * {
        a: int64,
        b: int64
    },
    var * {
        a: int64
    }
]
```

agoose77:
This PR adds a new `ak.merge_union_of_lists`, which pushes unions inside of list types, i.e.
```
Union[List[X], List[Y]] → List[Union[X, Y]]
```

This is useful in combination with `ak.merge_union_of_records`, which cannot operate upon `Union[var * Record, ...]`.

Fixes #2211

- [ ] Support `is_option` types in `ak.merge_union_of_lists`
- [ ] Support `is_option` types in `ak.merge_union_of_options`

### DIFF
--- a/src/awkward/operations/ak_merge_union_of_lists.py
+++ b/src/awkward/operations/ak_merge_union_of_lists.py
@@ -1,0 +1,99 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+
+import awkward as ak
+from awkward._nplikes.numpylike import NumpyMetadata
+
+np = NumpyMetadata.instance()
+cpu = ak._backends.NumpyBackend.instance()
+
+
+def merge_union_of_lists(array, axis=-1, *, highlevel=True, behavior=None):
+    """
+    Args:
+        array: Array-like data (anything #ak.to_layout recognizes).
+        axis (int): The dimension at which this operation is applied.
+            The outermost dimension is `0`, followed by `1`, etc., and negative
+            values count backward from the  innermost: `-1` is the innermost
+            dimension, `-2` is the next level up, etc.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+
+    Simplifies unions of lists, e.g.
+
+        >>> array = ak.concatenate(([["a", "b"]], [[1, 2, 3]]))
+
+    into lists of unions, i.e.
+
+        >>> ak.merge_union_of_lists(array, axis=0)
+        <Array [['a', 'b'], [1, 2, 3]] type='2 * var * union[string, int64]'>
+    """
+    with ak._errors.OperationErrorContext(
+        "ak.merge_union_of_lists",
+        {"array": array, "axis": axis, "highlevel": highlevel, "behavior": behavior},
+    ):
+        return _impl(array, axis, highlevel, behavior)
+
+
+def _impl(array, axis, highlevel, behavior):
+    behavior = ak._util.behavior_of(array, behavior=behavior)
+    layout = ak.to_layout(array, allow_record=False)
+
+    def apply(layout, depth, backend, **kwargs):
+        posaxis = ak._util.maybe_posaxis(layout, axis, depth)
+        if depth < posaxis + 1 and layout.is_leaf:
+            raise ak._errors.wrap_error(
+                np.AxisError(f"axis={axis} exceeds the depth of this array ({depth})")
+            )
+        elif depth == posaxis + 1 and layout.is_union:
+            if all(x.is_list for x in layout.contents):
+                if not all(x.is_list for x in layout.contents):
+                    raise ak._errors.wrap_error(NotImplementedError)
+
+                contents = [
+                    x.to_ListOffsetArray64(start_at_zero=True) for x in layout.contents
+                ]
+                index_nplike = backend.index_nplike
+
+                # Compute new offsets of outermost list by taking union of sublist lengths!
+                nums = index_nplike.asarray(
+                    ak.contents.UnionArray.simplified(
+                        layout.tags,
+                        layout.index,
+                        [
+                            ak.contents.NumpyArray(x.stops.data - x.starts.data)
+                            for x in contents
+                        ],
+                    ),
+                    dtype=np.int64,
+                )
+                stops = index_nplike.cumsum(nums)
+                offsets = index_nplike.empty(stops.size + 1, dtype=stops.dtype)
+                offsets[0] = 0
+                offsets[1:] = stops
+
+                # The new tags are given by the original tags, repeated by the sublist counts
+                new_tags = index_nplike.repeat(index_nplike.asarray(layout.tags), nums)
+
+                # To compute the per-tag index, we first project each content, and then flatten it
+                new_contents = [
+                    ak._do.flatten(layout.project(tag), axis=1)
+                    for tag, _ in enumerate(contents)
+                ]
+                # Now loop over each flattened array, and build a simple range index
+                new_index = index_nplike.empty(new_tags.size, dtype=np.int64)
+                for tag, flattened in enumerate(new_contents):
+                    # Index into flattened content, and
+                    index = index_nplike.arange(flattened.length, dtype=np.int64)
+                    new_index[new_tags == tag] = index
+
+                # Now we have a union over these contents
+                content = ak.contents.UnionArray.simplified(
+                    ak.index.Index8(new_tags), ak.index.Index64(new_index), new_contents
+                )
+                return ak.contents.ListOffsetArray(ak.index.Index64(offsets), content)
+
+    out = ak._do.recursively_apply(layout, apply)
+    return ak._util.wrap(out, highlevel=highlevel, behavior=behavior)
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -263,10 +263,6 @@ def fill_none(layout: Content, value: Content) -> Content:
     return layout._fill_none(value)
 
 
-def num(layout, axis):
-    return layout._num(axis, 0)
-
-
 def mergeable(one: Content, two: Content, mergebool: bool = True) -> bool:
     return one._mergeable_next(two, mergebool=mergebool)
 
--- a/src/awkward/operations/ak_concatenate.py
+++ b/src/awkward/operations/ak_concatenate.py
@@ -238,6 +238,8 @@ def _impl(arrays, axis, mergebool, highlevel, behavior):
                 offsets = backend.index_nplike.empty(
                     len(nextinputs[0]) + 1, dtype=np.int64
                 )
+                # TODO `maybe_out` is supposed to be allowed to make a copy
+                # but we don't anticipate that here.
                 offsets[0] = 0
                 backend.index_nplike.cumsum(counts, maybe_out=offsets[1:])
 


src/awkward/operations/ak_merge_union_of_lists.py
```diff
@@ -0,0 +1,99 @@
...
+    def apply(layout, depth, backend, **kwargs):
+        posaxis = ak._util.maybe_posaxis(layout, axis, depth)
+        if depth < posaxis + 1 and layout.is_leaf:
+            raise ak._errors.wrap_error(
+                np.AxisError(f"axis={axis} exceeds the depth of this array ({depth})")
+            )
+        elif depth == posaxis + 1 and layout.is_union:
+            if all(x.is_list for x in layout.contents):
+                if not all(x.is_list for x in layout.contents):
+                    raise ak._errors.wrap_error(NotImplementedError)
```
agoose77, line 53: TODO: add support for all list types, add tests, etc.

jpivarski:
Will this work for arbitrarily many lists deep? Will it work in the presence of any option-types (keeping in mind that unions and options already interact)?

Most importantly, does it solve @ivirshup's problem?

agoose77:
> Most importantly, does it solve @ivirshup's problem?

Yes, in its current form :)

> Will this work for arbitrarily many lists deep? Will it work in the presence of any option-types (keeping in mind that unions and options already interact)?

This is another question (hence the draft status)! My approach with these functions is to keep their scope small; operating only upon parent-child layouts where possible. This means that there are cases where multiple functions are needed:
```python
>>> x = ak.concatenate(([{'x': 1}], [{'y':2}]))
>>> ak.merge_union_of_records(x)
<Array [{x: 1, y: None}, {x: None, ...}] type='2 * {x: ?int64, y: ?int64}'>
```
vs
```python
>>> y = ak.concatenate(([{'x': 1}], [None,{'y':2}]))
>>> ak.merge_union_of_records(y)
<Array [{x: 1}, None, {y: 2}] type='3 * union[{x: int64}, ?{y: int64}]'>
```
This latter example needs the user to first call `ak.merge_option_of_records`:
```python
>>> y = ak.concatenate(([{'x': 1}], [None,{'y':2}]))
>>> ak.merge_option_of_records(y)
<Array [{x: 1}, {y: None}, {y: 2}] type='3 * union[{x: int64}, {y: ?int64}]'>
>>> ak.merge_union_of_records(ak.merge_option_of_records(y))
<Array [{x: 1, y: None}, ..., {x: None, ...}] type='3 * {x: ?int64, y: ?int64}'>
```

My view is that the same should apply here, if possible.

agoose77:
I started writing
> This implies the need for `merge_option_of_lists`, that pushes the option inside the lists. Otherwise, `Union[List[X], Option[List[Y]]]` would pass through this function unchanged. Though I think this is well motivated, do you agree?

Whilst a new `merge_option_of_lists` is motivated, the need is different actually. `Union[Record[X], Option[Record[Y]]]` *can't* be losslessly-simplified[^loss]  (otherwise the union wouldn't exist). Meanwhile, we know that `Union[List[X], List[Y]]` *can* be simplified in such a manner. Therefore, the way that options interact is an additional degree of freedom.

[^loss]: where "lossless" means `tolist()` producing an identical result, in this discussion.

jpivarski:
Well, if it works one list deep, then it can be applied repeatedly (at a specified `axis`).

I didn't want the scope to balloon in the direction of needing yet another function for merging options. This is starting to sound like something that would be cleaner to do some other way.

@ivirshup, why are you concatenating different types? We should understand the use-case better. (I usually think of concatenation for same-type or what ought to be same-typed data, such as partitions of a large dataset. How is your use-case different?)

agoose77:
>  We should understand the use-case better. 

This is an important question w.r.t the original motivation. I'm continuing this discussion given that these functions are intrinsically useful irrespective of this problem.


> Well, if it works one list deep, then it can be applied repeatedly (at a specified axis).

Yes, the problem is when you have mixed option — non-option. This is something that we can (and should) handle. The benefit of adding `ak.merge_option_of_lists` is therefore mostly tangential to this discussion. My discussion above was framed in terms of what we've currently implemented. It actually does make sense for us to support `Union[Option[X], Y]`.

> I didn't want the scope to balloon in the direction of needing yet another function for merging options. This is starting to sound like something that would be cleaner to do some other way.

I'm on the same page. And, although I took a wide route, I don't think we need a new function in this case. We should just be handling unions given that we know what the result should look like.

ivirshup:
> We should understand the use-case better. (I usually think of concatenation for same-type or what ought to be same-typed data, such as partitions of a large dataset. How is your use-case different?)

The use case is really "concatenation of very similar types". E.g the likely use cases for [`pd.concat`](https://pandas.pydata.org/docs/reference/api/pandas.concat.html) and [`xr.concat`](https://docs.xarray.dev/en/stable/generated/xarray.concat.html).

I can give a toy example, but let me know if you'd like something more concrete (it will get a bit bio-y).

Lets say a certain experiment results in data that looks like `n * var * {measure_a: float, measure_b: float}`. I've annotated an experiment with some computed value per measurement (or maybe an annotation) so now it looks like: `n * var * {measure_a: float, measure_b: float, additional_value: int}`. Now I want to combine this with some other experiment that doesn't have the additional value, and has the datashape of the base case. I want to maintain the values of `additional_value` for the observations that have it. 

awkward will give a union type here. This is an issue for a number of reasons, including that they're a little difficult to reason about, to export to other libraries and it'll error if I try to access `"additional_value"` on the combined dataset.

<details>
<summary> </summary>

```python
import awkward as ak, pandas as np, numpy as np

exp_a = ak.Array([
    [{"measure_a": .42, "measure_b": .3, "additional_value": 3}, {"measure_a": .26, "measure_b": .03, "additional_value": 9}],
    [{"measure_a": .29, "measure_b": .01, "additional_value": 0}, {"measure_a": .41, "measure_b": .62, "additional_value": 78}, {"measure_a": .31, "measure_b": .54, "additional_value": 62}],
])

exp_b = ak.Array([
    [{"measure_a": .42, "measure_b": .3}],
    [{"measure_a": .29, "measure_b": .01}, {"measure_a": .41, "measure_b": .62}, {"measure_a": .31, "measure_b": .54}],
    [{"measure_a": .26, "measure_b": .03}, {"measure_a": .41, "measure_b": .62}, {"measure_a": .31, "measure_b": .54}, {"measure_a": .41, "measure_b": .62}, {"measure_a": .31, "measure_b": .54}],
])

ak.concatenate([exp_a, exp_b]).type.show()
# 5 * union[
#     var * {
#         measure_a: float64,
#         measure_b: float64,
#         additional_value: int64
#     },
#     var * {
#         measure_a: float64,
#         measure_b: float64
#     }
# ]
```

</details>

If I were to do this in pandas, I would have a dataframe like `n * {observation: int, measure_a: float, measure_b: float, additional_value: int}` and `n * {entry: int, sub_entry:int, measure_a: float, measure_b: float}`. I could call `pd.concat([experiment_a, experiment_b], join="outer")` and end up with `n * {observation: int, measure_a: float, measure_b: float, additional_value: ?int}`

<details>
<summary> </summary>

```python
exp_a_df = ak.to_dataframe(exp_a).astype({"additional_value": pd.Int64Dtype()})
exp_b_df = ak.to_dataframe(exp_b)

pd.concat([exp_a_df, exp_b_df], join="outer")
#                 measure_a  measure_b  additional_value
# entry subentry                                        
# 0     0              0.42       0.30                 3
#       1              0.26       0.03                 9
# 1     0              0.29       0.01                 0
#       1              0.41       0.62                78
#       2              0.31       0.54                62
# 0     0              0.42       0.30              <NA>
# 1     0              0.29       0.01              <NA>
#       1              0.41       0.62              <NA>
#       2              0.31       0.54              <NA>
# 2     0              0.26       0.03              <NA>
#       1              0.41       0.62              <NA>
#       2              0.31       0.54              <NA>
#       3              0.41       0.62              <NA>
#       4              0.31       0.54              <NA>
```

</details>

The `join` argument from `xr.concat` allows very similar behavior.

-------

This PR will get a little closer to the goal here. There are a few other tasks I would need to do:

* Check if the result is mergable. I'm not sure I want to bother merging arrays where I can't eliminate unions.
* Iteratively call either `merge_unions_of_records`, `merge_unions_of_lists` (based on current top level union) until unions are eliminated.

This should be manageable, but "checking if the result it mergable" may be difficult, and each call to one of the `merge_unions` functions may be expensive (e.g. I think `Union[var * {a}, var * {b}] -> var * Union[{a}, {b}] -> var * {?a, ?b}`  allocates a lot more than a 1 step process would).

jpivarski:
Maybe we're going about this in too complicated a way, trying to fix a union after it has been created. The situation of having new data with more fields than old data is a familiar one, which I'd call schema evolution. If it always involves record types, such as

```python
>>> old_data = ak.Array([[{"a": 1}], [], [{"a": 2}]])
>>> old_data.show(type=True)
type: 3 * var * {
    a: int64
}
[[{a: 1}],
 [],
 [{a: 2}]]
>>> new_data = ak.Array([[], [{"a": 3.3, "b": 3.3}, {"a": 4.4, "b": 4.4}]])
>>> new_data.show(type=True)
type: 2 * var * {
    a: float64,
    b: float64
}
[[],
 [{a: 3.3, b: 3.3}, {a: 4.4, b: 4.4}]]
```

just concatenating them makes an undesirable union,

```python
>>> ak.concatenate((old_data, new_data)).show(type=True)
type: 5 * union[
    var * {
        a: int64
    },
    var * {
        a: float64,
        b: float64
    }
]
[[{a: 1}],
 [],
 [{a: 2}],
 [],
 [{a: 3.3, b: 3.3}, {a: 4.4, b: 4.4}]]
```

but if we explicitly set missing values for the records that don't have them,

```python
>>> old_data["b"] = None
>>> old_data.show(type=True)
type: 3 * var * {
    a: int64,
    b: ?unknown
}
[[{a: 1, b: None}],
 [],
 [{a: 2, b: None}]]
```

then they'll concatenate the way you want them without even needing the special function:

```python
>>> ak.concatenate((old_data, new_data)).show(type=True)
type: 5 * var * {
    a: float64,
    b: ?float64
}
[[{a: 1, b: None}],
 [],
 [{a: 2, b: None}],
 [],
 [{a: 3.3, b: 3.3}, {a: 4.4, b: 4.4}]]
```

This kind of procedure doesn't have to be manual:

```python
>>> newer_data = ak.Array([[{"b": 5, "c": []}, {"b": 6.6, "c": [1, 2]}]])

>>> all_fields = set(old_data.fields) | set(new_data.fields) | set(newer_data.fields)
>>> all_fields
{'a', 'c', 'b'}
>>> for field in all_fields:
...     if field not in old_data.fields:
...         old_data[field] = None
...     if field not in new_data.fields:
...         new_data[field] = None
...     if field not in newer_data.fields:
...         newer_data[field] = None
... 
>>> ak.concatenate((old_data, new_data, newer_data)).show(type=True)
type: 6 * var * {
    a: ?float64,
    b: ?float64,
    c: option[var * int64]
}
[[{a: 1, b: None, c: None}],
 [],
 [{a: 2, b: None, c: None}],
 [],
 [{a: 3.3, b: 3.3, c: None}, {a: 4.4, b: 4.4, c: None}],
 [{a: None, b: 5, c: []}, {a: None, b: 6.6, c: [1, ...]}]]
```

But it is strictly acting on records; there's no equivalent of this for unions of other types, such as `union[int, str]`. But it sounds to me like the problem you're trying to solve is records getting or losing attributes, and they can change in type from, say, integers to floats, but they don't change from integers to strings.

Integers → strings doesn't promote, regardless of whether they're in records or not:

```python
>>> even_newer_data = ak.Array([[{"a": "hello"}]])
>>> ak.concatenate((old_data, new_data, newer_data, even_newer_data)).show(type=True)
type: 7 * union[
    var * {
        a: ?float64,
        b: ?float64,
        c: option[var * int64]
    },
    var * {
        a: string
    }
]
[[{a: 1, b: None, c: None}],
 [],
 [{a: 2, b: None, c: None}],
 [],
 [{a: 3.3, b: 3.3, c: None}, {a: 4.4, b: 4.4, c: None}],
 [{a: None, b: 5, c: []}, {a: None, b: 6.6, c: [1, ...]}],
 [{a: 'hello'}]]
```

ivirshup:
> Maybe we're going about this in too complicated a way, trying to fix a union after it has been created.

Yes, I think this was actually the approach I had been exploring. It starts to fall down if the user provided unions or you hit nested types like:

```
n * {gene_id: str, exons: var * {exon_start: int, exon_end: int}}
```

If you look at the example code for the relevant issue in AnnData (https://github.com/scverse/anndata/issues/898), I was hoping to use the `merge_unions_*` logic to figure out resultant types.

For simple record types, I'd been doing a similar union of records to achieve the above (as discussed: https://github.com/scikit-hep/awkward/issues/1668#issuecomment-1237947580).

jpivarski:
It doesn't fall down all the way:

```python
>>> old_data = ak.Array([{"gene_id": "meow", "exons": [{"exon_start": 100}]}])
>>> new_data = ak.Array([{"gene_id": "woof", "exons": [{"exon_start": 200, "exon_end": 300}]}])

>>> old_data["exons", "exon_end"] = None

>>> ak.concatenate((old_data, new_data)).show(type=True)
type: 2 * {
    gene_id: string,
    exons: var * {
        exon_start: int64,
        exon_end: ?int64
    }
}
[{gene_id: 'meow', exons: [{exon_start: 100, ...}]},
 {gene_id: 'woof', exons: [{exon_start: 200, ...}]}]
```

It has become more complicated to find out what fields are needed and where they're needed, but it's still possible:

```python
>>> new_data.fields
['gene_id', 'exons']

>>> new_data["exons"].fields
['exon_start', 'exon_end']

>>> new_data["exons", "something_deeper"].fields
...
```

(so I suppose there would be some tree-recursion needed there). All of these are working because string tuple-items in square brackets means nested record fields.

It's starting to look like a complex function, but if we can generalize it and give it a good name, that might be the way to go. This method differs from `ak.merge_union_of_records` in that it:

   * works at all list depths; option-type data are automatically handled (no special cases)
   * would have to include the concatenation as part of the operation, whereas `ak.merge_union_of_records` could be separate.

Maybe it should be a non-default option of `ak.concatenate`? Maybe not; it sounds too special-purpose for a generic (and NumPy-extending) function like `ak.concatenate`. Maybe a longer name: `ak.concatenate_merge_records`?

agoose77:
@jpivarski and I discussed this in our meeting today. We proposed the addition of a new `ak.broadcast_fields()` (or similar name) function that broadcasts the field structure of multiple arrays, e.g.
```python
>>> arrays = ak.broadcast_fields([
    [{'x': 1}],
    [{'y': 2}],
    [{'z': 2}],
])
>>> arrays[0].type.show()
3 * {
    x: ?int64,
    y: ?int64,
    z: ?int64
}
```
This function would find the common type of missing fields, and add them in the appropriate places. Where there is no common type, "common type" means "union of types". 

You could then use this function with `ak.concatenate`, i.e.
```python
ak.concatenate(
    ak.broadcast_fields(...)
)
```

By creating unions as deep as possible, this function would not require `ak.merge_union_of_lists`.

Does this sound like a good approach?

grst:
sounds good to me! Happy to give it a try when you have a prototype. 

One question that I have from your example: When `broadcast_fields` is called with a list of arrays, would it again return a list of arrays or a single array? Your first example looks like the former, your second like the latter...

agoose77:
> Your first example looks like the former, your second like the latter...

Yes, a mistake on my part! I've corrected the example.

agoose77:
Closing this PR in favour of #2267

PR #2236, opened 2023-02-13
fix: support options in `ak.merge_union_of_records`

agoose77:
In #2235 it was raised that `merge_union_of_records` should support records wrapped in options. This PR considers that case.

- [x] Add tests

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -1031,9 +1031,16 @@ class TypeTracer(NumpyLike):
         try_touch_data(x)
         return (TypeTracerArray._new(np.int64, (unknown_length,)),) * len(x.shape)
 
+    def where(
+        self, condition: ArrayLike, x1: ArrayLike, x2: ArrayLike
+    ) -> TypeTracerArray:
+        condition, x1, x2 = self.broadcast_arrays(condition, x1, x2)
+        result_dtype = numpy.result_type(x1, x2)
+        return TypeTracerArray._new(result_dtype, shape=condition.shape)
+
     def unique_values(self, x: ArrayLike) -> TypeTracerArray:
         try_touch_data(x)
-        return TypeTracerArray._new(x.dtype, shape=(None,))
+        return TypeTracerArray._new(x.dtype, shape=(unknown_length,))
 
     def concat(self, arrays, *, axis: int | None = 0) -> TypeTracerArray:
         if axis is None:
--- a/src/awkward/operations/ak_merge_union_of_records.py
+++ b/src/awkward/operations/ak_merge_union_of_records.py
@@ -154,7 +154,8 @@ def _impl(array, axis, highlevel, behavior):
             if any(x.is_option for x in layout.contents):
                 # We'll create an outermost indexed-option type, which re-instates the missing values
                 # The values of this don't matter; we'll compute them as a dense index (0 ... N)
-                # where N is the number of non-null items
+                # where N is the number of non-null items. This indexes into the record-of-unions, which
+                # already handles the item ordering
                 outer_option_dense_index = backend.index_nplike.empty(
                     layout.length, dtype=np.int64
                 )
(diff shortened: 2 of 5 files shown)


src/awkward/_nplikes/typetracer.py
```diff
@@ -1031,9 +1031,16 @@ def nonzero(self, x: ArrayLike) -> tuple[TypeTracerArray, ...]:
         try_touch_data(x)
         return (TypeTracerArray._new(np.int64, (unknown_length,)),) * len(x.shape)
 
+    def where(
+        self, condition: ArrayLike, x1: ArrayLike, x2: ArrayLike
+    ) -> TypeTracerArray:
+        condition, x1, x2 = self.broadcast_arrays(condition, x1, x2)
+        result_dtype = numpy.result_type(x1, x2)
```
agoose77, line 1038: TODO: implement `result_type` for all nplikes. This coincides with the planned introduction of the `metadata` submodule

src/awkward/operations/ak_merge_union_of_records.py
```diff
@@ -75,93 +145,121 @@ def apply(layout, depth, backend, **kwargs):
...
-                            )
-
-                    outer_field_contents.append(
-                        ak.contents.UnionArray.simplified(
-                            ak.index.Index8(field_tags),
-                            ak.index.Index64(field_index),
-                            field_contents,
+                        # Find dense index into non-null items of this content for the outer optiontype.
+                        outer_option_index[is_this_tag] = backend.index_nplike.where(
+                            is_non_null, outer_option_index[is_this_tag], -1
```
agoose77: This is something that might be better as a kernel (and perhaps we already have a kernel that is applicable here).

src/awkward/operations/ak_merge_union_of_records.py
```diff
@@ -41,32 +41,102 @@ def _impl(array, axis, highlevel, behavior):
...
+                    ak.index.Index8(field_tags),
+                    ak.index.Index64(field_index),
+                    field_contents,
+                )
             )
-        else:
-            raise ak._errors.wrap_error(TypeError(layout))
+        return ak.contents.RecordArray(
+            outer_field_contents, all_fields, backend=backend
+        )
```
agoose77, line 136: This code has not really changed, I only eliminated the index inversion code (this is now handled separately in the new code)

agoose77:
@jpivarski  I think this is ready for review. It passes some simple tests, and the logic seems intuitive, but it's obviously complex. I'd be keen to know if you recall any kernels I can use instead of `np.where` (or I can look tomorrow), and if anything looks amiss or can be done more performantly.

jpivarski (approved):
Yes, it is complex, but the intentions are well documented in comments. I've read through them.

Regarding `nplike.where` vs. a dedicated kernel: very few kernels are shared by more than one operation, and I wouldn't be able to tell offhand if there was one that could be applied to this case.

Since it is possible with standard NumPy operations, that's preferable, apart from the performance hit of doing multiple passes over the data.

src/awkward/operations/ak_merge_union_of_records.py
```diff
@@ -75,93 +145,116 @@ def apply(layout, depth, backend, **kwargs):
...
+
+                # We'll rebuild the union to include only the non-null items.
+                next_contents = []
+                for tag, content in enumerate(layout.contents):
+                    is_this_tag = backend.index_nplike.asarray(layout.tags) == tag
+
+                    # Rewrite union index of indexed types
+                    if content.is_indexed:
+                        content_index = backend.index_nplike.asarray(content.index)
+                        next_index[is_this_tag] = current_index[is_this_tag][
```
agoose77: This should be inner[outer], not outer[inner]

src/awkward/operations/ak_merge_union_of_records.py
```diff
@@ -75,93 +145,116 @@ def apply(layout, depth, backend, **kwargs):
...
+                # This should have the same length as the original union, and its index should be "dense"
+                # (contiguous, monotonic integers; or -1). Therefore, we can directly compute it from the "sparse"
+                # tags index, which has the same length as the original union, and has only missing items set to -1.
+                outer_option_dense_index = compact_option_index(
+                    next_tags_sparse, backend=backend
+                )
+
+                # Find length of the new (dense) tags array of the inner union
+                total_length = 0
+                for array in inner_union_index_parts:
```
agoose77: Don't need to compute this, could assert though

PR #2240, opened 2023-02-15
fix: use `simplified` in `ak._do.merge_as_union`

### LINKED ISSUE #2239: Can't concatenate more than 3 types into Union
### Version of Awkward Array

2.0.7

When attempting to concatenate more than 3 awkward arrays with different RecordTypes, the concatenation fails because it refuses to create the respective UnionType. Not sure how this relates to #2235, because essentially, I don't want the union, but an OptionType instead. 

```python
>>> ak.concatenate(
    [
        ak.Array(x)
        for x in [
            [{"a": 3}],
            [{"c": 3}],
            [{"d": 3}],
            [{"e": 3}]
        ]
    ]
)
TypeError: while calling

    ak.concatenate(
        arrays = [<Array [{a: 3}] type='1 * {a: int64}'>, <Array [{c: 3}] typ...
        axis = 0
        mergebool = True
        highlevel = True
        behavior = None
    )

Error details: UnionArray cannot contain union-types in its 'contents' (UnionArray); try UnionArray.simplified instead
```
<details>
<summary>Stacktrace</summary>

```pytb
~/tmp/ipykernel_993961/1534422542.py in <cell line: 1>()
----> 1 ak.concatenate(
      2     [
      3         ak.Array(x)
      4         for x in [
      5             [{"a": 3}],

~/anaconda3/envs/scirpy_dev/lib/python3.9/site-packages/awkward/operations/ak_concatenate.py in concatenate(arrays, axis, mergebool, highlevel, behavior)
     43         },
     44     ):
---> 45         return _impl(arrays, axis, mergebool, highlevel, behavior)
     46 
     47 

~/anaconda3/envs/scirpy_dev/lib/python3.9/site-packages/awkward/operations/ak_concatenate.py in _impl(arrays, axis, mergebool, highlevel, behavior)
    109             else:
    110                 collapsed = ak._do.mergemany(batch)
--> 111                 batch = [ak._do.merge_as_union(collapsed, x)]
    112 
    113         out = ak._do.mergemany(batch)

~/anaconda3/envs/scirpy_dev/lib/python3.9/site-packages/awkward/_do.py in merge_as_union(one, two)
    307     )
    308 
--> 309     return ak.contents.UnionArray(tags, index, contents, parameters=None)
    310 
    311 

~/anaconda3/envs/scirpy_dev/lib/python3.9/site-packages/awkward/contents/unionarray.py in __init__(self, tags, index, contents, parameters)
     74                 )
     75             if content.is_union:
---> 76                 raise ak._errors.wrap_error(
     77                     TypeError(
     78                         "{0} cannot contain union-types in its 'contents' ({1}); try {0}.simplified instead".format(

TypeError: while calling

    ak.concatenate(
        arrays = [<Array [{a: 3}] type='1 * {a: int64}'>, <Array [{c: 3}] typ...
        axis = 0
        mergebool = True
        highlevel = True
        behavior = None
    )

Error details: UnionArray cannot contain union-types in its 'contents' (UnionArray); try UnionArray.simplified instead
```

</details>

With only three elements, this works: 
```python
>>> ak.concatenate(
    [
        ak.Array(x)
        for x in [
            [{"a": 3}],
            [{"c": 3}],
            [{"d": 3}],
        ]
    ]
)
[{a: 3},
 {c: 3},
 {d: 3}]
----------------
type: 3 * union[
    {
        a: int64
    },
    {
        c: int64
    },
    {
        d: int64
    }
]
```

**Context:**
I've been playing with the new awkward in AnnData Feature with real data and found this when concatenating a list of 32 AnnData objects, each of which contains a list-of-lists-of-records awkward array. 

CC @ivirshup

agoose77:
## TL;DR
- Remove untestable logic in `UnionArray._mergemany`
- Fix merging of multiple layouts that can be simplified
- Move mergeability check from `UnionArray.__init__` to `_validity_error`

## TL

`UnionArray._merge_strategy` always returns an empty sequence for `tail`. Therefore, any logic depending upon `tail` being non-empty can be removed, as it's more than just an implementation detail: unions always merge everything in one go.

We should use `simplified` to merge arrays that are not known to produce a canonical union. At present we could handle this explicitly; if the contents aren't unions and aren't mergeable, we can use `UnionArray(...)`, but in future this will become more complex if we absorb index arrays into unions as well.

Mergeable checks are "slow" as in the ideal case they must validate at every level of the tree. These should be done in the validity error checks instead.

Fixes #2239

### DIFF
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -306,7 +306,7 @@ def merge_as_union(one: Content, two: Content) -> ak.contents.UnionArray:
         )
     )
 
-    return ak.contents.UnionArray(tags, index, contents, parameters=None)
+    return ak.contents.UnionArray.simplified(tags, index, contents, parameters=None)
 
 
 def mergemany(contents: list[Content]) -> Content:
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -271,7 +271,9 @@ def mergeable(one: Content, two: Content, mergebool: bool = True) -> bool:
     return one._mergeable_next(two, mergebool=mergebool)
 
 
-def merge_as_union(contents: Sequence[Content]) -> ak.contents.UnionArray:
+def merge_as_union(
+    contents: Sequence[Content], parameters=None
+) -> ak.contents.UnionArray:
     length = sum([c.length for c in contents])
     first = contents[0]
     tags = ak.index.Index8.empty(length, first.backend.index_nplike)
@@ -291,7 +293,7 @@ def merge_as_union(contents: Sequence[Content]) -> ak.contents.UnionArray:
         )
         offset += content.length
 
-    return ak.contents.UnionArray(tags, index, contents, parameters=None)
+    return ak.contents.UnionArray(tags, index, contents, parameters=parameters)
 
 
 def mergemany(contents: list[Content]) -> Content:
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -293,7 +293,9 @@ def merge_as_union(
         )
         offset += content.length
 
-    return ak.contents.UnionArray(tags, index, contents, parameters=parameters)
+    return ak.contents.UnionArray.simplified(
+        tags, index, contents, parameters=parameters
+    )
 
 
 def mergemany(contents: list[Content]) -> Content:
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import copy
-from collections.abc import Mapping, MutableMapping
+from collections.abc import Mapping, MutableMapping, Sequence
 from numbers import Integral
 
 import awkward as ak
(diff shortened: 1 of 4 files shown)


agoose77:
We should merge these as multiple batches rather than using simplified.

src/awkward/_do.py
```diff
@@ -271,42 +271,29 @@ def mergeable(one: Content, two: Content, mergebool: bool = True) -> bool:
...
+        content._handle_error(
+            content.backend["awkward_UnionArray_fillindex_count", index.dtype.type](
+                index.data, offset, content.length
+            )
         )
-    )
+        offset += content.length
 
-    return ak.contents.UnionArray(tags, index, contents, parameters=None)
+    return ak.contents.UnionArray(tags, index, contents, parameters=parameters)
```
agoose77: @jpivarski this PR currently fails part of the test suite. The reason is that our many-batch merging scheme can produce batches whose non-adjacent merge results ("partial merges") can themselves be further merged. We discussed during our meeting that this is a possibility, yet our simplification tests check:
```python
for content in contents[1:]:
    if content._mergeable_next(contents[0], ...):
        raise RuntimeError
```
and our simplification algorithm performs an exhaustive `N^2` loop

Interestingly, this means it is actually possible to produce non-canonical union arrays, e.g. if `B` and `C` are mergeable in `[A, B, C]`, we wouldn't detect it in the initialiser. I don't have a solution here, I just wanted to raise it.

For this particular PR, the easiest fix would again be to use `UnionArray.simplified`, which will ensure that any mergeable partial merges are indeed merged. We could also implement our own `N(N-1)/2` algorithm that performs batch merging, but it's diminishing returns. A nudge in the right direction would be nice.

agoose77: We discussed that this should be done, but outside of the initialiser.

agoose77:
@jpivarski pinging your for another review!

Also, I lost my meeting notes (my PC crashed), so whilst I *think* I've covered everything we discussed, let me know if I've missed anything.

jpivarski (approved):
Yes, this is all correct. I did a deep reading of the new `merge_as_union` (looking up the kernel implementations) and it all looks good.

This is ready to merge!

PR #2241, opened 2023-02-15
fix: don't merge non-union parameters

agoose77:
## TL;DR
- Fix old parameter logic in `UnionArray._reverse_merge`

## TL

`UnionArray._reverse_merge` used to try and reconstruct parameters from the `other` layout. Now we just let the `.simplified` logic do this. We only want to take the union parameters in specific circumstances, and  parameters of the contents should only be merged with parameters of the union when they intersect. `UnionArray.simplified` already handles

### DIFF
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1064,14 +1064,8 @@ class UnionArray(Content):
                 AssertionError("FIXME: handle UnionArray with more than 127 contents")
             )
 
-        parameters = ak.forms.form._parameters_union(
-            self._parameters,
-            other._parameters,
-            exclude=ak.forms.form.reserved_nominal_parameters,
-        )
-
         return ak.contents.UnionArray.simplified(
-            tags, index, contents, parameters=parameters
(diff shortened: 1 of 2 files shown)


src/awkward/contents/unionarray.py
```diff
@@ -1064,14 +1064,8 @@ def _reverse_merge(self, other):
...
 
-        parameters = ak.forms.form._parameters_union(
-            self._parameters,
-            other._parameters,
-            exclude=ak.forms.form.reserved_nominal_parameters,
-        )
-
         return ak.contents.UnionArray.simplified(
-            tags, index, contents, parameters=parameters
+            tags, index, contents, parameters=self._parameters
```
jpivarski, line 1068: What if `self` and `other` are _both_ UnionArrays? Should their parameters be merged, then?

jpivarski (approved):

src/awkward/contents/unionarray.py
agoose77: Simplified handles this case, which is nice!

Issue #2214, opened 2023-02-06
IndexError: boolean index did not match indexed array in awkward 2.x vs 1.x


raymondEhlers:
### Version of Awkward Array

2.0.6

### Description and code to reproduce

Following up on my previous report, I've found another case where the behavior varies from awkward 1.x to 2.x in an unexpected manner. The rough outline is that I'm trying to apply a mask to select some subset of nested data. In awkward 1.x, this works fine (I did this with a behavior, which you can see a glimpse of the functionality in the trace - here it's a bit for convenience). In awkward 2.x, this raises an IndexError (trace below):

```python
Cell In [6], line 1
----> 1 jets.jet_splittings.iterative_splittings(jets.subjets)

File ~/software/dev/mammoth/src/mammoth/framework/analysis/jet_substructure.py:387, in JetSplittingArray.iterative_splittings(self, subjets)
    379 def iterative_splittings(self, subjets: SubjetArray) -> SubjetArray:
    380     """Retrieve the iterative splittings.
    381
    382     Args:
   (...)
    385         The splittings which are part of the iterative splitting chain.
    386     """
--> 387     return cast(SubjetArray, self[subjets.iterative_splitting_index])

File ~/software/dev/mammoth/src/mammoth/framework/analysis/jet_substructure.py:253, in SubjetArray.iterative_splitting_index(self)
    250 @property
    251 def iterative_splitting_index(self) -> AwkwardArray[int]:
    252     """Indices of splittings which were part of the iterative splitting chain."""
--> 253     return self.parent_splitting_index[self.part_of_iterative_splitting]

File ~/software/dev/mammoth/.venv/lib/python3.10/site-packages/awkward/highlevel.py:956, in Array.__getitem__(self, where)
    527 """
    528 Args:
    529     where (many types supported; see below): Index of positions to
   (...)
    953 have the same dimension as the array being indexed.
    954 """
    955 with ak._errors.SlicingErrorContext(self, where):
--> 956     out = self._layout[where]
    957     if isinstance(out, ak.contents.NumpyArray):
    958         array_param = out.parameter("__array__")

File ~/software/dev/mammoth/.venv/lib/python3.10/site-packages/awkward/contents/content.py:509, in Content.__getitem__(self, where)
    508 def __getitem__(self, where):
--> 509     return self._getitem(where)

File ~/software/dev/mammoth/.venv/lib/python3.10/site-packages/awkward/contents/content.py:554, in Content._getitem(self, where)
[...]

PR #2246, opened 2023-02-15
fix: boolean slicing with non-packed arrays

agoose77:
In fixing #2214, I missed the fact that `to_ListOffsetArray64(True)` provides no guarantee about the _length_ of the content. This PR fixes that oversight, by using `ak._do.flatten` which ensures the length of the result is correct.

### DIFF
--- a/src/awkward/_slicing.py
+++ b/src/awkward/_slicing.py
@@ -493,14 +493,19 @@ def _normalise_item_bool_to_int(item: Content, backend: Backend) -> Content:
         if item_backend.nplike.known_data:
             item = item.to_ListOffsetArray64(True)
             localindex = ak._do.local_index(item, axis=1)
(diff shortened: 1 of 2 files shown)


src/awkward/_slicing.py
```diff
@@ -522,7 +527,9 @@ def _normalise_item_bool_to_int(item: Content, backend: Backend) -> Content:
     ):
         if item_backend.nplike.known_data:
             if isinstance(item_backend.nplike, Jax):
-                raise wrap_error("This slice is not supported for JAX differentiation.")
+                raise wrap_error(
+                    TypeError("This slice is not supported for JAX differentiation.")
+                )
             # missing values as any integer other than -1 are extremely rare
```
agoose77, line 533: Unrelated change.

jpivarski (approved):

PR #2247, opened 2023-02-16
docs: improve ragged indexing docs

agoose77:
## TL;DR
- Slightly rework `how-to-filter-ragged.md`.
- Add `how-to-filter-masked.md`.
- Rename "ragged indexing" to "Awkward indexing".

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -99,9 +99,9 @@ subtrees:
                       - file: user-guide/how-to-filter-cut-mask
                         title: "Cuts vs. masks [todo]"
                       - file: user-guide/how-to-filter-ragged
-                        title: "Slicing lists within arrays"
+                        title: "Using ragged arrays"
                       - file: user-guide/how-to-filter-masked
-                        title: "Slices with missing values [todo]"
+                        title: "Using arrays with missing values"
 
               - file: user-guide/how-to-restructure
                 title: "Restructuring data"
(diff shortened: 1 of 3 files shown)


agoose77:
@jpivarski I noticed recently that boolean indexing is not strictly required to have the same shape as the underlying array, which is a reflection of the fact that we normalise boolean arrays to integers without knowledge of the array being indexed. 

As I see it, this is a policy decision. If we want to permit this, then we don't need to fix anything. If not, then we probably need to avoid this normalisation and explicitly handle the boolean arrays in each content's `_getitem_XXX`.

jpivarski (approved):
This is new text, and it looks good; explaining not just the slicing rules but how to use them with `argmax` and such.

We can get another tutorial "for free" by _moving_ this docstring:

https://github.com/scikit-hep/awkward/blob/082e4857607959edbee13a439a476274df537c26/src/awkward/highlevel.py#L530-L948

into the tutorial area where it will be more visible.

We've used three words now for mostly the same thing: "jagged" (I started with that because Wikipedia preferred it), "ragged" (this is what I should have used, because it's more widespread in the SciPy community), and "awkward" (new here). Are you using a different word than "ragged" because it also includes missing values? I wonder if "ragged, masked indexing" might be better, since it ties in with a word the reader might already know.

I try to use capitalization consistently and have decided to capitalize "Awkward Array" and even "Awkward" when it's used as an adjective: "Awkward indexing." If it's lowercase, it will less likely be recognized as a brand name, and then it takes on the ordinary English meaning of "clumsy or difficult."

PR #2251, opened 2023-02-16
fix: set dtype in `full_like`

### LINKED ISSUE #2250: `ak.full_like` does not fill with correct value when `dtype` is changed
### Version of Awkward Array

2.0.8

For example:
```python
>>> import awkward as ak
>>> ak.full_like([True], 2, dtype=int)
<Array [1] type='1 * int64'>
```

I would expect the array to get filled with `dtype(fill_value)`, so obviously just `2` in this case. This is the behavior in NumPy:

```python
>>> import numpy as np
>>> np.full_like([True], 2, dtype=int)
array([2])
```

agoose77:
This PR directly sets `dtype` in the `nplike.full_like` call, which I think is all we need to do here.

Fixes #2250

### DIFF
--- a/src/awkward/operations/ak_full_like.py
+++ b/src/awkward/operations/ak_full_like.py
@@ -99,65 +99,7 @@ def _impl(array, fill_value, highlevel, behavior, dtype):
         nplike = layout.backend.nplike
         index_nplike = layout.backend.index_nplike
 
-        if fill_value is _ZEROS and (
-            layout.parameter("__array__") in {"bytestring", "string"}
-        ):
-            name = layout.parameter("__array__")
-
-            asbytes = nplike.frombuffer(b"", dtype=np.uint8)
-            return ak.contents.ListArray(
-                ak.index.Index64(
-                    index_nplike.zeros(layout.length, dtype=np.int64),
-                    nplike=index_nplike,
-                ),
-                ak.index.Index64(
-                    index_nplike.zeros(layout.length, dtype=np.int64),
-                    nplike=index_nplike,
-                ),
-                ak.contents.NumpyArray(
-                    asbytes,
-                    parameters={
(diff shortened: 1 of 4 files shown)


agoose77:
@jpivarski I also added support for `EmptyArray` if a `dtype` is passed. Unlike `ak.values_astype`, I didn't add a `including_unknown` flag because this seems like a bugfix and a sensible choice. Is there a reason that we added `including_unknown` to `values_astype` besides backwards compatibility? Would you like me to do the same here?

agoose77:
It *also* changes the result of one test, which I think is an improvement. Let me know if you disagree.

jpivarski:
The reason is that sometimes an unknown type fills in for a NumpyArray, and sometimes it fills in for something else.

In a situation like this:

```python
>>> partition_1 = ak.from_json("[[1.1, 2.2], [3.3]]")
>>> partition_1.type.show()
2 * var * float64
>>> 
>>> partition_2 = ak.from_json("[[]]")
>>> partition_2.type.show()
1 * var * unknown
>>> 
>>> partition_3 = ak.from_json("[]")
>>> partition_3.type.show()
0 * unknown
```

you wouldn't want to fill in the unknown types as a byproduct of turning the floats into integers or something. You'd want these unknowns to stay unknown until concatenation.

But there are other situations in which you want unknown to become numeric, like `np.array([])`.

So I think it would be a good idea for both `values_astype` and `full_like` (with `dtype`) to have a boolean `including_unknown` argument.

agoose77:
> But there are other situations in which you want unknown to become numeric

That's a good point, and links to our conversations yesterday!

jpivarski (approved):
I looked over this when it was in the draft stage, and just did again now. I think it's great and ready to be merged!

PR #2252, opened 2023-02-16
chore: use scikit-build-core 0.2's build-dir

henryiii:
Awkward-cpp: initial build: 29 seconds & 174 files
Rebuild: 9.8 seconds and 5 files


agoose77:
We discussed this already, I shall merge! Thanks @henryiii, that's a task ticked off.

jpivarski:
Note to self: because of this and #2184 (which also touched a file in `awkward-cpp`), we'll need a new `awkward-cpp` version before releasing the next `awkward` patch release.

PR #2258, opened 2023-02-21
feat: add 'highlevel' and 'behavior' arguments to 'ak.from_rdataframe' function

### LINKED ISSUE #2256: add `highlevel` and `behavior` arguments to `from_rdataframe`
> I just noticed that `from_rdataframe` doesn't have `highlevel` and `behavior` arguments, though almost all of the `ak.*` operations do. As it is, this assumes `highlevel=True, behavior=None`. As possible, low-priority future work (independently of this PR), it would be nice to add these arguments.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/pull/2234#pullrequestreview-1306048587_

ianna:
Fixes https://github.com/scikit-hep/awkward/issues/2256

### DIFF
--- a/src/awkward/operations/ak_from_rdataframe.py
+++ b/src/awkward/operations/ak_from_rdataframe.py
@@ -6,9 +6,9 @@ import awkward as ak
 def from_rdataframe(
     rdf,
     columns,
-    offsets_type="int64_t",
-    keep_order=False,
     *,
+    keep_order=False,
+    offsets_type="int64_t",
     highlevel=True,
     behavior=None,
 ):
--- a/src/awkward/operations/ak_from_rdataframe.py
+++ b/src/awkward/operations/ak_from_rdataframe.py
@@ -1,6 +1,9 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
 import awkward as ak
+from awkward._nplikes.numpylike import NumpyMetadata
+
+np = NumpyMetadata.instance()
 
 
 def from_rdataframe(
@@ -8,7 +11,7 @@ def from_rdataframe(
     columns,
     *,
     keep_order=False,
-    offsets_type="int64_t",
+    offsets_type="int64",
     highlevel=True,
     behavior=None,
 ):
@@ -18,9 +21,10 @@ def from_rdataframe(
             Awkward Array.
         columns (str or iterable of str): A column or multiple columns to be
             converted to Awkward Array.
-        offsets_type (str): A C++ type of the ListOffsetArray offsets.
         keep_order (bool): If set to `True` the columns with Awkward type will
             keep order after filtering.
+        offsets_type (str): A `NumpyType.primitive` type of the ListOffsetArray
+            offsets: `"int32"`, `"uint32"` or `"int64"`.
         highlevel (bool): If True, return an #ak.Array; otherwise, return
             a low-level #ak.contents.Content subclass.
         behavior (None or dict): Custom #ak.behavior for the output array, if
@@ -67,6 +71,20 @@ def _impl(data_frame, columns, highlevel, behavior, offsets_type, keep_order):
                 f"'columns' must be a string or an iterable of strings, not {columns!r}"
             )
         )
+
+    if not isinstance(offsets_type, str) or offsets_type not in (
+        "int32",
+        "uint32",
+        "int64",
+    ):
+        raise ak._errors.wrap_error(
+            TypeError(
+                "'offsets_type' must be a string in (int32, uint32, int64), "
+                "not {}".format(repr(offsets_type))
+            )
+        )
+    else:
+        offsets_type = f"{offsets_type}_t"
 
     out = ak._connect.rdataframe.from_rdataframe.from_rdataframe(
         data_frame,
--- a/src/awkward/operations/ak_from_rdataframe.py
+++ b/src/awkward/operations/ak_from_rdataframe.py
@@ -3,13 +3,24 @@
 import awkward as ak
 
 
-def from_rdataframe(rdf, columns, offsets_type="int64_t", keep_order=False):
+def from_rdataframe(
+    rdf,
+    columns,
+    highlevel=True,
+    behavior=None,
+    offsets_type="int64_t",
+    keep_order=False,
+):
     """
     Args:
         rdf (`ROOT.RDataFrame`): ROOT RDataFrame to convert into an
             Awkward Array.
         columns (str or iterable of str): A column or multiple columns to be
             converted to Awkward Array.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
         offsets_type (str): A C++ type of the ListOffsetArray offsets.
         keep_order (bool): If set to `True` the columns with Awkward type will
             keep order after filtering.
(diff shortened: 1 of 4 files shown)


src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -3,13 +3,24 @@
...
 
-def from_rdataframe(rdf, columns, offsets_type="int64_t", keep_order=False):
+def from_rdataframe(
+    rdf,
+    columns,
+    highlevel=True,
+    behavior=None,
+    offsets_type="int64_t",
+    keep_order=False,
+):
```
agoose77: I think we need to add these at the end of the argument list, so that we don't break existing positional uses. We also want to make `ak.operations.*` functions use keyword-only arguments where it makes sense, and that currently includes `highlevel` and `behavior`, e.g.:
```suggestion
    offsets_type="int64_t",
    keep_order=False,
    *,
    highlevel=True,
    behavior=None,
):
```

At some point, I think we'd want `offsets_type` and `keep_order` to be keyword-only too, but that will be a breaking change that we might need to do by a deprecation cycle.

src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -3,13 +3,24 @@
...
             Awkward Array.
         columns (str or iterable of str): A column or multiple columns to be
             converted to Awkward Array.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
         offsets_type (str): A C++ type of the ListOffsetArray offsets.
         keep_order (bool): If set to `True` the columns with Awkward type will
             keep order after filtering.
```
agoose77:
```suggestion
        offsets_type (str): A C++ type of the ListOffsetArray offsets.
        keep_order (bool): If set to `True` the columns with Awkward type will
            keep order after filtering.
        highlevel (bool): If True, return an #ak.Array; otherwise, return
            a low-level #ak.contents.Content subclass.
        behavior (None or dict): Custom #ak.behavior for the output array, if
            high-level.
```

agoose77:
@ianna sorry, didn't realise this was still a draft!

ianna:
> @ianna sorry, didn't realise this was still a draft!

no worries - thanks for the suggestions!

src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -3,7 +3,15 @@
...
-def from_rdataframe(rdf, columns, offsets_type="int64_t", keep_order=False):
+def from_rdataframe(
+    rdf,
+    columns,
+    offsets_type="int64_t",
+    keep_order=False,
+    *,
+    highlevel=True,
+    behavior=None,
+):
```
jpivarski: I think it would be more natural to put the separator (`*`) between non-keyword-only and keyword-only arguments directly after `columns`.

Generally speaking, any boolean parameters should be keyword-only because a bare word "`False`" or "`True`" has little information, but also any rarely used arguments, which definitely includes `offsets_type`.

```suggestion
def from_rdataframe(
    rdf,
    columns,
    *,
    keep_order=False,
    offsets_type="int64_t",
    highlevel=True,
    behavior=None,
):
```

src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -30,14 +42,16 @@ def from_rdataframe(rdf, columns, offsets_type="int64_t", keep_order=False):
         {
             "rdf": rdf,
             "columns": columns,
+            "highlevel": highlevel,
+            "behavior": behavior,
             "offsets_type": offsets_type,
             "keep_order": keep_order,
         },
```
jpivarski: The order of this dict needs to match the order of arguments, since it's used in the error message.

```suggestion
        {
            "rdf": rdf,
            "columns": columns,
            "keep_order": keep_order,
            "offsets_type": offsets_type,
            "highlevel": highlevel,
            "behavior": behavior,
        },
```

jpivarski (commented):
While we're talking interface, users of Awkward Array would not be expecting to express `offsets_type` as a C++ type name; they'd be expecting a `NumpyType.primitive` string, which (for `ListOffsetArray.offsets`) can be one of

  * `"int32"`
  * `"uint32"`
  * `"int64"`

Since this is a complete set of strings that would ever be allowed, the implementation can be thin: just checking for one of these three and immediately appending a `"_t"`. Outside of this function, no one needs to know that it passes around a C++ type almost everywhere.

src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -1,18 +1,37 @@
...
+def from_rdataframe(
+    rdf,
+    columns,
+    *,
+    keep_order=False,
+    offsets_type="int64",
+    highlevel=True,
+    behavior=None,
+    with_name=None,
+):
```
jpivarski:
```suggestion
def from_rdataframe(
    rdf,
    columns,
    *,
    keep_order=False,
    offsets_type="int64",
    with_name=None,
    highlevel=True,
    behavior=None,
):
```

because `highlevel` and `behavior` are generally last. Here's `ak.zip` as an example:

https://github.com/scikit-hep/awkward/blob/3535fbb8b4bc3443c09ba595620610fee8be0b74/src/awkward/operations/ak_zip.py#L9-L19

src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -1,18 +1,37 @@
...
         keep_order (bool): If set to `True` the columns with Awkward type will
             keep order after filtering.
+        offsets_type (str): A `NumpyType.primitive` type of the ListOffsetArray
+            offsets: `"int32"`, `"uint32"` or `"int64"`.
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+        with_name (None or str): Gives tuples and records a name that can be
+            used to override their behavior (see #ak.Array).
```
jpivarski:
```suggestion
        offsets_type (str): A `NumpyType.primitive` type of the ListOffsetArray
            offsets: `"int32"`, `"uint32"` or `"int64"`.
        with_name (None or str): Gives tuples and records a name that can be
            used to override their behavior (see #ak.Array).
        highlevel (bool): If True, return an #ak.Array; otherwise, return
            a low-level #ak.contents.Content subclass.
        behavior (None or dict): Custom #ak.behavior for the output array, if
            high-level.
```

src/awkward/operations/ak_from_rdataframe.py
```diff
@@ -30,14 +49,21 @@ def from_rdataframe(rdf, columns, offsets_type="int64_t", keep_order=False):
         {
             "rdf": rdf,
             "columns": columns,
-            "offsets_type": offsets_type,
             "keep_order": keep_order,
+            "offsets_type": offsets_type,
+            "highlevel": highlevel,
+            "behavior": behavior,
+            "with_name": with_name,
```
jpivarski:
```suggestion
            "keep_order": keep_order,
            "offsets_type": offsets_type,
            "with_name": with_name,
            "highlevel": highlevel,
            "behavior": behavior,
```

jpivarski (approved):
Great! My only quibble is with the order of the `with_name` argument. By the way, I think it's good to have a `with_name` here because `ak.from_rdataframe` is L1 (see https://github.com/scikit-hep/awkward/issues/2262#issuecomment-1440223884).

I'll apply these and merge the PR. Thanks!

PR #2267, opened 2023-02-23
feat: add `ak.broadcast_fields`

### LINKED ISSUE #2211: `ak.merge_union_of_records` not merging `Union[var * {a: int, b:int}, var * {a: int}]`

agoose77:
This implementation of `ak.broadcast_fields` is just a starting point for discussion. It creates union-types for missing fields, though it does not descend through unions or attempt to coerce any other types.

Fixes #2211 through the new `ak.broadcast_fields` function, which can be used in conjunction with `ak.concatenate`

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -411,6 +411,17 @@ class RecordArray(Content):
         else:
             return out[: self._length]
 
+    def maybe_content(self, index_or_field) -> Content:
+        if self.has_field(index_or_field):
+            return self.content(index_or_field)
+        else:
+            return ak.contents.IndexedOptionArray(
+                ak.index.Index64(
+                    self._backend.index_nplike.full(self.length, -1, dtype=np.int64)
+                ),
+                ak.contents.EmptyArray(),
+            )
+
     def _getitem_nothing(self) -> Content:
         return self._getitem_range(0, 0)
 
--- a/src/awkward/forms/unionform.py
+++ b/src/awkward/forms/unionform.py
@@ -220,8 +220,8 @@ class UnionForm(Form):
 
     @property
     def fields(self):
-        fieldslists = [cont.fields for cont in self._contents]
-        return list(set.intersection(*[set(x) for x in fieldslists]))
+        all_fields = [f for cont in self._contents for f in cont.fields]
+        return ak._util.unique_list(all_fields)
 
     @property
     def is_tuple(self):
--- a/src/awkward/forms/unionform.py
+++ b/src/awkward/forms/unionform.py
@@ -1,5 +1,6 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
+from collections import Counter
 from collections.abc import Iterable
 
 import awkward as ak
@@ -220,8 +221,8 @@ class UnionForm(Form):
 
     @property
     def fields(self):
-        all_fields = [f for cont in self._contents for f in cont.fields]
-        return ak._util.unique_list(all_fields)
+        field_counts = Counter([f for c in self._contents for f in c.fields])
+        return [f for f, n in field_counts.items() if n == len(self._contents)]
 
     @property
     def is_tuple(self):
--- a/src/awkward/operations/ak_broadcast_fields.py
+++ b/src/awkward/operations/ak_broadcast_fields.py
@@ -79,6 +79,9 @@ def _impl(arrays, highlevel, behavior):
         else:
             raise ak._errors.wrap_error(AssertionError("unexpected content type"))
 
+    # Like broadcast_and_apply, we want to walk into each layout, correct the structure, and then rebuilt the arrays
+    # We do this using "pull back" functions that accept a child content, and return the top-level layout. Unlike
+    # layout.copy, the pull-back functions can be arbitrarily deep: the closures maintain the structure of the array
     def recurse(inputs):
         # Descend to records, identities, or leaves
         pullbacks, next_inputs = zip(
@@ -118,8 +121,12 @@ def _impl(arrays, highlevel, behavior):
 
             # We only want to recurse into non-missing fields, so we build this list separately as a generator
             recursed_field_layouts = iter(recurse(layouts_to_recurse))
+
             # Now we build the final list of layouts for this field, choosing between the recursion result and the
             # original layout according to whether the layout was recursed into
+            # The pattern here is that `layouts_for_field` maintains positional correspondence with the `records`,
+            # but uses `None` as a token for "recursed". In this case, we take the layout from `recursed_field_layouts`
+            # using the knowledge that `len(layouts_to_recurse)` corresponds to the number of `None`s
             layouts_by_field.append(
                 [
                     next(recursed_field_layouts) if layout is None else layout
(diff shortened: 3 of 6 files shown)


agoose77:
@jpivarski ignore this PR as it stands: it's wrong, and ATM just a place for me to get a feel for the problem. We would have spoken about this during our meeting this week, that I missed. 

There are lots of edge cases to think about, so I was hoping that we could run through them. The purpose of this function is to make record array" compatible" with one another, like broadcasting, but without paying heed to the length of each dimension. We could interpret that more or less loosely; what should we do for arrays with different _dimensions_? Previously we've spoken about the dimensions being an entirely independent d.o.f, so I'm currently working on that basis.

In general, I'm thinking of this function being exclusively scoped to the "field" behavior, i.e. not concerned about options or lists being compatible between layouts. 

In the following examples, we don't care about the array contents, so I'll only display the types with
```python
for x in arrays:
    x.show(type=True)
```

## Missing fields, same dimension
```python
ak.broadcast_fields(
        [{"x": 1, "c": 3}],
        [{"y": 2, "c": 3}],
        [{"z": 2, "c": 3}],
)
```

Either we introduce the option for all fields where it exists for one:
```python
1 * {
    x: ?int64,
    c: int64,
    y: ?int64,
    z: ?int64
}
1 * {
    x: ?int64,
    c: int64,
    y: ?int64,
    z: ?int64
}
1 * {
    x: ?int64,
    c: int64,
    y: ?int64,
    z: ?int64
}
```

or we never introduce an unneeded option:
```python
1 * {
    x: int64,
    c: int64,
    y: ?int64,
    z: ?int64
}
1 * {
    x: ?int64,
    c: int64,
    y: int64,
    z: ?int64
}
1 * {
    x: ?int64,
    c: int64,
    y: ?int64,
    z: int64
}
```


## Missing fields, different dimension
```python
ak.broadcast_fields(
        [{"x": {"y": 1, "z": 2}}],
        [{"x": [  {"y": 1}  ]}],
)
```

We ignore the dimension difference, and walk down each content tree until we reach a record:
```python1 * {
    x: {
        y: int64,
        z: int64
    }
}
1 * {
    x: var * {
        y: int64,
        z: ?int64
    }
}
```

## Union types
I didn't get this far yet.

src/awkward/contents/recordarray.py
```diff
@@ -411,6 +411,17 @@ def content(self, index_or_field: str | SupportsIndex) -> Content:
...
+        if self.has_field(index_or_field):
+            return self.content(index_or_field)
+        else:
+            return ak.contents.IndexedOptionArray(
+                ak.index.Index64(
+                    self._backend.index_nplike.full(self.length, -1, dtype=np.int64)
+                ),
+                ak.contents.EmptyArray(),
+            )
+
```
agoose77, line 424: I think this is simple enough and useful enough to warrant inclusion. Easily moved on this, though.

src/awkward/forms/unionform.py
```diff
@@ -220,8 +220,8 @@ def branch_depth(self):
 
     @property
     def fields(self):
-        fieldslists = [cont.fields for cont in self._contents]
-        return list(set.intersection(*[set(x) for x in fieldslists]))
+        all_fields = [f for cont in self._contents for f in cont.fields]
+        return ak._util.unique_list(all_fields)
```
agoose77: I think this is an improvement; `set.intersection` isn't ordered, and we do care about field ordering. This changes makes the order depend upon position in the `contents` list.

src/awkward/operations/ak_broadcast_fields.py
```diff
@@ -0,0 +1,130 @@
...
+    Args:
+        arrays: Array-like data (anything #ak.to_layout recognizes).
+        highlevel (bool): If True, return an #ak.Array; otherwise, return
+            a low-level #ak.contents.Content subclass.
+        behavior (None or dict): Custom #ak.behavior for the output array, if
+            high-level.
+
+    Return a list of arrays whose types contain the same number of fields. Unlike
+    #ak.broadcast_arrays, this function does not require record types to occur at the
+    same depths. Where fields are missing from a record ...
```
agoose77: TODO

src/awkward/forms/unionform.py
agoose77: This was also a mistake :laugh:, I've replaced it with an ordered intersection.

src/awkward/contents/recordarray.py
jpivarski: Yeah, that's a nice way to generalize.

src/awkward/forms/unionform.py
```diff
@@ -220,8 +221,8 @@ def branch_depth(self):
 
     @property
     def fields(self):
-        fieldslists = [cont.fields for cont in self._contents]
-        return list(set.intersection(*[set(x) for x in fieldslists]))
+        field_counts = Counter([f for c in self._contents for f in c.fields])
+        return [f for f, n in field_counts.items() if n == len(self._contents)]
```
jpivarski, line 225: The old algorithm was trying to say, "If a field name is in all contents, include it." The old algorithm may have lost field order, and that's a problem because this `fields` property should be ordered.

The new algorithm counts how many times a name is seen, and if that's equal to the number of union contents, it's a returned field. It definitely maintains field order, which is better than the old code. But is it always right? I suppose that one union content can't have the same field name twice, so if a field name ever reaches the number of union contents, it must have been in every content.

So I believe that it works! But an essential part of the proof is that a given field name can only appear in a union content zero or one times.

tests/test_2267_broadcast_fields.py
```diff
@@ -0,0 +1,29 @@
...
+    a, b, c = ak.broadcast_fields(
+        [{"x": 1, "z": {"a": 2, "b": 3}}],
+        [{"y": 2}],
+        [{"y": 3, "z": {"c": 9}}],
+    )
+    assert ak.almost_equal(a, [{"x": 1, "z": {"a": 2, "b": 3, "c": None}, "y": None}])
+    assert ak.almost_equal(b, [{"x": None, "z": None, "y": 2}])
+    assert ak.almost_equal(
+        c, [{"x": None, "z": {"a": None, "b": None, "c": 9}, "y": 3}]
+    )
```
jpivarski, line 29: This is exactly what we want for results. It's nice the way that this maintains field order: I wasn't expecting that.

Although the order of field names is an important and preserved part of a record array, record arrays are always compared for type equality, merging, concatenation, etc. in a way that allows fields from different arrays to not have the same order.

This is a nice touch, though!

jpivarski (approved):
Although I didn't have comments on the `ak.broadcast_fields` algorithm itself, I did look it over and it all looks good. I wasn't expecting field order to be maintained among the different arrays, but it's a nice feature to have.

Actually, it's an especially nice feature if the user's next operation after this is `ak.unzip`.

jpivarski:
I think it's ready to merge, and I'll let you do it, in case you have offline edits.

tests/test_2267_broadcast_fields.py
agoose77: > Although the order of field names is an important and preserved part of a record array, record arrays are always compared for type equality, merging, concatenation, etc. in a way that allows fields from different arrays to not have the same order.

I hadn't noticed that!

I ran into problems in v1 when building Parquet datasets that (IIRC) depended upon field ordering, so I make it an explicit check to ensure we keep order where possible.

src/awkward/forms/unionform.py
agoose77: Yes, this is an assumption. If we don't check this, we certainly should as it seems highly irregular (only L2 users could even access repeated fields, right?)

PR #2271, opened 2023-02-27
ci: test coffea & dask

### LINKED ISSUE #1702: Add CI workflow to test third party packages against master
### Version of Awkward Array

main

boost-histogram clones hist's master, and checks that it works against boost-histogram. We could do this to ensure that we don't break people e.g. `anndata` (if they have a test suite).

- vector 
- anndata

@henryiii mentioned `mypy_primer` as the inspiration for this.

If we pin, then we can treat these as workflow tests (non-required).

agoose77:
Closes #1702

This PR adds third-party tests to our CI, to improve visibility over downstream usages. We expect that these tests will fail, for now, because we don't implement everything required for a 100% typetracer workflow. In time, the number of test failures should decrease.


agoose77:
@jpivarski does this workflow look reasonable to you, to merge as a non-required workflow?

.github/workflows/downstream-test.yml
```diff
@@ -0,0 +1,98 @@
...
+        with:
+          repository: CoffeaTeam/coffea
+          path: coffea
+
+      - name: Install coffea
+        run: python -m pip install -v "./coffea[dev,dask]"
+
+      - name: Run coffea tests
+        run: python -m pytest -vv -rs tests/test_dask*
+        working-directory: coffea
```
jpivarski, line 99: It looks like these are all the new tests: dask-awkward and Coffea.

The Coffea ones took 18 minutes because of things like Spark. Although it's not great to have to maintain a fork, perhaps we could pick and choose which tests to run?

jpivarski (approved):
This is a good set! It seems to be failing in dask-awkward right now (i.e. not even testing Coffea yet).

Oh, as a _non-required_ test, yes.

.github/workflows/downstream-test.yml
agoose77: We should only be running the `test_dask*` coffea tests. Did you see this locally?

jpivarski: No, in my haste, I missed this:

> `tests/test_dask`

agoose77: Fab, I hope that constrains it to a much more reasonable set. I did notice just how long the Coffea test suite takes otherwise :S

lgray: Now that we've got user defined stepping in uproot and dak/ak are coming along I can finally overhaul Runner/Executor in coffea to not be completely eager-mode. This will reduce the time needed to run takes by quite a lot.

I can also implement some tests that track necessary columns expectations with complex examples, which is rather fast, so that we can easily catch any drift there.

agoose77:
I believe that Jim's work to integrate everything in https://github.com/CoffeaTeam/integration-test would supersede this.

PR #2274, opened 2023-02-28
docs: fix build for latest theme

### LINKED ISSUE #2273: Tests are failing in "Build Docs", possibly due to new Sphinx theme
Is pydata_sphinx_theme broken, perhaps? It had a 0.13 release 12 hours ago.

_Originally posted by @henryiii in https://github.com/scikit-hep/awkward/issues/2268#issuecomment-1448934363_

For example, https://github.com/scikit-hep/awkward/actions/runs/4297198800/jobs/7489883737

This open issue, https://github.com/pydata/pydata-sphinx-theme/issues/1149, has the same error message and they say

> I think it ought to be possible to run our basic dev commands (test, docs, docs-live) in the dev environment we recommend contributors to use

which sounds like something we do.

agoose77:
Fixes #2273

### DIFF
--- a/docs/_templates/breadcrumbs.html
+++ b/docs/_templates/breadcrumbs.html
@@ -1,4 +1,0 @@
-{%- extends "sphinx_rtd_theme/breadcrumbs.html" %}
-
-{% block breadcrumbs_aside %}
-{% endblock %}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -84,8 +84,8 @@ html_theme = "pydata_sphinx_theme"
 html_show_sourcelink = True
 html_theme_options = {
     "logo": {
-        "image_light": "image/logo-300px.png",
-        "image_dark": "image/logo-300px-white.png",
+        "image_light": "_static/image/logo-300px.png",
+        "image_dark": "_static/image/logo-300px-white.png",
     },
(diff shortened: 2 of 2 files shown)


docs/_templates/breadcrumbs.html
```diff
@@ -1,4 +0,0 @@
-{%- extends "sphinx_rtd_theme/breadcrumbs.html" %}
-
-{% block breadcrumbs_aside %}
-{% endblock %}
```
jpivarski, line 4: We had a custom breadcrumbs.html to remove some feature, probably an "edit on GitHub" link that didn't work because of how things were set up. Is all of that obsolete now?

jpivarski (approved):
You can merge it if the answer to the question below is "yes."

docs/_templates/breadcrumbs.html
agoose77: This currently doesn't do anything as we don't use `sphinx_rtd_theme`. However, `pydata_sphinx_theme` just added breadcrumbs, so now this no-op template loads and breaks things. We should probably re-introduce the feature that this template provided, on a more selective basis; generated listings, unlike the user guide, are not editable. I'll look into that in a new PR :)

PR #2276, opened 2023-03-01
fix: keep column order in all tests

ianna:


tests/test_1473_from_rdataframe.py
```diff
@@ -9,6 +9,7 @@
 
 ROOT = pytest.importorskip("ROOT")
 
+ROOT.ROOT.EnableImplicitMT(1)
```
ianna, line 12: While there is a [ROOT::DisableImplicitMT()](https://root.cern/doc/master/namespaceROOT.html#af6b6120e5533533bfe589e5ed91845f0) in C++ it looks like my version of ROOT does not have a Python equivalent. Here `EnableImplicitMT(1) `creates a thread-pool of size 1.

tests/test_1781_rdataframe_snapshot.py
```diff
@@ -49,7 +52,11 @@ def test_data_frame_vec_of_vec_of_real(tmp_path):
     )
     assert ak_array_in.to_list() == ak_array_out["x"].to_list()
 
-    with pytest.raises(SystemError):
+    # With `ROOT.ROOT.EnableImplicitMT(1)` a SystemError becomes a Warning:
```
ianna, line 55: This is kind of strange that an exception becomes a warning. The warning is caught beneath, but it is empty...

ianna (commented):
@agoose77 - I hope it fixes random test failures.

jpivarski (approved):
Now we don't get to test the "number of threads > 1" cases in the testing suite, but I can see why a global, mutable setting like `ROOT.EnableImplicitMT()` can't be reliably used in some tests and not others. This seems to be the best we can do.

PR #2280, opened 2023-03-03
feat: make `header-only` libraries a CMake project

agoose77:
This PR adds a CMake target to the header-only libraries so that third-party consumers can directly use `add_subdirectory` or `FetchContent`. This should improve the UX of consuming the header-only API, by ensuing that both the include path _and_ the compile flags are appropriately set.

The motivation for bundling the headers separately from the `awkward`/`awkward-cpp` packages is that they don't themselves require a Python interpreter to be useful. Therefore, we want to find a distribution solution that avoids that requirement. @jpivarski and I discussed this, and settled on the default workflow being the use of a CMake helper. I suggested `ExternalProject`, but, upon using it I remembered that it runs at build-time, rather than configure time. `FetchContent` is a configure-time analogue, which plays more nicely with modern CMake practices.

See [the preview](http://preview.awkward-array.org.s3-website.us-east-1.amazonaws.com/agoose77/fix-bundle-headers/user-guide/how-to-use-header-only-layoutbuilder.html#how-to-use-layout-builders) for the user guide which describes this new target. The guide provides a CMake snippet that loads the headers for a given version from the published artefact. This will only work with the _next_ Python release, i.e. 2.1.0.

```cmake
include(FetchContent)

set(AWKWARD_VERSION "v2.1.0")

FetchContent_Declare(
  awkward-headers
  URL      https://github.com/scikit-hep/awkward/releases/download/${AWKWARD_VERSION}/header-only.zip
)
FetchContent_MakeAvailable(awkward-headers)
```

It's possible to add the header-only targets to a CMake application using Git instead of a release artefact. I think the artefact-based solution is better, though, because it doesn't introduce a dependency upon the header locations in our repo (in case people start depending upon this in their own build scripts).

Additionally, this PR adds a note to the user-guide on how to return the C++-allocated arrays to Python using capsules.


agoose77:
@jpivarski this is still a WIP (need to update documentation), but I did realise that we _do_ publish the headers in a `tar.gz` during the `deploy` workflow: https://github.com/scikit-hep/awkward/actions/runs/4196422156

We just need to explicitly attach them to the release artefacts. I'll make a note to update `CONTRIBUTING.md` to reflect this.

agoose77:
@ianna IIRC you mentioned that the header-only library needs to be compiled with C++17. Is that so, i.e is our lower bound for `awkward-cpp` therefore also C++17?

header-only/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+cmake_minimum_required(VERSION 3.15...3.24)
+
+project(
+  awkward-headers
+  LANGUAGES CXX
+  VERSION 1.0.0)
```
agoose77, line 8: We can version these independently.

ianna:
> @ianna IIRC you mentioned that the header-only library needs to be compiled with C++17. Is that so, i.e is our lower bound for `awkward-cpp` therefore also C++17?

No, C++17 is the requirement for the header-only `LayoutBuilder`. The `GrowableBuffer` is fine with C++11. Only the latter is used in the `awkward-cpp`, so keeping the C++11 requirement there is ok, IMHO.

agoose77:
> No, C++17 is the requirement for the header-only LayoutBuilder. The GrowableBuffer is fine with C++11. Only the latter is used in the awkward-cpp, so keeping the C++11 requirement there is ok, IMHO.

Ah, OK! In which case, I've split the header-only library into smaller targets, so that we can set the CXX version for each one; only the `awkward::layout-builder` target needs CXX 17.

header-only/CMakeLists.txt
```diff
@@ -0,0 +1,71 @@
...
+set_target_properties(
+  builder-options growable-buffer
+  PROPERTIES CXX_STANDARD 11
+             CXX_STANDARD_REQUIRED YES
+             CXX_EXTENSIONS NO)
+
+# Build test suite?
+if(BUILD_TESTS)
+  add_subdirectory(tests)
+endif(BUILD_TESTS)
```
ianna, line 54: I tried `cmake tests`:
```bash
(base) yana@iannas-macbook-pro-2 header-only % cmake tests
CMake Warning (dev) in CMakeLists.txt:
  No project() command is present.  The top-level CMakeLists.txt file must
  contain a literal, direct call to the project() command.  Add a line of
  code such as

    project(ProjectName)

  near the top of the file, but after cmake_minimum_required().

  CMake is pretending there is a "project(Project)" command on the first
  line.
This warning is for project developers.  Use -Wno-dev to suppress it.

-- The C compiler identification is AppleClang 13.0.0.13000029
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Warning (dev) in CMakeLists.txt:
  No cmake_minimum_required command is present.  A line of code such as

    cmake_minimum_required(VERSION 3.24)

  should be added at the top of the file.  The version specified may be lower
  if you wish to support older CMake versions for this project.  For more
  information run "cmake --help-policy CMP0000".
This warning is for project developers.  Use -Wno-dev to suppress it.

-- Configuring done
CMake Warning (dev) at CMakeLists.txt:4 (add_executable):
  Policy CMP0063 is not set: Honor visibility properties for all target
  types.  Run "cmake --help-policy CMP0063" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  Target "test_1494-layout-builder" of type "EXECUTABLE" has the following
  visibility properties set for CXX:

    CXX_VISIBILITY_PRESET
    VISIBILITY_INLINES_HIDDEN

  For compatibility CMake is not honoring them for this target.
Call Stack (most recent call first):
  CMakeLists.txt:14 (addtest_nolibs)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:4 (add_executable):
  Policy CMP0063 is not set: Honor visibility properties for all target
  types.  Run "cmake --help-policy CMP0063" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  Target "test_1542-growable-buffer" of type "EXECUTABLE" has the following
  visibility properties set for CXX:

    CXX_VISIBILITY_PRESET
    VISIBILITY_INLINES_HIDDEN

  For compatibility CMake is not honoring them for this target.
Call Stack (most recent call first):
  CMakeLists.txt:15 (addtest_nolibs)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:4 (add_executable):
  Policy CMP0063 is not set: Honor visibility properties for all target
  types.  Run "cmake --help-policy CMP0063" for policy details.  Use the
  cmake_policy command to set the policy and suppress this warning.

  Target "test_1560-builder-options" of type "EXECUTABLE" has the following
  visibility properties set for CXX:

    CXX_VISIBILITY_PRESET
    VISIBILITY_INLINES_HIDDEN

  For compatibility CMake is not honoring them for this target.
Call Stack (most recent call first):
  CMakeLists.txt:16 (addtest_nolibs)
This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:18 (target_link_libraries):
  Policy CMP0028 is not set: Double colon in target name means ALIAS or
  IMPORTED target.  Run "cmake --help-policy CMP0028" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  Target "test_1494-layout-builder" links to:

    awkward::layout-builder

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:19 (target_link_libraries):
  Policy CMP0028 is not set: Double colon in target name means ALIAS or
  IMPORTED target.  Run "cmake --help-policy CMP0028" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  Target "test_1542-growable-buffer" links to:

    awkward::growable-buffer

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

This warning is for project developers.  Use -Wno-dev to suppress it.

CMake Warning (dev) at CMakeLists.txt:20 (target_link_libraries):
  Policy CMP0028 is not set: Double colon in target name means ALIAS or
  IMPORTED target.  Run "cmake --help-policy CMP0028" for policy details.
  Use the cmake_policy command to set the policy and suppress this warning.

  Target "test_1560-builder-options" links to:

    awkward::builder-options

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

This warning is for project developers.  Use -Wno-dev to suppress it.

-- Generating done
-- Build files have been written to: /Users/yana/Projects/PR2280/awkward/header-only
(base) yana@iannas-macbook-pro-2 header-only % ls
CMakeCache.txt		CMakeFiles		CMakeLists.txt		Makefile		README.md		builder-options		cmake_install.cmake	growable-buffer		layout-builder		tests
(base) yana@iannas-macbook-pro-2 header-only % make
[ 16%] Building CXX object CMakeFiles/test_1494-layout-builder.dir/test_1494-layout-builder.o
/Users/yana/Projects/PR2280/awkward/header-only/tests/test_1494-layout-builder.cpp:3:10: fatal error: 'awkward/LayoutBuilder.h' file not found
#include "awkward/LayoutBuilder.h"
         ^~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/test_1494-layout-builder.dir/test_1494-layout-builder.o] Error 1
make[1]: *** [CMakeFiles/test_1494-layout-builder.dir/all] Error 2
make: *** [all] Error 2
(base) yana@iannas-macbook-pro-2 header-only % 

```

docs/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -229,9 +235,56 @@ std::cout << builder.form() << std::endl;
...
+We want NumPy to own the array buffers, so that they get deleted when the Awkward Array goes out of Python scope, not when the LayoutBuilder goes out of C++ scope. For the hand-off, one can allocate memory for those buffers in Python, presumably with `np.empty(nbytes, dtype=np.uint8)` and get `void*` pointers to these buffers by casting the output of `numpy_array.ctypes.data` (pointer as integer). Then we can pass everything over the border from C++ to Python using e.g. `pybind11`'s `py::buffer_protocol` for the buffers.  
+
+Alternatively, the Python _capsule_ system can be used to tie the lifetime of the allocated buffers to the calling Python scope. `pybind11` makes this fairly trivial, and also permits us to invoke Python code _from_ C++. We can use this approach to call `ak.from_buffers` in order to build an `ak.Array`:
+```cpp
+template <typename T>
+py::object snapshot_builder(const T& builder)
```
agoose77, line 243: @ianna I'm assuming we can const-ref the builder, but let me know if I've made an oversight here in terms of best practice!

docs/user-guide/how-to-use-header-only-layoutbuilder.md
```diff
@@ -229,9 +235,56 @@ std::cout << builder.form() << std::endl;
...
+        
+        // Create capsule that frees the allocated data when out of scope
+        py::capsule free_when_done(it.second, [](void *data) {
+            uint8_t* dataPtr = reinterpret_cast<uint8_t*>(data);
+            delete[] dataPtr;
+        });
+
+        // Adopt the memory filled by `to_buffers` as a NumPy array
+        // We only need to return a "buffer" here, but py::array_t let's
+        // us associate a capsule for destruction, so we use that!
```
agoose77:
```suggestion
        // Adopt the memory filled by `to_buffers` as a NumPy array
        // We only need to return a "buffer" here, but py::array_t let's
        // us associate a capsule for destruction, which means that 
        // Python can own this memory. Therefore, we use py::array_t
```

ianna (commented):
@agoose77 - thanks! It looks good. I'm not sure I like the fine granularity for the headers, but it's just my personal view and I'm sure I'll get used to it :-) I tried to build the tests in place - the way I did before - and got lots of warnings. We may want to mention in the guide how to test the header-only installation by building and running the tests locally.

docs/user-guide/how-to-use-header-only-layoutbuilder.md
ianna: Correct - it's mutable :-)

header-only/CMakeLists.txt
agoose77: Thanks for trying this out! 

I've had to change how the tests are run (sorry!), as they're now a component of the header-only project. You can see the change in our CI configuration, but here's the excerpt of interest:

```cmake
cmake \
    -DBUILD_TESTS=ON \
    -B build \
    -S <PATH-TO-HEADER-ONLY>
cmake --build build
```

In this snippet, on my machine the test binaries are then found in `build/tests/`

agoose77: >  I'm not sure I like the fine granularity for the headers, but it's just my personal view and I'm sure I'll get used to it :-) 

Hehe, I'm open to alternative approaches! I settled on this finer-grained solution so that we can specify CXX 14 and CXX 17 for the growable-buffer and layout-builder respectively. I can't (off the top of my head) think of any other way of doing this that let's us just re-use the build configuration between `awkward-cpp` and third party usage.

ianna: Great! Thanks, it works for me.

ianna (approved):
@agoose77 - Looks great, thanks!

header-only/CMakeLists.txt
jpivarski: Independently of `awkward-cpp`, I assume.

Does the version need major, minor, and patch numbers, when the project it's embedded in has much less version-granularity? (Or are the three digits a CMake requirement?)

jpivarski:
> @ianna IIRC you mentioned that the header-only library needs to be compiled with C++17. Is that so, i.e is our lower bound for `awkward-cpp` therefore also C++17?

Our previous motivation for C++11 (and C++14 for Windows because it didn't have a well-defined C++11 target) was to be able to build with the compilers that come with the original `manylinux` images (from back when we were supporting Python 2.7). That constraint has been released by abandoning old Python versions. I _think_ that all of our wheel-building jobs are capable of C++17.

If that's the case, then we can just declare all of Awkward Array's C++ to have a minimum of C++17, just for simplicity. I don't know of any use-cases to motivate old C++ compilers, and the old use-case was our wheel-building workflows.

jpivarski (approved):
> I suggested `ExternalProject`, but, upon using it I remembered that it runs at build-time, rather than configure time. `FetchContent` is a configure-time analogue, which plays more nicely with modern CMake practices.

Okay, I do remember our conversation about `ExternalProject`, but in that conversation I had been assuming that it was configure-time, rather than compile-time, so I agree that this is better.

I see that this PR also includes explicit documentation for passing from C++ to Python, which answers https://github.com/conda-forge/awkward-cpp-feedstock/issues/13#issuecomment-1455519893 in a centralized place. Thanks!

agoose77:
> If that's the case, then we can just declare all of Awkward Array's C++ to have a minimum of C++17, just for simplicity.

OK, if we decide we can move to C++ 17, we can keep the targets as the separation probably makes sense!

Issue #2143, opened 2023-01-19
2.1.0 Release Checklist


agoose77:
Order of operations:

- [x] Last awkward-cpp release for awkward 2.0.x
- [x] Finish #2276, #2269, #2260, #2280
- [x] Last awkward 2.0.x patch release
- [x] Bump NumPy minimum version to 1.17 in both awkward and awkward-cpp # https://github.com/scikit-hep/awkward/pull/2282
- [x] Release 2.1.0

PR #2282, opened 2023-03-04
chore: bump NumPy version

agoose77:
This PR bumps the minimum NumPy version to 1.17.0, in line with #2143 

@jpivarski can you remind me why we settled on 1.17 instead of aligning with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)? I know we made this decision, but can't recall why.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -192,13 +192,12 @@ jobs:
           - "numpy"
         include:
           - python-version: '3.7'
-            numpy-package: "numpy==1.14.5"
+            numpy-package: "numpy==1.17.0"
 
     runs-on: ubuntu-20.04
 
     env:
       PIP_ONLY_BINARY: cmake
-      NUMPY_VERSION: '${{ matrix.numpy-version }}'
 
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
> can you remind me why we settled on 1.17 instead of aligning with [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html)?

I didn't recognize that NEP 29 is a community-wide recommendation, rather than something NumPy is doing and other projects ad-hoc chose to emulate it. But I went back and read it and indeed:

> This is an unusual NEP in that it offers recommendations for community-wide policy and not for changes to NumPy itself. Since a common place for SPEEPs (Scientific Python Ecosystem Enhancement Proposals) does not exist and given NumPy’s central role in the ecosystem, a NEP provides a visible place to document the proposed policy.

This is anticipating the development of [SPECs](https://scientific-python.org/specs/), particularly [SPEC 0](https://scientific-python.org/specs/spec-0000/).

For Awkward 2.1, we should continue with what we said we would do and only bump NumPy up to 1.17. However, let's also plan on adopting either NEP 29 or SPEC 0. Hopefully, they'll be the same recommendation, but if they differ, it depend on when SPEC 0 gets finalized. (For simplicity, to avoid changing our policy twice if we can do that in a good timeframe, like this year.)

Also, don't merge this one yet. I'd like to do a final patch release of 2.0 without this PR, so that this is the only change from the last 2.0 to the first 2.1 (in case it makes somebody's debugging task easier...).

jpivarski:
The pr-on-hold label will only be on this until we're ready for the 2.0 → 2.1 transition, which could be as early as today.

PR #2283, opened 2023-03-06
fix: apply subset of fixes for downstream dask usage

agoose77:
## TL;DR
- Adds `cumsum` to `TypeTracer`
- Removes `to_rectilinear` from nplikes, replaces with `layout.to_backend_array`


src/awkward/operations/ak_concatenate.py
```diff
@@ -240,7 +242,7 @@ def action(inputs, depth, **kwargs):
                     all_flatten.append(f)
 
                 offsets = backend.index_nplike.empty(
-                    len(nextinputs[0]) + 1, dtype=np.int64
+                    nextinputs[0].length + 1, dtype=np.int64
```
jpivarski, line 245: Good catch!

jpivarski (approved):
This looks good, although there's at least one use of `to_rectilinear` in the tests, which is where this is failing right now.

https://github.com/scikit-hep/awkward/blob/6cde4b3fd8f36b51fba202274e56e82cea129e9f/tests/test_1247_numpy_to_rectilinear_ndarray.py#L13

src/awkward/operations/ak_concatenate.py
agoose77: > This looks good, although there's at least one use of `to_rectilinear` in the tests, which is where this is failing right now.
> 
> https://github.com/scikit-hep/awkward/blob/6cde4b3fd8f36b51fba202274e56e82cea129e9f/tests/test_1247_numpy_to_rectilinear_ndarray.py#L13

I could have sworn I ran the test suite locally. Oh, well, thank goodness for CI!

jpivarski: It came up in the Linux-ROOT test; it might be a test that requires ROOT in the specific environment to run. Although I see that another one failed, too, so maybe not.

jpivarski:
In c569a71813ea27a0faae2031797733ec2dbad3d9, I simply removed the failing test. It was testing an aspect of `to_rectilinear`, but since the function no longer exists, I'd say that the property we intended to test is now obsolete.

PR #2285, opened 2023-03-06
feat: validate `axis` in L1

agoose77:
Long-term, we want to take advantage of L2 being developer-focused to simplify the API signatures. This PR ensures that L1 checks that `axis` is `None` or int-like. 

In general, we don't need to validate every argument at L1 - some are validated in L2. But anywhere that we expect L2 to be stricter than L1, we should do this.


jpivarski (approved):
This looks good! I didn't see any L2 functions that take `axis` as an argument (plenty of L3/L4 functions), so all public functions have the same regularization applied.

It's ready to merge.

PR #2288, opened 2023-03-07
feat: add examples to header-only module

agoose77:
This PR adds some examples to the header-only subdirectory to demonstrate usage.

- [x] Update `copy-cpp-headers` to avoid copying examples into `awkward-cpp` / `_connect`
- [x] Implement Cython example
- [x] Implement pybind11 example

Anyone is welcome to join in on this PR — I'll make a note of what I'm doing in the description. Assign yourself if you're working on it.


agoose77:
@jpivarski thought you *might* like to work on the the Cython example. Not required, but in case you want to do some awkward coding. Anyone's welcome, though!

agoose77:
@jpivarski requesting a review to see if this is heading in the right direction. It's been forever since I used Cython (honestly, I avoid it like the plague), so any extra eyes are welcome.

header-only/examples/cython/_demo.pyx
```diff
@@ -0,0 +1,26 @@
...
+
+    buffers = {}
+    # Convert from raw buffers to NumPy arrays
+    it = array_buffers.buffers.begin();
+    while (it != array_buffers.buffers.end()):
+        # Pull out name, buffer, and lookup length of buffer
+        name = deref(it).first
+        buffer = deref(it).second
+        nbytes = array_buffers.buffer_nbytes[name]
+        inc(it)
```
jpivarski: I don't think it's necessary to use a `while` loop on a C++ iterator, or to have to use `deref` and `inc` functions explicitly. [This StackOverflow answer](https://stackoverflow.com/a/33310456/1623645) has a natural-looking for loop over `std::map` where one of the other answers has a `while` loop like the one you used. Presumably, they're interchangeable.

jpivarski (commented):
> @jpivarski requesting a review to see if this is heading in the right direction. It's been forever since I used Cython (honestly, I avoid it like the plague), so any extra eyes are welcome.

I think this is right. Nothing jumps out at me as problematic.

I've only ever used Cython for a few in-Jupyter examples, so my knowledge of it doesn't run very deep. If you're worried about it working in the test but not being idiomatic or being an antipattern for performance reasons, `cythonize` has an [annotate](https://cython.readthedocs.io/en/latest/src/tutorial/cython_tutorial.html) argument that produces a visualization of the Python content hidden in each statement. (Although even with that, it took me years to work the inefficiencies out of my demo.)

I'll review it again when it's closer to being done. I think maybe I've found a spot that can be made easier to read (important in a demo):

agoose77:
I'm not sure whether to modify `copy-cpp-headers.py` to avoid copying the examples. If we copy the examples into awkward-cpp, then every change to these examples will require a new build of awkward-cpp before awkward can be released. I don't know how stable these headers will be, to determine whether this is a reasonable proposition. Meanwhile, it might be sensible to bundle the examples along with everything else, so that the `awkward-headers` within awkward-cpp is "complete".

header-only/examples/cython/_demo.pyx
agoose77: I had a hunch there must be something like this natively available. This is much nicer!

header-only/examples/cython/_demo.pyx
```diff
@@ -0,0 +1,24 @@
...
+    buffers = {}
+    # Convert from raw buffers to NumPy arrays
+    for it in array_buffers.buffers:
+        # Pull out name, buffer, and lookup length of buffer
+        name = it.first
+        buffer = it.second
+        nbytes = array_buffers.buffer_nbytes[name]
+
+        # Store buffer in Python dict
+        buffers[name.decode('UTF-8')] = np.asarray(<np.uint8_t[:nbytes]>buffer)
```
ianna: @agoose77 - does it transfer ownership of the C++ `ArrayBuffers` buffers?

ianna (commented):
@agoose77 -  very nice! I have one question - should have checked it myself ;-)

header-only/examples/cython/_demo.pyx
agoose77: Oooh :facepalm:!! Not sure how _that_ got lost between writing the `pybind11` example and the `cython` one. I love the review process :sweat_smile:

ianna: or `PyArray_ENABLEFLAGS(buffers[...], np.NPY_OWNDATA)`?

agoose77: > or `PyArray_ENABLEFLAGS(buffers[...], np.NPY_OWNDATA)`?

So, I'm not an expert here, but what I've read so far implies that NumPy would use its own allocator to free the memory if one uses `NPY_OWNDATA`. I *think* they map onto `malloc` and `free`, but my gut is that being explicit in using the cython view that definitely calls `free` might be safer?

ianna: > > or `PyArray_ENABLEFLAGS(buffers[...], np.NPY_OWNDATA)`?
> 
> So, I'm not an expert here, but what I've read so far implies that NumPy would use its own allocator to free the memory if one uses `NPY_OWNDATA`. I _think_ they map onto `malloc` and `free`, but my gut is that being explicit in using the cython view that definitely calls `free` might be safer?

Looking into the cython code: it looks like it also uses `malloc` and `free`. Perhaps, allocating the `ArrayBuffers` buffers should be also done with `malloc`. `new` allocated memory is on a `Free Store` and `malloc` one is on the `Heap` - technically it could be the same.

agoose77: Yes, that's a good point. I was willing to sweep _that_ one under the rug, but we should be proper here!

jpivarski (approved):
These will be very helpful, thanks.

Merge when you're ready.

agoose77:
Thanks @ianna and @jpivarski for the reviews, very helpful!

PR #2293, opened 2023-03-08
fix: support typetracer in `ak.unflatten`

agoose77:
This PR fixes assumptions of length and known values that dask-awkward's test-suite fails.

### DIFF
--- a/src/awkward/operations/ak_unflatten.py
+++ b/src/awkward/operations/ak_unflatten.py
@@ -94,7 +94,7 @@ def _impl(array, counts, axis, highlevel, behavior):
 
     if ak._util.is_integer_like(counts):
         # Regularize unknown values to unknown lengths
-        if not backend.index_nplike.known_data:
+        if is_unknown_scalar(counts) or counts is unknown_length:
             counts = unknown_length
         else:
             counts = int(counts)
--- a/src/awkward/operations/ak_unflatten.py
+++ b/src/awkward/operations/ak_unflatten.py
@@ -1,9 +1,9 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
-import numbers
 
 import awkward as ak
 from awkward._nplikes.numpylike import NumpyMetadata
+from awkward._nplikes.shape import unknown_length
 
 np = NumpyMetadata.instance()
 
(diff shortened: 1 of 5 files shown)


src/awkward/operations/ak_unflatten.py
```diff
@@ -116,18 +121,22 @@ def _impl(array, counts, axis, highlevel, behavior):
         if counts.ndim != 1:
             raise ak._errors.wrap_error(ValueError("counts must be one-dimensional"))
 
-        if not issubclass(counts.dtype.type, np.integer):
+        if not np.issubdtype(counts.dtype, np.integer):
             raise ak._errors.wrap_error(ValueError("counts must be integers"))
 
-        current_offsets = backend.index_nplike.empty(len(counts) + 1, dtype=np.int64)
+        current_offsets = backend.index_nplike.empty(counts.size + 1, dtype=np.int64)
```
jpivarski, line 128: So... `counts` was originally obtained via `to_layout` (layouts have `.length`, not `.size`), but then it was replaced by `to_backend_array`. I assume that means it could be NumPy, CuPy, a JAX DeviceArray, or a TypeTracerArray. All of these have a `.size`?

For a NumPy array, the `.size` is the product of `shape`, not its first element. Is there any way that the NumpyArray might be multidimensional? I suppose not, if it's a valid `counts`.

jpivarski (approved):
We talked about this today, and I thought that if threading a typetracer through the whole function is difficult, it could be a candidate for a specialized path. But it looks like you did it!

I don't see any tests, though. (Tests for the specific case of `ak.unflatten` _with_ typetracers.) Do they come up in dask-awkward?

src/awkward/operations/ak_unflatten.py
agoose77: Yes, a previous PR explicitly ensures that `counts` is a backend array, all of which have `.size`. We have an explicit check on `counts.ndim` to ensure that it's 1D!

agoose77: > I don't see any tests, though. (Tests for the specific case of ak.unflatten with typetracers.) Do they come up in dask-awkward?

Dask-awkward does test this, because it brought this to my attention from their test suite. However, we should really test it for ourselves (I'm being lazy :face_with_head_bandage:). I'll add some tests before we merge.

jpivarski (commented):
:heavy_check_mark: on tests/test_2293_unflatten_typetracer.py

agoose77:
Ah, I noticed a final corner-case: `counts` can be known for typetracer backends, and I'm trying to preserve that kind of information where possible. I modified the PR to normalise `counts` if its value is not known, rather than testing if the backend has `known_data`.

PR #2294, opened 2023-03-09
fix: view for unknown scalar

agoose77:
This small PR fixes the assumption in `TypeTracerArray.view` that an array's shape always contains at least one element.


jpivarski (approved):
Sounds like something one would want to be able to do.

PR #2295, opened 2023-03-09
fix: `ak.cartesian` for typetracer

agoose77:
@jpivarski this is a tiny PR that follows #2329. This was open first, and I thought it would be longer than it has transpired to be.

### DIFF
--- a/src/awkward/operations/ak_cartesian.py
+++ b/src/awkward/operations/ak_cartesian.py
@@ -304,7 +304,10 @@ def _impl(arrays, axis, nested, parameters, with_name, highlevel, behavior):
         indexes = [
             ak.index.Index64(backend.index_nplike.reshape(x, (-1,)))
             for x in backend.index_nplike.meshgrid(
-                *[backend.index_nplike.arange(len(x), dtype=np.int64) for x in layouts],
+                *[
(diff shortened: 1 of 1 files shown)


src/awkward/operations/ak_cartesian.py
```diff
@@ -398,10 +402,14 @@ def apply(x, i):
...
+                if all(
+                    x.length is not unknown_length and x.length == 0 for x in inputs
+                ):
                     inputs = [
                         x.content
-                        if isinstance(x, ak.contents.RegularArray) and x.size == 1
+                        if isinstance(x, ak.contents.RegularArray)
+                        and x.size is not unknown_length
+                        and x.size == 1
                         else x
```
agoose77: @jpivarski / @ioanaif do you remember what the purpose of this step is? I can spend some time figuring it out, but if you can think off the top of your head, it would be helpful!

I've "fixed" it, but I have no real understanding of the intention here, so I don't feel comfortable in considering the PR done! :)

agoose77: Note to self: make this blow up to find a reproducer & counterexample. See what typetracer does thereafter.

jpivarski (approved):

PR #2297, opened 2023-03-09
fix: common backend in functions accepting multiple arrays

agoose77:
Similarly to #2175, we want to coerce arguments of functions that accept multiple arrays to the same backend. This should only succeed if the coercion is zero copy, i.e. typetracer + backend, or backend + backend.


jpivarski (approved):
As I understand it, this is to ensure that typetracer + any other backend = typetracer and any other mixture of backends raises an error.

agoose77:
Yes, that's it! It's possible that should we ever have a same-device backend that we could support mixed backends distinct from typetracer, but that's not something one can currently do and I don't think it makes much sense.

PR #2301, opened 2023-03-09
ci: fix parsing of docs version

agoose77:
The [2.1 docs deployment](https://github.com/scikit-hep/awkward/actions/runs/4359154751/jobs/7620722224) pushed the docs to the `/doc/v2.1/` subdirectory instead of `/doc/2.1/`. This PR fixes the version parsing logic to correct this for future major/minor releases.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -314,7 +314,7 @@ jobs:
         id: sync-stable
         run: |
           # Take only leading version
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Right—before [v1.9.0rc12](https://github.com/scikit-hep/awkward/releases/tag/v1.9.0rc12), tags didn't have a `v` in them; now they do.

Looking at the tag list reminds me that we're missing some recent `awkward-cpp-*` tags. I had forgotten that they're not automatic.

agoose77:
@jpivarski I'll update the contributing-guide too!

PR #2305, opened 2023-03-10
fix: only convert NEP18 arguments to layouts if required

### LINKED ISSUE #2303: NEP18-dispatched functions raise an error if any arguments are non-native endian arrays
### Version of Awkward Array

HEAD

We can pass, for example, an Awkward Array and a NumPy array to `np.array_equal` (without having an explicit implementation of `ak.array_equal`) because we're overloading it with NEP-18 (`__array_function__`).

```python
>>> np.array_equal(ak.Array(np.arange(10, dtype="<u4")), np.arange(10, dtype="<u4"))
True
```

This function is supposed to be insensitive to the dtypes of its arguments.

```python
>>> np.array_equal(ak.Array(np.arange(10, dtype="<u4")), np.arange(10, dtype="<f4"))
True
```

Including endianness.

```
>>> np.array_equal(np.arange(10, dtype="<u4"), np.arange(10, dtype=">u4"))
True
```

But this fails:

```python
>>> np.array_equal(ak.Array(np.arange(10, dtype="<u4")), np.arange(10, dtype=">u4"))
  File "<__array_function__ internals>", line 180, in array_equal
  File "/home/jpivarski/irishep/awkward/src/awkward/highlevel.py", line 1392, in __array_function__
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 73, in array_function
    return _array_function_no_impl(func, types, args, kwargs, behavior)
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 59, in _array_function_no_impl
    rectilinear_args = tuple(_to_rectilinear(x) for x in args)
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 59, in <genexpr>
    rectilinear_args = tuple(_to_rectilinear(x) for x in args)
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 42, in _to_rectilinear
    return ak.operations.to_numpy(arg)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_to_numpy.py", line 41, in to_numpy
    return _impl(array, allow_missing)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_to_numpy.py", line 48, in _impl
    layout = ak.to_layout(array, allow_record=False)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_to_layout.py", line 51, in to_layout
    return _impl(array, allow_record, allow_other, regulararray=regulararray)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_to_layout.py", line 84, in _impl
    return ak.operations.from_numpy(
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_from_numpy.py", line 50, in from_numpy
    return ak._util.from_arraylib(
  File "/home/jpivarski/irishep/awkward/src/awkward/_util.py", line 709, in from_arraylib
    layout = recurse(array, mask)
  File "/home/jpivarski/irishep/awkward/src/awkward/_util.py", line 667, in recurse
    data = ak.contents.NumpyArray(array)
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/numpyarray.py", line 95, in __init__
    ak.types.numpytype.dtype_to_primitive(self._data.dtype)
  File "/home/jpivarski/irishep/awkward/src/awkward/types/numpytype.py", line 49, in dtype_to_primitive
    raise ak._errors.wrap_error(
TypeError: while calling

    ak.to_numpy(
        array = numpy.ndarray([0 1 2 ... 7 8 9])
        allow_missing = True
    )

Error details: unsupported dtype: dtype('>u4'). Must be one of

    bool, int8, uint8, int16, uint16, int32, uint32, int64, uint64, float32, float64, complex64, complex128, datetime64, timedelta64, float16, float128, complex256

or a datetime64/timedelta64 with units (e.g. 'datetime64[15us]')
```

because somewhere in the implementation of `ak.Array.__array_function__`, we try to convert the non-native endian NumPy array into an Awkward Array, and we don't support that.

The implementation will have to become insensitive to this, possibly by ensuring that all non-Awkward arguments in `__array_function__` are converted to native endian, or maybe by taking a new policy that non-native endian NumPy arrays are copy-converted to native endian in `to_layout` and/or `from_numpy`. That would introduce a hidden cost for non-native endian arrays, but I think that would be acceptable. It's not like GPU copying, where users really want to be aware of this; endianness is a nuisance, rather than something people want to track.

agoose77:
Issue #2303 raises a couple of important points;
- We don't need to build a layout only to immediately convert it to a backend array
- We should not convert to NumPy eagerly; it's possible to use NEP18 on cuda `cp.ndarray,` the same should hold for `ak.Array` with `backend="cuda"`.

This PR fixes #2303 by 
- Only converting to layouts if we identify a backend, and the object is not a backend array
- Convert all arguments to a common backend

### DIFF
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -1,11 +1,15 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+import collections
 import functools
 import inspect
+from collections.abc import Iterable
+from itertools import chain
 
 import numpy
 
 import awkward as ak
-from awkward._nplikes import nplike_of
+from awkward._backends import Backend, backend_of, common_backend
+from awkward._nplikes import to_nplike
 from awkward._util import numpy_at_least
 from awkward.contents.numpyarray import NumpyArray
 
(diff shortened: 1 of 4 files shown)


src/awkward/_connect/numpy.py
```diff
@@ -35,15 +40,47 @@ def convert_to_array(layout, args, kwargs):
...
+    stack = collections.deque(args)
+    while stack:
+        arg = stack.popleft()
+        # If the argument declares a backend, easy!
+        backend = backend_of(arg, default=None)
+        if backend is not None:
+            yield backend
+        # Is this object something we already associate with an nplike?
+        elif isinstance(arg, (tuple, list)):
+            stack.extend(arg)
```
agoose77, line 59: I don't love that we have to do two passes over the arguments list here. There are other ways to do this, but this seems like the most elegant approach. My feeling is that we want to provide native implementations for these functions if they're commonly needed, and further that we are mainly concerned with array-bound performance, so the fact that we visit the args twice is not the end of the world.

src/awkward/_nplikes/__init__.py
```diff
@@ -73,6 +75,10 @@ def nplike_of(*arrays, default: D = _UNSET) -> NumpyLike | D:
                 if cls.is_own_array(array):
                     nplikes.add(cls.instance())
                     break
+            else:
+                # Explicitly test non-layout ArrayBuilder types
+                if isinstance(array, (_ext.ArrayBuilder, ak.highlevel.ArrayBuilder)):
+                    nplikes.add(Numpy.instance())
```
agoose77, line 81: I don't love that we need to import layout-specific types here. I'd prefer to do this via e.g. attribute inspection. However, the proper way would be to define a `backend` property on the high & lowlevel array-builder types, which would expose `backend` to the L1 API.

agoose77:
@jpivarski I'm not super happy with this PR. I would prefer there to be fewer calls to `backend_of` et al. That said, this is elegant enough that I think it's "fine". Let me know your thoughts?

src/awkward/_connect/numpy.py
jpivarski: I think that doing two passes over all of the arguments is fine. In fact, I probably wouldn't have bothered with `collections.deque` to make a stack, since the built-in list is already stack-like, with `append` and `pop`. Oh, I see: `popleft`. Well, built-in list is still good for that.

(The number of arguments in a function call is never very large.)

src/awkward/_nplikes/__init__.py
jpivarski: ArrayBuilder is not a layout, and anything that an ArrayBuilder builds is naturally on the CPU, ready for `Numpy.instance()`. This doesn't seem like it's overreaching to me.

jpivarski (approved):
I think this is a fine solution to the problem.

The codecov test is holding up merging; updating to main will probably jostle it enough.

aburke1605:
Hi all, tracking issue https://github.com/scikit-hep/awkward/issues/2303, is there a rough time estimate of a new release that will contain this fix? \
Thanks

agoose77:
Hi @aburke1605, thanks for the ping! I've been away this week, but I'll pick this up again next week and should have it fixed then. We will then be due a new Awkward release, I think!

PR #2306, opened 2023-03-13
feat: use cppyy for JIT

ianna:


tests/test_2306_cppyy_git.py
```diff
@@ -0,0 +1,202 @@
...
+        return out;
+    }
+    """
+
+    cppyy.set_debug()
+    cppyy.cppdef(source_code_cpp)
+
+    # FIXME: a tuple of parameters: move it to __cpptype__??
+    def convert(array):
+        return (0, len(array), 0, array._lookup.arrayptrs, 0)
```
ianna: These are the parameters needed to generate this array dataset. It corresponds to a `generator.dataset()` output: `awkward::ListArray_x9C5rj3ig8(0, length, 0, reinterpret_cast<ssize_t*>(ptrs), 0)`

ianna (commented):
@jpivarski - I guess, this `convert` function can be our dunder/magic method that @wlav can call when cppyy sees an `awkward.Array`.

tests/test_2306_cppyy_git.py
ianna: then it's simply:
```python
out = cppyy.gbl.go_fast_cpp_2[f"awkward::{array.cpptype}"](convert(array))
```

tests/test_2306_cppyy_git.py
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        return out;
+    }
+    """
+
+    cppyy.cppdef(source_code_cpp)
+
+    out = cppyy.gbl.go_fast_cpp_2[array.cpptype](array.__castcpp__())
+    assert out == ak.sum(array["y"])
```
ianna: `array.__castcpp__()` returns `array.cpppars` - a Python tuple of parameters that are passed to a C++ `array.cpptype` constructor.

ianna (commented):
@jpivarski - please, have a look.

src/awkward/highlevel.py
```diff
@@ -1460,6 +1462,55 @@ def __bool__(self):
...
+    @property
+    def cpppars(self):
+        """
+        The C++ dataset parameters needed to construct the C++ type of
+        this Array when it is used in cppyy.
+
+            cpppars (None or tuple): Generated on demand when the Array needs to be passed
+                to a C++ function defined by cppyy. The C++ type is defined
+                in #ak.Array.cpptype.
+        """
```
jpivarski: As I understand it, this is temporary; it is a function that will be taken over by `__castcpp__` when cppyy supports it and then this property will be removed from public view. Right?

(If that's true, could it be hidden with an underscore to express our intent to eventually remove it from public view? Actually, this could just be `__castcpp__`, called explicitly for now, but it will be implicit in the future. That would communicate better to Wim what we want the API to look like.)

src/awkward/highlevel.py
```diff
@@ -1460,6 +1462,55 @@ def __bool__(self):
...
+        if self._cpptype is None:
+            self._cpptype = self.cpptype
+
+        if self._cpppars is None:
+            self._cpppars = (0, len(self), 0, self._lookup.arrayptrs, 0)
+
+        return self._cpppars
```
agoose77: I would be tempted _not_ to expose this property unless it's highly likely that the end user would need to do this? I think it makes sense for `numba_type` to be exposed, as Numba allows users to work with type objects. But, my guess here is that this is a cppyy-only feature?

tests/test_2306_cppyy_git.py
jpivarski: I just deleted some comments I was making on the first two tests because this third example illustrates the kind of interface we want to provide (except that cppyy should automatically convert `array` into `array.__castcpp__()`).

We don't want to make users write f-strings because C++ uses curly brackets a lot, and they become `{{` and `}}`. Besides being confusing or ugly, the doubling of curly brackets makes it much more labor-intensive to copy C++ code from an f-string to a non-f-string (or C++ source file). Also, the error messages from doing the wrong thing would be surprising. It was fine for us to use in our internal code for RDataFrame, but we shouldn't require users to write that. Fortunately, C++ provides a solution: templates!

jpivarski (approved):
This is looking good so far!

Don't worry about any copy-by-value of data structures smaller than, say, 64 bytes. The 2-value tuple (`struct { ssize_t length; ssize_t* ptr }`, 16 bytes) is fine to copy, and the small ArrayViews are even bigger (32 bytes, I think). Also, the hand-off between Python and C++ has a lot more overhead than copying an object.

Some questions:

  * Does this still work if a function takes two array arguments? (That's something that can't be done in RDataFrame, a motivation for this more direct kind of interface.) The user-written template should have two template parameters.
  * If a function is called on the same array twice, does it get recompiled? Hopefully not.
  * If a function is called on the same Form of array twice, does it get recompiled? Hopefully not.
  * Will array return values be another PR? As in Numba, you'd only be able to return a direct view of part of the original array that was passed in, not an entirely new array.[^1] (There's LayoutBuilder for that.) The returned view needs to pass ownership, just as it does in Numba. This is the reason that all ArrayViews in Numba carry a `PyObject*`, and that has to be passed in, a third argument of `cpppars`, along with `length` and `ptr`.

[^1]: One use-case for being able to return a part of the original array is to be able to write a function that iterates through an array and stops and returns a record of interest.

src/awkward/highlevel.py
ianna: Yes, I think, there is a meeting tomorrow. I'll bring it up. Thanks.

src/awkward/highlevel.py
ianna: Thanks, @agoose77! I agree, it is temporary - it should all go to a magic/dunder function that will be called by `cppyy`. This would be its new feature :-)

wlav:
For consistency, I ended up calling the implicit method `__cast_cpp__`, but that's open to change. An example of use, that we can use to flesh out the details, is here: https://github.com/wlav/cppyy/blob/master/test/test_advancedcpp.py#L901

This is in repo master. I cut 3.0.0 yesterday, b/c Windows and now also Mac no longer support Clang9, but it's been a bit of a rush job and I've already found some issues, so I hope to cut 3.0.1 soon.

src/awkward/highlevel.py
ianna: done

src/awkward/highlevel.py
ianna: done

ianna:
> For consistency, I ended up calling the implicit method `__cast_cpp__`, but that's open to change. An example of use, that we can use to flesh out the details, is here: https://github.com/wlav/cppyy/blob/master/test/test_advancedcpp.py#L901
> 
> This is in repo master. I cut 3.0.0 yesterday, b/c Windows and now also Mac no longer support Clang9, but it's been a bit of a rush job and I've already found some issues, so I hope to cut 3.0.1 soon.

Thanks, @wlav ! Everything works with `cppyy` built from source. 

@jpivarski - I've restricted the `cppyy` version to `3.0.1` - the minimum required for Awkward-cppyy GIT. I will write documentation tomorrow. Unless you'd like me to add more tests to this PR it can go in.

@agoose77 - please, have a look if the comments are understandable. Thanks!

tests/test_2306_cppyy_git.py
ianna: cppyy automatically converts array into its C++ instance.

ianna (commented):
@jpivarski - everything works as expected!

src/awkward/cppyy.py
```diff
@@ -0,0 +1,30 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import awkward as ak
+
+_has_checked_version = False
+
+
+def register_and_check(array=None, cpp_type=None):
```
jpivarski: This doesn't actually use `array` or `cpp_type` (nor should it).

```suggestion
def register_and_check():
```

src/awkward/cppyy.py
```diff
@@ -0,0 +1,30 @@
...
+        import cppyy
+    except ImportError as err:
+        raise ImportError(
+            """install the 'cppyy' package with:
+
+pip install cppyy
+
+or
+
+conda install -c conda-forge cppyy"""
```
jpivarski: Users need to be warned. If they just try this and have ROOT installed, it would fail somehow, either in the installation phase or when trying to import it.

```suggestion
            """install the 'cppyy' package with:

pip install cppyy

or

conda install -c conda-forge cppyy

Note that this must be in a different venv or conda environment from ROOT, if you have installed ROOT.
"""
```

tests/test_2306_cppyy_git.py
```diff
@@ -0,0 +1,118 @@
...
+            }
+        }
+
+        return out;
+    }
+    """
+
+    cppyy.cppdef(source_code_cpp)
+
+    out = cppyy.gbl.go_fast_cpp_2[array.cpptype](array)
```
jpivarski, line 117: This is great! The fact that users can write functions with this interface really streamlines it. Thanks, @wlav, for cppyy 3.0.1!

I still have the question (for @ianna) of whether the same array twice or a different array with the same `Form` (and therefore the same `array.cpptype`, including the classes that get defined in the evaluation of that property) would initiate a new compilation. Another call of this function with the same-typed (same `Form`) array _should_ be able to reuse precompiled code, and that becomes important for a Dask worker evaluating partitions of an array.

jpivarski (approved):
After committing the suggested changes, I think it's ready to go.

Yes, documentation next would be great, thanks!

tests/test_2306_cppyy_git.py
ianna: > This is great! The fact that users can write functions with this interface really streamlines it. Thanks, @wlav, for cppyy 3.0.1!
> 
> I still have the question (for @ianna) of whether the same array twice or a different array with the same `Form` (and therefore the same `array.cpptype`, including the classes that get defined in the evaluation of that property) would initiate a new compilation. Another call of this function with the same-typed (same `Form`) array _should_ be able to reuse precompiled code, and that becomes important for a Dask worker evaluating partitions of an array.

@jpivarski  - by default the generated C++ code is cached and compiled only once - thanks to you - yes, you've thought about it before and implemented it here:
https://github.com/scikit-hep/awkward/blob/main/src/awkward/_connect/cling.py#L47#L57

agoose77:
@ianna cool! This will be _so_ useful going forward :sunglasses: 

Sorry only get these comments in after merging, I was eating!

Just to check — this is a pretty niche method; most Array's won't be using cppyy. I wonder whether we should exclusively move this to an `ak.` function, so that we don't add any more methods to `Array`'s namespace? I'd be similiarly tempted to do the same for `numba_type` in the long run. @jpivarski you might also have thoughts on this proposal.

Secondly, could we rename this to `cpp_type`, to mirror `numba_type`?

I can make a PR if people agree on any of these suggestions.

jpivarski:
@agoose77 Actually, I recommended the public property in `ak.Array`'s namespace, since `type` and `numbatype` were already there—although I had missed the fact that the Numba one now has an underscore. Moving it from `cpptype` to `cpp_type` before the next release is a good idea.

In general, we try to keep the `ak.Array` namespace available for subclass method and record-field names, but L1 types are the sort of things that you'd want to have quick access to. Also, the one that is using the best real estate is `type`: it's quite likely that a record array would have a field named "type", and now it will have to be accessed with square brackets. However, `type` is also the most prominent (least-niche) one.

Although the fraction of physicists using cppyy directly may be small because of the interference of ROOT[^1], there's no such issue for general-science use-cases. Depending on how we manage to promote it, this interface may not be so niche. Most of the general-science users of Awkward so far have found the Numba bridge to be an essential feature, and there are things you can't do with Numba, such as connect to libraries that use C++ types in the interfaces. (Also, dealing with specialized types in Numba, such as convincing it to use 32-bit floats, is cumbersome.)

If you want to add the underscore, so that we have `numba_type` and `cpp_type`, please do. @ianna will be documenting next, so it's just a matter of using the new name.

[^1]: As a reminder, I'd like to try to take this up. Most PyROOT users don't know that the library they're using is called cppyy; if one module name has to move, it seems reasonable to me that it be the one in ROOT. Even changing it to `ROOT.cppyy` would solve the issue, and it's more tidy for a singly installable package to have only one top-level module name, anyway. I should ask in a ROOT meeting if that's something they're open to, even as an external PR.

agoose77:
> @agoose77 Actually, I recommended the public property in ak.Array's namespace, since type and numbatype were already there

Fine! For posterity, my motivations to remove this from the namespace are:
- it's niche; most users aren't likely to use cppyy, or at least, aren't likely to be calling `.cpp_type` very often. `ak.cpp_type(arr)` is not much extra typing
- it sets a precedent for future decisions

And probably some other ones ... can't remember!

I consider, however, that we're probably already on the same page; in your view, the pros outweigh the cons? 

As such, I'll make a PR just to rename the attribute! :)

Thanks team.

ianna:
@agoose77 - I think, I prefer a method because we have to keep a `lookup` lifetime connected to an `array` lifetime.

agoose77:
> @agoose77 - I think, I prefer a method because we have to keep a lookup lifetime connected to an array lifetime.

Hmm, yes, good point. In this case the solution would be to only hide the public `cpp_type` getter; i.e. only have `ak.Array._cpp_type`. I like this less, though, because there's no clear suggestion that this is different to the existing private variable (which would have to be renamed). 

We have a similar L3-L4 problem with contents, but in that case we have a protocol system (`ak._do`) that makes it clear which private methods are truly _private_.

Given how niche a name `cpp_type` is (whoever is going to want to use that for an analysis?), I think making this a public method is the right call.

jpivarski:
I forgot that it also has to change the state of the `ak.Array` (for the `lookup` lifetime). The same is true of `numba_type`.

If you remember, I kind-of also wanted to get rid of the `ak.type(...)` function and leave only the property. It's not much more _typing_ (with a keyboard), it's just unnatural when you're working interactively. It's often the case that you've built up an expression and then think, "Wait—I should check to see what type this is." Doing that by adding "`.type`" at the end is more natural than having to wrap the whole expression in parentheses to make the function call. Right after checking its type, the next thing you'd likely do is remove the "`.type`" and get the array itself.

This is the same argument for having a `to_list()` (and `tolist()`) method, rather than relying only on the `ak.to_list(...)` function. And the `show()` method doesn't even have a functional counterpart. This argument doesn't apply to everything, but there are a few things you want an object oriented interface to when working interactively. Types are basic, and even though `cpp_type` and `numba_type` are less used, it makes sense, organizationally, to find them in the same place as `type`.

agoose77:
> Doing that by adding ".type" at the end is more natural than having to wrap the whole expression in parentheses to make the function call. 

Agreed. In the phase space of "how convenient would this be — how often would users do this — how valuable a name is this", `.type` is fairly far from the origin. Meanwhile, `.numba_type` is _much_ less frequently used, even in analyses with Numba, yet it is also less useful as a name for users to overload. 

I think we will need to make decisions that appear inconsistent if you look at only one of these axes, but it's the holistic picture that we should be considering. On balance, the convenience of `self._cpp_type = ...` is a deciding factor for me here! :)

PR #2307, opened 2023-03-13
feat: new nplike finding mechanism

agoose77:
## TL;DR
This PR replaces a hard-coded loop-based search with a faster per-type lookup for each nplike.

### Context
We have an `nplike_of` L3 function in Awkward so that we can determine the most suitable array library among a set of complex input objects. This mechanism needs to extend beyond identifying the array library for the array type to layouts, builders, and potentially other objects.

Our existing mechanism for finding the appropriate nplike for a given object is flawed, due to the following points:
1. Our L1 API should not expose `backend` or `nplike`
2. Our L2 APIs cannot all be made homogeneous, e.g. `_ext.ArrayBuilder` should not have a backend (it has no concept of such a thing)
3. Each call to `nplike_for` performs a non-cached lookup of several attributes. For array types, it loops over each nplike
4. The set of nplikes is hard-coded

Extending `nplike_for` to include `ArrayBuilder` would require more customisation of this logic. At the same time, PRs like #2305 would benefit from the search for an nplike being as fast as possible.

### Details
In general, we can do better by generalising the search mechanism. This PR makes such a change, by introducing a type-cached "finder" system, that allows us to register direct nplike lookups for each type. When there is a cache miss, the system walks over the list of finders until a suitable finder is found, and memoizes the result under the type of the object.

There is now a `register_nplike_finder_factory` function that lets the caller register a finder factory that returns a `NumpyLikeFinder`:
```python
@register_nplike_finder_factory
def finder(cls) -> NumpyLikeFinder:
    if issubclass(cls, SomeType):
        return finder_for_SomeType
```
The `NumpyLikeFinder` is responsible for taking an instance of the registered type and returning its associated nplike:
```python
def finder_for_SomeType(obj: SomeType) -> NumpyLike:
    return Numpy.instance()
```

### DIFF
--- a/src/awkward/_nplikes/__init__.py
+++ b/src/awkward/_nplikes/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 from collections.abc import Collection
 
 import awkward as ak
+from awkward._nplikes.finder import find_nplike_for
 from awkward.typing import TYPE_CHECKING, TypeVar
 
 if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


agoose77:
Closely related to this is the question of where the boundary between `nplike` and `backend` begins.

In writing this PR, I think it's actually fairly clear; there should be a `nplike_of(array)` function that lazily evaluates `nplike.is_own_array_type`, and a `backend_of` function that does what the finder system here does.

I'll therefore changes this PR to reflect that separation of concerns.

agoose77:
Closing this in favour of a similar PR.

Issue #567, opened 2020-12-03
Memory leak in ArrayBuilder/GrowableBuffer


jpivarski:
My response to @sbuse's report on [Scikit-HEP/awkward-array](https://gitter.im/Scikit-HEP/awkward-array) (Gitter).

I did a dirty but conclusive-enough experiment (watching htop while running commands in a Zoom meeting). The following consistently increased memory linearly: 100 MB in 80 seconds.

```python
% python -i -c 'import awkward as ak; import numpy as np; import gc'
>>> for i in range(1000000):
...   a = ak.ArrayBuilder()
...   a.integer(i)
...   del a
...   tmp = gc.collect()
... 
```

and the following did not increase by even 10 MB in 80 seconds (where 10 MB is the level of noise—other applications allocating and freeing memory on my system).

```python
% python -i -c 'import awkward as ak; import numpy as np; import gc'
>>> for i in range(1000000):
...   a = np.array([i])
...   b = ak.Array(a)
...   del a
...   del b
...   tmp = gc.collect()
... 
```

The first is a simplified version of yours: your example creates arrays from Python data, which internally invoke the ArrayBuilder. The second makes Awkward Arrays by wrapping NumPy arrays. Your example,

```python
% python -i -c 'import awkward as ak; import numpy as np; import gc'
>>> for i in range(1000000):
...   a = ak.Array([i])
...   del a
...   tmp = gc.collect()
... 
```

is pretty much a combination of the two steps: ArrayBuilder, then wrap as ak.Array. Doing the above explicitly accumulated 120 MB in 80 seconds.

So it sounds like this is a memory leak in ArrayBuilder, very likely the GrowableBuffer that gets allocated is somehow not getting freed. The memory that GrowableBuffer allocates is in a `std::shared_ptr` that should be kept alive only by the fact that the ArrayBuilder is held as a Python reference. In my first example, `del a` should have dropped the Python reference count to zero and `gc.collect()` should have deleted the ArrayBuilder, then GrowableBuffer instance, which should have dropped the `std::shared_ptr` reference count to zero to immediately free the memory. That doesn't seem to be happening, but I know where to look.

(For future triage: memory leaks are _bugs_, not _performance issues_.)

jpivarski:
PR #570 fixes it:

```python
>>> import awkward as ak
>>> import gc
>>> builder = ak.ArrayBuilder()
>>> builder.integer(123)
CPU  malloc at 0x56286a4b4a50 (8192 bytes)
[...]

PR #2311, opened 2023-03-13
fix: DECREFing PyObject*s from Python C API calls in v1.

### LINKED ISSUE #1127: Massive memory leak with awkward.Array
### Version of Awkward Array

1.5.1

### Python version

Python 3.6 and 3.8

### OS

Ubuntu 18.04 and macOS Big Sur

### Problem

Using `ak.Array` causes memory to be leaked. The following program causes the memory (RAM) consumption to grow indefinitely

```
import awkward as ak

while True:
    ak.Array([1])
```

yes, that is the whole program.

### Rate of leakage

Memory increase can be tracked with utilities (e.g. htop) or from Python itself. Using `psutil`, memory usage in megabytes can be printed with the following code

```
import os

import awkward as ak
import psutil


process = psutil.Process(os.getpid())

while True:
    ak.Array([1])

    usage = process.memory_info().rss / 1000 / 1000
    print(usage)
```

which (for me) shows a steady RAM usage increase at a rate of about 2 MB/s.

I briefly tried to find the cause for this with various memory-profiling tools but could not find the cause (I tried three tools and none of them even saw any change in object counts/sizes)

## Expected behavior

I see no good reason for why this should be happening, I would expect `ak.Array` to cause no memory leakage.

### LINKED ISSUE #1280: Memory leak when iterating over array
### Version of Awkward Array

1.7.0

Hi, 

A student I am working with noticed what appears to be a memory leak when iterating over array. I've replicated the issue with the following snippet. 
```
import psutil

A = ak.Array([[0,1],[0],[1,0]]*10000000)
my_sum = np.zeros(len(A))
i = 0
for event in A:
    my_sum[i] =  np.sum(event)
    i += 1
    if i %10000 == 0:
        print('RAM memory % used:', psutil.virtual_memory()[2])
```
A (much faster) solution without the leak to do the sum on the awkward array directly
```
my_sum = np.sum(A, axis=1) 
```
But I figured I should report the issue here in case it is helpful.

### LINKED ISSUE #2275: Unbounded memory use when iterating over RecordArray
### Version of Awkward Array

1.10.2

When using iterators for a RecordArray multiple times, we see apparently unbounded growth of the RSS of our code, which eventually causes them to be killed by the OOM-killer. Not entirely sure if this is coming from uproot or awkward, but the fact that it depends on the way the RecordArray is accessed suggests that it's an awkward issue.

To reproduce:

```
import uproot
import psutil, gc

FILE='data_ceph/larlumi/root/run_00358031.root'

def itr(tree):
    records=tree.arrays(['PosInTrain','EventNumber'])
    records['PosInTrain']*records['EventNumber']
    
def itr2(tree):
    records=tree.arrays(['PosInTrain','EventNumber'])
    for q in records:
        pass

p = psutil.Process()
print('Array-at-a-time operations')
for j in range(10):
    print(p.memory_full_info().rss)
    with uproot.open(FILE) as f:
        t=f['myTree']
        itr(t)
    gc.collect()
print(p.memory_full_info().rss)

print()
print('Iteration')
for j in range(10):
    print(p.memory_full_info().rss)
    with uproot.open(FILE) as f:
        t=f['myTree']
        itr2(t)
    gc.collect()
print(p.memory_full_info().rss)
```

The test file can be obtained from root://utatlas.its.utexas.edu//data_ceph/onyisi/larlumi/root/run_00358031.root . The first set of iterations ("Array-at-a-time") will increase and level out RSS use, while "Iteration" will grow on each loop by a similar amount. Explicitly switching off the caches in uproot.open() doesn't affect this.

The issue appears in awkward 1.10.2 / uproot 4.3.7. I can confirm that it seems to be fixed in awkward 2.0.8 / uproot 5.0.3 ; however the coffea dependencies mean we can't upgrade to this version at this time.

(And yes we do have reasons why we can't use the array-at-a-time operations ... and numba makes them pretty fast anyway ...)

### LINKED ISSUE #2310: awkward1 `form.parameters` exhibits unbounded memory growth
This is an issue for all present users of coffea until they migrate to the to-be released coffea 2023 (previously 0.8), due to how nanoevents manipulates the form.

```
uproot version = '4.3.7'
awkward version = '1.10.2'
```

The following code will grow in memory continuously, this becomes a problem for very long running jobs (and often kills people's analysis runs over the course of hours):
```python3
import uproot
import awkward

i = 0
x = uproot.open({"tests/samples/nano_dy.root": "Events"})

while True:
    print("manipulating form",i)
    i += 1
    for _, branch in x.iteritems():
        form = branch.interpretation.awkward_form(None)
        form = uproot._util.awkward_form_remove_uproot(awkward, form)
```

Upon commenting out this last line the unbounded memory growth stops.

jpivarski:
A memory leak in Coffea was traced to this: we were calling Python C API functions (just [PyUnicode_DecodeUTF8](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_DecodeUTF8)) that returned new refs and then passing them directly to pybind11 `py::object` (or its subclasses') constructors. I had thought that the `py::object` constructor would steal this reference and implicitly call `Py_DECREF` on it when the `py::object` enters its destructor, but this is evidently not the case. (The memory leak in #2310 toggles exactly with adding `Py_DECREF` into [parameters2dict](https://github.com/scikit-hep/awkward/blob/198cf6188bc70a19bb413a1bd750f93e605a8e57/src/python/content.cpp#L1737-L1752
).)

In this PR, I have added `Py_DECREF` calls as late as possible in every branch of the program flow on the raw `PyObject*` objects directly returned by `PyUnicode_DecodeUTF8`. I got all of these cases, and there are no other Python C API function calls that return `PyObject*` in the codebase. All of the other Python C API functions called,

   * [PyObject_CheckBuffer](https://docs.python.org/3/c-api/buffer.html#c.PyObject_CheckBuffer)
   * [PyBytes_Check](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_Check)
   * [PyBytes_Size](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_Size)
   * [PyBytes_AsString](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_AsString)

return C types, not `PyObject*`. (The Python documentation says the `char*` that `PyBytes_AsString` returns is owned by Python and must not be released. In our code, we copy it immediately.)

This PR also updates the version number in anticipation of an Awkward v1 bug-fix release, and it fixes a test that now fails a NumPy deprecation warning (no relationship to the actual bug; it's just that `main-v1` is an old branch).

There is no equivalent action to be performed on Awkward `main` because none of this stuff exists there.


jpivarski:
Check: are the following memory leaks now fixed?

  - [ ] #567
  - [x] #1127
  - [x] #1280
  - [x] #2275

I don't remember why I thought that #567 was an unresolved memory leak. (The first part of that issue addressed a different leak in ArrayBuilder; afterward, I had thought we had hints of another one, but the conversation that's there doesn't look conclusive to me now.)

jpivarski:
@agoose77 I disabled tests of `ak.to_`/`from_rdataframe` and JAX. The RDataFrame tests were testing the last `ak._v2.*` version, and the JAX tests were testing the v1 version, neither of which are supported anymore. Hopefully, the remaining tests will pass.

Do you expect that enough of the infrastructure will work to do a PyPI 1.10.3 release the normal way—by creating a release in GitHub targeting the `main-v1` branch (after this gets merged)?

What would be necessary to get it from PyPI into conda-forge? Does that pick up any new PyPI releases, even if they're out of order?

agoose77:
I think the PyPI release should work just fine with the `wheels.yml` workflow. I'm not so sure about how we would release a patch version for an old major version on conda-forge. I think it should be possible, but let me do some digging.

src/python/content.cpp
```diff
@@ -1740,13 +1740,17 @@ parameters2dict(const ak::util::Parameters& in) {
...
-                                         cppvalue.length(),
-                                         "surrogateescape"));
+    PyObject* tmp1 = PyUnicode_DecodeUTF8(cppkey.data(),
+                                          cppkey.length(),
+                                          "surrogateescape");
+    PyObject* tmp2 = PyUnicode_DecodeUTF8(cppvalue.data(),
+                                          cppvalue.length(),
+                                          "surrogateescape");
+    py::str pykey(tmp1);
+    py::str pyvalue(tmp2);
```
agoose77:
```suggestion
    py::str pykey = reinterpret_steal<py::str>(PyUnicode_DecodeUTF8(cppkey.data(),
                                          cppkey.length(),
                                          "surrogateescape"));
    py::str pyvalue = reinterpret_steal<py::str>(PyUnicode_DecodeUTF8(cppvalue.data(),
                                          cppvalue.length(),
                                          "surrogateescape"));
```

agoose77: I think we can do this using `pybind11`: https://pybind11.readthedocs.io/en/stable/reference.html#_CPPv4I0E18reinterpret_borrow1T6handle

src/python/content.cpp
```diff
@@ -1740,13 +1740,17 @@ parameters2dict(const ak::util::Parameters& in) {
...
     out[pykey] = py::module::import("json").attr("loads")(pyvalue);
+    Py_DECREF(tmp1);
+    Py_DECREF(tmp2);
```
agoose77:
```suggestion
```

src/python/content.cpp
```diff
@@ -1761,20 +1765,26 @@ template <typename T>
...
-                                       cppvalue.length(),
-                                       "surrogateescape"));
-  return py::module::import("json").attr("loads")(pyvalue);
+  PyObject* tmp1 = PyUnicode_DecodeUTF8(cppvalue.data(),
+                                        cppvalue.length(),
+                                        "surrogateescape");
+  py::str pyvalue(tmp1);
+  py::object out = py::module::import("json").attr("loads")(pyvalue);
+  Py_DECREF(tmp1);
+  return out;
```
agoose77, line 1774:
```suggestion
  py::str pyvalue = reinterpret_steal<py::str>(PyUnicode_DecodeUTF8(cppvalue.data(),
                                        cppvalue.length(),
                                        "surrogateescape"));
  return py::module::import("json").attr("loads")(pyvalue);
```

src/python/content.cpp
agoose77: (And repeat for all instances)

agoose77:
I checked on Matrix, and it looks like we just need to maintain a `v1` branch on our feedstock. That should take care of it. Once we make the PyPI release, I can revert back to the v1 feedstock for a branch to make the release.

jpivarski:
Alternative to using the `py::` namespace in front of `reinterpret_steal`:

```c++
#include <pybind11/pytypes.h>
```

But I like the namespace to denote all things pybind11.

tests/v2/test_0355-mixins.py
```diff
@@ -54,59 +54,13 @@ def weighted_add(self, other):
...
-    assert to_list(abs(one)) == [
-        [1.4866068747318506, 2.973213749463701, 4.459820624195552],
-        [],
-        [5.946427498927402, 7.433034373659253],
-    ]
-    assert to_list(one.distance(wtwo)) == [
-        [0.14142135623730953, 0.0, 0.31622776601683783],
-        [],
-        [0.4123105625617664, 0.0],
-    ]
```
agoose77, line 112: Did we get rid of this because it's flukey?

agoose77 (approved):

tests/v2/test_0355-mixins.py
jpivarski: Yes. It's been failing in the last digits on only some platforms. This is a test of the not-completely-done v2 inside `ak._v2.*`, so I don't think we need to work very hard to save it.

jpivarski: In principle, we could remove all of `ak._v2` and `tests/v2` from `main-v1`, but I don't want to make such a large diff in this PR.

agoose77: I'd prefer to keep some of the tests so we don't fully regress, but I agree this is transitional code and should be treated as such. We didn't make long-term guarantees about v2-in-v1.

PR #2314, opened 2023-03-14
fix: support unknown scalars in `__getitem__`

agoose77:
## TL;DR
- Add support for 0D arrays as array index arguments

### DIFF
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -68,7 +68,6 @@ class NumpyKernel(BaseKernel):
     def _cast(cls, x, t):
         if issubclass(t, ctypes._Pointer):
             # Do we have a NumPy-owned array?
-            # TODO should kernels strip nplike wrapper? Probably
             if Numpy.is_own_array(x):
                 if x.ndim > 0:
                     return ctypes.cast(x.ctypes.data, t)
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -70,7 +70,10 @@ class NumpyKernel(BaseKernel):
             # Do we have a NumPy-owned array?
             # TODO should kernels strip nplike wrapper? Probably
             if Numpy.is_own_array(x):
-                return ctypes.cast(x.ctypes.data, t)
+                if x.ndim > 0:
+                    return ctypes.cast(x.ctypes.data, t)
+                else:
+                    return x
             # Or, do we have a ctypes type
             elif hasattr(x, "_b_base_"):
                 return ctypes.cast(x, t)
--- a/src/awkward/_regularize.py
+++ b/src/awkward/_regularize.py
@@ -27,7 +27,7 @@ def is_integer(x) -> bool:
 
 
 def is_array_like(x) -> bool:
-    return hasattr(x, "shape") and hasattr(x, "dtype")
+    return hasattr(x, "shape") and hasattr(x, "dtype") and hasattr(x, "T")
 
 
 def is_integer_like(x) -> bool:
--- a/src/awkward/_slicing.py
+++ b/src/awkward/_slicing.py
@@ -6,7 +6,7 @@ import operator
 import awkward as ak
 from awkward._backends import Backend
 from awkward._errors import wrap_error
-from awkward._nplikes import nplike_of, to_nplike
+from awkward._nplikes import to_nplike
 from awkward._nplikes.jax import Jax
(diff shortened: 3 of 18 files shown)


agoose77:
## Supporting 0D arrays and Python scalars
> **Note**
> The following is just context for posterity, it is not necessary for reviewing this PR.

In a previous PR, I unified the typetracer scalar and 0D arrays to lift the degeneracy of scalar types (there are only 0D arrays for typetracer now). However, there are places in the code where we currently assume that we have integers. The logical first step is to ask "can we unify these to a single type?"It would be nice to have a "canonical" scalar type, be it `int` or `array()` so that we can make more assumptions in our code.

- On the one hand, I'd prefer to support scalars as 0D arrays across all nplikes; otherwise, calling `int()` on a CuPy array to convert it to a Python scalar will require a sync which, although not moving a lot of data, is not ideal. 
- On the other hand, if we _require_ that the scalar types live on the GPU (i.e., only accept 0-D arrays), then Python scalars would need to be allocated on the device.

More generally, it's important to be able to disambiguate between "known scalar" and "unknown scalar" for typetracer arrays. Given that we can't mix array types, this means using Python scalars for known values, and 0D arrays for unknown values. This means that we have two distinct value types: `int` and `unknown-scalar`, regardless of whether the latter's actually a scalar or a 0D array.

With the above discussion, it's clear to me that there is no single type that describes a valid int-like index; the set cardinality is 2:
- `int`
- `TypeTracerArray(np.integer, shape=())`). 

For non-typetracer arrays, we can also accept the `np.generic` scalar instances, i.e. 
- `int`
- `np.integer[generic]`
- `np.array(dtype=np.integer, shape=())`

We *could* also accept `np.generic` instances for typetracer, but I'll describe why that's not necessary below.

The missing type for typetracer is the unknown equivalent of `int`. If we introduce an "unknown-int" type, though, we must now consider the following types for an index into a `TypeTracerArray`:
- `unknown-int`
- `int`
- `TypeTracerArray(np.int64, shape=())`
- `TypeTracerArray(np.int64, shape=(...,))`

With a canonicalisation rule, this would become
- `unknown-int`
- `int`
- `TypeTracerArray(np.int64, shape=(...,))`

Ultimately, it seems that permitting 0-D arrays as "int-like" objects is the cleanest path forward. This PR modifies the kernel dispatch so that 0D arrays are converted to `np.generic` instances with `.item()`, so within the `Content` protocols like `_getitem_XXX`, we may have any int-like object. That's what this PR now does.

agoose77:
## Supporting mixed known-unknown scalars

Right now, the `tuple` index pathway invokes `nplike.broadcast_arrays`, which loses information about integer indexes under typetracer; these are promoted to 0D typetracer arrays before broadcasting. This means that
```python
array[0]
```
and
```python
array[0,]
```
are not the same case; the latter loses information about the index value, i.e. cannot detect out-of-bounds at typetracer time.

I've been trying to preserve true Python integers so that we can detect out-of-bounds indexing in e.g. `RegularArray` or 1D arrays with known-lengths. Perhaps this is too ambitious a goal? @jpivarski what do you think, is this a reasonable tradeoff?

agoose77:
The failing test is a `repr` of `array(0)` vs `0`. This should probably be fixed by using `str` instead of `repr`, but that might not be a blanket rule, as oftentimes we want `repr`.

agoose77:
@jpivarski following from our discussion, I just changed the error string to account for the `array(0)` instead of `0`. The motivation for this change is that this is part of the error message that refers to the internal subslice, not the external user-authored slice, and we're free to normalise those things as we see fit.

jpivarski (approved):
> @jpivarski following from our discussion, I just changed the error string to account for the `array(0)` instead of `0`. The motivation for this change is that this is part of the error message that refers to the internal subslice, not the external user-authored slice, and we're free to normalise those things as we see fit.

Agreed.

I don't think it was necessary to introduce a `NO_HEAD` constant, since `()` is completely unique, but I don't think it's a bad thing, either.

This is good to go!

agoose77:
@jpivarski we need something with an identity so we can avoid equality comparison (we want to support unknown scalars). That's why I used this constant. I might have made an oversight!

jpivarski:
I see! `()` is also a length-zero sequence. Got it.

agoose77:
Actually, in this case it's even more peculiar;  `__eq__` raises an exception for unknown scalars, and currently it is always invoked for the first switch case(`if x == ()`). Using `is` we can bypass this and test identity.

PR #2315, opened 2023-03-15
fix: uneccessary use of `nplike_of`

agoose77:
In preparation for a simplification of `nplike` identification, this PR eliminates redundant `nplike_of` usages, and/or replaces them with `backend_of`.

In future, `nplike_of` will only accept array objects (not layouts, etc).

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -9,7 +9,6 @@ import itertools
 from collections.abc import Sequence
 
 import awkward as ak
-from awkward._nplikes import nplike_of
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._nplikes.shape import unknown_length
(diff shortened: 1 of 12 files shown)


jpivarski (approved):
This is different from the one I've been seeing develop, and it looks pretty simple: making the calls to get nplikes more uniform in preparation for a refactoring. Looks good!

> In future, `nplike_of` will only accept array objects (not layouts, etc).

By "array objects," do you mean buffers, like `np.ndarray`, `cp.ndarray`, or `TypeTracerArray`, not `ak.Array`? (It seems like that's what you mean.)

agoose77:
> By "array objects," do you mean buffers, like np.ndarray, cp.ndarray, or TypeTracerArray, not ak.Array? (It seems like that's what you mean.)

In an ecosystem with N array types, you'd think I'd be more precise by now ;)

Yes, I mean the nplike buffers. The rewrite of #2307 will simplify `nplike_of` to simply cache the nplike which returns `True` for `is_own_array_type`, and make `backend_of` do the heavy lifting w.r.t to inspecting layouts, arraybuilder, etc.

PR #2318, opened 2023-03-16
fix: jupyterlite dependency changes

agoose77:
`jupyterlite` partitioned into `jupyterlite-core` and `jupyterlite`. We pin `jupyterlite`, but now we need to pin the root package.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -11,7 +11,7 @@ myst-nb
 sphinx-external-toc
 jupyterlite-sphinx
 ipyleaflet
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Pinned to a specific version? Well, I guess that makes sense because our documentation site is an application, not a library.

agoose77:
> Pinned to a specific version? Well, I guess that makes sense because our documentation site is an application, not a library.

Yes, we need to pin the pyodide version which is closely related. So, for now, it's a hard pin!

PR #2319, opened 2023-03-16
feat: add support for arrays in `from_buffers`

agoose77:


src/awkward/operations/ak_from_buffers.py
```diff
@@ -157,11 +159,28 @@ def getkey(form, attribute):
...
-        return array.byteswap(inplace=False)
-    else:
+    if nplike.is_own_array(buffer):
+        array = buffer.view(dtype)
+
+        if array.ndim != 1:
+            raise ak._errors.wrap_error(
+                TypeError(f"dimensionality of array should be 1, not ({array.ndim})")
+            )
+        # Require 1D
```
jpivarski: Does `ak.to_buffers` ever return a multidimensional array? For instance, from a `NumpyArray` node? We want

```python
ak.from_buffers(*ak.to_buffers(x)) == x
```

to continue to be true.

agoose77: I was getting a bit caught up with not breaking existing ingest, yet I didn't even think about round-tripping. I've fixed this, and ensured that we don't copy any data.

jpivarski (approved):
Looks good! It's nice that it had already been factored out into `_from_buffer`, so that this could be added in only one place (apart from ensuring that `reshape` is fully implemented).

agoose77:
@jpivarski agreed!

agoose77:
@douglasdavis this *hopefully* should be what we discussed today :)

PR #2324, opened 2023-03-20
refactor: hide `awkward.typing`

agoose77:
This PR just performs a simple rename to hide `ak.typing` (it's supposed to be private for now).

PR #2325, opened 2023-03-20
refactor: move `_parameters_XXX` functions into `_parameters` submodule

agoose77:
> **Warning**
> For ease, depends upon #2324, which should be merged first.

The same rationale applies here as for other renames: the `_parameters` functions are L3, and should existing in an L3 submodule rather than L3 names in an L2 submodule.

There is, in my mind, a stronger reason to create a new submodule under `awkward` than `awkward.forms`, but I would easily be moved on this.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -12,7 +12,11 @@ from awkward._behavior import find_custom_broadcast
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._nplikes.shape import unknown_length
-from awkward._parameters import parameters_are_empty, parameters_intersect
+from awkward._parameters import (
+    parameters_are_empty,
+    parameters_are_equal,
+    parameters_intersect,
+)
 from awkward._typing import Any, Callable, Dict, List, TypeAlias, Union
 from awkward._util import unset
 from awkward.contents.bitmaskedarray import BitMaskedArray
@@ -241,7 +245,7 @@ def all_or_nothing_parameters_factory(
         first_parameters = input_parameters[0]
         # Ensure all parameters match, or set parameters to None
         for other_parameters in input_parameters[1:]:
-            if not ak.forms.form._parameters_equal(first_parameters, other_parameters):
+            if not parameters_are_equal(first_parameters, other_parameters):
                 break
         else:
             parameters = first_parameters
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -12,6 +12,7 @@ from awkward._behavior import find_custom_broadcast
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._nplikes.shape import unknown_length
+from awkward._parameters import _parameters_intersect, _parameters_is_empty
 from awkward._typing import Any, Callable, Dict, List, TypeAlias, Union
 from awkward._util import unset
 from awkward.contents.bitmaskedarray import BitMaskedArray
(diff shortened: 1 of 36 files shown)


jpivarski (approved):
Yes, the fact that these parameter-manipulation functions were in `awkward.forms.form` was bothering me; I had an intention to eventually move them to `awkward._util`. But since you broke down `awkward._util` into more specialized helper modules, it makes sense for `awkward._parameters` to be a new helper module.

You asked before about a flat structure versus a hierarchical structure, and I argued for a flat structure because it's easier to find things. (An ontology that makes sense at one time doesn't always make sense at another time, when you're trying to find something.)

But there are now so many helper modules that maybe they should be collected under _one_ subdirectory. Like, maybe everything that used to be in _util.py (plus these parameter-manipulation functions) should be under

```
awkward/_util/__init__.py       uncategorized (or not yet categorized)
awkward/_util/parameters.py     parameter-related functions
awkward/_util/...               the others that have recently been split out
```

?

When we're hunting for the code to check on something/answer someone's question, we'd have to know if what we're looking for is a main function or a helper function, but that's the roughly the same as the L2/L3 distinction. "Roughly" because I think it still makes sense for `_connect` to not be considered utility/helper functions: they're categorized by all being connections to third-party libraries.

This is merely a suggestion. When I'm looking for code that answers a question I'm having, I'll adjust to what you decide. If I can't find it, I'll `git pull` my local repo and search with `grep -r`.

agoose77:
Incidentally, this relates to a concept of module organisation that I'm still working out. My opinion has changed over the years, and Awkward provides a particularly interesting example of a package with lots of interrelated code. I *think* the main rules that I could write down for my preferences are:
- Flat is better than nested
   - Grouping is important, but it is possible to group subunits fairly endlessly
- Smaller modules admit easier groupings
- Named modules are better than `utils.py`
- Explicit imports (`from ak._util import x`) are better than implict attribute lookups (`ak._util.x`)
  - Shorter lines are generally better
  - Type hints are particularly much more readable

With a preference for "readability" (as I define it), this approach seems to work quite well. There are, or course, tradeoffs:
- Cyclic import errors are more likely with `from module import symbol` if one *also* imports submodules in `__init__`; Python can resolve import cycles as long as no-one tries to read the module dictionaries during resolution time. 
   - Therefore, `from x.y.z` is more common than `from x import z` (where `y` exports `z`).
- Ambiguous symbol names are harder to locate
   - Use of meaningful names is therefore important! (e.g. `nplike_of` vs `of`)


To address these shortcomings, I'm pushing imports to the module scope (rather than relying on `ak._util.xxx` to work at invocation time) so that we can try and maximally decouple our modules. That will implicitly help to draw the boundaries between code (by promoting new modules) that we can then place into different packages (directories with `__init__.py`, given how overloaded the term "package" is in Python). 

> But there are now so many helper modules that maybe they should be collected under one subdirectory. Like, maybe everything that used to be in _util.py (plus these parameter-manipulation functions) should be under [...]

This is something I'm open to, too. My approach, which I think mirrors yours, is to figure this out when we need to, rather than up front. I've long accepted that it's better to solve these things iteratively rather than all in one go. Specifically, how much nesting, and what kind of naming structure to use. There is always a risk that everything can go in `_util` if you're generalising enough. Right now, I don't see a huge need to introduce `_util`, but again, I'm open to it if you think it makes sense.

jpivarski:
Iteratively is good. One way to stay grounded is to remember what this is for: **efficiently finding the code that you're looking for**. Thus,

> * Named modules are better than `utils.py`

The advantage of nested hierarchies is not having to look at everything in the world when you're looking for a specific thing; the disadvantage is that if the nesting scheme is not natural or guessable at the time of lookup, searching for something would involve back-tracking.

Search tools like grep, GitHub search, and even just the browser incremental search[^1] shift the balance toward flatter hierarchies, but only up to a point. Having a lot of little files in the top-level was a problem when we needed to change `VERSION_INFO` frequently (and it's still a little problem navigating to `uproot/version.py`; it would be better if they were near the start of the alphabet!).

[^1]: Incremental search in a web browser was a _huge_ plus when it was added. It was enough to make me switch from... I think Netscape to Mozilla. At that time, the search was `/` and if the highlighted match was a link, `enter` would follow the link. Navigating the web by keystrokes was great; now you have to have an extra `tab` step to go from highlighted match to follow it, which is a little stumbling block, but not too bad.

PR #2062, opened 2023-01-04
docs: move to newest pyodide toolchain

agoose77:
Our existing CI requires several pieces of information to remain in sync in order to setup WASM building. With the new pyodide release, we can compute this (e.g. EMSDK version) at build time. This PR creates an additional `requirements-wasm.txt` that contains only the `pyodide-build` dependency meaning that all dependencies are contained in requirements files. 

This PR will fail whilst https://github.com/pyodide/pyodide/issues/3415 is being fixed, modulo other bugs in the latest release.

PR #2326, opened 2023-03-20
docs: fix pyodide (again!)

agoose77:
This PR reworks the fix in #2318. That fix was valid, but we need to wait until we can build for pyodide 0.22.0, which is given by #2062.

I usually use the docs release to rapidly test "in-the-wild" Awkward Array. I noticed that the build is currently broken.


agoose77:
This is actually not quite-right for this branch

jpivarski (approved):
:+1:

PR #2327, opened 2023-03-20
fix: expose array interface for CUDA

agoose77:
`ak.index.Index(existing_index)` doesn't work for CuPy arrays. We rely on being able to do this in our internal layout logic.

I think what happens is that CuPy asks NumPy to read this array, which then fails as what's returned is _not_ a NumPy array.

We can generalise our index to expose the various array interfaces. In the longer term, we should also expose DLPack.

### DIFF
--- a/src/awkward/index.py
+++ b/src/awkward/index.py
@@ -149,8 +149,13 @@ class Index:
     def __len__(self):
         return self.length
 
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
That's a good idea.

PR #2329, opened 2023-03-21
fix: test untested path in `ak.cartesian` & broadcasting

agoose77:
This PR: 
- removes a special case for zero-length arrays from `ak.cartesian`
- fixes missing case in `broadcast_and_apply` that required this patch
- fixes and tests a set of inputs that was broken in `ak.cartesian`

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -844,11 +844,7 @@ def apply_step(
                     elif isinstance(x, RegularArray):
                         nextinputs.append(x.content[: x.size * x.length])
                     else:
-                        raise ak._errors.wrap_error(
-                            AssertionError(
-                                "encountered non list-type despite all_same_offsets requiring lists"
-                            )
-                        )
+                        nextinputs.append(x)
 
                 outcontent = apply_step(
                     backend,
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -160,7 +160,7 @@ def all_same_offsets(backend: ak._backends.Backend, inputs: list) -> bool:
                 my_offsets = index_nplike.empty(0, dtype=np.int64)
             else:
                 my_offsets = index_nplike.arange(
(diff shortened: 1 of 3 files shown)


agoose77:
@jpivarski as this touches broadcasting code, I'd appreciate your eyes to take a close look at the changes to `_broadcasting.py` in order to assert that I've not introduced a regression. It looks like we don't actually pull out the contents of `RegularArray`, which we should be doing according the context of the branch (given that `all_same_offsets` checks the `RegularArray` offsets).

src/awkward/_broadcasting.py
```diff
@@ -160,7 +160,7 @@ def all_same_offsets(backend: ak._backends.Backend, inputs: list) -> bool:
                 my_offsets = index_nplike.empty(0, dtype=np.int64)
             else:
                 my_offsets = index_nplike.arange(
-                    0, x.content.length, x.size, dtype=np.int64
+                    0, x.content.length + 1, x.size, dtype=np.int64
```
agoose77, line 163: We want a range that is _inclusive_ of the endpoints, which is not the case by default.

src/awkward/_broadcasting.py
```diff
@@ -841,7 +841,8 @@ def continuation():
                         else:
                             lencontent = backend.index_nplike.max(stops)
                             nextinputs.append(x.content[:lencontent])
-
+                    elif isinstance(x, RegularArray):
+                        nextinputs.append(x.content[: x.size * x.length])
```
agoose77, line 845: We test for `RegularArray` in `all_same_offsets`, but here we weren't actually pulling out the content.

src/awkward/_broadcasting.py
jpivarski: We do? Oh, yes—we do! Offsets always have to be one longer than the array itself; that's right. Good catch!

jpivarski (approved):
Fantastic; I'm glad you found this.

PR #2331, opened 2023-03-21
feat!: rename cpptype

agoose77:
This PR follows up #2306 and renames `cpptype` to `cpp_type`, to align with our PEP8 naming conventions!

### DIFF
--- a/tests/test_2306_cppyy_git.py
+++ b/tests/test_2306_cppyy_git.py
@@ -60,7 +60,7 @@ def test_array_as_type():
     )
 
     source_code_cpp = f"""
-    double go_fast_cpp({array.cpptype} awkward_array) {{
+    double go_fast_cpp({array.cpp_type} awkward_array) {{
         double out = 0.0;
 
         for (auto list : awkward_array) {{
@@ -114,5 +114,5 @@ def test_array_as_templated_type():
 
     cppyy.cppdef(source_code_cpp)
 
-    out = cppyy.gbl.go_fast_cpp_2[array.cpptype](array)
+    out = cppyy.gbl.go_fast_cpp_2[array.cpp_type](array)
     assert out == ak.sum(array["y"])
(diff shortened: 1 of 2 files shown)


ianna:
@agoose77 - you'd need to build cppyy from source to test it :-)

jpivarski (approved):
Since this is such a simple PR, I'd accept it without being formally tested. (It was also the case that #2306 wasn't tested in CI, only on @ianna's computer.) Eventually, cppyy 3.0.1 will get into PyPI and it will be tested in CI.

For this PR,

```
% git checkout agoose77/feat-rename-cpp-type
Branch 'agoose77/feat-rename-cpp-type' set up to track remote branch 'agoose77/feat-rename-cpp-type' from 'origin'.
Switched to a new branch 'agoose77/feat-rename-cpp-type'
% fgrep -r cpptype src/awkward tests
```

demonstrates that there are no more instances of the word "`cpptype`", and I can see from the diff that every one of them has been given an underscore (no other change). I would consider that sufficient to accept it.

agoose77:
I agree with the rationale here; I did a simple grep, and checked beforehand as to whether there were any hits that were not to be replaced. I'll get a test env up and running soon, but for now I am happy to merge.

tests/test_2306_cppyy_git.py
```diff
@@ -114,5 +114,5 @@ def test_array_as_templated_type():
 
     cppyy.cppdef(source_code_cpp)
 
-    out = cppyy.gbl.go_fast_cpp_2[array.cpptype](array)
+    out = cppyy.gbl.go_fast_cpp_2[array.cpp_type](array)
```
ianna, line 117: @agoose77 - local tests pass!

ianna (commented):
@agoose77 - I can confirm - the change is complete and all the tests pass. Thanks!

tests/test_2306_cppyy_git.py
agoose77: @ianna thanks! I built a wheel last night but I didn't think I needed to build the whole stack, and apparently that was a mistake! I was going to re-run it this morning, but you've saved me the trouble!

PR #2332, opened 2023-03-21
fix: Numba string reference count memory leak.

### LINKED ISSUE #2330: Memory leak when using array of strings
### Version of Awkward Array

2.1.1

Unbounded memory growth when using arrays with strings. Running the following code will demonstrate the issue. In each loop iteration, memory usage keeps on increasing

``` python
import awkward as ak

from tqdm.cli import tqdm

def f(arr):
    result = 0
    for i in range(len(arr)):
        result += (arr[i] == "mov")
    return result


for i in tqdm(range(100)):
    arr = ak.Array(['aaaaa' * 100 * 1024] * 1024)
    f(arr)
```

jpivarski:
When creating a string in Numba using [PyUnicode_FromKindAndData](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromKindAndData) or a bytes object from [PyBytes_FromStringAndSize](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromStringAndSize), the returned object is a new reference, and _I guess_ Numba also adds a reference to it when we pass it on.

Underline and highlight "_I guess_". I only know that from the fact that this code change fixes a memory leak that is present without it. Maybe Numba is incrementing the count in `pyapi.to_native_value` because if the new `pyapi.decref(pystr)` is _before_ that line, it segfaults. (It can't go any later, since it must be within the GIL-acquired region.)

The memory leak is pretty well measured by the test, which I'm also uncomfortable with because it's measuring the process's total RSS memory use, and if pytest runs in parallel, maybe another test will boost the memory use. The 30 MiB figure in the test is exactly what is produced by the memory leak (determined by experimentation, but also note that it's 1024×1024 3-character strings in UTF-8). With the memory leak, the memory of the process, controlled by garbage collection, grows by exactly 30 MiB; without the memory leak, it grows by exactly zero. But that's when I run pytest sequentially. The 90% fudge factor is to allow it to run in parallel.

This test also introduces the need to have psutil as a testing dependency.

@agoose77, if you see a better way to test this, let me know. Or maybe we just don't test it? I've tested it manually and we now know that there needs to be a `pyapi.decref` there. It's not like that part of the code is going to be changed—we don't haphazardly remove `pyapi.incref` and `pyapi.decref`.

@ianna, do you see anything in Numba that justifies this `pyapi.decref`?


jpivarski:
Okay, I actually figured out what's going on here (in https://github.com/scikit-hep/awkward/issues/2330#issuecomment-1478682845).

Numba's string and byte object types ([UnicodeType](https://github.com/numba/numba/blob/4b8b176b5bdd2f93200b1eca727411c3e1657858/numba/core/types/misc.py#L530-L537) and [Bytes](https://github.com/numba/numba/blob/4b8b176b5bdd2f93200b1eca727411c3e1657858/numba/core/types/containers.py#L96-L103)) are _not_ Python types. I'm just using the Python API to build them. `pyapi.to_native_value` is a general Numba function for converting Python objects into Numba objects, and it's the Numba objects that escaped this function; the Python objects were ignored. Therefore, the `pyapi.decref(pystr)` is absolutely necessary to delete the no-longer-used Python object, which was a new reference from [PyUnicode_FromKindAndData](https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_FromKindAndData) or [PyBytes_FromStringAndSize](https://docs.python.org/3/c-api/bytes.html#c.PyBytes_FromStringAndSize).

Also, it's in principle possible to create Numba strings or byte objects directly, without going through Python, and therefore avoid acquiring the GIL. That would be an optimization for the future.

Also @ianna, this clears up some mysteries that we had looking at Numba strings. First, there was a time when I wasn't able to find where we handle strings ("Somehow, it's being done correctly, but I can't figure out where!"): it was in `awkward/behaviors/string.py`, not `awkward/_connect/numba/*`. But also, some of the confusions I had about ArrayBuilder and string handling (or byte object handling) was related to the fact that I had thought that Numba was using Python objects for strings and byte objects, rather than its own.

jpivarski:
@agoose77, what do you think? Now that the reason for the `pyapi.decref` is understood, can we remove the test, so that we don't get bad surprises when running pytest in parallel? It could be acceptable to take this fix without a test.

I'll mark you as reviewer.

agoose77:
Memory regressions are hard to test for. We definitely could guard these tests behind a flag, so that they're runnable when explicitly requested. 

Equally, I think a better test here is to ensure the refcounting is valid. However, given that the test for this is ~100% identical to writing the actual implementation, I'd be happy to just remove the test.

jpivarski:
This reference is inaccessible. Earlier, I was trying to use `sys.getrefcount` for the test, but no—it's lost inside of a Numba-compiled block and doesn't participate in NRT (Numba Runtime). I don't see any way to get at the reference count from the outside.

jpivarski:
I'd be more comfortable with removing the test, but it exists in this PR's history if we come back and want to set up a Numba memory suite. (This wouldn't be the only thing to check.)

agoose77:
> This reference is inaccessible. Earlier, I was trying to use `sys.getrefcount` for the test, but no—it's lost inside of a Numba-compiled block and doesn't participate in NRT (Numba Runtime). I don't see any way to get at the reference count from the outside.

Oh yes, that would be the case! I'm glad we've caught *another* memory bug; that's one fewer left in the wild!

ianna:
Fantastic! Thanks for fixing it!

agoose77 (approved):

requirements-test.txt
```diff
@@ -3,6 +3,7 @@ jax[cpu]>=0.2.15;sys_platform != "win32"
 numba>=0.50.0;python_version < "3.11"
 numexpr
 pandas>=0.24.0
+psutil
```
jpivarski:
```suggestion
```

PR #2333, opened 2023-03-21
fix: `ak.full_like` with unknown arrays

agoose77:
We already test for unknown scalars, but unknown arrays slip through! In general, this logic should test for scalar fill values in addition to whether they're known.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -17,7 +17,6 @@ from awkward._typing import (
     Literal,
     Self,
     SupportsIndex,
-    TypeVar,
 )
 
 np = NumpyMetadata.instance()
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Fixing `ak.full_like` should also fix `ak.ones_like` and `ak.zeros_like`, right? Because they're implemented in terms of `ak.full_like`.

If you're adding a test for `ak.full_like`, just copy-paste it for `ak.ones_like` and `ak.zeros_like` (removing the fill value) because I think they would automatically work.

agoose77:
TODO: this should simplify into broadcasting.

jpivarski:
Should this be `pr-on-hold` or closed?

agoose77:
I'll close for now.

PR #2335, opened 2023-03-22
fix: generate an array view when an Array C++ type is requested

ianna:


jpivarski (approved):
This was pulled out of #2334 because we should let the docs PR wait until cppyy 3.0.1 is released and we want the bug-fix (separate from docs) to go in as soon as possible.

@ianna, has this passed local tests (with cppyy 3.0.1 installed on your comptuer)? If so, we can merge it. The strictness of actually testing it in CI will come when cppyy 3.0.1 is on PyPI.

ianna:
> This was pulled out of #2334 because we should let the docs PR wait until cppyy 3.0.1 is released and we want the bug-fix (separate from docs) to go in as soon as possible.
> 
> @ianna, has this passed local tests (with cppyy 3.0.1 installed on your comptuer)? If so, we can merge it. The strictness of actually testing it in CI will come when cppyy 3.0.1 is on PyPI.

Yes, it does pass local tests.

PR #2345, opened 2023-03-30
fix: generate RDataSource API based on ROOT attribute

### LINKED ISSUE #2344: Warning in `RDataSource::Initialise` in ROOT 6.28
### Version of Awkward Array

2.1.1

Rename legacy `RDataSource::Initialise` to `RDataSource::Initialize` when minimum required ROOT version is fixed to 6.28:
```python
Warning in <RDataSource::Initialise>: Initialise is deprecated. Please rename it to "Initialize" (with a z).
```

ianna:
fixes issue #2344

### DIFF
--- a/src/awkward/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_connect/rdataframe/from_rdataframe.py
@@ -126,6 +126,7 @@ def from_rdataframe(
     column_types = {}
     result_ptrs = {}
     contents = {}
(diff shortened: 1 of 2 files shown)


ianna (commented):
@jpivarski - updated as you've suggested to check an attribute rather than a version. The final test is documentation. The warning should disappear from it  :-)

jpivarski (approved):

PR #2348, opened 2023-03-31
fix: use cppyy include

ianna:


jpivarski (approved):
I agree; this is cleaner.

PR #2349, opened 2023-03-31
feat: implemented GrowableBuffer in Numba as a start toward LayoutBuilder

jpivarski:
@ianna, this is a complete implementation of GrowableBuffer (with `dtype`, `append`, `extend`, `snapshot` public methods) in Numba. It's also a good example to follow, instead of the ArrayView iterators, for a mutable Numba object like (future) LayoutBuilder.

The public API doesn't exactly follow the one in C++ because we can provide more high-level conveniences here. The C++ ArrayBuilder has a multi-step process to provide a length to Python, let Python allocate arrays of the right length, and then fill them, but all of that can be a single `snapshot` method here because Numba can ensure that Python owns the output arrays.

The lowered `GrowableBuffer` object is actually immutable and passed by value (`StructModel`). In order to ensure identity among all the objects created from one `GrowableBuffer` by assignment, all of the mutable data are in NumPy arrays and a `numba.typed.List`. Numba maintains the reference counts and referential identity of these objects; we take advantage of the fact that they do so. This way, if you

```python
>>> @nb.njit
>>> def f(x):
...     return x, x
...
>>> growablebuffer1 = GrowableBuffer(np.float32)
>>> growablebuffer2, growablebuffer3 = f(growablebuffer1)
```

then all three `growablebuffer1`, `growablebuffer2`, `growablebuffer3` share the same state: appending to any one of them appends to all of them.

In order to do this, I had to put two attributes, `_length` and `_pos`, into a 2-element NumPy array. If it weren't for trying to avoid confusions about mutability, they would have been simple attributes.

The `numba.typed.List` for a growable list of NumPy array `_panels` is also essential. This Python implementation of `GrowableBuffer` can't be used without Numba. I didn't like that conclusion, but the _only_ reason we'd have a pure Python version of these things is to provide an outside-of-Numba playground for practicing before moving things to Numba, so I guess it's not that bad.

The Numba implementations of `_add_panel`, `append`, `extend`, and `snapshot` (the only "substantial" functions) are all implemented in Numba's high-level extension API. I used the high-level API as much as possible. Quite a few underscored/hidden attributes and methods also exist in Numba, some of them implemented with the low-level API, and they're not all the same as the ones in Python. For instance, we have `_length_get()` in Numba and `_length` in Python.

The interpretation of the `resize` parameter might be different from how it is in C++. If so, it's because we never finished updating that. Now that we have panels, it no longer makes sense to grow an array exponentially with `resize` close to the golden ratio (we approximated with `1.5`); it would be much better for `resize` to only change the size once:

  * the first panel has size `initial` (1024 items)
  * _all subsequent panels_ (2nd, 3rd, 4th, ...) all have the same size `initial*resize` (8192 items)

because (1) many buffers are never going to exceed their first panel and (2) having a bunch of same-sized chunks in memory space is less likely to get fragmented. If the C++ `GrowableBuffer` doesn't have this behavior yet, it should—not having that was an oversight.

I checked all of the reference-counting and they're correct. I couldn't do it in the test because pytest seemed to be changing the reference counts—could it be that it's holding onto objects so that it can prepare a richer error message?

### DIFF
--- a/src/awkward/_connect/numba/growablebuffer.py
+++ b/src/awkward/_connect/numba/growablebuffer.py
@@ -158,6 +158,7 @@ for member in ("panels", "length_pos", "resize"):
 def GrowableBufferType_dtype(growablebuffer):
     def getter(growablebuffer):
         return growablebuffer._panel[0].dtype
+
     return getter
 
 
@@ -165,6 +166,7 @@ def GrowableBufferType_dtype(growablebuffer):
 def GrowableBufferType_length(growablebuffer):
     def getter(growablebuffer):
         return growablebuffer._length_pos[0]
+
     return getter
 
 
@@ -172,6 +174,7 @@ def GrowableBufferType_length(growablebuffer):
 def GrowableBufferType_pos(growablebuffer):
     def getter(growablebuffer):
         return growablebuffer._length_pos[1]
+
     return getter
 
 
@@ -204,15 +207,29 @@ def GrowableBufferType_box(typ, val, c):
     GrowableBuffer_obj = c.pyapi.unserialize(c.pyapi.serialize_object(GrowableBuffer))
     from_data_obj = c.pyapi.object_getattr_string(GrowableBuffer_obj, "_from_data")
 
-    growablebuffer = numba.core.cgutils.create_struct_proxy(typ)(c.context, c.builder, value=val)
-    panels_obj = c.pyapi.from_native_value(typ.panels, growablebuffer.panels, c.env_manager)
-    length_pos_obj = c.pyapi.from_native_value(typ.length_pos, growablebuffer.length_pos, c.env_manager)
-    resize_obj = c.pyapi.from_native_value(typ.resize, growablebuffer.resize, c.env_manager)
+    growablebuffer = numba.core.cgutils.create_struct_proxy(typ)(
+        c.context, c.builder, value=val
+    )
+    panels_obj = c.pyapi.from_native_value(
+        typ.panels, growablebuffer.panels, c.env_manager
+    )
+    length_pos_obj = c.pyapi.from_native_value(
+        typ.length_pos, growablebuffer.length_pos, c.env_manager
+    )
+    resize_obj = c.pyapi.from_native_value(
+        typ.resize, growablebuffer.resize, c.env_manager
+    )
 
-    out = c.pyapi.call_function_objargs(from_data_obj, (panels_obj, length_pos_obj, resize_obj))
+    out = c.pyapi.call_function_objargs(
+        from_data_obj, (panels_obj, length_pos_obj, resize_obj)
+    )
 
     # decref PyObjects
     c.pyapi.decref(GrowableBuffer_obj)
     c.pyapi.decref(from_data_obj)
+
+    c.pyapi.decref(panels_obj)
+    c.pyapi.decref(length_pos_obj)
+    c.pyapi.decref(resize_obj)
 
     return out
--- a/src/awkward/_connect/numba/growablebuffer.py
+++ b/src/awkward/_connect/numba/growablebuffer.py
@@ -1,0 +1,80 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import numba
+import numba.core.typing
+import numba.core.typing.ctypes_utils
+import numpy
+
+
+class GrowableBuffer:
+    def __init__(self, dtype, *, initial=1024, resize=10.0):
+        self._panels = [numpy.zeros((initial,), dtype=dtype)]
+        self._last_panel = self._panels[-1]
+        self._length = 0
+        self._last_panel_length = 0
+        self._resize = resize
+
+    def __repr__(self):
+        return f"<GrowableBuffer({self._last_panel.dtype!r}) len {self._length}>"
+
+    def __len__(self):
+        return self._length
+
+    def append(self, datum):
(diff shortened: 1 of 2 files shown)


jpivarski:
Oh, and I'm done with this PR, so if you approve, go ahead and merge it. Thanks!

tests/test_2349_growablebuffer_in_numba.py
```diff
@@ -0,0 +1,508 @@
...
+
+
+def test_unbox():
+    @numba.njit
+    def f1(x):
+        x
+        return 3.14
+
+    growablebuffer = GrowableBuffer(np.int32)
+    f1(growablebuffer)
```
ianna, line 128:
```suggestion
    assert f1(growablebuffer) == 3.14
```

src/awkward/_connect/numba/growablebuffer.py
```diff
@@ -0,0 +1,407 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+import numba
+import numba.core.typing.npydecl
+import numpy
+
+
+class GrowableBuffer:
+    def __init__(self, dtype, *, initial=1024, resize=8.0):
```
ianna, line 9: I'm not sure about the default value of `resize=8.0`. I agree, it should be large enough. Does it have to be a `Float`?

ianna (approved):
@jpivarski - looks great! I like the tuple approach to keep the track of current data length and a position in the last panel! 

I've added a suggestion to assert the result - just to be sure the function runs as expected. Please, feel free to either accept or reject it. I'm approving this PR as is. Please, merge it.

src/awkward/_connect/numba/growablebuffer.py
jpivarski: With the algorithm, "first panel is small, all the rest are much larger but equal to each other," the factor separating the first panel size from the rest should be a fairly large number. I was debating between 8 and 16; they're both an order of magnitude, but 8 is `2**3` and 3 is an odd number, whereas 4 is evenly half of a byte. (Shrug.)

This would ideally be set by a performance study. But that performance study would have to use some realistic test cases, since the thing that the new resize algorithm is trying to model is the steep distribution of "many small buffers, a few large ones." The actual distribution of number of small buffers (how small?) to number of large buffers (how large?) would matter.

Now that it's large, you're right that it no longer needs to be a floating point number. In the old algorithm, there's some literature saying that

$$\phi = \frac{1 + \sqrt{5}}{2} \approx 1.5$$

is the right `resize` factor, so that's why it was a float. We want both the C++ GrowableBuffer and the Numba GrowableBuffer to use the same algorithm and parameters, and if the C++ algorithm still needs to be changed (I don't know whether it does or not), leaving it as a floating point number but changing the algorithm would be the easiest change. It appears in a lot of places, just to be passed from one function to another, so changing the data type is hard, but including the `if (panels_.size() == 1)` clause in one place is easy.

tests/test_2349_growablebuffer_in_numba.py
jpivarski: Thanks for the suggestion, but this function returns 3.14 just so that it can return something. Checking its value doesn't test anything related to the GrowableBuffer implementation itself. Just calling `f1` is enough to verify that it doesn't segfault.

What I really wanted to do here is call `f1` repeatedly in a loop and ensure that the `PyObject`s in the GrowableBuffer Python implementation (the `numba.typed.List` and the NumPy arrays) don't leak. I did that test offline, but when I tried to incorporate it into the test suite, I was getting extra references. They weren't growing with the number of iterations (an indication of a leak), but they differed from the test values and couldn't be asserted. I know that pytest rewrites the test code, to instrument it with better diagnostics, and perhaps it was rewriting it in a way that acquired additional references.

Since you'll be writing more Numba code, we should keep this in mind, that we need to find a better way to check Python reference counts. Moreover, just checking them on the Python side is not always enough, since some Python objects are created in Numba functions, such as boxing and unboxing functions. Since we don't have any reference to those objects in Python, there's no way to call `sys.getrefcount` on them. CPython has an—undocumented!—Py_RefCnt function, but it's not exposed through Numba's `c.pyapi`.

It's also the case that a Python reference count is the first object in a `PyObject*`, so it can be accessed by unsafe casting. I did that horrible thing here: https://github.com/jpivarski-talks/2019-04-08-picscie-numpy/blob/master/09-jailbreaking-python.ipynb

```python
# The snake eats its own tail: reading PyObject refcount without changing it.
class PyObject(ctypes.Structure): pass
PyObject._fields_ = [
    ("ob_refcnt", ctypes.c_size_t),
    ("ob_type", ctypes.POINTER(PyObject)),
    # there are more struct fields; different Python types have a different number of them
]

hello = b"Hello, world!"
ptr = PyObject.from_address(id(hello))
print("ptr =", ptr)

# a few reference counts
print("\nBefore:")
print("ptr.ob_refcnt =", ptr.ob_refcnt, "sys.getrefcount(hello) =", sys.getrefcount(hello))

# a lot more reference counts
biglist = [hello] * 1000
print("\nAfter:")
print("ptr.ob_refcnt =", ptr.ob_refcnt, "sys.getrefcount(hello) =", sys.getrefcount(hello))
```

Oh, but about this suggested change: I'm going to not take it, so that I can merge the PR now.

src/awkward/_connect/numba/growablebuffer.py
ianna: Thanks, @jpivarski !

PR #2351, opened 2023-04-03
refactor: move `merge_as_union` to `ak_concatenate`

agoose77:
`ak._do` serves a particular purpose; to provide entrypoints to L3 functions in the `Content` class. I briefly outlined how this should ultimately be organised (in my view) [here](https://github.com/scikit-hep/awkward/issues/2108).

For now, the only user of this L3 function is `ak_concatenate`, so I'm placing it in that module.

Also, this PR removes an unused L3 function.

### DIFF
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -2,7 +2,7 @@
 from __future__ import annotations
 
 import copy
-from collections.abc import Mapping, MutableMapping, Sequence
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
Makes sense.

The organization of the merging logic has changed quite a bit, but in sensible ways.

Since this is a straightforward PR, I'll assume that you're done with it and merge it now.

PR #2353, opened 2023-04-03
fix: make `from_iter` require iterables!

agoose77:
Although `from_iter` can understand non-iterables, I think we really want to restrict it to *only* returning layouts/arrays.

### DIFF
--- a/src/awkward/operations/ak_from_iter.py
+++ b/src/awkward/operations/ak_from_iter.py
@@ -1,6 +1,8 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 __all__ = ("from_iter",)
 
+from collections.abc import Iterable
+
 from awkward_cpp.lib import _ext
 
 import awkward as ak
@@ -72,6 +74,13 @@ def from_iter(
 
 
 def _impl(iterable, highlevel, behavior, allow_record, initial, resize):
+    if not isinstance(iterable, Iterable):
+        raise ak._errors.wrap_error(
+            TypeError(
+                f"cannot produce an array from a non-iterable object ({type(iterable)!r})"
+            )
+        )
+
     if isinstance(iterable, dict):
         if allow_record:
             return _impl(
@@ -89,6 +98,7 @@ def _impl(iterable, highlevel, behavior, allow_record, initial, resize):
                 )
             )
 
+    # Ensure that tuples are treated as iterables, not records
     if isinstance(iterable, tuple):
         iterable = list(iterable)
 
--- a/src/awkward/record.py
+++ b/src/awkward/record.py
@@ -5,6 +5,7 @@ import copy
 from collections.abc import Iterable
 
 import awkward as ak
+from awkward._backends import Backend
 from awkward._behavior import get_record_class
 from awkward._layout import wrap_layout
 from awkward._nplikes.numpylike import NumpyMetadata
@@ -246,6 +247,16 @@ class Record:
             return tuple(contents)
         else:
             return dict(zip(self._array.fields, contents))
+
+    def to_backend(self, backend: Backend | str | None = None) -> Self:
+        if backend is None:
+            backend = self._array.backend
+        else:
+            backend = ak._backends.regularize_backend(backend)
+        if backend is self._array.backend:
+            return self
+        else:
+            return Record(self._array._to_backend(backend), self._at)
 
     def __copy__(self) -> Self:
         return self.copy()


src/awkward/operations/ak_to_layout.py
```diff
@@ -113,7 +113,7 @@ def _impl(array, allow_record, allow_other, regulararray):
         return ak.operations.from_arrow(array, highlevel=False)
 
     elif isinstance(array, (str, bytes)):
-        return ak.operations.from_iter([array], highlevel=False)[0]
+        return ak.operations.from_iter(array, highlevel=False)
```
jpivarski: Does this still work?

```python
want_to_broadcast["new field"] = b"single value"
```

or this?

```python
ak.zip({"some things": ak.Array(...), "another": b"single value"})
```

We want to see bytestrings and strings as atoms for broadcasting, and they may be provided as Python objects that we want interpreted as `ak.Array([b"single value"])`.

src/awkward/operations/ak_from_iter.py
```diff
@@ -98,14 +109,16 @@ def _impl(iterable, highlevel, behavior, allow_record, initial, resize):
...
         backend="cpu",
         byteorder=ak._util.native_byteorder,
-        highlevel=highlevel,
+        highlevel=False,
         behavior=behavior,
         simplify=True,
-    )[0]
+    )
+
+    return wrap_layout(outer[0], highlevel=highlevel, behavior=behavior)
```
jpivarski: This doesn't change behavior, does it? It looks to me like it's neutral, but is using `wrap_layout` to be consistent with other functions that do so.

src/awkward/operations/ak_from_iter.py
```diff
@@ -72,6 +75,13 @@ def from_iter(
 
 
 def _impl(iterable, highlevel, behavior, allow_record, initial, resize):
+    if not isinstance(iterable, Iterable):
+        raise ak._errors.wrap_error(
+            TypeError(
+                f"cannot produce an array from a non-iterable object ({type(iterable)!r})"
+            )
+        )
```
jpivarski, line 82: As I understand it, this is the only part that does what the PR description says it's doing: excluding the possibility of passing non-iterables to `from_iter`.

jpivarski (commented):
Just comments for now because I'm a little wary of this one. I get that you're trying to make the function name more literal, and therefore more guessable without recourse to the documentation. I'm hesitant because there might be some things people expect would work, even though they're not literally iterables.

If the tests pass, then if there are any such cases, they're not explicitly being tested.

src/awkward/operations/ak_from_iter.py
agoose77: It *does* change behaviour, and I originally documented it in another pr. Oops!

If we don't do this, then behaviour classes with custom get item overloads can return non layout types. In my view the way we're using getitem here is an implementation detail rather than something users should play around with.

The specific example of what breaks without this change is when a string is passed into `from_iter` - in main, this returns a python string. With this PR, it returns an array of characters (strings are iterables of char).

I think that's a good change, though it is be breaking if users want a python string to round trip. It's your call whether this is a bug fix!

jpivarski: If a string/bytestring is passed directly into `ak.from_iter` or `ak.Array` without being wrapped in a length-1 list, it wouldn't be too surprising to see it turned into an array of characters. But what about when it's being picked up implicitly as in the examples in https://github.com/scikit-hep/awkward/pull/2353#discussion_r1156261075?

Replace the string/bytestring with a scalar value like `3`, and you'd expect it to broadcast as though it were `[3]`. A string in a broadcasting situation should behave like an atom, like `3`, not as a list of characters that would have to have the same length as the array it's being broadcasted against.

```python
all_my_data["new field"] = "label"
```

looks like it ought to duplicate `"label"` and attach it to `all_my_data` as a `new field`. If this tries to broadcast it instead, it fail if the length of `all_my_data` is not equal to the length of `"label"`, a confusing situation whether it works or doesn't.

Is this going to happen, considering that `"label"` has to be turned into an Awkward Array before the broadcasting can take place, and that conversion probably uses `ak.from_iter`?

jpivarski: It does, but it also does so in `main`.

```python
>>> import awkward as ak
>>> all_my_data = ak.Array([{"x": 1}, {"x": 2}, {"x": 3}, {"x": 4}, {"x": 5}])
>>> all_my_data["new field"] = "label"
<Array [{x: 1, 'new field': 108}, ..., {...}] type='5 * {x: int64, "new fie...'>
>>> all_my_data.show()
[{x: 1, 'new field': 108},
 {x: 2, 'new field': 97},
 {x: 3, 'new field': 98},
 {x: 4, 'new field': 101},
 {x: 5, 'new field': 108}]
```

It didn't in Awkward 1.10.2.

```python
>>> import awkward as ak
>>> all_my_data = ak.Array([{"x": 1}, {"x": 2}, {"x": 3}, {"x": 4}, {"x": 5}])
>>> all_my_data["new field"] = "label"
>>> all_my_data.tolist()
[
    {'x': 1, 'new field': 'label'},
    {'x': 2, 'new field': 'label'},
    {'x': 3, 'new field': 'label'},
    {'x': 4, 'new field': 'label'},
    {'x': 5, 'new field': 'label'},
]
```

Do you agree that this is a bug? I can open an issue, but I don't know if this PR is going to be a step toward or away from that.

agoose77: Yes, this behaviour is a bug. I authored this PR as part of a larger one that needs the ability to promote scalars to arrays. This behaviour overlaps with it, so I'll think about how to connect the two.

src/awkward/operations/ak_from_iter.py
agoose77: Let's constrain this PR to _just_ that change.

jpivarski (approved):
I see that the PR is only about rejecting non-iterables, and you can merge when ready.

Should I turn the problem with broadcasting strings into an issue?

PR #2354, opened 2023-04-03
fix: convert ufunc arguments to same backend

agoose77:


jpivarski (approved):
Definitely. You can merge this when ready.

PR #2357, opened 2023-04-04
docs: update mybinder link to example notebook

raybellwaves:
I didn't find the link to https://mybinder.org/ very helpful here. I think it makes sense to like the mybinder where you can run awkward


jpivarski (approved):
Thanks, that looks like an oversight.

PR #2358, opened 2023-04-04
docs: add import statement

raybellwaves:
Added a import statement here to make it more copy-pastable


jpivarski (approved):
Okay, this was being brief to minimize screen clutter, but you're right that being able to copy-paste it as-is is nice.

jpivarski:
@all-contributors please add @raybellwaves for docs

PR #2359, opened 2023-04-05
refactor: improve broadcasting logic readability

agoose77:
This PR splits the continuation in broadcasting into separate closures that we can more easily refactor.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -442,7 +442,7 @@ def apply_step(
     # This whole function is one big switch statement.
     def continuation():
         # Any EmptyArrays?
-        if any(isinstance(x, EmptyArray) for x in inputs):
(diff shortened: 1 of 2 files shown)


src/awkward/_broadcasting.py
```diff
@@ -1006,19 +945,96 @@ def continuation():
...
+        # Any list-types?
+        elif any(x.is_list for x in contents):
+            return broadcast_any_list()
+
+        # Any RecordArrays?
+        elif any(x.is_record for x in contents):
+            return broadcast_any_record()
 
         else:
             raise ak._errors.wrap_error(
```
jpivarski, line 1040: In the end, this is the summary, the high-level overview of how broadcasting works. Splitting it out like this really helps to see this big picture! (And it's safer in that it limits variable scope.)

jpivarski (approved):

PR #2365, opened 2023-04-05
feat: add `ak.enforce_type`

### LINKED ISSUE #1805: Allow specifying type for empty Array

The API does not allow one to create an empty ak.Array with a given type. It is missing the equivalent of a dtype argument.

```
a = Array([[1.2], []])  # succeeds, type inferred
b = Array([])  # succeeds, but type '0 * unknown'
```

Please allow the user to pass a `type` keyword, so that they can create empty arrays with an appropriate type.

### LINKED ISSUE #2338: Make a function that assigns types to unknowns from an overall type/Form description
Note to self: the "prototype method" of assigning a type is a good hint of how to write that type-setting function. We have a way of turning a Form into a length-0 array (it's an L2 function), and if you concatenate that length-0 array onto the array with unknowns, it would give the appropriate type to those unknowns. If the types conflict, you'd get unions, which might be the right thing.

This should be turned into an issue, for brainstorming how to make that function.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/issues/2337#issuecomment-1481370021_

agoose77:
> **Warning**
> Depends upon #2424 and #2425

Closes #1805, closes #2338 (?) by adding a function `ak.enforce_type` that converts an existing layout to a new high-level type. The following rules apply:

- Indexed layouts are projected out if their type changes.
- Option types can be either:
  - added
  - removed (if there are no missing values)
- Unions can either:
  - grow
  - shrink (if the lost content(s) are not referenced by any tags)
  - remain equivalent
  - change by a single content
  - project to a single type
  - convert to a single type
- Records can be converted:
  - tuple→tuple 
     - N contents to M types (N < M) if the trailing (M-N) types are options
     - N contents to M types (N > M) drops the trailing (N-M) contents
  - record→record
      - matching fields are kept, missing fields dropped, new fields added (but must be option types)
- List types can become regular/irregular (regularity is validated with known data) 
- NumPy arrays can change dtypes
- Unknown contents (`EmptyArray`) can:
  - be removed at any node,
  - be added at any node, provided that they're wrapped in an option


jpivarski:
It's looking good so far!

src/awkward/operations/ak_almost_equal.py
```diff
@@ -105,35 +105,89 @@ def visitor(left, right) -> bool:
...
-                        left.data, right.data, rtol=rtol, atol=atol, equal_nan=False
+            # If one type is timelike, so must the other be, etc.
+            if (
+                np.issubdtype(left.dtype, np.datetime64)
+                ^ np.issubdtype(right.dtype, np.datetime64)
+            ) or (
+                np.issubdtype(left.dtype, np.timedelta64)
+                ^ np.issubdtype(right.dtype, np.timedelta64)
+            ):
+                return False
```
agoose77: Should we support unit conversions in this function, e.g. seconds evaluated against minutes? If not, then we can remove this part.

src/awkward/operations/ak_almost_equal.py
```diff
@@ -105,35 +105,89 @@ def visitor(left, right) -> bool:
...
+
+            # Now project out the contents, and check for equality
+            for i, j in zip(left_tag_order, right_tag_order):
+                if not visitor(
+                    left.project(i).to_packed(), right.project(j).to_packed()
+                ):
+                    return False
+            return True
+
```
agoose77: Previously, this logic did not accommodate unions with different orderings of `contents` (and correspondingly re-ordered `tags`).

src/awkward/operations/ak_almost_equal.py
jpivarski: Converting time units, like seconds into minutes, would be in the same category as converting between other dtypes, like `np.float32` to `np.int64` (and may be just as lossy). I'd expect this to be in scope. Is it even any different from converting between other dtypes?

It seems to be the same:

```python
>>> np.array(["2021", "2022", "2023"], "datetime64[15s]").astype("datetime64[12m]")
array(['2021-01-01T00:00', '2022-01-01T00:00', '2023-01-01T00:00'],
      dtype='datetime64[12m]')
```

so there's no reason to _prevent_ it.

src/awkward/operations/ak_almost_equal.py
jpivarski: Does that mean that this is a bug-fix on a PR about a distinct new feature? (Ideally, they'd be separated, but in a pinch, combining them is okay.)

agoose77: A bugfix that this test suite uses. I'll move it into a new PR that sequences before this one.

src/awkward/operations/ak_almost_equal.py
agoose77: This particular fragment pertains to `almost_equal`. For datetimes, it doesn't make sense to define a "nearly equal" if we don't have both datetimes or deltas. However, as for allowing different units, I need to check NumPy's behaviour more closely in the analogous functions :)

jpivarski (approved):
I read over the tests, the main public function API, and the coding style within `ak.enforce_type`, and everything looks very nice! Let me know if there are particular points you want me to evaluate in detail.

agoose77:
@jpivarski fab! Thank you, it's a lot of code. I'm heading off until tomorrow; I plan to revisit the rules for records and tuples.

Given that record fields can't be renamed with this function, record - > record shouldn't need to compute permutations. Similarly, if we require that tuples can't be reordered), we don't need permutations there. I think removing these d.o.f should simplify things.

agoose77:
@jpivarski fab! Thank you, it's a lot of code. I'm heading off until tomorrow; I plan to revisit the rules for records and tuples.

Given that record fields can't be renamed with this function (or fields added/removed), record → record shouldn't need to compute permutations. Similarly, if we require that tuples can't be reordered), we don't need permutations there. I think removing these d.o.f should simplify things.

agoose77:
So, the original implementation of this function permuted record contents, and required that the permutation had matching fields if both layout and type were tuples. I don't think re-ordering records makes sense for this function; that's a fundamentally different operation at which we can only guess by convertability.

So, now this function respects fields/positions.

agoose77:
@jpivarski the motivation for the "thunk" pattern was to separate convertibility tests from the actual conversion logic. However, the only practical need for this was for the branches that occured with unions and records.

In the above commits & comments, you can see that I decided to make records *and* unions more strict; records cannot convert between tuple-record, and unions must now be strictly equal to the type in all "existing" contents, unless the total number of contents does not change (in which case, up-to one content can differ in order to be converted).

Regarding these changes for unions, it makes reasoning about them far simpler; we only have to consider type equality, rather than type _enforceability_.

As such, the need for a two-phase conversion is gone, and this logic is much easier to write as a direct recursive function. I found the need for a `layout_equals_type` function; although we can just write `layout.form.type.is_equal_to`, I suspect we want to avoid generating the type at multiple points during the tree.

Now, an existing example of where we keep the separate "can I do X" and "do X" logic is the the content merging protocol, which is defined over `_mergeable_next` and `_mergemany`. I don't think the motivations for separating these two processes holds here; the mergeable logic essentially looks like:
1. is X mergeable with Y?
2. if so, merge, else create union
3. proceed ...

whereas in our case, if something _isn't_ enforceable to `T`, that's an error (and we blow up).

So, I'm happy that this code no longer separates those two code pathways.

N.B., to be explicit I'm not talking about "thunks" vs two separate functions, rather I'm referring to the need to have "can I do X" at all.

agoose77:
@jpivarski this is converging on a final implementation. I'd benefit from the following:
- Thoughts on whether `layout_equals_type` should be a public function, e.g. defined on `layout` as `def is_equal_to_type()`?
- Thoughts on rules regarding unions and records (see main PR description for details)

src/awkward/operations/ak_almost_equal.py
jpivarski: Yes, datetimes (of any unit) can only be compared with datetimes, and timedeltas (of any unit) can only be compared with timedeltas.

jpivarski:
> * Thoughts on whether `layout_equals_type` should be a public function, e.g. defined on `layout` as `def is_equal_to_type()`?

Just for wording, a layout _has_ a type, but it can't be _equal to_ a type. It's not a big deal for the internal function, but would be for an external API.

If an array has both `ak.Array.type == t` and `ak.Array.layout.has_type(t)`, users will wonder which one they should use. It looks to me like `layout_equals_type` does more than just type equality because it allows for some fields to be `unknown_length` and the types are still considered equal. From inside the codebase, it can be clear that such a thing is needed, but from outside the codebase, that's a subtle difference.

I think we should _not_ expose this as a public API unless it's needed for some application. Before we get to this one, dask-awkward and Coffea are reaching into L3 functions, and those are functions with demonstrated need to be L2.

> * Thoughts on rules regarding unions and records (see main PR description for details)

A few:

> * Indexed layouts are projected out.

Are they necessarily projected out? If the node above the IndexedArray and the node below the IndexedArray do not need to be changed (determined via types only, not values), how about if the IndexedArray is left as-is? Such a determination can only be made on the bottom-up return-pass of the recursion (i.e. after recursing down, when you're on the way back up again, you'll be able to tell that the IndexedArray's `content` didn't change).

The reason that's valuable is because it minimally touches RecordArrays with many fields; maybe most of them are unaffected. The RecordArray may be within an IndexedArray (a common case).

> * Option types are packed, and can be either:
>   * added
>   * removed (if there are no missing values)

Great!

> * Unions can either:
>   * grow
>   * shrink (if the lost content(s) are not referenced by any tags)
>   * remain equivalent
>   * change by a single content

Great!

> * Records can be converted:
>   * tuple→tuple
>   * record→record (fields must match)

Maybe fields of a tuple can be dropped if they're at the end (tuples get shorter). That might be expected, though I can't think of any use-cases in which it would matter.

For record→record, it can be converted into a subset of fields, right?

Also for record→record, can it be converted into a superset of fields if the new fields are option-type? That would be a useful feature.

> * NumPy arrays can change dtypes

Great!

> * 1D list types cannot be changed

What about

  * regular → variable-length (always)
  * variable-length → regular (runtime exception if the lengths don't match)?

We already have `ak.from_regular` and `ak.to_regular`, but forcing users to do that part of the type conversion as a separate step (and figure out the `axis` parameters) is unfriendly.

> * Unknown contents (`EmptyArray`) can be added/removed at any node

By "removing EmptyArray" (outside the context of a RecordArray field), do you mean "convert EmptyArray into any type"? That's what I would expect: just create a zero-length array from the Form and replace the EmptyArray with that.

agoose77:
> List conversion is already there - I can see how my description _didn't_ give that thought! Let me update it. 

Records and tuples can drop fields, let's add this. I didn't notice that we can introduce none's for _new_ contents - that would be a good feature too! I hadn't originally thought it would be sensible / possible.

Currently we touch all indexed nodes if we walk into the layout, and we walk into all nodes to set parameters. We could perhaps early exit if a type matches, but that would be double walking the tree at each level.

jpivarski:
Double-walking is better than touching, since double-walking is metadata-only and touching means that a whole column will be loaded.

Setting parameters shouldn't touch any buffers; that's a different part of the `Content` instance. If you need to make a new `Content` instance with different parameters, it can just reference the `Index` or buffer that the original `Content` had without touching.

The usefulness of adding new record fields that are all `None` is because this function will be used in normalizing dynamic types as specific static types. Suppose some system doesn't make attributes if their value would be missing (i.e. `{"x": 123}` instead of `{"x": 123, "y": null}`).

  * If all instances of the attribute are present, ArrayBuilder would make the field non-option-type, and the user would need to insert their knowledge that it should be option-type with no missing values.
  * If only some instances of the attribute are present, ArrayBuilder would make the field option-type (instead of making a union of the record with and without the field: that was a choice).
  * If no the instances of the attribute are present, ArrayBuilder would not and could not make the field. The user would need to insert their knowledge that it should be there.

So I think this would come up quite a bit.

agoose77:
> Setting parameters shouldn't touch any buffers; that's a different part of the Content instance. If you need to make a new Content instance with different parameters, it can just reference the Index or buffer that the original Content had without touching.

Absolutely. The point I was making above was that a naive single-walk implementation will have to project index types, even if this is never required by the leaves. I've added a type check that will double-walk the tree for any indexed types encountered. I imagine this won't be too bad for most layouts. We could easily weakly memoize it if needs be down the road.

agoose77:
> 1D list types cannot be changed

Ah, now I'm back at my keyboard, I see why this was confusing. This is an obscure reference to strings - they're 1D lists according to `purelist_depth`.

agoose77:
> * If only some instances of the attribute are present, ArrayBuilder would make the field option-type (instead of making a union of the record with and without the field: that was a choice).

Crucially this point; I had forgotten about all-None options being possible stand-ins! 

> If an array has both ak.Array.type == t and ak.Array.layout.has_type(t), users will wonder which one they should use. It looks to me like layout_equals_type does more than just type equality because it allows for some fields to be unknown_length and the types are still considered equal. From inside the codebase, it can be clear that such a thing is needed, but from outside the codebase, that's a subtle difference.

Yes, this is my concern as well. Actually, type equality also supports these kind of options (though we do need to do a pass over our equality operations to ensure they all have the same semantics), but in this particular case the sole motivation is performance.

agoose77:
TODO: allow strings to move between strings and bytestrings, and allow string abstraction to be erased?

agoose77:
## Discussion of code design exploration

This function was harder to write than I anticipated. Not because it's inherently difficult, but writing readable code that covers all optimisation paths is fiddly.

For the problem of avoiding projections unless absolutely necessary, there are two solutions; only project if the type changes (related to #2434 — is it worse to operate on unseen parts of a buffer, or to project each layout on the way down?), or propagate a flag _upwards_ that indicates if projection (for the immediate child) is required. This PR implements the former, which means that we'll likely walk the layout multiple times. Therefore, if this is slow, we should use a weak LRU cache to store the results of the layout-type check.

The second approach is one that I played around with. This would use something _like_ the thunk pattern again, though I realised that generators are a perfect mechanism for implementing this, as they:
- can yield multiple values
- can resume operation with new inputs

The thunk pattern that I used before to separate type-data validation would have been more ergonomic using generators too, i.e. instead of
```python
def build_some_layout(layout, type_):
    build_child = builder_of(layout.content, type_.content)
    if layout_has_wrong_type(layout, type_):
        raise TypeError
    else:
        def thunk(this):
            packed_layout = this.to_packed()
            return packed_layout.copy(content=build_child(packed_layout.content))
        return thunk
```
we could write
```python
def build_some_layout(layout, type_):
    builder_child = builder_of(layout.content, type_.content)
    # Wake generator
    builder_child.send(None)
   
    if layout_has_wrong_type(layout, type_):
        raise TypeError
    else:
        # Receive from sender (parent)
        this = yield
        packed_layout = this.to_packed()
        # Send child layout to child builder, get result
        new_content = builder_child.send(packed_layout.content)
        # Yield result to recipient (parent)
        yield packed_layout.copy(content=new_content)
```
This same pattern lets us yield e.g. whether the content will need packed data, e.g.
```python
def builder_any_option(layout, type_):
    content_builder = builder_for(layout, type_.content)
    # Does our child need packed data?
    needs_packed = next(content_builder)
    # We don't need it ourselves, so just yield child's request
    layout = yield needs_packed
    yield ak.contents.UnmaskedArray(layout, parameters=type_.parameters)
```

The proper way that generators are used is to _return_ the result, which is stored in the `value` attribute of the thrown `StopIteration` error. However, using exceptions for flow control is not ideal, as it's slow. Python avoids this in the normal case by bypassing the exception handling machinery, but we'd need to manually catch the exception with `try: except:`. Though one could make that ergononic, it would still be slow.

agoose77:
@jpivarski I've implemented all of the changes above. See the description for the current set of rules.

The use of `_layout_has_type` may be a performance problem. Right now, I'm assuming not, and we can optimise it if so. 

Could you give this another review, just to see whether anything stands out?

jpivarski:
Are you sure that using exceptions for control flow is slow? That may be in C++, since exceptions require signal handling or an OS call or something, but in Python, exceptions are not fundamentally different from other control flow.

Also, I've heard explicit claims that option exceptions are not slower than other instructions, and can therefore be used in this way. I don't _like_ to do that because the syntax of the block is cumbersome, but that can be fixed by hiding it in a function call.

agoose77:
> Are you sure that using exceptions for control flow is slow? That may be in C++, since exceptions require signal handling or an OS call or something, but in Python, exceptions are not fundamentally different from other control flow.

Running inside a try-except is slower irrespective of whether the exception is raised. In Python 3.11, there is [a fast-path](https://bugs.python.org/issue40222) that reduces the cost of unhandled exceptions.

> Also, I've heard explicit claims that option exceptions are not slower than other instructions, and can therefore be used in this way. I don't like to do that because the syntax of the block is cumbersome, but that can be fixed by hiding it in a function call.

I don't have any numbers, but I'm fairly confident that writing EAFP is slower if the common-case is to raise an exception. In the context of the above, the exception would _always_ be raised.

```ipython
In [4]: d = {}
   ...:
   ...: def f1():
   ...:     if 0 in d: 
   ...:         return d[0]
   ...:     else: 
   ...:         return None
   ...:
   ...: def f2():
   ...:     try: 
   ...:         return d[0]
   ...:     except KeyError: 
   ...:         return None
   ...:

In [5]: %timeit f1()
86.8 ns ± 0.589 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [6]: %timeit f2()
230 ns ± 4.68 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)
```

jpivarski:
I've heard wrong. I guess, then, that it's good that I've been avoiding it. This is one of the things people used to say about Python—that it makes exceptions lightweight and we should use them for everything—but perhaps that was just a relative statement, for people coming from C++.

jpivarski (approved):
I see that the thunk-based organization is gone, which does make this easier to reason about, coming into it without background knowledge. (+1 for simple code.)

The tests are exhaustive. I'd say that this feature is ready. There were several people who were waiting for this; Hans, AnnData, and maybe another. Let's see if it works for them!

agoose77:
> I've heard wrong. I guess, then, that it's good that I've been avoiding it.

Not entirely; if you don't hit the slow path, it's not as bad. After all, Python encourages EAFP than LBYL. 

That said, I can't reproduce these numbers at all today :exploding_head: I'm actually seeing the reverse:
```ipython
In [7]: d = {}
   ...:
   ...:
   ...: def f1():
   ...:     if 0 in d:
   ...:         return d[0]
   ...:     else:
   ...:         return None
   ...:
   ...: def f2():
   ...:     try:
   ...:         return d[0]
   ...:     except KeyError:
   ...:         return None
   ...:

In [8]:

In [8]: timeit f1
26.4 ns ± 0.19 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [9]: timeit f2
18 ns ± 0.243 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [10]: d = {0:0}

In [11]: timeit f1
26 ns ± 0.25 ns per loop (mean ± std. dev. of 7 runs, 10,000,000 loops each)

In [12]: timeit f2
17.8 ns ± 0.192 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
```

I might be missing something obvious, but I'm inclined to suggest that my system was running something else in the background yesterday.

I'm curious now as to what happened yesterday (it was many factors slower for a start). In any case, prior discussions on Python have been that exceptions are slower but not sufficiently that it's better to LBYL. Obviously, "slow" is application-specific, and it might not be a problem for us. 

I'm going to do some more tests on this, because I'm curious to get a robust feel for it. Perhaps the situation has changed (Python 3 has seen a lot of changes).

agoose77:
There are actually two manners in which we might want to "erase" a union:
1. Projecting to a single variant of a union, e.g. `union[int64, string] → string`
2. Converting to a common type, e.g. `union[{x: int64}, {y: int32}] → {x: ?int64, y: ?int32}`

This PR previously implemented (1), but (2) is also useful (e.g. AnnData's case).

<details><summary>Discussion on union erasure</summary>
I was playing around with some demos to see whether this function covers all useful bases, and noticed a fairly glaring omission (hence my recent commits).

For `union[X, Y] → Z`, we previously had the requirement that `Z ∈ {X, Y}`. This rule is a "projection rule" — the first variant type that matches this should be the only type actually referenced by the union tags. This is predictable for typetracer and non-typetracer.

However, this means that we cannot erase a union through conversion, which is a motivator for something _like_ this function, e.g.
```python
a1 = ak.Array([{"a": "aaa", "b": "bbb"}])
a2 = ak.Array([{"b": "bbb", "c": "ccc"}])
array = ak.concatenate([a1, a2])
result = ak.enforce_type(array, "{a: ?string, b: string, c: ?string}")
```

Clearly, there are two use cases:
1. Projecting to a single variant of a union, e.g. `union[int64, string]`
2. Converting to a common type, e.g. `union[{x: int64}, {y: int32}] → {x: ?int64, y: ?int32}`

One solution is to first see whether it's possible to convert from _all_ contents to a common type. If this fails (by type), then we assume it's projection rather than conversion. This is predictable only from type information. However, it would fail when the user wants to project a union to a single type, but the union _looks_ (at type level) to be convertible . 

E.g.
```python
import awkward as ak

this = ak.contents.RecordArray(
    [
        ak.contents.RegularArray(
            ak.contents.NumpyArray([0, 1, 2]),
            3
        ),
        ak.contents.RegularArray(
            ak.contents.NumpyArray([0, 1, 2]),
            3
        )
    ],
    ["x", "y"]
)

that = ak.contents.RecordArray(
    [
        ak.contents.ListOffsetArray(
            ak.index.Index64([0]),
            ak.contents.NumpyArray([3])
        )
    ],
    ["y"]
)

array = ak.concatenate(
    [
        this, that
    ]
)
res = ak.enforce_type(array, "{y: option[3 * int64]}")
```

Another solution is to let the user signal their intent. The easiest way to do this is with a flag (`erase_union_by_projection`) passed to `ak.enforce_type`. Obviously, this then imposes this rule upon all union erasures, but I think it might be the most straightforward. Alternatively, we could re-use the parameter system to localise the setting to a specific type branch, but that feels hacky to me (having parameters be reserved for an operation, and not actually appear on the result).
</details> 

@jpivarski to support both erasure mechanisms, I've implemented a `union_erasure` enum argument that lets users disambiguate this case. Do you like this UX, or can you think of a better approach?

jpivarski:
Yes, (2) is also useful. I can see `ak.enforce_type` being a superset-tool that includes the functionality of others, like `ak.merge_union_of_records`. The difference with the all-in-one tool is that the user has to write a whole type, while `ak.merge_union_of_records` knows what type you need.

I'll get back to your second question after I look at the code again.

jpivarski:
Okay, yeah: why does the user need to specify more information than the desired type (that is, setting the `union_erasure` enum)?

Isn't it the case that if we have a "convert" situation, asking for "project" would just be an error/mismatching type, and vice-versa?

jpivarski:
If `array` has type

```
union[{x: int64}, {y: int32}]
```

and the user asks for type

```
{x: ?int64, y: ?int32}
```

then only "convert" will work. If the user asks for type

```
{x: int64}
```

then only "project" will work. I'm trying to think of counterexamples or ways to claim this more generally, but I don't see how there would be an ambiguous case, something that the enum would be needed to disambiguate. The implementation could just enter both cases and follow the one that succeeds (or points out that both fail in the error message).

I'm a little worried about it from a UX perspective because it would be hard for a user to know that what they want to do is "convert" or that what they want to do is "project"—they'll have an idea in mind of what they want, but won't know the vocabulary. Examples in documentation that don't look quite like their own example won't be helpful. It would be best if we could just take the exact type they want to convert into as given and do anything possible to make it that type (though I understand it's a problem if two different _array_ outputs can be made to match the same type, by following different methods).

agoose77:
Maybe it's not a problem. I was worried about converting an empty list to a regular array; the regular type of an empty list has `size=0`. That doesn't merge against other regular arrays. 

But, maybe just being a little smarter there is all we need.

agoose77:
OK @jpivarski this is now nearly *done*.

I've summarised the rules in the docstring.

Since your last review, I've:
* added a new rule - introducing `option[unknown]` in place of any existing type
* removed the `union_erasure` flag in favour of figuring this out from the type[^note]
* added docstrings with doctest examples
* added an `_type_is_enforceable` which tests enforceability _and_ whether the operation will require a projection of the current node contents
* fixed some bugs 

[^note]: I was worried about some corner cases here, such as length-zero regulararrays. After some experimenting, I have since convinced myself that we are able to do this robustly.

If you get 5 minutes, I'd appreciate if you could play around with some conversions to ensure it works :P

On my remaining todo list:
- check typetracer arrays
- strictly check layouts, rather than `almost_equal`

jpivarski (approved):
This looks great! I heartily approve it for merging!

You said "nearly done," so I'll let you decide when to merge it.

agoose77:
I've now finished with this for now. There are no typetracer tests, but that is because the layout equality checks need to be extended to handle this. I'll do so in a follow up PR.

PR #2370, opened 2023-04-06
refactor: wrap exceptions at catch time

### LINKED ISSUE #2369: Duck-type operations on `ak.Array` and `ak.Record` stringify the array
### Version of Awkward Array

main

Python's `hasattr` function works by catching an `AttributeError`. We intercept missing attributes for `ak.Array` and `ak.Record`, and throw a rewritten exception in its place.

i.e.
```python
hasattr(array, "no-key")
```
invokes the array-repr machinery due to traceback formatting.

This is bad for several reasons (performance, typetracer, ...)

agoose77:
Fixes #2369 by deferring stringifying until the top-level exception is caught by the error context. These exceptions leave Awkward's scope, and therefore count as true exceptions. Otherwise, we're decorating exceptions that might not be shown to the user, doing more busywork as a result!

NB, there are other ways to fix #2369, but this one also removes a lot of boilerplate!

### DIFF
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -1,6 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 from __future__ import annotations
 
+import sys
 import threading
 import warnings
 from collections.abc import Mapping, Sequence
@@ -50,24 +51,39 @@ class ErrorContext:
         try:
             # Handle caught exception
             if exception_type is not None:
-                raise self.rewrite_exception(exception_type, exception_value)
+                self.handle_exception(exception_type, exception_value)
         finally:
             # Step out of the way so that another ErrorContext can become primary.
             if self.primary() is self:
                 self._slate.__dict__.clear()
 
-    def rewrite_exception(self, cls: type[E], exception: E) -> E:
-        if issubclass(cls, (NotImplementedError, AssertionError)):
-            # Raise modified exception
-            exc = cls(
-                str(exception)
-                + "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
-            )
-            exc.__cause__ = exception
+    def handle_exception(self, cls: type[E], exception: E) -> E:
+        if sys.version_info >= (3, 11, 0, "final"):
+            return self.decorate_exception(cls, exception)
         else:
-            exc = cls(self.format_exception(exception))
-            exc.__cause__ = exception
-        return exc
+            raise self.decorate_exception(cls, exception)
+
+    def decorate_exception(self, cls: type[E], exception: E) -> E:
+        if sys.version_info >= (3, 11, 0, "final"):
+            if issubclass(cls, (NotImplementedError, AssertionError)):
+                exception.add_note(
+                    "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
+                )
+            else:
+                exception.add_note(self.note)
+            return exception
+        else:
+            if issubclass(cls, (NotImplementedError, AssertionError)):
+                # Raise modified exception
+                new_exception = cls(
+                    str(exception)
+                    + "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
+                )
+                new_exception.__cause__ = exception
+            else:
+                new_exception = cls(self.format_exception(exception))
+                new_exception.__cause__ = exception
+            return new_exception
 
     def format_argument(self, width, value):
         from awkward import contents, highlevel, record
@@ -144,6 +160,10 @@ class ErrorContext:
     def format_exception(self, exception: Exception) -> str:
         raise NotImplementedError
 
+    @property
+    def note(self) -> str:
+        raise NotImplementedError
+
 
 class OperationErrorContext(ErrorContext):
     _width = 80 - 8
@@ -191,6 +211,10 @@ class OperationErrorContext(ErrorContext):
         return out
 
     def format_exception(self, exception):
+        return f"{exception}.\n{self.note}"
+
+    @property
+    def note(self) -> str:
         arguments = []
         for name, valuestr in self.arguments.items():
             if isinstance(name, str):
@@ -199,8 +223,7 @@ class OperationErrorContext(ErrorContext):
                 arguments.append(f"\n        {valuestr}")
 
         extra_line = "" if len(arguments) == 0 else "\n    "
-        return f"""{exception}.
-
+        return f"""
 This error occurred while calling
 
     {self.name}({"".join(arguments)}{extra_line})"""
@@ -245,8 +268,11 @@ class SlicingErrorContext(ErrorContext):
         return out
 
     def format_exception(self, exception):
-        return f"""{exception}.
+        return f"{exception}.\n{self.note}"
 
+    @property
+    def note(self) -> str:
+        return f"""
 This error occurred while attempting to slice
 
     {self.array}
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -199,11 +199,11 @@ class OperationErrorContext(ErrorContext):
                 arguments.append(f"\n        {valuestr}")
 
         extra_line = "" if len(arguments) == 0 else "\n    "
-        return f"""while calling
+        return f"""{exception}.
 
-    {self.name}({"".join(arguments)}{extra_line})
+This error occurred while calling
 
-Error details: {str(exception)}"""
+    {self.name}({"".join(arguments)}{extra_line})"""
 
 
 class SlicingErrorContext(ErrorContext):
@@ -245,20 +245,15 @@ class SlicingErrorContext(ErrorContext):
         return out
 
     def format_exception(self, exception):
-        if isinstance(exception, str):
-            message = exception
-        else:
-            message = f"Error details: {str(exception)}"
+        return f"""{exception}.
 
-        return f"""while attempting to slice
+This error occurred while attempting to slice
 
     {self.array}
 
 with
 
-    {self.where}
-
-{message}"""
+    {self.where}"""
 
     @staticmethod
     def format_slice(x):
@@ -300,26 +295,14 @@ with
 
 
 def index_error(subarray, slicer, details: str = None) -> IndexError:
-    detailsstr = ""
+    message = ""
     if details is not None:
-        detailsstr = f"""
+        message = f": {details}"
 
-Error details: {details}."""
-
-    error_context = ErrorContext.primary()
-    if not isinstance(error_context, SlicingErrorContext):
-        # Note: returns an error for the caller to raise!
-        return IndexError(
-            f"cannot slice {type(subarray).__name__} with {SlicingErrorContext.format_slice(slicer)}{detailsstr}"
-        )
-
-    else:
-        # Note: returns an error for the caller to raise!
-        return IndexError(
-            error_context.format_exception(
-                f"at inner {type(subarray).__name__} of length {subarray.length}, using sub-slice {error_context.format_slice(slicer)}.{detailsstr}"
-            )
-        )
+    # Note: returns an error for the caller to raise!
+    return IndexError(
+        f"cannot slice {type(subarray).__name__} (of length {subarray.length}) with {SlicingErrorContext.format_slice(slicer)}{message}"
+    )
 
 
 ###############################################################################
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -299,40 +299,6 @@ with
             return repr(x)
 
 
-def wrap_error(
-    exception: Exception | type[Exception], error_context: ErrorContext = None
-) -> Exception:
-    """
-    Args:
-        exception: exception object, or exception type to instantiate
-        error_context: context in which error was raised
-
-    Wrap the given exception, instantiating it if needed, to ensure meaningful error messages.
-    """
-    return exception
-    if isinstance(exception, type) and issubclass(exception, Exception):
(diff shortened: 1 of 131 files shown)


agoose77:
@jpivarski this PR fixes #2369 by handling error rewriting at the bottom of the stack (the error context manager). With Python's existing machinery, there's no nice way <3.10 to rewrite an exception without using `raise exc from old_exc`, which partially duplicates the message. 

Can you live with that? 

Otherwise, in 3.10 we can switch to using the `add_note` API. For <3.10, we could just modify the exception in-place. 

Example traceback:
```pytb
  File "/home/angus/Git/awkward/src/awkward/highlevel.py", line 950, in __getitem__
    out = self._layout[where]
  File "/home/angus/Git/awkward/src/awkward/contents/content.py", line 555, in __getitem__
    return self._getitem(where)
  File "/home/angus/Git/awkward/src/awkward/contents/content.py", line 600, in _getitem
  File "/home/angus/Git/awkward/src/awkward/contents/regulararray.py", line 495, in _getitem_next
    return nextcontent._getitem_next(nexthead, nexttail, advanced)
  File "/home/angus/Git/awkward/src/awkward/contents/indexedarray.py", line 395, in _getitem_next
    return self._getitem_next_field(head, tail, advanced)
  File "/home/angus/Git/awkward/src/awkward/contents/content.py", line 348, in _getitem_next_field
    return self._getitem_field(head)._getitem_next(nexthead, nexttail, advanced)
  File "/home/angus/Git/awkward/src/awkward/contents/indexedarray.py", line 294, in _getitem_field
    self._content._getitem_field(where, only_fields),
  File "/home/angus/Git/awkward/src/awkward/contents/recordarray.py", line 476, in _getitem_field
    return self.content(where)
  File "/home/angus/Git/awkward/src/awkward/contents/recordarray.py", line 411, in content
    out = self.form_cls.content(self, index_or_field)
  File "/home/angus/Git/awkward/src/awkward/forms/recordform.py", line 155, in content
    index = self.field_to_index(index_or_field)
  File "/home/angus/Git/awkward/src/awkward/contents/recordarray.py", line 405, in field_to_index
    return self.form_cls.field_to_index(self, field)
  File "/home/angus/Git/awkward/src/awkward/forms/recordform.py", line 132, in field_to_index
    raise (
awkward._errors.FieldNotFoundError: no field 'z' in record with 2 fields

  File "/home/angus/.config/JetBrains/PyCharm2023.1/scratches/scratch_415.py", line 21, in <module>
    array[0, "z"]
  File "/home/angus/Git/awkward/src/awkward/highlevel.py", line 949, in __getitem__
  File "/home/angus/Git/awkward/src/awkward/_errors.py", line 53, in __exit__
    raise self.rewrite_exception(exception_type, exception_value)
awkward._errors.FieldNotFoundError: while attempting to slice

    <Array [{x: 100, y: 200}, {x: 300, ...}] type='2 * {x: int64, y: int64}'>

with

    (0, 'z')

Error details: no field 'z' in record with 2 fields
```

The "pure" solution here is probably to even avoid repeating the error details, and have the final exception just be the top level "error occured whilst slicing XXX with YYY", but I reckon you'll not go for that!

jpivarski:
Yeah, we assume that most people won't scroll to learn details about an exception. If the last thing in the stack trace only says what function they called with what arguments, there would be no information (at that place) saying what actually went wrong.

Since this PR removes the AK001 `ak._errors.wrap_error` boilerplate, the diff is huge. It would definitely be a plus if the alternative is nicer. But where is the alternative? **What should I look at to see how it's done now?**

I'd be a little uncomfortable with modifying the call stack (what if some other library is also modifying it?) or modifying exceptions in place (if they're not ours, how can we be sure it would work?). Knowing that we can switch to `add_note` when that becomes the minimum Python is comforting—it means that even if the solution isn't great, it won't be permanent. Python 3.10 can become the minimum-supported Python [on October 5](https://endoflife.date/python), [2025](https://youtu.be/zKQfxi8V5FA).

Will it satisfy the original goal of [providing meaningful error reporting for GPU processes that run asynchronously on a background thread](https://github.com/scikit-hep/awkward/discussions/1321)?

src/awkward/_errors.py
```diff
@@ -43,9 +47,27 @@ def __enter__(self):
...
+            # Step out of the way so that another ErrorContext can become primary.
+            if self.primary() is self:
+                self._slate.__dict__.clear()
+
+    def rewrite_exception(self, cls: type[E], exception: E) -> E:
+        if issubclass(cls, (NotImplementedError, AssertionError)):
+            # Raise modified exception
+            exc = cls(
+                str(exception)
+                + "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward-1.0/issues"
```
lgray: Silly comment, but shouldn't you update this URL?

jpivarski: Sure, although it redirects. But this one is being shown to users, so it would be less confusing if it were fixed. I'll put in a Suggested Change.

jpivarski:
```suggestion
                + "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
```

agoose77:
We already re-build exceptions (assuming that we can pass in args) ; I don't think it's much more of a stretch to directly change their args in place. The downside is if someone is holding on to the exception, they'll see that it updates in place.

I think the core point of this PR - to defer formatting of the array until we know the user will see this exception - is the right thing to do; not only for performance - we want to decorate any exception, not just those that we raise ourselves.

We can actually rewrite traceback / frames, but I agree - that's not ideal.

I'll keep thinking on this, though!

src/awkward/_errors.py
```diff
@@ -287,6 +309,7 @@ def wrap_error(
 
     Wrap the given exception, instantiating it if needed, to ensure meaningful error messages.
     """
+    return exception
     if isinstance(exception, type) and issubclass(exception, Exception):
         try:
             exception = exception()
```
jpivarski: Is `wrap_error` now dead code? (It was always L3; we can remove it.)

It looks like this is making it inert by returning the argument and then making everything that comes after it dead. Is that debugging code?

jpivarski (changes requested):
The exception rewrite looks safe: it's not assuming anything particular about the `exception.args`, which can be interpreted in different ways by different exceptions, but only the `__cause__`, which is only interpreted one way.

But it also looks like you turned off `wrap_error` by making it an identity function, and there are statements after the `return` statement. It looks like that's still in a debugging code state.

src/awkward/_errors.py
agoose77: Yes, this was a discussion pr. Must have slipped on the draft button!

agoose77:
I've reworked the error message ordering, to better support the "decorate an existing error" approach that will be favoured by `add_note`:
```pytb
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 119, in count
    return _impl(array, axis, keepdims, mask_identity, highlevel, behavior)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 128, in _impl
    out = ak._do.reduce(
  File "/home/angus/Git/awkward/src/awkward/_do.py", line 318, in reduce
    raise ValueError(
ValueError: axis=4 exceeds the depth of the nested list structure (which is 1)

  File "/home/angus/.config/JetBrains/PyCharm2023.1/scratches/scratch_415.py", line 15, in <module>
    ak.count(array, axis=4)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 98, in count
  File "/home/angus/Git/awkward/src/awkward/_errors.py", line 53, in __exit__
    raise self.rewrite_exception(exception_type, exception_value)
ValueError: axis=4 exceeds the depth of the nested list structure (which is 1).

    ak.count(
        array = <Array [{x: 100, y: 200}, {...}] type='2 * {x: int64, y: in...'>
        axis = 4
        keepdims = False
        mask_identity = False
        highlevel = True
        behavior = None
    )
```

You can now easily see the duplicated message `ValueError: axis=4 exceeds the depth of the nested list structure (which is 1)`. I don't hate this, though; the final error message is readable, and Python clearly shows that the original error _causes_ the final error.

Python 3.10 would eliminate the final traceback, and just add notes directly. Here's a similar effect to `add_note`, by overwriting `exception_value.args`:
```pytb
  File "/home/angus/.config/JetBrains/PyCharm2023.1/scratches/scratch_415.py", line 15, in <module>
    ak.count(array, axis=4)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 119, in count
    return _impl(array, axis, keepdims, mask_identity, highlevel, behavior)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 128, in _impl
    out = ak._do.reduce(
  File "/home/angus/Git/awkward/src/awkward/_do.py", line 318, in reduce
    raise ValueError(
ValueError: axis=4 exceeds the depth of the nested list structure (which is 1).

    ak.count(
        array = <Array [{x: 100, y: 200}, {...}] type='2 * {x: int64, y: in...'>
        axis = 4
        keepdims = False
        mask_identity = False
        highlevel = True
        behavior = None
    )

```

jpivarski (approved):

jpivarski:
I like the fact that we're following the `add_note` behavior now, so that we'll be able to move to that exclusively without observable changes in 2025.

I've approved this and it can be merged when ready. I'll take a look at this one test error.

agoose77:
> I like the fact that we're following the add_note behavior now, so that we'll be able to move to that exclusively without observable changes in 2025.

I will just confirm how a 3.11 implementation would look. I also made a mistake; `add_note` is actually a 3.11 feature. I think it might have originally been 3.10, but got bumped?

> I've approved this and it can be merged when ready. I'll take a look at this one test error.

Thanks! It's just a `str(exception)` case that we know will be broken here.

jpivarski:
The test failures are because the error message text has changed. (It's always a dubious thing to make a test depend on the exact text of some human-readable thing, like a `__repr__` or a `__str__`, but it was expedient to do so.)

I was going to just change them so the tests would pass, but I don't think I like the new format. This is `str(err)`:

```
cannot slice EmptyArray (of length 0) with 0: array is empty.

    <Array [] type='0 * unknown'>

with

    0
```

The "details" strings were partial sentences (not capitalized and no period) because that's a Python convention and the error formatting used that. Now it's backward.

But a bigger issue than the grammar/punctuation is that the bottom of a Python stack trace should say what the error is (just as the top of a Java stack trace does): that's where we've been trained to look (and anyone new to Python should learn that convention). The "details" string is the error; the "This error occurred while attempting to slice" is the context. Even though users are more likely to be able to use the "while attempting to slice" information, since it refers to something they can change in their code, we need to have the actual error at the bottom.

jpivarski:
> `add_note` is actually a 3.11 feature

October 4, 2026, then. There isn't a song about that. Although it would be more appropriate to the video, since _Metropolis_ was set in 2026.

agoose77:
Here's the Python 3.11 equivalent using `add_note`:
```pytb
  File "/home/angus/.config/JetBrains/PyCharm2023.1/scratches/scratch_415.py", line 15, in <module>
    ak.count(array, axis=4)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 119, in count
    return _impl(array, axis, keepdims, mask_identity, highlevel, behavior)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/angus/Git/awkward/src/awkward/operations/ak_count.py", line 128, in _impl
    out = ak._do.reduce(
          ^^^^^^^^^^^^^^
  File "/home/angus/Git/awkward/src/awkward/_do.py", line 318, in reduce
    raise ValueError(
ValueError: axis=4 exceeds the depth of the nested list structure (which is 1)

    ak.count(
        array = <Array [{x: 100, y: 200}, {...}] type='2 * {x: int64, y: in...'>
        axis = 4
        keepdims = False
        mask_identity = False
        highlevel = True
        behavior = None
    )
```

I can add support for this _now_ so that we can eventually remove the old code.

jpivarski:
After some discussion on Slack, I've acquiesced. The new `add_note` always adds to the bottom of exception text, so post-3.11 Python users will have to scroll their eyes up to the `ExceptionClassName:`, past the notes that 20 different libraries have all added to it. I think that could get out of hand, but we should adopt the future patterns.

Our "this error occurred" message could get rather long. Hopefully, it doesn't get in the way of people looking for the actual error. In particular, the file path and line number (above `ExceptionClassName:`) is typically the most important information. (For some things, like LaTeX, it is the _only_ information I can ever make use of.)

When originally writing this, I had thought about including an "`ak.whatever` called from" file path and line number, like this:

```

    ak.count(
        array = <Array [{x: 100, y: 200}, {...}] type='2 * {x: int64, y: in...'>
        axis = 4
        keepdims = False
        mask_identity = False
        highlevel = True
        behavior = None
    )

on line 123 of /Users/me/my_awesome_analysis_code.py
```

which would make being on the bottom of the stack trace text even more useful for users who are fixing their code, rather than finding the place in our code where the exception actually happened. I decided against it because it would require us to query the call stack every time an `ak.whatever` function is called. (I haven't measured the performance impact—just wary of it.)

But regardless of this possible improvement, I guess we should put our notes at the bottom, as is standard. So I'm in favor of this PR adhering to the new `add_note` way of formatting the text.

agoose77:
OK, sounds good to me. I'm personally not against this style, but I appreciate that it's not to everyone's liking.

For users, there is https://rich.readthedocs.io/en/stable/traceback.html which is one of several libraries that provides a nicer exception formatter, but must be installed manually (or via `sitecustomize`).

PR #2390, opened 2023-04-11
refactor: implement `Backend` registry (2 of 2)

agoose77:
## TL;DR

> **Warning** 
> Depends upon #2389

#2389 implements a cached lookup for nplikes. This PR implements a similar mechanism for backends. Unlike #2389, this PR implements a lookup registry; types can register a factory function that accepts a type, and returns a finder. This is used to locate backends for e.g. `Content`.

## TL
This PR intends to clarify which objects support `Backend` resolution, and to improve the performance of the search. Although the existing mechanism is trivial, it is called frequently throughout the codebase, so it is important to keep it simple. 
The existing implementation attempts to duck-type our way to find the backends. However, in practice we are constrained in what kind of interface we can define, such that this approach is not scalable. 

This PR replaces kludgy logic with an explicit lookup mechanism that can be specialised per type.


jpivarski (approved):
This is a good internal refactoring, but as a question that must be answered before merging: is this going to affect the Coffea and dask-awkward updates?

Coffea might be using L3 nplike/backend resolution as a stepping stone during development (it has to eventually all be L2, but can be L3 while figuring out what needs to be done, independent of how to interface with it properly).

With this PR, do the Coffea tests still work?

agoose77:
> With this PR, do the Coffea tests still work?

For my local install, `coffea` tests fail on main (both Awkward & Coffea), and if I downgrade `awkward` to v2.1.2, they still fail. The number of failures does not change if I bump `awkward` to this PR. Therefore, I'm comfortable merging it with the assumption that I'm not making things _worse_.

I'll ping @lgray to figure out whether I should be testing a different branch to `main`.

lgray:
@agoose77 let me check - there are some xfails right now because of some strange interactions with dask that are not your fault which I have not marked.

lgray:
@agoose77 coffea main should now pass tests if everything is ok. :-)

agoose77:
I ran coffea's test suite (and dask-awkward) on this branch, and saw no regressions :) Merging!

PR #2393, opened 2023-04-12
fix: clarify scalar promotion in `to_layout`

agoose77:
Unifies scalar type handling in `ak.to_layout` such that scalars are _always_ promoted to 1D arrays. Where possible, e.g. `np.generic` objects, the dtype is preserved.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -64,14 +64,13 @@ def length_of_broadcast(inputs: Sequence) -> int | type[unknown_length]:
     return maxlen
 
 
-def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
+def broadcast_pack(inputs: Sequence) -> list:
     maxlen = length_of_broadcast(inputs)
     nextinputs = []
     for x in inputs:
         if isinstance(x, Record):
             index = x.backend.index_nplike.full(maxlen, x.at, dtype=np.int64)
             nextinputs.append(RegularArray(x.array[index], maxlen, 1))
-            isscalar.append(True)
         elif isinstance(x, Content):
             nextinputs.append(
                 RegularArray(
@@ -81,16 +80,14 @@ def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
                     parameters=None,
                 )
             )
-            isscalar.append(False)
         else:
             nextinputs.append(x)
-            isscalar.append(True)
 
     return nextinputs
 
 
-def broadcast_unpack(x, isscalar: list[bool], backend: Backend):
-    if all(isscalar):
+def broadcast_unpack(x, isscalar: bool, backend: Backend):
+    if isscalar:
         if not backend.nplike.known_data or x.length == 0:
             return x._getitem_nothing()._getitem_nothing()
         else:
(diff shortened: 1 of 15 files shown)


jpivarski (approved):
This is good! (As we talked about in our meeting today.) The tests aren't passing yet, but the intention looks right to me.

src/awkward/operations/ak_isclose.py
```diff
@@ -43,16 +44,20 @@ def isclose(
...
 
 def _impl(a, b, rtol, atol, equal_nan, highlevel, behavior):
-    one = ak.operations.to_layout(a)
-    two = ak.operations.to_layout(b)
+    one, one_is_scalar = _to_layout_detailed(
+        a, allow_record=True, allow_other=False, regulararray=True
+    )
+    two, two_is_scalar = _to_layout_detailed(
+        b, allow_record=True, allow_other=False, regulararray=True
+    )
```
agoose77, line 52: Don't love the verbosity here, but two isn't enough cases for a loop imo.

src/awkward/operations/ak_mask.py
```diff
@@ -104,11 +104,11 @@ def mask(array, mask, *, valid_when=True, highlevel=True, behavior=None):
...
         layoutarray, layoutmask = inputs
-        if isinstance(layoutmask, ak.contents.NumpyArray):
-            m = backend.nplike.asarray(layoutmask)
-            if not issubclass(m.dtype.type, (bool, np.bool_)):
+        if layoutmask.is_numpy:
+            m = layoutmask.data
+            if not np.issubdtype(m.dtype, np.bool_):
                 raise ValueError(f"mask must have boolean type, not {repr(m.dtype)}")
-            bytemask = ak.index.Index8(m.view(np.int8))
+            bytemask = ak.index.Index8(m.view(np.int8)).to_nplike(backend.index_nplike)
```
agoose77, line 111: General cleanups

src/awkward/operations/ak_nan_to_num.py
```diff
@@ -54,82 +56,47 @@ def nan_to_num(
...
+            return (
+                ak.contents.NumpyArray(
                     backend.nplike.nan_to_num(
                         layout.data,
-                        nan=nan,
-                        posinf=posinf,
-                        neginf=neginf,
+                        nan=nan_layout.data,
+                        posinf=None if posinf_layout is unset else posinf_layout.data,
+                        neginf=None if neginf_layout is unset else neginf_layout.data,
```
agoose77, line 85: This preserves some of the performance aspect to not requiring a broadcast (the default case, but not the all-scalars case). Functionally we could just remove this along with the above `unset if obj is None` logic, but this ensures we only create an array if an argument is given. The most general solution would explicitly check for scalars, and only broadcast those which aren't scalar (as originally written).

src/awkward/operations/ak_with_field.py
```diff
@@ -109,22 +110,21 @@ def action_is_record(input, **kwargs):
...
+                    first_content = base.contents[0]
+                    what, _ = broadcast_arrays(
+                        [what, first_content[: base.length]],
+                        depth_limit=1,
+                        left_broadcast=True,
+                        right_broadcast=True,
+                        broadcast_parameters_rule="one_to_one",
+                        highlevel=False,
+                        behavior=behavior,
                     )
```
agoose77, line 128: It feels like there should be a more elegant way to recurse _into_ the record array, but  I think this is actually the simplest solution.

src/awkward/_broadcasting.py
```diff
@@ -1081,4 +1078,4 @@ def broadcast_and_apply(
         },
     )
     assert isinstance(out, tuple)
-    return tuple(broadcast_unpack(x, isscalar, backend) for x in out)
+    return tuple(broadcast_unpack(x, return_scalar, backend) for x in out)
```
agoose77, line 1081: This is the main change; the caller determines whether the result should be a scalar.

PR #2394, opened 2023-04-12
fix: update `GrowableBuffer::move_to` algorithm

ianna:
Here is a memory allocation comparison by two algorithms: `move_to` and `concatenate`: the first releases each panel after it copies the data from it, the second copies all the data, then clears all the panels at the end.
Screenshot 2023-04-12 at 16 34 50


jpivarski (approved):
As we talked about today, this is a big improvement. The most important part is that the `panel_` destructors are no longer recursive on the stack, but use `std::move` to delete them in a `while` loop. It's also nice that the panels are deleted as soon as possible, rather than all at the end of the concatenation-copy.

PR #2396, opened 2023-04-13
fix: unify C++ and Python `GrowableBuffer::extend` algorithm

ianna:
As discussed: a `GrowableBuffer::extend` algorithm should add a single panel of a data size. The next panel added after that one should be a size of `initial*resize`.

Note, the first panel is always a size of `initial`. All next panels added after the first one should be a size of `initial*resize`.

### DIFF
--- a/src/awkward/_connect/numba/growablebuffer.py
+++ b/src/awkward/_connect/numba/growablebuffer.py
@@ -368,11 +368,19 @@ def GrowableBuffer_extend(growablebuffer, data):
         pos = growablebuffer._pos_get()
 
         available = len(growablebuffer._panels[-1]) - pos
-        while len(data) > available:
-            growablebuffer._add_panel()
+        remaining = len(data)
+
+        if remaining > available:
+            panel_length = int(
+                math.ceil(len(growablebuffer._panels[0]) * growablebuffer._resize)
+            )
+
+            growablebuffer._panels.append(
+                numpy.empty((max(remaining, panel_length),), dtype=growablebuffer.dtype)
+            )
+            growablebuffer._pos_set(0)
             available += len(growablebuffer._panels[-1])
 
-        remaining = len(data)
         while remaining > 0:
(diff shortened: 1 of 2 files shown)


src/awkward/_connect/numba/growablebuffer.py
```diff
@@ -87,10 +89,11 @@ def extend(self, data):
...
 
     def _add_panel(self):
-        panel_length = len(self._panels[-1])
-        if len(self._panels) == 1:
-            # only resize the first time, and by a large factor (C++ should do this, too!)
-            panel_length = int(numpy.ceil(panel_length * self._resize))
+        panel_length = (
+            self._initial
+            if len(self._panels) < 1
+            else int(numpy.ceil(self._initial * self._resize))
```
agoose77: Small query — can we use `math.ceil` here? Trying to avoid `numpy` unless required so we can use strict checks for it being used outside of `nplikes` :)

ianna: done!

ianna:
@jpivarski - this completes unification of the C++ and Python and Numba algorithms. Please, have a look. Thanks!

src/awkward/_connect/numba/growablebuffer.py
jpivarski: This change is fine, but the Numba connector is allowed to use NumPy indiscriminately.

jpivarski (approved):
Great!

This `GrowableBuffer` data type is still hidden. It can't operate without Numba, even in pure Python mode, because the list of panels is a `nb.typed.List`. There isn't a way to avoid that and share state between pure Python `GrowableBuffers` and Numba `GrowableBuffers`, so that's just how it has to be.

Therefore, you'll want to make this visible in a Numba-specific spot, such as `ak.numba.GrowableBuffer`. (It can't be allowed to import Numba until someone constructs one.)

PR #2397, opened 2023-04-13
fix: don't touch for `ascontiguousarray`

agoose77:
This is because touching should really only occur if the operation result _depends_ upon the array value, which it does not in this instance.

This PR 
1. allows `dask-awkward` to set `pretend_contiguous` on the `ndarray.dtype` for our "fake" buffers, ensuring that nplikes do not force them to become contiguous.
2. does not touch arrays in `ascontiguousarray`

The plan is to remove (1) later, once `dask-awkward` passes in its own special content types.

### DIFF
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -6,6 +6,7 @@ from abc import abstractmethod
 from typing import Any, Callable
 
 import awkward as ak
+from awkward._nplikes.cupy import Cupy
 from awkward._nplikes.jax import Jax
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
@@ -69,6 +70,7 @@ class NumpyKernel(BaseKernel):
         if issubclass(t, ctypes._Pointer):
             # Do we have a NumPy-owned array?
             if Numpy.is_own_array(x):
+                assert Numpy.is_c_contiguous(x), "kernel expects contiguous array"
                 if x.ndim > 0:
                     return ctypes.cast(x.ctypes.data, t)
                 else:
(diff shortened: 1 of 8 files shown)


jpivarski (approved):
We talked about this in our meeting, and the key point is

> The plan is to remove (1) later, once `dask-awkward` passes in its own special content types.

I'm not comfortable using an under-documented NumPy feature (`np.dtype.metadata`), particularly since `pretend_contiguous` is a statement about its `shape` and `strides` (everything but the `dtype`). However, the `dtype` follows the array, so it's expedient to do it this way for now.

This will go into the next release.

lgray:
What do we need to do over in dask awkward to make it function as expected?

I'm not sure if we've actually hit this one yet, given that the other PR fixed the issue I was running into with Yi-Mu.

jpivarski:
We'll be able to avoid relying on `np.dtype.metadata` by completing https://github.com/dask-contrib/dask-awkward/pull/203 with a Dummy version of `NumpyArray` ("Dummy" → "Placeholder"). By rehydrating arrays on Dask workers with Dummy subclasses of `Index` and `NumpyArray`, we'll be able to make operations like `ascontiguousarray` be no-ops for the rehydrated array, just as they are (with this PR) for the TypeTracerArray.

The current method of rehydrating with zero-strided (but otherwise real NumPy) arrays doesn't do that.

What I'm addressing here is a preferred refactoring; I don't know if you were asking about something more fundamental, like why this PR exists at all.

lgray:
Oh I was trying to understand what additional PRs would be needed in dask-awkward before minting another release, and it sounds like None until we run into this problem there.

PR #2403, opened 2023-04-14
feat: make `ak.numba.GrowableBuffer` visible 

ianna:
@HDembinski - FYI 
Hopefully, this GrowableBuffer in Numba provides a solution for an issue discussed in https://github.com/scikit-hep/uproot5/discussions/307


src/awkward/numba.py
```diff
@@ -1,5 +1,9 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
+import math
+
+import numpy  # noqa: TID251
```
ianna, line 5: @jpivarski and @agoose77 - has our policy changed? I get the following `ruff` error: 
```
TID251 `numpy` is banned: Use `numpy = ak._nplikes.Numpy.instance()` instead
```

ianna (commented):
@jpivarski - as discussed, the Python `GrowableBuffer` has been moved to the `numba` module and Numba is initialised when it is constructed. Please, have a look. Thanks

src/awkward/numba.py
jpivarski: The Numba submodules should be exempt from any "use nplike, not numpy" rule. These are

```
src/awkward/numba.py         # public API
src/awkward/_connect/numba   # implementation
```

In the Numba code, we have to make concrete assumptions about array buffers as being pointers to data in main memory—the arrays can't be abstract. The Numba-CUDA stuff switches NumPy for CuPy in the `Lookup` class, and Numba unifies these types under the `nb.types.Array` class; we use their unification, not ours.

src/awkward/numba.py
```diff
@@ -33,6 +37,119 @@ def register_and_check():
     _register()
 
 
+class GrowableBuffer:
+    def __init__(self, dtype, *, initial=1024, resize=8.0):
+        register_and_check()
+
+        import numba
```
jpivarski, line 44: I have verified that this `GrowableBuffer` class is exactly the one removed from src/awkward/_connect/numba/growablebuffer.py with this

```python
        register_and_check()
￼
￼        import numba
```

added. (This is a note for posterity, since git isn't able to tell us that this is a move, rather than a random addition and deletion.)

tests/test_2403_growablebuffer_in_numba.py
```diff
@@ -0,0 +1,398 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
```
jpivarski: It's not necessary to move tests from tests/test_2349_growablebuffer_in_numba.py to tests/test_2403_growablebuffer_in_numba.py, and anyway, they weren't moved: they were copied. In this git branch, all of the tests in this file are also in the other file—they're redundantly run twice. The only difference is that this one is accessing them through the public API (`ak.numba.GrowableBuffer`).

The original file, tests/test_2349_growablebuffer_in_numba.py, has two more tests than this one:

  * `test_python_append`
  * `test_python_extend`

so you could just delete the new file, tests/test_2403_growablebuffer_in_numba.py. If you want to test the fact that `GrowableBuffer` can be accessed from `ak.numba`, then just modify the old file. Although the tests are named after the issue or PR when they were first introduced, it's not wrong to modify them afterward. git tracks all of those changes; the reason for the issue/PR naming is to explain why a given test exists: it helps us to look up the original motivation.

jpivarski (approved):
Only one request: please remove tests/test_2403_growablebuffer_in_numba.py (see below).

Otherwise, this PR looks great, and when the above is done, it can be merged.

tests/test_2403_growablebuffer_in_numba.py
ianna: done

HDembinski:
Does this offer additional functionality compared to or perform better than creating a list in Numba and then converting it to a numpy array on return? To my knowledge, lists in Numba are implemented like vectors and conversion to numpy arrays is a zero copy operation inside numba.

agoose77:
I've not been following @ianna's work too closely, but `GrowableBuffer` has different memory characteristics to `std::vector`-like containers, which should improve its relative performance for certain use cases. In particular, for the case where you're just appending a lot of data and don't need to look at the buffer until the point that you need a contiguous buffer.

Both `std::vector` and `GrowableBuffer` use resize-on-grow strategies, but the latter only grows the size of the _next_ panel, rather than resizing the entire array into a single contiguous block (and copying contents).

When building an Awkward Array, that's a common use case. I'm not sure whether we e.g. plan to make it possible to look _into_ the array, but it should be possible to make that reasonably performant (keep a list of offsets, binary search into them). However, I don't know if that's on the cards. I assume we'll wait for a need @ianna? :)

jpivarski:
> Does this offer additional functionality compared to or perform better than creating a list in Numba and then converting it to a numpy array on return?

So... that's a great idea! This `GrowableBuffer` is a Numba-reimplementation of our C++ `GrowableBuffer`. We needed this in C++ at first because we needed sharable buffer data (`std::vector::data` is not safely sharable), and then later because we gave up on sharing and could then take advantage of the fact that the buffer didn't need to be contiguous (`std::vector::data` must be contiguous, and therefore reallocate-and-copy as it grows). The current algorithm makes discontiguous panels as it grows, in which the first panel is small and all subsequent panels are large, since it's common to have many small buffers and a few large ones.

233456786-587d1886-d56e-4aaa-a835-51bf3e1ea781

(where `initial=1024` items and `resize=8` by default). A `std::vector` grows exponentially (and therefore logarithmically often) in order to preserve contiguousness, and therefore it can overallocate by at most its resize factor (usually `1.5`, right?). By allocating equal-sized panels, we only overallocate by `initial × resize`, which doesn't scale with the total length of the `GrowableBuffer` as the `std::vector` does.

So keeping this logic in C++ is a good thing for controlling memory use, and the speed is close to optimal if our `resize` factor is big enough that the panel-handling overhead does not dominate over the data copy that we have to do anyway.

In Numba, however, things are different. The same argument about controlling memory use holds, but this Numba-`GrowableBuffer` actually uses `nb.typed.List` to manage its list of panels, which are `nb.types.ArrayType`. All of these things have to be reference-counted, unlike the C++ implementation, which is a linked list of `std::unique_ptrs`. Worse still, we have to put integer metadata (total length and length of the last panel) into a `nb.types.ArrayType` to ensure that all references to the `GrowableBuffer` share the same mutable state. So there are multiple ways in which this `GrowableBuffer` could be slower than just using Numba's `nb.typed.List` directly.

I found that to be the case: this `GrowableBuffer` is a few times slower than an `nb.typed.List` of numbers.



For your work, @HDembinski, I think you should continue to use `nb.typed.List`. @ianna pointed you to this because you had asked for it a few years ago. Although `nb.typed.List` existed (since 2019) when you asked in scikit-hep/uproot5/discussions/307 (2021), I didn't know about it, and I had had bad experiences with Numba's old reflected list.

For our work, I'm not sure whether we should use this `GrowableBuffer` _in Numba_ or not. (No question about using the C++ `GrowableBuffer` in C++.) For now, we won't change anything, but in the future, we might replace the Numba-`GrowableBuffer`'s implementation with a simple `nb.typed.List` internally—it would become a thin wrapper. That would still have value because we want to maintain some symmetry between the C++ `LayoutBuilder` and the Numba `LayoutBuilder`'s implementations; it would be easier to compare them if both are interacting with an object named `GrowableBuffer`.

It's also useful to have this `GrowableBuffer` because (as I said at the beginning), it's an illustration of how to write a mutable Numba extension type, a model for writing the `LayoutBuilder` classes.

HDembinski:
Great answer, thank you. Your implementation sounds like how a C++ std::deque is implemented, although the details could be different.

PR #2406, opened 2023-04-17
ci: attach header-only artifact to release

agoose77:
I've tested this in a [playground repo](https://github.com/agoose77/github-action-playground/actions/runs/4722005250), and it works. It's unfortunate that we have to rebuild the zip, but in this instance, we don't need the zip to have the same SHA or anything like that.


jpivarski (approved):
This will be very helpful, thanks!

PR #2407, opened 2023-04-17
fix: return empty list if broadcasting no arrays

agoose77:
`ak._broadcasting.broadcast_and_apply` expects at least one argument that has a backend, which is not satisfied if no arrays are passed. It also fails if no array-like objects are passed, but that's for another PR.


jpivarski (approved):
This still qualifies as a type-only test—a typetracer could give the same result. Good to catch this special case.

PR #2408, opened 2023-04-18
feat: LayoutBuilder in Numba

### LINKED ISSUE #1659: Implement LayoutBuilder in Numba

To be fully performant, LayoutBuilder in Numba needs to be a reimplementation, using Numba's lowering language. Starting with GrowableBuffer, which has been requested on its own [ref]. (In fact, the GrowableBuffer would be a good way to learn how to lower things in Numba.)

Now that the LayoutBuilder API has stabilized, it would be a good time to reimplement exactly that API in Numba.

ianna:
- [x] `BitMaskedType`: `append`, `extend`, `append_null` (0 args), `extend_null` (1 int arg)
- [x] `ByteMaskedType`: `append`, `extend`, `append_null` (0 args), `extend_null` (1 int arg)
- [x] Remove `EmptyRecordType`:  
- [x] `RecordType` must have at least one content
- [x] `TupleType` must have at least one content
<strike>- [ ] also remember to remove it from C++</strike> - as a separate PR
- [x] `EmptyType`
- [x] `IndexedOptionType`: `append`, `extend`, `append_null` (0 args), `extend_null`(1 int arg). (`append_index` and `extend_index`?)
- [x] `IndexedType`: `append`, `extend`. (`append_index` and `extend_index`?)
- [x] `ListOffsetType`: `begin_list`->content, `end_list` (<strike>both</strike> modify offsets)
- [x] `ListType`: `begin_list`->content (modify starts), `end_list` (modify stops) <strike>(both modify offsets)</strike>
- [x] `NumpyType`: `append`, `extend`
- [x] `RecordType`: `field` (Literal string) alias, `content` (Literal int and string) method
- [x] `RegularType`: `begin_list`->content, `end_list` (does nothing?), content property, size 0 is not allowed
- [x] `TupleType`: `index` (Literal int) alias, `content` (Literal int) method
- [x] `UnionType`: `append_index` (Literal int) updates tags and index -> content, and `content` (Literal int) method
- [x] `UnmaskedType`: `append`, `extend`


tests/test_2408_layoutbuilder_in_numba.py
```diff
@@ -0,0 +1,842 @@
...
+    f1(builder)
+
+    builder = lb.List(np.int8, lb.Numpy(np.int64))
+    f1(builder)
+
+    builder = lb.List(np.int32, lb.Empty())
+    f1(builder)
+
+    # FIXME:
+    builder = lb.ListOffset(np.int32, lb.ListOffset(np.int64, lb.Numpy(np.int64)))
```
ianna: @jpivarski - I need some help. I wonder if I need to use `numba.deferred_type` here. Currently, the first `ListOffset` type is correct, but the content of the second one is a `bound method NumpyType.type`:
```
src/awkward/_connect/numba/layoutbuilder.py:581: in __init__
    super().__init__(dmm, fe_type, members)
        __class__  = <class 'awkward._connect.numba.layoutbuilder.ListOffsetModel'>
        dmm        = <numba.core.datamodel.manager.DataModelManager object at 0x7fd318b6d040>
        fe_type    = ak.numba.lb.ListOffset(int32, ak.numba.lb.ListOffset(int64, ak.numba.lb.Numpy(int64)))
        members    = [('offsets', ak.GrowableBuffer(int32)),
 ('content',
  ak.numba.lb.ListOffset(int64, <bound method NumpyType.type of <class 'awkward._connect.numba.layoutbuilder.NumpyType'>>))]
        self       = <awkward._connect.numba.layoutbuilder.ListOffsetModel object at 0x7fd3197dd3a0>

```

ianna (commented):
@jpivarski - Could you, please, have a look? Thanks!

tests/test_2408_layoutbuilder_in_numba.py
jpivarski: I've never used `deferred_type` before (I had to look it up) and I doubt it's needed here. You'd have to defer a type for circular references, but all layout builders are strictly tree-like.

Can you say something more about what the issue is? I don't have any context.

Could it be that you want the return value of a `type` method/property and not the function-as-a-value? Is `NumpyType` the Numba type of a `lb.Numpy` instance?

jpivarski:
### Defining the semantics

What methods should all the builders have and what should they do? (Anything that has `append` has an `extend`.) Everything has a `__len__`. They have the indexes they need (`ListOffset` has `offsets`, etc.) as properties.

- `Numpy`: `append` passes a value to `GrowableBuffer.append`
- `Empty`: has no methods; can't be filled
- `ListOffset`:
  - has `content` property
  - `begin_list` returns `content`
  - `end_list` appends the length of `content` to `offsets`
- `Regular`:
  - has `content` property
  - `begin_list` returns `content`
  - `end_list` does nothing
- `IndexedOption`:
  - has `content` property
  - `append_valid` appends the length of `content` to `index` and returns `content`
  - `append_invalid` appends a `-1` to the `index` and returns nothing
- `ByteMasked`:
  - has `content` property
  - `append_valid` appends `valid_when` as a byte to `mask` and returns `content`
  - `append_invalid` appends `not valid_when` as a byte to `mask` and returns `content`
- `BitMasked`:
  - has `content` property and a length-2 NumPy array of `length` and `bit_position` (to avoid a modulo)
  - `append_valid` appends `valid_when` as a bit to `mask` and updates `length` and `bit_position` and returns `content`
  - `append_invalid` appends `not valid_when` as a bit to `mask` and updates `length` and `bit_position` and returns `content`
- `Unmasked`: has `content` property
- `Record`:
  - has nonempty `contents` property
  - `fields` property, `content(string literal)` method that returns an element of the `contents`
- `Tuple`:
  - has nonempty `contents` property
  - `content(integer literal)` method that returns an element of the `contents`
- `Union`:
  - has `contents` property of length >= 2
  - `append_content(integer literal)` method that appends the literal integer to `tags`, appends the length of that one `content` to `index` and returns that one `content`

We're dropping `List` and `Indexed`, since they don't produce arrays that are usefully different from what you can construct with the above. This is similar to the reasoning about dropping `EmptyRecord`, `EmptyTuple`.

There are types you can't make with this: empty records and tuples, and `RegularArray` with `size == 0`. Given an array made with the above, you can get all of those with slices.

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2602 @@
...
+
+    @classmethod
+    def _from_buffer(cls, data):
+        out = cls.__new__(cls)
+        out._data = data
+        out._parameters = None
+        return out
+
+    def __repr__(self):
+        return f"ak.numba.lb.Numpy({self._data.dtype}, parameters={self._parameters})"
```
agoose77: I'd prefer to drop the module prefix for these types so that we can alias them (e.g. export them under `ak.numba`) without the names mismatching, and also to keep in-line with convention.

tests/test_2408_layoutbuilder_in_numba.py
```diff
@@ -0,0 +1,1508 @@
...
+
+    f_compiled_index(builder)
+
+    @numba.njit(debug=True)
+    def f_runtime_index(x, indx):
+        i = x._contents[indx]
+        return i
+
+    # FIXME: Tuple needs a compile-time index
+    # f_runtime_index(builder, 0)
```
ianna: @jpivarski - I'm not sure how to deal with the fact that Numba does not support runtime indexing of a Tuple.

ianna (commented):
@jpivarski - I think, apart from the Tuple indexing issue, most what we have discussed is done. I will re-check with a fresh eye tomorrow. Please, let me know if you have ideas how to handle the Tuple better. Thanks!

tests/test_2408_layoutbuilder_in_numba.py
jpivarski: Since each field of a Tuple can have a different type, indexing can't be runtime.

```python
    @numba.njit
    def f_runtime_index(x, indx):
        i = x._contents[indx]
        return i
```

_should_ fail to compile because if `indx == 0`, the return type of the function would be one thing, and if `indx == 1`, the return type of the function would be another thing, etc.

Just as Record requires a literal string to select a field, Tuple requires a literal integer to select a field.

tests/test_2408_layoutbuilder_in_numba.py
```diff
@@ -0,0 +1,1462 @@
...
+    assert field_index(builder, "one") == 0
+    assert field_index(builder, "two") == 1
+    assert field_index(builder, "three") == 2
+
+    with pytest.raises(ValueError):
+        field_index(builder, "four")  # ValueError: tuple.index(x): x not in tuple
+
+    fill(builder)
+    array = builder.snapshot()
+    assert ak.to_list(array) == [{"one": 1.1, "three": "o", "two": 1}]  # ???
```
ianna: Hmm... I looked into the order issue: printing here the `array` from the `snapshot` looks correct and so does the `print(ak.to_list(array))`:
```python
[{'one': 1.1, 'two': 1, 'three': 'o'}]
```

The `assert ak.to_list(array)  == [{'one': 1.1, 'two': 1, 'three': 'o'}]` in Python prompt.

ianna (commented):
@jpivarski and @agoose77 - All is done except for the `@numba.extending.overload_method(RecordType, "content")` where it should take a `Record.content(Literal[str])`. 

Please, have a look when you have time and comment. Thanks!

tests/test_2408_layoutbuilder_in_numba.py
ianna: I forgot to mention that I tested it with Python 3.9 + Numba 0.56 and Python 3.10 + Numba 0.57. The result is the same: the assert check of `ak.to_list(array)` is sorted by field.

jpivarski:
By the way, I'm not ignoring this—it's a large PR that I'll need to allocate a large block of time to.

agoose77 (approved):
Huh, I didn't get the tag in my notifications for this (well, I must have done, but I didn't see it).

src/awkward/_connect/numba/layoutbuilder.py
jpivarski: This qualified class name is not needed here, in `LayoutBuilder.__repr__`, or following, in `LayoutBuilder.type`, but it is needed in the `NumpyType.name`, because Numba uses that string to determine type equality. Fully qualifying it there makes sense, but here we can have a shorter string, though it would have to be one that tells the world that it's a LayoutBuilder "`Numpy`" and not some other kind of "`Numpy`".

Perhaps

```
    def __repr__(self):
        return f"layoutbuilder.Numpy({self._data.dtype}, parameters={self._parameters})"
```

agoose77: The prefix for the `__repr__` that makes sense to me is `ak.numba`, with the intention that users would use the public namespace for importing them. If we didn't need that guarantee, I'd just use the `{__module__}.{__qualname__}`

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+        return self._builder
+
+    def clear(self):
+        return self._builder.clear()
+
+    def is_valid(self, error: str):
+        return self._builder.is_valid(error)
+
+    def _type(self, typestrs):
+        raise NotImplementedError
```
jpivarski: This looks like an abstract base class for all of the Python-accessible builders, a role similar to [ak.forms.Form](https://github.com/scikit-hep/awkward/blob/f2783ba480b5f9f668b687e327d0a5a453b82e5b/src/awkward/forms/form.py#L335). It doesn't look like an outer wrapper in the style of [ak.types.ArrayType](https://github.com/scikit-hep/awkward/blob/f2783ba480b5f9f668b687e327d0a5a453b82e5b/src/awkward/types/arraytype.py#L11) or [ArrayBuilder](https://github.com/scikit-hep/awkward/blob/main/awkward-cpp/include/awkward/builder/ArrayBuilder.h), and layout builders shouldn't have a wrapper, anyway (because they're targeting layouts, not high-level arrays).

However,

  * An abstract base class shouldn't be directly constructable, and this one has an `__init__` that looks, to a user, like something they should use. We didn't prohibit direct construction of `Form` by putting an exception in `Form.__init__` (although we could have, to make it more clear), but in the place where the base class has an init-helper, its name is [_init](https://github.com/scikit-hep/awkward/blob/f2783ba480b5f9f668b687e327d0a5a453b82e5b/src/awkward/forms/form.py#L345), to make it more clear that this isn't something users should use.
  * This `__init__` takes a `builder` argument and exposes a `builder` parameter, but not all subclasses of `LayoutBuilder` have a single child, which is anyway named `content`, rather than `builder`. The abstract base class shouldn't define a `builder`.
  * `clear` and `is_valid` seem to be methods used on all subclasses, so it's good to have them here with `raise NotImplementedError` (or better, `raise AssertionError("missing implementation")`) to make sure that none of them go unimplemented. This can be done better with Python-typing "protocols," though this PR should not get involved in adding Python type hints; that's too open-ended of a project.
  * All `LayoutBuilders` have lengths, so `__len__` would be another good method to provide a stub for here.
  * Same for `numbatype` and `snapshot`.
  * Also, all `LayoutBuilders` would have a `parameters` property. Since that's generic (there's no subclass-specific `parameters` handling), this class can handle that entirely, and the `parameters` property can be removed from subclasses.
  * I don't know why there's a `_type` method, or why it takes `typestrs`. On the other hand, `LayoutBuilders` will all need to produce Forms, so stubbing a `form` property at this level would be a good idea.

```suggestion
class LayoutBuilder:
    def _init(self, parameters):
        self._parameters = parameters

    @property
    def parameters(self):
        return self._parameters

    def __len__(self):
        raise AssertionError("missing implementation")

    def numbatype(self):
        raise AssertionError("missing implementation")

    def snapshot(self):
        raise AssertionError("missing implementation")

    @property
    def form(self):
        raise AssertionError("missing implementation")

    def clear(self):
        raise AssertionError("missing implementation")

    def is_valid(self, error: str):
        raise AssertionError("missing implementation")
```

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+    elif isinstance(builder, Unmasked):
+        return Unmasked.numbatype(builder)
+    elif isinstance(builder, Record):
+        return Record.numbatype(builder)
+    elif isinstance(builder, Tuple):
+        return Tuple.numbatype(builder)
+    elif isinstance(builder, Union):
+        return Union.numbatype(builder)
+    elif isinstance(builder, tuple):
+        return numba.types.Tuple([to_numbatype(it) for it in builder])
```
jpivarski: Why is this not following the pattern?

```suggestion
    elif isinstance(builder, tuple):
        return Tuple.numbatype(builder)
```

src/awkward/_connect/numba/layoutbuilder.py
jpivarski: The `parameters` won't be all that useful in a Numba-compiled function, since it returns a JSON-like dict of data. Its data type would differ from one layout builder's `parameters` to another's, but a single builder's `parameters` would at least be a compile-time constant.

It's pretty common for a single parameter to be a string, and perhaps this is the only case that should be implemented in the compiled context. To do that, we'll need a function that only returns one parameter (or `None`). At the Python level, that would be a method on `LayoutBuilder` like

```python
    def parameter(self, name):
        return self._parameters.get(name)
```

I'll show the Numba-typing for this below.

jpivarski: It also looks like you implemented a `length` property, rather than `__len__`, in the compiled context, so let's change `__len__` to

```python
    @property
    def length(self):
        raise AssertionError("missing implementation")
```

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+        return builder
+
+
+class LayoutBuilderType(numba.types.Type):
+    def __init__(self, parameters):
+        super().__init__(name="ak.numba.lb.LayoutBuilder()")
+
+    @property
+    def parameters(self):
+        return numba.types.StringLiteral(self._parameters)
```
jpivarski: You can't (easily) provide a type for all of the `properties` (and it's not a string; it's a dict). But you can provide a type for one property, if it happens to be a string, as described above in https://github.com/scikit-hep/awkward/pull/2408/files#r1243026912.

```suggestion
    @property
    def parameter(self, name):
        if name in self._parameters:
            return numba.types.StringLiteral(self._parameters[name])
        else:
            raise TypeError(f"LayoutBuilder.parameters does not have a {name!r}")
```

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+        return NumpyType(
+            numba.from_dtype(self.dtype), numba.types.StringLiteral(self._parameters)
+        )
+
+    @property
+    def _length(self):
+        return len(self._data)
+
+    def __len__(self):
+        return self._length
```
jpivarski:
```suggestion
    def __len__(self):
        return len(self._data)
```

and similarly for all other `LayoutBuilder` subclasses. The `Content` subclasses have extra properties for lengths, but that's because the lengths can be unknown. Here, the lengths are always going to be non-negative integers, and `__len__` is fine to use.

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+
+    @property
+    def dtype(self):
+        return self._data.dtype
+
+    @property
+    def parameters(self):
+        return self._parameters
+
```
jpivarski:
```suggestion
```

The `LayoutBuilder` abstract class manages all of the parameters; we don't have to (and shouldn't) do that in any subclasses.

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+    elif isinstance(builder, tuple):
+        return numba.types.Tuple([to_numbatype(it) for it in builder])
+    else:
+        return builder
+
+
+class LayoutBuilderType(numba.types.Type):
+    def __init__(self, parameters):
+        super().__init__(name="ak.numba.lb.LayoutBuilder()")
+
```
jpivarski: This is another abstract class; don't define an `__init__`.

```suggestion
    def _init(self, parameters):
        self._parameters = parameters
```

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+    def parameters(self):
+        return numba.types.StringLiteral(self._parameters)
+
+    @property
+    def data(self):
+        return ak.numba.GrowableBufferType(self._dtype)
+
+    @property
+    def length(self):
+        return numba.types.float64
```
jpivarski: These Numba type classes would be easier to define if you inherit your implementations in `LayoutBuilderType`.

```suggestion
class NumpyType(LayoutBuilderType):
    def __init__(self, dtype, parameters):
        super().__init__(
            name=f"ak.lb.Numpy({dtype!r}, parameters={parameters!r})"
        )
        self._dtype = dtype
        self._init(parameters)

    @property
    def dtype(self):
        return self._dtype

    @property
    def data(self):
        return ak.numba.GrowableBufferType(self._dtype)
```

(And similarly with all other Numba type classes for `LayoutBuilder` subclasses.)

jpivarski: Oh, I made the type disambiguator `"ak.lb.Numpy"` because this doesn't need to be the literal path in Awkward (it wasn't, anyway), it just needs to be distinct enough that no other Numba extensions are likely to use it. We've claimed the `"ak."` prefix, and there's no need to follow that up with "`numba.`" since this string is only used in Numba. `"lb."` will be enough to clarify that these are the LayoutBuilder part of Awkward's Numba extensions.

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+        *,
+        valid_when=True,
+        parameters=None,
+        initial=1024,
+        resize=8.0,
+    ):
+        self._mask = GrowableBuffer(dtype=dtype, initial=initial, resize=resize)
+        self._content = content
+        self._valid_when = valid_when
+        self._parameters = parameters
```
jpivarski: This is an interface that users will need to use, so we can leave out the `dtype` that they don't get control over.

```suggestion
@final
class ByteMasked(LayoutBuilder):
    def __init__(
        self,
        content,
        *,
        valid_when=True,
        parameters=None,
        initial=1024,
        resize=8.0,
    ):
        self._mask = GrowableBuffer(dtype=np.dtype(np.bool_), initial=initial, resize=resize)
        self._content = content
        self._valid_when = valid_when
        self._parameters = parameters
```

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+########## ListOffset #########################################################
+
+
+@final
+class ListOffset(LayoutBuilder):
+    def __init__(self, dtype, content, *, parameters=None, initial=1024, resize=8.0):
+        self._offsets = GrowableBuffer(dtype=dtype, initial=initial, resize=resize)
+        self._offsets.append(0)
+        self._content = content
+        self._parameters = parameters
```
jpivarski: All `dtypes` accepted from the user need to be regularized to a common form, such as NumPy's `dtype` object.

```suggestion
class ListOffset(LayoutBuilder):
    def __init__(self, dtype, content, *, parameters=None, initial=1024, resize=8.0):
        self._offsets = GrowableBuffer(dtype=np.dtype(dtype), initial=initial, resize=resize)
        self._offsets.append(0)
        self._content = content
        self._parameters = parameters
```

tests/test_2408_layoutbuilder_in_numba.py
```diff
@@ -0,0 +1,1462 @@
...
+
+    builder.extend(range(8))
+
+    assert ak.to_list(f12(builder)) == list(range(8))
+
+    builder.extend(range(8))
+
+    assert ak.to_list(f12(builder)) == list(range(8)) + list(range(8))
+
+
```
jpivarski: It doesn't need to be possible to do a `snapshot` inside of the compiled context. And if you did, the memory that the `ArrayView` points to (which is what an `ak.Array` is in a compiled context) couldn't be owned by anything, so there's no way to even do it without changing our ownership model.

```suggestion
```

tests/test_2408_layoutbuilder_in_numba.py
```diff
@@ -0,0 +1,1462 @@
...
+            builder.append(x)
+
+    @numba.njit
+    def append_single(builder, x):
+        builder.append(x)
+
+    @numba.njit
+    def snapshot(builder):
+        return builder.snapshot()
+
```
jpivarski: Perform the `snapshot` outside of the compiled context.

```suggestion
```

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+    def is_valid(self, error: str):
+        return True
+
+    def snapshot(self) -> ArrayLike:
+        """
+        Converts the currently accumulated data into an #ak.Array.
+        """
+        return ak.Array(
+            ak.contents.NumpyArray(self._data.snapshot(), parameters=self._parameters)
+        )
```
jpivarski: Since these are layout builders, the `snapshots` should build layouts (`ak.contents.Content` subclasses).

```suggestion
    def snapshot(self) -> ak.contents.Content:
        return ak.contents.NumpyArray(self._data.snapshot(), parameters=self._parameters)
```

In Python, you have the `snapshot` methods returning `ak.Array` (and since you know the exact type, you didn't need to annotate it as `ArrayLike`), and then the recursion had to unpack each level down with `.layout` in order to pack it back up with a new `ak.Array`. Returning the layout directly is a lot more straightforward.

If you want to have a convenience method returning an `ak.Array`, you can have two functions: one recurses, and the other packages up the output of the recursive function in the high-level wrapper. For instance, you could have this:

```python
class LayoutBuilder:  # the abstract class
    def snapshot(self) -> ak.Array:
        return ak.Array(self.snapshot_layout())

class Numpy(LayoutBuilder):  # each subclass
    def snapshot_layout(self) -> ak.contents.Content:
        # recursively calls `snapshot_layout`
```

(I don't have opinions on whether the recursive function is named `snapshot` or `snapshot_layout` and the wrapping function is named `snapshot_array` or `snapshot`, as long as they're a consistent pair.)

src/awkward/_connect/numba/layoutbuilder.py
```diff
@@ -0,0 +1,2547 @@
...
+
+@numba.extending.overload_method(NumpyType, "snapshot")
+def Numpy_snapshot(builder):
+    def snapshot(builder):
+        # FIXME: returns 'numpy.ndarray'
+        return builder.data.snapshot()
+
+    return snapshot
+
+
```
jpivarski: The `FIXME` says it: it's an incorrect implementation because the Python version returns some sort of an Awkward Array (`ak.Array` or `ak.contents.Content` subclass) and the Numba implementation returns a NumPy array. But you don't have to (and can't) implement these in the compiled context.

```suggestion
```

They also weren't consistently implemented, and only this one was tested, which is probably why it had to temporarily return a NumPy array (which is possible, instead of an Awkward Array, which is impossible).

tests/test_2408_layoutbuilder_in_numba.py
jpivarski: It's possible to drill down into this by finding out if the `builder.contents` and `builder.fields` are also out of order. Or maybe after the snapshot, the `array.fields` and `array.contents` of the `RecordArray`. At some point, they get out of order.

The order `"one", "three", "two"` is suggestive of alphabetical sorting, but with only three items, that might be a coincidence. This might be some random hash order that looks alphabetical for these three.

src/awkward/_connect/numba/layoutbuilder.py
jpivarski: Although there's a rule that `reprs` that look like they could be copy-pasted as code should be copy-pasteable as code, that rule has to assume some imports. Usually, the assumption is that you've imported the class name as an unqualified word, so that

```python
    def __repr__(self):
        return f"{type(self).__name__}({', '.join(self._constructor_args)})"
```

is a good `repr`.

But the `{__module__}` in this case is going to be very long: `awkward.numba.layoutbuilder`. It won't even abbreviate `awkward` to `ak`.

Using "`layoutbuilder.`" as a prefix was imagining the hypothetical world in which the user has already imported

```python
from awkward.numba import layoutbuilder
```

just as NumPy's array dtypes assume that the user has already imported

```python
from numpy import array, dtype
```

jpivarski:
Most of the comments I've made above apply to all `LayoutBuilder` subclasses, but I only wrote them for the first instances to avoid cluttering this PR thread.

I also suggested that all of the `LayoutBuilder` subclasses should have a `form` property, so that their Form (and from that, the Type) can be determined without doing a snapshot. This property does not need to be implemented in the compiled context (and can't be, since Forms haven't been lowered to Numba).

Those `form` implementations would be a straightforward recursion, and can be tested against the `snapshot`'s Forms in the tests.

tests/test_2408_layoutbuilder_in_numba.py
ianna: done

tests/test_2408_layoutbuilder_in_numba.py
ianna: done

jpivarski:
As you said in the group meeting today, this PR is now done(ish?) and reviewable, so I'll take it out of draft status and remind myself to review it again.

jpivarski (approved):
I reviewed this by checking it out and trying several examples, and they all looked fine. For example,

```python
import numba as nb
import numpy as np

import awkward as ak
import awkward.numba.layoutbuilder as lb

@nb.njit
def doit(builder):
    x = 0
    y = 0.0
    for i in range(10):
        subbuilder = builder.begin_list()
        x_builder = subbuilder.content("x")
        y_builder = subbuilder.content("y")
        for _ in range(i):
            x_builder.append(x)
            y_builder.append(y)
            x += 1
            y += 1.1
        builder.end_list()

builder = lb.ListOffset(
    np.int64,
    lb.Record([lb.Numpy(np.int32), lb.Numpy(np.float64)], ["x", "y"]),
)

doit(builder)

ak.Array(builder.snapshot()).show(type=True)
```

produces

```
type: 10 * var * {
    x: int32,
    y: float64
}
[[],
 [{x: 0, y: 0}],
 [{x: 1, y: 1.1}, {x: 2, y: 2.2}],
 [{x: 3, y: 3.3}, {x: 4, y: 4.4}, {x: 5, y: 5.5}],
 [{x: 6, y: 6.6}, {x: 7, y: 7.7}, {x: 8, y: 8.8}, {x: 9, y: 9.9}],
 [{x: 10, y: 11}, {x: 11, y: 12.1}, {...}, {x: 13, ...}, {x: 14, y: 15.4}],
 [{x: 15, y: 16.5}, {x: 16, y: 17.6}, {...}, ..., {x: 19, ...}, {x: 20, y: 22}],
 [{x: 21, y: 23.1}, {x: 22, y: 24.2}, ..., {x: 26, ...}, {x: 27, y: 29.7}],
 [{x: 28, y: 30.8}, {x: 29, y: 31.9}, ..., {x: 34, ...}, {x: 35, y: 38.5}],
 [{x: 36, y: 39.6}, {x: 37, y: 40.7}, ..., {x: 43, ...}, {x: 44, y: 48.4}]]
```

I had to move the Python interface classes (the `builder` that has to be created outside of the compiled context) to a public place to be able to get them with

```python
import awkward.numba.layoutbuilder as lb
```

(This is how I think we should present it to the world. We'll encourage the use of the `lb` abbreviation, here and in C++: `lb::ListOffset`, etc.)

When the tests finish, I'll merge it. Thanks, @ianna!

Issue #2364, opened 2023-04-05
ak.to_numpy turns an empty array of type string into an empty array of type float64


jpivarski:
### Version of Awkward Array

HEAD

### Description and code to reproduce

```python
>>> array = ak.Array(["one", "two", "three"])

>>> ak.to_numpy(array)
array(['one', 'two', 'three'], dtype='<U5')

>>> array[0:0].type.show()
0 * string

>>> ak.to_numpy(array[0:0])
array([], dtype=float64)
```

Same for bytestrings.

PR #2410, opened 2023-04-18
fix: partially fix string broadcasting

### LINKED ISSUE #1683: String broadcasting is not producing sensible results
### Version of Awkward Array

main

As Jim and I discussed today, string broadcasting is currently not producing reliable results for string-string, or even string-non string cases. In particular many broadcasts violate the appearance that strings are atoms:

```python3
>>> import awkward as ak
>>> x = ak._v2.Array([["one", "two"], ["three", "four"]])
>>> ak._v2.broadcast_arrays(x[1:], x[:-1])
ValueError: while calling (from <ipython-input-42-6e4db831c1ff>, line 1)
    ak._v2.broadcast_arrays(
        arrays = (<Array [['one', 'two']] type='1 * var * string'>, <Array [[...
        kwargs = {}
    )
Error details: cannot broadcast nested list (in compiled code: https://github.com/scikit-hep/awkward-1.0/blob/1.10.0rc1/src/cpu-kernels/awkward_ListArray_broadcast_tooffsets.cpp#L27)
```

In this case, broadcasting an Array of strings against an Array of integers produces arrays that have the same structure as though the outer `__array__ = "string"` were missing (i.e. broadcasting happens against the underlying characters array):
```python3
>>> import awkward as ak
>>> x = ak._v2.Array([["one", "two"], ["three", "four"]])
>>> y = ak._v2.Array([[1,2],[3, 4]])
>>> ak._v2.broadcast_arrays(x, y)
[<Array [['one', 'two'], ['three', 'four']] type='2 * var * string'>,
 <Array [[[1, 1, 1], [2, 2, 2]], [[...], ...]] type='2 * var * var * int64'>]
```

agoose77:
Partially fixes #1683 for the trivial string-string (same depth) case. The other cases will take longer, so for now we just add xfail tests.

### DIFF
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -222,7 +222,9 @@ def _impl(
         inputs.append(y.to_backend(backend))
 
     def action(inputs, depth, **kwargs):
-        if depth == depth_limit or all(x.purelist_depth == 1 for x in inputs):
+        if any(x.is_option for x in inputs):
+            return None
+        elif depth == depth_limit or all(x.purelist_depth == 1 for x in inputs):
             return tuple(inputs)
         else:
             return None
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -222,9 +222,14 @@ def _impl(
         inputs.append(y.to_backend(backend))
 
     def action(inputs, depth, **kwargs):
-        if any(x.is_option for x in inputs):
-            return None
-        elif depth == depth_limit or all(x.purelist_depth == 1 for x in inputs):
+        # The depth limit is the depth at which we must return, i.e.
+        # the _first_ layout at that depth
+        if depth == depth_limit:
+            return tuple(inputs)
+        # Walk through non-leaf nodes
+        elif all(
+            x.purelist_depth == 1 and not (x.is_option or x.is_indexed) for x in inputs
+        ):
             return tuple(inputs)
         else:
             return None
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -222,7 +222,7 @@ def _impl(
         inputs.append(y.to_backend(backend))
 
     def action(inputs, depth, **kwargs):
-        if depth == depth_limit or all(x.is_numpy for x in inputs):
+        if depth == depth_limit or all(x.purelist_depth == 1 for x in inputs):
             return tuple(inputs)
         else:
             return None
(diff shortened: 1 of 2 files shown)


src/awkward/operations/ak_broadcast_arrays.py
```diff
@@ -222,7 +222,7 @@ def _impl(
         inputs.append(y.to_backend(backend))
 
     def action(inputs, depth, **kwargs):
-        if depth == depth_limit or all(x.is_numpy for x in inputs):
+        if depth == depth_limit or all(x.purelist_depth == 1 for x in inputs):
```
agoose77: This is non-linear in time (for the tree depth). I'm assuming it's small enough not to matter.

jpivarski: Yes. It's also why we don't want functions like `purelist_depth` to _create_ new Forms, just to evaluate the one property. That's why we have `Content.purelist_depth` be a duck-typeable reference to the same function as `Form.purelist_depth`. We want to be able to do these non-linear metadata operations quickly enough. (Traversing the tree more than once is okay, building it more than once isn't.)

jpivarski (approved):

src/awkward/operations/ak_broadcast_arrays.py
```diff
@@ -222,7 +222,14 @@ def _impl(
...
 
     def action(inputs, depth, **kwargs):
-        if depth == depth_limit or all(x.is_numpy for x in inputs):
+        # The depth limit is the depth at which we must return, i.e.
+        # the _first_ layout at that depth
+        if depth == depth_limit:
+            return tuple(inputs)
+        # Walk through non-leaf nodes
+        elif all(
+            x.purelist_depth == 1 and not (x.is_option or x.is_indexed) for x in inputs
```
agoose77, line 231: Trying fairly hard not to special case strings here!

agoose77:
@jpivarski this PR changes how empty arrays behave; they are returned without conversion.

e.g. (old)
```pycon
>>> ak.broadcast_arrays([], [1,2,])
[<Array [] type='0 * float64'>, <Array [] type='0 * int64'>]


```
vs (new)
```pycon
>>> ak.broadcast_arrays([], [1,2,])
[<Array [] type='0 * unknown'>, <Array [] type='0 * int64'>]
```

We touched on an aspect of this https://github.com/scikit-hep/awkward/issues/2364, do you agree with this change?

I think in-general we should remove this conversion logic from `broadcast_and_apply`, as we want to avoid converting to an arbitrary dtype.

jpivarski:
I don't see why broadcasting should change a dtype, so the new behavior is better. It's also small enough and corner-case enough that we don't need to do a deprecation cycle. (Also, it's unclear how we could, since it's a difference in return value, not arguments.)

agoose77:
Ok! I'll make another PR to remove empty conversion in general, which could also (if we want) remove conversion for reducers too.

PR #2411, opened 2023-04-19
feat!: raise error for out-of-bounds axis in `ak.cartesian`

agoose77:
There are probably other functions that have this problem, but I'll start with `ak.cartesian`!

Currently, `ak.cartesian` accepts any axis that `>0`, as long as the negative depth is not ambiguous, or disagrees with the value for other layouts. This PR ensures that `axis` lies within the range `0 <= axis < max_depth`, where `max_depth` is the maximum `purelist_depth` of all layouts.

### DIFF
--- a/src/awkward/operations/ak_cartesian.py
+++ b/src/awkward/operations/ak_cartesian.py
@@ -217,6 +217,8 @@ def _impl(arrays, axis, nested, parameters, with_name, highlevel, behavior):
             ).to_backend(backend)
             for name, layout in arrays.items()
         }
+        layouts = list(array_layouts.values())
+        fields = list(array_layouts.keys())
 
     else:
         arrays = list(arrays)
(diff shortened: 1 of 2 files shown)


src/awkward/operations/ak_cartesian.py
```diff
@@ -236,30 +240,30 @@ def _impl(arrays, axis, nested, parameters, with_name, highlevel, behavior):
...
     if posaxis is None or posaxis < 0:
-        raise ValueError("negative axis depth is ambiguous")
+        raise np.AxisError("negative axis depth is ambiguous")
+    # Ensure other layouts have same positive value for axis
     for layout in layouts[1:]:
         if maybe_posaxis(layout, axis, 1) != posaxis:
-            raise ValueError(
+            raise np.AxisError(
                 "arrays to cartesian-product do not have the same depth for negative axis"
             )
```
jpivarski, line 252: Things like this—ValueError instead of AxisError—are a relic of the pybind11 exception limitations. We had to produce C++ exceptions that would get remapped to Python exceptions (because we also wanted to keep the Python-aware code separate from the Python-unaware code, a separation that we still maintain), and so couldn't raise AxisError in the places where it should have been raised. Thanks for catching these, most of the other cases have been converted over already.

jpivarski (approved):

PR #2412, opened 2023-04-19
refactor: add custom axis error

agoose77:
> **Warning**
> This PR needs to be finalised to account for #2411 

Ergonomically, I'd prefer to import an exception from our errors module, instead of raising `np.AxisError`. This doesn't have to be a new class; we could just import the name from `numpy`. However, we should probably prefer to define our own exception types instead of directly re-using NumPy's. For most people, this won't change anything, but it means that we can detect axis errors in Awkward's code vs those from NumPy. This might not be a useful feature, though.

I haven't yet figured out how we should export these exceptions for public use.

### DIFF
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -125,8 +125,6 @@ class ErrorContext:
                 valuestr = f"repr-raised-{type(err).__name__}"
 
         elif isinstance(value, np.ndarray):
-            import numpy  # noqa: TID251
-
             if not numpy.__version__.startswith("1.13."):  # 'threshold' argument
                 prefix = f"{type(value).__module__}.{type(value).__name__}("
                 suffix = ")"
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -6,6 +6,8 @@ import threading
 import warnings
 from collections.abc import Mapping, Sequence
 
+import numpy  # noqa: TID251
+
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._typing import TypeVar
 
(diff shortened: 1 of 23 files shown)


src/awkward/_errors.py
```diff
@@ -364,3 +364,7 @@ def deprecate(
 
 class FieldNotFoundError(IndexError):
     ...
+
+
+class AxisError(numpy.AxisError):
+    ...
```
jpivarski: How about a reference, instead of an empty subclass?

jpivarski (approved):
Getting `AxisError` from NumPy in a single place (_errors.py), instead of all over the place, gives us more control, such that it can be changed at any time.

For this argument, which you gave in the first comment of this PR, you don't need `ak._errors.AxisError` to be a subclass of `np.AxisError`; it can be a direct reference to get all of the benefits.

How about doing this in minimal steps—just make it a direct reference now and promote it to being a subclass if we ever need that? Not for any strong reason, just the YAGNI principle. Since you've now localized all references to a single place, that replacement would be easy to do.

Users _can_ see the difference if they check `type(err) is np.AxisError` instead of `isinstance(err, np.AxisError)`. Since a general Python principle strongly recommends the latter, we'd be justified in making the breaking change from referencing `np.AxisError` to subclassing it, but how about if we only do that when we have an actual reason for doing so? (A reason we can point to if somebody complains.) I don't think we have such a reason _yet_.

src/awkward/_errors.py
agoose77: There's no good reason not to do this, I think it's a good suggestion. It was really an aesthetic thought, which is not a good argument in favour :)

agoose77:
```suggestion
AxisError = numpy.AxisError
```

PR #2413, opened 2023-04-19
feat!: don't broadcast `EmptyArray` in `broadcast_and_apply`

agoose77:
Currently, we're too eager to convert `EmptyArray`s into known `NumpyArray`s. This PR starts the process of unwinding this, making `ak.broadcast_arrays` preserve the unknown array.

The planned existence of `ak.enforce_type` should ensure that we can deprecate this logic (in this PR), so that e.g. ufunc operations on unknown arrays throw warnings.

At a high level, we should permit `EmptyArray`s instead of `NumpyArray`s where it is not required that the programmer choose an arbitrary dtype. i.e., if the code requires a boolean dtype, it can accept an `EmptyArray`, whereas if the code supports many dtypes (reducers) it fails.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -24,7 +24,6 @@ from awkward._util import unset
 from awkward.contents.bitmaskedarray import BitMaskedArray
 from awkward.contents.bytemaskedarray import ByteMaskedArray
 from awkward.contents.content import Content
-from awkward.contents.emptyarray import EmptyArray
 from awkward.contents.indexedarray import IndexedArray
 from awkward.contents.indexedoptionarray import IndexedOptionArray
 from awkward.contents.listarray import ListArray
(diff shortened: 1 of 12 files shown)


jpivarski:
We now have an `ak.enforce_type`. Can this PR be closed?

agoose77:
I think the contrary; we want to be stricter now that it's easier to remove unknown arrays through `enforce_type`.

jpivarski:
So `EmptyArray` would not be allowed to broadcast with `NumpyArray` at all? Even though someone could reassign `EmptyArray`, that sounds inconvenient. (Broadcasting is only ever about convenience.)

agoose77:
Rather that it won't be coerced during broadcasting. This PR name should really imply that broadcasting won't handle unknown arrays, just as it doesn't handle NumPy arrays.

jpivarski:
So the broadcasting logic can be simplified by just removing the `EmptyArray` case? That sounds good! But it would probably be easier to start over, rather than bring this PR up-to-date, right?

PR #2421, opened 2023-04-21
fix: consistently set a `resize` default value to 8 - that is passed to GrowableBuffer

### LINKED ISSUE #2419: Default value of resize needs to be set consistently for the new GrowableBuffer algorithm
### Version of Awkward Array

HEAD

Since GrowableBuffer now interprets `resize` as the difference in size between panel 1 and all subsequent panels:

equation

instead of

equation

https://github.com/scikit-hep/awkward/blob/3fc5f65554b1e35d4a4eab7a81b418b80e99fc58/header-only/growable-buffer/awkward/GrowableBuffer.h#L449-L455

The default value for `resize` should now be `16` (or what optimal value did you find, @ianna?), rather than `1.5`. Unfortunately, it is assigned all over the place, because it has to be a default for a lot of "user starts here" functions.

```bash
% fgrep -r 1.5 awkward-cpp/src src
awkward-cpp/src/python/content.cpp:      }), py::arg("initial") = 1024, py::arg("resize") = 1.5)
awkward-cpp/src/python/forth.cpp:               py::arg("output_resize_factor") = 1.5)
src/awkward/highlevel.py:    def __init__(self, *, behavior=None, initial=1024, resize=1.5):
src/awkward/operations/ak_mean.py:        <Array [1.5, nan, 4.5] type='3 * float64'>
src/awkward/operations/ak_mean.py:    because there are three lists, the first has mean `1.5`, the second is
src/awkward/operations/ak_from_json.py:    resize=1.5,
src/awkward/operations/ak_from_iter.py:    resize=1.5,
```

https://github.com/scikit-hep/awkward/blob/3fc5f65554b1e35d4a4eab7a81b418b80e99fc58/awkward-cpp/src/python/content.cpp#L198-L200

(The awkward-cpp/src/python/forth.cpp one shouldn't change yet because `ForthOutputBuffer` is not based on `GrowableBuffer`.)

https://github.com/scikit-hep/awkward/blob/3fc5f65554b1e35d4a4eab7a81b418b80e99fc58/src/awkward/highlevel.py#L2238

https://github.com/scikit-hep/awkward/blob/3fc5f65554b1e35d4a4eab7a81b418b80e99fc58/src/awkward/operations/ak_from_json.py#L31

https://github.com/scikit-hep/awkward/blob/3fc5f65554b1e35d4a4eab7a81b418b80e99fc58/src/awkward/operations/ak_from_iter.py#L21

ianna:
The new GrowableBuffer algorithm resizes its panels once. The plots show that 8 is a good compromise.
Screenshot 2023-04-21 at 16 50 13
Screenshot 2023-04-21 at 16 44 43

fixes issue #2419

### DIFF
--- a/awkward-cpp/src/python/content.cpp
+++ b/awkward-cpp/src/python/content.cpp
@@ -197,7 +197,7 @@ make_ArrayBuilder(const py::handle& m, const std::string& name) {
   return (py::class_<ak::ArrayBuilder>(m, name.c_str())
       .def(py::init([](const int64_t initial, double resize) -> ak::ArrayBuilder {
         return ak::ArrayBuilder({initial, resize});
-      }), py::arg("initial") = 1024, py::arg("resize") = 1.5)
+      }), py::arg("initial") = 1024, py::arg("resize") = 8)
       .def_property_readonly("_ptr",
                              [](const ak::ArrayBuilder* self) -> size_t {
         return reinterpret_cast<size_t>(self);
--- a/docs/reference/ak.builder.ArrayBuilder.rst
+++ b/docs/reference/ak.builder.ArrayBuilder.rst
@@ -10,11 +10,11 @@ object is wrapped by :class:`ak.ArrayBuilder`.
 underscores after "begin" and "end," but that hasn't happened in the
 low-level interface, yet or possibly at all.)
 
-.. py:class:: ArrayBuilder(initial=1024, resize=1.5)
+.. py:class:: ArrayBuilder(initial=1024, resize=8)
 
 .. py:method:: ArrayBuilder.__getitem__(where)
 
-.. py:method:: ArrayBuilder.__init__(initial=1024, resize=1.5)
+.. py:method:: ArrayBuilder.__init__(initial=1024, resize=8)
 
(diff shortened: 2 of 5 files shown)


jpivarski (approved):
This was a study of the C++ `GrowableBuffer`, right?

The unfilled reserved memory reaches a limit at `resize=16`. Did the study always fill `GrowableBuffers` up to a particular `length`? Because if you allow any `length`, then it can have as much as `initial × resize - 1` unfilled reserved for any `resize` factor.

The speed seems to be independent of `resize` factor, which suggests that all of these panels, including the `initial` sized one, are big enough for the panel-overhead to be subdominant. I'm not taking that increase in time between `resize=8` and `resize=16` as significant (it's much smaller than a factor of 2 and it goes in the wrong direction). I'm happy enough to go with a factor of 8 because it's like an order of magnitude.

Thanks for doing the study! We can be confident that we're not introducing a bad parameterization.

I think this is ready to be merged.

jpivarski:
We don't need codecov to pass, and it's stuck, so I'll just merge it manually.

PR #2424, opened 2023-04-25
fix: handle reordered contents in `ak.almost_equal`

agoose77:
`ak.almost_equal` made some poor assumptions about content orderings, which this PR fixes. Now, unions are considered equal against unions with different content orders (as long as the tags align equally). Records are compared by field for non-tuples, addressing the same problem with records.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -5,7 +5,13 @@ import math
 
 import numpy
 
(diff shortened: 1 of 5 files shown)


jpivarski (approved):
I see that this was pulled out of https://github.com/scikit-hep/awkward/pull/2365#discussion_r1175775791

Approved!

We can skip the codecov that didn't run. Maybe we should make it a non-required test, considering how often it fails to start.

jpivarski:
I removed codecov as a required test. (It was a weak requirement anyway; the test coverage would have to drop way down for it to fail.)

PR #2426, opened 2023-04-25
fix: `is_equal_to` bug in `UnionArray`

agoose77:
The tags were being compared by identity, not equality.

### DIFF
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1583,7 +1583,7 @@ class UnionArray(Content):
 
     def _is_equal_to(self, other, index_dtype, numpyarray):
         return (
-            self.tags == other.tags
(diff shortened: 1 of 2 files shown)


src/awkward/contents/unionarray.py
```diff
@@ -1583,7 +1583,7 @@ def _to_backend(self, backend: Backend) -> Self:
 
     def _is_equal_to(self, other, index_dtype, numpyarray):
         return (
-            self.tags == other.tags
+            self.tags.is_equal_to(other.tags)
```
jpivarski, line 1586: Yikes! That's right, this `==` is either not defined on Index or it would be non-coercible to a scalar Boolean.

jpivarski (approved):
This is a simple, clear-cut fix; I'll merge it now.

PR #2427, opened 2023-04-25
feat: canonicalise `union[?X, Y]` into `union[?X, ?Y]`

### LINKED ISSUE #2371: `union[X, ?Y]` should be canonicalized to `union[?X, ?Y]`
Although records and unions have a canonical `contents` order that they preserve, it would be correct to say that two records or two unions are equal if they have the same set of `contents`, regardless of order (`zip(fields, contents)` in the case of records, which puts an ordering back in for tuples).

I was trying some examples with our new canonicalization rules to see if we'd have to distinguish between types like

```
A = union[option[X], Y]
```

and

```
B = union[X, option[Y]]
```

I found that an attempt to create

```
C = option[union[X, Y]]
```

forces you to go through `simplified` and makes

```
D = union[option[X], option[Y]]
```

but it's still possible to create an asymmetric case (`A` or `B`). How about if we squash that avenue as well? That is, an attempt to construct

```python
UnionArray(tags, index, [option_type_array, non_option_type_array])
```

would (with a future rule) fail and force you to use `UnionArray.simplified` (already required wherever the codebase makes `UnionArray`, and then `UnionArray.simplified` would wrap all the non-option type arrays in an `UnmaskedArray` if any one of the provided `contents` is option-type.

How about it? Then you'd have fewer cases to consider when checking equality of types: you wouldn't have to worry about whether `A` and `B` should be considered equivalent because they can't exist (they always be `D`).

This comment can become a new issue, if you like that idea.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/pull/2368#discussion_r1160806288_

agoose77:
Closes #2371

### DIFF
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -117,6 +117,8 @@ class UnionArray(Content):
 
         if len(contents) < 2:
             raise TypeError(f"{type(self).__name__} must have at least 2 'contents'")
+
+        n_seen_options = 0
         for content in contents:
             if not isinstance(content, Content):
                 raise TypeError(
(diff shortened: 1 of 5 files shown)


jpivarski (approved):
I scanned over it, and this is more straightforward than I thought it would be. Nice!

agoose77:
Yeah, originally I thought we'd need to be careful about where to introduce the options. However, I think this is fine; options always merge parameters with non options at the `.content` level, so merging early won't cause problems there, I think.

PR #2437, opened 2023-04-28
feat!: deprecate `length_zero_array` highlevel argument

### LINKED ISSUE #2435: Deprecate `highlevel=True` for `Form.length_zero_array`

It seems to me that `Form.length_zero_array` is an L2 function, and therefore doesn't need to be concerned with behaviours or `ak.Array`s. I propose that we deprecate the `highlevel=True` variant of this function, to keep the separation of concerns.

agoose77:
Closes #2435

### DIFF
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -9,6 +9,7 @@ from collections.abc import Mapping
 import awkward as ak
 from awkward._backends.numpy import NumpyBackend
 from awkward._behavior import find_typestrs
+from awkward._errors import deprecate
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._nplikes.shape import unknown_length
 from awkward._parameters import parameters_union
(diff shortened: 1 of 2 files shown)


src/awkward/forms/form.py
```diff
@@ -427,6 +428,12 @@ def _type(self, typestrs):
     def length_zero_array(
         self, *, backend=numpy_backend, highlevel=True, behavior=None
     ):
+        if highlevel:
```
jpivarski, line 431: It's correct to trigger the warning on `highlevel` (ignoring `behavior`) because `behavior` has no effect when `highlevel is False`.

src/awkward/forms/form.py
```diff
@@ -427,6 +428,12 @@ def _type(self, typestrs):
...
+            deprecate(
+                "The `highlevel=True` variant of `Form.length_zero_array` is now deprecated. "
+                "Please use `ak.Array(form.length_zero_array(...), behavior=...)` if an `ak.Array` is required.",
+                version="2.3.0",
```
jpivarski, line 435: 2.3.0 is the correct starting date, since we're in 2.1.x.

jpivarski (approved):
This is good! I've also put it in the wiki, where these deprecations are consolidated.

https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20

agoose77:
Thanks! That was my next step, you've saved me some time!

Issue #2440, opened 2023-04-28
Performance regressions compared to ak1


lgray:
### Version of Awkward Array

2.1.4

### Description and code to reproduce

Long story short in the `trijet["p4"] = trijet.j1 + trijet.j2 + trijet.j3` I am seeing up to a 2.5x longer execution time in ak2 in the following code. 

I see also about 2x more peak RSS usage in top, but using python's internal max RSS monitor from resource they show nearly the same in terms of max RSS (but that may not count something?). I'll see if I can get a better measure.

Here's the code (you will need to `pip install pyinstrument`):
```python3
import numpy as np
import awkward as ak

from pyinstrument import Profiler

np.random.seed(1)

behavior = {}

@ak.mixin_class(behavior)
class LorentzVector:
    @property
    def mass2(self):
        """Squared `mass`"""
        return self.t**2 - self.x**2 - self.y**2 - self.z**2

    @property
    def mass(self):
        r"""Invariant mass (+, -, -, -)
        :math:`\sqrt{t^2-x^2-y^2-z^2}`
        """
        return np.sqrt(self.mass2)
    
    @ak.mixin_class_method(np.add, {"LorentzVector"})
    def add(self, other):
        """Add two vectors together elementwise using `x`, `y`, `z`, and `t` components"""
        return ak.zip(
            {
                "x": self.x + other.x,
                "y": self.y + other.y,
                "z": self.z + other.z,
                "t": self.t + other.t,
            },
            with_name="LorentzVector",
            behavior=self.behavior,
        )

rows = 2_000_000

counts = np.rint(np.random.lognormal(mean=1.0, sigma=0.8, size=rows)).astype(np.int64)
total = np.sum(counts)
print(np.mean(counts), np.min(counts), np.max(counts))
x = np.random.uniform(low=-250.0/1.74, high=250.0/1.74, size=total)
y = np.random.uniform(low=-250.0/1.74, high=250.0/1.74, size=total)
z = np.random.uniform(low=-250.0/1.74, high=250.0/1.74, size=total)
t = np.random.uniform(low=250.0, high=500.0, size=total)
btag = np.random.uniform(low=0.0, high=1.0, size=total)

jets = ak.zip(
    {
        "x": ak.unflatten(x, counts),
        "y": ak.unflatten(y, counts),
        "z": ak.unflatten(z, counts),
        "t": ak.unflatten(t, counts),
        "btag": ak.unflatten(btag, counts),
    },
[...]

PR #2441, opened 2023-05-01
test: growablebuffer move_to

ianna:
test if releasing memory earlier helps with https://github.com/scikit-hep/awkward/issues/2440


ianna:
@lgray - this PR implements a more aggressive memory release policy in building Awkward Arrays. Though, I do not see much difference with your test on my Mac. Could you, please, test it within a realistic analysis and see if it makes any difference? Thanks!

lgray:
Hi @ianna we're largely finding, I think, that macos is kind of weird when it comes to memory. The answers are very different w.r.t. linux. 

I'll try this out but if you have a linux machine to try it on - I'd check there!

jpivarski:
This PR is addressing an issue that has already been closed for other reasons, and it's months old. I'm going to guess that it is not needed anymore and will close it. If I'm wrong, we can reopen it.

ianna (commented):
@jpivarski - this PR tests an existing code. I don't think I need to keep the branch.

PR #2444, opened 2023-05-02
fix: proper unknown size broadcasting for all-regular

### LINKED ISSUE #2439: Assigning `[]` or `{}` into RecordArray typetracer fails
### Version of Awkward Array

2.1.4

The following script:
```python3
import json
import awkward as ak
import numpy as np

fromjson = {
    "class": "RecordArray",
    "fields": ["muon"],
    "contents": [
        {
            "class": "ListOffsetArray",
            "offsets": "i64",
            "content": {
                "class": "RecordArray",
                "fields": ["pt"],
                "contents": [
                    {
                        "class": "NumpyArray",
                        "primitive": "int64",
                        "inner_shape": [],
                        "parameters": {},
                        "form_key": "muon_pt!",
                    },
                ],
                "parameters": {},
                "form_key": "muon_record!",
            },
            "parameters": {},
            "form_key": "muon_list_outer!",
        },
    ],
    "parameters": {},
    "form_key": "outer!",
}

form = ak.forms.from_json(json.dumps(fromjson))

ttlayout, report = ak._nplikes.typetracer.typetracer_with_report(form, forget_length=True)

ttarray = ak.Array(ttlayout)

ttmuons = ttarray.muon

array = ak.Array(ttlayout.form.length_zero_array(highlevel=False))

array["emptydict"] = {}

muons = array.muon

muons["emptydict"] = {}

array["emptylist"] = []

muons["emptylist"] = []

# all four of the following fail

ttarray["emptydict"] = {}

ttmuons["emptydict"] = {}

ttarray["emptylist"] = []

ttmuons["emptylist"] = []
```
results in (and similar variants for different types being assigned):
```
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/highlevel.py", line 1027, in __setitem__
    self._layout = ak.operations.with_field(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_with_field.py", line 47, in with_field
    return _impl(array, what, where, highlevel, behavior)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_with_field.py", line 153, in _impl
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_broadcasting.py", line 1062, in broadcast_and_apply
    out = apply_step(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_broadcasting.py", line 1041, in apply_step
    return continuation()
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_broadcasting.py", line 1014, in continuation
    return broadcast_any_list()
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_broadcasting.py", line 563, in broadcast_any_list
    outcontent = apply_step(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_broadcasting.py", line 1027, in apply_step
    result = action(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_with_field.py", line 144, in action
    out = ak.contents.RecordArray(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/recordarray.py", line 161, in __init__
    raise TypeError(
TypeError: RecordArray 'contents' must use the same array library (backend): TypeTracerBackend vs NumpyBackend

  File "form_madness12.py", line 59, in <module>
    ttarray["emptylist"] = []
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/highlevel.py", line 1030, in __setitem__
    self._numbaview = None
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_errors.py", line 56, in __exit__
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_errors.py", line 66, in handle_exception
TypeError: RecordArray 'contents' must use the same array library (backend): TypeTracerBackend vs NumpyBackend

    ak.Array.__setitem__(
        self = <Array-typetracer [...] type='## * {muon: var * {pt: int64}}'>
        field_name = 'emptylist'
        field_value = []
    )
```

agoose77:
This PR touches the fairly important broadcasting logic to fix some of the unknown-length cases.

In general, it removes `if backend.nplike.known_data` blocks in favour of logic that explicitly handles unknown lengths.

Fixes #2439

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -568,9 +568,7 @@ def apply_step(
                     if dimsize_known_to_be_zero:
                         nextinputs.append(x.content[:0])
                     # If we have a known size=1 content, then broadcast it to the dimension size
-                    elif (
-                        dimsize_greater_than_one_if_known > 1 and x_size_known_to_be_one
-                    ):
+                    elif dimsize_greater_than_one_if_known and x_size_known_to_be_one:
                         nextinputs.append(
                             x.content[: x.length * x.size]._carry(
                                 size_one_carry_index, allow_lazy=False
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -67,6 +67,7 @@ def length_of_broadcast(inputs: Sequence) -> int | type[unknown_length]:
 def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
     maxlen = length_of_broadcast(inputs)
     nextinputs = []
+
     for x in inputs:
         if isinstance(x, Record):
             index = x.backend.index_nplike.full(maxlen, x.at, dtype=np.int64)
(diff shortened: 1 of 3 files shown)


src/awkward/operations/ak_with_field.py
```diff
@@ -81,8 +86,14 @@ def _impl(base, what, where, highlevel, behavior):
...
         behavior = behavior_of(base, what, behavior=behavior)
-        base = ak.operations.to_layout(base, allow_record=True, allow_other=False)
+        backend = backend_of(base, what, default=cpu)
+
+        base = ak.operations.to_layout(
+            base, allow_record=True, allow_other=False
+        ).to_backend(backend)
         what = ak.operations.to_layout(what, allow_record=True, allow_other=True)
+        if isinstance(what, (ak.contents.Content, ak.record.Record)):
+            what = what.to_backend(backend)
```
agoose77, line 96: These should be the same backend!

src/awkward/_broadcasting.py
```diff
@@ -497,68 +498,100 @@ def broadcast_any_record():
...
                             )
+                        )
+                    # Any unknown values or sizes are assumed to be correct as-is
+                    elif (
+                        dim_size is unknown_length
+                        or x.size is unknown_length
+                        or x.size == dim_size
+                    ):
+                        nextinputs.append(x.content[: x.length * x.size])
                     else:
```
agoose77, line 586: @jpivarski do you think that these rules are reasonable? It means that if we lose length information, we assume that neither the 1-length or 0-length broadcasting applies.

src/awkward/_broadcasting.py
```diff
@@ -76,7 +77,7 @@ def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
             nextinputs.append(
                 RegularArray(
                     x,
-                    x.length if x.backend.nplike.known_data else 1,
+                    x.length,
```
agoose77, line 80: We're now propagating unknown lengths into the layout tree

src/awkward/operations/ak_with_field.py
jpivarski: That's right, and if they're different, the new data being assigned into the array (the `what`) is the one that should move/be changed.

src/awkward/_broadcasting.py
jpivarski: If we don't know a RegularArray size at type-tracer time, we won't enter any specialized length-0 or length-1 logic, but we also don't know that the lengths are incompatible, so we'd return a type-tracer of unknown length—not raise an error.

At run-time (on a Dask worker), the length with either be length-0, length-1, or something else, and that something else will either be compatible or incompatible. The type-tracer time return value (an array with unknown length) is either compatible with the run-time result or the run-time result is an error. That's fine.

The only bad cases would be

  * if the type-tracer returned a result that is incompatible with the run-time result, such as a different number of dimensions (the type reported by the type-tracer is wrong), or
  * if the type-tracer raises an error that doesn't get raised at run-time (users would be able to do something in Awkward Array that they can't do in dask-awkward, and it would be hard for us to explain why).

As I understand it, what you have above is not either of these two bad cases.

jpivarski (approved):
This is good—see comments below.

src/awkward/_broadcasting.py
agoose77: I think so. The form should be preserved, at least. What might be different is that if the `_carry` only happens on the non-typetracer branch, then the form of the result would differ (`ListArray` vs `ListOffsetArray`) if the content is a `ListOffsetArray`. 

We have previously asserted that we want the "same" form for typetracer operations, but the above would break that. It's already a problem though — if lengths become type information, then we have to make an arbitrary choice.

PR #2447, opened 2023-05-09
feat: add validity error for duplicate record field

### LINKED ISSUE #2296: Add uniqueness of `fields` to `_validity_error`
### Version of Awkward Array

main

It's currently possible to create "invalid" `RecordArrays` that contain duplicate fields. We should check this.

agoose77:
Closes #2296


jpivarski (approved):
Right, because we want to avoid $\mathcal{O}(n \, \log n)$ checks in the constructor, so it will at least be a validity check.

PR #2448, opened 2023-05-09
ci: annotate pytest failures on GHA

### LINKED ISSUE #2405: Use pytest plugin to annotate failures on GHA

See https://pypi.org/project/pytest-github-actions-annotate-failures/

agoose77:
Closes #2405

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,7 +91,7 @@ jobs:
         run: pipx run nox -s diagnostics -- --check-spec-sorted
 
       - name: Install test requirements
-        run: python -m pip install -v -r requirements-test.txt
+        run: python -m pip install -v -r requirements-test.txt pytest-github-actions-annotate-failures
 
       - name: Test specification
         if: steps.cache-awkward-cpp-wheel.outputs.cache-hit != 'true'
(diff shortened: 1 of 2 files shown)


agoose77:
See `test_XXXX_failure.py` for a test that is deliberately failing to demonstrate this feature.

jpivarski (approved):
I see, although PyLint doesn't like tests that so obviously would always fail. Perhaps you can trick it like this:

```python
import sys

assert sys.platform == "wacky"
```

PyLint can't know what all of the platform types are, and that this might actually run on a wacky system someday. We, however, know that it will not. (If you want to be more sure of it, generate a long random name.)

agoose77:
> I see, although PyLint doesn't like tests that so obviously would always fail.

I'm not quite following here — I added these failing tests to trigger a failed run with annotations (on GitHub)
image

Incidentally, I don't see any pylint warnings about this trivial assertion (either on GHA, or locally)

jpivarski:
Oh! This was the intention! I had thought these were PyLint failures because PyLint has always been plugged into this system (correlating errors to code in GitHub). Also, PyLint could be smart enough to find out that this assertion would always fail, and that's why I was suggesting a way to fool it. But I misunderstood what was supposed to happen here.

agoose77:
Ah, I see! Well, if you're happy with this feature, I'll remove the failing test!

PR #2449, opened 2023-05-09
fix: `ak.to_numpy` does not properly convert lists of empty string

### LINKED ISSUE #2364: ak.to_numpy turns an empty array of type string into an empty array of type float64

agoose77:
Fixes #2364

### DIFF
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -1428,7 +1428,13 @@ class ListArray(Content):
         )
 
     def _to_backend_array(self, allow_missing, backend):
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
Good idea. That would be a surprise, only happening for empty arrays.

PR #2454, opened 2023-05-11
feat!: remove use of `typestr` in `ak.types.Type`

### LINKED ISSUE #1945: Check typestring handling
### Version of Awkward Array

main

`RegularType` has a case for regular strings / bytestrings. We never see this in the rendered type information, because we pass in the typestr to the `RegularType` object (I assume, anyway, I haven't checked).

Additionally, we have logic to quote Datashape identifiers. These don't look complete between `OptionType` and `RecordType`. This might be deliberate, but we should check and explain this logic with a comment or two.

agoose77:
Awkward associates `("__typestr__", "name")` behavior keys with custom typestrings for named types (records, arrays). Right now, that's implemented using a `typestr` argument. 

This PR drops `typestr` (deprecates), and instead introduces a `behavior` dict on the highlevel type, mirroring what already happens for arrays/records. Custom typestrings are resolved from this dict.

@jpivarski and I discussed this, and I was initially in favour of just building out the `typestr` argument usage. However, it feels to me that passing the behaviour to mirror our array/content implementation seems nicer (and keeps typestrings as contextual properties, rather than an intrinsic part of the type object)

This PR does the following:
- Remove string special casing in typestring resolution
- Simplifies typestring resolution
- Fixes a bug in `ak.types.from_datashape` (doesn't use new `ScalarType`)
- Changes typestring of regular strings to `string[N]` 

I also noticed some quirks that we should raise now:
- Awkward seems to support fixed-length strings (e.g. `ak.to_regular(["hell", "worl"]).type.show()`), but this pathway is never seen as we usually pass in a custom `typestr`.  
  I think the rule here is that typestrings are always used in favour of the type's default format, but the type can embellish that e.g. with fixed-length information?
- `repr(type)` shows custom typestrings. I think this should not be the case; typestrings are an aesthetic thing, not a fundamental part of the representation.

Closes #1945

### DIFF
--- a/src/awkward/_behavior.py
+++ b/src/awkward/_behavior.py
@@ -6,6 +6,7 @@ from collections.abc import Mapping
 
 import awkward as ak
 from awkward._nplikes import ufuncs
+from awkward._typing import Any
 
 
 def overlay_behavior(behavior: dict | None) -> Mapping:
@@ -125,34 +126,18 @@ def find_ufunc(behavior, signature):
                 return custom
 
 
-def find_typestrs(behavior):
+def find_record_typestr(behavior: None | Mapping, parameters: None | Mapping[str, Any]):
+    if parameters is None:
+        return None
     behavior = overlay_behavior(behavior)
-    out = {}
-    for key, typestr in behavior.items():
-        if (
-            isinstance(key, tuple)
-            and len(key) == 2
-            and key[0] == "__typestr__"
-            and isinstance(key[1], str)
-            and isinstance(typestr, str)
-        ):
-            out[key[1]] = typestr
-    return out
+    return behavior.get(("__typestr__", parameters.get("__record__")))
 
 
-def find_typestr(parameters, typestrs):
-    if parameters is not None:
-        record = parameters.get("__record__")
-        if record is not None:
-            typestr = typestrs.get(record)
-            if typestr is not None:
-                return typestr
-        array = parameters.get("__array__")
-        if array is not None:
-            typestr = typestrs.get(array)
-            if typestr is not None:
-                return typestr
-    return None
+def find_array_typestr(behavior: None | Mapping, parameters: None | Mapping[str, Any]):
+    if parameters is None:
+        return None
+    behavior = overlay_behavior(behavior)
+    return behavior.get(("__typestr__", parameters.get("__array__")))
 
 
 def behavior_of(*arrays, **kwargs):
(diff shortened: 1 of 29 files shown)


jpivarski (approved):
Actually passing the whole `behavior` around is fine for symmetry. Now that there isn't the Python/C++ split, there isn't a technical reason why this can't be done, and there's precedent for passing `behavior` around in `_reduce_next`; this isn't even as deep as that. (What I'm saying is that although a separation between `behavior` at high-level/not at low-level was originally intended, there are good reasons to give it up.)

> * Awkward seems to support fixed-length strings (e.g. `ak.to_regular(["hell", "worl"]).type.show()`), but this pathway is never seen as we usually pass in a custom `typestr`.

Fixed-length strings (and bytestrings) are intended, but they could be under-tested.

>   I think the rule here is that typestrings are always used in favour of the type's default format, but the type can embellish that e.g. with fixed-length information?

Strings (`string[5]` versus `string`) are the only case where I'd expect that. I wouldn't want to complicate the API by generalizing it, but maybe it could apply to only the string typeclass?

> * `repr(type)` shows custom typestrings. I think this should not be the case; typestrings are an aesthetic thing, not a fundamental part of the representation.

"Aesthetic": sort of. It's for [Datashape](https://datashape.readthedocs.io/en/latest/overview.html) compliance (which is loose, since we have to extend Datashape).

agoose77:
> Strings (string[5] versus string) are the only case where I'd expect that. I wouldn't want to complicate the API by generalizing it, but maybe it could apply to only the string typeclass?

I can't think of a reason _not_ to generalise to non-strings equally as strongly as I can think of a reason. I'm tempted to kick this bucket down the road until someone wants it as a feature?

jpivarski:
Sure, that sounds good. Since it only applies to RegularArrays, it's only overloads of `__array__`, not `__record__`. An overload for an `__array__` with a custom name that isn't a string is a... ? I can't think of what would fit those constraints.

agoose77:
> Sure, that sounds good. Since it only applies to RegularArrays, it's only overloads of `__array__`, not `__record__`. An overload for an `__array__` with a custom name that isn't a string is a... ? I can't think of what would fit those constraints.

It would be a custom array type, which has regular list semantics. It would only have this custom repr if the user _also_ sets `__typestr__`, which is fairly "you can do it if you want, but it might not be meaningful" in my view?

agoose77:
Based upon the above, I take your approval as go to merge!

Issue #2433, opened 2023-04-26
Implement nominal type inheritance


agoose77:
### Description of new feature

Library authors are currently required to do a reasonable amount of work to extend the types of other libraries. For example, to extend the `vector` types, one must both inherit from the Python classes and register them under new names, but also re-define all of the ufunc overloads in `ak.behavior`.

A related discussion is happening in #2432, whereby a new "typeclass" mechanism is proposed for string-like types. It's my belief #2432 is proposing a solution to the question "is this type like another type". We could implement such logic by defining type inheritance. For example, consider a custom vector type:
```python
ak.behavior["__supertype__", "CustomVector"] = "Vector"
```

Despite the motivations above, I can't think of many concrete use cases, yet. One example would be if the user evaluates `CustomVector + Vector`. Right now, that's an error. With the concept of super and subtypes, the result of `CustomVector + Vector` is now clearly defined; it should invoke the overload for `Vector + Vector`, i.e. return a `Vector`.

This issue is mainly book-keeping; I don't think it's a pressing issue, and I'd want to think it through more before deeming it a sensible direction to take things.

jpivarski:
Here is what one must do to define a "subclass" for a record behavior now:

```python
import numpy as np
import awkward as ak
import vector
vector.register_awkward()


class MomentumRecord4DWithCharge(vector.backends.awkward.MomentumRecord4D):
    pass


class MomentumArray4DWithCharge(vector.backends.awkward.MomentumArray4D):
    pass


ak.behavior["Momentum4DWithCharge"] = MomentumRecord4DWithCharge
ak.behavior["*", "Momentum4DWithCharge"] = MomentumArray4DWithCharge


def add_them(x, y):
    vector_x = ak.with_name(x, "Momentum4D")
    vector_y = ak.with_name(y, "Momentum4D")
    vector_out = ak.with_name(vector_x + vector_y, "Momentum4DWithCharge")
    vector_out["charge"] = x["charge"] + y["charge"]
    return vector_out


ak.behavior[np.add, "Momentum4DWithCharge", "Momentum4DWithCharge"] = add_them
```

Then, given two arrays with the new name:

```python
array1 = ak.zip({
    "px": [[1.1, 2.1], [], [3.1]],
    "py": [[1.2, 2.2], [], [3.2]],
    "pz": [[1.3, 2.3], [], [3.3]],
    "E": [[10.0, 20.0], [], [30.0]],
    "charge": [[+1, -1], [], [-1]],
[...]

PR #2456, opened 2023-05-11
feat!: rework string abstraction

### LINKED ISSUE #2432: Simplify customisation that is used to implement strings

There has been a long-standing awareness that our string abstraction is both a special case (privileged in our codebase) and also something that users might want to customise. Consequently, we have mechanisms like custom broadcasting that users can hook into to implement the same features.

Now that Awkward is more mature, we can more conclusively say that few/no users are actually making use of these mechanisms (@jpivarski might have data on this, or I can grep it!)
However, this might be a chicken-and-egg problem; the extension system is not sufficiently mature or comprehensive that users could implement strings without Awkward's special casing.

Specifically, we have known bugs that are trickier to solve for our extension mechanisms, such as string-string broadcasting. 

There are clear use cases for extending strings, e.g. IP addresses:
- Users want IP addresses to broadcast like strings (string-like)
- Users want IP addresses to have IP address methods (IP address-like)

Now, IP addresses could be implemented as tuples (@jpivarski could you expand on the reason not to do this?), but let's consider that we want to implement them as strings. Really, as @jpivarski pointed out in a meeting today, strings are a special example of binary data (an escape hatch, of sorts). 

It would appear that the best path forward is to simplify our extension mechanisms to make string-like custom types _easier_, whilst removing the kinds of dials that users can turn.
Namely, we could introduce a string-like table to `ak.behavior` that enumerates the string-like type names (that are used with `__array__`). This enables other types to gain string-like broadcasting, without opening the door to esoteric broadcasting rules. Note that to gain string-like _methods_, the names placed in this table should correspond to array classes that _inherit_ from the string behaviour classes. Note that, only some of Awkward's logic is implemented in behaviour classes — much is implemented inside the Awkward code base.

Specifically, string-like names imply the following 2D list containing a 1D `NumpyArray` of `uint8`. Concretely the "string-like" typeclass (borrowed from Haskell) means that the list of list of `uint8` should be treated as a list of atoms. By applying such a restrictive definition, we can avoid ambiguity when handling all of the string-related Awkward rules.

This relates to another conversion on type _inheritance_. That will be discussed separately.

## Related Issues
- #1682

agoose77:
Closes #2432 

- [x] Support for string-likes in `getitem`
- [x] Support for string-likes in `_remove_structure`
- [x] Support for string-likes in `_to_backend_array`
- [x] Support for string-likes in `_combinations`
- [x] Support for string-likes in `_argsort_next`
- [x] Support for string-likes in `_sort_next`
- [x] Support for string-likes in `_unique`
- [x] Support for string-likes in `_is_unique`
- [x] Support for string-likes in `_numbers_to_type`
- [x] Support for string-likes in `_to_arrow`
- [x] Support for string-likes in `_reduce_next`
- [ ] Validity error for string-likes (list of numpy invariant) [^1]
- [ ] Figure out whether to make `behavior: Mapping` internally 

[^1]: should we allow overloading validity error for any type name?

### DIFF
--- a/src/awkward/_behavior.py
+++ b/src/awkward/_behavior.py
@@ -144,10 +144,20 @@ def find_array_typestr(
     return behavior.get(("__typestr__", parameters.get("__array__")), default)
 
 
-def is_subtype(behavior: None | Mapping, type_name: str, supertype_name: str) -> bool:
+def is_subtype(
+    behavior: None | Mapping,
+    type_name: str | None,
+    supertype_names: str | tuple[str, ...],
+) -> bool:
+    if type_name is None:
+        return False
+
+    if isinstance(supertype_names, str):
+        supertype_names = (supertype_names,)
+
     behavior = overlay_behavior(behavior)
     while type_name is not None:
-        if type_name == supertype_name:
+        if type_name in supertype_names:
             return True
         type_name = behavior.get(("__super__", type_name))
     return False
--- a/src/awkward/_behavior.py
+++ b/src/awkward/_behavior.py
@@ -71,25 +71,6 @@ def find_custom_cast(obj, behavior):
     return None
 
 
-def find_custom_broadcast(layout, behavior):
-    behavior = overlay_behavior(behavior)
-    custom = layout.parameter("__array__")
-    if not isinstance(custom, str):
-        custom = layout.parameter("__record__")
-    if not isinstance(custom, str):
-        custom = layout.purelist_parameter("__record__")
-    if isinstance(custom, str):
-        for key, fcn in behavior.items():
-            if (
-                isinstance(key, tuple)
-                and len(key) == 2
-                and key[0] == "__broadcast__"
-                and key[1] == custom
-            ):
-                return fcn
-    return None
-
-
 def find_ufunc_generic(ufunc, layout, behavior):
     behavior = overlay_behavior(behavior)
     custom = layout.parameter("__array__")
--- a/src/awkward/_behavior.py
+++ b/src/awkward/_behavior.py
@@ -6,6 +6,7 @@ from collections.abc import Mapping
 
 import awkward as ak
 from awkward._nplikes import ufuncs
+from awkward._typing import Any
 
 
 def overlay_behavior(behavior: dict | None) -> Mapping:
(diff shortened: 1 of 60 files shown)


agoose77:
This PR starts work on nominal type inheritance, with the a new `stringlike` type (inherited from by `string` and `bytestring`). This type can be used to build new binary stream views that are neither strings nor bytes. 

Introducing nominal inheritance raises the question of where else it should apply. For example, with ufuncs of `X + Y`, if only `X + X` is defined and `Y` is a subtype of `X`, do we support this? The answer is whether we are implementing OOP or type classes. I think I prefer type classes, i.e. `X + Y` is only permitted if `X` and `Y` belong to the same type (unless the user implements an explicit overload)

In https://github.com/scikit-hep/awkward/issues/2433 we touch on this. I'm in agreement that something which performs the `with_name` dance, and expands the permutations of `X, Y` would be useful.

Incidentally, maybe we should expose the `is_subtype` functionality so that library authors can test it.

agoose77:
@jpivarski it would be useful to discuss this effort in our weekly meeting. In anticipation of that, here's what I think we'll need to touch upon:

Fundamentally, this PR is redefining strings as something that is behavior-dependent. I think this is the _right_ move; strings are a kind of composed type that requires this additional context. However, this means that many of our API methods need to be modified to accept the behavior dictionary, and this is challenging for descriptors like properties, e.g. `Form.purelist_depth` et al.

A trivial solution to all of this is to set the behavior in a top-level context. I'm not really sure that I like this suggestion — whilst it's a lot cleaner internally, it means that L2 users would need to be aware of such a context, i.e. we're just kicking the problem to a much less visible part of the room.

We can of-course make this API change if we need to. We can introduce new methods like
- ```python
  def depth(
      behavior: Mapping, stop_at_strings: bool = True, stop_at_records: bool = True
  ) -> tuple[bool, int]:
      ...
  ```
- ```python
  def depth_limits(
      behavior: Mapping, stop_at_strings: bool = True, stop_at_records: bool = True
  ) -> tuple[int, int]:
      ...
  ```
which give us more control than the existing `purelist_depth`, `minmax_depth`, `branch_depth`.

I've been thinking — is there any way to avoid needing `behavior` everywhere? One approach would be to introduce a nominal _subtype_, i.e. custom strings have the following parameters
```python
{
    "__array__": "string",
    "__flavor__": "my-string"
}
```

This is inelegant for a couple of reasons:
1. `__array__` no longer exclusively defines an array's type;
2. Subtype relations are now defined on every instance of an array instead of in a global table (`ak.behavior`).
3. Rules for how the standard flavor interacts with custom flavors would need to be considered.

So, I'm not yet in favor of this second approach, but I mention it for posterity.

jpivarski:
My thoughts, copied from Slack:

I think I have exactly the same quantum admixture of opinions about introducing a `behavior` argument everywhere versus introducing a `__flavor__` parameter. I prefer the `behavior` everywhere but the `__flavor__` idea is interesting, _maybe_ better.

`__flavor__` can only be a string, matched to Python classes in a similar way as `__array__`, but a different implementation. That's the downside.

You end up with "multiple ways to do it."

It's probably for the best that all of those depth-parameters become a single method with arguments. They were parameters to simplify writing the pybind11, which is all ancient history now.

agoose77:
> My thoughts, copied from Slack:
> 
> I think I have exactly the same quantum admixture of opinions about introducing a behavior argument everywhere versus introducing a __flavor__ parameter. I prefer the behavior everywhere but the __flavor__ idea is interesting, maybe better.

Good to know we both feel similarly :stuck_out_tongue: It would have been nice if you had held a very strong opinion in one direction so that we didn't need to weigh up both arguments ;)


> `__flavor__` can only be a string, matched to Python classes in a similar way as `__array__`, but a different implementation. That's the 
> 
> You end up with "multiple ways to do it."

Yep, big drawback.

> It's probably for the best that all of those depth-parameters become a single method with arguments. They were parameters to simplify writing the pybind11, which is all ancient history now.

I think _this_ can be a starting point — implementing this can happen orthogonally to whether behaviour should _also_ be an argument.

agoose77:
At this point, I'm fairly confident that we are not going to implement custom strings using nominal inheritance, so this PR can be closed!

PR #2459, opened 2023-05-15
refactor: reorganise sentinel definitions

agoose77:
For a while we've wanted a `Sentinel` class, but it's not been clear where to put it such that our imports don't tangle. After #2131 and friends, we have much leaner modules, and it's safe to put this in `ak._util`.

Also, generally PEP8 encourages uppercase constants, which I consider sentinels to be.

### DIFF
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -2,18 +2,16 @@
 from __future__ import annotations
 
 import os
+import struct
 import sys
 from collections.abc import Collection
 
 import packaging.version
 
-from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._typing import TypeVar
 
-np = NumpyMetadata.instance()
-
 win = os.name == "nt"
-bits32 = np.iinfo(np.intp).bits == 32
+bits32 = struct.calcsize("P") * 8 == 32
 
 # matches include/awkward/common.h
 kMaxInt8 = 127  # 2**7  - 1
--- a/src/awkward/_util.py
+++ b/src/awkward/_util.py
@@ -80,18 +80,21 @@ def identifier_hash(str):
     )
 
 
-# FIXME: introduce sentinel type for this
-class _Unset:
+class Sentinel:
+    """A class for implementing sentinel types"""
+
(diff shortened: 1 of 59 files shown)


src/awkward/_util.py
```diff
@@ -2,18 +2,16 @@
...
 import packaging.version
 
-from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._typing import TypeVar
 
-np = NumpyMetadata.instance()
-
 win = os.name == "nt"
-bits32 = np.iinfo(np.intp).bits == 32
+bits32 = struct.calcsize("P") * 8 == 32
```
agoose77, line 15: This is the only non-sentinel change

jpivarski: Either way, you'd asking the system whether `sizeof(size_t)` is equal to 4. `np.iinfo` goes through NumPy's code and `struct.calcsize` goes through Python's, but this is determined by whatever compiled NumPy and Python. (And if they weren't both 32-bit or both 64-bit, lots of things would fail.)

But okay.

jpivarski (approved):
All-caps for constants adheres to Python's style. Aesthetically, these things appear in Sphinx-generated documentation, and I thought that

```python
copy(parameters=unset)
```

looked better than

```python
copy(parameters=UNSET)
```

but in retrospect, this is better anyway. We might as well tell users (in the documentation) that `UNSET` is some constant.

src/awkward/_util.py
agoose77: Yep - the motivation for this is to remove the nplike import in favour of a built-in, so that we can decouple modules (we don't need to, but it's long term preferable)

agoose77: Yep - the motivation for this is to remove the nplike import in favour of a built-in, so that we can decouple modules (we don't need to, but it's long term preferable)!

agoose77: Yep - the motivation for this is to remove the nplike import in favour of a built-in, so that we can decouple modules (we don't need to, but it's long term preferable)!

PR #2460, opened 2023-05-15
fix: handle indexed nodes in `ak.almost_equal`

agoose77:
This PR:
- Adds handling of `IndexedArray` nodes to `ak.almost_equal` (these come from carrying a `RecordArray`)
- Replaces use of `layout.to_packed()` with explicit packing logic

(2) makes our code more predictable; there are exemptions to packing rules that make it more of an obfuscation than code re-use to rely on `to_packed()`.

By simplifying the switch case, we now don't explicitly catch non-implemented cases. I think, however, this logic is already exhaustive in the valid combinations (`list-list`, `option-option`, etc.)


jpivarski (approved):
I remember this conversation. It's the right thing to do.

Issue #2462, opened 2023-05-16
Running over zero-length arrays, time spent mostly in function overhead


lgray:
### Version of Awkward Array

2.2.0

### Description and code to reproduce

N.B.: This is lower priority, but it scales with analysis size/complexity!

When running column optimization in dask awkward a typetracer is passed through the whole analysis code.
This triggers all relevant function calls for the analysis and produces an accurate accounting of data needs which is great, but as analyses grow larger (especially with systematics overhead!) the tracing begins to take quite some time for processing no data.

I understand this is only so reducible, but responsiveness is definitely part of user experience and it would be sad if we lost users because of really excellent core functionality taking too long.

Here is a pyinstrument trace showing what functions stacks are grabbing the most time in a mostly realistic analysis with full systematics:
```
  _     ._   __/__   _ _  _  _ _/_   Recorded: 08:25:40  Samples:  3840
 /_//_/// /_\ / //_// / //_'/ //     Duration: 3.983     CPU time: 3.988
/   _/                      v4.3.0

Program: yimu_test5.py

3.983 <module>  yimu_test5.py:1
└─ 3.983 necessary_columns  dask_awkward/lib/inspect.py:11
   └─ 3.983 _necessary_columns  dask_awkward/lib/optimize.py:392
      └─ 3.979 _get_column_reports  dask_awkward/lib/optimize.py:335
         ├─ 3.900 get_sync  dask/local.py:551
         │  └─ 3.898 get_async  dask/local.py:350
         │     ├─ 3.382 fire_tasks  dask/local.py:452
         │     │  └─ 3.331 submit  dask/local.py:539
         │     │     └─ 3.292 batch_execute_tasks  dask/local.py:234
         │     │        └─ 3.292 <listcomp>  dask/local.py:238
         │     │           └─ 3.287 execute_task  dask/local.py:214
         │     │              └─ 3.279 _execute_task  dask/core.py:82
         │     │                 └─ 3.254 __call__  dask/optimization.py:987
         │     │                    └─ 3.246 get  dask/core.py:128
         │     │                       └─ 3.211 _execute_task  dask/core.py:82
         │     │                          ├─ 1.186 __array_ufunc__  awkward/highlevel.py:1289
         │     │                          │  └─ 1.138 array_ufunc  awkward/_connect/numpy.py:219
         │     │                          │     ├─ 0.496 _array_ufunc_custom_cast  awkward/_connect/numpy.py:147
[...]

PR #2464, opened 2023-05-17
feat: improve performance of behavior lookups

agoose77:
Whilst it is not profitable for us to try and optimise Awkward's metadata performance too strongly, these fixes are ones that I've been thinking of making for a while, and #2462 justifies it.


jpivarski (approved):
Right—I always found it a little uncomfortable that we had to search through a dict by iteration in O(n) time, though I thought it couldn't be helped. The dict was just for interface, not for lookup efficiency. But these changes look good!

PR #2466, opened 2023-05-22
fix: convert index to shape item

agoose77:
This bug was caught during dask-awkward development.


jpivarski (approved):
Right, `offsets[-1]` is being converted from a data value into a shape item. Good catch!

jpivarski (approved):
Right, `offsets[-1]` is being converted from a data value into a shape item. Good catch!

PR #2469, opened 2023-05-23
fix: touch offsets in kernel call

agoose77:
This kernel reads the values of `offsets`, so it should trigger a touch.

I recall seeing a place in awkward where we didn't make these touches before invoking this function; really, we should perform touches as-close-to the functions that would actually touch the

### DIFF
--- a/src/awkward/contents/listarray.py
+++ b/src/awkward/contents/listarray.py
@@ -386,6 +386,7 @@ class ListArray(Content):
     def _broadcast_tooffsets64(self, offsets):
         if not self.backend.index_nplike.known_data:
             self._touch_data(recursive=False)
+            offsets.touch_data()
         return ListOffsetArray._broadcast_tooffsets64(self, offsets)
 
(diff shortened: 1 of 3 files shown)


agoose77:
@jpivarski I've been thinking about a more robust solution to this problem; the typetracer kernel implementations should touch any typetracer arguments, removing the need to remember to do this. Do you see any issue to this before I get started in a later PR?

jpivarski (approved):
Looks good!

As for automatically touching TypeTracer attributes in a kernel call, be sure that it's only for those TypeTracerArrays that are actually passed into the kernel. (Also, I think it already does that. Isn't there a `touch_data` in the TypeTracer generic implementation of kernel calls?)

agoose77:
> [...] (Also, I think it already does that. Isn't there a touch_data in the TypeTracer generic implementation of kernel calls?)

No, surprisingly we do this manually at each call site! I think we can remove all of those statements accordingly.

jpivarski:
That sounds like a good idea.

PR #2471, opened 2023-05-24
fix!: raise error when flattening strings

### LINKED ISSUE #2455: `ak.flatten` flattens strings with `axis != None`
### Version of Awkward Array

main

We are leaning towards strings being a robust abstraction — if you want to erase a string, remove the parameters (or use `ak.enforce_type`). 

However, there are some holes in this, notably with `ak.flatten`:
```python
>>> ak.flatten(["hello", "moto"], axis=1)
'hellomoto'
```

We should ensure that this raises an error.

agoose77:
We consider strings to be a relatively strong abstraction, so we should error when trying to flatten them at axis=0. To be able to flatten these arrays, users should change their type, e.g. with `ak.ensure_type`

```python
>>> import awkward as ak
>>> ak.flatten(["this", "and", "that"], axis=1)
ValueError: array of strings cannot be directly flattened. To flatten this array, drop the `"__array__"="string"` parameter using `ak.enforce_type`, `ak.with_parameter`, or `ak.without_parameters`/

    ak.flatten(
        array = ['this', 'and', 'that']
        axis = 1
        highlevel = True
        behavior = None
    )
>>> ak.flatten(
...     ak.enforce_type(["this", "and", "that"], "var * char"),
... axis=1)
'thisandthat'
```

Fixes #2455

### DIFF
--- a/src/awkward/contents/listoffsetarray.py
+++ b/src/awkward/contents/listoffsetarray.py
@@ -718,6 +718,15 @@ class ListOffsetArray(Content):
             raise AxisError("axis=0 not allowed for flatten")
 
         elif posaxis is not None and posaxis + 1 == depth + 1:
+            if (
+                self.parameter("__array__") == "string"
+                or self.parameter("__array__") == "bytestring"
+            ):
+                raise ValueError(
+                    "array of strings cannot be directly flattened. "
+                    'To flatten this array, drop the `"__array__"="string"` parameter using '
+                    "`ak.enforce_type`, `ak.with_parameter`, or `ak.without_parameters`/"
+                )
             listoffsetarray = self.to_ListOffsetArray64(True)
             stop = listoffsetarray.offsets[-1]
             content = listoffsetarray.content._getitem_range(0, stop)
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
You're right, I had missed this. (I don't remember seeing it and thinking, "I'll get back to this." I just didn't see it at all.)

It's reasonable to say that strings are un-flattenable, with instructions for how to make it not a string and therefore flattenable. On the other hand, there are probably a lot of other methods like this one. Perhaps it could be generalized inside `recursively_apply` and `broadcast_and_apply`, but there are also operations defined as layout methods that would have to be excluded one by one (like `flatten`).

Like, should we exclude `ak.local_index` on a string? (Maybe someone would want positions of each character...)

agoose77:
> It's reasonable to say that strings are un-flattenable, with instructions for how to make it not a string and therefore flattenable. On the other hand, there are probably a lot of other methods like this one. Perhaps it could be generalized inside recursively_apply and broadcast_and_apply, but there are also operations defined as layout methods that would have to be excluded one by one (like flatten).
> 
> Like, should we exclude ak.local_index on a string? (Maybe someone would want positions of each character...)

I agree. My current stance is that `__array__` is meaningful, and we should always respect it. If users want e.g. character positions, they must drop the string abstraction.

Meanwhile, I *think* `recursively_apply` could implement a flag. I don't think we want to preclude the ability to use it to recurse into the character array, but you're right that we should enforce this decision automatically instead of opt-in.

PR #2472, opened 2023-05-24
fix: always touch arguments to kernels

agoose77:
This PR ensures that any buffers passed into a kernel are always touched. In time, we can start removing the manual touching that is associated with kernel calls.


jpivarski (approved):
This is exactly what we want. I know I intended to do this—touching all arguments that go into a kernel here at the TypeTracer kernel implementation—but I suppose that I didn't remember that before finishing it.

PR #2473, opened 2023-05-24
chore: add more type hints!

agoose77:
We should keep adding type hints, so I made another pass.

### DIFF
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -306,13 +306,13 @@ class BitMaskedArray(Content):
             parameters=self._parameters,
         )
 
-    def _touch_data(self, recursive):
+    def _touch_data(self, recursive: bool):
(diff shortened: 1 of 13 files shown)


jpivarski (approved):
More type hints are always good.

(I'm working on the API survey now, and type hints are essential for identifying that someone is using `XYZ.method` because type inference is a necessary step toward knowing that it's `XYZ`'s method, rather than `ABC`'s method. We're using [jedi.Script.infer](https://jedi.readthedocs.io/en/latest/docs/api.html#jedi.Script.infer) to [obtain the types of variables](https://github.com/jpivarski-talks/2023-05-09-chep23-analysis-of-physicists/blob/6ea152ebc9f6e3d1b0f1b5d044848aafa6dba816/analysis/smarter-ast-scan.py#L294-L296) in user code, but that type inference won't work without [type hints or docstrings with types](https://jedi.readthedocs.io/en/latest/docs/usage.html#type-hinting).)

PR #2474, opened 2023-05-24
feat!: drop string broadcasting overloading

agoose77:
We are planning to remove this feature and replace it with more string-focussed customisation. This PR removes the custom broadcasting in anticipation for that change.

Partially addresses #2432


jpivarski (approved):
That's right, this is the project we discussed, in which string behaviors will be hard-coded, but new behaviors can opt-into behaving like strings. (This is just the hard-coding step.)

PR #2475, opened 2023-05-24
feat!: deprecate `NumpyArray.strides`

agoose77:
We have no internal need for this property, and I can see that we might want to avoid relying on such information being trivially available. It's not guaranteed that all backends implement this e.g. `jax.Array`.

### DIFF
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -8,7 +8,7 @@ from awkward._backends.backend import Backend
 from awkward._backends.dispatch import backend_of
 from awkward._backends.numpy import NumpyBackend
 from awkward._backends.typetracer import TypeTracerBackend
-from awkward._errors import AxisError
+from awkward._errors import AxisError, deprecate
 from awkward._layout import maybe_posaxis
 from awkward._nplikes import to_nplike
 from awkward._nplikes.jax import Jax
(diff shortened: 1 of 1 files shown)


jpivarski (commented):
Wait—I don't think this is a good idea. `strides` is still an important part of the `NumpyArray` data model, like `shape`. Since we pass through `shape` (instead of making people type `x.data.shape`), by symmetry, we should pass through `strides`, too (or neither).

JAX can't have anything but C-contiguous strides, but it does have strides. If the `JAXDeviceArray` class doesn't have a `strides` member, then its `strides` can be computed from `shape` and `itemsize`. In fact, if `x.data.strides` doesn't exist (because `x.data` is a `JAXDeviceArray`), `x.strides` would be a more reliable, generic way to get strides because `NumpyArray.strides` can compute it if `self._backend.has_strides` is false.

```python
>>> array = np.arange(2*3*5).reshape(2, 3, 5)
>>> array.flags.c_contiguous
True
>>> array.shape
(2, 3, 5)
>>> array.strides
(120, 40, 8)
>>> array.itemsize
8
>>> tuple(np.cumprod((array.itemsize,) + array.shape[:0:-1])[::-1])
(120, 40, 8)
```

agoose77:
My thinking was - not all Array objects implement strides, so we don't need to pass it through the `NumpyArray` unless there's a good reason. In the back of my mind, I'm thinking about minimising how many details we expose about our arrays beyond "it's contiguous" and "it's not contiguous".

In any case, we can close this, I won't push it!

PR #2476, opened 2023-05-25
refactor: cleanup and type hints

agoose77:
This PR is a bag of changes:
- Add some additional type hints
- Refactor logic that depends upon `NumpyArray.is_contiguous` (to be more explicit).
- Replace use of `.strides` with nplike method


jpivarski (approved):
Looks good!

I like the use of computed strides. The strides always exist, even for backends for which there is no choice about what the strides may be.

PR #2477, opened 2023-05-25
fix: check `start_at_zero` in `_compact_offsets`

agoose77:
This PR 
- fixes `start_at_zero` in `_compact_offsets`
- improves the performance of `to_ListOffsetArray64` via `_broadcast_tooffsets64`. 
- removes kernels that can be replaced with `nplike` calls[^perf]

In `origin/main`, calling `to_ListOffsetArray64` for an array that does not have `int64` `offsets` invokes `_broadcast_tooffsets64`, which performs a `carry` of the list content. In this instance, we do not need to carry; a simple range-slice is sufficient.

Along the way, some kernels that aren't strictly required have been replaced with nplike calls.

[^perf]: n.b. this _may_ lead to a performance regression. We should re-introduce these kernels in such cases.


jpivarski (approved):
I looked at this earlier when we talked about it on a Slack Huddle. I think it's good to replace Awkward1-era kernels that can now be replaced by nplike calls. It also means there will be fewer CUDA kernels to implement.

agoose77:
Something peculiar is going on with the x86 test. The traceback doesn't initially make sense to me ­— I'll need to dig into this further.

agoose77:
@jpivarski the tests are failing on 32-bit systems. It looks likely NumPy doesn't support `int64` types in some contexts (I think they use `intp` under the hood) without unsafely casting. 

I think this means we *can't* drop the `size=1` broadcasting kernel, as we need to support this dtype. Does this follow?

Alternatively, we forcibly cast, knowing that `intp` is expected to represent all valid sizes of allocatable objects.

jpivarski:
We can forcibly cast (and do that in other circumstances). That's the reason that "64" is in the name of the function; we really mean `int64_t` and not `ssize_t`. If NumPy is doing `intp` casts, then we should explicitly cast to `int64`. That's essentially what the Awkward1-era kernels were doing (with lots of `(int64_t)xyz` in the code), and it doesn't make sense to bring back the kernels when that's semantically equivalent to casting NumPy (and casting NumPy gets us to smaller code).

agoose77:
> We can forcibly cast (and do that in other circumstances). That's the reason that "64" is in the name of the function; we really mean `int64_t` and not `ssize_t`.

To be clear, this is currently the bug - NumPy internally performs a cast to `int32` (which I assume is what `intp` aliases to), and this is not a safe cast, so it fails. We would need to cast `counts` to `intp` (or `int32`) for this to work, which is an unsafe cast to a smaller width on 32-bit systems (see 16e40ef). I'm somewhat comfortable doing that; in-memory Awkward arrays are limited in length by the same rules that permit `intp` to represent any array length in C. So, even though `Index64` offsets are larger than `intp` on 32-bit, they can only hold meaningful values `< np.iinfo(np.intp).max`. This axiom would fall down if we used this logic in e.g. Dask, however, it is not a problem given that Dask manages these indices separately.

jpivarski (approved):
We've been talking about this/looking at it in our meeting, and it looks good to me!

I think an assertion on 32-bit is the best way to ensure that the 64-bit → 32-bit conversion is not lossy (even though it seems like a theorem that it should not be).

PR #2478, opened 2023-05-26
feat: add `Index._touch_XXX` methods

agoose77:
This brings parity with `Content._touch_data`, slightly improving readability.

### DIFF
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -316,14 +316,12 @@ class BitMaskedArray(Content):
         )
 
     def _touch_data(self, recursive: bool):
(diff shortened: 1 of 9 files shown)


jpivarski (approved):
In our meeting today, I +1'ed this idea (in another PR's context): having the `if not backend.known_data:` inside the `touch_data` or `touch_shape`.

On a related note, when we were looking at possible inclusion of Awkward in Tiled, we had to use the `_touch_data` hidden method, just as Coffea already has. This is a good candidate for becoming a public API. What we've learned from this phase of experimenting with it as a private interface is that we only need touch-data and not also touch-shape.

agoose77:
OK - we'll make a separate PR to make these methods public, then!

PR #2479, opened 2023-05-26
feat: add placeholder array

agoose77:
As a memory optimisation, `dask-awkward` needs to be able to introduce placeholder arrays for buffers whose typetracers never touch data. 

This PR adds a placeholder array that is guaranteed to survive basic non-touching operations, with dtype and length metadata.

A consequence of this PR is the expectation that these placeholders may have `unknown-length` shapes. Concrete nplikes like CuPy don't support this via e.g. `shape_item_as_index`, but it might be that this is acceptable.


jpivarski (approved):
As we discussed in our meeting today, a `PlaceholderArray` that can be used in any `Backend` (NumPy, CuPy, JAX, etc.) would have to be different from the `TypeTracerArray` whose `Backend` is only `TypeTracer`, and that's fine. Even though both are placeholdery, they hold places in different ways for different reasons and even have different implementations (since `TypeTracerArray` has to be computable and predict types).

I also think this belongs in the Awkward codebase itself, rather than in dask-awkward (or elsewhere), because Awkward needs to know about it to not complain that it found a `PlaceholderArray` where it was expecting an `np.ndarray`.

src/awkward/_broadcasting.py
```diff
@@ -594,69 +600,28 @@ def broadcast_any_list():
...
-            parameters = parameters_factory(len(outcontent))
-            return tuple(
-                ListOffsetArray(offsets, x, parameters=p)
-                for x, p in zip(outcontent, parameters)
-            )
-
         # Not all regular, but all same offsets?
         # Optimization: https://github.com/scikit-hep/awkward-1.0/issues/442
-        elif all_same_offsets(backend, inputs):
+        elif index_nplike.known_data and all_same_offsets(backend, inputs):
```
agoose77, line 605: This is an optimisation, so we don't need to take it for typetracer

src/awkward/_broadcasting.py
```diff
@@ -594,69 +600,28 @@ def broadcast_any_list():
...
-                    if starts.length == 0 or stops.length == 0:
+                    if (starts.length is not unknown_length and starts.length == 0) or (
+                        stops.length is not unknown_length and stops.length == 0
+                    ):
                         nextinputs.append(x.content[:0])
                     else:
-                        lencontent = backend.index_nplike.max(stops)
+                        lencontent = index_nplike.index_as_shape_item(
+                            index_nplike.max(stops)
+                        )
```
agoose77, line 624: This is not strictly required; `known_data` implies that all index values are nearly valid shape items. Technically, `index_as_shape_item` also calls `int()` to avoid propagation of `int64` etc types.

src/awkward/_nplikes/array_module.py
```diff
@@ -92,6 +101,9 @@ def arange(
         *,
         dtype: np.dtype | None = None,
     ) -> ArrayLike:
+        assert not isinstance(start, PlaceholderArray)
+        assert not isinstance(stop, PlaceholderArray)
+        assert not isinstance(step, PlaceholderArray)
```
agoose77, line 106: For now, be certain we don't see placeholders in unexpected locations

agoose77:
@jpivarski I added some assertions, and fixed some broadcasting bugs (where known sizes in typetracer weren't correct). If you'd be so kind as to pass your eyes over the changes, then merge at your discretion?

One aspect to this PR that is on my radar is whether introducing `unknown_length` in non `known_data=False` contexts might be problematic. The `index_nplike.shape_item_as_index` will always fail for `unknown_length` (it doesn't map to a valid index), so any code that e.g. uses the length of a placeholder `Index` (with unknown length) in some secondary calculation would fail.

That is to say - we have been assuming that `touch_shape` is not needed, and we have not been using it in dask-awkward. But, it's possible that we do need the shape in some contexts.

I suppose this is where our re-hydration PR falls short of the ideal case, where we can query the buffer dtype and sizes from the metadata without reading the contents themselves. This would obviate the need for `unknown_length` in `PlaceholderArray`.

agoose77:
Then again, what's the alternative? It's possible (I don't know) that it's non-trivial to determine only the lengths of the offsets of a ragged column in Parquet. In such a case, we could not apply this optimisation if we also required known lengths.

Perhaps this serves in favour of `unknown_length` outside of typetracer contexts, and suggests that we should be mindful of this in code that e.g. creates new `Index` objects to take a known length.

jpivarski:
In Parquet version 2 (not universally adopted), it's possible to read just the definition and repetition levels and reconstruct the `offsets` of a jagged array without the `content`.

But If it's only the overall length (single integer) that you're talking about, that's always available in metadata and can be read independently of the data for any version of Parquet.

I'll be reviewing this PR soon.

jpivarski (approved):
Wow, virtually everything has to be checked for

```python
assert not isinstance(x, PlaceholderArray)
```

But only in nplike implementations. With that abstraction, we thankfully don't have to pepper all of the src/awkward/operations code with these checks. It's pulling its weight!

This looks ready to go, and since you've given me the go-ahead, I'll merge it.

PR #2486, opened 2023-06-02
fix: make starts and stops return length-1 arrays for empty lists

### LINKED ISSUE #2484: Reducers raise IndexError on masked three dimensional array
### Version of Awkward Array

2.2.1

```python
import awkward as ak
mask = ak.Array([False, False])
array = ak.mask(ak.unflatten(ak.mask(ak.unflatten([5, 6], [1,1]), [True, True]), [1,1]), mask)

ak.any(array, axis=2)
```
The last line raises `IndexError: index 0 is out of bounds for axis 0 with size 0`. However, I would except it to return `<Array [None, None] type='2 * option[var * ?bool]'>`
If `mask` has any elements that are true, it works.
I also tested this with `ak.all` and `ak.sum` with the same results.

agoose77:
This PR brings `ListOffsetArray.starts` and `stops` in line with `ListArray.starts`, by returning length-1 arrays if the list is empty. 

Fixes #2484

### DIFF
--- a/src/awkward/contents/listoffsetarray.py
+++ b/src/awkward/contents/listoffsetarray.py
@@ -175,11 +175,17 @@ class ListOffsetArray(Content):
 
     @property
     def starts(self) -> Index:
-        return self._offsets[:-1]
+        if self._offsets.length is unknown_length or self._offsets.length > 1:
+            return self._offsets[:-1]
+        else:
+            return self._offsets[:1]
 
     @property
-    def stops(self):
-        return self._offsets[1:]
+    def stops(self) -> Index:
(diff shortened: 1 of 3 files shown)


agoose77:
Wait, this is wrong (and the tests show it ... didn't run the full test suite locally as I'm working remotely). `ListArray.starts` has length zero for an empty list which makes sense!

jpivarski:
That's right; `offsets` is length-1 for an empty array of lists, but `starts` and `stops` are length-0 for an empty array of lists, regardless of whether we're talking about `ListOffsetArray` or `ListArray` (though we can't always make `offsets` for a `ListArray`).

PR #2488, opened 2023-06-02
fix: use offsets for starts to handle length-0 arrays

### LINKED ISSUE #2484: Reducers raise IndexError on masked three dimensional array

agoose77:
Unlike `offsets`, `starts` is not guaranteed to always have at least one element. This kind of bug has been dealt with elsewhere, but these are a few outstanding places.

Fixes #2484

### DIFF
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -949,10 +949,13 @@ class ByteMaskedArray(Content):
         if not branch and negaxis == depth:
             return out
         else:
-            if out.is_list:
-                out_content = out.content[out.offsets[0] :]
-            elif out.is_regular:
+            if out.is_regular:
                 out_content = out.content
+            elif out.is_list:
+                # The `outindex` that will index into `out_content` is 0-based, so we should ensure that we normalise
+                # the list content to start at the first offset.
+                # We know that we don't have any list type other than `ListOffsetArray`
+                out_content = out.content[out.offsets[0] :]
             else:
                 raise ValueError(
                     "reduce_next with unbranching depth > negaxis is only "
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -954,7 +954,6 @@ class ByteMaskedArray(Content):
             elif isinstance(out, ak.contents.ListOffsetArray):
                 # The `outindex` that will index into `out_content` is 0-based, so we should ensure that we normalise
                 # the list content to start at the first offset.
-                # We know that we don't have any list type other than `ListOffsetArray`
                 out_content = out.content[out.offsets[0] :]
             else:
                 raise ValueError(
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -950,7 +950,7 @@ class ByteMaskedArray(Content):
             return out
         else:
             if out.is_list:
-                out_content = out.content[out.starts[0] :]
+                out_content = out.content[out.offsets[0] :]
             elif out.is_regular:
                 out_content = out.content
             else:
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -949,9 +949,9 @@ class ByteMaskedArray(Content):
         if not branch and negaxis == depth:
             return out
         else:
-            if out.is_regular:
+            if isinstance(out, ak.contents.RegularArray):
                 out_content = out.content
-            elif out.is_list:
+            elif isinstance(out, ak.contents.ListOffsetArray):
                 # The `outindex` that will index into `out_content` is 0-based, so we should ensure that we normalise
                 # the list content to start at the first offset.
                 # We know that we don't have any list type other than `ListOffsetArray`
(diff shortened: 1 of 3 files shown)


jpivarski:
If all that we know about `out` is that it is a list array, it might not have an `offsets`. (`ListArray.offsets` does not exist). Also, if the array is empty, `offsets[0]` can be any value, possibly even one that doesn't exist in the `content`.

```python
>>> listoffsetarray = ak.Array([[1, 2, 3]])[0:0]
>>> listoffsetarray.layout.offsets.data[0] = 1000
>>> listoffsetarray.layout
<ListOffsetArray len='0'>
    <offsets><Index dtype='int64' len='1'>
        [1000]
    </Index></offsets>
    <content><NumpyArray dtype='int64' len='3'>[1 2 3]</NumpyArray></content>
</ListOffsetArray>
>>> listoffsetarray
<Array [] type='0 * var * int64'>
```

It's not a good thing to use in a slice of `content`.

At this point in the code, an empty array has to be considered as a separate case. In context, it looks like you'll want

```python
out.content[0:0]
```

for that case.

Previously, these had used `out.starts` because `starts` and `stops` exist for all three list array types. But even then, if list `i` within an array of lists is empty, then `starts[i] == stops[i]` but these values can be anything at all.

```python
>>> listarray = ak.Array([[1, 2, 3], [], [4, 5]])[[True, True, True]]
>>> listarray.layout.starts.data[1] = 1000
>>> listarray.layout.stops.data[1] = 1000
>>> listarray.layout
<ListArray len='3'>
    <starts><Index dtype='int64' len='3'>
        [   0 1000    3]
    </Index></starts>
    <stops><Index dtype='int64' len='3'>
        [   3 1000    5]
    </Index></stops>
    <content><NumpyArray dtype='int64' len='5'>[1 2 3 4 5]</NumpyArray></content>
</ListArray>
>>> listarray
<Array [[1, 2, 3], [], [4, 5]] type='3 * var * int64'>
```

So the old code wasn't right, either. Maybe this?

```python
starts, stops = out.starts.data, out.stops.data
nonempty = starts != stops
if nplike.count_nonzero(nonempty) == 0:
    return out.content[0:0]
else:
    return out.content[nplike.min(starts[nonempty]) : nplike.max(stops[nonempty])]
```

Or is the `out` at this point in the code constructed in such a way that we know enough about its structure to not have to do these defensive checks?

jpivarski:
Since `out = next._reduce_next(...)`, we know that it's not an out-of-order `ListArray`. The `nplike.min` and `nplike.max` are overkill.

We don't know what `axis` it is, so there's no guarantee that the first list is nonempty. (If `axis` is less than the level we're at, then I think it would always be true that each list has greater than or equal to the length of all that follow it. But this point in the code is not guaranteed to be in a case like that.)

```python
starts, stops = out.starts.data, out.stops.data
nonempty = starts != stops
nonempty_starts = starts[nonempty]
if len(nonempty_starts) == 0:
    return out.content[0:0]
else:
    return out.content[nonempty_starts : stops[nonempty]]
```

with appropriate handling of typetracer arrays.

Is there any reason why the original code only has a lower bound on the slice and not an upper bound? If there really shouldn't be an upper bound, I guess you don't need `stops[nonempty]`.

agoose77:
I should have added more explanation here, because it's definitely something that should have comments.

Whilst we're checking `is_list`, we also know that the result of `out` can be only `RegularArray` or `ListOffsetArray` types, i.e. not `ListArray`. So we can safely use `offsets` here. The reason that we _only_ trim the start is because the slice is done only to ensure that `outindex` (which starts at 0) aligns with the content.

As such, I don't *think* we need to change anything here, but I should add comments to elucidate this.

Wait, though, `is_list` should come _after_ `is_regular`, so there is a bug, although it doesn't actually have an effect, as `RegularArray` has an `offsets` too.

jpivarski:
`RegularArray` can _now_ have `offsets`. It's probably historical. Even though you could rely on `RegularArray.offsets` and have fewer if statements, those offsets have to be generated; it would be a waste just to check the first element.

Perhaps the if-predicates should be `isinstance(out, RegularArray)` and `isinstance(out, ListOffsetArray)`, rather than using the `is_regular` and `is_list` methods, since you want to make assumptions about the specific `Content` subclass that `out` is? You can put an `AssertionError` in the else clause.

What you're saying about `out` and `outindex` rings a bell...

agoose77:
> `RegularArray` can _now_ have `offsets`. It's probably historical. Even though you could rely on `RegularArray.offsets` and have fewer if statements, those offsets have to be generated; it would be a waste just to check the first element.

Yep — and to be clear, I mean to say that by providing this interface, the misordered `if` statements never caused a noticeable bug when reducing with regular dimensions! We definitely want to use both branches to maximise performance.

> Perhaps the if-predicates should be `isinstance(out, RegularArray)` and `isinstance(out, ListOffsetArray)`, rather than using the `is_regular` and `is_list` methods, since you want to make assumptions about the specific `Content` subclass that `out` is? You can put an `AssertionError` in the else clause.

I like being able to avoid the cross-content explicit type checks, but in this case I think that would indeed be clearer.

If you're happy with this, please merge!

jpivarski (approved):
Yes, great!

PR #2492, opened 2023-06-05
fix: starts handling in `RegularArray._reduce_next`

### LINKED ISSUE #2490: Positional reducers produce strange results for `axis != -1` on regular types
### Version of Awkward Array

main

```python
import awkward as ak
import numpy as np

x = np.arange(2 * 3).reshape(2, 3)

y = ak.from_numpy(x, regulararray=True)
y.show()

result = ak.argmin(y, axis=0, keepdims=True)
result.show()
```

agoose77:
This PR fixes #2490, which was caused by a bug that #1811 introduced.

To recap our reduction assumptions:
- `parents` *must* be _locally_ contiguous, e.g. `[ 0 0 2 2 1 1 ]` or `[ 0 0 1 1 2 2 ]`
- `parents` *may not* be _dense_, e.g. `[ 1 1 2 2 ]` or `[ 0 0 2 2 3 3 ]`

:scream: Both of these assumptions were incorrectly applied in #1811. I won't make any excuses, it shouldn't have happened! #1811 introduced code that assumed `parents` are both globally contiguous (`[ 1 1 2 2 ]`) and dense (`[ 0 0 1 1 2 2 ]`). Under those constraints, we can use `arange`, but in the more general case, we need to use a kernel to compute this.

This PR introduces a new explicit assumption:
- `starts` has exactly `len(starts) == max(parents)+1`
If this assumption fails, then we may over-allocate `starts`, but that would be a programming error.

This PR uses the `awkward_ListOffsetArray_reduce_nonlocal_nextstarts_64` kernel to compute `nextstarts`. It is likely possible to compute this more exactly, but for correctness and ease, I'm re-using the list kernel for this.

### DIFF
--- a/src/awkward/contents/regulararray.py
+++ b/src/awkward/contents/regulararray.py
@@ -1018,7 +1018,11 @@ class RegularArray(Content):
                 )
             )
             nextstarts = ak.index.Index64.empty(
-                starts.length * self._size, nplike=index_nplike
+                # `starts` must have at least enough elements for the largest `nextparent` to index into
+                # The upper bound for this value is given by `nextlen` (each item in this list belonging
+                # to a distinct reduction), but the length of `starts` should equate to `maxnextparents - 1`.
+                starts.length * self._size,
+                nplike=index_nplike,
             )
             assert (
                 nextstarts.nplike is index_nplike and nextparents.nplike is index_nplike
--- a/src/awkward/contents/regulararray.py
+++ b/src/awkward/contents/regulararray.py
@@ -1464,5 +1464,5 @@ class RegularArray(Content):
 
     def _is_equal_to(self, other, index_dtype, numpyarray):
         return self._size == other.size and self._content.is_equal_to(
-            self._content, index_dtype, numpyarray
+            other._content, index_dtype, numpyarray
         )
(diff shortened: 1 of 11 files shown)


src/awkward/contents/regulararray.py
```diff
@@ -1456,5 +1465,5 @@ def _to_backend(self, backend: Backend) -> Self:
 
     def _is_equal_to(self, other, index_dtype, numpyarray):
         return self._size == other.size and self._content.is_equal_to(
-            self._content, index_dtype, numpyarray
+            other._content, index_dtype, numpyarray
```
agoose77, line 1468: Unrelated bug (causing tests to always pass)!

jpivarski: I guess `self._content` is always equal to `self._content`. Good catch!

jpivarski (approved):
This looks good. At first, I was concerned about the

```python
len(starts) == max(parents) + 1
```

constraint, since computing the `max` can go awry if `len(parents) == 0` and has troubles with type-tracers, but that's not how it's actually implemented.

And that criterion is always true: the length of the conceptual/ideal `starts` is always greater than `max(parents) + 1`. You set the length of a `RegularArray`'s `content`'s `starts` to `len(starts) * size`, which is always correct (and may be bigger than `max(parents) + 1`).

(It is a weakness of the `parents` representation of jaggedness that it can't tell us about any empty lists at the end of an array.)

PR #2493, opened 2023-06-05
ci: include `kernel-specification.yml` in hash

agoose77:
Our CI wasn't including the kernel specification in the wheel hash key computation. This led to a cached wheel failing in #2492

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,7 +70,7 @@ jobs:
         uses: actions/cache@v3
         with:
           path: awkward-cpp/dist
-          key: ${{ github.job }}-${{ matrix.python-version }}-${{ matrix.python-architecture }}-${{ hashFiles('awkward-cpp/**') }}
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
Aha—that's what you meant. Yes, any change to kernel-specification.yml should force a recompilation.

What else is there?

  * .github/workflows (maybe changes there automatically trigger cache to be flushed)
  * dev/ because these scripts create some of the files that are used in compilation
  * header-only/ is used in the precompiled C++, not just JIT
  * .clang-format, .clang-tidy, .cmake-format.yml?
  * .gitmodules affects which pybind11 version we're using
  * cibuildwheel.toml?
  * noxfile.py for the same reasons as dev/

In principle, a change in any of the above could change awkward-cpp.

agoose77:
@jpivarski actually, I've confused myself. The whole point of running `pipx run nox -s prepare -- --headers --signatures --tests` as a primary step is to ensure that most of the "state" of the repo is contained in `awkward-cpp`; we run the `build` tool from _inside_ `awkward-cpp` so it makes sense that we don't depend upon external state.

Therefore, I'm not sure how we ran into that cache invalidation bug. I'll try out some ideas on a PR.

PR #2494, opened 2023-06-05
chore: upgrade CI micromamba action

agoose77:


jpivarski (approved):
Pinning to a version is better than `@main`.

This looks done: I'll merge it.

PR #2499, opened 2023-06-05
feat: add `ak.typetracer.length_one_if_typetracer`

### LINKED ISSUE #2496: request for `ak.typetracer.length_one_if_typetracer(array)`

When tracing ML algorithm evaluation meant for evaluation within dask_awkward it is often necessary to provide non-empty data to the ML algorithm to get a meaningful output array. A length-one array of zeros, like `form.length_one_array()`, is sufficient for this task and would significantly simplify code that needs to perform these kinds of evaluations. 

Example of where this would simplify library code: https://github.com/CoffeaTeam/coffea/pull/798/files#diff-f929094235cdc99a2b652540733611357165a0a3b99cb26aa74fea51907f8344R280-R341 

As with `ak.typetracer.empty_if_typetracer(array)`, when calling `ak.typetracer.length_one_if_typetracer(array)` the `array` argument should have its data recursively touched if `array` is a typetracer.

jpivarski:
Fixes #2496. Hopefully, the combination of `empty_if_typetracer` and `length_one_if_typetracer` implementations is not confusing.

Any thoughts on the name, like "`singleton_if_typetracer`"? I'm just wanting to ensure that the symmetry between the two functions is apparent, although "singleton" doesn't sound like a good way to describe an array with length 1.

### DIFF
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -5,14 +5,15 @@ import itertools
 import json
 import re
 from collections.abc import Mapping
+from functools import reduce
 
 import awkward as ak
 from awkward._backends.numpy import NumpyBackend
 from awkward._errors import deprecate
 from awkward._nplikes.numpylike import NumpyMetadata
-from awkward._nplikes.shape import unknown_length
+from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._parameters import parameters_union
-from awkward._typing import Final, JSONMapping, JSONSerializable
+from awkward._typing import Final, Iterator, JSONMapping, JSONSerializable
 
 np = NumpyMetadata.instance()
 numpy_backend = NumpyBackend.instance()
@@ -28,6 +29,15 @@ reserved_nominal_parameters: Final = frozenset(
         ("__array__", "categorical"),
     }
 )
+
+
+index_size_bytes: Final = {
+    "i8": 1,
+    "u8": 1,
+    "i32": 4,
+    "u32": 4,
+    "i64": 8,
+}
 
 
 def from_dict(input: dict) -> Form:
@@ -452,6 +462,9 @@ class Form:
             simplify=False,
         )
 
+    def _smallest_zero_buffer_lengths(self) -> Iterator[ShapeItem]:
+        raise NotImplementedError
+
     def length_one_array(self, *, backend=numpy_backend, highlevel=True, behavior=None):
         # A length-1 array will need at least N bytes, where N is the largest dtype (e.g. 256 bit complex)
         # Similarly, a length-1 array will need no more than 2*N bytes, as all contents need at most two
@@ -459,12 +472,19 @@ class Form:
         # create a buffer of this length (2N) and instruct all contents to use it (via `buffer_key=""`).
         # At the same time, with all index-like metadata set to 0, the list types will have zero lengths
         # whilst unions, indexed, and option types will contain a single value.
+        def max_prefer_unknown(this: ShapeItem, that: ShapeItem) -> ShapeItem:
+            if this is unknown_length:
+                return this
+            if that is unknown_length:
+                return that
+            return max(this, that)
+
+        buffer_length = reduce(max_prefer_unknown, self._smallest_zero_buffer_lengths())
+
         return ak.operations.ak_from_buffers._impl(
             form=self,
             length=1,
-            container={
-                "": b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
-            },
+            container={"": b"\x00" * buffer_length},
             buffer_key="",
             backend=backend,
             byteorder=ak._util.native_byteorder,
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -471,13 +471,90 @@ class Form:
                 return that
             return max(this, that)
 
+        container = {}
+
+        def prepare(form, multiplier):
+            form_key = f"node-{len(container)}"
+
+            if isinstance(form, (ak.forms.BitMaskedForm, ak.forms.ByteMaskedForm)):
+                if form.valid_when:
+                    container[form_key] = b"\x00" * multiplier
+                else:
+                    container[form_key] = b"\xff" * multiplier
+                return form.copy(form_key=form_key)  # DO NOT RECURSE
+
+            elif isinstance(form, ak.forms.IndexedOptionForm):
+                container[form_key] = b"\xff\xff\xff\xff\xff\xff\xff\xff"  # -1
+                return form.copy(form_key=form_key)  # DO NOT RECURSE
+
+            elif isinstance(form, ak.forms.EmptyForm):
+                # no error if protected by non-recursing node type
+                raise TypeError(
+                    "cannot generate a length_one_array from a Form with an "
+                    "unknowntype that cannot be hidden (EmptyForm not within "
+                    "BitMaskedForm, ByteMaskedForm, or IndexedOptionForm)"
+                )
+
+            elif isinstance(form, ak.forms.UnmaskedForm):
+                return form.copy(content=prepare(form.content, multiplier))
+
+            elif isinstance(form, (ak.forms.IndexedForm, ak.forms.ListForm)):
+                container[form_key] = b"\x00" * (8 * multiplier)
+                return form.copy(
+                    content=prepare(form.content, multiplier), form_key=form_key
+                )
+
+            elif isinstance(form, ak.forms.ListOffsetForm):
+                # offsets length == array length + 1
+                container[form_key] = b"\x00" * (8 * (multiplier + 1))
+                return form.copy(
+                    content=prepare(form.content, multiplier), form_key=form_key
+                )
+
+            elif isinstance(form, ak.forms.RegularForm):
+                size = form.size
+
+                # https://github.com/scikit-hep/awkward/pull/2499#discussion_r1220503454
+                if size is unknown_length:
+                    size = 1
+
+                return form.copy(content=prepare(form.content, multiplier * size))
+
+            elif isinstance(form, ak.forms.NumpyForm):
+                dtype = ak.types.numpytype.primitive_to_dtype(form._primitive)
+                size = multiplier * dtype.itemsize
+                for x in form.inner_shape:
+                    if x is not unknown_length:
+                        size *= x
+
+                container[form_key] = b"\x00" * size
+                return form.copy(form_key=form_key)
+
+            elif isinstance(form, ak.forms.RecordForm):
+                return form.copy(
+                    # recurse down all contents
+                    contents=[prepare(x, multiplier) for x in form.contents]
+                )
+
+            elif isinstance(form, ak.forms.UnionForm):
+                # both tags and index will get this buffer, but index is 8 bytes
+                container[form_key] = b"\x00" * (8 * multiplier)
+                return form.copy(
+                    # only recurse down contents[0] because all index == 0
+                    contents=(
+                        [prepare(form.contents[0], multiplier)] + form.contents[1:]
+                    ),
+                    form_key=form_key,
+                )
+
+            else:
+                raise AssertionError(f"not a Form: {form!r}")
+
         return ak.operations.ak_from_buffers._impl(
-            form=self,
+            form=prepare(self, 1),
             length=1,
-            container={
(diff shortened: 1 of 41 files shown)


agoose77:
@jpivarski I'm a bit reluctant to further expose `length_one_array`. I noticed an issue the other day, but had yet to get around to it. Your PR reminded me of the problem; there exists a subset of forms for which `length_one_array` is not able to return:

```pycon
>>> import awkward as ak
>>> import numpy as np
>>> buffer = ak.from_numpy(np.zeros((4, 5, 6)))
>>> result = buffer.layout.form.length_one_array(highlevel=False)
  File "/home/angus/.config/JetBrains/PyCharm2023.1/scratches/scratch_571.py", line 5, in <module>
    result = buffer.layout.form.length_one_array(highlevel=False)
  File "/home/angus/Git/awkward/src/awkward/forms/form.py", line 462, in length_one_array
    return ak.operations.ak_from_buffers._impl(
  File "/home/angus/Git/awkward/src/awkward/operations/ak_from_buffers.py", line 146, in _impl
    out = reconstitute(form, length, container, getkey, backend, byteorder, simplify)
  File "/home/angus/Git/awkward/src/awkward/operations/ak_from_buffers.py", line 190, in reconstitute
    data = _from_buffer(
  File "/home/angus/Git/awkward/src/awkward/operations/ak_from_buffers.py", line 171, in _from_buffer
    array = nplike.frombuffer(buffer, dtype=dtype, count=count)
  File "/home/angus/Git/awkward/src/awkward/_nplikes/array_module.py", line 60, in frombuffer
    return self._module.frombuffer(buffer, dtype=dtype, count=count)
ValueError: buffer is smaller than requested size

Process finished with exit code 1

```

i.e., `purelist_regular` arrays of >1 dimension.

Given that this is an intrinsic limitation, I wonder whether we should forgo exposing this to L2?

jpivarski:
I must have forgotten about that. Yes, a `RegularArray` can have any `size`, and the buffer that is provided has to allow for those sizes. If we have `RegularArray` inside `RegularArray` (with any nesting), those sizes multiply.

How big the buffer needs to be is knowable. I could add a `how_big` helper function to recursively walk over the Form and predict it. Alternatively, we can try to make the array, catch this exception, double the buffer size, and try again (in a loop). It will terminate after logarithmically many steps.

Nevertheless, I'd be happier with a `how_big` helper function. It shouldn't be too complicated, and working through it would be a reminder of whether there are any other issues beyond `RegularArray`. I can't think of any right now.

agoose77:
> How big the buffer needs to be is knowable. 

Yes, this is true. I can only think of regular arrays for now, and it should be fairly trivial to write. Let's move forward with that idea.

agoose77:
@jpivarski I think I prefer `length_one_if_typetracer`, but for completeness, what about `unary_if_typetracer`?

jpivarski:
`unary_if_typetracer` doesn't really help. So far, I think the best is still `length_one_if_typetracer`.

By the way, I intend to get to `how_big` later today, as well as your requests for review. They're all open tabs.

agoose77:
> By the way, I intend to get to how_big later today

Just to be clear — I made some commits in this PR that tackle that! Let me know if you want to take a different approach.

src/awkward/forms/numpyform.py
```diff
@@ -206,3 +207,10 @@ def _select_columns(self, index, specifier, matches, output):
 
     def _column_types(self):
         return (ak.types.numpytype.primitive_to_dtype(self._primitive),)
+
+    def _smallest_zero_buffer_lengths(self) -> Iterator[ShapeItem]:
+        if len(self.inner_shape) > 0:
+            raise NotImplementedError
```
jpivarski: Just like `ak.forms.regularform._smallest_zero_buffer_lengths`, this can multiply all of the `inner_shape` items except the last (unless I'm off-by-one).

src/awkward/forms/form.py
```diff
@@ -452,19 +462,29 @@ def length_zero_array(
...
+                return this
+            if that is unknown_length:
+                return that
+            return max(this, that)
+
+        buffer_length = reduce(max_prefer_unknown, self._smallest_zero_buffer_lengths())
+        if buffer_length is unknown_length:
+            raise NotImplementedError(
+                "cannot create length_zero_array from a form containing unknown shape items"
+            )
```
agoose77: This implementation of `_smallest_zero_buffer_lengths` is designed to avoid the call to `max` until the top level. This is a defensive approach; I'm not exactly clear on how this will tie in with typetracer / unknown shapes over time, so for now it's easy to error in a single place.

src/awkward/forms/numpyform.py
agoose77: Oops, that was a todo!

jpivarski:
Do you mean `_smallest_zero_buffer_lengths`? It looks good, although it seems like the name should say it's for a length-1 array. I don't understand the "`zero_buffer`" in the name. A length-0 array is easy to estimate: only `ListOffsetArray` is going to require a non-zero `Index`, and the largest that `Index` can be is 8 bytes. It's the length-1 case that's trickier.

This function only needs to return the maximum of a quantity that is never negative. It seems more straightforward to me to implement that as a reducer with identity `0` instead of an iterator. Then, when `ak.forms.regularform._smallest_zero_buffer_lengths` has to multiply what its content returned by its `size`, it only has to multiply one value; not an iterator of values. (Multiplication by a constant integer `size` is a homomorphism over maximization: $\mbox{size} \cdot \max(x_1, x_2) = \max(\mbox{size} \cdot x_1, \mbox{size} \cdot x_2)$. It's enough to multiply the winner of the maximization on `RegularForm.content` by `size`, rather than multiply each contender.)

agoose77:
> Do you mean _smallest_zero_buffer_lengths? It looks good, although it seems like the name should say it's for a length-1 array. I don't understand the "zero_buffer" in the name. A length-0 array is easy to estimate: only ListOffsetArray is going to require a non-zero Index, and the largest that Index can be is 8 bytes. It's the length-1 case that's trickier.

Yes, long day ;)

> This function only needs to return the maximum of a quantity that is never negativ

Yes - I touch on this inline, but essentially, I want to avoid the need to invoke `max` within the recursion, because it frees us up to not worry about `unknown_length` until the top-level

src/awkward/forms/form.py
jpivarski: That's a good way to write the `max` function in a typetracer-agnostic way. It's true that the max of `x` and unknown is unknown.

The intended use of this function is to turn a typetracer layout into a length-1 non-typetracer array. The typetracer might have `RegularArray.size is unknown_length`. That's the only way that an actual unknown length could come into this (the outermost length of the typetracer layout is usually unknown, but we'll be replacing that unknown length with 1.)

What kind of array would you want in that situation, @lgray? If an internal `RegularArray.size is unknown_length`, would you want an array with `RegularArray.size == 0` or maybe `1`? For the application, it might not matter, or you might want to pass a default down the recursion.

jpivarski:
Maybe call `max_prefer_unknown` within the recursion?

agoose77:
> max_prefer_unknown

Yes, I just didn't want to propagate this detail via an import & dispatch. But, it's all a bit unclear to me whether the iterator-approach is a solution in search of a problem. Right now with placeholder arrays, unknown lengths, etc, there are things to be figured out about how far into a layout `unknown_length` gets, both for typetracer and non-typetracer.

src/awkward/forms/form.py
lgray: For my particular use case I'm generally adding a batching dimension for ML evaluation, which is unknown. It will always be the outermost dimension.

lgray: But in general, length one for unknowns makes the most sense to me, given the name of the function and the intent.

agoose77: @jpivarski we now also have `unknown_length` in `known_data` contexts. It's being used for the re-hydration PR. I'm not very comfortable with that, but equally choosing an arbitrary "wrong" length is not the solution there either.

jpivarski: Answering

> For my particular use case I'm generally adding a batching dimension for ML evaluation, which is unknown. It will always be the outermost dimension.

Yeah, but suppose that batch should have regular shape `?? * 50 * 123 * ?? * 10`. The outer `??` is unknown because it's a batch—that's fine—but what do we do about any inner `??` that might exist? In a non-typetracer array, this will resolve to a known integer, but it's unknown in the Dask graph/typetracer pass because it was derived from a value of some other array. (Maybe it was ragged data that's getting padded by the longest list, which isn't known until you have concrete arrays.) When you call `length_one_if_typetracer` in this instance, can we insert a `0`, a `1`, or a default you pass down to fill in the inner `??`?

lgray: Passing a 1 makes the most sense IMO.

jpivarski: We can't let an `unknown_length` be used in a rehydrated (non-typetracer) array. But it might be appropriate to replace all of those inner `RegularArray.size is unknown_length` with `0` or `1`. Given what it's used for (rehydrated arrays are supposed to never be touched), it might not be wrong.

jpivarski:
Just to be clear, the size of the buffer only needs to be derived for `length_one_if_typetracer`.

For `empty_if_typetracer`, the buffer never needs to be larger than 8 bytes.

Buffer length for `length_one_if_typetracer` is a maximization over:

  * `if BitMaskedArray`: 1 and **do not recurse**
  * `if ByteMaskedArray`: 1 and **do not recurse**
  * `if EmptyArray`: raise an error because if you could recurse here, it's unprotected and `length_one_if_typetracer` will be trying to make a length-1 `EmptyArray`
  * `if IndexedArray`: `index.itemsize` and recurse into `self.content`
  * `if IndexedOptionArray`: `index.itemsize` and **do not recurse**
  * `if ListArray`: `starts.itemsize` and recurse into `self.content`
  * `if ListOffsetArray`: `2 * offsets.itemsize` and recurse into `self.content`
  * `if NumpyArray`: `self.itemsize * reduce(operator.mul, self.inner_shape)`
  * `if RecordArray`: maximum of recursing into `self.contents`
  * `if RegularArray`: `self.size` times the result of recursing into `self.content`
  * `if UnionArray`: `index.itemsize` and the result of recursing into `self.contents[0]` (because the only tag will be `0`; `EmptyArray` is never legal in a `UnionArray`)
  * `if UnmaskedArray`: recurse into `content`

For `length_one_if_typetracer`, the `container` passed to `from_buffers` will need to arrange for the `IndexedOptionArrays` to get `-1` instead of `0` (i.e. almost all buffers get `b"\x00\x00\x00\x00\x00\x00\x00\x00"` but `IndexedOptionArrays` get `b"\xff\xff\xff\xff\xff\xff\xff\xff"`). This is because, if you have an `EmptyArray`, it is very likely inside of an option-type array of all `None`. Similarly, `BitMaskedArray` and `ByteMaskedArray` get `b"\x00"` if `valid_when is True` and `b"\xff"` if `valid_when is False`, so that they produce `None` as well.

Buffers can be longer than they need to be (i.e. `b"\xff\xff\xff\xff\xff\xff\xff\xff"` is just as good as `b"\xff"`).

It would probably be easiest to make the recursive function that finds the buffer size also make the buffers, so that it can give `BitMaskedArrays`, `ByteMaskedArrays`, and `IndexedOptionArrays` the appropriate choice of zeros or ones.

src/awkward/forms/form.py
agoose77: The main discrepancy with rehydration is that it doesn't track `shape_touched`, so we might depend upon the shape. A common example would be broadcasting `NumpyArray`; these don't need to be touched, but they must have compatible shapes. After rehydration, any arbitrary length we choose won't be guaranteed to agree with the other layout.

jpivarski: A length-1 regular dimension can agree with any broadcasting. (Fortunately!)

lgray:
On names: perhaps, given the `empty` vs. `length_one`. Singleton doesn't make sense on reading. What about `length_zero_if_typetracer` and `length_one_if_typetracer`? To me, at least, it is then crystal clear what's going on.

Can slap a deprecation on `empty_if_typetracer` and change downstream code.

jpivarski:
> Can slap a deprecation on `empty_if_typetracer` and change downstream code.

This is a good suggestion. Then `length_zero_if_typetracer` and `length_one_if_typetracer` are symmetric. The `empty_if_typetracer` function can raise a warning and then pass to `length_zero_if_typetracer`.

@agoose77, should the name-change be in this PR?

agoose77:
> @agoose77, should the name-change be in this PR?

Yep, that seems like the right decision

jpivarski:
@agoose77 I've added the name change and deprecation warning. It's also in the [wiki](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20).

PR #2502, opened 2023-06-06
fix: request non-option results for overloaded positional record reducers

### LINKED ISSUE #2501: Positional record-reducers cannot return masked results
### Version of Awkward Array

main

We provide a `mask` argument that indicates to custom reducers whether a mask should be returned, for performance. For positional reducers, the proper thing would be to set `mask=False` always; we have to do more work for masked results in order to rebuild the index values correctly, and whilst we can use e.g. `fill_none` on the result, the `mask` argument becomes meaningless in such a scenario. Instead, we can set `mask` to `False` for positional reducers, meaning that the user should always return non-option types.

agoose77:
The rationale in #2458 for the `mask` argument is to avoid the unnecessary copying involved in constructing identity elements for reductions that are ultimately masked. There is no similar motivation for positional reducers. Unlike value reducers, the results of positional reductions are passed into a set of kernels to adjust their shifts / starts. Even if the final result is masked, these kernels must operate upon a contiguous buffer of index values. As such, it would be preferable if the library author _did not_ return masked arrays in such cases.

This PR closes #2501 by asking user reducer implementations to return non-masked arrays in such cases, and raises an error if they do not honour the `mask` argument.

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -929,9 +929,27 @@ class RecordArray(Content):
                 )
             )
 
+            # Positional reducers ultimately need to do more work when rebuilding the result
+            # so asking for a mask doesn't help us!
+            reducer_should_mask = mask and not reducer.needs_position
             out = _apply_record_reducer(
-                reducer_recordclass, self, mask, outoffsets, behavior
+                reducer_recordclass,
+                self,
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Okay, that makes sense.

Issue #2468, opened 2023-05-23
Support for units


HealthyPear:
### Description of new feature

(Apologies if this is already supported in some way I didn't find it in the docs or in the suggested links when opening the issue)

I am working with a data format in which each event is a row in a table and it's a ragged array.

For context, this is how each event is written to file:
- some columns relate to the general properties of the event
- 2 of these columns, say "A" and "B" are integers
- a second set of columns pertaining to properties related to A
- a third set of columns pertain to properties related to B

As you can guess:
- columns of A contain arrays long how much A says and the same for B
- each event is independent, so a column "A_bla" becomes automatically a ragged array
- most columns describe physical quantities which have units stored in the header of the file

I am currently opening this file as a dictionary of numpy arrays and building 3 [astropy QTables](https://docs.astropy.org/en/stable/table/): "event_table", "A_table", "B_table". Of course from there one can play around with masking, filtering, etc...so my (immediate) use case is kind of fixed.

Said this, I cannot stop thinking that one of these files is basically an " awkward table"!

At the same time:
- astropy devs say that support for ragged arrays is a numpy problem (see e.g. astropy/astropy#14014)
- numpy devs have no intention to support this use-case (see e.g. numpy/numpy#22603 - which is kind of the point of why awkward array exists!)

Soooo....what about adding units, maybe compatible with [astropy quantities](https://docs.astropy.org/en/stable/api/astropy.units.Quantity.html#astropy.units.Quantity) or with a similar object? 😄

jpivarski:
This sounds like a good use-case for #1391. That would propagate units, though it would not convert them.

Or maybe it could be some custom behaviors? https://awkward-array.org/doc/main/reference/ak.behavior.html That would even make it possible to convert appropriately. (The `NumpyArray` nodes would have two parameters, `__array__: "units"` and `__units__: "light year"` and there would be a custom `ak.Array` subclass with some mathematical operations overloaded, such as `np.add`—maybe only addition.)

I can try to write a prototype. This may be complex enough that we should build it in, rather than making users implement it—so the new feature would be a new set of behaviors in src/awkward/behaviors. It would be the only `__array__` behaviors that are not string-like, and @agoose77 and I were trying to think if there would be anything like that.

agoose77:
@gpiert thanks for opening #2788 regarding this! 

It's planned that we add support for units through `pint`. We're working our way through features to get there :)

jpivarski:
[...]

PR #2503, opened 2023-06-06
feat!: deprecate `to_NumpyForm`'s ``dtype` argument

agoose77:
Whilst refactoring I noticed that we accept a `dtype` argument in `EmptyForm`. I propose we deprecate this form, in favour of `primitive`. This is not a big change, but I think it would be sensible.

### DIFF
--- a/src/awkward/_connect/cling.py
+++ b/src/awkward/_connect/cling.py
@@ -446,7 +446,7 @@ namespace awkward {{
 
 def togenerator(form, flatlist_as_rvec):
     if isinstance(form, ak.forms.EmptyForm):
-        return togenerator(form.to_NumpyForm(np.dtype(np.float64)), flatlist_as_rvec)
+        return togenerator(form.to_NumpyForm(primitive="float64"), flatlist_as_rvec)
 
     elif isinstance(form, ak.forms.NumpyForm):
         if len(form.inner_shape) == 0:
--- a/src/awkward/_connect/numba/arrayview.py
+++ b/src/awkward/_connect/numba/arrayview.py
@@ -106,7 +106,7 @@ def code_to_function(code, function_name, externals=None, debug=False):
 
 def to_numbatype(form):
     if isinstance(form, ak.forms.EmptyForm):
-        return to_numbatype(form.to_NumpyForm(np.dtype(np.float64)))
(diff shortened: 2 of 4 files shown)


jpivarski (approved):
In fact, the method with primitives preserves more dtype information than the method with dtypes, oddly enough.

```python
>>> f1 = ak.forms.numpyform.from_dtype(np.dtype("timedelta64[15s]"))
>>> f1
NumpyForm('timedelta64', parameters={'__unit__': '15s'})
>>> f1.length_one_array()
<Array [0 generic time units] type='1 * timedelta64[unit="15s"]'>
>>> ak.to_numpy(f1.length_one_array())
array([0], dtype=timedelta64)
```

versus

```python
>>> f2 = ak.forms.numpyform.NumpyForm("timedelta64[15s]")
>>> f2
NumpyForm('timedelta64[15s]')
>>> f2.length_one_array()
<Array [0 seconds] type='1 * timedelta64[15s]'>
>>> ak.to_numpy(f2.length_one_array())
array([0], dtype='timedelta64[15s]')
```

So this looks like a good direction to go in. The `from_dtype` function moving the time units into a `__units__` parameter is counterproductive and it will interfere with plans to put general units into a `__units__` parameter, so this should be changed.

To deprecate: add a `time_units_as_parameter: bool` to `from_dtype`, which raises a deprecation warning and becomes `True` if unspecified during the deprecation period, and after the deprecation period the default becomes `False`.

agoose77:
Following from @jpivarski's observation above, I realised that I hadn't noticed our parameter-based datetime handling. We had a discussion on Slack, and I think we know that `__units__` is likely the name that #2468 will use for registry-based units. This is distinct datetime units, which are a different subset.

I'm interested in the idea that we could change the primitive to be "datetime64", and add a special "__datetime_unit__" parameter to store this state. `NumpyArray` would enforce the presence of such a parameter for the given dtype.

The benefits of this suggestion are that we can simplify our primitive strings, and avoid the need to parse the "unit" information, by placing it in a structural location (parameters)

This PR is on hold until we firmly resolve what to do with the two implementations of time-unit handling.

agoose77:
@jpivarski I'm planning upon deprecating the time-unit logic in a separate PR. Is this good to go?

EDIT: you approved already, so I'm happy to merge.

PR #2505, opened 2023-06-07
refactor: improve kernel error handling logic (1 of 3)

agoose77:
For a while I've been intending upon refactoring the error handling logic for kernels, so that it's not associated with `Content`.

This PR does two things:
- Moves `_selfless_handle_error` from `Content` to `Backend.maybe_kernel_error`
- Simplifies `_handle_error` and renames to `Content._maybe_index_error` (name TBD)

### DIFF
--- a/src/awkward/_backends/backend.py
+++ b/src/awkward/_backends/backend.py
@@ -39,9 +39,11 @@ class Backend(Singleton, ABC):
         reducer: ak._reducers.Reducer,
         layout: ak.contents.NumpyArray,
         parents: ak.index.Index,
+        starts: ak.index.Index,
+        shifts: ak.index.Index | None,
         outlength: int,
     ) -> ak.contents.NumpyArray:
-        return reducer.apply(layout, parents, outlength)
+        return reducer.apply(layout, parents, starts, shifts, outlength)
 
     def apply_ufunc(self, ufunc, method, args, kwargs):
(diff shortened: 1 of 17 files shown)


src/awkward/_backends/backend.py
```diff
@@ -34,14 +35,35 @@ def index_nplike(self) -> NumpyLike:
...
+        if error.filename is None:
+            filename = ""
+        else:
+            filename = " (in compiled code: " + error.filename.decode(
+                errors="surrogateescape"
+            ).lstrip("\n").lstrip("(")
+
+        message = error.str.decode(errors="surrogateescape")
+
+        assert not error.pass_through
```
agoose77, line 63: See #2504 for why this assertion holds

jpivarski: Yes, it's dead code. The kernels part of Awkward had to be valid for both v1 and v2 at the same time. The error-reporting record is now fairly out of date, but we'd have to change it carefully, if at all. I don't remember whether the Numba implementation depends on that exact struct or not.

jpivarski (approved):
This is a much-needed refactoring. The kernel-error handling was getting pretty crusty.

Although it touches a lot of files, most of them are just swapping names (and making the error-handling call "selfless"). It looks straightforward enough to me that I can just merge it.

PR #2506, opened 2023-06-07
refactor: move JAX reducer dispatch to `ak._do.reduce` (2 of 3)

agoose77:
> **Warning**
> Depends upon #2505 

This PR rewrites the JAX reducer dispatch to happen _before_ reduction takes place. This `prepare_` design is also applied to ufuncs, and permits us to use a reference implementation e.g. `ak._reducers.Min` as both an implementation _and_ token that can be replaced by different backends.

There are no bugfixes here, just nicer-to-maintain.

### DIFF
--- a/src/awkward/_backends/backend.py
+++ b/src/awkward/_backends/backend.py
@@ -39,9 +39,11 @@ class Backend(Singleton, ABC):
         reducer: ak._reducers.Reducer,
         layout: ak.contents.NumpyArray,
         parents: ak.index.Index,
+        starts: ak.index.Index,
+        shifts: ak.index.Index | None,
         outlength: int,
     ) -> ak.contents.NumpyArray:
-        return reducer.apply(layout, parents, outlength)
+        return reducer.apply(layout, parents, starts, shifts, outlength)
 
     def apply_ufunc(self, ufunc, method, args, kwargs):
         return getattr(ufunc, method)(*args, **kwargs)
--- a/src/awkward/_backends/backend.py
+++ b/src/awkward/_backends/backend.py
@@ -18,7 +18,6 @@ T = TypeVar("T", covariant=True)
 KernelKeyType: TypeAlias = Tuple[str, Unpack[Tuple[np.dtype, ...]]]
 KernelType: TypeAlias = "Callable[..., KernelError | None]"
 
-
 class UfuncLike(Protocol):
     def __call__(self, *args: ArrayLike, **kwargs) -> ArrayLike:
         ...
--- a/src/awkward/_backends/backend.py
+++ b/src/awkward/_backends/backend.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 from abc import ABC, abstractmethod
 
 import awkward as ak
+from awkward._kernels import KernelError
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyLike, NumpyMetadata
 from awkward._singleton import Singleton
@@ -15,7 +16,7 @@ numpy = Numpy.instance()
 
 T = TypeVar("T", covariant=True)
 KernelKeyType: TypeAlias = Tuple[str, Unpack[Tuple[np.dtype, ...]]]
-KernelType: TypeAlias = Callable[..., None]
+KernelType: TypeAlias = Callable[..., KernelError | None]
 
 
 class Backend(Singleton, ABC):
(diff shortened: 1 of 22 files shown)


src/awkward/_backends/backend.py
```diff
@@ -19,6 +19,11 @@
 KernelType: TypeAlias = "Callable[..., KernelError | None]"
 
 
+class UfuncLike(Protocol):
+    def __call__(self, *args: ArrayLike, **kwargs) -> ArrayLike:
+        ...
```
agoose77, line 24: NumPy ufuncs provide a suite of methods / features that we don't currently support across all backends. I think at this stage, we have no internal use for these other ufuncs and we don't do any packing to ensure that ufuncs only touch "real" data. Therefore, we should only support `ufunc.__call__`; most other things should perhaps be dedicated kernels.

src/awkward/_backends/typetracer.py
```diff
@@ -43,18 +43,3 @@ def _coerce_ufunc_argument(self, x):
...
-        for x in args:
-            if isinstance(x, TypeTracerArray):
-                x.touch_data()
-                shape = x.shape
-
-            numpy_args.append(self._coerce_ufunc_argument(x))
-
-        assert shape is not None
-        tmp = getattr(ufunc, method)(*numpy_args, **kwargs)
-        return self._typetracer.empty((shape[0],) + tmp.shape[1:], dtype=tmp.dtype)
```
agoose77, line 60: We already implement ufunc support for typetracers via NEP-13, so removing this avoids a distinction between the ufunc-via-NumpyArray and ufunc-via-typetracer

src/awkward/_connect/jax/reducers.py
```diff
@@ -1,74 +1,153 @@
...
+    def _return_dtype(cls, given_dtype):
+        return np.int64
+
+    def apply(
+        self,
+        array: ak.contents.NumpyArray,
+        parents: ak.index.Index,
+        outlength: ShapeItem,
+    ) -> ak.contents.NumpyArray:
         raise RuntimeError("Cannot differentiate through argmin")
```
agoose77, line 59: I wanted to be more explicit about the mapping between reducers; this is slightly more type-based than name based.

src/awkward/_nplikes/typetracer.py
```diff
@@ -506,6 +506,8 @@ def _apply_ufunc(self, ufunc, *inputs):
             assert not isinstance(x, PlaceholderArray)
             try_touch_data(x)
 
+        inputs = [x.content if isinstance(x, MaybeNone) else x for x in inputs]
+
```
agoose77, line 510: In general, we want `MaybeNone` to erase in contexts where a `None` would just error. We probably need to do more here, but this seems like a good start.

This was motivated by a failing test.

src/awkward/_reducers.py
```diff
@@ -34,8 +34,17 @@ def preferred_dtype(self) -> DTypeLike:
...
+        self,
+        array: ak.contents.NumpyArray,
+        parents: ak.index.Index,
+        outlength: ShapeItem,
+    ) -> ak.contents.NumpyArray:
+        raise NotImplementedError
+
     @classmethod
-    def return_dtype(cls, given_dtype: DTypeLike) -> DTypeLike:
+    def _return_dtype(cls, given_dtype: DTypeLike) -> DTypeLike:
```
agoose77, line 38: These underscored methods are internal-only, so they should be private.

src/awkward/_backends/backend.py
jpivarski: It's true that we only ever use `ufunc.__call__`. Awkward v0 used `ufunc.reduceat` to implement `axis=-1` reducers, but none of that code survives.

I just grepped the codebase for all the other ufunc methods. We _only_ use `__call__`, even in NEP-13:

https://github.com/scikit-hep/awkward/blob/33750f362b0398b4d30936a8572cf1c77faa1edf/src/awkward/_connect/numpy.py#L213-L215

src/awkward/_backends/typetracer.py
jpivarski: Okay, I had missed that duplication.

src/awkward/_reducers.py
jpivarski: That's true. If we were exposing the `Reducer` classes at all, the `return_dtype` is something one might want to know about it, but they're in a hidden module.

jpivarski: Actually, it's because they're in a hidden module that I didn't bother to hide the methods. There only needs to be one underscore between `ak.` and object. Maybe the distinction you're making here is between L3 and L4?

jpivarski (approved):
Most of what I see here are type annotations, protocols, and some hardening (basing decisions on types, rather than names).

I don't think it's necessary to hide a method in a hidden class (`ak._reducers.Reducer._return_dtype` only needed one leading underscore), but it makes sense if it's a distinction between L3 and L4.

agoose77:
> but it makes sense if it's a distinction between L3 and L4.

exactly that!

PR #2507, opened 2023-06-07
refactor: move responsibility for positional reducers to `Reducer.apply`

agoose77:
This PR moves the responsibility for "fixing" positional reducers into `_reducers`, such that reducers are self-contained. The same is not done for `RecordArray`, because this is effectively a separate reduction implementation that happens to use the same kernels.

### DIFF
--- a/src/awkward/_reducers.py
+++ b/src/awkward/_reducers.py
@@ -39,6 +39,8 @@ class Reducer(Protocol):
         self,
         array: ak.contents.NumpyArray,
         parents: ak.index.Index,
+        starts: ak.index.Index,
+        shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
         ...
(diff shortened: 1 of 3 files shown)


src/awkward/_reducers.py
```diff
@@ -68,6 +70,53 @@ def _maybe_other_type(cls, dtype: DTypeLike) -> DTypeLike:
         return type
 
 
+def apply_positional_corrections(
+    reduced: ak.contents.NumpyArray,
+    parents: ak.index.Index,
+    starts: ak.index.Index,
+    shifts: ak.index.Index | None,
+):
```
agoose77, line 84: This is a free function rather than a method because I want to avoid a complex inheritance heirarchy that the JAX reducers need to mimic.

jpivarski (approved):
This makes sense. I had felt that `_reduce_next` was a reasonable place to put reducer-stuff, and it's only in `NumpyArray._reduce_next`, but `ak._reducers` is also a reasonable place and this removes noise from numpyarray.py, which deals with many non-reducer things as well.

PR #2509, opened 2023-06-07
refactor: simplify reducer boilerplate (3 of 3)

agoose77:
I've intended upon refactoring the reducers in `ak._reducers` for a while. This pass should preserve clarity and readability.


jpivarski (approved):
I reviewed this by merging the changes, which forced me to think about which methods are going away and which are new. Breaking it down into a protocol and implementation is a clean, formal way of specifying it.

PR #2518, opened 2023-06-09
feat!: deprecate the time `__unit__` parameter

agoose77:
In future we anticipate the introduction of a units system that uses this parameter. Moreover, datetimes should use the `primitive` to represent these types, as it's a fundamental type.


src/awkward/forms/numpyform.py
```diff
@@ -11,14 +12,26 @@
...
         inner_shape = dtype.shape
         dtype = dtype.subdtype[0]
 
-    if issubclass(dtype.type, (np.datetime64, np.timedelta64)):
+    if time_units_as_parameter is UNSET:
+        time_units_as_parameter = True
+
+    if time_units_as_parameter:
+        deprecate(
+            "from_dtype conversion of datetime units to generic `datetime` types is deprecated, "
```
jpivarski:
```suggestion
            "from_dtype conversion of temporal units to generic `datetime64` and `timedelta64` types is deprecated, "
```

Just to be a little clearer.

jpivarski (approved):
This looks great!

All of the tests that assert the `DeprecationWarnings` will have to be changed later, but it will be obvious when the time comes.

PR #2519, opened 2023-06-09
fix: str of `KeyError` for <3.11

### LINKED ISSUE #2513: Error formatting is broken (an error in the error handling)
### Version of Awkward Array

HEAD

I have a real error and should be getting a properly formatted error message, but there's an error in the error-handling.

To reproduce it:

```python
import awkward as ak
f = ak.Array([[1, 2, 3], [], [4, 5]]).layout.form
ak.from_buffers(f, 0, {"": b"\x00\x00\x00\x00\x00\x00\x00\x00"}, buffer_key="{form_key}")
```

The error message is

```
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_from_buffers.py", line 89, in from_buffers
    return _impl(
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_from_buffers.py", line 146, in _impl
    out = reconstitute(form, length, container, getkey, backend, byteorder, simplify)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_from_buffers.py", line 349, in reconstitute
    raw_array = container[getkey(form, "offsets")]
KeyError: 'None'

  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_from_buffers.py", line 89, in from_buffers
    return _impl(
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 56, in __exit__
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 71, in handle_exception
KeyError: "'None'\n\nThis error occurred while calling\n\n    ak.from_buffers(\n        form = ListOffsetForm-instance\n        length = 0\n        container = {'': b'\\x00\\x00\\x00\\x00\\x00\\x00\\x00\\x00'}\n        buffer_key = '{form_key}'\n        backend = 'cpu'\n        byteorder = '<'\n        highlevel = True\n        behavior = None\n    )"
```

The `KeyError: 'None'` is the actual error, and it was supposed to be decorated like this:

```
KeyError: 'None'

    ak.from_buffers(
        form = ListOffsetForm-instance
        length = 0
        container = {'': b'\x00\x00\x00\x00\x00\x00\x00\x00'}
        buffer_key = '{form_key}'
        backend = 'cpu'
        byteorder = '<'
        highlevel = True
        behavior = None
```

The line number didn't shift between my copy of Awkward and `main`:

https://github.com/scikit-hep/awkward/blob/be876a0ae3d78de29a33c635cdcc75315c4b1740/src/awkward/_errors.py#L71

I'm in the second branch because I'm not using Python 3.11 (with its exception decorators).

```python
>>> sys.version_info
sys.version_info(major=3, minor=9, micro=15, releaselevel='final', serial=0)
```

agoose77:
This fixes #2513

### DIFF
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -37,10 +37,8 @@ class AwkwardKeyError(builtins.KeyError):
         return super(Exception, self).__str__()
 
 
-# Pretend to be builtins!
 AwkwardKeyError.__name__ = "KeyError"
 AwkwardKeyError.__qualname__ = "KeyError"
-AwkwardKeyError.__module__ = "builtins"
 
 
 class ErrorContext:
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -1,6 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 from __future__ import annotations
 
+import builtins
 import sys
 import threading
 import warnings
@@ -29,6 +30,17 @@ class PartialFunction:
 
     def __call__(self):
         return self.func(*self.args, **self.kwargs)
+
+
+class AwkwardKeyError(builtins.KeyError):
+    def __str__(self):
+        return super(Exception, self).__str__()
+
+
+# Pretend to be builtins!
+AwkwardKeyError.__name__ = "KeyError"
+AwkwardKeyError.__qualname__ = "KeyError"
+AwkwardKeyError.__module__ = "builtins"
 
 
 class ErrorContext:
@@ -86,6 +98,9 @@ class ErrorContext:
                     str(exception)
                     + "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
                 )
+                new_exception.__cause__ = exception
+            elif issubclass(cls, KeyError):
+                new_exception = AwkwardKeyError(self.format_exception(exception))
                 new_exception.__cause__ = exception
             else:
                 new_exception = cls(self.format_exception(exception))
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -32,13 +32,9 @@ class PartialFunction:
         return self.func(*self.args, **self.kwargs)
 
 
-class AwkwardKeyError(builtins.KeyError):
+class KeyError(builtins.KeyError):
     def __str__(self):
         return super(Exception, self).__str__()
-
-
-AwkwardKeyError.__name__ = "KeyError"
-AwkwardKeyError.__qualname__ = "KeyError"
 
 
 class ErrorContext:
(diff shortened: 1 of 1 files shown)


src/awkward/_errors.py
```diff
@@ -31,6 +32,17 @@ def __call__(self):
...
 
+class AwkwardKeyError(builtins.KeyError):
+    def __str__(self):
+        return super(Exception, self).__str__()
+
+
+# Pretend to be builtins!
+AwkwardKeyError.__name__ = "KeyError"
+AwkwardKeyError.__qualname__ = "KeyError"
+AwkwardKeyError.__module__ = "builtins"
```
jpivarski: I'm wary of this. It could lead to someone being very confused that this type `is not KeyError` for some reason.

I was thinking that we would name the class `ak._errors.KeyError`, so it displays its name as `KeyError`, but it's distinguishable by its `__module__`. Setting `__module__ = "builtins"` sounds like we're setting up for a disaster.

jpivarski (changes requested):
I'm uncomfortable about our class pretending to _be_ the `builtins.KeyError`. It can be named `KeyError`, but let's make it distinguishable by `__module__`, so that when someone is digging into a bug, we're not making the solution harder for them.

(I'm assuming that experts who are trying to resolve bugs would know that classes can have the same name in different modules.)

src/awkward/_errors.py
agoose77: That's fair. In general, it's not good practice to use 'is' like this, but I think you make a reasonable argument. Let's rework this in line with making our error distinguishable from the built-in.

jpivarski (approved):

jpivarski:
I made a counter-edit to indicate what I meant. If we're saying its name is `KeyError`, then we should let its name be `KeyError`; module encapsulation is for making these sorts of same-name-but-distinct references.

It means that we can never use unqualified `KeyError` in the text of this file—it always has to be `builtins.KeyError`—but that's normal. Uproot's reading.py has to do that with the function named `open`, and ak_zip.py has to do that with the function named `zip`.

src/awkward/_errors.py
```diff
@@ -97,8 +93,8 @@ def decorate_exception(self, cls: type[E], exception: E) -> E:
                     + "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
                 )
                 new_exception.__cause__ = exception
-            elif issubclass(cls, KeyError):
-                new_exception = AwkwardKeyError(self.format_exception(exception))
+            elif issubclass(cls, builtins.KeyError):
+                new_exception = KeyError(self.format_exception(exception))
```
agoose77, line 97: I prefer not to shadow the `KeyError` builtin, as it's not something one usually overrides in a global namespace. Our `AwkwardKeyError` isn't something I anticipated even L3 awkward modules importing, so changing `__name__` didn't both me. With some time having elapsed, I have changed my mind on that, so I'm now in favour of this change.

PR #2520, opened 2023-06-09
fix(do not merge / example): do not touch data in complex slices

lgray:
It turns out that removing these two lines fixes the overtouching problem in the example complex analysis attached below (where you can run `dask.compute(output)` and have it succeed using prior versions of dask_awkward/awkward with far fewer read columns).

With the this PR applied the `necessary_columns` returns to previous the previous list.

The rest of https://github.com/scikit-hep/awkward/commit/19f62aaac1f77c8d8096730f56dead782260d400 seems to be perfectly fine.

@douglasdavis @martindurant @jpivarski @agoose77

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -382,7 +382,6 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
                     or is_unknown_integer(item)
                     or is_unknown_array(item)
                 ):
-
                     if is_unknown_scalar(item):
                         item = self.nplike.promote_scalar(item)
 
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -382,8 +382,6 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
                     or is_unknown_integer(item)
                     or is_unknown_array(item)
                 ):
-                    try_touch_data(item)
-                    try_touch_data(self)
 
                     if is_unknown_scalar(item):
                         item = self.nplike.promote_scalar(item)


agoose77:
It's late here, but at face value, we should touch data in a complex slice - this would fail at runtime if these data were bogus.

It might be that we end up in this path unnecessarily, or that it represents a case that would actually require touching data.

The reproducer you showed looks like it should touch data, because it assigns in place, which would broadcast.

lgray:
No worries on it being late, it's also Friday.

I can confirm that the built and optimized task graph with this change applied runs all the way through and has touched all and only the correct data. 

~7000 layer task graph, ~10k events. 

Without this patch it tries to load about 2-3x more columns than are needed. 

The reproducer shouldn't require loading "MET_pt" to calculate only "Muon_pt".

Conclusion - something is fishy with these lines. And then fishier still since "MET_pt" doesn't get optimized away even with this patch in place for the repro.

lgray:
After sleeping on it and then thinking more - this is a separate issue from the dask_awkward one.

- the complex example has been running fine since April, touching exactly the correct columns
- these two lines drastically alter the touching behavior, aggressively expanding the number of columns touched 
  - when we already know the set of columns pre awkward 2.2.2 was enough
  
 The dask_awkward issue as more to do `dak.necessary_columns(uproot_dask_array.field)` being somehow affected by `uproot_dask_array["newfield"] = uproot_dask_array.unrelated_field` where it is obvious that you need not consider touching `unrelated_field` to `dask.compute(uproot_dask_array.field)`.

agoose77:
I've looked at the example with a fresh pair of eyes, and I believe that this "bug" is deliberate. To be clear, this is not the intended end-user experience, and I'll expand upon this point. What you're seeing in dask-contrib/dask-awkward#280 is an example of https://github.com/dask-contrib/dask-awkward/pull/184#issuecomment-1573437965.


By definition, `with_field` _should_ touch all the offset buffers down to the leaves, because it uses broadcasting. This means that `Array.__setitem__` can actually change the structure of the container array:
```pycon
>>> array = ak.to_regular(ak.zip({'x':[[1],[2],[3],[4]]}))
>>> array
<Array [[{x: 1}], [{x: 2}], [{...}], [{x: 4}]] type='4 * 1 * {x: int64}'>
>>> array['y'] = [[1,2,3],[4,5,6],[7], []]
>>> array
<Array [[{x: 1, y: 1}, {...}, {...}], ..., []] type='4 * var * {x: int64, y...'>
```

I asked on Slack whether you were using nanoevents. In fact, I should have been more precise and asked "Are you using nanoevents _remapping_?". I would expect nano-events arrays that _do not_ have the nanoevents remapping to have this problem. Whilst we know that, due to the way that the data were serialised, `MET_xxx` and `Muon_xxx` have the same offsets, at read-time they are considered different arrays with distinct `form-keys`. As such, we must touch both of the arrays (via `column.to_list`) to ensure that the offsets are included. The benefit of reading these as remapped nano-events is that the `form_keys` can be such that the `MET` and `Muon` are sub-branches of a root node with common offsets. That's why scikit-hep/uproot5#901 is important; it allows us to encode this additional information into the structure.

This is effectively a limitation of how column optimisation is done in dask-awkward; it's an optimisation over structure, rather than low-level buffers. This is not a bad thing, just a statement of how things work!

lgray:
Here's the more complex test I am talking about https://github.com/dask-contrib/dask-awkward/files/11710067/complex_repro.zip. Try this with and without patch (and then also with awkward 2.2.1 and dask_awkward 2023.6.0) and you will be able to reproduce my claims.

lgray:
To answer your question: I can cause the bug in dask_awkward *without* nanoevents remapping.

This over-greedy columns stuff here I can't figure out how to make a non-complex reproducer.

lgray:
And I agree with you on with_field for typetracers and how that touches in a single statement - but necessary_columns / dask.compute should be fed only the information they need to calculate the exact output requested. I have a feeling this one will smooth out as time goes by and we evolve in the next couple weeks. For sure, the optimization over structure rather than low level knowledge of inputs, which we are about get, will yield a better answer.

Whatever the underlying issue is that's in this PR, which I think is separate from the dask_awkward issue (they just appeared related at first), is of a slightly different flavor.

agoose77:
I replied to @lgray on Slack, but I'll repost it here for posterity, so that others can track why this PR is being closed.

@lgray I'm not fully tracking the notion that there are two issues here. Let me restate what I think is going on, and you can hopefully tell me if we're on the same page.


I would expect that existing workflows using PyPI releases of `awkward`, `uproot`, and the latest `main` of `dask-awkward` will lead to increased touching such that workflows become slower to execute. This is because;
* https://github.com/scikit-hep/awkward/commit/19f62aaac1f77c8d8096730f56dead782260d400 fixes a long-standing bug in which advanced slices (that evaluate pointers in NumPy) weren't considered as touching operations
* `ak.with_field` et al. use broadcasting, which performs these kinds of index operations
* awkward-dask assumes that only columns of the same parentage have the same offsets, i.e. contents of the same `RecordArray`. This is not how Awkward actually works; buffers (offsets) can be re-used across unrelated layouts.
* `uproot.dask` returns arrays of type `{field1: var * ..., field2: var * ..., field3: var * ...}` instead of `var * {field1: ..., field2: ..., field3: ...}`, meaning that there's insufficient information to know that the offsets of all fields are identically.

The solution is therefore to encode information that offsets are shared. Arrays with no shared parentage like those returned from `uproot.dask` need some auxiliary information that encodes this, e.g. https://github.com/scikit-hep/uproot5/pull/900:
* nanoevents can rebuild the form such that the type is `var * {field1: ..., field2: ..., field3: ...}`, meaning that the optimisation logic can properly detect this case. Although the broadcasting in e.g. `with_field` still touches the offsets all the way to the record layout, it transpires that they're the same form key, and the non-read field _itself_ is not read.

Without the change that fixed touching and caused this new behaviour, results would vary depending upon whether executed via dask or awkward, which would be bad.

PR #2523, opened 2023-06-13
fix: protect ak.to_parquet against memory explosion when args are swapped.

jpivarski:
If an Awkward Array is passed as `destination` to `fsspec.core.url_to_fs(destination)`, the memory use explodes before Python crashes. The array doesn't even need to be very large to send fsspec into dozens of GB; I think it might be converting the array into a string and interpreting `[` `]` `{` `}` as wildcards, a combinatorially large file list.

Anyway, it's very common for me to forget the argument order in I/O functions (`ak.to_parquet`, `json.dump`, etc.) and I just try one until it works. It needs to be the case that trying the wrong one produces an immediate error message.

### DIFF
--- a/src/awkward/_regularize.py
+++ b/src/awkward/_regularize.py
@@ -50,30 +50,6 @@ def is_non_string_like_sequence(obj) -> bool:
     return not isinstance(obj, (str, bytes)) and isinstance(obj, Sequence)
 
 
-def regularize_path(path):
-    """
-    Converts pathlib Paths into plain string paths (for all versions of Python).
-    """
-    is_path = False
-
-    if isinstance(path, getattr(os, "PathLike", ())):
-        is_path = True
-        path = os.fspath(path)
-
-    elif hasattr(path, "__fspath__"):
-        is_path = True
-        path = path.__fspath__()
-
-    elif path.__class__.__module__ == "pathlib":
-        import pathlib
-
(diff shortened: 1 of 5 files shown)


jpivarski:
Improvements: this function isn't going through the standard `ErrorContext`, so the error message contains text about the fact that this is happening in `ak.to_parquet` and what the correct order of arguments ought to be. These tidbits wouldn't be necessary if there was an error context.

Also, only strings for `destination`? That's what the docstring says. What about accepting a `pathlib.Path`? [pq.ParquetWriter](https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetWriter.html) accepts a file-like object—we could in principle pass on that capability, but we'd have to bypass a lot of fsspec to do it.

raymondEhlers:
Just poking around and saw this, so feel free to disregard - although it's just a nice quality of life thing, it would be nice to have support for Path. Every time I use `to_parquet`, I forget to convert from Path, so if it's not a big burden, it would be a helpful addition

jpivarski:
That one's easy. Allowing file-like objects to pass through to the `ParquetWriter`, to take advantage of its ability to use them, would be harder.

agoose77 (approved):
This looks good! I made a change to open up support for any path-likes, are you happy with these modifications? I'll just test them locally to ensure that there are no corner-cases.

PR #2527, opened 2023-06-16
feat!: canonicalise `union` of `indexed`

### LINKED ISSUE #2192: IndexedArrays can be absorbed into a UnionArray's index (unless categorical)
### Version of Awkward Array

HEAD

Currently, UnionArrays are allowed to contain IndexedArrays:

```python
array = ak.Array(
    ak.contents.UnionArray.simplified(
        ak.index.Index8([0, 0, 0, 1, 1, 1, 1, 1]),
        ak.index.Index64([0, 1, 2, 0, 1, 2, 3, 4]),
        [
            ak.from_iter(["one", "two", "three"], highlevel=False),
            ak.contents.IndexedArray(
                ak.index.Index64([4, 3, 2, 1, 0]),
                ak.from_iter([1.1, 2.2, 3.3, 4.4, 5.5], highlevel=False),
            )
        ],
    )
)
```

has layout

```python
>>> array.layout
<UnionArray len='8'>
    <tags><Index dtype='int8' len='8'>[0 0 0 1 1 1 1 1]</Index></tags>
    <index><Index dtype='int64' len='8'>[0 1 2 0 1 2 3 4]</Index></index>
    <content index='0'>
        <ListOffsetArray len='3'>
            <offsets><Index dtype='int64' len='4'>
                [ 0  3  6 11]
            </Index></offsets>
            <content><NumpyArray dtype='uint8' len='11'>
                [111 110 101 116 119 111 116 104 114 101 101]
            </NumpyArray></content>
        </ListOffsetArray>
    </content>
    <content index='1'>
        <IndexedArray len='5'>
            <index><Index dtype='int64' len='5'>[4 3 2 1 0]</Index></index>
            <content><NumpyArray dtype='float64' len='5'>[1.1 2.2 3.3 4.4 5.5]</NumpyArray></content>
        </IndexedArray>
    </content>
</UnionArray>
```

But because of

  * https://github.com/scikit-hep/awkward/blob/0.1.15/docs/theory/arrays-are-functions.pdf
  * page 22‒23 https://indico.cern.ch/event/833895/contributions/3577882/
  * page 25 https://www.thestrangeloop.com/2019/jagged-ragged-awkward-arrays.html, starting at 18:24 in https://youtu.be/2NxWpU7NArk?t=1104

the application of the `index[tags == tag]` array in UnionArray is associative with the application of the `index` in the IndexedArray. The latter can be merged into the former with no loss of information, and then we're using less memory and performing less indirection in computations. (Is it enough to matter for any applications? Unclear, because we like to avoid UnionArrays in general, but it is a net win.)

This issue would be resolved by doing a final pass at the end of `UnionArray.simplified` to check for any IndexedArrays in `contents`, not IndexedOptionArrays, that are not categorical. (IndexedOptionArrays can't be merged into `UnionArray.index` because they have `-1` for missing values, which `UnionArray.index` does not, and we don't want to apply this to categorical IndexedArrays because then we'd have to put the `__array__: "categorical"` parameter on the UnionArray, which downstream code is not expecting, and anyway it would lose information because if some of the `UnionArray.contents` were categorical and others not, we'd lose knowledge of which is which.)

If the `outindex` (`index` of the UnionArray that will be returned) is a new array, created inside the `UnionArray.simplified` function, it can be modified in place, something like

```python
for tag, content in enumerate(outcontents):
    if isinstance(content, IndexedArray) and content.parameter("__array__") != "categorical":
        # only want to affect the part of the UnionArray.index for this tag
        selection = (outtags == tag)
        # function-composition of this part of the UnionArray.index with the IndexedArray.index
        outindex[selection] = content.index[outindex[selection]]
        # now we don't have an IndexedArray anymore
        outcontents[tag] = content.content
```

If not, then `outindex` can be copied before the first modification.

Then `UnionArray.__init__` can forbid non-categorical IndexedArrays as `contents`. Even fewer combinations to worry about: yay!

We should consider the case of IndexedArray-of-RecordArray, which is created by `_carry` with `allow_lazy=True`. This would eliminate the IndexedArray that lazily carries the RecordArray, but the RecordArray would still be lazily carried (that is, the `_carry` would still not be propagated to all of the RecordArray's `contents`) because that laziness is in the `UnionArray.index` now. When/if the RecordArray ever gets projected out of this UnionArray, if that happens with `allow_lazy=True`, then it's _still_ lazy (not any worse for having made this optimization). But if it gets projected out with `allow_lazy=False`, then that specific case could have a performance degradation due to this change. That seems hyperspecialized, though.

agoose77:
This PR:
- Deprecates the `merge` argument to `UnionArray.simplified` — since the previous introduction of canonicalisation checks, this argument can no longer be meaningfully `False` in contexts where layouts are mergeable.
- Add a new rule for the simplification of `IndexedArray`s inside unions.

Closes #2192

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -1276,7 +1276,7 @@ class Content:
         return (
             self.__class__ is other.__class__
             and len(self) == len(other)
-            and type_parameters_equal(self.parameters, other.parameters)
+            and type_parameters_equal(self._parameters, other._parameters)
             and self._is_equal_to(other, index_dtype, numpyarray)
         )
 
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -835,7 +835,6 @@ class IndexedOptionArray(Content):
             index,
             contents,
             parameters=self._parameters,
-            merge=True,
             mergebool=True,
         )
 
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1155,7 +1155,10 @@ class UnionArray(Content):
 
             # This pathway is covered by `.simplified`, but we can avoid an extra array operation
             # by performing this now
-            elif isinstance(array, ak.contents.IndexedArray):
+            elif (
+                isinstance(array, ak.contents.IndexedArray)
+                and array.parameter("__array__") != "categorical"
+            ):
                 assert nexttags.nplike is index_nplike
                 self._backend.maybe_kernel_error(
                     self._backend[
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -8,7 +8,7 @@ from collections.abc import Iterable, MutableMapping, Sequence
 
 import awkward as ak
 from awkward._backends.backend import Backend
-from awkward._errors import AxisError
+from awkward._errors import AxisError, deprecate
 from awkward._layout import maybe_posaxis
 from awkward._nplikes.jax import Jax
 from awkward._nplikes.numpy import Numpy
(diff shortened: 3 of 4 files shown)


jpivarski (approved):
I know we had this conversation, but I'll reiterate that I agree that `IndexedArray` inside a `UnionArray` can get absorbed into the `UnionArray`'s `index`, and therefore should be. Unless it's categorical, which is a user-visible type.

I added the `UnionArray.simplified` `merge`-argument deprecation to the wiki. (And maybe was too vocal about it being more like a bug-fix than a deprecation, but I don't think it matters what we call it.)

You can merge when ready!

PR #2528, opened 2023-06-16
feat!: deprecate string behavior classes

agoose77:
As discussed in #2432, we don't appear to use these behavior classes anywhere. This PR deprecates them, and removes them from the global behavior dictionary.

This PR _doesn't_ remove the behavior classes for `byte` and `char`, because these _are_ used, including for `==` overloads.

- [x] Deprecate `string` and `bytestring` behavior classes

### DIFF
--- a/docs/user-guide/how-to-convert-python.md
+++ b/docs/user-guide/how-to-convert-python.md
@@ -314,7 +314,7 @@ ak.Array([b"one", b"two", b"three", b"four"]).to_list()
 Advanced topic: the rest of this section may be skipped if you don't care about internal representations.
 ```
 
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is good. It removes all dependence on `StringBehavior` and `ByteStringBehavior`, making them essentially dead code, apart from the fact that they're a public API. It would be very weird for someone to be using these classes downstream, now that they're not a functional part of Awkward Array anymore, but deprecating, rather than deleting, is the right thing to do.

Let's merge this so that it gets into 2.3.0 on July 1. I've already added it to the [list of deprecations](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20).

PR #2530, opened 2023-06-19
feat: add array dispatcher (non-inline)

agoose77:
## Goal 
It would be helpful for users of `dask-awkward` if we could provide a mechanism to use the `ak.` namespace on these third-party array objects, e.g.
```python
import awkward as ak

import dask_awkward as dak

array = ak.Array([[False, False, True, True], [], [True]])
darray = dak.from_awkward(array, npartitions=3)

dtotal = ak.sum(darray, axis=-1, keepdims=True)
dresult = ak.any(dtotal > 1, axis=-1, keepdims=True)

print(dresult.compute())
```

This PR implements a mechanism similar to NEP-18 for delegating to array objects for the concrete operation implementation. This behaves as follows:
- each high-level function implements a dispatcher that yields possible array-like objects that may implement this protocol
- if _any_ array-like implements the `__awkward_function__` protocol, we return the result of invoking that function
- if multiple arrays implement an `__awkward_function__`, the first as yielded by the `dispatcher` wins. In future we could require that `__awkward_function__` be defined _on the class_ and require that only a single _type_ implement this method.

For `dask-awkward` this function might look as follows:
```python
@classmethod
def __awkward_function__(cls, func, dispatched, args, kwargs):
    assert all(isinstance(x, cls) for x in dispatched)

    import dask_awkward
	name = func.__qualname__
    try:
        impl = getattr(dask_awkward, name)
    except AttributeError:
        raise NotImplementedError from None
    return impl(*args, **kwargs)
```

### Notes
- Unlike NumPy's mechanism, we don't try and support the case that there are multiple third-party arrays. I think this use case fairly unlikely at this stage, so at least for now it is not necessary to cater for it.
- The dispatcher must have an identical signature to the implementation. This avoids the need to use something like `inspect.signature` to prepare the arguments upon each invocation, i.e. a design tradeoff. This is also checked by the decorator upon first invocation. We could also perform such a test at import time, or only in our CI, but I think better to ensure this always runs for an invocation.

### DIFF
--- a/src/awkward/_dispatch.py
+++ b/src/awkward/_dispatch.py
@@ -1,0 +1,40 @@
+from functools import wraps
+from inspect import signature
+
+from awkward._errors import OperationErrorContext
+from awkward._typing import Callable, Iterator, TypeAlias, TypeVar
+
+DispatcherType: TypeAlias = Iterator
+
+
+T = TypeVar("T", bound=Callable)
+
+
+def high_level_function(dispatcher=None):
+    """Decorate a high-level function such that it may be overloaded by third-party array objects"""
+
+    def decorator(func: T) -> T:
+        # Let's check the signature!
+        if not (dispatcher is None or signature(func) == signature(dispatcher)):
+            raise RuntimeError(
+                f"Array dispatcher for {func.__qualname__} is incompatible with implementation"
+            )
+
+        @wraps(func)
+        def dispatch(*args, **kwargs):
+            # NOTE: this decorator assumes that the operation is exposed under `ak.`
+            with OperationErrorContext(f"ak.{func.__qualname__}", args, kwargs):
+                if dispatcher is not None:
+                    dispatched = tuple(dispatcher(*args, **kwargs))
+                    for arg in dispatched:
+                        try:
+                            custom_impl = arg.__awkward_function__
+                        except AttributeError:
+                            continue
+                        return custom_impl(dispatch, dispatched, args, kwargs)
+
+                return func(*args, **kwargs)
+
+        return dispatch
+
+    return decorator
--- a/src/awkward/_dispatch.py
+++ b/src/awkward/_dispatch.py
@@ -1,4 +1,3 @@
-import os
 from functools import wraps
 from inspect import signature
 
@@ -15,16 +14,21 @@ def high_level_function(dispatcher=None):
     """Decorate a high-level function such that it may be overloaded by third-party array objects"""
 
     def decorator(func: T) -> T:
-        # Don't always check something that should be correct at author time
-        if "AWKWARD_CHECK_DISPATCH_SIGNATURES" in os.environ and not (
-            dispatcher is None or signature(func) == signature(dispatcher)
-        ):
-            raise RuntimeError(
-                f"Array dispatcher for {func.__qualname__} is incompatible with implementation"
-            )
+        signatures_validated = False
 
         @wraps(func)
         def dispatch(*args, **kwargs):
+            nonlocal signatures_validated
+
+            # Don't always check something that should be correct at author time
+            if not (signatures_validated or dispatcher is None):
+                if signature(func) == signature(dispatcher):
+                    signatures_validated = True
+                else:
+                    raise RuntimeError(
+                        f"Array dispatcher for {func.__qualname__} is incompatible with implementation"
+                    )
+
             # NOTE: this decorator assumes that the operation is exposed under `ak.`
             with OperationErrorContext(f"ak.{func.__qualname__}", args, kwargs):
                 if dispatcher is not None:
--- a/src/awkward/_dispatch.py
+++ b/src/awkward/_dispatch.py
@@ -1,3 +1,4 @@
+import os
 from functools import wraps
 from inspect import signature
 
@@ -14,8 +15,10 @@ def high_level_function(dispatcher=None):
     """Decorate a high-level function such that it may be overloaded by third-party array objects"""
 
     def decorator(func: T) -> T:
-        # Let's check the signature!
-        if not (dispatcher is None or signature(func) == signature(dispatcher)):
+        # Don't always check something that should be correct at author time
+        if "AWKWARD_CHECK_DISPATCH_SIGNATURES" in os.environ and not (
+            dispatcher is None or signature(func) == signature(dispatcher)
+        ):
             raise RuntimeError(
                 f"Array dispatcher for {func.__qualname__} is incompatible with implementation"
             )
--- a/src/awkward/operations/ak_all.py
+++ b/src/awkward/operations/ak_all.py
@@ -3,7 +3,7 @@ __all__ = ("all",)
 import awkward as ak
 from awkward._behavior import behavior_of
 from awkward._connect.numpy import UNSUPPORTED
-from awkward._errors import with_operation_context
+from awkward._dispatch import high_level_function
 from awkward._layout import wrap_layout
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._regularize import regularize_axis
@@ -11,7 +11,19 @@ from awkward._regularize import regularize_axis
 np = NumpyMetadata.instance()
 
 
-@with_operation_context
+def _dispatcher(
+    array,
+    axis=None,
+    *,
+    keepdims=False,
+    mask_identity=False,
+    highlevel=True,
+    behavior=None,
+):
+    yield array
+
+
+@high_level_function(_dispatcher)
 def all(
     array,
     axis=None,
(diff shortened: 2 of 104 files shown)


agoose77:
The `dispatcher` mechanism is something we might want to use for the high-level array formatting, too. Right now we don't do anything special to format containers-of-arrays, and we detect these using a recursive (depth-limited) search. For now, I'm not going to think about it :)

agoose77:
I also noticed a couple of places where L3 functions were L2 and being decorated as high-level functions. This PR corrects those.

agoose77:
These test failures are odd. I can't reproduce them locally, and I've not touched any of the code that I'd expect to trigger them.

jpivarski:
Do the tests running directly on `main` work?

agoose77:
Strangely not: https://github.com/scikit-hep/awkward/actions/runs/5312769508/jobs/9617751770. But my test suite passes locally, and these were passing historically. 

The codecov workflow _is_ passing on main. This workflow does run a different set of steps, so it's possible that a dependency is at fault. I think I'll try and spin up a clean environment image to test in.

agoose77:
A note on the design. We could have the dispatch logic performed _in_ the implementation, by prepending the dispatch yield statements:
```python
from inspect import isgenerator
from functools import wraps
import builtins


def dispatched(func):
    @wraps(func)
    def wrapper(*args, **kwargs):
        gen_or_result = func(*args, **kwargs)
        if isgenerator(gen_or_result):
            try:
                while True:
                    arg = next(gen_or_result)
                    try:
                        impl = arg.__awkward_function__
                    except AttributeError:
                        continue
                    else:
                        return impl(wrapper, *args, **kwargs)
            except StopIteration as err:
                return err.value
        else:
            return gen_or_result

    return wrapper


@dispatched
def ak_any(array):
    yield array

    return builtins.any(array)


@dispatched
def ak_all(array):
    yield array

    return builtins.all(array)


array = [1, 2, 3, 0]
result = ak_any(array)
print(result)


class AnyToAllArray(list):
    ...

    def __awkward_function__(self, func, *args, **kwargs):
        if func is ak_any:
            (array,) = args
            return ak_all(list(array))
        elif func is ak_all:
            (array,) = args
            return ak_all(list(array))
        raise NotImplementedError(func)


result = ak_any(AnyToAllArray(array))
print(result)

```

I'm not sure whether this is better or just more fun to design. Probably the latter!

src/awkward/_dispatch.py
```diff
@@ -0,0 +1,47 @@
+from functools import wraps
+from inspect import signature
```
jpivarski, line 2: Tools from `inspect` have a habit of not working when the function is defined on the fly (i.e. not in an imported module). That would definitely be true if you were asking for the source code or line number, but I don't know about `inspect.signature`. If it's implemented using other pieces of `inspect`, it would have this problem. But I don't know if it does, since a `function.__code__` object contains all the information one would need in order to reconstruct a signature...

src/awkward/operations/ak_all.py
```diff
@@ -3,15 +3,27 @@
...
+    highlevel=True,
+    behavior=None,
+):
+    yield array
+
+
+@high_level_function(_dispatcher)
 def all(
     array,
     axis=None,
```
jpivarski, line 29: By adding `with_operation_context`, you managed to eliminate the need to manually duplicate each high-level signature, but it looks like `_dispatcher` make you have to put them all back. What's worse, now you also need to specify the defaults twice.

If you can extract the signature from third-party overloaded `__awkward_function__`s, can't you extract the signature from our own high-level functions?

You can chain decorators, if that helps: it doesn't need to be passed as an argument to `high_level_function`. Would there be a way to do it like

```python
@with_operation_context
@_dispatcher
def all(
    ...
)
```

and have `_dispatcher` pull the signature off of the `all` function that gets passed to it?

jpivarski (commented):
For the test failures, maybe the tests are caching something? Can you reset all caches (with a big hammer) to find out?

It looks like you find out if a given `__awkward_function__` applies by seeing if it raises an error or lacks a `__qualname__` (?); the usual way is for the function to _return_ (not raise) `NotImplemented` (not `NotImplementedError`). Would that make sense here?

src/awkward/_dispatch.py
agoose77: `inspect.signature` is now fairly fundamental to how `inspect` works. As such, it's reliable for any function you can generate.

src/awkward/operations/ak_all.py
agoose77: > By adding with_operation_context, you managed to eliminate the need to manually duplicate each high-level signature, but it looks like _dispatcher make you have to put them all back. What's worse, now you also need to specify the defaults twice.

Partly. This PR requires copy-and-paste signatures, which is easier to check and keep in sync (we can test the signature programmatically).

> If you can extract the signature from third-party overloaded __awkward_function__s, can't you extract the signature from our own high-level functions?

The purpose of `dispatcher` is to make Awkward's mechanism aware of which arguments are array-like. We don't _need_ to use an identical signature, but if the dispatch mechanism is a separate function, then any modification to the signature (such as permitting the dispatcher to just use non-keyword-only, non-position-only arguments arguments) requires converting between the two. This could (should) be done using `inspect.Signature.bind()`, which adds some overhead (I mention this [here](https://github.com/scikit-hep/awkward/pull/2530#issue-1763535738)). I'm not yet clear on where the equilibrium point is between convenience & expressiveness vs performance in this aspect of the code, but being mindful of the graph optimisation cost penalty suffered by coffea workflows in dask-awkward, I am reluctant to make things slower out of the gate.

> it doesn't need to be passed as an argument to high_level_function. Would there be a way to do it like

Yes, originally I had this inverted; the decorator was added to `ak.any`, followed by
```python
@ak.any.dispatcher
def _dispatcher(...):
    ...
```

I ended up preferring the upfrontness of this PR (inspired by NumPy's implementation).

---

We can otherwise use a naive argument-inspection implementation that is permitted to look for arrays recursively in tuples and lists. I don't think that's worth pursuing though; there will always be types e.g. iterators that we can't (shouldn't) inspect, and it seems silly not to encode up-front the knowledge about which arguments (or their contents) are arrays.

Another approach that avoids the problem entirely is to just do everything inline: https://github.com/scikit-hep/awkward/pull/2531

agoose77: > it doesn't need to be passed as an argument to high_level_function. Would there be a way to do it like

Yes, originally I had this inverted; the decorator was added to `ak.any`, and then 
```python
@ak.any.dispatcher
def _dispatcher(...):
    ...
```

I ended up preferring the up-frontness of this PR (inspired by NumPy's implementation).

src/awkward/_dispatch.py
agoose77: > It looks like you find out if a given __awkward_function__ applies by seeing if it raises an error or lacks a __qualname__ (?); the usual way is for the function to return (not raise) NotImplemented (not NotImplementedError). Would that make sense here?

No — this is actually the client implementation (e.g. `dask-awkward`). The idea is that unlike NEP-18, we don't need to try multiple implementations to handle a function. If any array-like argument implements the Awkward protocol, it becomes responsible for handling the call (and must error if it can't). This simplifies things considerably, and I don't think that we need to support the multiple-array-type case (e.g. dask-awkward mixed with some other library). Do you feel this is too restrictive? In future, should we decide to relax this constraint, we could (and interpret `NotImplementedError` as `return NotImplemented`).

jpivarski: Yes, since there's a pathway to extend this to the `return NotImplemented` pattern in the future, let's go ahead with it.

Version 1: if you write a `__awkward_function__` method, you must provide an implementation. That implementation might raise an exception, but that exception will be passed to the user and will look like an error. (I.e. we won't catch `NotImplementedError` or any other exception raised by a third-party `__awkward_function__`. We only anticipate one downstream library implementing `__awkward_function__` methods, and that's dask-awkward.

Version 2: if you write a `__awkward_function__` method, you can `return NotImplemented`, and then we'll check the other arguments for `__awkward_function__` methods. The value of this would just to be "normal," to adhere to a common pattern in the Python world.

Adding version 2 after version 1 has been out for a while wouldn't invalidate the libraries written for version 1, since they have to not `return NotImplemented` in order to adhere to the version 1 protocol. The only odd case is if `NotImplemented` was the _desired output_ of an Awkward function that a third-party library wanted to return. That's very unlikely. But we could plug that gap by adding a restriction to the version 1 protocol: "It is illegal to return `NotImplemented`." That would make it like a reserved bit.

src/awkward/operations/ak_all.py
jpivarski: Manually duplicated arguments look bad in the source code, but if they're automatically checked, then they're not actually-bad for maintenance. Let's go with this, then.

agoose77:
Closing in favour of #2531

PR #2531, opened 2023-06-19
feat: add array dispatcher (inline)

agoose77:
This PR is an alternative to #2530. It moves the dispatch mechanism _into_ the implementation, which perhaps improves readability. 

Here's an example of `broadcast_arrays` with this dispatch mechanism variant:

```python

@high_level_function
def broadcast_arrays(
    *arrays,
    depth_limit=None,
    broadcast_parameters_rule="one_to_one",
    left_broadcast=True,
    right_broadcast=True,
    highlevel=True,
    behavior=None,
):
    # Dispatch
    yield arrays

    # Implementation
    return _impl(
        arrays,
        depth_limit,
        broadcast_parameters_rule,
        left_broadcast,
        right_broadcast,
        highlevel,
        behavior,
    )
```

Third-party arrays can define the `__awkward_function__` method to intercept `ak.XXX` functions, e.g.
```python
@classmethod
def __awkward_function__(cls, func, array_likes, args, kwargs):
    import dask_awkward
	name = func.__qualname__   

    try:
        impl = getattr(dask_awkward, name)
    except AttributeError:
        return NotImplemented

    return impl(*args, **kwargs)
```

### Limitations
1. High-level function results cannot be generators if they are decorated with `high_level_function`. In this case, the bare error-handler context manager should be used.

### DIFF
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -126,7 +126,7 @@ def broadcast_arrays(
             output = []
             for inner in outer:
                 output.append(x + inner)
-            yield arrayput
+            yield output
 
     where `x` has the same value for each `inner` in the inner loop.
 
@@ -179,7 +179,7 @@ def broadcast_arrays(
          [[6.6]]]
     """
     # Dispatch
-    yield from arrays
+    yield arrays
 
     # Implementation
     return _impl(
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -5,7 +5,7 @@ from awkward._backends.dispatch import backend_of
 from awkward._backends.numpy import NumpyBackend
 from awkward._behavior import behavior_of
 from awkward._connect.numpy import UNSUPPORTED
-from awkward._errors import with_operation_context
+from awkward._dispatch import high_level_function
 from awkward._layout import wrap_layout
 from awkward._nplikes.numpylike import NumpyMetadata
 
@@ -13,7 +13,19 @@ np = NumpyMetadata.instance()
 cpu = NumpyBackend.instance()
 
 
-@with_operation_context
+def _dispatcher(
+    *arrays,
+    depth_limit=None,
+    broadcast_parameters_rule="one_to_one",
+    left_broadcast=True,
+    right_broadcast=True,
+    highlevel=True,
+    behavior=None,
+):
+    yield from arrays
+
+
+@high_level_function(_dispatcher)
 def broadcast_arrays(
     *arrays,
     depth_limit=None,
@@ -126,7 +138,7 @@ def broadcast_arrays(
             output = []
             for inner in outer:
                 output.append(x + inner)
-            yield output
+            yield arrayput
 
     where `x` has the same value for each `inner` in the inner loop.
 
--- a/src/awkward/operations/ak_broadcast_arrays.py
+++ b/src/awkward/operations/ak_broadcast_arrays.py
@@ -13,19 +13,7 @@ np = NumpyMetadata.instance()
 cpu = NumpyBackend.instance()
 
 
-def _dispatcher(
-    *arrays,
-    depth_limit=None,
-    broadcast_parameters_rule="one_to_one",
-    left_broadcast=True,
-    right_broadcast=True,
-    highlevel=True,
-    behavior=None,
-):
-    yield from arrays
-
-
-@high_level_function(_dispatcher)
+@high_level_function
 def broadcast_arrays(
     *arrays,
     depth_limit=None,
@@ -190,6 +178,10 @@ def broadcast_arrays(
          [],
          [[6.6]]]
     """
+    # Dispatch
+    yield from arrays
+
+    # Implementation
     return _impl(
         arrays,
         depth_limit,
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,7 +27,6 @@ env:
   # This will ensure that the hash of the awkward-cpp directory remains
   # constant for unchanged files, meaning that it can be used for caching
   SOURCE_DATE_EPOCH: "1668811211"
-  AWKWARD_CHECK_DISPATCH_SIGNATURES: 1
 
 jobs:
   Windows:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,6 +27,7 @@ env:
   # This will ensure that the hash of the awkward-cpp directory remains
   # constant for unchanged files, meaning that it can be used for caching
   SOURCE_DATE_EPOCH: "1668811211"
+  AWKWARD_CHECK_DISPATCH_SIGNATURES: 1
 
 jobs:
   Windows:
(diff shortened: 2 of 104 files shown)


src/awkward/operations/ak_broadcast_arrays.py
```diff
@@ -126,7 +126,7 @@ def broadcast_arrays(
             output = []
             for inner in outer:
                 output.append(x + inner)
-            yield output
+            yield arrayput
```
jpivarski: Did you find-and-replace `out` → `array` to get "`arrayput`"?

jpivarski:
Oh, I don't think so: now every `ak.*` function is both a function _and_ a generator? An object being iterable is a user-visible property, and who knows what kind of confusing situation someone might get into by having this very surprising property on all of our functions. Moreover, the reason/rationale isn't obviously related to the purpose you're trying to achieve (i.e. it's purely technical).

Since this is an alternative, let's go with the original.

src/awkward/operations/ak_broadcast_arrays.py
agoose77: Ha, yes, but not in the same pass. This one snuck in despite a thorough check (not thorough enough ...)

agoose77:
> [N]ow every ak.* function is both a function and a generator? An object being iterable is a user-visible property, and who knows what kind of confusing situation someone might get into by having this very surprising property on all of our functions. 

Users won't see this unless they go digging into the source code; the decorator ensures that the function behaves like a function.

> Moreover, the reason/rationale isn't obviously related to the purpose you're trying to achieve (i.e. it's purely technical).

Hmm, I'm not sure I agree. We're constrained by several factors here: readability, code duplication, and ultimately the language features at our disposal. In this context, generators let us exit the execution of the function, and pass control flow to the caller. We can otherwise do this with separate functions (see #2530), a gnarly explicit dispatch mechanism (see below), or something like a closure (see below also!)

### Callback
```python
class DispatchTarget:
    def __init__(self, target):
        self.target = target

def maybe_dispatch(*arrays):
    for array in arrays:
        if hasattr(array, '__awkward_function__'):
            return DispatchTarget(array.__awkward_function__)

def dispatched(func):
    def wrapper(*args, **kwargs):
        dispatch_or_result = func(*args, **kwargs)
        if isinstance(dispatch_or_result, DispatchTarget):
            return dispatch_or_result.target(wrapper, args, kwargs)
        else:
            return dispatch_or_result
    return wrapper
```
that yields
```python
@dispatched
def any(array, ...):
	dispatch = maybe_dispatch(array)
    if dispatch is not None:
        return dispatch
    else:
        return ...
```

This gets messier if there are multiple arguments that need to be handled separately.

I find the generator-based approach simpler:
```python
@dispatched
def any(array, ...):
    yield array
```

### Closure
```python
def dispatched(func):
    def wrapper(*args, **kwargs):
        impl, *arrays = func(*args, **kwargs)
        for array in arrays:
            if hasattr(array, '__awkward_function__'):
                return array.__awkward_function__(
                    wrapper, args, kwargs
                )
        return impl()
    return wrapper
```
```python
@dispatched
def any(array, ...):
    def impl():
        ...
    return impl, array
```

jpivarski:
> the decorator ensures that the function behaves like a function.

Oh, I didn't recognize that. In that case, what I said after that point is a moot point.

agoose77:
@douglasdavis if we push ahead with this PR, then `dask-awkward` will be able to define a `__awkward_function__` classmethod (or instance method, at present) on `dask_awkward.Array` that enables `ak.high_level_function(dask_array)`. As it stands, if any array-like argument has this method, we dispatch to it instead of calling our own implementation.

Will this process be sufficient for dask-awkward? Is it reasonable to have `dask_awkward` check the types of the arguments in each function to ensure you're not mixing dask and non-dask awkward arrays? Would we want the ability for another third-party library to take precedence?

douglasdavis:
I'll take this PR for a spin and see how it feels!

douglasdavis:
one note:
dask_awkward relies heavily on `ak.from_json` in testing and this PR appears to have broken it:


```python
In [1]: import awkward as ak
   ...: 
   ...: 
   ...: with open("file.json", "wt") as f:
   ...:     text = '{"a":[1,2,3]}\n{"a":[4,5,6,7]}'
   ...:     print(text, file=f)
   ...: 
   ...: 
   ...: with open("file.json", "rb") as f:
   ...:     a = ak.from_json(f, line_delimited=True)
   ...: 

In [2]: a
Out[2]: <Array [] type='0 * unknown'>
```

douglasdavis:
In general this does look like a reasonable interface to me. This was all it took:

```diff
diff --git a/src/dask_awkward/lib/core.py b/src/dask_awkward/lib/core.py
index b53a8f0..7fb2729 100644
--- a/src/dask_awkward/lib/core.py
+++ b/src/dask_awkward/lib/core.py
@@ -1198,6 +1198,17 @@ class Array(DaskMethodsMixin, NDArrayOperatorsMixin):
         """Clear the divisions of a Dask Awkward Collection."""
         self._divisions = (None,) * (self.npartitions + 1)
 
+    def __awkward_function__(self, func, args, kwargs):
+        import dask_awkward
+
+        if any(isinstance(arg, ak.Array) for arg in args):
+            raise TypeError("cannot mix awkward.Array and dask_awkward.Array")
+
+        fn_name = func.__qualname__
+        fn = getattr(dask_awkward, fn_name)
+        res = fn(*args, **kwargs)
+        return res
+
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         if method != "__call__":
             raise RuntimeError("Array ufunc supports only method == '__call__'")
```

And this works as expected:


```python
In [1]: ds = dak.from_parquet("s3://ddavistemp/hpq", storage_options={"anon": True})

In [2]: ak.min(ak.fill_none(ak.min(ds.muons.pt, axis=1), -1.0))
Out[2]: dask.awkward<min, type=Scalar, dtype=Unknown>

In [3]: ak.min(ak.fill_none(ak.min(ds.muons.pt, axis=1), -1.0)).compute()
Out[3]: -1.0
```

agoose77:
Closing in favour of #2531

agoose77:
@douglasdavis OK, after decorating all of our internal functions, it's clear that there is little need for the ability to use multiple yield statements (which would follow from fiddly control-flow to unpack arguments that contain array-like objects). So, we can simplify the dispatch logic to expect a single yield statement. As such, this means that we _can_ pass the list of array-like objects, enabling dask-awkward to quickly reject the overload as an error.

Practically, this means dispatch functions changing from
```python
@high_level_function
def broadcast_arrays(*arrays, ...):
    yield from arrays
    return ...
```
to
```python
@high_level_function
def broadcast_arrays(*arrays, ...):
    yield arrays
    return ...
```

agoose77:
I'm done with this PR now, just added an error for multiple yields! Are you happy for it to be merged @jpivarski?

jpivarski:
Yes! This is a good design. The only question I have is about the dask-awkward implementation: why

```python
if any(isinstance(arg, ak.Array) for arg in args):
    raise TypeError("cannot mix awkward.Array and dask_awkward.Array")
```

Why not `dak.from_awkward`?

That's not a question for this PR, though. This is ready to merge.

agoose77:
I think in general it might be better to ensure that users load data in a dask-friendly manner. Whilst I suppose for things like `fill_none`, passing a small `ak.Array` is sensible, we don't want to encourage mixing these arrays in contexts that it is harmful to performance? I'm not yet sure on that, though — I'd be interested to see how things develop in dask-awkward, as you suggest.

PR #2115, opened 2023-01-12
fix: `TypeTracerArray` binary operators, `ak.Array.__str__` with a typetracer, attempts to call `touch_data` on non-typetracers, ...?

jpivarski:

PR #2534, opened 2023-06-20
fix: patch NDArrayOperatorsMixin

### LINKED ISSUE #2533: `__class__` assignment fails with `mixin_class` decorator
### Version of Awkward Array

main

Setting `__class__` to a class built by `ak.mixin_class` is failing for `numpy>=1.25.0`.

```python
import awkward as ak

behavior = {}


class Point:
    ...


class MyPoint(Point, ak.Array):
    ...


behavior["*", "Point"] = MyPoint  # type("MyPoint", (Point, ak.Array), {})

ak.operations.zip(
    {
        "x": ak.Array([1, 1]),
        "y": ak.Array([1, 1]),
    },
    with_name="Point",
    behavior=behavior,
)

```

agoose77:
Fixes #2533 by vendoring our own version of `NDArrayOperatorsMixin`

### DIFF
--- a/src/awkward/_connect/numexpr.py
+++ b/src/awkward/_connect/numexpr.py
@@ -2,6 +2,8 @@
 import sys
 import warnings
 
+from packaging.version import parse as parse_version
+
 import awkward as ak
 from awkward._behavior import behavior_of
 from awkward._layout import wrap_layout
@@ -25,9 +27,7 @@ or
         ) from err
     else:
         if not _has_checked_version:
-            if ak._util.parse_version(numexpr.__version__) < ak._util.parse_version(
-                "2.7.1"
-            ):
+            if parse_version(numexpr.__version__) < parse_version("2.7.1"):
                 warnings.warn(
                     "Awkward Array is only known to work with numexpr 2.7.1 or later"
                     "(you have version {})".format(numexpr.__version__),
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -6,6 +6,7 @@ from collections.abc import Iterable
 from itertools import chain
 
 import numpy
+from packaging.version import parse as parse_version
 
 import awkward as ak
 from awkward._backends.backend import Backend
@@ -20,13 +21,13 @@ from awkward._layout import wrap_layout
 from awkward._nplikes import to_nplike
 from awkward._regularize import is_non_string_like_iterable
 from awkward._typing import Iterator
-from awkward._util import Sentinel, numpy_at_least
+from awkward._util import Sentinel
 from awkward.contents.numpyarray import NumpyArray
 
 # NumPy 1.13.1 introduced NEP13, without which Awkward ufuncs won't work, which
 # would be worse than lacking a feature: it would cause unexpected output.
 # NumPy 1.17.0 introduced NEP18, which is optional (use ak.* instead of np.*).
-if not numpy_at_least("1.13.1"):
+if parse_version(numpy.__version__) < parse_version("1.13.1"):
     raise ImportError("NumPy 1.13.1 or later required")
 
 
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -3,6 +3,8 @@
 import json
 from collections.abc import Iterable, Sized
 
+from packaging.version import parse as parse_version
+
 import awkward as ak
 from awkward._backends.numpy import NumpyBackend
 from awkward._nplikes.numpy import Numpy
@@ -29,7 +31,7 @@ or
 """
 
 else:
-    if ak._util.parse_version(pyarrow.__version__) < ak._util.parse_version("7.0.0"):
+    if parse_version(pyarrow.__version__) < parse_version("7.0.0"):
         pyarrow = None
         error_message = "pyarrow 7.0.0 or later required for {0}"
 
--- a/src/awkward/_nplikes/numpy.py
+++ b/src/awkward/_nplikes/numpy.py
@@ -2,8 +2,8 @@
 from __future__ import annotations
 
 import numpy
+from packaging.version import parse as parse_version
 
-import awkward as ak
 from awkward._nplikes.array_module import ArrayModuleNumpyLike
 from awkward._nplikes.dispatch import register_nplike
 from awkward._nplikes.numpylike import ArrayLike, NumpyMetadata
@@ -57,7 +57,7 @@ class Numpy(ArrayModuleNumpyLike):
         bitorder: Literal["big", "little"] = "big",
     ):
         assert not isinstance(x, PlaceholderArray)
-        if ak._util.numpy_at_least("1.17.0"):
+        if parse_version(numpy.__version__) >= parse_version("1.17.0"):
             return numpy.packbits(x, axis=axis, bitorder=bitorder)
         else:
             assert axis is None, "unsupported argument value for axis given"
@@ -85,7 +85,7 @@ class Numpy(ArrayModuleNumpyLike):
         bitorder: Literal["big", "little"] = "big",
     ):
         assert not isinstance(x, PlaceholderArray)
-        if ak._util.numpy_at_least("1.17.0"):
+        if parse_version(numpy.__version__) >= parse_version("1.17.0"):
             return numpy.unpackbits(x, axis=axis, count=count, bitorder=bitorder)
         else:
             assert axis is None, "unsupported argument value for axis given"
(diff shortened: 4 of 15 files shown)


jpivarski (commented):
Why would NumPy introduce `__slots__` on a mixin? (Mixins are supposed to not be in charge of what data a class has, and therefore not define the `__init__` and attributes. Defining `__slots__` forces an exact set of data attributes, and therefore make the class not a mixin. #2533 is exactly the reason why mixins are okay for multiple inheritance and classes that define data are not.)

Since we have our own definition of `NDArrayOperatorsMixin`,

https://github.com/scikit-hep/awkward/blob/7f9be54cb71c21c255cf251f9d7aee38f6d5c09e/src/awkward/_connect/numpy.py#L315-L401

(though it was removed in #2115), I'd prefer to just use our definition and therefore control it. It means that we'll need to be on top of any new additional unary/binary operators (like `@`), but we don't have to be on the defensive about upstream technicalities like this.

douglasdavis:
numpy issue https://github.com/numpy/numpy/issues/22876 and PR https://github.com/numpy/numpy/pull/23113

jpivarski:
Oh, so my reading of this is that

  * NumPy only introduced `__slots__ = ()`
  * mixins without `__slots__` are not mixable into classes with `__slots__` (worse: Python just _ignores_ the subclassed `__slots__`)
  * a superclass/mixin with `__slots__ = ()` is unrestrictive about what slots downstream classes define.

By that logic, the NumPy developers concluded that there could be no downside to it. In #2533, we seem to have discovered a downside: `__class__` can't be assigned to by other classes that don't define `__slots__` because they have a different memory model.

Maybe a different solution to this is to define `__slots__` for `ak.Array` and `ak.Record`? That would put them in the same memory model as `NDArrayOperatorsMixin`-with-`__slots__`. If we do that, would that require downstream libraries from Awkward (i.e. Coffea and Vector) to also define `__slots__` in their behavior classes?

We don't want an infectious implementation requirement to spread to all downstream libraries that use Awkward, especially if the error message is cryptic, like

```
TypeError: __class__ assignment: 'MyAwesomeBehavior' object layout differs from 'NDArrayOperatorsMixin'
```

If defining `__slots__` for `ak.Array` and `ak.Record` infects downstream libraries with the need to define `__slots__` on their behaviors, then let's define our own `NDArrayOperatorsMixin` without `__slots__`. (At worst, downstream libraries will try to use `__slots__` and Python will ignore their performance hints.)

If defining `__slots__` for `ak.Array` and `ak.Record` has no implications for downstream libraries, let's define `__slots__` for `ak.Array` and `ak.Record`.

agoose77:
The way that the layout checker works is that it has a notion of type-supertype "compatibility", and it attempts to find a compatible common base. 

To generate our mixin classes for _both_ records and arrays, we do something like

```python
class ArrayImpl(MixinType, ak.Array):
    ...
```

The logic checking type-supertype compatibility for multiple inheritance walks down the `tp_base` (or `__base__` in Python) slot, and checks whether it is "compatible" with the type. If the child has a `__dict__` and the parent does not, this is `False`.

Let's imagine that we add slots to `ak.Array` and the _generated_ classes for records and arrays. If the _user_ doesn't define `__slots__`, we'll end up with a dict, and the compatibility check fails. Conversely, without `__slots__`, the base for the mixin (`ArrayImpl`) is `object`, whilst for `ak.Array` the base is `NDArrayOperatorsMixin`, which also fails.

Of the two approaches, it's probably safer to assume that the user _doesn't_ define `__slots__` for these mixin classes, because doing so would currently fail:
```python
import awkward as ak


behavior = {}


class Point:
    __slots__ = ()


class MyPoint(Point, ak.Array):
    ...


behavior["*", "Point"] = MyPoint

ak.operations.zip(
    {
        "x": ak.Array([1, 1]),
        "y": ak.Array([1, 1]),
    },
    with_name="Point",
    behavior=behavior,
)
```

For the same reasons outlined above: `MyPoint.__base__` is `Point`, which is the final "compatible base" for `MyPoint`, and is not the same base as that of `ak.Array`.

I think, therefore, the safest immediate step is to drop the slots, by using our own mixin. For our ecosystem, I think there's lesser benefit to `__slots__` usage; we don't expect high-frequency usage of array attributes to dominate over the array kernels. I might be wrong, though.

src/awkward/_connect/numexpr.py
```diff
@@ -2,6 +2,8 @@
 import sys
 import warnings
 
+from packaging.version import parse as parse_version
```
agoose77, line 5: Unrelated cleanup, but whilst we're here.

src/awkward/_connect/numpy.py
```diff
@@ -20,13 +21,13 @@
...
 from awkward._typing import Iterator
-from awkward._util import Sentinel, numpy_at_least
+from awkward._util import Sentinel
 from awkward.contents.numpyarray import NumpyArray
 
 # NumPy 1.13.1 introduced NEP13, without which Awkward ufuncs won't work, which
 # would be worse than lacking a feature: it would cause unexpected output.
 # NumPy 1.17.0 introduced NEP18, which is optional (use ak.* instead of np.*).
-if not numpy_at_least("1.13.1"):
+if parse_version(numpy.__version__) < parse_version("1.13.1"):
```
agoose77, line 30: Unrelated cleanup, but whilst we're here.

src/awkward/_connect/pyarrow.py
```diff
@@ -3,6 +3,8 @@
 import json
 from collections.abc import Iterable, Sized
 
+from packaging.version import parse as parse_version
```
agoose77, line 6: Unrelated cleanup, but whilst we're here.

src/awkward/_nplikes/numpy.py
```diff
@@ -2,8 +2,8 @@
 from __future__ import annotations
 
 import numpy
+from packaging.version import parse as parse_version
```
agoose77, line 5: Unrelated cleanup, but whilst we're here.

src/awkward/_util.py
```diff
@@ -24,16 +22,6 @@
...
 
 
-def parse_version(version):
-    return packaging.version.parse(version)
-
-
-def numpy_at_least(version):
-    import numpy  # noqa: TID251
-
-    return parse_version(numpy.__version__) >= parse_version(version)
```
agoose77, line 34: Unrelated cleanup, but whilst we're here.

src/awkward/numba.py
```diff
@@ -27,7 +28,7 @@ def register_and_check():
         ) from err
 
     if not _has_checked_version:
-        if ak._util.parse_version(numba.__version__) < ak._util.parse_version("0.50"):
+        if parse_version(numba.__version__) < parse_version("0.50"):
```
agoose77, line 31: Unrelated cleanup, but whilst we're here.

tests/test_1294_to_and_from_parquet.py
```diff
@@ -8,6 +8,7 @@
 import awkward as ak
 
 pyarrow_parquet = pytest.importorskip("pyarrow.parquet")
+fsspec = pytest.importorskip("fsspec")
```
agoose77, line 12: `ffspec` must have recently become optional elsewhere; we didn't depend upon it explicitly in these tests, which started failing on Mac.

jpivarski:
> I think, therefore, the safest immediate step is to drop the slots, by using our own mixin. For our ecosystem, I think there's lesser benefit to `__slots__` usage; we don't expect high-frequency usage of array attributes to dominate over the array kernels. I might be wrong, though.

Let's go with that. I was also thinking that `__slots__` aren't going to help much for `ak.Array`, considering that it contains so many large, nested, tree-like objects.

tests/test_1294_to_and_from_parquet.py
jpivarski: `ak.to_parquet` and `ak.from_parquet` rely fairly heavily on it, something that I am reminded of when I make environment.yml files for tutorials. Perhaps the pathway ought to be cleared for the case when we have a path to a local file or want to pass a file-like object directly into `pyarrow.parquet.ParquetFile` or `ParquetWriter`, so that fsspec is only needed for remote files. That's an aside.

agoose77: My initial reaction is to make `fsspec` a required dependency for IO (not for the `awkward` package itself, but we could provide extras to include it), given how fundamental it is to other parts of the Python ecosystem. I recall there being pushback on that from collaboration maintainers, but you'll have a better recollection of that conversation than I will, I suspect!

agoose77:
I've taken your above comment as approval for merging!

PR #2536, opened 2023-06-20
fix: improve performance of `Form.select_columns`

### LINKED ISSUE #2535: form.select_columns takes an hour for ~1500 columns
### Version of Awkward Array

2.2.3

Encountered when first implementing https://github.com/scikit-hep/uproot5/pull/905. 
There appear to be some implementations of `_select_columns` for the derived form classes that are very slow.

Reproducer, which may not take an hour but is still excessively slow:
```python3
import uproot

events = uproot.dask({"https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root": "Events"})
form = events._meta.layout.form
form.select_columns(form.columns())
```
This is what https://github.com/scikit-hep/uproot5/pull/905 was doing with column-optimization turned off until I worked around it, but you can imagine if the number of columns is large it's still a problem!

agoose77:
Fixes #2535

In `RecordForm._select_columns`, we perform an `N * M` iteration to test for matches. Previously, we were invoking `fnmatch.fnmatchcase`, which uses a 256-wide cache to compile & translate the regex pattern (IIRC `re.compile` also caches, with 512-wide, so that's likely our true effective cache size for the `compile` part). This means that in an `N * M` loop where `N` and `M` are greater than 256, the cache fails to improve performance. This PR identifies the pathological case of `N * M` containing mostly fixed strings, and replaces the calls to `fnmatchcase` with `string in set_of_strings`.

We could just implement our own version of `fnmatchcase` (3 lines) with a bigger cache, but that just defers the problem.

In practice, we know that if the specifiers given are _unique_ fixed strings, we can reduce this loop to `N` iterations. For now, I've not imposed this optimisation, and instead just made the `* M` loop much faster by leveraging the `item in set` time complexity.

### DIFF
--- a/src/awkward/forms/bitmaskedform.py
+++ b/src/awkward/forms/bitmaskedform.py
@@ -197,15 +197,23 @@ class BitMaskedForm(Form):
     def _columns(self, path, output, list_indicator):
         self._content._columns(path, output, list_indicator)
 
-    def _select_columns(self, match_specifier):
-        return BitMaskedForm(
-            self._mask,
-            self._content._select_columns(match_specifier),
-            self._valid_when,
-            self._lsb_order,
-            parameters=self._parameters,
-            form_key=self._form_key,
+    def _select_columns(
+        self, match_specifier, prune_interior_leaves: bool, is_inside_record: bool
+    ):
+        next_content = self._content._select_columns(
+            match_specifier, prune_interior_leaves, is_inside_record
         )
+        if next_content is None:
+            return None
+        else:
+            return BitMaskedForm(
+                self._mask,
+                next_content,
+                self._valid_when,
+                self._lsb_order,
+                parameters=self._parameters,
+                form_key=self._form_key,
+            )
 
     def _column_types(self):
         return self._content._column_types()
--- a/src/awkward/forms/bitmaskedform.py
+++ b/src/awkward/forms/bitmaskedform.py
@@ -197,10 +197,10 @@ class BitMaskedForm(Form):
     def _columns(self, path, output, list_indicator):
         self._content._columns(path, output, list_indicator)
 
-    def _select_columns(self, index, specifier, matches, output):
+    def _select_columns(self, match_specifier):
         return BitMaskedForm(
             self._mask,
-            self._content._select_columns(index, specifier, matches, output),
+            self._content._select_columns(match_specifier),
             self._valid_when,
             self._lsb_order,
             parameters=self._parameters,
--- a/src/awkward/forms/bitmaskedform.py
+++ b/src/awkward/forms/bitmaskedform.py
@@ -197,8 +197,8 @@ class BitMaskedForm(Form):
     def _columns(self, path, output, list_indicator):
         self._content._columns(path, output, list_indicator)
 
-    def _prune_columns(self, is_inside_record: bool) -> Self | None:
-        next_content = self._content._prune_columns(is_inside_record)
+    def _prune_columns(self, is_inside_record_or_union: bool) -> Self | None:
+        next_content = self._content._prune_columns(is_inside_record_or_union)
         if next_content is None:
             return None
         else:
--- a/src/awkward/forms/bitmaskedform.py
+++ b/src/awkward/forms/bitmaskedform.py
@@ -1,7 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 import awkward as ak
 from awkward._parameters import type_parameters_equal
-from awkward._typing import final
+from awkward._typing import Self, final
 from awkward._util import UNSET
 from awkward.forms.form import Form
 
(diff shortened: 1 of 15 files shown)


jpivarski (approved):
Okay, yeah—if someone has hundreds of wildcard patterns, we can expect that to be expensive, but if they're all fixed, unescaped strings, they should be hash-map lookups. If the strings can be interned, so much the better (but this correction with `item in set` should already be enough).

It shouldn't be too hard to generate a `RecordForm` with 10,000 fields as a test. Now that the time complexity has been cut down from $\mathcal{O}(n^2)$ to $\mathcal{O}(n log n)$ for the case of fixed strings (right?), a test with 10,000 fields shouldn't be an unbearable pause. A test that sits and churns for hours would be as noticeable as a failed test.

agoose77:
The time complexity of my fix was still going to be unkind to large arrays; though it used a set to look-up fixed strings, it still did this `N * M` times in order to conform to the existing selection interface.

This isn't really good enough, so I've simplified the existing code to make it possible to drop down to $\mathcal{O}(n \log m)$. This is done through a new `_SpecifierMatcher` object that partitions the current column of selections into fixed-strings and patterns, for which only patterns have $\mathcal{O}(n \times m)$ (for smaller $m$).

I just need to implement union trimming support.

agoose77:
The original logic had some pruning behavior. After logical column selection by pattern matching, we have the following rules (where `Record()` indicates a record with no contents).

- `Record(Record())` prunes to `Record()`
- `Record()` (top-level) does not prune
- `Union(Record(), X)` prunes to `Union(X)`
- `Union(X)` prunes to `X`
- `Union()` prunes to `Unknown`

This means that all but the outermost record can disappear if inner fields are not selected, e.g.
```python
import awkward as ak

form = ak.forms.from_dict(
    {
        "class": "ListOffsetArray",
        "offsets": "i64",
        "content": {
            "class": "ListOffsetArray",
            "offsets": "i64",
            "content": {
                "class": "RecordArray",
                "fields": ["x"],
                "contents": [
                    {
                        "class": "ListOffsetArray",
                        "offsets": "i64",
                        "content": {
                            "class": "RecordArray",
                            "fields": ["y"],
                            "contents": ["int64"],
                        },
                    }
                ],
            },
        },
    }
)
print(form.select_columns(["x.z"]))
```
which gives
```json
{
    "class": "ListOffsetArray",
    "offsets": "i64",
    "content": {
        "class": "ListOffsetArray",
        "offsets": "i64",
        "content": {
            "class": "RecordArray",
            "fields": [],
            "contents": []
        }
    }
}
```

This PR implements a `prune_unions_and_records` flag that performs a separate pruning pass, 
```python
print(form.select_columns(["x.z"], prune_unions_and_records=False))
```
so we can see the un-pruned variant:
```json
{
    "class": "ListOffsetArray",
    "offsets": "i64",
    "content": {
        "class": "ListOffsetArray",
        "offsets": "i64",
        "content": {
            "class": "RecordArray",
            "fields": [
                "x"
            ],
            "contents": [
                {
                    "class": "ListOffsetArray",
                    "offsets": "i64",
                    "content": {
                        "class": "RecordArray",
                        "fields": [],
                        "contents": []
                    }
                }
            ]
        }
    }
}
```

Like the original code, we only prune interior forms (with respect to an outermost (maybe nested) record or union.

agoose77:
@jpivarski you've marked this as ready in your review. After your previous comment, I decided to go further and change the time complexity. Would you mind reviewing one final time?

jpivarski (approved):
It looks like `_prune_columns` avoids a walk down-and-back-up subtrees, which would improve the time complexity if trees are deep (and would reduce time, anyway). That looks like an improvement as well, and I know there are `select_columns` tests that verify correctness, so everything here looks good and ready to merge.

When you run the tests interactively, does `test_very_large_record` exhibit a noticeable pause? I guess it must be quick enough that it's not hours (or GitHub Actions would have rejected it).

agoose77:
> It looks like `_prune_columns` avoids a walk down-and-back-up subtrees, which would improve the time complexity if trees are deep (and would reduce time, anyway). That looks like an improvement as well, and I know there are `select_columns` tests that verify correctness, so everything here looks good and ready to merge.

Hmm, I'm not quite sure if I understand this point properly. `_prune_columns` will visit every layout node, and rebuilds the layout from the bottom up. This will be slower than doing everything in one pass, but I don't think it's going to be majorly problematic; all of our operations assume we can rebuild layouts fairly cheaply.

> When you run the tests interactively, does `test_very_large_record` exhibit a noticeable pause? I guess it must be quick enough that it's not hours (or GitHub Actions would have rejected it).

With 1e6 fields under instrumentation, building the form takes ~4s on my machine, and the selection `form.select_columns(form.fields)` takes ~10s. That time is dominated by setup logic like `glob.escape` and building the various tables. I think that's fine; it's fast enough.

agoose77:
I don't think my reply above will put this at odds with merging: this directly improves the existing $\mathcal{O}(n^2)$ behavior. Therefore, I'm merging :)

jpivarski:
That looks good/sounds good. I was being verbal about my interpretation of `_prune_columns`, which apparently was wrong.

agoose77:
> I was being verbal about my interpretation 

Oh, I know - and it's useful; I can get a good sense of your thought processes! :)

PR #2538, opened 2023-06-22
chore: update pyodide parts a bit

henryiii:


henryiii:
Hitting:

```
Run python3 -m pip install -r docs/requirements-wasm.txt
[13](https://github.com/scikit-hep/awkward/actions/runs/5349678787/jobs/9701234421#step:5:14)
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'docs/requirements-wasm.txt'
```

Which appears on main, too.

awkward-cpp/CMakeLists.txt
```diff
@@ -30,16 +30,13 @@ set(CMAKE_CXX_EXTENSIONS OFF)
 set(CMAKE_CXX_VISIBILITY_PRESET hidden)
 set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
 
-if(APPLE)
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
-  set(CMAKE_INSTALL_RPATH "@loader_path")
-elseif(EMSCRIPTEN)
+if(CMAKE_SYSTEM_NAME STREQUAL Emscripten)
```
agoose77, line 33: Yes! This is much better.

awkward-cpp/CMakeLists.txt
```diff
@@ -105,6 +102,7 @@ target_link_libraries(awkward PRIVATE awkward-cpu-kernels-static ${CMAKE_DL_LIBS
 target_link_libraries(awkward PUBLIC awkward-parent)
 
 # Third tier: Python modules.
+find_package(Python REQUIRED COMPONENTS Interpreter Development.Module)
```
agoose77, line 105: Doesn't `pybind11` handle this for us? We don't link against these targets in this module.

henryiii: It defaults to the old FindPythonLibs/FindPythonInterp. You can use a variable to force the new search, or just search for it first.

If you build outside of scikit-build-core and want PyPy support, it's better to use the variable.

henryiii: (this is going to matter in CMake 3.27 which removes FindPythonLibs/FindPythonInterp if you set 3.27 as your max or min version)

PR #2543, opened 2023-06-26
refactor: drop unused `if`

### LINKED ISSUE #2457: `NumpyArray` should not be string-like
### Version of Awkward Array

main

At present, we have [code that implies `NumpyArray` can be a string](https://github.com/scikit-hep/awkward/blob/f3ce29e0a6014ebbce8a722826abc6fc7ae14791/src/awkward/contents/numpyarray.py#L665) (list of char/byte). Whilst this is only a single line of code, it makes me realise that we don't have a clear policy here. Without the ability to store parameters on a child content, we would need to implement this _at present_ with special logic on `NumpyArray` to add these parameters when e.g. pulling out a string. 

I don't think this is necessary; 2D `NumpyArray`s are only really useful if you have a non-contiguous layout, as it defers the need to render it contiguous. Given that strings are already converted into `ListArray` layouts with `from_numpy`, I think we could impose the restriction that `NumpyArray` cannot have `__array__ = "string"`, and simply require that a `RegularArray` be used for such a purpose.

Thoughts? @jpivarski

agoose77:
Closes #2457

### DIFF
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -674,9 +674,7 @@ class NumpyArray(Content):
 
     def _numbers_to_type(self, name, including_unknown):
         if (
-            self.parameter("__array__") == "string"
-            or self.parameter("__array__") == "bytestring"
-            or self.parameter("__array__") == "char"
+            self.parameter("__array__") == "char"
             or self.parameter("__array__") == "byte"
         ):
             return self


jpivarski (approved):
Yes, at the level of `NumpyArray`, the `__array__` parameter can't be `"string"` or `"bytestring"` without the array being invalid.

This is a simple one; I'll merge it.

Issue #1682, opened 2022-09-07
Remove special-casing for strings


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

The internals of Awkward have some hard-coded logic for handling strings (and categoricals). We ought to extend our behaviour system so that it is not necessary to special-case these types.

Examples:
https://github.com/scikit-hep/awkward/blob/1ebf1dbe732b6b634fab95c8406498a59b71695f/src/awkward/_v2/contents/listoffsetarray.py#L830-L840
https://github.com/scikit-hep/awkward/blob/1ebf1dbe732b6b634fab95c8406498a59b71695f/src/awkward/_v2/contents/listoffsetarray.py#L1203-L1247

etc.

agoose77:
@jpivarski suggested adding an `is_string` helper function, which special cases string fast-paths. We can build-out a more general mechanism, which we can then disable down the road.

jpivarski:
For name-collisions: there's a function in both Awkward and Uproot's _util called `isstr`, parallel with `isint`, which checks to see if an object is as Python `str`. `isstr` can be retired/removed, since its purpose was to handle Python 2 vs 3 strings, though that means turning a lot of `ak._util.isstr(x)` into `isinstance(x, str)` (and `x` can be complex). Incidentally, we'll still need `isint` because it standardizes _which_ integers we accept (no booleans but yes NumPy integers).

Since `is_string` always applies to things that have parameters, maybe it could be a method on the Metadata superclasses we talked about? And since we're talking about strings-via-parameters, maybe name it `parameter_is_string` (`string`, `bytestring`, `anystring`, with `anystring` being the most popular, I think).

agoose77:
Just updating with my latest ideas.

I think about this occasionally, and I recognise that it will be tricky to implement. At present, we use the forms to test things like `purelist_depth`. Any PR that wants to make string handling less special-cased would likely need to make forms more aware of behaviours, either through an initialiser argument (harming serialisation), or remove the ability to ask a form for its depth without providing the behaviour, e.g. `form.purelist_depth(behavior)`.

I can definitely see value in someone limiting the true depth of an array via behaviours, but it seems non-trivial to get there from where we are now.

agoose77:
Closed by #2547 (we went the other way, and made strings internal!)

PR #2547, opened 2023-06-27
feat: move string features into core

agoose77:
This PR tackes #1682 by moving existing behavior-based string logic into the core Awkward routines.

- [x] Adds `__bytes__` to `ak.Array`, to replace the `ByteBehavior.__bytes__`
- [x] Reworks `__getitem__` bypass logic
- [x] Removes `__add__` and `__radd__` support for `chat`/`byte` arrays.
- [x] Hardcodes ufunc handling for strings
- [x] Hardcodes typestr handling for strings / chars

### DIFF
--- a/docs/user-guide/how-to-convert-python.md
+++ b/docs/user-guide/how-to-convert-python.md
@@ -340,12 +340,6 @@ ak.Array(["one", "two", "three", "four"]) == ak.Array(
 
 (Without this overloaded behavior, the string comparison would yield `[True, True, True]` for `"one" == "one"` and would fail to broadcast `"three"` and `"thirty three"`.)
 
-Special behaviors for strings are implemented using the same {data}`ak.behavior` mechanism that you might use to give special behaviors to Arrays and Records.
-
-
-```{code-cell} ipython3
-ak.behavior[np.equal, "string", "string"]
-```
 
 The fact that strings are really just variable-length lists is worth keeping in mind, since they might behave in unexpectedly list-like ways. If you notice any behavior that ought to be overloded for strings, recommend it as a [feature request](https://github.com/scikit-hep/awkward-1.0/issues/new?assignees=&labels=feature&template=feature-request.md&title=).
 
--- a/src/awkward/_connect/avro.py
+++ b/src/awkward/_connect/avro.py
@@ -292,6 +292,7 @@ class ReadAvroFT:
                     parameters={"__array__": "char"},
                     form_key=f"node{form_next_id+1}",
                 ),
+                parameters={"__array__": "string"},
                 form_key=f"node{form_next_id}",
             )
             declarations.append(f"output node{form_next_id+1}-data uint8 \n")
@@ -793,6 +794,7 @@ class ReadAvroFT:
                         parameters={"__array__": "char"},
                         form_key=f"node{form_next_id+2}",
                     ),
+                    parameters={"__array__": "string"},
                     form_key=f"node{form_next_id+1}",
                 ),
                 parameters={"__array__": "categorical"},
--- a/src/awkward/_prettyprint.py
+++ b/src/awkward/_prettyprint.py
@@ -4,7 +4,6 @@ import numbers
 import re
 
 import awkward as ak
-from awkward._layout import wrap_layout
 from awkward._nplikes.numpy import Numpy
 
 numpy = Numpy.instance()
@@ -33,38 +32,6 @@ def alternate(length):
 
 
 is_identifier = re.compile(r"^[A-Za-z_][A-Za-z_0-9]*$")
-
-
-# avoid recursion in which ak.Array.__getitem__ calls prettyprint
-# to form an error string: private reimplementation of ak.Array.__getitem__
-
-
-def get_at(data, index):
-    out = data._layout._getitem_at(index)
-    if isinstance(out, ak.contents.NumpyArray):
-        array_param = out.parameter("__array__")
-        if array_param == "byte":
-            return ak._util.tobytes(out._raw(numpy))
-        elif array_param == "char":
-            return ak._util.tobytes(out._raw(numpy)).decode(errors="surrogateescape")
-    if isinstance(out, (ak.contents.Content, ak.record.Record)):
-        return wrap_layout(out, data._behavior)
-    else:
-        return out
-
-
-def get_field(data, field):
-    out = data._layout._getitem_field(field)
-    if isinstance(out, ak.contents.NumpyArray):
-        array_param = out.parameter("__array__")
-        if array_param == "byte":
-            return ak._util.tobytes(out._raw(numpy))
-        elif array_param == "char":
-            return ak._util.tobytes(out._raw(numpy)).decode(errors="surrogateescape")
-    if isinstance(out, (ak.contents.Content, ak.record.Record)):
-        return wrap_layout(out, data._behavior)
-    else:
-        return out
 
 
 def custom_str(current):
@@ -107,14 +74,14 @@ def valuestr_horiz(data, limit_cols):
             return 2, front + back
 
         elif len(data) == 1:
-            cols_taken, strs = valuestr_horiz(get_at(data, 0), limit_cols)
+            cols_taken, strs = valuestr_horiz(data._getitem(0), limit_cols)
             return 2 + cols_taken, front + strs + back
 
         else:
             limit_cols -= 5  # anticipate the ", ..."
             which = 0
             for forward, index in alternate(len(data)):
-                current = get_at(data, index)
+                current = data._getitem(index)
                 if forward:
                     for_comma = 0 if which == 0 else 2
                     cols_taken, strs = valuestr_horiz(current, limit_cols - for_comma)
@@ -187,7 +154,7 @@ def valuestr_horiz(data, limit_cols):
                 which += 1
 
                 target = limit_cols if len(fields) == 1 else half(limit_cols)
-                cols_taken, strs = valuestr_horiz(get_field(data, key), target)
+                cols_taken, strs = valuestr_horiz(data._getitem(key), target)
                 if limit_cols - cols_taken >= 0:
                     front.extend(strs)
                     limit_cols -= cols_taken
@@ -244,7 +211,7 @@ def valuestr(data, limit_rows, limit_cols):
         front, back = [], []
         which = 0
         for forward, index in alternate(len(data)):
-            _, strs = valuestr_horiz(get_at(data, index), limit_cols - 2)
+            _, strs = valuestr_horiz(data._getitem(index), limit_cols - 2)
             if forward:
                 front.append("".join(strs))
             else:
@@ -287,9 +254,7 @@ def valuestr(data, limit_rows, limit_cols):
                         key_str = key_str[1:]
                 else:
                     key_str = key + ": "
-            _, strs = valuestr_horiz(
-                get_field(data, key), limit_cols - 2 - len(key_str)
-            )
+            _, strs = valuestr_horiz(data._getitem(key), limit_cols - 2 - len(key_str))
             front.append(key_str + "".join(strs))
 
             which += 1
--- a/src/awkward/_prettyprint.py
+++ b/src/awkward/_prettyprint.py
@@ -4,6 +4,7 @@ import numbers
 import re
 
 import awkward as ak
+from awkward._layout import wrap_layout
 from awkward._nplikes.numpy import Numpy
 
 numpy = Numpy.instance()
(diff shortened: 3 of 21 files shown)


agoose77:
From #873, whilst the `__iter__` and `__getitem__` logic on _string_ arrays will give us Python strings, the same operations on arrays of `char` do not. 

The reason for this is that #873 implements the `uint8[] → str` transformation in `ak.Array.__getitem__`. We can't do the same for `char`, because the `NumpyArray._getitem_at` operation on `char` returns an integer, with no metadata about the fact that it is a unicode code point. 

The "obvious" solution is to make `NumpyArray._getitem_at` aware of `char`/`byte`, and return the Python type. But, for symmetry reasons that makes me wonder whether `ListOffsetArray` should do the same for strings, and the existing logic on `ak.Array` be removed.

I can't clearly see what the risks of doing this are. My hunch is that it would be fine; anything interacting with layouts on a per-item basis can pay the cost of string conversion. But, your thoughts are welcome here @jpivarski

jpivarski:
I think you should go with the "obvious" solution of adding the `parameter` check in `NumpyArray._getitem_at`, just as it is in `ListOffsetArray._getitem_at`. We can say that the reason the check is duplicated is for performance: putting together a string by repeatedly calling `NumpyArray._getitem_at` and having it repeatedly check the `parameters` would be a waste when it can be done once at the list level.

The weirdness of having lists check for stringiness and return a `str` and also having list contents check for stringiness and return a `str` is due to Python: the characters of a string are not a different type from the string itself. Formally, you'd expect them to be different, as they are in C (`char` vs `char*`) or C++ (`char` vs `std::string`) or Java (`char` vs `java.lang.String`). The weirdness is evident in that this doesn't terminate or raise an error:

```python
def iterate_over(thing):
    for x in thing:
        print(x)
        iterate_over(x)

iterate_over("a string")
```

But to comply with Pythonic expectations, both list types and `NumpyArray` should check `parameters` and return a `str`.

agoose77:
> I think you should go with the "obvious" solution of adding the `parameter` check in `NumpyArray._getitem_at`, just as it is in `ListOffsetArray._getitem_at`. We can say that the reason the check is duplicated is for performance: putting together a string by repeatedly calling `NumpyArray._getitem_at` and having it repeatedly check the `parameters` would be a waste when it can be done once at the list level.
> 
> The weirdness of having lists check for stringiness and return a `str` and also having list contents check for stringiness and return a `str` is due to Python: the characters of a string are not a different type from the string itself. Formally, you'd expect them to be different, as they are in C (`char` vs `char*`) or C++ (`char` vs `std::string`) or Java (`char` vs `java.lang.String`). The weirdness is evident in that this doesn't terminate or raise an error:
> 
> ```python
> def iterate_over(thing):
>     for x in thing:
>         print(x)
>         iterate_over(x)
> 
> iterate_over("a string")
> ```
> 
> But to comply with Pythonic expectations, both list types and `NumpyArray` should check `parameters` and return a `str`.

Could you clarify this slightly? The `ListOffsetArray` (et al.) and `NumpyArray` cater to different types; `char` is set on `NumpyArray`, whilst `string` is set on `ListOffsetArray`. Whilst it's hard to get a bare `NumpyArray(..., parameters={"__array__": "char"})`, it's not impossible (`ak.Array(ak.to_layout("hello"))`), and it is also trivial to pull out a character from a string (`ak.Array(["hello"])[0, 0]`). Pulling out characters and whole strings should both involve casting to a Python type (str, and str again!). 

The question for me is whether this should always happen in `Content._getitem_at`, rather than only enforced at the highlevel. We can't actually enforce the `char` logic at highlevel for the reasons outlined above.

jpivarski:
Oh, that's right: it was handled in `ak.Array.__getitem__` (high-level) because there are multiple `Content._getitem` paths that can get you to a string or char: direct `_getitem_at` and indirect `_getitem_next`. We don't want one to return a length-1 `str` and the other to return an `np.uint8`.

Yeah, this test only happens at high-level:

https://github.com/scikit-hep/awkward/blob/f2783ba480b5f9f668b687e327d0a5a453b82e5b/src/awkward/highlevel.py#L950-L958

and not at low-level:

https://github.com/scikit-hep/awkward/blob/f2783ba480b5f9f668b687e327d0a5a453b82e5b/src/awkward/contents/listoffsetarray.py#L290-L303

https://github.com/scikit-hep/awkward/blob/f2783ba480b5f9f668b687e327d0a5a453b82e5b/src/awkward/contents/numpyarray.py#L295-L303

-------------

```python
>>> del ak.behavior["char"]
>>> ak.Array(ak.to_layout("hello"))
<Array [104, 101, 108, 108, 111] type='5 * char'>
>>> ak.Array(ak.to_layout("hellothere"))[0]
104
```

So the problem is conveying the information that the final result of the slice is a char from the last step of `Content.__getitem__`, which could be `_getitem_at`, `_getitem_next`, or (maybe?) something else, to the `ak.Array.__getitem__` routine, which does not deeply inspect the `Content`'s `__getitem__` machinery, just the final result. By the time it gets a `np.uint8`, the information about it being a char is gone.

Well, `NumpyArray._getitem_at`, `NumpyArray._getitem_next`, etc. could return `np.character`...?

-------

No: here's a better idea. Note that Awkward slicing does not make a distinction between string and bytestring: if you index the contents of a Unicode string, your indexes are not counting Unicode characters; they're counting bytes. That was a decision to keep the slicing rules simple; a Unicode-aware codepoint-index slice could be implemented in a future strings module. (It would need compiled routines to get all the variable-width UTF-8 rules right.)

```python
>>> money = ak.Array(["$", "¢", "€", "💰"])

>>> ak.num(money)
<Array [1, 2, 3, 4] type='4 * int64'>

>>> money.layout
<ListOffsetArray len='4'>
    <offsets><Index dtype='int64' len='5'>
        [ 0  1  3  6 10]
    </Index></offsets>
    <content><NumpyArray dtype='uint8' len='10'>
        [ 36 194 162 226 130 172 240 159 146 176]
    </NumpyArray></content>
</ListOffsetArray>

>>> money[0, 0]
36
>>> money[1, 0]
194
>>> money[1, 1]
162
>>> money[2, 0]
226
>>> money[2, 1]
130
>>> money[2, 2]
172
>>> money[3, 0]
240
>>> money[3, 1]
159
>>> money[3, 2]
146
>>> money[3, 3]
176
```

Considering that `ak.Array.__getitem__` into a string _does not give you Unicode characters_, we can't expect it to give you a length-1 string. It's showing you _part_ of a UTF-8 encoding.

So this could just be a policy decision: we always show individual-item selections of a string as integers. That's what Python 3 users expect of `bytes` but not `str`:

```python
>>> list("$¢€💰")
['$', '¢', '€', '💰']

>>> list("$¢€💰".encode("utf-8"))
[36, 194, 162, 226, 130, 172, 240, 159, 146, 176]
```

The policy can be that Awkward returns integers from both bytestrings and strings because it's not really selecting characters, anyway. I see a stronger case that this is the _right_ thing to do, though it's not obvious at first and we may find ourselves explaining it as a gotcha.

In other words,

```python
>>> del ak.behavior["char"]

>>> ak.Array(ak.to_layout("hello"))
<Array [104, 101, 108, 108, 111] type='5 * char'>

>>> ak.Array(ak.to_layout(["hello", "there"]))
<Array ['hello', 'there'] type='2 * string'>
```

is fine as-is. What do you think of that?

agoose77:
> is fine as-is. What do you think of that?

In general, I strongly prefer this. I'd really like to drop as much magic from these routines as possible, so that we can more easily reason about what's going on here!

I was looking to modify `NumpyArray.to_list()` to return integers instead of bytestrings for `char`, so this all fits in very nicely.

src/awkward/_connect/avro.py
```diff
@@ -292,6 +292,7 @@ def rec_exp_json_code(
                     parameters={"__array__": "char"},
                     form_key=f"node{form_next_id+1}",
                 ),
+                parameters={"__array__": "string"},
```
agoose77, line 295: Avro deserialiser was missing `"string"` names

agoose77:
OK, the latest iteration is thus:
- `to_list` still converts `char` arrays to a single string.
- `N * char` → `str` conversion still happens in `Array.__getitem__` & `Array.__iter__` (we want to be able to access the low-level `char` nodes at low-level, e.g. in `from_iter` with strings).
- `ak.prettyprint` renders `char` arrays as `[101, 93, 50, ...]` instead of `"e]2..."`. This is to ensure that parameters are still visible. If we _really_ want this to render like a string, we can.
- `char` elements are always returned as integers, because they may be partical codepoints

jpivarski:
Having char _arrays_ convert to `str`/`bytes` sounds like a good idea, and it's probably implemented that way either because your or I expected it or someone asked for it.

_Individual items_ pulled from a char array should be integers, because they could be a partial codepoint.

The only odd one out is the pretty-print of a char array as a list of integers. Any alternative would be a very special case, since it would be a pretty-printed array that doesn't begin and end with `[` and `]`. I'd say we don't need to do that unless asked, when asked. Changing `repr` is not what I would consider a backward-incompatible change, since no one should be depending on the formatting of a `repr` (Python anti-pattern).

src/awkward/operations/ak_to_list.py
```diff
@@ -77,7 +78,7 @@ def _impl(array):
     elif isinstance(array, Mapping):
         return {k: _impl(v) for k, v in array.items()}
 
-    elif isinstance(array, Iterable):
+    elif is_non_string_like_iterable(array):
```
agoose77, line 81: Fix for passing iterable str!

agoose77:
I believe this is good to go. Overall, it removes a feature: the ability to add two `char` / `byte` arrays. It isn't very easy to identify this from existing repo usage given, but my feeling is that this is very niche. Due to `Array._getitem`, it should be difficult to obtain a bare char array unless one performs `ak.to_layout("hello world")`.

docs/user-guide/how-to-convert-python.md
```diff
@@ -340,12 +340,6 @@ ak.Array(["one", "two", "three", "four"]) == ak.Array(
 
 (Without this overloaded behavior, the string comparison would yield `[True, True, True]` for `"one" == "one"` and would fail to broadcast `"three"` and `"thirty three"`.)
 
-Special behaviors for strings are implemented using the same {data}`ak.behavior` mechanism that you might use to give special behaviors to Arrays and Records.
-
-
-```{code-cell} ipython3
-ak.behavior[np.equal, "string", "string"]
-```
```
agoose77, line 348: Now that strings are built-in, there's no exception here.

agoose77:
@jpivarski could you check whether I've added the typer and lowering in the best place? It seems to wrap the getitem result, but I don't know enough about what to expect here!

jpivarski:
> @jpivarski could you check whether I've added the typer and lowering in the best place? It seems to wrap the getitem result, but I don't know enough about what to expect here!

While talking about this on Zoom, I looked into it, and I think that checking at the level of `ContentType.getitem_at_check` is fine. The string check _could_ be done only in the list types, but going to the extra work of changing it to do that is unnecessary. Upon entering a Numba context, we assume that the array is fully valid; `__array__: string` would only be on list-types, so the check on all other types would always return `False`.

src/awkward/_connect/avro.py
jpivarski: I'm not surprised—this isn't an unlikely mistake. (We didn't spend much time on this module because it was a stepping-stone to the real problem.)

src/awkward/_prettyprint.py
```diff
@@ -35,38 +34,6 @@ def alternate(length):
...
-    if isinstance(out, ak.contents.NumpyArray):
-        array_param = out.parameter("__array__")
-        if array_param == "byte":
-            return ak._util.tobytes(out._raw(numpy))
-        elif array_param == "char":
-            return ak._util.tobytes(out._raw(numpy)).decode(errors="surrogateescape")
-    if isinstance(out, (ak.contents.Content, ak.record.Record)):
-        return wrap_layout(out, data._behavior)
-    else:
-        return out
```
jpivarski: Why is it okay to remove these? I think the problem they were addressing is that there was an error in constructing the error message, which is not easily tested in automated tests.

Also, they avoid the rather complex dispatch of `Content._getitem`, which checks for any kind of slice, with `Content._getitem_at` or `Content._getitem_field`. That's not (just) a performance thing, but it also avoids the possibility of errors in that dispatch, which can be a problem if  it needs to print the array when there's an error.

jpivarski (approved):
All good, except for the removal of `get_at` and `get_field` in prettyprint.py. I think we might still need these, so that a simple error doesn't explode into a `RecursionError`.

PR #2548, opened 2023-06-27
fix: support dict-likes in form.from_dict, convert fields to list

agoose77:
`ak.forms.RecordForm` accepts an iterable for `fields`, but doesn't canonicalise it. This PR addresses that, and also relaxes the `dict` requirement in `forms.from_dict`. This just seems reasonable; it will be a bit slower than checking dict, but I'm hoping this isn't a performance critical path.


jpivarski (approved):
Both are good, and this is pretty simple, so I'll merge it.

PR #2549, opened 2023-06-29
feat: add new `__list__` parameter for list-based array types

agoose77:
This PR adds a new `__list__` nominal type parameter to associate with behaviors. It should only be added to lists. In addition to setting an array-class for `ListOffsetArray` (et al.) via `layout.parameter("__list__")`, we can also lookup the list in a purelist sense, allowing list-of-type to have an array class too.

Setting `__list__` and `__record__` an error for non-intended types, but that should be another PR (there are fixes that need to be made to both `validity_error` and broadcasting).

### DIFF
--- a/docs/reference/ak.behavior.md
+++ b/docs/reference/ak.behavior.md
@@ -520,9 +520,7 @@ First, one must define the behavior class
 ```python
 class ReversibleArray(ak.Array):
     def reversed(self):
-        index = ak.local_index(self)
-        reversed_index = index[..., ::-1]
-        return self[reversed_index]
+        return self[..., ::-1]
 ```
 To make this behavior class available to new arrays, it must be associated with its name
 ```python
--- a/docs/reference/ak.behavior.md
+++ b/docs/reference/ak.behavior.md
@@ -510,41 +510,46 @@ One may wish to disable this by type-checking, since the functionalities are rat
 ## Adding behavior to arrays
 
 Occasionally, you may want to add behavior to an array that does not contain
-records. A good example of this is to implement strings: strings are not a
-special data type in Awkward Array as they are in many other libraries, they
-are a behavior overlaid on arrays.
+records. Historically, this mechanism was used to help implement strings (although)
+strings have since been made a part of Awkward's internals, alongside categorical types.
 
-There are four predefined string behaviors:
+Awkward Array supports adding behaviors to the list-types in an array. Let's suppose that 
+one wishes to define a special list that defines a `reversed()` method, equivalent to `array[..., ::-1]`.
 
-- {class}`ak.CharBehavior`: an array of UTF-8 encoded characters;
-- {class}`ak.ByteBehavior`: an array of unencoded characters;
-- {class}`ak.StringBehavior`: an array of variable-length UTF-8 encoded strings;
-- {class}`ak.ByteStringBehavior`: an array of variable-length unencoded bytestrings.
-
-All four override the string representations (`__str__` and `__repr__`),
-but the string behaviors additionally override equality:
-
+First, one must define the behavior class
 ```python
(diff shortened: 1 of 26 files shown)


agoose77:
The addition of the new `__list__` nominal type parameter provides the ability to set `__typestr__` for a list type. However, unlike records which can sensibly be considered atoms, lists are usually part of a deeper structure that we care about seeing the type of. Having
`N * my-list` instead of `N * my-list * int64` is worse. I wonder if we should think about this before merging this PR.

docs/reference/ak.behavior.md
```diff
@@ -510,41 +510,46 @@ One may wish to disable this by type-checking, since the functionalities are rat
...
 
+First, one must define the behavior class
 ```python
->>> ak.Array(["one", "two", "three"]) == ak.Array(["1", "TWO", "three"])
-<Array [False, False, True] type='3 * bool'>
+class ReversibleArray(ak.Array):
+    def reversed(self):
+        index = ak.local_index(self)
+        reversed_index = index[..., ::-1]
+        return self[reversed_index]
```
jpivarski: This can also be implemented as

```python
def reversed(self):
    return self[..., ::-1]
```

but the original is fine: it's some computation that one might want to do—it doesn't matter what that computation is.

src/awkward/_parameters.py
```diff
@@ -4,6 +4,8 @@
 
 from awkward._typing import JSONMapping, JSONSerializable
 
+TYPE_PARAMETERS = ("__array__", "__list__", "__record__", "__categorical__")
```
jpivarski, line 7: :heavy_check_mark:

jpivarski (approved):
Very nice! I like the new documentation examples, too.

Since this PR is more complicated and you might think of something you need to add at the last minute, I'll let you merge it. I think it's done.

docs/reference/ak.behavior.md
agoose77: Indeed, I hesitated on adding this. Originally I was going to add a string-related method that inverted the character order. I'm going to change this because it is not an important part of the messaging.

agoose77:
```suggestion
    def reversed(self):
        return self[..., ::-1]
```

agoose77:
```suggestion
    def reversed(self):
        return self[..., ::-1]
```

jpivarski:
> The addition of the new `__list__` nominal type parameter provides the ability to set `__typestr__` for a list type. However, unlike records which can sensibly be considered atoms, lists are usually part of a deeper structure that we care about seeing the type of. Having `N * my-list` instead of `N * my-list * int64` is worse. I wonder if we should think about this before merging this PR.

Sorry that I missed this question!

A `__typestr__` for `__list__` would be useful because a common reason to override a list would be to define an opaque binary blob. The developer of that library would want the type of the hidden data to show up as "`my-list`", _without_ any hint about the contents. So, for instance,

```
1000 * var * PNG-image
```

for an array of lists of overloaded bytestrings that are interpreted as PNG images.

The `__typestr__` was introduced to make it possible to label strings as `string` and adhere to Datashape, as part of the goal of making string entirely user-definable. I think that was already gone at the time when Awkward 1 was converted to Awkward 2.

Since then, the `__typestr__` was useful for overloaded records, in which the library developer wanted to hide the details of the record structure. (I think this was in some version of Coffea.)

The bytestring use-case (I can't think of any others, involving `__list__`) would also want to hide the internal structure of the list, for the same reason. In fact, that was also true when `__typestr__` was set to `"string"`.

So yes, I think it would be a good addition, but it can be applied in the same way that it is for `__record__`, which is simpler than trying to let the inner contents through (or parameterize/templatize `__typestr__` in some way).

PR #2550, opened 2023-06-29
fix: check validity error properly

agoose77:
The parameters check for validity error does not recurse in the existing implementation

### DIFF
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -349,9 +349,6 @@ def reduce(
 
 
 def validity_error(layout: Content, path: str = "layout") -> str:
-    paramcheck = layout._validity_error_parameters(path)
-    if paramcheck != "":
-        return paramcheck
     return layout._validity_error(path)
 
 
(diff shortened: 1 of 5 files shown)


jpivarski (approved):
I don't think it was failing to recurse before: my reading of the red/removed code is that it would only stop recursion if it hit an error, which is what it's supposed to do. (I didn't construct a test, though.)

Anyway, the new code also looks correct but it's better from an OOP point of view.

This is a simple PR and it looks like you're done; I'll merge it immediately.

agoose77:
> I don't think it was failing to recurse before: my reading of the red/removed code is that it would only stop recursion if it hit an error, which is what it's supposed to do. (I didn't construct a test, though.)



The intention was that we have cross-layout checks in a separate method. However, only `ak._do.validity_error` checked this for the top-level layout (hence the test). We might still want a dedicated `ak.Content._validity_error_parameters` function in future (if we have such checks), but this is L3 and we don't currently need it.

PR #2552, opened 2023-06-30
fix: string broadcasting

agoose77:
There are still some outstanding problems with string broadcasting, that pertain to parameter replacement in `broadcast_and_apply`. 

The main cases are covered [in the tests](https://github.com/scikit-hep/awkward/blob/agoose77/fix-string-broadcasting/tests/test_2410_string_broadcast.py) as `xfail`.

This PR fixes the broadcasting rules for parameters, with the following changes:

1. New list-types are not included in the parameter broadcasting, using a sentinel "NO_PARAMETERS" token
2. Record and union broadcasting considers only the parameters of the record/union layouts respectively; the non-record/non-union contents are nested in the records/unions that are build.


src/awkward/_broadcasting.py
```diff
@@ -194,12 +194,7 @@ class BroadcastParameterRule(str, enum.Enum):
     NONE = "none"
 
 
-BroadcastParameterFactory: TypeAlias = Callable[
-    [int], List[Union[Dict[str, Any], None]]
-]
-
-
```
agoose77, line 201: We need to change the input parameters for each closure, so there's no benefit to this two-phase design any more.

src/awkward/_broadcasting.py
```diff
@@ -228,11 +224,9 @@ def all_or_nothing_parameters_factory(
     parameters are repeated, i.e. `[parameters, parameters, ...]`. Otherwise, a list
     of Nones is returned, i.e. `[None, None, ...]`.
     """
-    input_parameters = [
-        p for p in (_parameters_of(c) for c in inputs) if p is not NO_PARAMETERS
-    ]
```
agoose77, line 233: As we need to signal intent as well as parameter contents, we pass in the parameters directly. This means that the caller can pass `NO_PARAMETERS` which behaves like a no-op placeholder.

src/awkward/_broadcasting.py
```diff
@@ -308,28 +298,23 @@ def one_to_one_parameters_factory(
...
-                "cannot follow one-to-one parameter broadcasting rule for actions "
-                "which change the number of outputs."
-            )
-        return input_parameters
-
-    return apply
+    if n_outputs == len(parameters):
+        return [p if p is not NO_PARAMETERS else None for p in parameters]
+    else:
+        return intersection_parameters_factory(parameters, n_outputs)
```
agoose77, line 310: New default behaviour: perform the intersection when we don't have the same number of inputs as outputs

agoose77:
This PR is done! We will possibly need to revisit this again in future; broadcasting is a fiddly thing, but it's definitely an improvement, and removes some `xfail`s!

jpivarski (approved):
These are the right tests to make work. I had some comments on the implementation until I noticed that I was wrong, and deleted them.

It all looks good! Go ahead and merge when you're ready.

PR #2554, opened 2023-06-30
chore: move C++ format configuration

agoose77:
#2553 will also require a C++ release, so this PR can go in at the same time.


header-only/CMakeLists.txt
```diff
@@ -1,4 +1,5 @@
-# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+# BSD 3-Clause License; see
+# https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
```
jpivarski: I don't like the license being multi-line, even if it is a long line, but it looks like this is happening exactly once, in CMakeLists only (instead of every single C++ file). That's fine.

Actually, I'll try to see if it fits with 4 fewer characters.

```suggestion
# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
```

jpivarski (approved):
Sure, CMakeLists formatting.

I'm going to try the suggested commit below, to see if it's narrow enough for the line length.

agoose77:
Taking your "it's fine" as tacit approval :)

PR #2557, opened 2023-07-02
chore: drop Python 3.7

agoose77:
[Our deprecation schedule](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20) earmarked the July 2.3.0 release for dropping Python 3.7 support. This PR updates our CI, and package metadata, to reflect this change.

### DIFF
--- a/.github/workflows/packaging-test.yml
+++ b/.github/workflows/packaging-test.yml
@@ -59,7 +59,7 @@ jobs:
     - uses: pypa/cibuildwheel@v2.13.1
       env:
         CIBW_ARCHS_MACOS: universal2
-        CIBW_BUILD: cp39-win_amd64 cp37-manylinux_x86_64 cp38-macosx_universal2
+        CIBW_BUILD: cp39-win_amd64 cp310-manylinux_x86_64 cp38-macosx_universal2
       with:
         config-file: cibuildwheel.toml
         package-dir: awkward-cpp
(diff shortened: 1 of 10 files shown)


agoose77:
@henryiii I'd be interested in your thoughts about dropping lower bounds on dependencies that are already implied by the Python version. In this case, the NumPy lower bound is technically 1.18, but that's already imposed by the Python constraint of 3.8. What do you think about this, is it reasonable to remove the LB, or does it remove useful information?

jpivarski (approved):

agoose77:
@jpivarski did you have any thoughts RE removing the lower bound on NumPy (as we can use the minimum supported Python version as our LB)? I'm not aware of any problems in the wild, but wanted to see if anyone else has concerns.

jpivarski:
I noticed that, but I understood why it was happening: Python version ranges now _imply_ NumPy version ranges, so an explicit constraint on NumPy version is redundant.

In the wild, one could circumvent dependency managers when installing these packages. If they then run into an issue with NumPy not supporting NEP-13 or NEP-18, then they brought that problem upon themselves.

henryiii:
It does still help the solver a bit in rare cases (no need to download earlier versions to check compatibility), and has a small user readability benefit, but no strong opinion. If' you'd like to remove it (say to highlight that you support everything valid), that' fine too.

agoose77:
OK, given both of your feedback, I'll err on the side of being explicit here. It might make our lives slightly easier in future if our supported Python deprecation schedule changes relative to NumPy's.

PR #2559, opened 2023-07-02
fix: allow JAX tracers in `from_arraylib`

### LINKED ISSUE #2556: Jax tracer TypeError in multiplication
### Version of Awkward Array

2.2.4

While trying out awkward + jax I ran into the following:
```python
import jax
import awkward as ak

ak.jax.register_and_check()

a = ak.Array([[1.0, 2, 3], [5, 6]], backend="jax")

def f(x):
    return ak.sum(ak.sum(x) * x)

f(a)
jax.grad(f)(a)
```
results in
```python
[...]
TypeError: Jax tracers cannot be used with `ak.from_arraylib`

    numpy.multiply.__call__(
        JVPTracer-instance
        <Array [[...], [...]] type='2 * var * float32'>
    )
```
This seems to originate from the `ak.sum(x) * x` piece.

A corresponding `numpy` version (without any `awkward`) works fine:
```python
import jax
import numpy as np

a = np.array([1, 2, 3], dtype=float)

def f(x):
    return np.sum(np.sum(x) * x)

f(a)
jax.grad(f)(a)
```

Relevant library versions:
```
awkward                       2.2.4
awkward-cpp                   17
jax                           0.4.13
jaxlib                        0.4.13
numpy                         1.25.0
```

agoose77:
This should fix #2556, though I'm not sure why we had that check there. It looks intentional. Let's see what the test suite picks up.

### DIFF
--- a/src/awkward/_layout.py
+++ b/src/awkward/_layout.py
@@ -7,7 +7,6 @@ from awkward._backends.numpy import NumpyBackend
 from awkward._behavior import behavior_of
 from awkward._errors import AxisError
 from awkward._nplikes.dispatch import nplike_of
-from awkward._nplikes.jax import Jax
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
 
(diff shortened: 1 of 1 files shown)


agoose77:
So, the check that this PR removes was added in https://github.com/scikit-hep/awkward/pull/1763

I'm not totally clear on what the right solution is here. I think we probably want to allow tracers to be used as buffers for transient layouts, but ensure that the result is tracer free. As such, this PR needs a bit more time.

PR #2560, opened 2023-07-02
fix: convert unknown-length to index in `ListOffsetArray.to_RegularArray`

agoose77:
This fixes a bug that is raised during the strict dask-awkward test suite.


jpivarski (approved):
This makes sense and it looks very straightforward.

PR #2561, opened 2023-07-03
fix: prefer known to unknown lengths in broadcasting

agoose77:
This PR is a first pass over the broadcasting logic to change direction from "unknown lengths are infectious" to "known lengths are infectious". 

As we've discussed here previously, typetracer should only fail for ahead-of-time known errors. We defer validation of the unknown data to runtime, via a second pass using a known-data backend. As such, we can rewrite assertions of the form `assert unknown_value == x` with `unknown_value = x`, rather than propagating unknown values everywhere.

i.e. operations like
```python
require_equal_lengths(contents)
next_length = unknown_length
```
become
```python
require_equal_lengths(contents)
next_length = contents[0].length
```

I think we might have assumed this before; I'm just getting around to changing the code after  re-orienting my thinking a while back. 

Relatedly, slicing typetracer arrays should assume that the length succeeds, and use the concrete length, for obvious reason.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -66,8 +66,7 @@ def length_of_broadcast(inputs: Sequence) -> int | type[unknown_length]:
             if x.length is unknown_length:
                 has_seen_unknown_length = True
                 continue
-            else:
-                maxlen = max(maxlen, x.length)
+            maxlen = max(maxlen, x.length)
 
     if has_seen_unknown_length:
         return unknown_length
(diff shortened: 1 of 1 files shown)


src/awkward/_broadcasting.py
```diff
@@ -413,38 +426,39 @@ def broadcast_any_record():
...
+                                ", ".join(sorted(fields)),
+                                ", ".join(sorted(x.fields)),
+                            )
                         )
-                    )
-                # Records win over tuples
-                if istuple is UNSET or not x.is_tuple:
-                    istuple = False
+                    # Records win over tuples
+                    is_tuple = False
```
agoose77: This just benefited from a re-write

src/awkward/_broadcasting.py
```diff
@@ -47,20 +47,32 @@
...
 
 
+class BroadcastOptions(TypedDict):
+    allow_records: bool
+    left_broadcast: bool
+    right_broadcast: bool
+    numpy_to_regular: bool
+    regular_to_jagged: bool
+    function_name: str | None
+    broadcast_parameters_rule: BroadcastParameterRule
```
agoose77, line 57: Add type annotations!

jpivarski (approved):
Yes, "known lengths are infectious" is the right direction: if we try to broadcast length `N` with length `unknown`, then the broadcasted result is length `N` because `unknown` _might_ be equal to `1`. If it's not, we'll find out when a Dask worker tries to actually do it with no unknown lengths.

PR #2564, opened 2023-07-03
fix: broadcasting of regular strings

agoose77:
The recent changes for string broadcasting had a missing case for regular strings. Furthermore, the broadcasting logic for lists was growing in complexity. This PR fixes bugs with broadcasting of regular strings, and merges the optimised broadcasting logic into the fallback. This reduces the number of code branches that we have to keep in sync.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -677,7 +677,6 @@ def apply_step(
             # 2. mixed string-list (strings gain a dimension and broadcast to the non-string offsets)
             # 3. no strings (all lists broadcast to a single offsets)
             offsets_content = None
-            all_content_strings = True
             input_is_string = []
             for content in inputs:
                 if isinstance(content, Content):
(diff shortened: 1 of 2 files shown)


src/awkward/_broadcasting.py
```diff
@@ -144,54 +143,13 @@ def ensure_common_length(inputs, options: BroadcastOptions) -> ShapeItem:
...
-            return False
-
-    return True
+def offset_source_rank(list_content: ak.contents.Content) -> int:
+    if isinstance(list_content, ak.contents.ListOffsetArray):
+        return 0
+    elif isinstance(list_content, ak.contents.ListArray):
+        return 1
+    else:
+        raise AssertionError
```
agoose77, line 152: This might benefit from a better name. The purpose of this function is to score list-types in terms of the preference to use their offsets. ListOffsetArray can avoid an allocation (at the cost of reading the zeroeth value) to build offsets, whilst `ListArray` cannot. This might be premature optimisation; at one point I also scored regular arrays here, but that turned out to be a mistake.

jpivarski (approved):

PR #2566, opened 2023-07-04
feat!: apply retroactive deprecations

agoose77:
- Raise error for `parameters` that are neither `None` or `{}` in `EmptyForm` et al.
  - We still allow `parameters=None` and `parameters={}` in the constructor arguments, to ensure that content-naive workflows still work.
- Raise error for conversions of unions to backend arrays

### DIFF
--- a/src/awkward/contents/emptyarray.py
+++ b/src/awkward/contents/emptyarray.py
@@ -68,7 +68,7 @@ class EmptyArray(Content):
     is_leaf = True
 
     def __init__(self, *, parameters=None, backend=None):
-        if parameters is not None:
+        if not (parameters is None or parameters == {}):
             raise TypeError(f"{type(self).__name__} cannot contain parameters")
         if backend is None:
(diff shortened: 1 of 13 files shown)


src/awkward/contents/emptyarray.py
```diff
@@ -85,29 +82,26 @@ def copy(
...
     def __copy__(self):
         return self.copy()
 
     def __deepcopy__(self, memo):
-        return self.copy(parameters=copy.deepcopy(self._parameters, memo))
+        return self.copy()
 
     @classmethod
     def simplified(cls, *, parameters=None, backend=None):
-        if not (parameters is None or len(parameters) == 0):
```
agoose77, line 105: This is a more permissive test for the deprecations, but I think it's okay to break empty dicts here too.

jpivarski (approved):
Right: this makes good on the deprecation promise of forbidding `EmptyArray` with `parameters`.

PR #2570, opened 2023-07-05
test: Make cuda tests pass

kkothari2001:


src/awkward/_kernels.py
```diff
@@ -127,7 +127,7 @@ def calc_blocks(self, length):
         return length, 1, 1
 
     def _cast(self, x, t):
-        if issubclass(t, ctypes._Pointer):
+        if t:
             # Do we have a NumPy-owned array?
```
agoose77:
```suggestion
            # Do we have a CuPy-owned array?
```

jpivarski:
For @agoose77, this is the work that @kkothari2001 did to get `tests-cuda` working again. I'll make it a draft because, despite all of our CPU-bound automated tests passing, there are still 2 failing GPU tests.

I've scheduled "Set up CUDA test-runner for Awkward" for myself on July 17, since that's when I arrive at Princeton and I can talk to the Princeton computing people directly.

src/awkward/_nplikes/cupy.py
```diff
@@ -165,3 +165,9 @@ def is_c_contiguous(self, x: ArrayLike) -> bool:
             return True
         else:
             return x.flags["C_CONTIGUOUS"]
+
+    def broadcast_arrays(self, *arrays):
+        return [
+            self._module.array(x) if not isinstance(x, self._module.ndarray) else x
+            for x in super().broadcast_arrays(*arrays)
+        ]
```
agoose77: I think this should be handled at the call site - we want arguments to nplikes to be arrays already, for simplicity.

agoose77:
@jpivarski did you manage to follow up with the Princeton compute team regarding a CUDA runner? :)

jpivarski:
I had a conversation with @vineetbansal, who has set up a GPU test on an owned machine, but not in Princeton infrastructure. If we get this working, we know there are at least two use-cases that would justify it.

@vineetbansal, you said there was someone with whom we could follow this up. Who was that?

src/awkward/_kernels.py
```diff
@@ -127,7 +127,7 @@ def calc_blocks(self, length):
         return length, 1, 1
 
     def _cast(self, x, t):
-        if issubclass(t, ctypes._Pointer):
+        if t:
```
ianna:
```suggestion
        if type:
```

ianna: the `if` statement seems a bit short to me :-)

ianna (commented):
@kkothari2001 - Nice, thanks for cleaning it up!

jpivarski:
I think @kkothari2001 has finished as much of this PR as he's going to do. (Is that right? Let me know!) We should take it the rest of the way, and it would be best for one of us with direct access to a GPU. (That sounds like me.)

We don't have a dedicated GPU runner yet. Maybe instead I could put a weekly manual test on my calendar...

agoose77:
I have a GPU and this is on my todo list for this week! Happy for you to take the lead, but I can do it!

jpivarski:
We could rock-paper-scissors for it. `:)`

If you get a chance to look at it this week, please do. Beyond that, I can take on the weekly testing as a manual task. Calendars tell me what to do nowadays, so when it pops up, I'll run it and adjust to whatever small issues there are, which will both keep me in touch with changes in nplikes/backends and act as a constant reminder to get the dedicated GPU test-runner sorted out.

jpivarski:
As an update to this, on the problem of CI testing CUDA: there are emulators. Can they be used with CuPy (the only CUDA library we use)? There are no explicit indications that it will work, and all of these emulators are old.

  * [MCUDA](http://impact.crhc.illinois.edu/mcuda.aspx)
  * [gpgpu-sim_distribution](https://github.com/gpgpu-sim/gpgpu-sim_distribution/tree/dev)
  * [coriander](https://github.com/hughperkins/coriander)
  * [gpuocelot](https://github.com/gtcasl/gpuocelot)

Or maybe we can roll our own emulator. Since we only use the GPU through CuPy and CuPy-compiled `RawKernels`, perhaps we could mock CuPy with a combination of NumPy and cppyy. (That is, remove the "`__global__`" from the source code, compile it as C++, and run it in a loop or OpenMP threads.) This would not fully test its correctness, which would have to be occasionally tested on a real GPU, but it would catch issues like the ones that led to this PR—that the infrastructure changed and new kernels were added without equivalent updates for the CUDA side. Those are the kinds of issues that would require continuous testing, anyway—the more subtle concurrency issues are either encountered at development time or when moving to a different platform (for race conditions to surface), not everyday CI.

jpivarski:
Numba has a CUDA emulator (`export NUMBA_ENABLE_CUDASIM=1`), and that could help with our Numba-CUDA backend, but not the CuPy nplike.

agoose77:
@jpivarski I've made the tests pass on my machine by `xfail`-ing tests that require the missing kernel. I think that we should merge this, and @ManasviGoyal can work in a new branch?

src/awkward/_errors.py
```diff
@@ -18,6 +19,22 @@
...
+    """A proxy for a method of a weakly referenced object"""
+
+    def __init__(self, method):
+        self._this = weak_ref(method.__self__)
+        self._impl = method.__func__
+
+    def __call__(self, *args, **kwargs):
+        this = self._this()
+        method = self._impl.__get__(this, type(this))
+        return method(*args, **kwargs)
```
agoose77, line 37: The previous solution to long-lived reference cycles was to just clear the context state after exit. However, I'd forgotten that the error context is long-lived for GPU arrays, and therefore this solution was insufficient.

This PR introduces a weakref for the error context that is held _by_ the error context. As such, it should not cause any issues.

jpivarski:
I manually tested it, and I'm going to set up a calendar reminder for me to manually test it—perpetually—until I manage to set up a routine solution.

(I found and fixed one bug that way: 50614aa87b831fd7840ebfcb36b9e0f5466989d7.)

src/awkward/_kernels.py
jpivarski: This has been fixed.

src/awkward/_nplikes/cupy.py
jpivarski: This must have been addressed somehow—I can't find any text matching "broadcast" in cupy.py.

src/awkward/_errors.py
jpivarski: Okay, that's good to know. This error-handling context has changed several times, but it was originally introduced so that kernels can more fully occupy the device (#1321).

Here's what

```python
a = np.random.uniform(5, 10, 1000000)
b = np.random.uniform(10, 20, 1000000)
c = np.random.uniform(-0.1, 0.1, 1000000)
```

looks like, computing

```python
(-b + np.sqrt(b**2 - 4*a*c)) / (2*a)
```

download

It's pretty hard to get one to run immediately after the next because they're so quick compared to the time for Python to execute an instruction.

jpivarski (approved):
> @jpivarski I've made the tests pass on my machine by `xfail`-ing tests that require the missing kernel. I think that we should merge this, and @ManasviGoyal can work in a new branch?

I think that's a great idea, and certainly much better than letting this PR get any more stale. (Thank you, @kkothari2001, for starting it 19 weeks ago, and @agoose77 for getting it into a runable state.)

Now 61 tests pass with 8 skipped (xfail), 6 of which are needing the `awkward_ListArray_broadcast_tooffsets` kernel. That can be a good first task for @ManasviGoyal in her new project!

This PR can be merged if you're done, @agoose77, and I'll be testing the `tests-cuda` directory locally at least once a week until I can find some automated solution.

agoose77:
Thanks @jpivarski! I'm just setting up a Docker (podman) image to run this in a cron job. I intend for it to file a GitHub issue if it notices that something fails. So, let's hold the merge.

jpivarski:
It passed all tests, and I'll merge it now.

jpivarski:
Sorry that I missed this important contribution in the all-contributors list, @kkothari2001!

@all-contributors please add @kkothari2001 for code, test

PR #2572, opened 2023-07-05
fix: added kernel specializations for `awkward_ListOffsetArray_reduce_local_nextparents_64`

### LINKED ISSUE #2571: `awkward_ListOffsetArray_reduce_local_nextparents_64` only supports `int64` arguments
### Version of Awkward Array

main

See https://github.com/dask-contrib/dask-awkward/issues/307

@ianna, @ioanaif, or @jpivarski - is this something you'd be able to take a look at? I'm away for the next two weeks (and preparing for my viva tonight and tomorrow), so I won't be able to get to it for a while.

The bug here is that whilst our index kernels are always specialised to `int64`, they should support the various input index types that we know of. I think we should tackle this, as there are no guards before this kernel that ensure we don't end up with an `int32` index, and I believe we should support that case. 

The fix I *think* is to change the `nextparents` call such that it doesn't cast to `int64`, and then generate the appropriate overloads for the kernel. We could cast both index and nextparents, but that's more copying vs specialising the kernel.

N.B. there may be other kernels that have this problem. If you notice them whilst fixing this, great!

jpivarski:
This includes the awkward-cpp version number bump. Before we make another awkward release, we'll need the new awkward-cpp release, but the tests will warn us about that.

### DIFF
--- a/awkward-cpp/pyproject.toml
+++ b/awkward-cpp/pyproject.toml
@@ -7,7 +7,7 @@ build-backend = "scikit_build_core.build"
 
 [project]
 name = "awkward_cpp"
-version = "20"
+version = "21"
 dependencies = [
     "numpy>=1.18.0"
 ]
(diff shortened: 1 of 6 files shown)


agoose77 (approved):

jpivarski:
Okay, thanks! (I didn't want to bother you.)

agoose77:
@jpivarski not at all! But I appreciate it.

Just making a quick change that I noticed...

jpivarski:
I think those are unrelated to this fix, but also necessary. Thanks for noticing them!

PR #2583, opened 2023-07-17
fix: backport vendored mixin from #2534

agoose77:
This PR uses only the vendored `NDArrayOperatorsMixin` mixin class (without `__slots__`) from #2534


src/awkward/_v2/operations/ak_to_layout.py
```diff
@@ -95,7 +95,7 @@ def _impl(array, allow_record, allow_other, numpytype):
             numpytype,
         )
 
-    elif ak.nplike.is_jax_buffer(array) and type(array).__name__ == "DeviceArray":
+    elif ak.nplike.is_jax_buffer(array):
```
agoose77, line 98: Any `jaxlib` object is now permitted here, which mirrors v2.

jpivarski (commented):
Why is there a _v2 directory? I noticed that `NDArrayOperatorsMixin` is now defined twice, and then one of them is in the _v2 directory.

This is a good thing to do, though!

agoose77:
@jpivarski this is a backport for `main-v1`, which still has `_v2` (although it doesn't receive feature updates). In fact, our unofficial policy for `_v2` is that it doesn't get *much* bugfix support either, because so many bugs are fixed by the large-scale changes in `main`.

PR #2587, opened 2023-07-19
fix: don't decorate non-`Exception` subclasses

### LINKED ISSUE #2586: We shouldn't decorate the `KeyboardInterrupt` exception
### Version of Awkward Array

HEAD

If I press control-C (or ■ in Jupyter), I don't need to be told what high-level function failed because it didn't really "fail," it was cancelled.

```
KeyboardInterrupt: 

    <MomentumArray4D [[{pt: 63, eta: ..., ...}, ...], ...] type='299973 * v...'>

with

    'phi'
```

Similarly, we have to think about what will happen if someone cancels an asynchronous GPU process—does the background thread get informed? That's beyond this issue, though.

agoose77:
Fixes #2586

### DIFF
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -59,7 +59,11 @@ class ErrorContext:
     def __exit__(self, exception_type, exception_value, traceback):
         try:
             # Handle caught exception
-            if exception_type is not None and self.primary() is self:
+            if (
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
This looks simple enough to merge immediately. Thanks!

PR #2592, opened 2023-07-26
fix: use `to_backend_array()` instead of `asarray`

agoose77:
Succinctly, `asarray` is less powerful than `to_backend_array()`. It often goes through iterative type discovery, rather than leveraging our internal typed layouts. We should instead be using our internal `to_backend_array()`.


jpivarski (approved):
Okay to merge, and it looks like a simple enough change that you're done with it and I can merge it. Thanks!

PR #2595, opened 2023-07-27
feat: add support for expected termination of transform functions

agoose77:
#2593 demonstrates a case where it would be nice for `ak.transform` to be able to error if the user function never terminates.

This PR adds an `expect_return_value` argument, which causes `ak.transform` to throw a `RuntimeError` if the user transform is supposed to return a non-None value, but never terminates.


jpivarski (approved):
Okay, this is a safety check on `ak.transform` users.

It's good to merge, but I'll let you press the button in case there are any last fixes.

PR #2604, opened 2023-08-03
feat: unpickle arrays made in Awkward v1 (as v2).

jpivarski:
This will make @zbhatti's life easier.

Now it's possible to move arrays all the way from Awkward v0 to v1 to v2, although they need to be pickled and unpickled in each step.

1. In an environment with Awkward 0.x, pickle the array.
2. In an environment (possibly the same) with the `awkward0` library and `awkward<2` (v1), use pickle to load it as an `awkward0` array, then use the [ak.from_awkward0 (v1)](https://awkward-array.org/doc/1.10/_auto/ak.from_awkward0.html) function to convert it into a v1 array.
3. In the same environment, save the v1 array with pickle.
4. In a different environment with `awkward>=2` (v2), use pickle to load it as a v2 array.

This PR makes the last step possible.

@agoose77, we had talked about the possibility of doing this, and weren't sure if it was a good idea to reintroduce `ak._ext`. In v1, that was a compiled extension with all the C++ code in it. In this PR, it's an ordinary Python module providing synonyms for the `Form` classes, since this is where pickle will look for them. The downside of this is that `_ext` now has a very different function, and it's not in any way "external." However, it's also hidden with an underscore.

The second part to get this to work is for all of the `Form` subclasses to be unpicklable from their v1 formats _and_ their v2 formats. Fortunately, the v1 formats are always tuples and the v2 formats are always dicts, so they're easily distinguished.

This PR adds a `__setstate__` to each `Form` subclass (no `__getstate__`; we want the default v2 behavior of returning `self.__dict__`). If it's a tuple (v1), each subclass has to follow a different prescription to unpickle it, as all of the v1 `Form` subclasses were pickled in different ways.

If a v1 array had multiple partitions, only the first partition will be loaded (with no error). Something ought to be done about that; an error message at least (probably in `ak.Array.__setstate__`). A v1 array can't be pickled as virtual because it gets [ak.packed (v1)](https://awkward-array.org/doc/1.10/_auto/ak.packed.html).

In this PR, only the `Form` subclasses were modified. Nothing had to be done other than that (in `ak.from_buffers` or `ak.Array.__setstate__`).

There are tests for array types that aren't eliminated by [ak.packed (v1)](https://awkward-array.org/doc/1.10/_auto/ak.packed.html). The untestable ones are simple extrapolations from the tested ones. It'll be fine!


tests/test_2591_nan_to_num.py
```diff
@@ -11,5 +11,6 @@
 
 def test():
     ak.nan_to_num(
-        ak.Array([1, 2, 3], backend="jax"), nan=ak.Array([1, 2, 3], backend="jax")
+        ak.Array([1.1, 2.2, 3.3], backend="jax"),
+        nan=ak.Array([1.1, 2.2, 3.3], backend="jax"),
     )
```
jpivarski, line 16: This change had nothing to do with the PR. A new version of JAX complains that integers are not an inexact type, so I made them floats.

jpivarski:
We should let @zbhatti test this before merging.

pyproject.toml
```diff
@@ -77,6 +77,7 @@ path = "src/awkward/_version.py"
 include = [
     "/src",
     "/tests",
+    "/tests/samples",
```
jpivarski: @henryiii, the sdist test was failing because the pickle files that I made in Awkward v1 to test this weren't available. I could just include those files (`/tests/samples/*.pkl`), but all of the sample files should be included. I think it was passing tests before just because the sdist packaging test didn't have all optional dependencies installed. Therefore, it didn't test reading of Parquet and Avro files.

Adding the whole `/tests/samples` directory adds 220 kB. I think that's fine, even for an sdist. Do you have an opinion on that?

henryiii: I’d generally always include test files in SDist. Are these files gitignored or something?

matthewfeickert:
> We should let @zbhatti test this before merging.

@zbhatti in your environment to try this out you should follow the instructions here: https://github.com/scikit-hep/awkward#installation-for-developers (but with this branch `jpivarski/allow-unpickle-from-awkward1` checked out)

pyproject.toml
henryiii: These files are present if you just have `/tests`, hatchling includes the subdirectories automatically. Tested by `pipx run build --sdist` then `tar -tvf dist/awkward-2.3.1.tar.gz`.

pyproject.toml
```diff
@@ -85,7 +85,7 @@ include = [
     "/README.md",
     "/requirements-test.txt"
 ]
-artefacts = [
+artifacts = [
```
henryiii: This was a typo, so these were not included before. Are they needed?

agoose77: Are you SERIOUS? That's frustrating (and definitely my British fault). We certainly want these in our sdists.

jpivarski: I didn't know about that one. ([Cool graphs](https://grammarist.com/spelling/artefact-artifact/).)

We haven't coordinated British vs American spelling conventions in the codebase, but so far, it has always been without consequences. It's only a problem when a different spelling than intended goes quietly unnoticed, which isn't the case for variable and function names.

Maybe this issue should be passed up to Python itself: an unexpected word in pyproject.toml should raise an error, rather than be quietly ignored? Or maybe this specific case should be added to the codespell tool?

henryiii: There are two options - either Hatchling could provide a json-schema, and then validate-pyproject would be able to validate the `tool.hatch` section, and/or hatchling could gain a `strict-config` setting, which I don't see present. (Scikit-build-core will have a schema in 0.5). Not sure if @ofek wants to work on it or would accept work toward it (it really depends on the backend settings implementation - it was trivial with scikit-build-core, because it was declarative dataclass-based).

ofek: A strict option would be doable but a schema-based implementation would be a large effort

agoose77: I'm resolving this as I've corrected it in `main`.

pyproject.toml
agoose77: The `tests/samples` subdirectory is already included in our sdists, so I don't think this part of the PR is required. I'll revert it and see what's failing.

agoose77:
This is apparently easier than I'd anticipated, thanks for tackling it @jpivarski!

We could create shim classes in `_ext` to avoid mixing v1 logic into v2 forms, but I'm not *that* worried about it, so let's keep it as-is.

I've added support for partitioned arrays, which I think ticks all boxes?

src/awkward/highlevel.py
```diff
@@ -1400,14 +1444,33 @@ def __reduce__(self):
...
-        layout = ak.operations.from_buffers(
-            form,
-            length,
-            container,
-            highlevel=False,
-            buffer_key="{form_key}-{attribute}",
-            byteorder="<",
-        )
+        # If length is a sequence, we have awkward1
+        if isinstance(length, Sequence):
```
jpivarski, line 1450: Nice! I'm glad we have this way to identify an Awkward 1 partitioned array.

jpivarski:
> I've added support for partitioned arrays, which I think ticks all boxes?

That's great!

To fix CI, we'll need to merge in #2617.

It looks like the last successful test (5b1677c0a57342af1fddbc9ce7923c97f32ee1b1) passed the sdist tests.

https://github.com/scikit-hep/awkward/pull/2604#discussion_r1282580623 should have been a separate PR, but we can just wait for this one to get merged, since I think that's pretty soon.

We haven't heard from @zbhatti about whether this works in his case. We can wait a bit longer, but then move ahead anyway and if there are any specific problems with unpickling his files, we'll address them in another PR.

matthewfeickert:
> We can wait a bit longer, but then move ahead anyway and if there are any specific problems with unpickling his files, we'll address them in another PR.

@jpivarski as I'm working with him I think it is fine to merge whenever you're ready and he can follow up with additional questions.

PR #2606, opened 2023-08-03
fix: add inline to prevent multiple definitions

ManasviGoyal:
The following changes are made -
- Some methods were causing multiple definitions when including LayoutBuilder in another C++ header file so inline is added to `default_options`, `type_to_name`, and `type_to_numpy_like`.
- preprocessor `#define` is used as a temporary workaround in place of `default_options` as inline variables are supported starting from C++17. Once the requirements are updated to C++17, `default_options` can be inlined.
- Also, two of the tests in `test_1494-layout-builder.cpp` which included the EmptyArray case were failing due to missing parameter string in the form, so that has been fixed too by removing parameters from the `LayoutBuilder` code.

### DIFF
--- a/header-only/layout-builder/awkward/LayoutBuilder.h
+++ b/header-only/layout-builder/awkward/LayoutBuilder.h
@@ -13,13 +13,13 @@
 #include <string>
 #include <functional>
 
+/// @brief Object of {@link BuilderOptions BuilderOptions} which sets the
+/// values of the default options.
+#define AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS awkward::BuilderOptions(1024, 1)
+
 namespace awkward {
 
   namespace LayoutBuilder {
-
-    /// @brief Object of {@link BuilderOptions BuilderOptions} which sets the
-    /// values of the default options.
-    inline awkward::BuilderOptions default_options(1024, 1);
 
     /// @class Field
     ///
@@ -55,10 +55,10 @@ namespace awkward {
     class Numpy {
     public:
       /// @brief Creates a new Numpy layout builder by allocating a new buffer,
-      /// using `default_options` for initializing the buffer.
+      /// using `AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS` for initializing the buffer.
       Numpy()
           : data_(
-                awkward::GrowableBuffer<PRIMITIVE>(default_options)) {
+                awkward::GrowableBuffer<PRIMITIVE>(AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS)) {
         size_t id = 0;
         set_id(id);
       }
@@ -210,10 +210,10 @@ namespace awkward {
     class ListOffset {
     public:
       /// @brief Creates a new ListOffset layout builder by allocating a new `offset`
-      ///  buffer, using `default_options` for initializing the buffer.
+      ///  buffer, using `AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS` for initializing the buffer.
       ListOffset()
           : offsets_(
-                awkward::GrowableBuffer<PRIMITIVE>(default_options)) {
+                awkward::GrowableBuffer<PRIMITIVE>(AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS)) {
         offsets_.append(0);
         size_t id = 0;
         set_id(id);
@@ -1034,10 +1034,10 @@ namespace awkward {
     class IndexedOption {
     public:
       /// @brief Creates a new IndexedOption layout builder by allocating a new `index`
-      /// buffer, using `default_options` for initializing the buffer.
+      /// buffer, using `AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS` for initializing the buffer.
       IndexedOption()
           : index_(
-                awkward::GrowableBuffer<PRIMITIVE>(default_options)),
+                awkward::GrowableBuffer<PRIMITIVE>(AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS)),
             last_valid_(-1) {
         size_t id = 0;
         set_id(id);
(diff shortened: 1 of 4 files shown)


ManasviGoyal:
@ianna I notice that it gives an error - `inline variables require at least std:c++17` and `LayoutBuilder.h` is restricted to `std:c++11`.

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -19,7 +19,7 @@ namespace awkward {
 
     /// @brief Object of {@link BuilderOptions BuilderOptions} which sets the
     /// values of the default options.
-    awkward::BuilderOptions default_options(1024, 1);
+    inline awkward::BuilderOptions default_options(1024, 1);
```
ianna: This is an inlined variable - does it need to be?

ManasviGoyal:
@jpivarski `default_options` is an inline variable that's why it gives an error that it requires C++17.

`D:\a\awkward\awkward\header-only\layout-builder\awkward/LayoutBuilder.h(22,36): error C7525: inline variables require at least '/std:c++17' [D:\a\awkward\awkward\build\tests\test_1494-layout-builder.vcxproj]`

header-only/layout-builder/awkward/LayoutBuilder.h
ManasviGoyal: When we include `LayoutBuilder.h` in multiple source files, it results in multiple definitions of `default_options` leading to linker errors. So yes it needs to be inline.

```
test-fake.cpp:(.text+0x360): multiple definition of `std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const awkward::type_to_numpy_like<long>()'; /tmp/ccmHoZHl.o:fake.cpp:(.text+0x360): first defined here
/usr/bin/ld: /tmp/cculuiVU.o:(.bss+0x0): multiple definition of `awkward::LayoutBuilder::default_options'; /tmp/ccmHoZHl.o:(.bss+0x0): first defined here
```

jpivarski:
Ah, right! I missed that `default_options` is a variable.

If you want a quick-and-dirty way to ensure that this PR does not depend on #2608, you could replace the inline variable with a preprocessor `#define`. That name will leak into any code that `#includes` this file, without `namespace` qualifications, so the name would need to be qualified:

```c++
#define AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS awkward::BuilderOptions(1024, 1)
```

and then use `AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS` in place of `awkward::LayoutBuilder::default_options` everywhere.

------------------------

For completeness, all of the functions ought to be `inline`, as @henryiii pointed out. In LayoutBuilder.h, I see:

  * `default_options` (already addressed)
  * `class Empty` is not a template, but I'm not sure if the always-inline rule for header-only libraries applies to classes...

In utils.h, I see:

  * `type_to_name` is a templated function, but it has been instantiated for some concrete types, so I guess that needs to be inlined (already addressed)
  * `type_to_numpy_like`, same thing
  * `type_to_form` is a templated function, and I don't see any concrete instantiations, so I guess it's fine to leave it non-inlined
  * `visit_impl` is implemented in some concrete cases; maybe that needs to be inlined. It hasn't been: does anything bad happen if you try inlining it? (Do you know what it's used for, and do you have any tests that would verify that it's okay to inline?)

ManasviGoyal:
> If you want a quick-and-dirty way to ensure that this PR does not depend on #2608, you could replace the inline variable with a preprocessor `#define`. That name will leak into any code that `#includes` this file, without `namespace` qualifications, so the name would need to be qualified:
> 
> ```c++
> #define AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS awkward::BuilderOptions(1024, 1)
> ```
> 
> and then use `AWKWARD_LAYOUTBUILDER_DEFAULT_OPTIONS` in place of `awkward::LayoutBuilder::default_options` everywhere.
> 

Yes, this can be a temporary workaround until the version isn't updated to C++17. Should I add these changes to this PR?


> * `visit_impl` is implemented in some concrete cases; maybe that needs to be inlined. It hasn't been: does anything bad happen if you try inlining it? (Do you know what it's used for, and do you have any tests that would verify that it's okay to inline?)

`visit_impl` is a struct so it can't be inlined. `default_options`, `type_to_name`, and `type_to_numpy_like` are the ones that give linker errors so I think those might be the only one that needs to be inlined.

jpivarski:
Okay, then it sounds good. Let's use the preprocessor as a temporary workaround for not requiring C++17 yet. (Our next opportunity to do that is September 1.)

ianna (approved):
Thanks, @ManasviGoyal! Looks good to me.

header-only/tests/test_1494-layout-builder.cpp
```diff
@@ -531,7 +532,8 @@ test_ListOffset_Empty() {
           "\"class\": \"ListOffsetArray\", "
           "\"offsets\": \"i64\", "
           "\"content\": { "
-              "\"class\": \"EmptyArray\" "
+              "\"class\": \"EmptyArray\", "
+              "\"parameters\": {} "
```
agoose77: @ManasviGoyal why does this test change? Was it not being run before?

ManasviGoyal: When I run the the `LayoutBuilder` tests, the assertion is failing due to the missing `"parameters: {}"` described in the `EmptyArray` form. If this test is run while running all the tests, it's strange that it doesn't give any error despite the incomplete form.

jpivarski: None of the Forms should explicitly require a `parameters` key when reading from JSON. The `parameters` key is always extracted with `get`:

https://github.com/scikit-hep/awkward/blob/089f8430e4b95b8b6b84ce941f6d1423dd6f4518/src/awkward/forms/form.py#L36-L42

(`ak.forms.from_json` immediately calls `ak.forms.from_dict`. When `parameters` are `None`, it's the same as if they're `{}`.)

On top of that, EmptyArray is required to not have any `parameters`. Explicitly providing `None` or `{}` is benign, though:

https://github.com/scikit-hep/awkward/blob/089f8430e4b95b8b6b84ce941f6d1423dd6f4518/src/awkward/forms/emptyform.py#L15-L21

If these are C++ tests that are failing, and the C++ tests are more restrictive than what the Python tests require, then it's the test that should change.

ManasviGoyal: The C++ tests were failing because of the mismatch between the `EmptyArray` form definition in the code and the tests. If the parameters are not needed in this case, then the parameters part from the `EmptyArray` form in the `Layoutbuilder.h` code can be removed then we don't have to change the tests

https://github.com/scikit-hep/awkward/blob/aba0b8d5412a1e0db89a27d065ab5f2bb6759846/header-only/layout-builder/awkward/LayoutBuilder.h#L416-L419

jpivarski: Correct. It's also the only node type that sets it to a constant; all of the other LayoutBuilder node types take a `parameters` argument to pass on here.

ManasviGoyal: Yes exactly. I'll update the PR with the required changes.

PR #2610, opened 2023-08-03
fix: `assert nextlen == 0` in reducer should also allow `unknown_length`

### LINKED ISSUE #2607: TypeError when summing regular array of unknown length
### Version of Awkward Array

2.3.1

```python
import awkward as ak


x = ak.to_regular(ak.Array([[1.]])[[[False]]])
t = ak.Array(x.layout.to_typetracer(forget_length=True))
ak.sum(t, axis=1)
```
The last line raises an error: `TypeError: cannot compare unknown lengths against known values`.
It works fine when x is not regular, meaning `x = ak.Array([[1.]])[[[False]]]`.
In my case, this breaks dask-awkward functionality when trying to sum over regular dask-awkward arrays.

jpivarski:
@jrueb should test this to find out if it's the only blocker, or if there are other (related?) errors down the line.


jrueb:
This fixes the issue for me.

jpivarski:
I should have marked this ready for review last week when @jrueb said it worked. This one ought to go in today's release because it's fixing a bug encountered in the wild. Any chance you could sign off on it today, @agoose77?

agoose77 (approved):

jpivarski:
Thanks!

PR #2613, opened 2023-08-04
feat: export error classes

### LINKED ISSUE #2611: Errors should be exported to public namespace
### Version of Awkward Array

main

Now that we have a subset of custom error types, we should export them so that users can catch them.

agoose77:
Fixes #2611

Now we have `ak._errors` and `ak._errors`. This is unfortunate. The constraints I'm operating under are:
- public symbols should live in public modules
- private symbols should preferably live in private modules

There are enough internal symbols (deprecation, error handling) that we don't want to put these inside `ak.errors` in my opinion, hence the two modules.

I'm open to suggestions on how to structure this differently @jpivarski.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -304,6 +304,8 @@ mccabe.max-complexity = 100
 [tool.ruff.per-file-ignores]
 "dev/*" = ["T20", "TID251"]
 "src/awkward/numba/*" = ["TID251"]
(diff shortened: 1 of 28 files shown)


jpivarski (approved):
Just moving them into a public space is a fine idea. We now have both `errors.py` and `_errors.py`, but I think that's okay.

This PR doesn't look like it will have any amendments, so I'll merge it now.

PR #2616, opened 2023-08-04
feat: use pyarrow for string functions

jpivarski:
This PR adds support for vectorised string operations using PyArrow's `pyarrow.compute` module.
It implements all of the listed [string predicates](https://arrow.apache.org/docs/python/api/compute.html#string-predicates), [string transforms](https://arrow.apache.org/docs/python/api/compute.html#string-transforms), [string padding](https://arrow.apache.org/docs/python/api/compute.html#string-padding), [string trimming](https://arrow.apache.org/docs/python/api/compute.html#string-trimming), [string splitting](https://arrow.apache.org/docs/python/api/compute.html#string-splitting), [string component extraction](https://arrow.apache.org/docs/python/api/compute.html#string-component-extraction), [string joining](https://arrow.apache.org/docs/python/api/compute.html#string-joining), [string slicing](https://arrow.apache.org/docs/python/api/compute.html#string-slicing), and [string containment tests](https://arrow.apache.org/docs/python/api/compute.html#containment-tests).

### String predicates

- [X] [is_alnum](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_alnum.html) Classify strings as alphanumeric.
- [x] [is_alpha](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_alpha.html) Classify strings as alphabetic.
- [x] [is_decimal](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_decimal.html) Classify strings as decimal.
- [x] [is_digit](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_digit.html) Classify strings as digits.
- [x] [is_lower](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_lower.html) Classify strings as lowercase.
- [x] [is_numeric](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_numeric.html) Classify strings as numeric.
- [x] [is_printable](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_printable.html) Classify strings as printable.
- [x] [is_space](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_space.html) Classify strings as whitespace.
- [x] [is_upper](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_upper.html) Classify strings as uppercase.
- [X] [is_title](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_title.html) Classify strings as titlecase.
- [x] [is_ascii](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_is_ascii.html) Classify strings as ASCII.

### String transforms

- [x] [capitalize](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_capitalize.html) Capitalize the first character of input.
- [x] [length](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_length.html) Compute UTF8 string lengths.
- [x] [lower](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_lower.html) Transform input to lowercase.
- [x] [swapcase](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_swapcase.html) Transform input lowercase characters to uppercase and uppercase characters to lowercase.
- [x] [title](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_title.html) Titlecase each word of input.
- [x] [upper](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_upper.html) Transform input to uppercase.
- [x] [repeat](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_repeat.html) Repeat a binary string.
- [x] [replace_slice](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_replace_slice.html) Replace a slice of a string.
- [x] [reverse](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_reverse.html) Reverse input.
- [x] [replace_substring](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_replace_substring.html) Replace matching non-overlapping substrings with replacement.
- [x] [replace_substring_regex](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_replace_substring_regex.html) Replace matching non-overlapping substrings with replacement.

### String padding

- [x] [center](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_center.html) Center strings by padding with a given character.
- [x] [lpad](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_lpad.html) Right-align strings by padding with a given character.
- [x] [rpad](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_rpad.html) Left-align strings by padding with a given character.

### String trimming

- [x] [ltrim](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_ltrim.html) Trim leading characters.
- [x] [ltrim_whitespace](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_ltrim_whitespace.html) Trim leading whitespace characters.
- [x] [rtrim](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_rtrim.html) Trim trailing characters.
- [x] [rtrim_whitespace](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_rtrim_whitespace.html) Trim trailing whitespace characters.
- [x] [trim](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_trim.html) Trim leading and trailing characters.
- [x] [trim_whitespace](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_trim_whitespace.html) Trim leading and trailing whitespace characters.

### String splitting

- [x] [split_pattern](https://arrow.apache.org/docs/python/generated/pyarrow.compute.split_pattern.html) Split string according to separator.
- [x] [split_whitespace](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ascii_split_whitespace.html) Split string according to any ASCII whitespace.
- [x] [split_pattern_regex](https://arrow.apache.org/docs/python/generated/pyarrow.compute.split_pattern_regex.html) Split string according to regex pattern.

### String component extraction

- [x] [extract_regex](https://arrow.apache.org/docs/python/generated/pyarrow.compute.extract_regex.html) Extract substrings captured by a regex pattern.

### String joining

- [x] [join](https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_join.html) Join a list of strings together with a separator.
- [x] [join_element_wise](https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_join_element_wise.html) Join string arguments together, with the last argument as separator.

### String slicing

- [x] [slice](https://arrow.apache.org/docs/python/generated/pyarrow.compute.utf8_slice_codeunits.html) Slice string.

### Containment tests

- [x] [count_substring](https://arrow.apache.org/docs/python/generated/pyarrow.compute.count_substring.html) Count occurrences of substring.
- [x] [count_substring_regex](https://arrow.apache.org/docs/python/generated/pyarrow.compute.count_substring_regex.html) Count occurrences of substring.
- [x] [ends_with](https://arrow.apache.org/docs/python/generated/pyarrow.compute.ends_with.html) Check if strings end with a literal pattern.
- [x] [starts_with](https://arrow.apache.org/docs/python/generated/pyarrow.compute.starts_with.html) Check if strings start with a literal pattern.
- [x] [find_substring](https://arrow.apache.org/docs/python/generated/pyarrow.compute.find_substring.html) Find first occurrence of substring.
- [x] [find_substring_regex](https://arrow.apache.org/docs/python/generated/pyarrow.compute.find_substring_regex.html) Find location of first match of regex pattern.
- [x] [index_in](https://arrow.apache.org/docs/python/generated/pyarrow.compute.index_in.html) Return index of each element in a set of values.
- [x] [is_in](https://arrow.apache.org/docs/python/generated/pyarrow.compute.is_in.html) Find each element in a set of values.
- [x] [match_like](https://arrow.apache.org/docs/python/generated/pyarrow.compute.match_like.html) Match strings against SQL-style LIKE pattern.
- [x] [match_substring](https://arrow.apache.org/docs/python/generated/pyarrow.compute.match_substring.html) Match strings against literal pattern.
- [x] [match_substring_regex](https://arrow.apache.org/docs/python/generated/pyarrow.compute.match_substring_regex.html) Match strings against regex pattern.


jpivarski:
The `split_*` functions don't look fundamentally hard. They have to pass parameters (constants, not broadcasted arrays) and they output lists of strings instead of strings. If the `bytestring_to_string=True` option is needed for bytestrings (very unlikely!), then the string → bytestring repacking will need to be applied one level deeper (but I don't think that `bytestring_to_string=True` will be needed).

The `extract_regex` function is straightforward to call, but it returns an Arrow struct. If the record field names are formulaic (`"\1"`, `"\2"`, etc.), then we'd probably want to set the `fields` of the output to `None` so that it becomes an Awkward tuple. (We can assign to the output in place because it was just created by `ak.from_arrow`, and the `fields` is not the zero-copy data.)

The only remaining function that needs to be broadcasted (like `binary_repeat`) is `join_element_wise`. Unlike `binary_repeat`, there's no choice about whether to broadcast or not, and there can be more than two arrays to broadcast. (All of the `*strings` arrays [documented here](https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_join_element_wise.html) would have to be broadcasted). This will probably be the hardest one.

By comparison, `join` ([documented here](https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_join.html)) does not broadcast, but it needs to be applied at the level of lists of strings, not just strings (one level deep). It's a reducer.

Two of the containment tests take an array-like second argument: `index_in` and `is_in`. This argument is `value_set` (see [index_in](https://arrow.apache.org/docs/python/generated/pyarrow.compute.index_in.html) and [is_in](https://arrow.apache.org/docs/python/generated/pyarrow.compute.is_in.html), which is a set of strings that each string in the `values` is compared against, but _it is not broadcasted_ to `values`. So it's much easier than `join_element_wise`.

None of the remaining functions look sufficiently different from the ones that have been implemented so far as to be a problem.

agoose77:
Jim, this is a phenomenal piece of work. Looking forward to reviewing it.

P.S. I touched your PR description to fix the links. Correct anything if you feel I made a mistake.

agoose77:
I've added `split_whitespace`, `split_pattern`, and `split_pattern_regex`. I've noticed that these operations return nullable types, so the round-trip logic in `_get_action` doesn't correctly re-interpret the result as bytestrings if the input is a bytestring (and `bytestring_to_string=True`). 

As such, I added a new `_get_split_action` that handles the case where a returned list contains a nullable string`. I don't *think* we should ever get anything besides an `UnmaskedArray` here, as we never pass any null data to Arrow.

It's not clear to me whether we're passing a nullable type, or whether Arrow is just deciding to return a nullable type. My hunch was that if the input `StringArray` is nullable, then the result is nullable, but I can't check whether the type we build is considered `nullable` by arrow; `large_string` or `large_binary` don't seem to expose this information.

I haven't added tests here yet, but they should be fairly easy.

agoose77:
`binary_join` is not too tricky, but it looks like the Arrow kernel only supports nullable, _small_ strings. A first-pass is [here](https://gist.github.com/agoose77/28e5bb0250678e454356a85861a16368), although it should be cleaned up. I just played around to get it working, rather than thinking through robustly what the recursion function should look like.

When working on these functions, two questions come to mind:
1. Should we perform validation logic in Awkward, rather than relying on Arrow to error if an invalid type is passed?
2. Should we treat operations on incompatible types (e.g. string operations on non-strings, or on record arrays) as errors, rather than silently returning the original array?

On both of these points, I think the answer is yes. There are some Awkward functions for which being permissive is perhaps a good thing, e.g. `ak.values_astype`. However, in general we have type information available and I think it would be useful for users to be able to detect operations upon the "wrong" types as early as possible. Do you agree, @jpivarski?

src/awkward/contents/unmaskedarray.py
```diff
@@ -491,7 +491,7 @@ def _remove_structure(self, backend, options):
             return [self]
 
     def _drop_none(self) -> Content:
-        return self.to_ByteMaskedArray(True)._drop_none()
+        return self.content[:0]
```
jpivarski: I think this caused an error in an adjacent test:

```
________________________ test_UnmaskedArray_NumpyArray _________________________

    def test_UnmaskedArray_NumpyArray():
        a = ak.contents.unmaskedarray.UnmaskedArray(
            ak.contents.numpyarray.NumpyArray(np.array([0.0, 1.1, 2.2, 3.3]))
        )
    
>       assert to_list(ak.drop_none(a)) == [0.0, 1.1, 2.2, 3.3]
E       assert [] == [0.0, 1.1, 2.2, 3.3]
E         Right contains 4 more items, first extra item: 0.0
E         Full diff:
E         - [0.0, 1.1, 2.2, 3.3]
E         + []

a          = <UnmaskedArray len='4'>
    <content><NumpyArray dtype='float64' len='4'>[0.  1.1 2.2 3.3]</NumpyArray></content>
</UnmaskedArray>

tests/test_1904_drop_none.py:197: AssertionError
```

"Drop none" means remove the missing values and return data with a non option-type type. Removing _all_ elements from the array is not the right thing to do with an `UnmaskedArray`; it's the opposite: all elements are non-missing.

Converting the `UnmaskedArray` into a `ByteMaskedArray` was perhaps too computationally intensive; it could have been done by

```python
    def _drop_none(self) -> Content:
        return self.content
```

to remove option from the type, but not change the values at all.

src/awkward/operations/str/__init__.py
```diff
@@ -103,3 +108,81 @@ def action(layout, **absorb):
...
     return action
+
+
+def _erase_list_option(layout):
+    from awkward.contents.unmaskedarray import UnmaskedArray
+
+    assert layout.is_list
+    if layout.content.is_option:
+        assert isinstance(layout.content, UnmaskedArray)
+        return layout.copy(content=layout.content.content)
```
jpivarski, line 136: Since we only ever send non-missing strings to Arrow, it's fair to assume that the only outputs are non-missing as well, even if Arrow says that the type is potentially nullable.

Actually, I think that the non-nullable type is part of the information that's lost by setting `extensionarray=False`. But we want that because Arrow Compute only applies its string operations if it recognizes the type, and it doesn't recognize the type if it's an extension array.

jpivarski: > It's not clear to me whether we're passing a nullable type, or whether Arrow is just deciding to return a nullable type.

I'm pretty sure that Arrow is just deciding to return a nullable type. Arrow's type system does not allow it to express non-nullability (the _default_ is nullable) in the type objects themselves; non-nullability can only be expressed in fields. There's a two-level structure under each struct and Table: field (with name, nullability, and some other things), which contains type. If you have a raw list, not in a struct or Table, I don't think there's a place to put the non-nullability information.

These corner-cases are the reason we use extension arrays, to carry more information through Arrow and Parquet. But we can't use one here if we want Arrow Compute to recognize the input as strings.

src/awkward/contents/unmaskedarray.py
agoose77: Gah, that was not supposed to make its way upstream (and I'm glad the tests caught it). I had a brain fart earlier when figuring out something adjacent to the arrow work.

jpivarski:
> When working on these functions, two questions come to mind:
> 
> 1. Should we perform validation logic in Awkward, rather than relying on Arrow to error if an invalid type is passed?
> 2. Should we treat operations on incompatible types (e.g. string operations on non-strings, or on record arrays) as errors, rather than silently returning the original array?

1. We ensure the type of data that we send to Arrow. If there are any errors to catch after this, it could be hard to discover them all. (At least, it could be a big project.) Arrow's own error messages are _probably_ good enough, especially since we show which `ak.str.*` function it shows up in.
2. The value-added that Awkward provides is that the strings could be nested in some other data types. For instance, we add the ability to do crazy stuff like this:

```python
>>> ak.str.upper(ak.Array(["look", ["what", ["we", ["can", "do!"]]]])).show()
['LOOK',
 ['WHAT', ['WE', ['CAN', 'DO!']]]]
```

We wouldn't want to throw that away because it's not (exclusively) string-like. Maybe if the `recursively_apply` finishes without finding _any_ strings, maybe that could be an error case. I'll look into that.

jpivarski:
> By comparison, `join` ([documented here](https://arrow.apache.org/docs/python/generated/pyarrow.compute.binary_join.html)) does not broadcast, but it needs to be applied at the level of lists of strings, not just strings (one level deep). It's a reducer.

I was wrong: `binary_join` does broadcast. I'll try using your gist.

agoose77:
@jpivarski is [559073b](https://github.com/scikit-hep/awkward/pull/2616/commits/559073b1b047e099b5fce0dcd0930bd2154feedd) necessary? The arguments for e.g. `ak_center` are not handled via broadcasting.

agoose77:
@jpivarski I completed a preliminary pass of the containment tests. Would you be able to give them another eye? The error behaviour isn't right yet, but the non-error cases should work.

I'll give the docstrings a proper pass before final review, too. These are placeholder.

agoose77:
@jpivarski I think this is ready to go! I see the merit in being permissive for our string operations, and I think upon reflection that most of our functions are like this.

I have one outstanding API question that we need to figure out — whether to drop the `ak_` prefix for string function module names? I think we should, and it suggests that the module names in `ak.operations` should also be changed. I don't know the motivation for the `ak_` prefix; it helps with indicating that a module contains a high-level function, but that's not a strong motivation for keeping it.

jpivarski:
The `ak_` prefix was to distinguish between modules and the high-level functions they contain. Some of the high-level functions have the same names as Python built-ins, so I wanted something to not spread that degeneracy into the module names as well.

You're right that `ak_` makes a lot less sense for these functions, which aren't `ak.*` but `ak.str.*`. I wasn't sure what to do about that when starting this PR, but I just went with the flow. We could use a `ak_str_` prefix, adding boilerplate, but it would be less confusing. If someone has a nested, indented file-browser, digging more deeply into the file structure would indent to the right because these functions are in a directory named `str`, but also because their names have longer prefixes.

Maybe all of the modules should be prefixed by something neutral, like `m_` for module (not to be confused with "private" in C++), `h_` for high-level function, or `o_` for operation.

I don't think this needs to be decided now.

It's great that all of these functions are done! We should tell @martindurant that he can try them out on some log-file data whenever he gets a chance. I'll merge this (unclear who should review it) after testing against all other outstanding merges into main.

agoose77:
If we don't sort this now, well have to change the names in a breaking way down the road (only breaking for imports, which resolve file paths under the hood). That's already true for ak.operations.

I don't think we should make the ak.operations module private though - I think we want public apis to live in public modules. I'm still thinking that through, though.

jpivarski:
Technically true, but should people really be accessing `ak.XYZ` through `ak.operations.XYZ` or `ak.operations.ak_XYZ.XYZ` anyway? It seems like there should be only one way to access these functions; putting them inside directories is for our own organization. I'd be happy with `operations` → `_operations`.

martindurant:
I suppose this means there's no longer any need for the same functionality in awkward-pandas, but I do think it would like to keep a `str` accessor whatever you decide here, since that's the pandas pattern.

How is the performance?

Yes, agree that nice log datasets would be perfect;  https://github.com/logpai/loghub ?

jpivarski:
Performance should be whatever Arrow provides, which I expect to be good. (All of those functions are C++, exposed through Cython.)

It looks like the log files in that GitHub repo are not the full datasets. Some of them are supposedly 20‒30 GB, but the file itself looks like a few kB. Besides, if they really put many GB in a git repo, it would be very hard to clone.

We'd also need something to preprocess it, to get it into an Awkward format. That text processing would take some significant time.

Are any of them JSON? We already have an optimized JSON parser.

martindurant:
The full data files are at https://zenodo.org/record/8196385 (zip and tar.gz files). The repo contains all the data descriptions as READMEs.

jpivarski:
I renamed the modules in `ak.operations.str` as `akstr_*` to acknowledge that they're not in the `ak.*` namespace, but the `ak.str.*` namespace. I think this will at least be a non-confusing choice (and I thought the extra underscore between "ak" and "str" is gratuitous). Eventually, I think `ak.operations` should be private (L3), but that can be later and it will require a deprecation cycle.

I'm guessing this is not controversial, so I'll enable auto-merge, but you can stop it and we'll discuss if you have another idea.

docs/prepare_docstrings.py
```diff
@@ -303,7 +303,7 @@ def dofunction(link, linelink, shortname, name, astfcn):
         .replace(".behaviors.string", "")
     )
     shortname = re.sub(r"\.operations\.ak_\w+", "", shortname)
-    shortname = re.sub(r"\.operations\.str\.ak_\w+", ".str", shortname)
+    shortname = re.sub(r"\.operations\.str\.akstr_\w+", ".str", shortname)
```
jpivarski: Thanks!

PR #2617, opened 2023-08-07
fix: adjust for numexpr 2.8.5, which hid getContext's frame_depth argument

jpivarski:
In https://github.com/pydata/numexpr/commit/21ff376a0853aff6aea63343c6010d6434917319, `frame_depth` was changed to `_frame_depth`. We didn't need to set it, anyway, since we were selecting its default value. It came from code copied directly from NumExpr itself.


jpivarski:
(Merging this into `main` and then from `main` to active PRs should fix the CI errors we started to see over the weekend.)

PR #2620, opened 2023-08-08
refactor: add new `module` and `name` arguments to `high_level_function` decorator

agoose77:
This paves the way for #2616 to override the module for `ak.str` operations


agoose77:
@jpivarski already approved the rationale here, and it's so small that I'm happy just to merge and unblock.

PR #2622, opened 2023-08-08
docs: add guide on unflattening and grouping

agoose77:
This PR adds an initial implementation of the https://awkward-array.org/doc/main/user-guide/how-to-create-unflatten-group.html tutorial page. 

It's a hybrid user-guide and tutorial, following from the premise that we don't have enough resources to do both.

### DIFF
--- a/docs/_toc.yml
+++ b/docs/_toc.yml
@@ -52,7 +52,7 @@ subtrees:
                       - file: user-guide/how-to-create-strings
                         title: "Arrays of strings"
                       - file: user-guide/how-to-create-unflatten-group
-                        title: "Unflattening and grouping [todo]"
+                        title: "Unflattening and grouping"
                       - file: user-guide/how-to-create-arraybuilder
                         title: "ArrayBuilder (easy & general)"
                       - file: user-guide/how-to-create-constructors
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This looks good! The index heading still has a "[todo]" in it:

image

For group-by, I've been thinking that we might want to lean on Arrow even more and use its [group by aggregations](https://arrow.apache.org/docs/python/compute.html#grouped-aggregations). Same for [SQL-like joins](https://arrow.apache.org/docs/python/compute.html#table-and-dataset-joins).

Being NumPy-like, we don't intend to implement these features ourselves in Awkward Array, but we can make the best of both available in the same interface by passing Arrow's functionality through. If so, then the documentation on grouping would want to show the "Awkward native" way, involving flattening, run-lengths, and unflattening, as well as the "Arrow Compute" way, involving a direct call to that function.

agoose77:
> For group-by, I've been thinking that we might want to lean on Arrow even more and use its [group by aggregations](https://arrow.apache.org/docs/python/compute.html#grouped-aggregations). Same for [SQL-like joins](https://arrow.apache.org/docs/python/compute.html#table-and-dataset-joins).

Agreed. It doesn't make sense for us to re-invent the wheel unless we have to. The downside with Arrow is that it doesn't support all of the backends that we use, but there may be places that specialised Arrow support is worth having.

PR #2623, opened 2023-08-09
fix: support lists of strings in `ak.zip` with `optiontype_outside_record=True`

agoose77:
The `optiontype_outside_record` flag in `ak.zip` should permit lifting the option from a single record field to the outside of the record. However, with strings this does not happen; we test the wrong depth.


jpivarski (approved):
We talked about it. LGTM!

PR #2630, opened 2023-08-09
feat: add support for dictionary encoding from Arrow

agoose77:
This function is intended to replace `ak.to_categorical`, which is now deprecated. It is not called by the same name as the Arrow function that implements it, because the Awkward parlance is "convert to categorical", and I think we want our `ak.str` API to be Awkward-looking, even if it uses Arrow under the hood.

### DIFF
--- a/src/awkward/operations/str/__init__.py
+++ b/src/awkward/operations/str/__init__.py
@@ -1,18 +1,10 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
 # This set of string-manipulation routines is strongly inspired by Arrow:
-# https://arrow.apache.org/docs/python/api/compute.html#string-predicates
-# https://arrow.apache.org/docs/python/api/compute.html#string-transforms
-# https://arrow.apache.org/docs/python/api/compute.html#string-padding
-# https://arrow.apache.org/docs/python/api/compute.html#string-trimming
-# https://arrow.apache.org/docs/python/api/compute.html#string-splitting
-# https://arrow.apache.org/docs/python/api/compute.html#string-component-extraction
-# https://arrow.apache.org/docs/python/api/compute.html#string-joining
-# https://arrow.apache.org/docs/python/api/compute.html#string-slicing
-# https://arrow.apache.org/docs/python/api/compute.html#containment-tests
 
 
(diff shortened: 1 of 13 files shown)


src/awkward/operations/str/__init__.py
```diff
@@ -1,6 +1,16 @@
...
+# This set of string-manipulation routines is strongly inspired by Arrow:
 # https://arrow.apache.org/docs/python/api/compute.html#string-predicates
+# https://arrow.apache.org/docs/python/api/compute.html#string-transforms
+# https://arrow.apache.org/docs/python/api/compute.html#string-padding
+# https://arrow.apache.org/docs/python/api/compute.html#string-trimming
+# https://arrow.apache.org/docs/python/api/compute.html#string-splitting
+# https://arrow.apache.org/docs/python/api/compute.html#string-component-extraction
+# https://arrow.apache.org/docs/python/api/compute.html#string-joining
+# https://arrow.apache.org/docs/python/api/compute.html#string-slicing
+# https://arrow.apache.org/docs/python/api/compute.html#containment-tests
```
jpivarski: These are just scrolling to different places on the same webpage. To be useful as documentation, maybe pair each URL with the categories of imports below?

src/awkward/operations/str/__init__.py
```diff
@@ -73,6 +83,9 @@
 from awkward.operations.str.akstr_match_substring_regex import *
 from awkward.operations.str.akstr_starts_with import *
 
+# dictionary-encoding (exclusively for strings)
+from awkward.operations.str.akstr_to_categorical import *
```
jpivarski: Good: we'll introduce the new function under `ak.str` and deprecate the old one. (No overlapping.)

src/awkward/operations/str/__init__.py
agoose77: Sure!

src/awkward/operations/str/__init__.py
agoose77: Yep, it's in motion.

jpivarski (approved):
Great!

And I see that the deprecation is [already in the wiki](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20).

This is ready to merge.

PR #2631, opened 2023-08-09
feat: add CPU kernel for `to_numpy` support for strings/bytestrings

### LINKED ISSUE #2624: `to_numpy` uses `to_list` for strings
### Version of Awkward Array

main

We should implement a kernel that converts strings to UTF-8 and pads them such that NumPy can create a view over the array.

agoose77:
Currently, `to_numpy` (or rather, `to_backend_array`) uses `to_list()` for strings. We can do better than that, if we build the string view ourselves.

This PR prototypes those kernels; it could definitely be written to be faster, if needed.

In the interests of expediency, some of this code was taken from https://github.com/s22h/cutils. I'm not sure if it needs to be directly licensed as MIT, because it's mostly all adapted.

Closes #2624

### DIFF
--- a/src/awkward/contents/listoffsetarray.py
+++ b/src/awkward/contents/listoffsetarray.py
@@ -1986,10 +1986,61 @@ class ListOffsetArray(Content):
 
     def _to_backend_array(self, allow_missing, backend):
         array_param = self.parameter("__array__")
-        if array_param in {"bytestring", "string"}:
-            # As our array-of-strings _may_ be empty, we should pass the dtype
-            dtype = np.str_ if array_param == "string" else np.bytes_
-            return backend.nplike.asarray(self.to_list(), dtype=dtype)
+        if array_param == "string":
+            # Determine the widest string (in code points)
+            _max_code_points = backend.index_nplike.empty(1, dtype=np.int64)
+            backend[
+                "awkward_NumpyArray_prepare_utf8_to_utf32_padded",
+                self._content.dtype.type,
+                self._offsets.dtype.type,
+                _max_code_points.dtype.type,
+            ](
+                self._content.data,
+                self._offsets.data,
+                self._offsets.length,
+                _max_code_points,
+            )
+            max_code_points = backend.index_nplike.index_as_shape_item(
+                _max_code_points[0]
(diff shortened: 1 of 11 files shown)


agoose77:
@jpivarski your initial thoughts would be welcome here :)

src/awkward/contents/listoffsetarray.py
```diff
@@ -1986,10 +1986,73 @@ def _to_arrow(self, pyarrow, mask_node, validbytes, length, options):
...
+            max_code_points = backend.index_nplike.index_as_shape_item(
+                _max_code_points[0]
+            )
+            # Ensure that we have at-least length-1 bytestrings
+            if max_code_points is not unknown_length:
+                max_code_points = max(1, max_code_points)
+
+            # Allocate the correct size buffer
+            total_code_points = max_code_points * self.length
+            buffer = backend.nplike.empty(total_code_points, dtype=np.uint32)
```
jpivarski, line 2012: According to the comments and function names, this first step finds the maximum number of code points of any string in the array and uses that to allocate the buffer into which UTF-32 data will be written.

~~Shouldn't it be finding the maximum number of _bytes_, not _code points_? The size of the buffer into which UTF-32 data will be written is specified as a number of bytes.~~

Nope! I'm wrong! In UTF-32, each code point is a 4-byte integer, all the same size.

```python
buffer = backend.nplike.empty(total_code_points, dtype=np.uint32)
```

makes a `buffer` that is the correct number of bytes for UTF-32.

jpivarski (approved):
Looks good!

All it needs is a test, and that test should include variations in numbers of code points and also number of bytes per UTF-8 code points. There should be some in which the maximum number of bytes is achieved by having more code points and some in which the maximum number of bytes is achieved by having more bytes per code point.

These are 1-byte, 2-byte, 3-byte, and 4-byte UTF-8 sequences, each with exactly one code point:

```python
>>> [x.encode("utf-8") for x in ["$", "¢", "€", "💰"]]
[b'$', b'\xc2\xa2', b'\xe2\x82\xac', b'\xf0\x9f\x92\xb0']
```

PR #2632, opened 2023-08-09
fix: support `counts=len(array)` in `ak.unflatten`

agoose77:
It should be possible to invoke `ak.unflatten` with the length of the layout, such that it is wrapped in a length-1 `RegularArray`. However, this currently raises an error because the upper bound of permissible integers is taken to be `len(layout)-1`.


jpivarski (approved):
This looks like a straightforward bug-fix. I'll merge it; thanks!

PR #2634, opened 2023-08-10
fix: support non `int64` index in `ListArray`'s `pad_none`

agoose77:
This PR does two things:

1. Changes dtypes of a kernel call (`awkward_ListArray_rpad_axis1`)
   The kernel is written with the following template signature
   ```cpp
   T* toindex,
   const C* fromstarts,
   const C* fromstops,
   C* tostarts,
   C* tostops,
   ```
   i.e. `starts` and `tostarts` are expected to have the same dtypes. We can rewrite the kernel call site to ensure that this is true. This bug was observed after reading data from Arrow with 32-bit indices.
2. When using `pad_none`, we shouldn't copy the parameters of the content to the new mask node; this duplicates parameters, and doesn't consider the rules about e.g. whether the content is allowed to possess such parameters.[^note]
[^note]: this remains a general class of problem IIRC with broadcasting that we'll need to work on. However, this fix is also correct; we don't want duplicate parameters.

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -1021,9 +1021,7 @@ class Content:
                 ](index.data, target, self.length)
             )
 
-        return ak.contents.IndexedOptionArray.simplified(
-            index, self, parameters=self._parameters
-        )
+        return ak.contents.IndexedOptionArray.simplified(index, self)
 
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
Good catch; we use 64-bit so often that it's easy to miss cases that should work for 32-bit as well.

I'll merge it.

PR #2649, opened 2023-08-15
feat: add `__dlpack__`, `from_dlpack` support

### LINKED ISSUE #2648: Add support for DLPack

The most forward-looking device-agnostic (regular) array interchange specification is DLPack, which has support in JAX, NumPy, and CuPy. We should be able to ingest arrays that expose this interface, and perhaps also expose the `__dlpack__` and `__dlpack_device__` interfaces to allow interop with DLPack aware libraries.

agoose77:
This PR fixes #2648. 

- Adds `Array.__dlpack__(stream=None)` and `Array.__dlpack_device__()`
- Adds `ak.from_dlpack`
- Fixes copying from CUDA in `to_nplike`
- Replaces `Array.__array__` with `Array.__array_interface__` and `Array.__cuda_array_interface__` (as these are higher precedence, and symmetric to one-another. It's arbitrary, we could keep `__array__`, but I think `__array_interface__` should be fairly well supported).

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -1115,6 +1115,13 @@ class RecordArray(Content):
         contents = [x._to_backend_array(allow_missing, backend) for x in self._contents]
         if any(len(x.shape) != 1 for x in contents):
             raise ValueError(f"cannot convert {self} into np.ndarray")
+
+        if not backend.nplike.supports_structured_dtypes:
+            raise TypeError(
+                f"backend {backend.name!r} does not support structured "
+                f"dtypes required for converting {type(self).__name__} "
+                f"into a backend array"
+            )
         out = backend.nplike.empty(
             self.length,
             dtype=[(str(n), x.dtype) for n, x in zip(self.fields, contents)],
--- a/src/awkward/operations/ak_from_dlpack.py
+++ b/src/awkward/operations/ak_from_dlpack.py
@@ -1,25 +1,12 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 __all__ = ("from_dlpack",)
-from enum import IntEnum
 
+from awkward._connect.dlpack import DLPackDevice
 from awkward._dispatch import high_level_function
 from awkward._layout import from_arraylib, wrap_layout
 from awkward._nplikes.cupy import Cupy
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyLike
-
-
-class DLPackDevice(IntEnum):
-    CPU = 1  # CPU
-    CUDA = 2  # GPU
-    CUDA_PINNED = 3  # GPU & CPU
-    OPENCL = 4  # UNSUPPORTED
-    VULKAN = 7  # UNSUPPORTED
-    METAL = 8  # UNSUPPORTED
-    VPI = 9  # UNSUPPORTED
-    ROCM = 10  # GPU
-    ROCM_PINNED = 11  # GPU & CPU
-    CUDA_MANAGED = 13  # GPU & CPU
 
 
 @high_level_function()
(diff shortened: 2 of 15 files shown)


src/awkward/operations/ak_from_dlpack.py
```diff
@@ -0,0 +1,83 @@
...
+        # nplike = (Numpy if prefer_cpu else Cupy).instance()
+        nplike = Numpy.instance()
+    elif device_type == DLPackDevice.ROCM:
+        nplike = Cupy.instance()
+    elif device_type == DLPackDevice.ROCM_PINNED:
+        nplike = (Numpy if prefer_cpu else Cupy).instance()
+    elif device_type == DLPackDevice.CUDA_MANAGED:
+        nplike = (Numpy if prefer_cpu else Cupy).instance()
+    else:
+        raise AssertionError
```
agoose77, line 63: We *could* generalise this to include JAX. However, I think this is a lot simpler to reason about for now.

src/awkward/operations/ak_from_dlpack.py
```diff
@@ -0,0 +1,83 @@
...
+    CPU = 1  # CPU
+    CUDA = 2  # GPU
+    CUDA_PINNED = 3  # GPU & CPU
+    OPENCL = 4  # UNSUPPORTED
+    VULKAN = 7  # UNSUPPORTED
+    METAL = 8  # UNSUPPORTED
+    VPI = 9  # UNSUPPORTED
+    ROCM = 10  # GPU
+    ROCM_PINNED = 11  # GPU & CPU
+    CUDA_MANAGED = 13  # GPU & CPU
```
agoose77: This might need to live in `_nplikes` in future.

src/awkward/contents/recordarray.py
```diff
@@ -1115,6 +1115,13 @@ def _to_backend_array(self, allow_missing, backend):
         contents = [x._to_backend_array(allow_missing, backend) for x in self._contents]
         if any(len(x.shape) != 1 for x in contents):
             raise ValueError(f"cannot convert {self} into np.ndarray")
+
+        if not backend.nplike.supports_structured_dtypes:
+            raise TypeError(
+                f"backend {backend.name!r} does not support structured "
+                f"dtypes required for converting {type(self).__name__} "
+                f"into a backend array"
+            )
```
jpivarski, line 1124: You noticed this: okay, so maybe I didn't need to tell you.

jpivarski (approved):
This converts from DLPack arrays (`ak.from_dlpack`) and converts single-buffer Awkward Arrays to DLPack.

Just to be clear, that "single-buffer Awkward Arrays" is a proper subset of "Awkward Arrays that won't fail in `ak.to_numpy`":

```python
>>> array = ak.Array([
...     [{"x": 1, "y": 1.1}, {"x": 2, "y": 2.2}], 
...     [{"x": 3, "y": 3.3}, {"x": 4, "y": 4.4}],
... ])
>>> ak.to_numpy(array)    # also np.asarray(array)
array([[(1, 1.1), (2, 2.2)],
       [(3, 3.3), (4, 4.4)]], dtype=[('x', '<i8'), ('y', '<f8')])

>>> array = ak.Array([1.1, 2.2, None, 4.4, None, None, 7.7])
>>> ak.to_numpy(array)    # NOT np.asarray(array)
masked_array(data=[1.1, 2.2, --, 4.4, --, --, 7.7],
             mask=[False, False,  True, False,  True,  True, False],
       fill_value=1e+20)
```

In both of the above, `array` contains multiple buffers that might be on different GPUs in the `"cuda"` backend, so it's good that they'll fail to be converted to DLPack.

-------------

It looks good to me; go ahead and merge it if you don't have any last-minute changes.

src/awkward/contents/recordarray.py
agoose77: This is a cleanup more than a feature; NumPy _does_ support both DLPack and structured dtypes. However, NumPy doesn't support structured dtypes with DLPack, and will report that the conversion fails, e.g.
```pycon
>>> import awkward as ak
>>> array = ak.Array([{'x': 1}]).to_numpy()
>>> ptr = array.__dlpack__()
  File "...", line 5, in <module>
    ptr = array.__dlpack__()
          ^^^^^^^^^^^^^^^^^^
BufferError: DLPack only supports signed/unsigned integers, float and complex dtypes.
[('x', '<i8')]
```

PR #2650, opened 2023-08-15
feat: add support for `__arrow_array__`

### LINKED ISSUE #2647: Support `pyarrow.array(ak.Array(...))`

Arrow provides an `__arrow_array__` interface that we *could* support.

@jpivarski how do you feel about this? The questions that arise for me are - what defaults should we choose for `ak.to_arrow`? I'm mainly concerned about enabling the extension array or not. To my mind, the extension-array mechanism is useful for serialisation, but may cause problems for strict Arrow routines.

agoose77:
Fixes #2647

### DIFF
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -1116,9 +1116,9 @@ def form_handle_arrow(schema, pass_empty_field=False):
             return out
 
 
-def convert_to_array(layout, args, kwargs):
+def convert_to_array(layout, type=None):
     out = ak.operations.to_arrow(layout, extensionarray=False)
-    if args == () and kwargs == {}:
+    if type is None:
         return out
     else:
-        return pyarrow.array(out, *args, **kwargs)
+        return pyarrow.array(out, type=type)
--- a/src/awkward/_connect/pyarrow.py
+++ b/src/awkward/_connect/pyarrow.py
@@ -1114,3 +1114,11 @@ def form_handle_arrow(schema, pass_empty_field=False):
 
         else:
             return out
+
+
+def convert_to_array(layout, args, kwargs):
+    out = ak.operations.to_arrow(layout, extensionarray=False)
+    if args == () and kwargs == {}:
+        return out
+    else:
+        return pyarrow.array(out, *args, **kwargs)
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 
 __all__ = ("Array", "ArrayBuilder", "Record")
 
+import builtins
 import copy
 import html
 import io
@@ -1325,13 +1326,13 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         with ak._errors.OperationErrorContext("numpy.asarray", (self, *args), kwargs):
             return ak._connect.numpy.convert_to_array(self._layout, args, kwargs)
 
-    def __arrow_array__(self, *args, **kwargs):
+    def __arrow_array__(self, type=None):
         with ak._errors.OperationErrorContext(
-            f"{type(self).__name__}.__arrow_array__", (self, *args), kwargs
+            f"{builtins.type(self).__name__}.__arrow_array__", (self,), {"type": type}
         ):
             from awkward._connect.pyarrow import convert_to_array
 
-            return convert_to_array(self._layout, args, kwargs)
+            return convert_to_array(self._layout, type=type)
 
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
         """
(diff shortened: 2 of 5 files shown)


src/awkward/highlevel.py
```diff
@@ -1325,6 +1325,14 @@ def __array__(self, *args, **kwargs):
         with ak._errors.OperationErrorContext("numpy.asarray", (self, *args), kwargs):
             return ak._connect.numpy.convert_to_array(self._layout, args, kwargs)
 
+    def __arrow_array__(self, *args, **kwargs):
+        with ak._errors.OperationErrorContext(
+            f"{type(self).__name__}.__arrow_array__", (self, *args), kwargs
```
agoose77: I think we need to be clear that this is Awkward's code, not `pyarrow.array(...)`

jpivarski: Without this, the error message would show up in `ak.to_arrow`, since you're calling the high-level version of that and not its `_impl`. But users may be confused because they don't have any `ak.to_arrow` in their own code, so this helps.

src/awkward/_connect/pyarrow.py
```diff
@@ -1114,3 +1114,11 @@ def form_handle_arrow(schema, pass_empty_field=False):
...
         else:
             return out
+
+
+def convert_to_array(layout, args, kwargs):
+    out = ak.operations.to_arrow(layout, extensionarray=False)
+    if args == () and kwargs == {}:
+        return out
+    else:
+        return pyarrow.array(out, *args, **kwargs)
```
jpivarski: Your test passes, but I tried what ought to be the same thing and it didn't:

```python
>>> import pyarrow as pa
>>> a = pa.array([1.1, 2.2, 3.3, 4.4, 5.5])
>>> b = pa.array(a, type=pa.float64())
  File "pyarrow/array.pxi", line 317, in pyarrow.lib.array
  File "pyarrow/array.pxi", line 39, in pyarrow.lib._sequence_to_array
  File "pyarrow/error.pxi", line 144, in pyarrow.lib.pyarrow_internal_check_status
  File "pyarrow/error.pxi", line 100, in pyarrow.lib.check_status
pyarrow.lib.ArrowInvalid: Could not convert <pyarrow.DoubleScalar: 1.1> with type pyarrow.lib.DoubleScalar: tried to convert to double
```

(I was going to do a test to find out if wrapping an Arrow array with `pa.array` is zero-copy. It ought to be—I was just checking.)

It might be my pyarrow version (9.0.0): I checkout out this repo and tried to run the test and it failed (same error).

We currently support Arrow versions as low as 7.0.0. This is apparently a bug that they fixed in version $x$ such that $9 < x \le 12$. We could increase that threshold, except that Arrow is an important requirement for basic things like saving and reading files, and other packages have a tendency of putting upper bounds on the Arrow version. (`condastats` wanted to lower it to Arrow 6, which made it tricky to build an environment in which conda stats can actually be analyzed.)

We can't pass `pa.array` arguments into `to_arrow` because `to_arrow` doesn't usually call `pa.array` directly; it uses `pa.Array.from_buffers`, which doesn't take the same arguments. Arrow version $x$ is only needed for the `else` case of this `if`-`else`: maybe check the Arrow version there? If it's less than $x$, raise an error message saying that passing arguments requires Arrow version $x$?

Or maybe that's overkill and we should just explain that it's an Arrow bug, fixed in version $x$, when our users encounter this. It would at least require an Arrow version check in the test, so that we can run the tests with lower versions of Arrow. (And maybe that needs to be added to the CI matrix, not just NumPy minimum version, but Arrow, too.)

jpivarski: By the way, my choice would be for the last one: just ensure that our tests work with the Arrow versions we say we support, even if the Arrow version we support has known bugs when some (not all) operations are attempted.

jpivarski (changes requested):
I approve of the PR and how it's implemented, but something needs to be done about the Arrow bug it triggers in old versions of Arrow (see below).

src/awkward/_connect/pyarrow.py
agoose77: Good spot; I assumed that this part of the Arrow API was more stable than this. It looks like only 12.0.0 supports the casting that is performed by the test case in question. I've made it `skipif`

jpivarski: Some of the other tests weren't working for me, so I've added a minimum pyarrow to the build matrix. We don't depend on it quite as much as NumPy, but increasingly, and it has more radical changes from one version to the next.

agoose77: That's a good idea. Thanks!

jpivarski: Two of these failures are the new tests I was expecting, but some others are from pyarrow string functions. I'm going to test those locally now.

https://github.com/scikit-hep/awkward/actions/runs/5871689180/job/15921684610?pr=2650

jpivarski: They're all fine in my pyarrow 9.0.0, but apparently not in the pyarrow 7.0.0 that we say we support. It was good to test!

They're encoding mistakes. I can fix the string functions (in this PR) while you fix the casting.

jpivarski (approved):

src/awkward/_connect/pyarrow.py
jpivarski: Done!

jpivarski:
I think this one will pass because the only error in your last commit was in the strings.

Feel free to merge when you're ready!

PR #2651, opened 2023-08-16
fix: support non-hashable values in `parameters_union`

### LINKED ISSUE #340: `events[boolean mask]` computes fail with `scheduler="processes"` or a distributed `Client`
```python
import dask_awkward as dak
from coffea.nanoevents import NanoAODSchema, NanoEventsFactory
from distributed import Client

fnames = {"../../coffea_dev/coffea/tests/samples/nano_dy.root": "Events"}

events = NanoEventsFactory.from_root(
    fnames,
    schemaclass=NanoAODSchema,
    permit_dask=True,
).events()

mask = dak.num(events.Electron.pt) >= 1

(events[mask]).compute(scheduler="threads")

(events[mask]).compute(scheduler="processes")

with Client() as client:
    (events[mask]).compute()
```
From the above snippet, only the first compute works. The other two fail with
```
TypeError: unhashable type: 'dict'

    ak.to_packed(
        <Array [{PuppiMET: {...}, ...}, ..., {...}] type='20 * NanoEvents[P...'>
        highlevel = False
    )
```
This  happens in `awkward>=2.3.0` and is independent of `dask-awkward` or `uproot` version. However it is dependent on `coffea`'s nanoevents. I wasn't able to reproduce this if events are read with plain `uproot`.

I don't know which is the "correct" package to post the issue so I'm tagging @lgray and awkward people @jpivarski @agoose77.
If you'd like to move this issue, let me know.

I would also like to note that in my code (https://github.com/iasonkrom/egamma-tnp/blob/6e379ebb784869bf38f09f91f2b790aa91469b69/src/egamma_tnp/_tnpmodules.py#L37) where more things happen afterwards during compute, this fails only with `scheduler="processes"` and works with the distributed `Client`.

agoose77:
Fixes dask-contrib/dask-awkward#340

### DIFF
--- a/src/awkward/_parameters.py
+++ b/src/awkward/_parameters.py
@@ -95,7 +95,7 @@ def parameters_intersect(
         return None
 
     common_keys = iter(left.keys() & right.keys())
-    has_exclusions = bool(exclude)
+    has_exclusions = exclude is not None and len(exclude) > 0
 
     # Avoid creating `result` unless we have to
     result = None
@@ -132,7 +132,7 @@ def parameters_union(
     Returns the merged key-value pairs of `left` and `right` as a dictionary.
 
     """
-    has_exclusions = bool(exclude)
+    has_exclusions = exclude is not None and len(exclude) > 0
     items = []
     if left is not None:
         items.append(left.items())
--- a/src/awkward/_parameters.py
+++ b/src/awkward/_parameters.py
@@ -1,9 +1,8 @@
 from __future__ import annotations
 
-from collections.abc import Collection
 from itertools import chain
 
-from awkward._typing import Any, JSONMapping, JSONSerializable, Set
+from awkward._typing import JSONMapping, Set
 
 TYPE_PARAMETERS = ("__array__", "__list__", "__record__", "__categorical__")
 
@@ -81,13 +80,13 @@ def parameters_intersect(
     left: JSONMapping | None,
     right: JSONMapping | None,
     *,
-    exclude: Collection[tuple[str, JSONSerializable]] = (),
+    exclude: Set[str] = frozenset(),
 ) -> JSONMapping | None:
     """
     Args:
         left: first parameters mapping
         right: second parameters mapping
-        exclude: collection of (key, value) items to exclude
+        exclude: collection of keys to exclude
 
     Returns the intersected key-value pairs of `left` and `right` as a dictionary.
     """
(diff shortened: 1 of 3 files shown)


ikrommyd:
@agoose77 I pip installed from this branch and my reproducer on https://github.com/dask-contrib/dask-awkward/issues/340 and my code on https://github.com/iasonkrom/egamma-tnp work fine now. Looks good from my perspective but that's up to you to judge. Thanks!

src/awkward/_parameters.py
```diff
@@ -81,21 +81,21 @@ def parameters_intersect(
     left: JSONMapping | None,
     right: JSONMapping | None,
     *,
-    exclude: Collection[tuple[str, JSONSerializable]] = (),
+    exclude: Collection[tuple[str, Any]] | None = None,
```
jpivarski: The values of the parameters still need to be JSON serializable. I'm not sure how much that type hint would enforce it, though.

src/awkward/_parameters.py
```diff
@@ -81,21 +81,21 @@ def parameters_intersect(
...
+        exclude: collection of keys to exclude
 
     Returns the intersected key-value pairs of `left` and `right` as a dictionary.
     """
     if left is None or right is None:
         return None
 
     common_keys = iter(left.keys() & right.keys())
-    has_no_exclusions = len(exclude) == 0
+    has_exclusions = bool(exclude)
```
jpivarski: `bool(x)` is a rather brief way to say `len(x) != 0`. It works through Python's truthiness, but I would prefer it if Python required us to be more explicit. Nevertheless, this does look like something a flake8-style linter would suggest. `:)`

src/awkward/_parameters.py
```diff
@@ -130,7 +132,7 @@ def parameters_union(
     Returns the merged key-value pairs of `left` and `right` as a dictionary.
 
     """
-    has_exclusions = len(exclude) > 0
+    has_exclusions = bool(exclude)
```
jpivarski: Another use of `bool(x)` to mean `len(x) != 0`.

Some languages have an `is_empty` operator or function to be very explicit about it, but then people end up comparing `length` with zero anyway, maybe because they don't know about `is_empty`. In sufficiently strict type systems, `is_empty` is a different type than `length` and a compiler can use that...

jpivarski (approved):
It looks great! Merge it whenever you're ready.

src/awkward/_parameters.py
agoose77: Oops, we should keep that.

src/awkward/_parameters.py
agoose77: Hehe, I wondered if you'd mind this given that linters generally prefer it. Here, I want to mean both `x is not None` or `len(x) != 0`

jpivarski: I prefer `x is not None` and also `len(x) != 0` over `bool(x)`, but now that we're talking

```python
x is not None and len(x) != 0
```

Well, that is a lot more characters than

```python
bool(x)
```

~~I guess it's okay.~~ You know, the fact that I didn't realize that `x` might be `None` by reading the code is an argument in favor of the long expression. I know that what type hints are for, but the type hint is far away and long-time Python users have gotten used to looking for clues about a variable's type from nearby context. So I still kinda prefer the long expression, as long as black doesn't split it up onto multiple lines. (I don't think it would.)

PR #2652, opened 2023-08-16
fix: support placeholders in shape-only routines

agoose77:
When dask-awkward used its column optimisation to generate placeholder arrays, we were seeing placeholders in some `nplike` functions that were not prepared for them. This is because these nplike functions explicitly don't touch data for typetracers. 

This fix won't solve the re-hydration issues by itself, but it should get us closer and is fairly easy to reason about.

### DIFF
--- a/src/awkward/_nplikes/jax.py
+++ b/src/awkward/_nplikes/jax.py
@@ -79,6 +79,6 @@ class Jax(ArrayModuleNumpyLike):
     def strides(self, x: ArrayLike) -> tuple[ShapeItem, ...]:
         x.touch_shape()
         out = (x._dtype.itemsize,)
-        for item in reversed(x._shape):
+        for item in x._shape[-1:0:-1]:
             out = (item * out[0], *out)
         return out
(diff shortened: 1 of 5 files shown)


agoose77:
Amongst other things, this PR adds `nbytes` as a part of the `ArrayLike` protocol. `TypeTracerArray` will use the shape to return a maybe-unknown number of bytes (length), whilst `PlaceholderArray` returns 0. This is because placeholder arrays are explicitly data-less, whilst typetracers are stand-in arrays.

src/awkward/_nplikes/jax.py
```diff
@@ -80,6 +80,6 @@ def ascontiguousarray(self, x: ArrayLike) -> ArrayLike:
     def strides(self, x: ArrayLike) -> tuple[ShapeItem, ...]:
         x.touch_shape()
         out = (x._dtype.itemsize,)
-        for item in reversed(x._shape):
+        for item in x._shape[-1:0:-1]:
```
jpivarski, line 83: This is not just style: it's a change in behavior. Was the original wrong?

We talked about this on Zoom, and yes—it used to be wrong. Now it's right.

jpivarski (approved):
Looks good to me!

PR #2653, opened 2023-08-16
fix: remove unneeded cast

agoose77:
I noticed a bug during slicing for a record of ragged lists. This PR fixes the bug, which was introduced during the backend addition.

### DIFF
--- a/src/awkward/contents/listoffsetarray.py
+++ b/src/awkward/contents/listoffsetarray.py
@@ -417,7 +417,6 @@ class ListOffsetArray(Content):
         tail: tuple[SliceItem, ...],
         advanced: Index | None,
     ) -> Content:
-        advanced = advanced.to_nplike(self._backend.nplike)
         if head is NO_HEAD:
             return self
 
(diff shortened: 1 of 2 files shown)


src/awkward/contents/listoffsetarray.py
```diff
@@ -417,7 +417,6 @@ def _getitem_next(
         tail: tuple[SliceItem, ...],
         advanced: Index | None,
     ) -> Content:
-        advanced = advanced.to_nplike(self._backend.nplike)
         if head is NO_HEAD:
             return self
```
jpivarski, line 421:
```suggestion
        if ak._util.win:
            advanced = advanced.to_nplike(self._backend.nplike)
        if head is NO_HEAD:
            return self
```

It seems that only Windows needs this cast. (Maybe, depending on what the other Windows tests would have done if they had been allowed to run, and maybe the test would fail on Windows even without this line and we're looking at something else.)

https://github.com/scikit-hep/awkward/actions/runs/5884127678/job/15958232220?pr=2653

jpivarski (approved):
This will be ready to go as soon as the tests pass. I'm going to try my own suggestion to see if that's it.

src/awkward/contents/listoffsetarray.py
agoose77: I think I've fixed it - oversight on my part :)

jpivarski: In that case, I'll enable auto-merge. Disable it if you see something else.

PR #2654, opened 2023-08-17
fix: multiple-output ufuncs e.g. `divmod`

agoose77:
It turns out that we didn't yet test the `nout != 1` ufuncs, such as `divmod`. 
```python
>>> import awkward as ak
>>> import numpy as np
>>> array = ak.Array([
...     [1, 2, 3],
...     [4]
... ])
>>> np.divmod(array, 2)
[[[0, 1, 1, 2], [1, 0, 1, 0]], []]
```
This PR adds support for this ufunc, with some refactoring work along the way.
```python
>>> import awkward as ak
>>> import numpy as np
>>> array = ak.Array([
...     [1, 2, 3],
...     [4]
... ])
>>> np.divmod(array, 2)
(<Array [[0, 1, 1], [2]] type='2 * var * int64'>, <Array [[1, 0, 1], [0]] type='2 * var * int64'>)
```

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -57,20 +57,27 @@ class BroadcastOptions(TypedDict):
 
 
 def length_of_broadcast(inputs: Sequence) -> int | type[unknown_length]:
-    maxlen = 1
-
+    max_length: int | None = None
     has_seen_unknown_length: bool = False
     for x in inputs:
-        if isinstance(x, Content):
-            if x.length is unknown_length:
-                has_seen_unknown_length = True
-                continue
-            maxlen = max(maxlen, x.length)
+        if not isinstance(x, Content):
+            continue
+        if x.length is unknown_length:
+            has_seen_unknown_length = True
+        elif max_length is None:
+            max_length = x.length
+        else:
+            max_length = max(max_length, x.length)
 
     if has_seen_unknown_length:
-        return unknown_length
+        if max_length is None:
+            return unknown_length
+        else:
+            return max_length
+    elif max_length is None:
+        return 1
     else:
-        return maxlen
+        return max_length
 
 
 def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -106,14 +106,14 @@ def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
     return nextinputs
 
 
-def broadcast_unpack(x, isscalar: list[bool], backend: Backend):
+def broadcast_unpack(x, isscalar: list[bool]):
     if all(isscalar):
-        if not backend.nplike.known_data or x.length == 0:
+        if x.length is not unknown_length and x.length == 0:
             return x._getitem_nothing()._getitem_nothing()
         else:
             return x[0][0]
     else:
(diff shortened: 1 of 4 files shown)


agoose77:
@jpivarski in this PR so-far, I removed the unary pathway for ufuncs. I can't see why this was introduced in 0e0202a5cf54b5dbc2606861d058665b8d90fa07

It might just be that our broadcasting code has changed since then to make the need redundant. Are you OK with the removal?

src/awkward/_broadcasting.py
```diff
@@ -57,20 +57,27 @@ class BroadcastOptions(TypedDict):
...
-        return unknown_length
+        if max_length is None:
+            return unknown_length
+        else:
+            return max_length
+    elif max_length is None:
+        return 1
     else:
-        return maxlen
+        return max_length
```
agoose77, line 80: This code should now return a concrete length in the `unknown-length` - `int` case.

src/awkward/_broadcasting.py
```diff
@@ -99,14 +106,14 @@ def broadcast_pack(inputs: Sequence, isscalar: list[bool]) -> list:
...
+def broadcast_unpack(x, isscalar: list[bool]):
     if all(isscalar):
-        if not backend.nplike.known_data or x.length == 0:
+        if x.length is not unknown_length and x.length == 0:
             return x._getitem_nothing()._getitem_nothing()
         else:
             return x[0][0]
     else:
-        if not backend.nplike.known_data or x.length == 0:
+        if x.length is not unknown_length and x.length == 0:
```
agoose77, line 116: TypeTracer should be smart enough to just work here, now.

src/awkward/_connect/numpy.py
```diff
@@ -420,41 +420,14 @@ def action(inputs, **ignore):
...
-                assert isinstance(result, tuple) and len(result) == 1
-                return result[0]
-
-        out = ak._do.recursively_apply(
-            inputs[where],
-            unary_action,
-            behavior,
-            function_name=ufunc.__name__,
-            allow_records=False,
-        )
```
jpivarski, line 448: > @jpivarski in this PR so-far, I removed the unary pathway for ufuncs. I can't see why this was introduced in [0e0202a](https://github.com/scikit-hep/awkward/commit/0e0202a5cf54b5dbc2606861d058665b8d90fa07)

That commit is a PR merge; for finer granularity, the commit within that PR is

https://github.com/scikit-hep/awkward/pull/1183/commits/5b0a273e807a8b25ba8b02d2651af28f267823d7#diff-e7b6632c2ca4228bf21505ced52be60d4fda8813e00cbaf65cd1c5d2e228d721R242-R263

It was accompanied by a large batch of single-line assertions, some of which have only one `Content` in the ufunc (such as `+` with one `Content` and one Python number).

Broadcasting a single layout should be equivalent to `recursively_apply` on that one layout. The fact that numerous tests were probably going through that and they passed either way is good evidence that the "should be" holds. (I don't see anything special about it, that it has to be there, and the context of the commit isn't reminding me of any gotchas.)

jpivarski (approved):
Thanks for catching this! At one point, the return value for broadcasting actions had to be promoted from a single `Content` to a tuple of `Contents`, and I think it was because of ufuncs that return multiple values. But when that went it, it should have gone in with a test, like this one.

If you're done with this PR, go ahead and merge it.

PR #2659, opened 2023-08-18
fix: using `numba.core.errors.Numba<Error>` instead of `Error` in a Numba typing context.

jpivarski:
This is the only update needed to pass tests with [Numba 0.58.0rc1](https://numba.discourse.group/t/ann-numba-0-58-0rc1-and-llvmlite-0-41-0rc1/2078). However, that version of Numba still raises a warning if

```bash
export NUMBA_CAPTURED_ERRORS=new_style
```

is not explicitly set (should probably be changed in Numba itself). Most of the exceptions changed were `TypeErrors`, but a few were `ValueErrors` that really ought to have been type errors. (Not finding a record field by name is a type error in compiled code.)

[Vector](https://github.com/scikit-hep/vector) is already up-to-date with respect to `TypeError` versus `numba.TypingError`.

### DIFF
--- a/src/awkward/_connect/numba/arrayview.py
+++ b/src/awkward/_connect/numba/arrayview.py
@@ -5,6 +5,7 @@ import numba
 import numba.core.typing
 import numba.core.typing.ctypes_utils
 import numpy
+from numba.core.errors import NumbaTypeError
 
 import awkward as ak
 from awkward._behavior import behavior_of, overlay_behavior
@@ -420,7 +421,7 @@ class type_getitem(numba.core.typing.templates.AbstractTemplate):
                     viewtype, wheretype
                 )
             else:
-                raise numba.TypingError(
+                raise NumbaTypeError(
                     "only an integer, start:stop range, or a *constant* "
                     "field name string may be used as ak.Array "
                     "slices in compiled code"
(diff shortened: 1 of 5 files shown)


agoose77:
From [the deprecation notice,](https://numba.readthedocs.io/en/0.58.0rc1/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors), it seems to me that we should be using the Numba-aliased errors. I replaced our changes to `NumbaTypingError` with the appropriate specific error class. Hopefully this works ...

I also therefore am mixing import styles, but I think in time we are moving to `from XXX import YYY`.

jpivarski:
I'm not really sure what the intended difference is between `nb.TypingError` and `nb.NumbaTypeError`. They both participate in the new-style of `NUMBA_CAPTURED_ERRORS`:

```python
>>> issubclass(numba.TypingError, numba.NumbaError)
True
>>> issubclass(numba.NumbaTypeError, numba.NumbaError)
True
```

Oh, it turns out that `nb.NumbaTypeError` is just a subclass of `nb.TypingError`.

```python
>>> numba.TypingError.mro()
[<class 'numba.core.errors.TypingError'>, <class 'numba.core.errors.NumbaError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>]
>>> numba.NumbaTypeError.mro()
[<class 'numba.core.errors.NumbaTypeError'>, <class 'numba.core.errors.TypingError'>, <class 'numba.core.errors.NumbaError'>, <class 'Exception'>, <class 'BaseException'>, <class 'object'>]
```

I'm on board with `from XXX import YYY`, and in that style, `NumbaTypeError` is less ambiguous than `TypingError` would be. So I think these are fine counter-edits.

@ianna, what do you think?

ianna (approved):
I like it, it is more expressive. Thanks!

jpivarski:
Okay, then I'll merge!

PR #2660, opened 2023-08-18
feat: `Form.expected_from_buffers` for names/dtypes `ak.from_buffers` needs.

jpivarski:
Requested by @danielballan in https://github.com/bluesky/tiled/pull/549#issuecomment-1682987096.

### DIFF
--- a/src/awkward/contents/emptyarray.py
+++ b/src/awkward/contents/emptyarray.py
@@ -24,8 +24,9 @@ from awkward.index import Index
 if TYPE_CHECKING:
     from awkward._slicing import SliceItem
 
-np = NumpyMetadata.instance()
 numpy = Numpy.instance()
+
+np = NumpyMetadata.instance()
 
 
 @final
--- a/src/awkward/forms/bitmaskedform.py
+++ b/src/awkward/forms/bitmaskedform.py
@@ -251,3 +251,9 @@ class BitMaskedForm(Form):
                 parameters=parameters,
                 form_key=form_key,
             )
+
+    def _expected_from_buffers(self, getkey):
+        from awkward.operations.ak_from_buffers import _index_to_dtype
+
+        yield (getkey(self, "mask"), _index_to_dtype[self._mask])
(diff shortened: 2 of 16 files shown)


agoose77:
This looks pretty good! 

I'm not yet sure whether we want this logic to be implemented by the `Form` classes, though. My thinking is thus;

- We want to separate the concerns of `Form` and `Content`.
- `Content` handles `to_buffers`, and a top-level routine handles `from_buffers`

As such, we should probably have a top-level routine that handles `expected_from_buffers`. 

I'm also thinking about whether `Form.expected_from_buffers` should instead be accessed via `ak.expected_from_buffers`. We do have an array-aware method on `Form`: `Form.length_zero_array`, but that perhaps should be the exception rather than the rule.

I can make these changes — at this stage it's just musings.

jpivarski:
This kind of functionality should be level-2: public but not widely advertised. I don't think it should be a high-level `ak.*` function, for instance.

It could be a free function in the `ak.forms` namespace, like `ak.forms.from_dict` and `ak.forms.from_json`. Nothing in the name `expected_from_buffers` says that it's about Forms, rather than layouts, so

```python
ak.forms.expected_from_buffers(some_form)
```

would accomplish that as well as

```python
some_form.expected_from_buffers
```

I'm not actually in favor of the following, but just to consider all options:

```python
ak.from_buffers.expected(some_form)
```

agoose77:
> I don't think it should be a high-level ak.* function, for instance.

There're probably too many constraints here to make everything work. I've changed my mind; although `from_buffers` is written in the `ak_` namespace, there's nothing to suggest that we can't encode a canonical form representation in `ak.forms`, which is what this does. I'm +1, although I'd therefore like to move the functions that are imported from `ak_from_buffers` to `ak.forms` itself (my todo).

agoose77:
@jpivarski could you OK (nOK) my counter edits?

jpivarski:
> @jpivarski could you OK (nOK) my counter edits?

Okay!

PR #2661, opened 2023-08-18
feat: allow `form_key` to round-trip through `length_zero_array`

jpivarski:
**Note:** see below; this _deprecates_ `form_key != None` on `EmptyForm` and immediately changes some of the ways that `EmptyForm` works with `form_keys`. @agoose77, @douglasdavis, and @lgray probably want to know about this and weigh in on whether it will be a problem.

--------------------

This started as an observation that `UnionForm.simplified` is wrong because it strips the `form_keys` from all of its `contents` (in https://github.com/bluesky/tiled/pull/549#discussion_r1298819704). That's because it's implemented by a round-trip through length-zero `UnionArray.simplified`. (The `UnionArray.simplified` logic is too complex to replicate in `UnionForm`. It would be a maintenance burden to try to do so.)

I know that we've struggled with this aspect of `Form.length_zero_array` in the past: we occasionally need it to simulate what should happen to a Form, but the Form is carrying important `form_keys` that we don't want to just drop. Repopulating the round-tripped Form with the original `form_keys` is a hassle and error-prone.

Therefore, this PR introduces a new parameter, `"__form_key__"`, to carry `form_keys` through temporary arrays made by `Form.length_zero_array` and `Form.length_one_array`. It's opt-in.

```python
>>> from awkward.forms import *
>>> form = ListOffsetForm(
...     "i64",
...     RecordForm(
...         [
...             NumpyForm("float64", form_key="node2"),
...             ListOffsetForm(
...                 "i64", NumpyForm("int64", form_key="node4"), form_key="node3"
...             ),
...         ],
...         ["x", "y"],
...         form_key="node1",
...     ),
...     form_key="node0",
... )
>>> layout = form.length_zero_array(form_keys_to_parameters=True, highlevel=False)
>>> layout
<ListOffsetArray len='0'>
    <parameter name='__form_key__'>'node0'</parameter>
    <offsets><Index dtype='int64' len='1'>[0]</Index></offsets>
    <content><RecordArray is_tuple='false' len='0'>
        <parameter name='__form_key__'>'node1'</parameter>
        <content index='0' field='x'>
            <NumpyArray dtype='float64' len='0'>
                <parameter name='__form_key__'>'node2'</parameter>
                []
            </NumpyArray>
        </content>
        <content index='1' field='y'>
            <ListOffsetArray len='0'>
                <parameter name='__form_key__'>'node3'</parameter>
                <offsets><Index dtype='int64' len='1'>[0]</Index></offsets>
                <content><NumpyArray dtype='int64' len='0'>
                    <parameter name='__form_key__'>'node4'</parameter>
                    []
                </NumpyArray></content>
            </ListOffsetArray>
        </content>
    </RecordArray></content>
</ListOffsetArray>
```

Converting all of the `"__form_key__"` parameters back into `form_keys` (as opposed to `parameters` on the Form) is also opt-in, but a library developer would know about both by following examples:

```python
>>> print(layout.form)
{
    "class": "ListOffsetArray",
    "offsets": "i64",
    "content": {
        "class": "RecordArray",
        "fields": [
            "x",
            "y"
        ],
        "contents": [
            {
                "class": "NumpyArray",
                "primitive": "float64",
                "parameters": {
                    "__form_key__": "node2"
                }
            },
            {
                "class": "ListOffsetArray",
                "offsets": "i64",
                "content": {
                    "class": "NumpyArray",
                    "primitive": "int64",
                    "parameters": {
                        "__form_key__": "node4"
                    }
                },
                "parameters": {
                    "__form_key__": "node3"
                }
            }
        ],
        "parameters": {
            "__form_key__": "node1"
        }
    },
    "parameters": {
        "__form_key__": "node0"
    }
}
>>> print(layout.form_with_key_from_parameter)
{
    "class": "ListOffsetArray",
    "offsets": "i64",
    "content": {
        "class": "RecordArray",
        "fields": [
            "x",
            "y"
        ],
        "contents": [
            {
                "class": "NumpyArray",
                "primitive": "float64",
                "form_key": "node2"
            },
            {
                "class": "ListOffsetArray",
                "offsets": "i64",
                "content": {
                    "class": "NumpyArray",
                    "primitive": "int64",
                    "form_key": "node4"
                },
                "form_key": "node3"
            }
        ],
        "form_key": "node1"
    },
    "form_key": "node0"
}
```

And then there's the issue that `EmptyArrays` don't have `parameters`. For this mechanism to work on all node types, `EmptyArrays` can't have `form_keys`, either. That's okay: they can't have buffers, so they don't really need `form_keys`.

So this PR makes the following immediate changes, which I think are safe:

* `EmptyForm.__eq__` ignores the `form_key` when determining if two Forms are equal.
* `ak.forms.from_dict` (and `from_json`) ignores `form_key` when constructing `EmptyForm` from dicts (or JSON). It's important to allow the dict (or JSON) format continue to have `"parameters"` and `"form_key"` keys, for format version independence/schema evolution, but we no longer read these fields.
* `EmptyForm.to_dict` (and `to_json`) outputs `"form_key": None` if `verbose=True`, rather than outputting its actual `form_key`.
* All of the functions in Awkward Array that create `EmptyForms` create them without `form_keys`. In particular, `Content.form_with_key` _skips a number_ when constructing `form_key` names from `f"node{number}"`. This is so that old Forms that might be saved somewhere in JSON format don't have to be modified in a major way (changing all the numbers after the first `EmptyForm`) to get the same `form_keys` as the new `Content.form_with_key` outputs.

It also deprecates intentional, explicit passing of `form_key != None` in

* the `EmptyArray` constructor
* the `EmptyArray.copy` method
* the `EmptyArray.simplified` constructor

The deprecation warning turns into an error in Awkward Array **2.5.0** (much as non-empty `parameters` were removed in Awkward Array 2.2.0).

### DIFF
--- a/src/awkward/_connect/avro.py
+++ b/src/awkward/_connect/avro.py
@@ -222,7 +222,7 @@ class ReadAvroFT:
         if file["type"] == "null":
             aform = ak.forms.IndexedOptionForm(
                 "i64",
-                ak.forms.EmptyForm(form_key=f"node{form_next_id+1}"),
+                ak.forms.EmptyForm(),
                 form_key=f"node{form_next_id}",
             )
             declarations.append(f"output node{form_next_id+1}-data uint8 \n")
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -192,6 +192,31 @@ class Content:
     def form(self) -> Form:
         return self.form_with_key(None)
 
+    @property
+    def form_with_key_from_parameter(self) -> Form:
+        def getkey(layout):
+            return layout.parameter("__form_key__")
+
+        def remove_parameter(form):
+            if form._parameters is not None and "__form_key__" in form._parameters:
+                if len(form._parameters) == 1:
+                    form._parameters = None
+                else:
+                    form._parameters = {
+                        k: v for k, v in form._parameters.items() if k != "__form_key__"
+                    }
+
+            if hasattr(form, "_content"):
+                remove_parameter(form._content)
+
+            elif hasattr(form, "_contents"):
+                for x in form._contents:
+                    remove_parameter(x)
+
+        out = self._form_with_key(getkey)
+        remove_parameter(out)
+        return out
+
     def form_with_key(
         self, form_key: str | None | Callable = "node{id}", id_start: int = 0
     ) -> Form:
--- a/src/awkward/contents/emptyarray.py
+++ b/src/awkward/contents/emptyarray.py
@@ -101,7 +101,8 @@ class EmptyArray(Content):
         return cls(backend=backend)
 
     def _form_with_key(self, getkey: Callable[[Content], str | None]) -> EmptyForm:
-        return self.form_cls(form_key=getkey(self))
+        getkey(self)  # skip a number but don't add a form_key to EmptyForm
+        return self.form_cls()
 
     def _to_buffers(
         self,
--- a/src/awkward/forms/emptyform.py
+++ b/src/awkward/forms/emptyform.py
@@ -19,6 +19,8 @@ class EmptyForm(Form):
     def __init__(self, *, parameters: JSONMapping | None = None, form_key=None):
         if not (parameters is None or len(parameters) == 0):
             raise TypeError(f"{type(self).__name__} cannot contain parameters")
+        if form_key is not None:
+            deprecate(f"{type(self).__name__} cannot have a form_key", version="2.5.0")
         self._init(parameters=parameters, form_key=form_key)
 
     def copy(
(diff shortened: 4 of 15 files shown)


jpivarski:
Oh, and 59536896d9af7807d9855062b66824409393480e renames `Index._form_to_zero_length` to `Index._form_to_length_zero` for consistency with the other methods. This is a hidden name, so it should be safe to change, but I just want to double-check: @douglasdavis and @lgray, you're not calling this directly, right?

agoose77 (commented):
This is an important use case to satisfy, but we discussed this earlier today, and realised that we need to rethink the parameter mechanism because of the parameter merging rules. As such, I'm closing this, and we will open an issue. As for the `EmptyForm` changes, this will be another PR!

PR #2662, opened 2023-08-21
fix: raise error for invalid object in `from_dlpack`

agoose77:
When I wrote this branch in `from_dlpack`, I think I was considering support for the PyCapsule legacy interface. Apparently I never restored an error case.

Whilst we're at it, this PR also adds support for conversion of index objects to DLPack.


jpivarski (approved):
This looks good, and also a simple enough fix for me to merge right away. Thanks!

PR #2663, opened 2023-08-21
fix: fix/harden broadcasting through tuples

agoose77:
At some point, refactoring work broke broadcasting of tuples whereby the broadcast proceeds through the fields. 

This PR fixes that, and also prohibits broadcasting of tuples against records, even if the field names are considered compatible (`"0"` in a record has a mapping onto tuple slot 0). I think we want to meaningfully distinguish between records and tuples, so I made this change.


jpivarski (approved):
Thanks for catching that! This looks good and I'll merge it now.

PR #2665, opened 2023-08-22
feat: support out-of-band buffers in pickling

agoose77:
Pickle 5 supports out-of-band data, which supports zero-copy round-trips with pickle.


agoose77:
This will break with placeholder arrays, but that's probably fine - we can either fix it in future, or explicitly ban placeholders in pickling (in lieu of dask serialisation).

jpivarski (approved):
I'm glad that it was this easy!

PR #2670, opened 2023-08-24
fix: update string to search for in determing parquet column list separator

douglasdavis:
With pyarrow 13 it appears that `list.element` is used over `list.item` in some datasets. Looks like our original check to look for `list.element` was too strict (the final `.`). We also need to make this change in dask-awkward!

more info at https://github.com/dask-contrib/dask-awkward/issues/346

### DIFF
--- a/src/awkward/operations/ak_from_parquet.py
+++ b/src/awkward/operations/ak_from_parquet.py
@@ -115,7 +115,7 @@ def metadata(
     for column_metadata in parquetfile_for_metadata.schema:
         if (
             column_metadata.max_repetition_level > 0
-            and ".list.element." in column_metadata.path
+            and ".list.element" in column_metadata.path
         ):
             list_indicator = "list.element"
             break


jpivarski:
This distinction between `item` and `element` was a difference in convention between Arrow and Spark (and Spark was more technically correct: there was some third party that said that `element` is what should be used). This could be a part of a transition toward conformity.

Anyway, these are the only two names that need to be supported.

agoose77:
I am happy to merge this, unless @jpivarski has objections, I'll do so this evening.

jpivarski (approved):
Now that I'm back at a computer, rather than a phone, I noticed that this is a PR on Awkward, not dask-awkward.

The change appears to be all-inclusive. If `".list.element" in column_metadata.path` is needed to match new Arrow arrays and `".list.element." in column_metadata.path` was used before, the new predicate will match both. Also, it couldn't spuriously match something it's not intended to.

So I sign off on it as well, and since we're all in agreement, I'll merge it, too.

PR #2672, opened 2023-08-25
feat: add `touch_data` to `ak.typetracer`

### LINKED ISSUE #2668: Can `_touch_data` be made public?

I have a [use case](https://github.com/bluesky/tiled/pull/549/files#diff-40070db129d1d3432609597f627cdbcddf5f970b7bb8062862427e8501359ab1R32-R34) for calling the [`_touch_data`](https://github.com/search?q=repo%3Ascikit-hep%2Fawkward%20%22def%20_touch_data%22&type=code) methods directly. I wonder if that method can be promoted from a private method to "public but not widely advertised".

If the maintainers are receptive, I would be happy to contribute a PR. It would be easy enough to provide a backward-compatible implementation:

```py
    def touch_data(self):
        ...

    _touch_data = touch_data  # back-compat
```

if that is an important consideration here.

agoose77:
This was requested in #2668, and I've identified that Coffea and dask-awkward currently use our private APIs. We should just make this public. 

At this stage, I'd prefer to expose this via `ak.typetracer`, rather than `Content.touch_data`, as we can better control the public API (e.g. always recursive, return `self`), and most people are already working with `ak.Array` objects.

Closes #2668

### DIFF
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -423,3 +423,11 @@ def sort(
     starts = ak.index.Index64.zeros(1, nplike=layout.backend.index_nplike)
     parents = ak.index.Index64.zeros(layout.length, nplike=layout.backend.index_nplike)
     return layout._sort_next(negaxis, starts, parents, 1, ascending, stable)
+
+
+def touch_data(layout: Content, recursive: bool = True):
+    layout._touch_data(recursive)
+
+
+def touch_shape(layout: Content, recursive: bool = True):
+    layout._touch_shape(recursive)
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
Great!

I think that Coffea was also using `_touch_data` temporarily, waiting for some version of it to become public, right @lgray? If so, then this will also serve that purpose.

lgray:
Yes - there are a few places where we still call `_touch_data` though `length_zero_if_typetracer` got rid of most of them.

danielballan:
Thank you!

PR #2678, opened 2023-08-29
fix: guard `broadcast_and_apply` from mixed backends

agoose77:
This PR ensures that `broadcast_and_apply` throws an error if the given arrays do not share the same backend. 

At a high level, all `ak.XXX` operations that accept multiple array arguments should effectively compute
```python
backend = backend_of(*args, default=..., coerce_to_common=True)
layouts = [ak.to_layout(x).to_backend(backend) for x in args]
```

We _could_ add a helper to do this, although given that each `ak.XXX` function needs to set different layout conversion flags, this is probably best avoided.

There's an argument to be made for making `broadcast_and_apply` be more permissive and perform the coercion itself. However, I *think* we want to coerce to the same backend in the highlevel `ak.XXX` functions ASAP, i.e. before broadcasting. 

Whilst I was here, I cleaned up some of the routines that use `backend_of`.

### DIFF
--- a/src/awkward/operations/ak_linear_fit.py
+++ b/src/awkward/operations/ak_linear_fit.py
@@ -2,6 +2,7 @@
 __all__ = ("linear_fit",)
 import awkward as ak
 from awkward._backends.dispatch import backend_of
+from awkward._backends.numpy import NumpyBackend
 from awkward._behavior import behavior_of
 from awkward._dispatch import high_level_function
 from awkward._layout import wrap_layout
@@ -9,6 +10,7 @@ from awkward._nplikes import ufuncs
 from awkward._nplikes.numpylike import NumpyMetadata
 from awkward._regularize import regularize_axis
 
+cpu = NumpyBackend.instance()
 np = NumpyMetadata.instance()
 
 
@@ -76,7 +78,7 @@ def linear_fit(x, y, weight=None, axis=None, *, keepdims=False, mask_identity=Fa
 def _impl(x, y, weight, axis, keepdims, mask_identity):
     axis = regularize_axis(axis)
     behavior = behavior_of(x, y, weight)
-    backend = backend_of(x, y, weight, coerce_to_common=True)
+    backend = backend_of(x, y, weight, coerce_to_common=True, default=cpu)
     x = ak.highlevel.Array(
         ak.operations.to_layout(x, allow_record=False, allow_other=False).to_backend(
             backend
--- a/src/awkward/operations/ak_linear_fit.py
+++ b/src/awkward/operations/ak_linear_fit.py
@@ -76,22 +76,28 @@ def linear_fit(x, y, weight=None, axis=None, *, keepdims=False, mask_identity=Fa
 def _impl(x, y, weight, axis, keepdims, mask_identity):
     axis = regularize_axis(axis)
     behavior = behavior_of(x, y, weight)
+    backend = backend_of(x, y, weight, coerce_to_common=True)
     x = ak.highlevel.Array(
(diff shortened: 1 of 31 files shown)


src/awkward/operations/ak_where.py
```diff
@@ -87,57 +87,57 @@ def _impl1(condition, mergebool, highlevel, behavior):
...
+    x_layout = ak.operations.to_layout(x, allow_record=False, allow_other=True)
+    if isinstance(x_layout, ak.contents.Content):
+        x_layout = x_layout.to_backend(backend)
 
-    good_arrays = [akcondition]
-    if isinstance(left, ak.contents.Content):
-        good_arrays.append(left)
-    if isinstance(right, ak.contents.Content):
-        good_arrays.append(right)
-    backend = backend_of(*good_arrays, default=cpu)
```
agoose77, line 102: I think this was here to avoid passing non backend-aware objects to `backend_of`. This is not necessary any more (and hasn't been for a while), so it's OK to remove this.

tests/test_2678_same_backend.py
```diff
@@ -0,0 +1,19 @@
...
+import pytest  # noqa: F401
+
+import awkward as ak
+
+
+def test_where():
+    result = ak.where(
+        [True, False, False], ak.to_backend([1, 2, 3], "typetracer"), [4, 5, 6]
+    )
+    assert ak.backend(result) == "typetracer"
```
agoose77, line 13: This just clarifies that the changes in this PR didn't break mixed-backends for `ak.where`.

tests/test_2678_same_backend.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def test_almost_equal():
+    assert not ak.almost_equal(
+        [True, False, False], ak.to_backend([True, False, False], "typetracer")
+    )
```
agoose77, line 19: New mixed-backend support for `almost_equal`.

src/awkward/operations/ak_linear_fit.py
```diff
@@ -76,22 +76,28 @@ def linear_fit(x, y, weight=None, axis=None, *, keepdims=False, mask_identity=Fa
 def _impl(x, y, weight, axis, keepdims, mask_identity):
     axis = regularize_axis(axis)
     behavior = behavior_of(x, y, weight)
+    backend = backend_of(x, y, weight, coerce_to_common=True)
     x = ak.highlevel.Array(
-        ak.operations.to_layout(x, allow_record=False, allow_other=False),
+        ak.operations.to_layout(x, allow_record=False, allow_other=False).to_backend(
+            backend
+        ),
```
agoose77, line 85: Previously we could have had mixed backends until the ufuncs later in this routine. It is better to convert up-front.

src/awkward/operations/ak_where.py
jpivarski: I'm 90% sure you're right about that being the intention of this code.

tests/test_2678_same_backend.py
jpivarski: :heavy_check_mark:

tests/test_2678_same_backend.py
jpivarski: :heavy_check_mark:

src/awkward/operations/ak_linear_fit.py
jpivarski: :heavy_check_mark:

jpivarski (approved):
This all looks good. Thanks!

PR #2679, opened 2023-08-29
fix: support typetracer in `ak.str.` operations

### LINKED ISSUE #2675: Fix typetracer for string operations
### Version of Awkward Array

main

I don't think we can fix this via adding support for typetracers in `to_arrow`, as we don't want to start mimicking the arrow objects.

agoose77:
Fixes #2675 by introducing a typetracer branch in a new `_apply_through_arrow` helper. This helper only handles the single-argument case, because the multi-array cases are more complex (they need finer control over how things get converted to arrow).

Some functions in PyArrow [have bugs](https://github.com/apache/arrow/issues/37437) that require us to workaround by manually creating a layout from a form.

### DIFF
--- a/src/awkward/operations/str/__init__.py
+++ b/src/awkward/operations/str/__init__.py
@@ -3,89 +3,121 @@
 # This set of string-manipulation routines is strongly inspired by Arrow:
 
 
-# string predicates
-# https://arrow.apache.org/docs/python/api/compute.html#string-predicates
-from awkward.operations.str.akstr_is_alnum import *
-from awkward.operations.str.akstr_is_alpha import *
-from awkward.operations.str.akstr_is_decimal import *
-from awkward.operations.str.akstr_is_digit import *
-from awkward.operations.str.akstr_is_lower import *
-from awkward.operations.str.akstr_is_numeric import *
-from awkward.operations.str.akstr_is_printable import *
-from awkward.operations.str.akstr_is_space import *
-from awkward.operations.str.akstr_is_upper import *
-from awkward.operations.str.akstr_is_title import *
-from awkward.operations.str.akstr_is_ascii import *
-
 # string transforms
 # https://arrow.apache.org/docs/python/api/compute.html#string-transforms
 from awkward.operations.str.akstr_capitalize import *
-from awkward.operations.str.akstr_length import *
-from awkward.operations.str.akstr_lower import *
-from awkward.operations.str.akstr_swapcase import *
-from awkward.operations.str.akstr_title import *
-from awkward.operations.str.akstr_upper import *
-from awkward.operations.str.akstr_repeat import *
-from awkward.operations.str.akstr_replace_slice import *
-from awkward.operations.str.akstr_reverse import *
-from awkward.operations.str.akstr_replace_substring import *
-from awkward.operations.str.akstr_replace_substring_regex import *
 
 # string padding
 # https://arrow.apache.org/docs/python/api/compute.html#string-padding
 from awkward.operations.str.akstr_center import *
-from awkward.operations.str.akstr_lpad import *
-from awkward.operations.str.akstr_rpad import *
-
-# string trimming
-# https://arrow.apache.org/docs/python/api/compute.html#string-trimming
-from awkward.operations.str.akstr_ltrim import *
-from awkward.operations.str.akstr_ltrim_whitespace import *
-from awkward.operations.str.akstr_rtrim import *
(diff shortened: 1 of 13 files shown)


agoose77:
@jpivarski I've noticed that PyArrow doesn't seem to return proper-sized offsets for the following
```python
import pyarrow as pa
import pyarrow.compute as pc

print(
    pc.utf8_split_whitespace(
        pa.array([], type=pa.large_string()),
    ).buffers()
)

print(pa.array([], type=pa.large_string()).buffers())
```

Could you give this code-block a once-over, and confirm that we _should_ expect both sets of offsets buffers to have `size>=8`, i.e. the size of a single offset (64 bit)? I assume this is a PyArrow bug, but I want to be sure before reporting.

jpivarski:
It looks like `pc.utf8_split_whitespace` on arrays of `large_string` is returning arrays of lists of `string`.

```python
>>> [None if x is None else len(x)
...  for x in pa.array([], type=pa.large_string()).buffers()]
[None, 8, 0]
>>> # string-mask (None), string-offsets (8 bytes), string-data (0 bytes)

>>> [None if x is None else len(x)
...  for x in pc.utf8_split_whitespace(pa.array([], type=pa.large_string())).buffers()]
[None, 4, None, 4, 4]
>>> # list-mask (None), list-offsets (4), string-mask (None), string-offsets (4), string-data (4???)
```

versus

```python
>>> [None if x is None else len(x)
...  for x in pa.array([], type=pa.string()).buffers()]
[None, 4, 0]
>>> # string-mask (None), string-offsets (4 bytes), string-data (0 bytes)

>>> [None if x is None else len(x)
...  for x in pc.utf8_split_whitespace(pa.array([], type=pa.string())).buffers()]
[None, 4, None, 4, 4]
>>> # list-mask (None), list-offsets (4), string-mask (None), string-offsets (4), string-data (4???)
```

That's not obviously wrong: we have plenty of functions that take `Index32` inputs and return `Index64` outputs. Arrow's bias is toward 32-bit; 64-bit was a late addition before version 1.0. Maybe Arrow is even deciding on the index size of the output based on the lengths of the string values, which would be a problem for our Form-stability, but we can cast the outputs if we need to. We could just have a policy of always turning Arrow 32-bit output indexes into 64-bit—that way, it would never be value-dependent.

--------------

Nope, it really is an Arrow bug: the declared type is still `large_string`, despite the fact that the buffer is 4 bytes long:

```python
>>> pc.utf8_split_whitespace(pa.array([], type=pa.large_string())).type
ListType(list<item: large_string>)
>>> pc.utf8_split_whitespace(pa.array([], type=pa.string())).type
ListType(list<item: string>)
```

It's definitely wrong for Arrow to declare the type `large_string` and have only 4 bytes for the offset. (It needs a single `[0]` for the offsets of an empty string, just like Awkward.) The `large_` means 8-byte index.

I also find it a bit odd that the string data for the empty lists of strings returned by `pc.utf8_split_whitespace` is non-empty (the "`???`" above). I don't know what rules for too-large buffers Arrow allows; it would be allowed in Awkward, but perhaps unexpected in this case because I don't see what good it might do.

agoose77:
> It's definitely wrong for Arrow to declare the type large_string and have only 4 bytes for the offset. (It needs a single [0] for the offsets of an empty string, just like Awkward.) The large_ means 8-byte index.

Good, that matches my understanding here. I'll file a bug with PyArrow.

agoose77:
I've rebased the commits such that they could be applied in sequence. That might make for easier per-commit reviewing than the whole changeset in one go.

jpivarski (approved):
`string_to32=True, bytestring_to32=True` is a fine way to solve the problem of Arrow functions not being properly implemented for 64-bit.

Solving the typetracer problem in one place, through `_apply_through_arrow`, is a good idea, but I've noticed that only some of the functions use it; others use `to_arrow` and `from_arrow` manually. Is it the case that only some of them need it? (There are several different classes of function signatures, I know.) Even if only some of them need it, could they be made to all go through the same helper function, even if that helper function has a trivial action on the ones that don't need it, for regularity?

(If there's a good reason not to, then that's fine.)

jpivarski (approved):
Great! This unblocks dask-awkward, so that it can wrap these functions.

All of the changes are things that look plausible to me; having to coerce Forms in a few cases, dealing with Arrow sometimes adding option, sometimes not (different versions), 32-bit versus 64-bit, etc.

We now have a `pyarrow==7` test to make sure that this code is compliant with our range of Arrow versions.

So it all looks good! Merge when you're ready!

PR #2680, opened 2023-08-30
feat!: deprecate form key for forms that have no buffers

agoose77:
1. Raises `DeprecationWarning` if `form_key` is passed to constructor of non-buffer forms
   - `UnmaskedForm`
   - `EmptyForm`
   - `RecordForm`
2. No longer passes `form_key` to these types from form factories, such as pickle, `form.from_XXX` methods, `Content._form_with_key`
3. No longer outputs `form_key` in `form.to_dict(...)` for these types.


agoose77:
After discussing this with @jpivarski, there's no _strong_ remaining impetus to continue this work, so I'm closing it.

PR #2684, opened 2023-09-01
chore: bump numba and pyarrow test pins

agoose77:
These dependencies now support Python 3.11


jpivarski (approved):
Great, thanks! I'll merge this right away.

PR #2685, opened 2023-09-01
chore: bump C++ standard version

### LINKED ISSUE #2608: Increase the minimum C++ version to C++17
The reason is because we use `std::optional` and `std::variant` in header-only code that (currently) only gets compiled in ROOT/RDataFrame (and these are in tests), so we're _sometimes_ requiring C++17, so we should just be consistent about it, to avoid confusing people. Also, this is a good first step toward migrating from pybind11 to nanobind (which requires C++17).

https://github.com/scikit-hep/awkward/blob/ce63bf2b4afebd93e5b1b67d9985d66b51d37bc5/awkward-cpp/CMakeLists.txt#L53

Also need to set the minimum cibuildwheel MacOS version to 10.12. Maybe 10.14 (need to check).

agoose77:
- Bumps minimum C++ standard
- Bumps awkward-cpp version
- Fixes #2608

### DIFF
--- a/awkward-cpp/CMakeLists.txt
+++ b/awkward-cpp/CMakeLists.txt
@@ -50,7 +50,7 @@ file(GLOB_RECURSE LIBAWKWARD_SOURCES CONFIGURE_DEPENDS "src/libawkward/*.cpp")
 add_library(awkward-parent INTERFACE)
 target_compile_definitions(awkward-parent INTERFACE VERSION_INFO="${SKBUILD_PROJECT_VERSION}")
 target_include_directories(awkward-parent INTERFACE include)
-target_compile_features(awkward-parent INTERFACE cxx_std_11)
+target_compile_features(awkward-parent INTERFACE cxx_std_17)
 
 # C++ dependencies (header-only): RapidJSON
 target_include_directories(awkward-parent INTERFACE rapidjson/include)
--- a/awkward-cpp/CMakeLists.txt
+++ b/awkward-cpp/CMakeLists.txt
@@ -24,11 +24,7 @@ Please check https://github.com/scikit-hep/awkward#installation-for-developers t
 ")
 endif()
 
-# Defaults for properties in this directory (and below)
-set(CMAKE_CXX_EXTENSIONS OFF)
-set(CMAKE_CXX_VISIBILITY_PRESET hidden)
-set(CMAKE_VISIBILITY_INLINES_HIDDEN ON)
-
+# Setup the RPATH for built libraries
 if(CMAKE_SYSTEM_NAME STREQUAL Emscripten)
   message(STATUS "Using Emscripten")
   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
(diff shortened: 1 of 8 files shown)


agoose77:
@henryiii do you think the changes in this PR to use `target_compile_features` is appropriate?

henryiii:
Yes. It seems like it might get set in a few too many places - if you link to a PUBLIC or INTERFACE target with compile features set, then you get them too. (Not true with the properties).

agoose77:
> Yes. It seems like it might get set in a few too many places

This was my worry / hunch. I'll untangle the ones we don't need.

agoose77:
@henryiii how's this?

jpivarski (approved):
I'm in favor of this, though we should let @henryiii give it a quick look before merging it.

awkward-cpp/CMakeLists.txt
```diff
@@ -50,7 +49,12 @@ file(GLOB_RECURSE LIBAWKWARD_SOURCES CONFIGURE_DEPENDS "src/libawkward/*.cpp")
 add_library(awkward-parent INTERFACE)
 target_compile_definitions(awkward-parent INTERFACE VERSION_INFO="${SKBUILD_PROJECT_VERSION}")
 target_include_directories(awkward-parent INTERFACE include)
-target_compile_features(awkward-parent INTERFACE cxx_std_11)
+target_compile_features(awkward-parent INTERFACE cxx_std_17)
+set_target_properties(
+  awkward-parent
+  PROPERTIES CXX_VISIBILITY_PRESET hidden
+             VISIBILITY_INLINES_HIDDEN ON
+             CXX_EXTENSIONS NO)
```
henryiii: This is an interface target? Why is CMake letting you set these non-interface properties on an interface target? There's no built component to apply them to, and they are not inheritable last I checked.

henryiii: (Interface properties start with `INTERFACE_`)

agoose77: Oops, these should also be removed. Let me do another pass.

agoose77:
@henryiii your comments above RE the INTERFACE properties made me realise I was missing some understanding about this. The PR should now reflect my working knowledge:
1. `INTERFACE` libraries may only have `INTERFACE_` properties
2. `set_target_properties` are not inheritable via linking

I've changed our various `CMakeLists.txt` files to reflect this, and cherry picked 60f822e60f6e6fb2fe17d2f4f74d026b5e3c1f37 from #2687. Let me know if this looks wrong.

agoose77:
@henryiii sorry to ping you on this again, would you be able to take another look? This is a blocker for our next minor release.

jpivarski:
Once this goes in, we'll also need a new `awkward-cpp` version.

agoose77:
This PR actually bumps that version. Have I erred there; did we want to separate the process of bumping awkward-cpp into two commits?

jpivarski:
I checked the [Needs C++ Release](https://github.com/scikit-hep/awkward/actions/runs/6075343421/job/16481336480) test, but then realized that that was redundant because this PR touches the C++ as well. So I was just noting, as a reminder, that we'll need to run a Deploy C++, and this PR blocks that ~1 hour long process from starting.

henryiii (approved):

jpivarski:
Thanks, @henryiii!

Issue #2686, opened 2023-09-01
awkward-cpp header-only requires cmake 3.19 due to INTERFACE_LIBRARY property whitelist


wdconinc:
### Version of Awkward Array

awkward-cpp>=14

### Description and code to reproduce

On a standard [Compute Canada](https://docs.alliancecan.ca/wiki/Narval/en) system (narval), with cmake 3.16.5, `pip install awkward-cpp==14` (and more recent) fails to install with:
```console
[wdconinc@narval2 ~]$ pip install awkward-cpp==14
Defaulting to user installation because normal site-packages is not writeable
Looking in links: /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/gentoo2020/avx2, /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/gentoo/avx2, /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/gentoo2020/generic, /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/gentoo/generic, /cvmfs/soft.computecanada.ca/custom/python/wheelhouse/generic
Collecting awkward-cpp==14
  Downloading awkward-cpp-14.tar.gz (1.4 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 1.4/1.4 MB 425.5 kB/s eta 0:00:00
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
  Preparing metadata (pyproject.toml) ... done
Requirement already satisfied: numpy>=1.17.0 in ./.local/lib/python3.10/site-packages (from awkward-cpp==14) (1.25.2+computecanada)
Building wheels for collected packages: awkward-cpp
  Building wheel for awkward-cpp (pyproject.toml) ... error
  error: subprocess-exited-with-error
  
  × Building wheel for awkward-cpp (pyproject.toml) did not run successfully.
  │ exit code: 1
  ╰─> [66 lines of output]
      *** scikit-build-core 0.5.0 using CMake 3.16.5 (wheel)
      *** Configuring CMake...
      loading initial cache file build/cpython-310/CMakeInit.txt
      -- The CXX compiler identification is GNU 9.3.0
      -- Check for working CXX compiler: /cvmfs/soft.computecanada.ca/easybuild/software/2020/Core/gcccore/9.3.0/bin/c++
[...]

PR #2687, opened 2023-09-01
fix: support CMake < 3.19

henryiii:
Looking at #2686, seeing if it can be easily replicated before attempting fix.


agoose77:
Closed in favour of #2685

PR #2688, opened 2023-09-02
feat!: remove deprecations for 1.4.0 release

agoose77:
Per [the wiki](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20):

- `typestr` argument will be removed from `Form` objects.
- `Form.type_from_behavior(...)` will be removed in favour of `Form.type`.
- `ak.typetracer.empty_if_typetracer` will be removed in favour of `ak.typetracer.length_zero_if_typetracer`.
- `UnionArray.simplified`'s `merge` argument will be removed.
- The built-in string and categorical behaviors will be removed.


agoose77:
Hmm, looks like we probably shouldn't remove the `ak.behaviors.string` module (or categorical) just yet; any pickles whose array class is defined in these modules will thus fail to load. I propose to wait, and eventually we'll be able to remove these outright. Let's bump the removal to 6-12 months?

jpivarski (approved):
Thanks for making each deprecated feature removal a separate commit; it was much easier to review each commit separately!

> Hmm, looks like we probably shouldn't remove the `ak.behaviors.string` module (or categorical) just yet; any pickles whose array class is defined in these modules will thus fail to load. I propose to wait, and eventually we'll be able to remove these outright. Let's bump the removal to 6-12 months?

Good point—to maintain backward pickle compatibility, we can never fully remove them. But they can become hollow shells of their former selves, like the `ak._ext.Form` and subclasses in

https://github.com/scikit-hep/awkward/blob/519bba6ed2eec4e227994d2fd1a62b2a51f15e20/src/awkward/_ext.py#L3-L17

If all we're trying to do is have `pickle.load` not fail on old data, the `ak.behaviors.string.StringBehavior` etc. classes must continue to exist at that address, but they don't have to be special classes. We can keep the string.py and categorical.py modules in behaviors/, but `StringBehavior` etc. can be alternate names for `ak.Array` (i.e. no special class for them). After unpickling, the arrays would no longer carry any residue of pickle having declared that they should have these classes.

If anyone asks why we have these pass-throughs, the answer can be the same as for _ext.py: to preserve old pickles. (A comment can explain that in-situ.) That would minimize cognitive overhead.

agoose77:
Hmm, the thought had occurred to me but I dismissed it as non-viable. I'm not sure why, though; it's a good idea!

Now this can merge :)

PR #2696, opened 2023-09-06
fix: implement dlpack-device for ArrayBuilder

agoose77:
Also improve performance/fix dlpack for the same.


agoose77:
Oops, this is redundant with an ensuing PR.

PR #2697, opened 2023-09-06
fix: drop zero-cost views of `ak.Array`

### LINKED ISSUE #2695: zip + combinations clobbers data on x86?
### Version of Awkward Array

2.4.1

Starting from investigation in: https://github.com/dask-contrib/dask-histogram/issues/100

```python3
import hist
import numpy as np
import awkward as ak

if __name__ == "__main__":

    ar = ak.Array(200*np.random.random((100_000, 3)).astype(np.float32) + 32)

    electrons = ak.zip({"pt": ar})

    combos = ak.combinations(electrons, 2, fields=["first", "second"])

    for _ in range(10):
        x = hist.Hist.new.Variable([0, 10, 20, 30, 40, 80, 120, 200], name="x").Double()
        print(x.fill(x=ak.flatten(combos.second.pt)).values(flow=True))
```

results in:
```
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
[2.00000e+00 6.00000e+00 0.00000e+00 0.00000e+00 1.20520e+04 5.98710e+04
 6.01770e+04 1.20042e+05 4.78500e+04]
```
Note there are no underflows or values less than 32 in the dataset.

If I print out `ak.flatten(combos.second.pt)` before the histogram filling step the error magically goes away.
Still might be a histogramming issue.

agoose77:
In the words of Taylor Swift, I'm the problem — it's me.

In #2649, I exposed a DLPack view of Awkward Arrays, alongside the `__array_interface__` and `__cuda_array_interface__` properties. However, these are supposed to be zero copy, and non-owned views. As such, we can't return local temporary copies that leave scope. 

So, whoops, and here's a fix: we don't support DLPack or CUDA views of Awkward Arrays; they need to be explicitly converted to user-referenced backend arrays.

### DIFF
--- a/src/awkward/_connect/dlpack.py
+++ b/src/awkward/_connect/dlpack.py
@@ -1,11 +1,8 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 from __future__ import annotations
 
-__all__ = ("DLPackDevice", "get_layout_device", "to_dlpack")
+__all__ = ("DLPackDevice",)
 from enum import IntEnum
-
-from awkward._typing import Any
-from awkward.contents import Content
 
 
 class DLPackDevice(IntEnum):
@@ -19,26 +16,3 @@ class DLPackDevice(IntEnum):
     ROCM = 10  # GPU
     ROCM_PINNED = 11  # GPU & CPU
     CUDA_MANAGED = 13  # GPU & CPU
-
-
-def get_layout_device(layout: Content) -> tuple[int, int]:
-    while True:
-        if layout.is_numpy:
-            break
-        elif layout.is_regular:
-            layout = layout.content
-        else:
-            raise TypeError(
-                "Cannot determine the DLPack device for this array layout."
-                "DLPack is only supported for regular arrays."
-            )
-
-    return layout.data.__dlpack_device__()
-
-
-def to_dlpack(layout: Content, stream: Any = None) -> Any:
-    array = layout.to_backend_array(allow_missing=False)
-    if stream is None:
-        return array.__dlpack__()
-    else:
-        return array.__dlpack__(stream)
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -434,3 +434,11 @@ def array_ufunc(ufunc, method: str, inputs, kwargs: dict[str, Any]):
 
 def action_for_matmul(inputs):
     raise NotImplementedError
+
+
+def convert_to_array(layout, dtype=None):
+    out = ak.operations.to_numpy(layout, allow_missing=False)
+    if dtype is None:
+        return out
+    else:
+        return numpy.array(out, dtype=dtype)
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -23,7 +23,6 @@ import awkward._connect.hist
 from awkward._backends.dispatch import register_backend_lookup_factory
 from awkward._backends.numpy import NumpyBackend
 from awkward._behavior import behavior_of, get_array_class, get_record_class
-from awkward._connect.dlpack import get_layout_device, to_dlpack
 from awkward._layout import wrap_layout
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpylike import NumpyMetadata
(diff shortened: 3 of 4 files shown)


agoose77:
@lgray this will be in the next release, ASAP.

ikrommyd:
Fixes my code as seen in the output of the last cell here: https://github.com/iasonkrom/egamma-tnp/blob/debug_dask_histogram/debug_dask_histogram.ipynb

jpivarski (approved):
I had thought that `__cuda_array_interface__` and `__dlpack_device__` would only affect rectilinear arrays on GPUs, and these don't have applications yet.

That might still be true. It may be that the `__array_interface__` is the one that's causing the trouble. Yes, hist uses it:

```python
>>> class Something:
...     __array_interface__ = {"data": np.arange(10), "shape": (10,), "typestr": "<i8"}
... 
>>> hist.Hist.new.Reg(10, 0, 10).Double().fill(Something())
Hist(Regular(10, 0, 10, label='Axis 0'), storage=Double()) # Sum: 10.0
```

CuPy uses `__cuda_array_interface__` but nobody else does:

```python
>>> data = cp.arange(10)
>>> data.__cuda_array_interface__
{'shape': (10,), 'typestr': '<i8', 'descr': [('', '<i8')], 'stream': 1, 'version': 3, 'strides': None, 'data': (140416645071360, False)}
>>> class Something:
...     __cuda_array_interface__ = {"data": (140416645071360, False), "shape": (10,), "typestr": "<i8", "stream": 1, "version": 3}
... 
>>> cp.asarray(Something())
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

(The above worked, but both of the below didn't.)

```python
>>> np.asarray(Something())
array(<__main__.Something object at 0x7fb574e86cd0>, dtype=object)
>>> hist.Hist.new.Reg(10, 0, 10).Double().fill(Something())
  File "/home/jpivarski/mambaforge/lib/python3.9/site-packages/hist/basehist.py", line 232, in fill
    return super().fill(*args, *data, weight=weight, sample=sample, threads=threads)
  File "/home/jpivarski/mambaforge/lib/python3.9/site-packages/boost_histogram/_internal/hist.py", line 529, in fill
    self._hist.fill(*args_ars, weight=weight_ars, sample=sample_ars)  # type: ignore[arg-type]
TypeError: float() argument must be a string or a number, not 'Something'
```

I haven't managed to get the DLPack protocol to be identified by anything:

```python
>>> from awkward._connect.dlpack import DLPackDevice
>>> class Something:
...     def __dlpack_device__(self):
...         return DLPackDevice.CPU
...     def __dlpack__(self, stream=None):
...         return np.arange(10)
```

On Slack, @agoose77 said he'd split out the `__cuda_array_interface__` and `__dlpack_device__` removals into a different PR because they have a different motivation than the `__array_interface__` removal. The `__array_interface__` removal is for fixing #2695, and this is the right fix for that.

ikrommyd:
Would it be possible to roll out a patch 2.4.2 release?

PR #2700, opened 2023-09-08
feat: expose `ak.from_non_simplified_buffers`

agoose77:
Over in bluesky/tiled#549, there's a need to perform a minimal read of an Awkward Array that's already in `to_buffers` form. This scenario, slicing an array on the server itself, differs from the dask-awkward partial-read case, in that the array form is allowed (and expected) to change with the slice operation.

Our simplification logic has the ability to merge buffers e.g. if a union is no longer canonical. This means that this simplification step needs to happen _after_ the data have been read from disk. Upon discussing this with @jpivarski, we concluded that it would be useful to manipulate the form on the client, and perform a `from_buffers` that performs this simplification during reading.

This PR adds `ak.from_non_simplified_buffers`. @jpivarski suggested making this "hidden" as an L2 function. Whilst it still remains L2 by designation, I think the name is sufficient to export it in the `ak.operations` namespace and to publish the documentation. Our L2 layouts already introduce the `simplified` convention, so it should be clear what this function does.

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -2515,7 +2515,7 @@ class ArrayBuilder(Sized):
         form = ak.forms.from_json(formstr)
 
         with ak._errors.OperationErrorContext("ak.ArrayBuilder.snapshot", [], {}):
-            return ak.operations.ak_from_buffers._impl(
+            return ak.operations.from_non_simplified_buffers(
                 form,
                 length,
                 container,
@@ -2524,7 +2524,6 @@ class ArrayBuilder(Sized):
                 byteorder=ak._util.native_byteorder,
                 highlevel=True,
                 behavior=self._behavior,
-                simplify=True,
             )
 
     def null(self):
--- a/src/awkward/operations/__init__.py
+++ b/src/awkward/operations/__init__.py
@@ -37,6 +37,7 @@ from awkward.operations.ak_from_dlpack import *
 from awkward.operations.ak_from_iter import *
 from awkward.operations.ak_from_jax import *
 from awkward.operations.ak_from_json import *
+from awkward.operations.ak_from_non_simplified_buffers import *
 from awkward.operations.ak_from_numpy import *
 from awkward.operations.ak_from_parquet import *
 from awkward.operations.ak_from_rdataframe import *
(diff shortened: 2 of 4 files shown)


agoose77:
@jpivarski let me know how this sits with you in its current form? :)

jpivarski (commented):
With this name, it wouldn't come next to `ak.from_buffers` in a sorted list. (`ak.from_buffers_non_simplified`?) It's nearly identical to `from_buffers`, apart from one boolean argument, which makes for a lot of duplicated docstrings.

It would be better as a non-default argument in `ak.from_buffers`. I also think the word "simplify" isn't enough—even though we've defined it, it's a generic enough word that someone who reads it might not realize there's a technical definition to look up.

Maybe the argument could be named `allow_invalid_form` or `fix_invalid_form`? (Or replace "invalid" with "noncanonical," as the word you've used in the documentation?) It would be the last argument, apart from `highlevel` and `behavior`, to indicate that it's for experts.

What I had in mind for "hiding" it was more like making it an attribute of `ak.from_buffers`, like `ak.from_buffers.noncanonical`, or putting it in a module that's not promoted to the top-level `ak.*` namespace. If you want to keep it a separate function, rather than an argument in `ak.from_buffers`, it would be better in one of these hidden places, which would allow Tiled to move forward with it.

agoose77:
> With this name, it wouldn't come next to ak.from_buffers in a sorted list. (ak.from_buffers_non_simplified?) It's nearly identical to from_buffers, apart from one boolean argument, which makes for a lot of duplicated docstrings.

Yes, that occurred to me too. Then again, I don't mind that from a English standpoint.

> It would be better as a non-default argument in ak.from_buffers. I also think the word "simplify" isn't enough—even though we've defined it, it's a generic enough word that someone who reads it might not realize there's a technical definition to look up.

Ah, I thought you'd stood against that. In hindsight, I don't recall you ever saying that so I must have just got my wires crossed. I also prefer this! 

> What I had in mind for "hiding" it was more like making it an attribute of ak.from_buffers, like ak.from_buffers.noncanonical, or putting it in a module that's not promoted to the top-level ak.* namespace. 

Right, it also occurred to me to either:
- Create new `ak_from_buff...` and _not_ export via `from XXX import *`
   - Undecided on this
- Create a new attribute of `from_buffers`
   - This is cool, but I don't love it from a surprise perspective

Yet, given that we both support doing this via an argument, let's just do that and KISS.

jpivarski:
> * Create new `ak_from_buff...` and _not_ export via `from XXX import *`

A future maintainer is likely to think, "Oops, that was a mistake. Let me fix that..."

Let's both go with making it an argument. (You're right that I was against it in the call; I just think it's _better_ than a new high-level function.) The name of that argument should be relatively long and self-explanatory, like `allow_noncanonical_form`.

danielballan:
Is there a superseding PR? I could not quickly find one.

PR #2710, opened 2023-09-14
fix: unpickling of `unknown_length`

agoose77:
This PR makes `unknown_length` use the singleton machinery so that it's trivially pickleable.


jpivarski (approved):
Okay, sounds good! It's also straightforward enough that I'll merge it, too.

PR #2713, opened 2023-09-17
feat: expose simplification from `ak.from_buffers`

agoose77:
This PR supersedes #2700 in simplify adding a new `allow_noncanonical_form` argument to `from_buffers`. This is a much more concise API change.


agoose77:
@danielballan this PR should do it! Apologies, this slipped off my task list.

agoose77:
I've employed a rule that if `from_buffers` has a known length, then the buffer itself must have a known length. This means that placeholders inside unions would be required to have known lengths.

The purpose of imposing this constraint that `from_buffers` can be naive, and as such render the caller of `from_buffers` responsible for populating the lengths appropriately. However, I this might be that this is too restrictive, and perhaps we should just let `from_buffers` fill in unknown lengths when it has the ability to. In any case, such a change could be made at a later date.

jpivarski (approved):
In our conversations about what would be a good API, this is where we settled. I like the name and think that an argument on the single function is the best way to do it.

jpivarski:
> I've employed a rule that if `from_buffers` has a known length, then the buffer itself must have a known length. This means that placeholders inside unions would be required to have known lengths.

:heavy_check_mark:

I can imagine wanting to be selective about which fields of a record one reads, but not about which possibilities of a union array. Possibilities in a union are supposed to be a short list; records can have thousands of fields.

jpivarski:
I think this is ready to merge. (The code change is rather small; it's mostly an extensive test.)

danielballan:
Thanks to you both!

PR #2693, opened 2023-09-05
fix: support placeholder arrays in `ArrayModuleNumpyLike.frombuffer`

agoose77:
This PR:
- preserves `PlaceholderArray` instances between `ak.to_buffers` and `ak.from_buffers`
- fixes pickling of `Singleton` objects like `NumpyLike` and `Backend`

Fixes #2674

PR #2714, opened 2023-09-17
fix: support placeholders in `from_buffers`

agoose77:
With #2693 we support writing placeholder arrays using `to_buffers`, which is used for pickling re-hydrated arrays. This PR supports reading such arrays.

By design, this routine throws an error if a placeholder has an unknown length but a known length can be determined by the recursion; the caller should ensure that placeholders have valid lengths. We can relax this in future, but for now it makes things easier to reason about.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -53,15 +53,11 @@ class ArrayModuleNumpyLike(NumpyLike):
             return self._module.ascontiguousarray(x)
 
     def frombuffer(
-        self, buffer, *, dtype: np.dtype | None = None, count: int = -1
+        self, buffer, *, dtype: np.dtype | None = None, count: ShapeItem = -1
     ) -> ArrayLike:
         if isinstance(buffer, PlaceholderArray):
-            if count == -1:
-                return self.asarray(buffer)
-            else:
-                return self.asarray(buffer[:count])
-        else:
-            return self._module.frombuffer(buffer, dtype=dtype, count=count)
+            raise TypeError("placeholder arrays are not supported in `frombuffer`")
+        return self._module.frombuffer(buffer, dtype=dtype, count=count)
 
     def from_dlpack(self, x: Any) -> ArrayLike:
         return self._module.from_dlpack(x)
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -46,7 +46,12 @@ def from_dict(input: Mapping) -> Form:
 
     if input["class"] == "NumpyArray":
         primitive = input["primitive"]
-        inner_shape = input.get("inner_shape", [])
+        inner_shape = tuple(
+            [
+                unknown_length if item is None else item
+                for item in input.get("inner_shape", [])
+            ]
+        )
         return ak.forms.NumpyForm(
             primitive, inner_shape, parameters=parameters, form_key=form_key
         )
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -47,10 +47,8 @@ def from_dict(input: Mapping) -> Form:
     if input["class"] == "NumpyArray":
         primitive = input["primitive"]
         inner_shape = tuple(
-            [
-                unknown_length if item is None else item
-                for item in input.get("inner_shape", [])
-            ]
+            unknown_length if item is None else item
+            for item in input.get("inner_shape", [])
         )
         return ak.forms.NumpyForm(
             primitive, inner_shape, parameters=parameters, form_key=form_key
--- a/src/awkward/operations/ak_from_buffers.py
+++ b/src/awkward/operations/ak_from_buffers.py
@@ -178,10 +178,8 @@ def _reconstitute(form, length, container, getkey, backend, byteorder, simplify)
             byteorder=byteorder,
         )
         if form.inner_shape != ():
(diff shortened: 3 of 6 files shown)


src/awkward/_nplikes/array_module.py
```diff
@@ -53,15 +53,11 @@ def ascontiguousarray(self, x: ArrayLike) -> ArrayLike:
...
     ) -> ArrayLike:
         if isinstance(buffer, PlaceholderArray):
-            if count == -1:
-                return self.asarray(buffer)
-            else:
-                return self.asarray(buffer[:count])
-        else:
-            return self._module.frombuffer(buffer, dtype=dtype, count=count)
+            raise TypeError("placeholder arrays are not supported in `frombuffer`")
+        return self._module.frombuffer(buffer, dtype=dtype, count=count)
```
agoose77, line 60: I don't think we should support placeholders in `nplike.frombuffer`; they don't actually have any data. We could make exceptions to this rule, but for `frombuffer` this doesn't make any sense.

src/awkward/forms/form.py
```diff
@@ -46,7 +46,12 @@ def from_dict(input: Mapping) -> Form:
 
     if input["class"] == "NumpyArray":
         primitive = input["primitive"]
-        inner_shape = input.get("inner_shape", [])
+        inner_shape = tuple(
+            [
+                unknown_length if item is None else item
+                for item in input.get("inner_shape", [])
+            ]
+        )
```
agoose77: We should support forms with unknown-shape N-D NumpyArrays

src/awkward/_nplikes/array_module.py
jpivarski: I agree: even though we'd be loading the placeholder at the point where we normally call `frombuffer`, having that pass through `frombuffer` is an abuse of the name.

src/awkward/operations/ak_from_buffers.py
```diff
@@ -123,11 +125,27 @@ def _impl(
...
+    # contain placeholders
+    if count is unknown_length:
+        if not isinstance(buffer, PlaceholderArray):
+            raise AssertionError("Encountered unknown length for concrete buffer")
+        return PlaceholderArray(nplike, (unknown_length,), dtype)
+    # Known-length information implies that we should have known-length buffers here
+    # Therefore, placeholders without shape information are not permitted
+    elif isinstance(buffer, PlaceholderArray) or nplike.is_own_array(buffer):
+        # Require 1D buffers
         array = nplike.reshape(buffer.view(dtype), shape=(-1,), copy=False)
```
jpivarski, line 142: It's good that this can all happen in one place (i.e. that there's a `_from_buffer` function).

src/awkward/operations/ak_from_buffers.py
```diff
@@ -151,9 +169,7 @@ def _reconstitute(form, length, container, getkey, backend, byteorder, simplify)
     elif isinstance(form, ak.forms.NumpyForm):
         dtype = ak.types.numpytype.primitive_to_dtype(form.primitive)
         raw_array = container[getkey(form, "data")]
-        real_length = length
-        for x in form.inner_shape:
-            real_length *= x
+        real_length = length * math.prod(form.inner_shape)
```
jpivarski, line 172: I went and checked on when [math.prod](https://docs.python.org/3/library/math.html#math.prod) was added: Python 3.8, and we support Python 3.8 onwards, so this is fine.

src/awkward/operations/ak_from_buffers.py
```diff
@@ -182,12 +196,15 @@ def _reconstitute(form, length, container, getkey, backend, byteorder, simplify)
 
     elif isinstance(form, ak.forms.BitMaskedForm):
         raw_array = container[getkey(form, "mask")]
-        excess_length = int(math.ceil(length / 8.0))
+        if length is unknown_length:
+            next_length = unknown_length
+        else:
+            next_length = int(math.ceil(length / 8.0))
```
jpivarski, line 202: "Excess length" is terminology that I even copied over [into Julia](https://github.com/jpivarski/AwkwardArray.jl/blob/4f97922ded63937d45f4b4c3dd474f379d95fd3f/src/AwkwardArray.jl#L3435). But `next_length` is fine.

The idea here is that bit masks can only come in units of bytes, so $\frac{7}{8}$ of the time, you'll have more in your mask than you should. That's the sense in which it's an excess.

src/awkward/operations/ak_from_buffers.py
```diff
@@ -300,8 +323,13 @@ def _reconstitute(form, length, container, getkey, backend, byteorder, simplify)
...
         )
-        reduced_stops = stops[starts != stops]
-        next_length = 0 if len(starts) == 0 else backend.index_nplike.max(reduced_stops)
+        if isinstance(stops, PlaceholderArray):
+            next_length = unknown_length
+        else:
+            reduced_stops = stops[starts != stops]
+            next_length = (
+                0 if len(starts) == 0 else backend.index_nplike.max(reduced_stops)
+            )
```
jpivarski, line 332: This idiom is repeated often, which makes it look like it wants to be a helper function, but if it was a helper function, it would have to take a function of "what to do if not `PlaceholderArray`" as an argument, and that kind of callback-hell would be worse than this repetition. So I'm in favor of what you've done here.

jpivarski (approved):
This looks good! I'd say it's ready to be merged if you have no final edits.

PR #2718, opened 2023-09-20
fix: support concrete buffers in `TypeTracer.frombuffer`

agoose77:
`ak.zeros_like` should support typetracer. This PR ensures that concrete buffers, such as those used in `zeros_like`, are supported. I didn't add support for using typetracer arrays here, because we don't yet need it.


jpivarski (approved):
I agree; `ak.zeros_like` should support typetracer!

This looks like a straightforward enough PR that I'll merge it right away.

PR #2719, opened 2023-09-22
feat: add `buffer_key` to `typetracer_with_report`

agoose77:
This PR makes it possible to format the `form_key`s assigned to typetracer buffers via `ak.typetracer.typetracer_with_report`. Although the default behaviour is unchanged, this permits us to ensure that buffer keys are unique within a typetracer layout.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -1431,7 +1431,7 @@ def _attach_report(
 
     elif isinstance(layout, ak.contents.NumpyArray):
         assert isinstance(form, ak.forms.NumpyForm)
-        layout.data.form_key = form.form_key
+        layout.data.form_key = getkey(form.form_key, "data")
         layout.data.report = report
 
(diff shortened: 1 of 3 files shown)


src/awkward/_nplikes/typetracer.py
```diff
@@ -610,7 +611,7 @@ def populate_shape_and_items(node, dim):
                     if has_seen_leaf:
                         if len(node) != shape[dim - 1]:
                             raise ValueError(
-                                f"sequence at dimension {dim} does not match shape {shape[dim-1]}"
+                                f"sequence at dimension {dim} does not match shape {shape[dim - 1]}"
```
agoose77, line 614: Think this was a linter change, but I'm fine with it.

jpivarski (approved):
So, at the time that (public function) `typetracer_with_report` is called, we assume that users already know what `buffer_key` function they'll want to use? That does sound reasonable.

Merge when you're ready!

PR #2720, opened 2023-09-25
fix: rework to use numba typed list

ianna:


ianna (commented):
@jpivarski - I'd like to keep the branch. Thanks!

jpivarski:
Absolutely: it's a draft that is still in progress. I was only asking about the branches associated with closed PRs, merged PRs, and branches that are not associated with any PRs.

PR #2724, opened 2023-09-26
feat: add `recursive` argument to `expected_from_buffers`

agoose77:
In the work to support buffer-shape touching in dask-awkward, I found it useful to be able to disable recursion in `expected_from_buffers` in order to identify the buffers associated directly with each form node. I think this has merit, so I've created this PR.

### DIFF
--- a/src/awkward/forms/bitmaskedform.py
+++ b/src/awkward/forms/bitmaskedform.py
@@ -262,7 +262,8 @@ class BitMaskedForm(Form):
             )
 
     def _expected_from_buffers(
-        self, getkey: Callable[[Form, str], str]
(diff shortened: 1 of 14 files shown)


jpivarski (approved):
I don't see any examples in this PR in which the new `recursive` argument gets used, but in principle it could be useful to have.

The default should be `recursive=True`, as you have it here, since its primary use-case is to find out what the `containers` needs to have, and that's a question for a whole tree, not a single node.

agoose77:
> I don't see any examples in this PR in which the new recursive argument gets used, but in principle it could be useful to have.

The test file uses both variations :)

PR #2725, opened 2023-09-27
fix: make the examples consistent with the changes in #2553

ManasviGoyal:
I noticed that the header-only C++ examples didn't reflect the changes made in the C++ LayoutBuilder API in #2553. Fixed the code to add the required changes.

### DIFF
--- a/header-only/examples/cython/CMakeLists.txt
+++ b/header-only/examples/cython/CMakeLists.txt
@@ -4,7 +4,7 @@ project(demo LANGUAGES CXX C)
 
 # Download headers
 include(FetchContent)
-set(AWKWARD_VERSION "v2.0.10")
+set(AWKWARD_VERSION "v2.4.3")
 FetchContent_Declare(
   awkward-headers
   URL https://github.com/scikit-hep/awkward/releases/download/${AWKWARD_VERSION}/header-only.zip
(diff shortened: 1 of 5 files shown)


agoose77:
Thanks Manasvi! We really should add a test that these compile, let me modify our CI.

header-only/examples/cython/demo_impl.cpp
```diff
@@ -72,8 +72,8 @@ ArrayBuffers create_demo_array() {
                     NumpyBuilder<int32_t>>>
     > builder(fields_map);
 
-    auto &one_builder = builder.field<Field::one>();
-    auto &two_builder = builder.field<Field::two>();
+    auto &one_builder = builder.content<Field::one>();
```
ianna, line 75: @ManasviGoyal - thanks for spotting this. Indeed we have changed the API of the builders.

ianna (approved):
Good catch! Thanks.

jpivarski (approved):
Thanks for catching this! When everyone is in agreement, let's merge it.

header-only/examples/cython/CMakeLists.txt
```diff
@@ -18,11 +18,9 @@ if(NOT awkward-headers_POPULATED)
                    EXCLUDE_FROM_ALL)
 endif()
 
+find_package(Python REQUIRED COMPONENTS Interpreter Development.Module NumPy)
```
agoose77, line 21: This is not a required change (besides linking NumPy), but it's more modern.

agoose77 (approved):

PR #2732, opened 2023-10-02
feat: infer unknown lengths from context in `from_buffers`

agoose77:
This PR follows up from #2718 by relaxing some rules:

- Placeholder arrays no longer need known lengths if `from_buffers` is able to deduce the length. 
   This requirement was originally made for safety reasons — if something went wrong during rehydration etc., we would see placeholders without known lengths in contexts where they should be known. However, by imposing this restriction, the caller would need to recreate much of `from_buffers` to compute lengths from offset buffers. Thus, we relax this rule, and permit `from_buffers` to update placeholders (via a new placeholder) with new _known_ lengths.
- Non-placeholder arrays can be encountered in contexts where the outer length is not known.
   Originally we wanted this to be an error so that we could clearly identify strange touching behaviour. However, in general it should be possible to refer to the same buffer multiple times within a form, so we should remove this rule.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -249,7 +249,6 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
 
     @property
     def ndim(self) -> int:
-        self.touch_shape()
         return len(self._shape)
 
     @property
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -18,7 +18,6 @@ from awkward._parameters import (
     parameters_intersect,
     type_parameters_equal,
 )
-from awkward._regularize import is_integer
 from awkward._slicing import NO_HEAD
 from awkward._typing import TYPE_CHECKING, Callable, Final, Self, SupportsIndex, final
 from awkward._util import UNSET
@@ -145,8 +144,15 @@ class RecordArray(Content):
         parameters=None,
         backend=None,
     ):
-        if not (length is None or length is unknown_length):
-            length = int(length)  # TODO: this should not happen!
+        if length is not None and length is not unknown_length:
+            try:
+                length = int(length)  # TODO: this should not happen!
+            except TypeError:
+                raise TypeError(
+                    "{} 'length' must be a non-negative integer or None, not {}".format(
+                        type(self).__name__, repr(length)
+                    )
+                ) from None
         if not isinstance(contents, Iterable):
             raise TypeError(
                 "{} 'contents' must be iterable, not {}".format(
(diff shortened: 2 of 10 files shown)


src/awkward/contents/recordarray.py
```diff
@@ -179,53 +186,50 @@ def __init__(
...
+                elif content.length is unknown_length:
+                    length = unknown_length
+                    break
+                # `length` is set, can't be unknown_length
+                else:
+                    length = min(length, content.length)
+
+            # Touch everything else
+            for content in it_contents:
+                content._touch_shape(False)
```
agoose77, line 216: Before hand, we'd early-exit, and *not* touch all the shapes. But at runtime, we'll want to know all the shapes so we can pick the smallest.

src/awkward/contents/recordarray.py
```diff
@@ -179,53 +186,50 @@ def __init__(
...
+        # Otherwise
         elif length is not unknown_length:
+            # Ensure lengths are not smaller than given length.
             for content in contents:
-                if content.length is not unknown_length and content.length < length:
+                if (
+                    backend.index_nplike.known_data
+                    and content.length is not unknown_length
+                    and content.length < length
+                ):
```
agoose77, line 226: The length is given, so only inspect shapes if we have known data

src/awkward/_nplikes/typetracer.py
```diff
@@ -249,7 +249,6 @@ def nplike(self) -> TypeTracer:
 
     @property
     def ndim(self) -> int:
-        self.touch_shape()
```
agoose77, line 252: The dimensionality is _not_ the shape, it's the length of the shape tuple. I think we need to treat this as a must-be-knowable.

jpivarski: Yes, absolutely. There are three levels of knowledge about array metadata that you can have:

  * no information about `ndim`, like NumPy at compile-time (i.e. when compiling NumPy)
  * knowledge of `ndim` but not the values in `shape`, like a Numba function when JIT-compiling it
  * full knowledge of the `shape`, like JAX when JIT-compiling for XLA

It is absolutely reasonable for our model of arrays at Dask-DAG collection time (which is like a kind of compilation) to be like Numba's model. Knowing `ndim`, you know how many times you can `__getitem__(int)` before the type changes from an array to a scalar, for instance. That seems pretty fundamental.

src/awkward/contents/recordarray.py
jpivarski: Right. (And early exits are dangerous for typetracer code.)

src/awkward/contents/recordarray.py
jpivarski: Makes sense.

jpivarski (approved):
This looks very sensible. And it's a good procedure for strict rules during early development to be relaxed later as the system gets more understood.

PR #2735, opened 2023-10-04
fix: update backport for changes in Numba, NumExpr, Arrow, and RDataFrame

### LINKED ISSUE #2734: Awkward version 1.x has a conflict (just a warning?) with Numba 0.53
### Version of Awkward Array

1.x

In scikit-hep/vector#388, [a test based on Awkward 1.x](https://github.com/scikit-hep/vector/actions/runs/6394289115/job/17355399815?pr=388) is failing:

```
FAILED tests/backends/test_awkward_numba.py::test - numba.core.errors.LoweringError: Failed in nopython mode pipeline (step: native lowering)
Failed in nopython mode pipeline (step: nopython frontend)
Code using Numba extension API maybe depending on 'old_style' error-capturing, which is deprecated and will be replaced by 'new_style' in a future release. See details at https://numba.readthedocs.io/en/latest/reference/deprecation.html#deprecation-of-old-style-numba-captured-errors
Exception origin:
  File "/opt/hostedtoolcache/Python/3.11.5/x64/lib/python3.11/site-packages/awkward/_connect/_numba/arrayview.py", line 877, in generic
    raise TypeError(


File "tests/backends/test_awkward_numba.py", line 23:
    def extract(x):
        return x[2][0]
        ^

During: lowering "$20binary_subscr.4 = static_getitem(value=$8binary_subscr.2, index=0, index_var=$const18.3, fn=<built-in function getitem>)" at /home/runner/work/vector/vector/tests/backends/test_awkward_numba.py (23)
```

Awkward 1.x should either fix this (Coffea 0.7 users are using it, and Coffea uses Numba) or put an upper cap on Numba versions, which would be difficult because Numba is a soft/runtime dependency.

The warning about old-style and new-style Numba error capturing might be like #2659, which fixed the issue for Awkward 2.x. Maybe all this needs is for the Numba-handling in Awkward 1.x to get the same kind of fix.

It will require a new Awkward 1.x release, which will take some thinking to remember how to do.

ianna:
address issue https://github.com/scikit-hep/awkward/issues/2734


jpivarski:
With or without this PR (that is, `main-v1` or `ianna/numba-error-backport-to-v1`), I get 66 test failures. They might not be related to Numba. I'm going to work through the test failures to see what can be done.

jpivarski (approved):
I'm self-reviewing this because I don't think there's anything a second pair of eyes can add.

* @ianna updated all Numba errors to `NumbaTypeError` and `NumbaValueError`, like #2659.
* I updated our NumExpr interface like #2617.
* I updated our pyarrow-Parquet interface to allow `.list.element` in addition to `.list.item`, similar to #2670.
* I dropped failing RDataFrame tests without investigating why they were failing. This is only in the `_v2` subdirectory, which no one should be using now: the 1.x backport is for using 1.x. If anyone is using 2.x interfaces, they should be using Awkward 2.x.

After merging this into `main-v1`, I'll also publish a release as 1.10.5.

jpivarski:
I know that the Windows 32-bit wheels build and are successfully tested from those builds.

https://github.com/scikit-hep/awkward/actions/runs/6410147455/job/17402806110?pr=2735

There is no reason to struggle with build-test.yml, to get its Windows to see the correct version of Python when wheels.yml does it correctly.

So I'll just remove the 32-bit test from build-test.yml and merge.

ianna:
I like the **−68,438**  :-)

jpivarski:
Oh yeah, I ran into a few problems with v2 tests failing, and this is v1, so I just dropped all of the tests. In principle, I could drop the `awkward._v2` source code as well, but I didn't do that because I wasn't sure whether it would break anything. (I'd have to look more closely.)

You had me wondering, for a moment, whether there was an integer wrap-around error, since -68,438 looks a bit like `-2**16`.

PR #2740, opened 2023-10-04
fix: don't use `np.asarray` on `Index` or `Content` objects

agoose77:
`nplike.asarray` ultimately delegates to the underlying NumPy-like library's `asarray`. For NumPy, this invokes `obj.__array__()`. If our object is e.g. an `Index` that wraps a placeholder, this conversion fails because NumPy et al. are not aware of our placeholder type, and eagerly tries to convert it to a buffer.

This would also fail during typetracing time, but it may well be that there are branches where this gets skipped.

We could add our own protocol to the `Content` and `Index` types that permits `Content` and `Index` objects to pass through `asarray`, e.g.
```python
class Numpy:
    ...
    def asarray(self, obj, *, ...):
        if hasattr(obj, "_awkward_asarray_"):
            return obj._awkward_asarray_(self, ...)
        else:
            return numpy.asarray(obj, ...)

class Index:
    ...
    def _awkward_asarray_(self, nplike, ...):
        return nplike.asarray(self.data, ...)
```

but on balance I think it might be easier just to pull out `.data` by hand, because most of the time we should know that we have a `Content` / `Index` object.

At the same time, we've long prohibited `np.asarray` on e.g. `RegularArray`, and I think we should just do this for all Content nodes to be more thorough. Thus, this PR removes `NumpyArray.__array__`, and the same for `EmptyArray` et al.

I didn't change `Index`, because I suspect that's used _everywhere_ in the wild (in legitimate NumPy-only cases), and the argument for `ak.Array` being the high-level `np.asarray` interface does not apply. Thus, library authors must take care not to use `asarray` even in NumPy-only contexts iff. there's a chance that the incoming array is a placeholder.

So, this PR changes two things:
1. Don't pass in `Content` or `Index` arguments to `nplike.asarray`, as they nest the true buffers within themselves.
2. Don't support `Content.__array__` for any content type, for the reason discussed above.

### DIFF
--- a/src/awkward/_connect/numexpr.py
+++ b/src/awkward/_connect/numexpr.py
@@ -141,8 +141,13 @@ def re_evaluate(local_dict=None):
             or not isinstance(x, ak.contents.Content)
             for x in inputs
         ):
+            input_primitives = [
+                x.data if isinstance(x, ak.contents.NumpyArray) else x for x in inputs
+            ]
             return (
-                ak.contents.NumpyArray(numexpr.re_evaluate(dict(zip(names, inputs)))),
+                ak.contents.NumpyArray(
+                    numexpr.re_evaluate(dict(zip(names, input_primitives)))
+                ),
             )
         else:
             return None
--- a/src/awkward/_connect/numexpr.py
+++ b/src/awkward/_connect/numexpr.py
@@ -91,11 +91,14 @@ def evaluate(
             or not isinstance(x, ak.contents.Content)
             for x in inputs
         ):
+            input_primitives = [
+                x.data if isinstance(x, ak.contents.NumpyArray) else x for x in inputs
+            ]
             return (
                 ak.contents.NumpyArray(
                     numexpr.evaluate(
                         expression,
-                        dict(zip(names, inputs)),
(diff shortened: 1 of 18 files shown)


jpivarski (approved):
Okay, that makes sense. How did you find all of the instances where `__array__` was _implicitly_ being called?

My guess (how I'd go about it): by adding `raise Exception("HERE!")` in `__array__` and seeing where the tests fail.

agoose77:
Nearly; I removed the support for `NumpyArray` and `EmptyArray` for `__array__`, which leaves the base `Content.__array__` that already throws an exception. This was prohibited for the base class some time ago; we expect users who do not wish to be explicit about e.g. `ak.to_numpy` to have a high-level array for such things.

For index nodes, a manual regex was used. We could also remove `__array__` et al. from `Index`, I'm not yet sure how we feel; `Index.__array__` et al. are useful for third-party package authors, who probably aren't going to have much support for placeholders. By the same token, however, we probably need such package authors to support typetracer and placeholders in order to build a better dask ecosystem so they wouldn't need `__array__` either.

PR #2741, opened 2023-10-04
fix: swap names for `__dlpack__`

agoose77:
Somehow, when authoring #2649, I swapped the names for the query descriptor and the dlpack method itself.


jpivarski (approved):
Right.

I'll enable auto-merge.

PR #2742, opened 2023-10-05
ci: improve deployment environment UX

agoose77:
After becoming more familiar with environments, prompted by a comment made by @henryiii, I've decided to share our deployment environments across targets, and rename them accordingly. This should slightly improve UX.

I've also added URLs to the PyPI deployments, and a linter for the YML files in GH.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -120,7 +120,7 @@ jobs:
           - '3.9'
           - '3.8'
 
-    runs-on: macOS-11
+    runs-on: macos-11
 
     env:
       PIP_ONLY_BINARY: cmake,numpy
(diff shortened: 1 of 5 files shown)


.github/workflows/test.yml
```diff
@@ -120,7 +120,7 @@ jobs:
           - '3.9'
           - '3.8'
 
-    runs-on: macOS-11
+    runs-on: macos-11
```
agoose77, line 123: Appease new linter!

jpivarski: I would have capitalized it as MacOS...

jpivarski (approved):
It's my understanding that this will make GitHub recognize the environment as a PyPI deployment and add more buttons. We could use more buttons.

I think it's ready to merge!

PR #2744, opened 2023-10-06
fix: `__array__` accepts positional arguments

agoose77:


agoose77:
Merging as this is a tiny, well defined change.

PR #2747, opened 2023-10-08
fix: update cppyy module

ianna:
cppyy module relied on ak._utils to check the version.
Also we need to import cling. I'm not sure this is the best place, but it works:
```python 
import awkward._connect.cling  # noqa: F401
```

### DIFF
--- a/src/awkward/cppyy.py
+++ b/src/awkward/cppyy.py
@@ -1,6 +1,6 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 
-import awkward as ak
+import packaging.version
 
 _has_checked_version = False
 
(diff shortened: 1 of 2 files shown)


ianna (commented):
@agoose77 and @jpivarski - preparing for PyHEP I've discovered that cppyy test would fail. I wonder if this patch can be integrated in the next release asap. Thanks!

agoose77:
@ianna I presume we need to install cppyy from source to obtain the supported version? Perhaps we should add a CI job to do this.

When obtaining cppyy from Git, do you locally need to bump the version to 3.0.1?

ianna:
> @ianna I presume we need to install cppyy from source to obtain the supported version?

yes

> Perhaps we should add a CI job to do this.
> 
> When obtaining cppyy from Git, do you locally need to bump the version to 3.0.1?

yes

jpivarski (approved):
Yes, `ak._util.parse_version` was removed in https://github.com/scikit-hep/awkward/commit/11e61b6be1ff29001dd1f327dbe16b5fca06cafd#diff-5b25481afa0bba148a54d46dfc93af8141dfa50dcb399c55bb7df34f3af40f64

The dangerous-looking

```python
import awkward._connect.cling
```

happens inside of the non-inspectable `ak.Array.cpp_type` property, so it's safe.

I think this is ready to go and will merge it.

agoose77:
Thanks for merging this @jpivarski, I missed Ianna's reply!

PR #2748, opened 2023-10-11
fix: update cppyy version

ianna:
as discussed at the cppyy meeting the next cppyy version is going to be 3.1.0


agoose77 (approved):

agoose77:
It's my understanding that:

3.0.1 was a speculative version for cppyy. As such, bumping it to 3.1.0 is non-harmful; it's still a yet-unreleased version, and we require the latest version for this code.

Therefore, I've accepted and we can merge! Thanks @ianna

PR #2749, opened 2023-10-11
fix: don't call asarray on `Index` objects internally

agoose77:
This PR follows #2740. I've updated that PR's description for context, but let me restart the problem here!

`nplike.asarray` ultimately invokes NumPy's array interface mechanisms when it acts upon an `Index` or `Content` object. If we don't make this PR change, an array with placeholders that hits this codepath will lead to an exception being thrown.

As outlined in #2740, we *could* update `nplike.asarray` to handle these types, but I'm really in favour of having our internals be strict where possible; both to do less work, but also to make it easier to reason about.

So, this PR replaces any as-yet uncaught uses of `nplike.asarray(index)` with `index.data` if the `nplike` is unchanged, or `nplike.asarray(index.data)` if not.

I haven't been exhaustive; anything that isn't being run in our test suite won't have been caught, but I'm happy enough with our coverage that those will hopefully not be too many that we can't deal with them as they come in.

I haven't added anything to the test suite to catch this. I'm slightly reluctant to e.g. add a strict-mode for our test suite that throws an exception if `Index._array_interface_` is inspected. We could do this, though, and I'd welcome @jpivarski thoughts!

### DIFF
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -262,7 +262,7 @@ class IndexedArray(Content):
             self._index, self._content, parameters=self._parameters
         )
 
-    def mask_as_bool(self, valid_when=True):
+    def mask_as_bool(self, valid_when: bool = True) -> ArrayLike:
         if valid_when:
             return self._index.data >= 0
         else:
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -284,7 +284,7 @@ class IndexedOptionArray(Content):
             valid_when, lsb_order
         )
 
-    def mask_as_bool(self, valid_when=True):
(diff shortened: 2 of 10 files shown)


agoose77:
@jpivarski how do you feel about this PR? 

Namely — considering `nplike.asarray(index)` an "error" internally, expecting it to fail under placeholders, but _not_ adding any tests for this yet

Upfront it means we'll likely see exceptions that we could avoid by being more permissive in `nplike.asarray`, but I hope that this is just a bathtub model curve, and we'll have cleaner internals down the road.

jpivarski (approved):
To make our tests more robust against the possibility of accidentally introducing dependence on `Index.__array__` in the future, without forbidding it among end-users, we could add a private (L3) global setting that makes `Index.__array__` an error when that global setting is turned on (default is off). The tests could then turn it on and be more strict than user code.

For this PR specifically, it all looks good.

PR #2752, opened 2023-10-11
feat!: revert breaking component of `asarray` PR

agoose77:
This PR reverts the breaking parts of #2740 to introduce a deprecation period.


jpivarski (approved):
`Index.__array__` would need the same deprecation period.

PR #2754, opened 2023-10-12
fix: add `highlevel`, `behavior` arguments to composite reducers

agoose77:
A number of high-level operations in Awkward Array did not accept the `highlevel` or `behavior` arguments. I don't think we ever explicitly decided upon this, so I assume that it is an oversight in need of correction!

### DIFF
--- a/src/awkward/_layout.py
+++ b/src/awkward/_layout.py
@@ -36,6 +36,31 @@ def wrap_layout(content, behavior=None, highlevel=True, like=None, allow_other=F
     return content
 
 
+def maybe_highlevel_to_lowlevel(obj):
+    """
+    Args:
+        obj: an object
+
+    Calls #ak.to_layout and returns the result iff. the object is a high-level
+    Awkward object, otherwise the object is returned as-is.
+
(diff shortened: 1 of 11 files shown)


src/awkward/_layout.py
```diff
@@ -36,6 +36,31 @@ def wrap_layout(content, behavior=None, highlevel=True, like=None, allow_other=F
...
+            awkward.highlevel.Array,
+            awkward.highlevel.Record,
+            awkward.highlevel.ArrayBuilder,
+        ),
+    ):
+        return awkward.to_layout(obj)
+    else:
+        return obj
+
+
```
agoose77, line 63: We will get rid of this function once our policies for `ak.to_layout` of scalars are more well defined.

jpivarski (approved):
I think it was an oversight. The functions you found are all of the "pseudoreducers." They were added as a pack, all at once.

jpivarski:
This looks done, it's a straightforward fix, and it has been hours since you touched it, so I'm going to merge it.

PR #2756, opened 2023-10-13
chore: set `pyupgrade` version

agoose77:
When we invoke `pyupgrade`, it should be given the minimum Python version explicitly; it doesn't read `pyproject.toml`.


jpivarski (approved):
Sounds good. I'll merge this now.

PR #2757, opened 2023-10-13
feat: add `.attrs` to highlevel objects

### LINKED ISSUE #1391: xarray-style "attrs", global and per-record field

@philippemiron is converting data from NetCDF4 files into Awkward Arrays, and one of the features we lack is a place to put attributes. These can be descriptions, units, meanings of flags, etc., and a single field can have more than one of these (e.g. so that "units" are programmatically accessible). In general, any JSON-encodable data.

This applies at two levels: globally for a whole array, in such a way that all derived arrays pass on the attributes, and per-record field. The per-record field attributes should only be passed on if the meaning of the field is not changed, and should not be counted as part of the Content node's type. Also, these should be read and written to files as metadata wherever possible.

**Why not use parameter `__doc__`?**

This would work as a per-record field attribute that is passed on whenever the meaning of the field is not changed, and doesn't count as part of the Content node's type. However, it has to be a string, since this is what goes into the Python `__doc__` property (and therefore IPython and Jupyter help). The attributes have to be general JSON-encodable metadata.

Also, this only encodes per-record field attributes, not global attributes.

**Why not use a new parameter?**

It would only be accessible through idioms like `array.layout.parameters["attrs"]`. This is high-level data analyst information, and they shouldn't have to go through `layout`, which is mid-level, for library developers. Also, such an idiom would fail if the layout ever gets buried in another Content node (i.e. the user has to be _aware_ of layouts and how they change, which is not a high-level view!). For example, rearranging records in a RecordArray nests the RecordArray within an IndexedArray for performance reasons; so after certain kinds of slices, `array.layout.parameters["attrs"]` becomes `array.layout.content.parameters["attrs"]`, probably unexpectedly.

**Why not use behavior?**

It's an odd thing to do, but `behavior` is passed from an ak.Array to any new ak.Array derived from it. Any keys in `behavior` that aren't recognized are ignored. However, this would only encode global attributes, not per-record field attributes, and `behavior` (which typically contains function objects and class objects) is not serialized when writing files, or filled when reading files.

**What instead?**

This is really two new features:

   1. Per-record field metadata, as a parameter, that follows all the same rules as `__doc__`, but doesn't have to be a string and is exposed at high-level in a different way (one that allows non-strings). Parameters are already serializable.
   2. Global metadata that follows all the same rules as `behavior`, but is serializable.

They should have names like `attrs`, following [xarray's convention](https://xarray.pydata.org/en/stable/generated/xarray.DataArray.attrs.html), but the per-record field `attrs` should be distinguishable somehow from the global `attrs`.

Perhaps both of these should be dicts (JSON objects) and when we're looking at one field, the two dicts are overlaid, with the per-field keys taking precedence over the global keys? That sounds natural and would minimize the use of names, but it sounds like just the sort of thing that would break something in the future.

What does xarray do?

```python
>>> import xarray as xr
>>> dataarray = xr.DataArray([1, 2, 3], attrs={"wow": "look!"})
>>> dataset = xr.Dataset({"x": dataarray}, attrs={"wow": "wee!", "another": "thing"})
>>> dataset.attrs
{'wow': 'wee!', 'another': 'thing'}
>>> dataset.x.attrs
{'wow': 'look!'}
```

xarray never has this issue because DataArray and Dataset are different types and can never be confused. In Awkward, an ak.Array is an ak.Array is an ak.Array, regardless of whether it's an array of lists of records, an array of just the records, or an array of one of the fields of those records. The global attrs can propagate down when extracting the array of records from the array of lists of records, but when you get to the array of one of the fields of those records, there's a conflict. They should probably be named differently, since "per-field attributes" and "global attributes" are different concepts, but which one gets the exalted name of "`attrs`"? And what would the other one be called?

agoose77:
This will partially fix #1391 by adding `.attrs`, a dictionary of metadata that is associated with the top-level array, with the following semantics:
- Operations with multiple arrays take the set-union of properties, first wins
- Keys beginning with a special prefix (`@`) will be removed at serialisation time.

Whilst touching some of this code, I did some work on type hints and refactoring.


agoose77:
Design question (1) — should `array.attrs` _always_ return a dict, or can it return `None`?

### Return `None`
(invert the below for "always return a dict")  
- Convenient token for "no metadata", allowing optimisations (including non-Awkward cases)
- Harder to set metadata in a non-functional style (unless metadata already set)

jpivarski:
The `_parameters` are internally `None` or `dict` for performance, but the external `parameters` property always returns a `dict` for uniformity. In fact, the property creates the `dict` so that it can be mutably updated from that point onward.

I think this is a good compromise, and it would follow the principle of least surprise if `attrs` does the same thing. In fact, I think this is what `behaviors` does, too.

You were only asking about how it looks from the outside, so I'm voting "always return a dict." But since most users won't be using `attrs` (that's the situation we're starting from before this first implementation), so using `None` internally minimizes the performance impact of adding a new feature across all `ak.Arrays`.

If we really need a `has_attrs: bool` (more likely a `has_attr(name)`), then we can add it.

-------------

Oh, unlike `parameters`, I think we should _not_ conflate key-not-found with value-is-None. We should allow users to consider `None` a meaningful value for an attribute, distinct from not having that attribute. `attrs` is the users' space, much more so than `parameters`.

agoose77:
> The `_parameters` are internally `None` or `dict` for performance, but the external `parameters` property always returns a `dict` for uniformity. In fact, the property creates the `dict` so that it can be mutably updated from that point onward.
> ...
> I think this is a good compromise, and it would follow the principle of least surprise if attrs does the same thing. In fact, I think this is what behaviors does, too.

Actually, `Array.behavior` returns `None | dict`, making it possible to predict whether the array will use the global behavior lookup.

`attrs` doesn't have that benefit; there's only a single namespace to look at, so this motivation for exposing `None` disappears. I think it's therefore OK to hide the detail from users.

-------------

> Oh, unlike `parameters`, I think we should _not_ conflate key-not-found with value-is-None. We should allow users to consider `None` a meaningful value for an attribute, distinct from not having that attribute. `attrs` is the users' space, much more so than `parameters`.

Agreed.

agoose77:
This PR indicates that it would be nice to have something like
```python
ctx = HighLevelCtx()

layouts = ctx.finalise([
    ak.to_layout(ctx.maybe_highlevel(a)) for a in arrays
])

backend = ctx.backend
behavior = ctx.behavior

```

Rather than the existing stateless functional API that revisits inputs multiple times. This function would ensure all layouts have the proper "final" backend.

As such, it's on-hold for PR #2763 and another PR.

agoose77:
@jpivarski this is a big PR.

I've split it into commits that should be a bit more manageable.

I still need to do a pass to confirm that the various policy settings used by `ak.to_layout` are sensible, but I'd benefit from the main review to confirm that the (internal) API changes are a step in the right direction.

The purpose of the `HighLevelContext` addition to this PR was make the general idiom of pulling behaviors off of high-level object(s) and applying them to the results is less error prone. The use of `ctx.wrap_layout` leads to an Exception if the context hasn't been explicitly finalised, making it harder to pass the wrong `behavior` argument around.

There's a test failure pertaining to the choice to permit records, that I'll revisit.

src/awkward/_attrs.py
```diff
@@ -0,0 +1,44 @@
...
+    # An explicit 'attrs' always wins.
+    if attrs is not None:
+        return attrs
+
+    copied = False
+    for x in reversed(arrays):
+        x_attrs = attrs_of_obj(x)
+        if x_attrs is None:
+            continue
+        elif attrs is None:
```
jpivarski: (PyLint and Ruff are starting to make suggestions that go against what I think of as good coding practices. Here, it says, "unnecessary 'elif' after 'continue'." Well, yeah, that's so that the `if`-`elif`-`elif`-`else` block is complete. I think Ruff recently said a similar thing about a `return` statement.

While it's true that `elif` is allowing for a code path that can't happen, we're not thinking of these things as individual lines stepping sequentially, but as blocks and patterns. Breaking up a `if`-`elif`-`elif`-`else` block so that each line has the minimum necessary code flow reduces high-level patterns to low-level instruction-following.)

agoose77:
Do we have a formal policy regarding which functions should prohibit/support record objects? e.g. `ak.is_none` on a record.

jpivarski (commented):
It is a big PR with a lot of differences, although many of them come from adding the single extra `attrs` argument to lots of functions, which in many cases turns them from single-line to multi-line due to black.

The context object is a big difference. I see that this context object is private, for our own use, and I'm not certain that it's necessary for keeping track of two pieces of data (`behavior` and `attrs`), but it's fine.

(In your comment, I thought you were referring to a public context object, so that people can say,

```python
with ak.Context(behavior, attrs):
    ak.something...
```

although that doesn't make a lot of sense: one would want the `behavior` and `attrs` to be permanently glued to specific objects and not others. But I guess you were talking about a private context.)

I'm in favor of the direction that this PR is heading.

jpivarski:
> Do we have a formal policy regarding which functions should prohibit/support record objects? e.g. `ak.is_none` on a record.

I think it's fine. `ak.is_none(something)` should probably return `np.bool_(True)` if `something is None` and `np.bool_(False)` if `something` is any other scalar, including `ak.Record` or a single string. At least, that's what I would naively expect.

The function is doing something that could be done in other ways in Python, namely `something is None`, but Awkward functions partly overlap territory already covered by NumPy, so they can partly overlap territory already covered by Python, too—as long as the main focus is on things that can't be done other ways. It would be odd if it raised an error saying, "I could tell you the answer, but instead I'm going to tell you to ask your question in a different way," and it would lead to code that has to check

```python
if something is None:
    ...
elif not isinstance(something, ak.Array):
    ...
else:
    ...
```

If an Awkward function has a _natural_ extension into space covered by NumPy or Python, then we're reducing work on the user's end by handling array, scalar, and `None` in a consistent way.

src/awkward/_attrs.py
agoose77: Yes, the general best-practices are at odds with the way we write code for Awkward. I try to stick to what we've decided to do internally, but I'm not against following ruff etc; there's merit to being standardised and tool-friendly. Incidentally, my mental model for the early-return approach is that it simplifies the context of the program; there are fewer cases to consider. 

Personally I'd prefer to just let ruff do its thing rather than spend time tweaking it, but it depends upon how strongly you feel about the style!

agoose77:
> The context object is a big difference. I see that this context object is private, for our own use, and I'm not certain that it's necessary for keeping track of two pieces of data (behavior and attrs), but it's fine.

It's certainly not _necessary_. But, I found the `metadata_of` approach a bit more error-prone and verbose. I think the change to use `ctx` improves readability. The challenge here is not to over-optimise things, which is the motivation behind getting a sense check on the direction.

The rationale for `HighLevelContext` is that if one needs to obtain a layout object, we should be forced to consider its behavior and/or attrs.

agoose77:
I've changed `ak.to_regular` so that it errors upon encountering a bare record. That might be too strict.

agoose77:
@ianna this is the PR that I mentioned on Slack — would you be able to lend a hand in passing `attrs` through Numba? The only difference to `Array.behavior` is that we need to drop any `attrs` keys that start with a special token `"_transient_"`.

jpivarski:
The token for transient attrs is `_transient_`: that's a little long. How about a special symbol, like `@`?

I don't think that's relevant for Numba, though. All of the attrs can be passed through Numba because they don't need to be serialized. (They're attached to the type instance, and that's a compile-time thing, not a runtime thing. If it seems like they're needed in boxing, to produce the output array, I'm sure we can arrange a work-around that calls a function and puts these things together on the Python side, rather than the Numba runtime side.)

agoose77:
> The token for transient attrs is `_transient_`: that's a little long. How about a special symbol, like `@`?

I prefer that, actually!

> I don't think that's relevant for Numba, though. All of the attrs can be passed through Numba because they don't need to be serialized. (They're attached to the type instance, and that's a compile-time thing, not a runtime thing. If it seems like they're needed in boxing, to produce the output array, I'm sure we can arrange a work-around that calls a function and puts these things together on the Python side, rather than the Numba runtime side.)

Ah, in my recent memory we did rely on serialisation in Numba for behavior, and this would necessarily suffer the same restrictions. But, we have a strategy (`"@"` prefix) for dropping these problem keys.

jpivarski:
It will be a little surprising for transient attributes to not survive a trip through Numba. Users probably don't expect Numba to pass data through serialization, though it does.

It comes up in the boxing functions, which is the only place where a type (and the things attached to it, like `behaviors` and `attrs`) are used at runtime in Numba-compiled code, and therefore it serializes. But instead of trying to pass the type object directly into that code, we can reference it somehow with a Python function call. As long as we obtain the object via Python and not lowered Numba, we can avoid putting it through serialization.

agoose77:
I wasn't happy with the sharing of responsibilities that `HighLevelContent` was undertaking. I've modified the API such that `HighLevelContext` is only responsible for the behavior and attrs of the arrays it accepts. I also restored the context-manager behavior, so the name is even more apt.


To ensure that the layouts have the same backend, I added `ensure_same_backend`, which calls `to_backend` on any object which implements the method.


The purpose of this context manager is to make it difficult to forget to take the behavior and attrs from participating array objects:
1. The `ctx.unwrap` method is easy to type
2. The indentation of the context manager requires one to localise the layout handling code
3. `HighLevelContext` handles both attrs and behavior
4. `ctx.wrap` makes it possible to avoid referring to behavior or attrs at all

e.g. `ak.mean`:
```python
with HighLevelContext(behavior=behavior, attrs=attrs) as ctx:
    x_layout, weight_layout = ensure_same_backend(
        ctx.unwrap(x, allow_record=False, primitive_policy="error"),
        ctx.unwrap(
            weight,
            allow_record=False,
            allow_unknown=False,
            primitive_policy="error",
            none_policy="pass-through",
        ),
    )

x = ctx.wrap(x_layout)
weight = ctx.wrap(weight_layout, allow_other=True)
```

src/awkward/highlevel.py
```diff
@@ -1531,7 +1535,19 @@ def __reduce_ex__(self, protocol: int) -> tuple:
...
+        else:
+            attrs = without_transient_attrs(self._attrs)
+
+        return unpickle_array_schema_1, (
+            form.to_dict(),
+            length,
+            container,
+            behavior,
+            attrs,
+        )
```
agoose77, line 1574: @jpivarski this is something I'd like you to pay attention to — it means that newer Awkward Array pickles won't unpickle in older Awkward Array versions (even in the 2.x series). 

However, this PR will have to break unpickling in one form or another; it introduces a new state item that older Awkward versions won't be able to handle.

The intention of writing new, named functions is to encode the pickle scheme as a function name, rather than a version number alongside the pickle state. That way, older pickles are easily distinguished from newer ones in future versions of Awkward.

src/awkward/_pickle.py
```diff
@@ -72,11 +73,24 @@ def get_custom_reducer() -> PickleReducer | None:
...
+    finally:
+        _DISABLE_CUSTOM_REDUCER = old_value
+
+
+def custom_reduce(obj, protocol: int) -> tuple | NotImplemented:
+    if (plugin := get_custom_reducer()) is None or _DISABLE_CUSTOM_REDUCER:
         return NotImplemented
-    return plugin(obj, protocol)
+    else:
+        return plugin(obj, protocol)
```
agoose77, line 96: We want to be able to disable this custom reduction mechanism within our test suite. I think the easiest method is this one: temporarily disabling the custom reducer plugin.

agoose77:
> **Note**
> We should make a follow-up PR to deprecate `behavior` from `recursively_apply`

src/awkward/_attrs.py
jpivarski: Oh, we'll leave it. I was just complaining. Eventually these will become `match`-`case`, and then we'd have to re-figure-out how they were all linked up originally.

src/awkward/highlevel.py
jpivarski: Breaking forward-compatibility in pickles is bound to happen, I agree. Is this PR going to be in 2.5.0? That would be a good boundary at which to introduce such breakage. If someone is having trouble with pickled arrays, we can tell them that this is the boundary to avoid.

src/awkward/_pickle.py
jpivarski: Okay!

src/awkward/_connect/numba/arrayview.py
```diff
@@ -291,6 +301,9 @@ def lower_const_view(context, builder, viewtype, view):
     proxyout.pylookup = context.add_dynamic_addr(
         builder, id(lookup), info=str(type(lookup))
     )
+    proxyout.pyattrs = context.add_dynamic_addr(
+        builder, id(attrs), info=str(type(attrs))
+    )
```
jpivarski: One request: for the Numba implementation (not including ArrayBuilder), could you pass the `attrs` through by just attaching the `attrs` to the `lookup`? These are both Python objects, which is what makes them cumbersome to pass through lowered code, but we should only have to do that once. Every time the `lookup`'s `PyObject*` is copied, it slows down iteration in Numba-compiled functions. There is a one-to-one relationship between the `lookup` and the original array (the `lookup` is both holding the original array's references and also presenting them in a pointer/bytes-friendly form), so an array's `attrs` can be simply attached to its `lookup`.

Also, this `add_dynamic_addr` puts a pointer into the space of `static` variables. That's probably limited, so there's an implicit upper limit on the number of arrays that can be passed through Numba-compiled functions. I don't know what that limit is, but by making two pointers for every array, we approach it twice as fast.

In the boxing function, you'd be able to access the `attrs_obj` by

```python
c.pyapi.object_getattr_string(lookup_obj, "_attrs")
```

Oddly enough, we're serializing and unserializing the entire `behaviors` dict and that seems to be okay. It seems to me that the `behaviors` could be carried on the `lookup` as well. I'm surprised that no one in Coffea ran into the issue of putting a lambda function in `behaviors` and then not being able to pass it through a Numba-compiled function.

jpivarski (changes requested):
This is a huge work, and tests/test_2757_attrs_metadata.py is very extensive, parameterizing over so many functions to make sure that `attrs` are correctly passed through all of them.

My only request is about Numba: see below. If you have questions about that, I can help during our meeting tomorrow.

src/awkward/_connect/numba/arrayview.py
```diff
@@ -152,7 +153,17 @@ def to_numbatype(form):
...
 
-@numba.extending.typeof_impl.register(ak._lookup.Lookup)
+class NumbaLookup(Lookup):
+    def __init__(self, layout, attrs, generator=None):
+        super().__init__(layout, generator=generator)
+        self._attrs = attrs
+
+    @property
+    def attrs(self):
+        return self._attrs
```
agoose77, line 163: @jpivarski I've opted to extend the `Lookup` class to attach `attrs` — I don't think this makes sense for the base class.

Are you OK with this approach?

jpivarski: It didn't need to be a subclass—`Lookup` is already a "bag of stuff" and having more stuff (`attrs`) doesn't break the model.

A subclass is more complex than it needs to be, but it can be simplified later (not this PR).

jpivarski (approved):

Issue #1325, opened 2022-02-26
The "nan" versions of all the reducers and statistics functions


jpivarski:
### Description of new feature

Like `np.nanmean`, raised by @philippemiron.

jpivarski:
When [1.9.0rc2](https://github.com/scikit-hep/awkward-1.0/actions/runs/2193146418) is uploaded, you'll be able to

```bash
pip install 'awkward>=1.9.0rc2'
```

and

```python
import awkward._v2 as ak
ak.nanmean(...)
```

PR #2758, opened 2023-10-16
fix: export `nan` variants

agoose77:
We have high-level `ak.XXX` reducers that include `nan`-aware variants. We should export those!


jpivarski (approved):
Oops; that was added and tested in `awkward._v2` (https://github.com/scikit-hep/awkward/issues/1325#issuecomment-1103381403), but must have gotten lost when the submodule got promoted.

I'll merge this now.

PR #2759, opened 2023-10-16
fix: update `__class__` for both `layout` and `behavior` consistently

agoose77:
Presently, setting `.behavior` and `.layout` does not consistently update `__class__`. This PR rectifies the behavior such that `__class__` is updated for both attributes!


agoose77:
@jpivarski I reckon you might've seen this earlier, before I noticed that the CI was failing. I've fixed it now!

jpivarski (approved):
I thought we did this, but I guess they must have gotten out of sync. The helper function helps!

I'll merge this now.

PR #2761, opened 2023-10-19
refactor: drop `behavior` broadcasting

agoose77:
The `behavior` for one array given to broadcasting functions no longer _needs_ to be broadcast against the other array behaviors. Thus, let's remove that restriction.

This would break users who introduce behaviors on array X, and expect `ak.broadcast_arrays` to merge them.

### DIFF
--- a/src/awkward/_backends/dispatch.py
+++ b/src/awkward/_backends/dispatch.py
@@ -1,7 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 from __future__ import annotations
 
-from collections.abc import Set
+from collections.abc import Collection
 
 from awkward._backends.backend import Backend
 from awkward._nplikes.numpy import Numpy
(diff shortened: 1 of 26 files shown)


agoose77:
[5a2a95a](https://github.com/scikit-hep/awkward/pull/2761/commits/5a2a95ad8449ea849cf8375b294c7fe991380b49) is preparation for another PR that made sense here.

jpivarski (approved):
Now that string-broadcasting is built-in, no longer comes from `ak.behavior`, the `broadcast_and_apply` function no longer needs to ask for it as an argument, and that decouples n-ary → n-ary functions so that they don't have to mix `behaviors` anymore.

You're right that this changes how `ak.broadcast_arrays` works, and is therefore a technically breaking change, but it is extremely niche. Behaviors are generally thought of as global settings (Coffea, Vector), and arrays can carry their own `behavior` just in case they're not. Encountering different `behavior` on the inputs to `ak.broadcast_arrays` would be that surprising, exceptional case. _Depending_ on it [would be very weird](https://xkcd.com/1172/).

I think this can go in a patch release. And I'm happy with the changes as they are here; please go ahead and merge when you're ready!

PR #2765, opened 2023-10-22
fix: add policy for length erasure

### LINKED ISSUE #2764: Typetracer can't slice with `(int, str, int)`
### Version of Awkward Array

HEAD

This is either a bug in Tiled's implementation of Awkward slices or in Awkward, but from the error message, I think it's Awkward.

With an `array` that has `RecordType(ListType(RecordType(...))`, I can't slice

```python
array[0, "field_name", 0]
```

without getting

```
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/highlevel.py", line 1014, in __getitem__
    prepare_layout(self._layout[where]), self._behavior, allow_other=True
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/contents/content.py", line 519, in __getitem__
    return self._getitem(where)
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/contents/content.py", line 569, in _getitem
    return out._getitem_at(0)
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/contents/indexedarray.py", line 277, in _getitem_at
    return self._content._getitem_at(where)
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/contents/recordarray.py", line 447, in _getitem_at
    raise ak._errors.index_error(self, where)
IndexError: cannot slice RecordArray (of length 0) with 0

  File "/home/jpivarski/irishep/tiled/tiled/client/awkward.py", line 55, in __getitem__
    return self.read(slice=slice)
  File "/home/jpivarski/irishep/tiled/tiled/client/awkward.py", line 33, in read
    awkward.typetracer.touch_data(proxy_array[slice])
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/highlevel.py", line 1012, in __getitem__
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/_errors.py", line 67, in __exit__
  File "/home/jpivarski/mambaforge/envs/tiled/lib/python3.10/site-packages/awkward/_errors.py", line 82, in handle_exception
IndexError: cannot slice RecordArray (of length 0) with 0

    <Array-typetracer [...] type='## * {offsets: var * int64, file: var * {...'>

with

    (0, 'era', 0)
```

My work-around is https://github.com/jpivarski/tiled-uproot/pull/3/commits/586f3c565e8ed85eccd3d8891a71c270a96616f7.

The real `RecordArray` doesn't have length zero (or the work-around wouldn't work). Could this be a case in which an effect on a Form is computed by converting it to a length-0 array, then converting back? (If so, it will eventually get us in the Dask world, too.)

Cc: @danielballan

agoose77:
Fix #2764 by adding support for recursive erasure of length information. 

Instead of `forget_length: bool`, we have `forget_length: Literal["keep", "drop_outer", "drop_recursive"]`, which enables us to decide how deep the lengths are forgotten. For the `Form → Content` case, we always want recursively to drop lengths.

> **Warning**
> This PR introduces a deprecation in the `forget_length` argument.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -1465,7 +1465,7 @@ def typetracer_with_report(
     forget_length: bool = True,
 ) -> tuple[ak.contents.Content, TypeTracerReport]:
     layout = form.length_zero_array(highlevel=False).to_typetracer(
-        forget_length=forget_length
+        length_policy="drop_recursive" if forget_length else "keep"
     )
     report = TypeTracerReport()
     _attach_report(layout, form, report, getkey)
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -17,7 +17,15 @@ from awkward._parameters import (
 )
 from awkward._regularize import is_integer, is_integer_like
 from awkward._slicing import NO_HEAD
-from awkward._typing import TYPE_CHECKING, Callable, Final, Self, SupportsIndex, final
+from awkward._typing import (
+    TYPE_CHECKING,
+    Callable,
+    Final,
+    Literal,
+    Self,
+    SupportsIndex,
+    final,
+)
 from awkward._util import UNSET
 from awkward.contents.bytemaskedarray import ByteMaskedArray
 from awkward.contents.content import Content
@@ -307,13 +315,18 @@ class BitMaskedArray(Content):
         )
         self._content._to_buffers(form.content, getkey, container, backend, byteorder)
 
-    def _to_typetracer(self, forget_length: bool) -> Self:
+    def _to_typetracer(
+        self, length_policy: Literal["keep", "drop_outer", "drop_recursive"]
+    ) -> Self:
         tt = TypeTracer.instance()
+        mask = self._mask.to_nplike(tt)
         return BitMaskedArray(
-            self._mask.to_nplike(tt),
-            self._content._to_typetracer(False),
+            mask.forget_length() if length_policy != "keep" else mask,
+            self._content._to_typetracer(
+                "keep" if length_policy == "drop_outer" else length_policy
+            ),
             self._valid_when,
-            unknown_length if forget_length else self.length,
+            unknown_length if length_policy != "keep" else self.length,
             self._lsb_order,
             parameters=self._parameters,
         )
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -19,7 +19,15 @@ from awkward._parameters import (
 )
 from awkward._regularize import is_integer_like
 from awkward._slicing import NO_HEAD
-from awkward._typing import TYPE_CHECKING, Callable, Final, Self, SupportsIndex, final
+from awkward._typing import (
+    TYPE_CHECKING,
+    Callable,
+    Final,
+    Literal,
+    Self,
+    SupportsIndex,
+    final,
+)
 from awkward._util import UNSET
 from awkward.contents.content import Content
 from awkward.errors import AxisError
@@ -229,12 +237,16 @@ class ByteMaskedArray(Content):
         )
         self._content._to_buffers(form.content, getkey, container, backend, byteorder)
 
-    def _to_typetracer(self, forget_length: bool) -> Self:
+    def _to_typetracer(
+        self, length_policy: Literal["keep", "drop_outer", "drop_recursive"]
+    ) -> Self:
         tt = TypeTracer.instance()
         mask = self._mask.to_nplike(tt)
         return ByteMaskedArray(
-            mask.forget_length() if forget_length else mask,
-            self._content._to_typetracer(False),
+            mask.forget_length() if length_policy != "keep" else mask,
+            self._content._to_typetracer(
+                "keep" if length_policy == "drop_outer" else length_policy
+            ),
             self._valid_when,
             parameters=self._parameters,
         )
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -238,7 +238,7 @@ class Content:
         self,
         forget_length: bool = UNSET,
         *,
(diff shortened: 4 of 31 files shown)


jpivarski (commented):
This is an L2 public function, so we can, in principle, be strict with users. However, it doesn't seem to be necessary here. Changing

```python
forget_length: bool
```

into

```python
length_policy: Literal["keep", "drop_outer", "drop_recursive"]
```

requires calling functions to change, so so there's a deprecation cycle. But it could have been changed to

```python
forget_length: bool | Literal["false", "true", "recursive"]
```

such that `False` corresponds to `"false"`, `True` corresponds to `"true"` (dropping just the outer length, which is the old behavior), and `"recursive"` is the only new behavior: dropping lengths recursively.

That way, there would be no need to make downstream libraries change. Since we're strictly extending the old behavior (new set of options is a superset of the old set of options) and there's a natural way to keep using the old words while introducing a new word, this is less of an imposition. Only downstream libraries that want the new behavior need to do anything; others don't even need to know about the change. (In addition, I have a more immediate understanding of what "forget length" means than "length policy".)

As another alternative, perhaps

```python
forget_length: bool | Literal["recursive"]
```

to be tighter? It depends on whether it's important for all options to have identical type. (Well, they do, even now: it's a union type.)

Also, why string literals and not enums? Do any enums naturally map onto false and true?

agoose77:
> This is an L2 public function, so we can, in principle, be strict with users. However, it doesn't seem to be necessary here. Changing
> 
> ```python
> forget_length: bool
> ```
> 
> into
> 
> ```python
> length_policy: Literal["keep", "drop_outer", "drop_recursive"]
> ```
> 
> requires calling functions to change, so so there's a deprecation cycle. But it could have been changed to
> 
> ```python
> forget_length: bool | Literal["false", "true", "recursive"]
> ```
> 
> such that `False` corresponds to `"false"`, `True` corresponds to `"true"` (dropping just the outer length, which is the old behavior), and `"recursive"` is the only new behavior: dropping lengths recursively.

My personal preference is fairly strongly against mixing types like this, especially at L2 where we can move further away from interactive-friendly APIs towards stricter typing.

> That way, there would be no need to make downstream libraries change. Since we're strictly extending the old behavior (new set of options is a superset of the old set of options) and there's a natural way to keep using the old words while introducing a new word, this is less of an imposition. Only downstream libraries that want the new behavior need to do anything; others don't even need to know about the change. 

That is a benefit, though, of gradually extending the argument.

> (In addition, I have a more immediate understanding of what "forget length" means than "length policy".)

Naming things is hard :( I like `policy` because it implies to me that there's more than two options. 

> As another alternative, perhaps
> 
> ```python
> forget_length: bool | Literal["recursive"]
> ```
> 
> to be tighter? It depends on whether it's important for all options to have identical type. (Well, they do, even now: it's a union type.)
> 
> Also, why string literals and not enums? Do any enums naturally map onto false and true?

Enum's don't play as nicely with typing as I would like. I don't think you can type a function such that it accepts a literal string _or_ an enum value; even `StrEnum`s are considered different types. The benefit for using enums is the safety; a misspelled identifier throws a `NameError`. I'm finding with editor annotations that the use of type hints helps eliminate this kind of mistake, and it's my hope that we'll further concrete this once mypy is in the CI. Furthermore, defining classes for a single functions' options feels clunky vs inlining the literal types.

jpivarski:
My one "on the other hand" is that this is an L2 function and we can be strict. But on the first hand, it just seems like this will appear to downstream dependencies as deck-chair shuffling, forcing them to change their code and introduce a version dependence without an obvious benefit.

A union type is a single type. As I understand from your argument, it is the editors/IDEs that will most strongly pass on the benefit of this. (That's why a type hint is preferred over an enum.) Do the editors have a problem with union types? Does it not tab-complete or show you a menu of possible completions if the type is a union type?

agoose77:
> A union type is a single type. As I understand from your argument, it is the editors/IDEs that will most strongly pass on the benefit of this. (That's why a type hint is preferred over an enum.) Do the editors have a problem with union types? Does it not tab-complete or show you a menu of possible completions if the type is a union type?

No, LSP and other tools support unions just fine — we shouldn't discard a union for that reason. 

My rationale is essentially that this should always have supported more than one option *I think*; dropping lengths is a deeper operation than the top level. But I'm working on a few things at the moment, and this is not the hill to die on! So, if you've not been convinced in favour of deprecation in this PR, then I'm happy to kick the bucket down the road to a future "deprecation cycle" and/or not do it :)

jpivarski:
I'll need to look into it more deeply, but I don't see how this is what is needed to fix #2764. The outer length was the only length dropped from type-tracers for a reason. (Partitioning only happens at top level.) I'll let you know if I'm wrong, but I think it can be fixed by a smaller change.

So we'll put this aside for now. I'll make this PR a draft. If #2764 can be solved without it, we can close it.

agoose77:
> The outer length was the only length dropped from type-tracers for a reason. (Partitioning only happens at top level.) I'll let you know if I'm wrong, but I think it can be fixed by a smaller change.

Whilst it's true that we envisage dropping lengths for top-level partitions, I think we now use this function more generally than that. For example, taking an _existing_ array and converting it to typetracer will produce an array whose buffers nearly _all_ have known lengths. There may be cases where this is intended, but others where it is not. 

So, whilst we can certainly fix #2764 by doing something special for the typetracer factory method, I think handling all of the other known cases we want to support means adding more policies for length erasure.

agoose77:
After discussing this with @jpivarski in our weekly Zoom meeting, we've concluded that lengths should never be partially forgotten — it should be all or nothing, as *most* of the time we don't know lengths because we don't know buffers, i.e. interior nodes cannot have known lengths.

The only case that we'd lose by choosing recursive forgetfulness is when a concrete layout is partitioned into smaller parts. That's not something we do for dask-awkward, nor is it likely that useful.

PR #2766, opened 2023-10-23
feat: add `to_buffer` and helper methods in layout builder

ManasviGoyal:
Added some methods to be able to the features of `LayoutBuilder.h` in `C` where `std::map` can't be used.
- `to_buffer` method for each builder to copy and concatenate the data of each builder buffer to user-defined pointer based on the given node name instead of passing a map.
- some other helper functions to extract the names, sizes (in bytes) and the number of buffers in a separate namespace.

### DIFF
--- a/header-only/layout-builder/awkward/LayoutBuilder.h
+++ b/header-only/layout-builder/awkward/LayoutBuilder.h
@@ -143,6 +143,15 @@ namespace awkward {
             buffers["node" + std::to_string(id_) + "-data"]));
         }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-data"))
+          data_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
+      }
+
       /// @brief Copies and concatenates all the accumulated data in the builder
       /// to a map of user-allocated buffers.
       ///
@@ -322,6 +331,17 @@ namespace awkward {
         content_.to_buffers(buffers);
       }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-offsets"))
+          offsets_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
+        content_.to_buffer(buffer, name);
+      }
+
       /// @brief Copies and concatenates all the accumulated data in the builder
       /// to a map of user-allocated buffers.
       ///
@@ -403,6 +423,9 @@ namespace awkward {
 
       void
       to_buffers(std::map<std::string, void*>& /* buffers */) const noexcept {}
+
+      void
+      to_buffer(void* /* buffer */, const char* /* name */) const noexcept {}
 
       /// @brief Copies and concatenates all the accumulated data in the builder
       /// to a map of user-allocated buffers.
@@ -584,6 +607,17 @@ namespace awkward {
         for (size_t i = 0; i < fields_count_; i++)
           visit_at(contents, i, [&buffers](auto& content) {
             content.builder.to_buffers(buffers);
+          });
+      }
+
+      /// @brief Copies and concatenates the accumulated data in the buffers of the
+      /// builder contents to user-defined pointers if the given node name matches
+      /// with the node associated with that builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        for (size_t i = 0; i < fields_count_; i++)
+          visit_at(contents, i, [&buffer, &name](auto& content) {
+            content.builder.to_buffer(buffer, name);
           });
       }
 
@@ -797,6 +831,17 @@ namespace awkward {
           });
       }
 
+      /// @brief Copies and concatenates the accumulated data in the buffers of the
+      /// builder contents to user-defined pointers if the given node name matches
+      /// with the node associated with that builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        for (size_t i = 0; i < fields_count_; i++)
+          visit_at(contents, i, [&buffer, &name](auto& content) {
+            content.to_buffer(buffer, name);
+          });
+      }
+
       /// @brief Copies and concatenates all the accumulated data in the builder
       /// to a map of user-allocated buffers.
       ///
@@ -975,6 +1020,14 @@ namespace awkward {
       void
       to_buffers(std::map<std::string, void*>& buffers) const noexcept {
         content_.to_buffers(buffers);
+      }
+
+      /// @brief Copies and concatenates the accumulated data in the buffers of the
+      /// builder content to user-defined pointers if the given node name matches
+      /// with the node associated with that builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        content_.to_buffer(buffer, name);
       }
 
       /// @brief Copies and concatenates all the accumulated data in the builder
@@ -1173,6 +1226,17 @@ namespace awkward {
         content_.to_buffers(buffers);
       }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-index"))
+          index_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
+        content_.to_buffer(buffer, name);
+      }
+
       /// @brief Copies and concatenates all the accumulated data in the builder
       /// to a map of user-allocated buffers.
       ///
@@ -1300,6 +1364,14 @@ namespace awkward {
       void
       to_buffers(std::map<std::string, void*>& buffers) const noexcept {
         content_.to_buffers(buffers);
+      }
+
+      /// @brief Copies and concatenates the accumulated data in the buffers of the
+      /// builder content to user-defined pointers if the given node name matches
+      /// with the node associated with that builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        content_.to_buffer(buffer, name);
       }
 
       /// @brief Copies and concatenates all the accumulated data in the builder
@@ -1501,6 +1573,17 @@ namespace awkward {
         mask_.concatenate(reinterpret_cast<int8_t*>(
             buffers["node" + std::to_string(id_) + "-mask"]));
         content_.to_buffers(buffers);
+      }
+
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-mask"))
+          mask_.concatenate(reinterpret_cast<int8_t*>(buffer));
+        content_.to_buffer(buffer, name);
       }
 
       /// @brief Copies and concatenates all the accumulated data in the builder
@@ -1756,6 +1839,19 @@ namespace awkward {
         mask_.append(reinterpret_cast<uint8_t*>(
             buffers["node" + std::to_string(id_) + "-mask"]), mask_.length() - 1, 0, 1);
         content_.to_buffers(buffers);
+      }
+
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-mask")) {
+          mask_.concatenate_from(reinterpret_cast<uint8_t*>(buffer), 0, 1);
+          mask_.append(reinterpret_cast<uint8_t*>(buffer), mask_.length() - 1, 0, 1);
+        }
+        content_.to_buffer(buffer, name);
       }
 
       /// @brief Copies and concatenates all the accumulated data in the builder
@@ -2027,6 +2123,28 @@ namespace awkward {
         for (size_t i = 0; i < contents_count_; i++)
           visit_at(contents_, i, [&buffers](auto& content) {
             content.to_buffers(buffers);
+          });
+      }
+
+      /// @brief Copies and concatenates the accumulated data in the builder buffers to
+      /// user-defined pointers if the given node name matches with any one of the nodes
+      /// associated with the builder; otherwise, it searches the builder contents to
+      /// locate a matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        auto index_sequence((std::index_sequence_for<BUILDERS...>()));
+
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-tags")) {
+          tags_.concatenate(reinterpret_cast<TAGS*>(buffer));
+        }
+
+        else if (std::string(name) == std::string("node" + std::to_string(id_) + "-index")) {
+          index_.concatenate(reinterpret_cast<INDEX*>(buffer));
+        }
+
+        for (size_t i = 0; i < contents_count_; i++)
+          visit_at(contents_, i, [&buffer, &name](auto& content) {
+            content.to_buffer(buffer, name);
           });
       }
 
--- a/header-only/layout-builder/awkward/LayoutBuilder.h
+++ b/header-only/layout-builder/awkward/LayoutBuilder.h
@@ -2241,4 +2241,39 @@ namespace awkward {
   }  // namespace LayoutBuilder
 }  // namespace awkward
 
+namespace {
+
+  /// @brief Helper function to retrieve the names of the buffers.
+  template<typename LayoutBuilder>
+  std::vector<std::string> buffer_name_helper(const LayoutBuilder* builder) {
+    std::map <std::string, size_t> names_nbytes = {};
+    std::vector<std::string> buffer_name;
+    builder->buffer_nbytes(names_nbytes);
+    for (auto it: names_nbytes) {
+      buffer_name.push_back(it.first);
+    }
+    return buffer_name;
+  }
+
+  /// @brief Helper function to retrieve the sizes (in bytes) of the buffers.
+  template<typename LayoutBuilder>
+  std::vector<size_t> buffer_size_helper(const LayoutBuilder* builder) {
+    std::map <std::string, size_t> names_nbytes = {};
+    std::vector<size_t> buffer_size;
+    builder->buffer_nbytes(names_nbytes);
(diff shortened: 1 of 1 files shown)


header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -143,6 +143,15 @@ namespace awkward {
             buffers["node" + std::to_string(id_) + "-data"]));
         }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-data"))
+          data_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
```
jpivarski:
```suggestion
        if (std::string(name) == std::string("node" + std::to_string(id_) + "-data")) {
          data_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
        }
```

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -322,6 +331,17 @@ namespace awkward {
...
       }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-offsets"))
+          offsets_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
```
jpivarski:
```suggestion
        if (std::string(name) == std::string("node" + std::to_string(id_) + "-offsets")) {
          offsets_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
        }
```

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -587,6 +610,17 @@ namespace awkward {
...
 
+      /// @brief Copies and concatenates the accumulated data in the buffers of the
+      /// builder contents to user-defined pointers if the given node name matches
+      /// with the node associated with that builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        for (size_t i = 0; i < fields_count_; i++)
+          visit_at(contents, i, [&buffer, &name](auto& content) {
+            content.builder.to_buffer(buffer, name);
+          });
```
jpivarski:
```suggestion
        for (size_t i = 0; i < fields_count_; i++) {
          visit_at(contents, i, [&buffer, &name](auto& content) {
            content.builder.to_buffer(buffer, name);
          });
        }
```

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -797,6 +831,17 @@ namespace awkward {
...
 
+      /// @brief Copies and concatenates the accumulated data in the buffers of the
+      /// builder contents to user-defined pointers if the given node name matches
+      /// with the node associated with that builder.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        for (size_t i = 0; i < fields_count_; i++)
+          visit_at(contents, i, [&buffer, &name](auto& content) {
+            content.to_buffer(buffer, name);
+          });
```
jpivarski:
```suggestion
        for (size_t i = 0; i < fields_count_; i++) {
          visit_at(contents, i, [&buffer, &name](auto& content) {
            content.to_buffer(buffer, name);
          });
        }
```

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -1173,6 +1226,17 @@ namespace awkward {
...
       }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-index"))
+          index_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
```
jpivarski:
```suggestion
        if (std::string(name) == std::string("node" + std::to_string(id_) + "-index")) {
          index_.concatenate(reinterpret_cast<PRIMITIVE*>(buffer));
        }
```

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -1503,6 +1575,17 @@ namespace awkward {
...
       }
 
+      /// @brief Copies and concatenates the accumulated data in the builder buffer to
+      /// a user-defined pointer if the given node name matches with the node associated
+      /// with the builder; otherwise, it searches the builder contents to locate a
+      /// matching node.
+      void
+      to_buffer(void* buffer, const char* name) const noexcept {
+        if (std::string(name) == std::string("node" + std::to_string(id_) + "-mask"))
+          mask_.concatenate(reinterpret_cast<int8_t*>(buffer));
```
jpivarski:
```suggestion
        if (std::string(name) == std::string("node" + std::to_string(id_) + "-mask")) {
          mask_.concatenate(reinterpret_cast<int8_t*>(buffer));
        }
```

jpivarski (approved):
I can see how this will make it much more convenient to fill buffers from a C interface, which has to pass each buffer individually. This implementation puts the recursion in the LayoutBuilder classes themselves, rather than outside of them.

Is our C/C++ tidy formatter removing the curly brackets from single-statement `if`/`for` bodies? I'm surprised it's not enforcing them. According to https://stackoverflow.com/q/26111162/1623645, it's either named `readability-braces-around-statements` in clang-tidy or `InsertBraces` in clang-format.

ianna (changes requested):
@ManasviGoyal - good progress, thanks! 

Let's combine these helper functions to fill one map to avoid a potential miss-match between the map key and its value.

Here is how ChatGPT suggests the `std::map` handling :-)  

please, write a Python code and a C function using ctypes that wraps C++ to pass std::map to python

Certainly! To wrap a C++ function using `ctypes` that passes a `std::map` to Python, you'll need to create a C interface that can be accessed from Python. Here's an example:

1. **C++ Code** (`example.cpp`):

```cpp
#include <iostream>
#include <map>

extern "C" {
    typedef std::map<int, int> IntMap;

    IntMap* create_map() {
        return new IntMap();
    }

    void insert_to_map(IntMap* map, int key, int value) {
        (*map)[key] = value;
    }

    void print_map(IntMap* map) {
        for (const auto& entry : *map) {
            std::cout << entry.first << ": " << entry.second << std::endl;
        }
    }

    void delete_map(IntMap* map) {
        delete map;
    }
}
```

2. **C++ Compilation**:

```bash
g++ -shared -o example.so -fPIC example.cpp
```

This will create a shared library `example.so` that can be loaded by `ctypes` in Python.

3. **Python Code** (`example.py`):

```python
import ctypes

# Load the shared library
example = ctypes.CDLL('./example.so')

# Define types for C functions
example.create_map.restype = ctypes.c_void_p
example.create_map.argtypes = []

example.insert_to_map.argtypes = [ctypes.c_void_p, ctypes.c_int, ctypes.c_int]

example.print_map.argtypes = [ctypes.c_void_p]

example.delete_map.argtypes = [ctypes.c_void_p]

# Create a map
map_ptr = example.create_map()

# Insert some values
example.insert_to_map(map_ptr, 1, 10)
example.insert_to_map(map_ptr, 2, 20)
example.insert_to_map(map_ptr, 3, 30)

# Print the map
example.print_map(map_ptr)

# Clean up memory
example.delete_map(map_ptr)
```

4. **Run the Python Code**:

```bash
python example.py
```

This code will create a `std::map` in C++, insert some values, print the map, and then clean up the memory.

Please make sure to adjust the compilation and library loading paths as needed. Additionally, this example assumes a Linux-like environment; if you're on a different OS, the compilation and library loading process may be slightly different.

ianna:
> I can see how this will make it much more convenient to fill buffers from a C interface, which has to pass each buffer individually. This implementation puts the recursion in the LayoutBuilder classes themselves, rather than outside of them.
> 
> Is our C/C++ tidy formatter removing the curly brackets from single-statement `if`/`for` bodies? I'm surprised it's not enforcing them. According to https://stackoverflow.com/q/26111162/1623645, it's either named `readability-braces-around-statements` in clang-tidy or `InsertBraces` in clang-format.

I think the clang-tidy is reinforcing it :-) 
Since we do not have a `.clang-tidy` configuration in the project it uses the following settings:
```
---
Checks:          'clang-diagnostic-*,clang-analyzer-*'
WarningsAsErrors: ''
HeaderFilterRegex: ''
AnalyzeTemporaryDtors: false
FormatStyle:     none
User:            yana
CheckOptions:
  llvm-else-after-return.WarnOnConditionVariables: 'false'
  modernize-loop-convert.MinConfidence: reasonable
  modernize-replace-auto-ptr.IncludeStyle: llvm
  cert-str34-c.DiagnoseSignedUnsignedCharComparisons: 'false'
  google-readability-namespace-comments.ShortNamespaceLines: '10'
  cert-err33-c.CheckedFunctions: '::aligned_alloc;::asctime_s;::at_quick_exit;::atexit;::bsearch;::bsearch_s;::btowc;::c16rtomb;::c32rtomb;::calloc;::clock;::cnd_broadcast;::cnd_init;::cnd_s
ignal;::cnd_timedwait;::cnd_wait;::ctime_s;::fclose;::fflush;::fgetc;::fgetpos;::fgets;::fgetwc;::fopen;::fopen_s;::fprintf;::fprintf_s;::fputc;::fputs;::fputwc;::fputws;::fread;::freopen;::
freopen_s;::fscanf;::fscanf_s;::fseek;::fsetpos;::ftell;::fwprintf;::fwprintf_s;::fwrite;::fwscanf;::fwscanf_s;::getc;::getchar;::getenv;::getenv_s;::gets_s;::getwc;::getwchar;::gmtime;::gmt
ime_s;::localtime;::localtime_s;::malloc;::mbrtoc16;::mbrtoc32;::mbsrtowcs;::mbsrtowcs_s;::mbstowcs;::mbstowcs_s;::memchr;::mktime;::mtx_init;::mtx_lock;::mtx_timedlock;::mtx_trylock;::mtx_u
nlock;::printf_s;::putc;::putwc;::raise;::realloc;::remove;::rename;::scanf;::scanf_s;::setlocale;::setvbuf;::signal;::snprintf;::snprintf_s;::sprintf;::sprintf_s;::sscanf;::sscanf_s;::strch
r;::strerror_s;::strftime;::strpbrk;::strrchr;::strstr;::strtod;::strtof;::strtoimax;::strtok;::strtok_s;::strtol;::strtold;::strtoll;::strtoul;::strtoull;::strtoumax;::strxfrm;::swprintf;::
swprintf_s;::swscanf;::swscanf_s;::thrd_create;::thrd_detach;::thrd_join;::thrd_sleep;::time;::timespec_get;::tmpfile;::tmpfile_s;::tmpnam;::tmpnam_s;::tss_create;::tss_get;::tss_set;::unget
c;::ungetwc;::vfprintf;::vfprintf_s;::vfscanf;::vfscanf_s;::vfwprintf;::vfwprintf_s;::vfwscanf;::vfwscanf_s;::vprintf_s;::vscanf;::vscanf_s;::vsnprintf;::vsnprintf_s;::vsprintf;::vsprintf_s;
::vsscanf;::vsscanf_s;::vswprintf;::vswprintf_s;::vswscanf;::vswscanf_s;::vwprintf_s;::vwscanf;::vwscanf_s;::wcrtomb;::wcschr;::wcsftime;::wcspbrk;::wcsrchr;::wcsrtombs;::wcsrtombs_s;::wcsst
r;::wcstod;::wcstof;::wcstoimax;::wcstok;::wcstok_s;::wcstol;::wcstold;::wcstoll;::wcstombs;::wcstombs_s;::wcstoul;::wcstoull;::wcstoumax;::wcsxfrm;::wctob;::wctrans;::wctype;::wmemchr;::wpr
intf_s;::wscanf;::wscanf_s;'
  cert-oop54-cpp.WarnOnlyIfThisHasSuspiciousField: 'false'
  cert-dcl16-c.NewSuffixes: 'L;LL;LU;LLU'
  google-readability-braces-around-statements.ShortStatementLines: '1'
  cppcoreguidelines-non-private-member-variables-in-classes.IgnoreClassesWithAllMemberVariablesBeingPublic: 'true'
  google-readability-namespace-comments.SpacesBeforeComments: '2'
  modernize-loop-convert.MaxCopySize: '16'
  modernize-pass-by-value.IncludeStyle: llvm
  modernize-use-nullptr.NullMacros: 'NULL'
  llvm-qualified-auto.AddConstToQualified: 'false'
  modernize-loop-convert.NamingStyle: CamelCase
  llvm-else-after-return.WarnOnUnfixable: 'false'
  google-readability-function-size.StatementThreshold: '800'
...
```

jpivarski:
Not to be down on ChatGPT, but I prefer Manasvi's solution. Instead of creating an temporary object that gets filled and later has to be deleted, Manasvi's solution goes directly to the thing we ultimately want to do: get the buffers to the right depth of the LayoutBuilder tree so that they can be filled.

I'll admit that there are performance benefits to making the temporary `std::map`. In Manasvi's solution, every name is compared with the name of every node in the tree, and those are string comparisons, whereas with a `std::map<std::string, void*>`, there's only one tree-traversal for all buffers, and keys of the map are first checked by comparing hashes before string comparisons. At the absolute worst, there can be thousands of nodes in a tree, so we're comparing $\mathcal{O}(n^2)$ string comparisons with $\mathcal{O}(n)$ hashmap lookups, and $n \approx 1000$. A million string comparisons sounds bad, but I'll bet it's not any more than a few milliseconds, and it's only done once per file/Awkward Array. Since this is not a speed-critical part of the code, I would favor a million string comparisons over having to delete a temporary `std::map`.

But if we go with the temporary `std::map` (and all of that code would make more sense in the Kaitai-Awkward runtime, rather than here in Awkward Array), then the Python needs to ensure that the map pointer gets deleted, even if there are errors.

```python
# Create a map
map_ptr = example.create_map()

try:
    # Insert some values
    example.insert_to_map(map_ptr, 1, 10)
    example.insert_to_map(map_ptr, 2, 20)
    example.insert_to_map(map_ptr, 3, 30)

    # There's no reason to ever print the map

finally:
    # Clean up memory
    example.delete_map(map_ptr)
```

That `try`-`finally` logic applies to the LayoutBuilder's opaque pointer, too, but it would be a different `try`-`finally` because the `try` block has to encompass all of the code after the pointer is constructed and before it is deleted, and the LayoutBuilder and `std::map` would be constructed at different times. Even if they're constructed one after the other, the two `try` blocks would be nested because one is on a line immediately following the other.

(This is the kind of complexity I wanted to avoid by not having intermediate objects.)

jpivarski:
We have both a `.clang-tidy` and a `.clang-format` in [awkward-cpp](https://github.com/scikit-hep/awkward/tree/main/awkward-cpp), right? Do they apply to the header-only code?

ianna:
> We have both a `.clang-tidy` and a `.clang-format` in [awkward-cpp](https://github.com/scikit-hep/awkward/tree/main/awkward-cpp), right? Do they apply to the header-only code?

No, I don't think so

ianna:
> Not to be down on ChatGPT, but I prefer Manasvi's solution. Instead of creating an temporary object that gets filled and later has to be deleted, Manasvi's solution goes directly to the thing we ultimately want to do: get the buffers to the right depth of the LayoutBuilder tree so that they can be filled.
> 
> I'll admit that there are performance benefits to making the temporary `std::map`. In Manasvi's solution, every name is compared with the name of every node in the tree, and those are string comparisons, whereas with a `std::map<std::string, void*>`, there's only one tree-traversal for all buffers, and keys of the map are first checked by comparing hashes before string comparisons. At the absolute worst, there can be thousands of nodes in a tree, so we're comparing O(n2) string comparisons with O(n) hashmap lookups, and n≈1000. A million string comparisons sounds bad, but I'll bet it's not any more than a few milliseconds, and it's only done once per file/Awkward Array. Since this is not a speed-critical part of the code, I would favor a million string comparisons over having to delete a temporary `std::map`.
> 
> But if we go with the temporary `std::map` (and all of that code would make more sense in the Kaitai-Awkward runtime, rather than here in Awkward Array), then the Python needs to ensure that the map pointer gets deleted, even if there are errors.
> 
> ```python
> # Create a map
> map_ptr = example.create_map()
> 
> try:
>     # Insert some values
>     example.insert_to_map(map_ptr, 1, 10)
>     example.insert_to_map(map_ptr, 2, 20)
>     example.insert_to_map(map_ptr, 3, 30)
> 
>     # There's no reason to ever print the map
> 
> finally:
>     # Clean up memory
>     example.delete_map(map_ptr)
> ```
> 
> That `try`-`finally` logic applies to the LayoutBuilder's opaque pointer, too, but it would be a different `try`-`finally` because the `try` block has to encompass all of the code after the pointer is constructed and before it is deleted, and the LayoutBuilder and `std::map` would be constructed at different times. Even if they're constructed one after the other, the two `try` blocks would be nested because one is on a line immediately following the other.
> 
> (This is the kind of complexity I wanted to avoid by not having intermediate objects.)

I'm not worried about the performance, but rather a potential issues when a user loads multiple files, gets the sizes first for all files, then gets the names, etc. The results could be out of order. The helpers should be hidden.

And, yes, I think the anonymous helpers would make more sense in an auto-generated part, or outside of the LayoutBuilder header.

ManasviGoyal:
> And, yes, I think the anonymous helpers would make more sense in an auto-generated part, or outside of the LayoutBuilder header.
>
Yes, you are right. The anonymous helpers don't need to be in the `LayoutBuilder` header as they serve no purpose there. I don't think they will be needed anywhere else later besides the kaitai code. So, they can be moved to the generated code.

ManasviGoyal:
> Is our C/C++ tidy formatter removing the curly brackets from single-statement `if`/`for` bodies? I'm surprised it's not enforcing them. According to https://stackoverflow.com/q/26111162/1623645, it's either named `readability-braces-around-statements` in clang-tidy or `InsertBraces` in clang-format.

It's a bit uneven. There are few places where it is enforced while in other places it's not.

jpivarski:
> I'm not worried about the performance, but rather a potential issues when a user loads multiple files, gets the sizes first for all files, then gets the names, etc. The results could be out of order. The helpers should be hidden.

It's already the case that a single LayoutBuilder instance can't be used with multiple files. It's a stateful object that is created when a file is opened, goes through all the sequential steps of snapshot-generation when it's done reading, and then is deleted when the file is closed. That's true in both its C++ and its C interface—even in the C++ interface, the snapshot process involved multiple steps that have to go in order.

> And, yes, I think the anonymous helpers would make more sense in an auto-generated part, or outside of the LayoutBuilder header.

I would expect to find `std::map` generation in the Kaitai-Awkward runtime (not generated, but also not in the Awkward header-only code).

But Manasvi's solution of passing each buffer and name down through the LayoutBuilder tree is most easily expressed in the Awkward header-only code, so if you're going with that implementation, I agree with Manasvi's original idea that it should be here.

I'm in favor of merging this PR if all of the curly brackets are put in.

ManasviGoyal:
> I would expect to find `std::map` generation in the Kaitai-Awkward runtime (not generated, but also not in the Awkward header-only code).
> 
> But Manasvi's solution of passing each buffer and name down through the LayoutBuilder tree is most easily expressed in the Awkward header-only code, so if you're going with that implementation, I agree with Manasvi's original idea that it should be here.
> 
> I'm in favor of merging this PR if all of the curly brackets are put in.

I have removed the helper functions for now and fixed the curly braces (there were more in the code). Maybe we can this PR open and discuss further about this in tomorrow. I do have a few questions regarding what should go the the `kaitai-awkward-runtime` and what in the generated code.

Maybe we can have something like `utils.h` in the `kaitai-awkward-runtime`  for such code if we decide to put  it there.

ianna (approved):
@ManasviGoyal - great! thanks!

ianna:
@jpivarski - I'm happy with the PR. Please, feel free to merge it anytime. Thanks!

jpivarski (approved):
In the end, this just adds a

```c++
to_buffer(void* buffer, const char* name)
```

interface to LayoutBuilder, which is a useful thing to have.

PR #2769, opened 2023-10-24
fix: support all-`None` index in `awkward_Index_nones_as_index.cpp`

### LINKED ISSUE #2768: `IndexError` for `ak.argsort` on arrays with `None`
### Version of Awkward Array

2.4.6

It seems that `ak.argsort` does not work on arrays that are entirely `None`, though `ak.sort` works fine. I would have naively expected that if `sort` works on an array, `argsort` would also work on the array, so it is not clear to me if this behavior would be expected or not. Here is an example that reproduces the behavior: 
```
import awkward as ak
  
a = ak.Array([None,None,[None,None]])
print(ak.sort(a,axis=1))
print(ak.argsort(a,axis=1))
```  
The `ak.sort` line is ok (and prints `[None, None, [None, None]]`), while the `ak.argsort` line fails with the error below: 
```
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/contents/numpyarray.py", line 301, in _getitem_at
    out = self._data[where]
          ~~~~~~~~~~^^^^^^^
IndexError: index 2 is out of bounds for axis 0 with size 2

  File "/home/k.mohrman/coffea_dir/ewkcoffea/analysis/wwz/mre2.py", line 6, in <module>
    print(ak.argsort(a,axis=1))
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/highlevel.py", line 1233, in __str__
    return awkward._prettyprint.valuestr(self, 1, 80)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/_prettyprint.py", line 240, in valuestr
    _, strs = valuestr_horiz(data, limit_cols)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/_prettyprint.py", line 135, in valuestr_horiz
    cols_taken, strs = valuestr_horiz(current, limit_cols - 2)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/_prettyprint.py", line 117, in valuestr_horiz
    current = get_at(data, index)
              ^^^^^^^^^^^^^^^^^^^
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/_prettyprint.py", line 43, in get_at
    out = data._layout._getitem_at(index)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/contents/indexedoptionarray.py", line 308, in _getitem_at
    return self._content._getitem_at(self._index[where])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/k.mohrman/miniconda3/envs/ak-env/lib/python3.12/site-packages/awkward/contents/numpyarray.py", line 303, in _getitem_at
    raise ak._errors.index_error(self, where, str(err)) from err
IndexError: cannot slice NumpyArray (of length 2) with 2: index 2 is out of bounds for axis 0 with size 2
```

agoose77:
This PR fixes #2768 by adding support for the all-none case to our `awkward_Index_nones_as_index.cpp` kernel. It also ensures that if we're argsorting the final dimension, we drop the option type from that dimension i.e. for this argsort result `[10, 20, None, 5] → [3, 0, 1, 4]` the array doesn't contain any `None`.

I think we should probably refactor this logic, but right now that's not the highest priority, so I've just patched

### DIFF
--- a/awkward-cpp/src/cpu-kernels/awkward_Index_nones_as_index.cpp
+++ b/awkward-cpp/src/cpu-kernels/awkward_Index_nones_as_index.cpp
@@ -8,12 +8,17 @@ template <typename T>
 ERROR awkward_Index_nones_as_index(
   T* toindex,
   int64_t length) {
-  int64_t last_index = 0;
+  int64_t n_non_null = 0;
+  // Assuming that `toindex` comprises of contiguous integers, and is zero-based
+  // Compute the number of non-null values to determine our starting index
   for (int64_t i = 0; i < length; i++) {
-    toindex[i] > last_index ? last_index = toindex[i] : last_index;
+    if (toindex[i] != -1) {
+        n_non_null++;
+    }
   }
+  //
   for (int64_t i = 0; i < length; i++) {
-    toindex[i] == -1 ? toindex[i] = ++last_index : toindex[i];
+    toindex[i] == -1 ? toindex[i] = n_non_null++ : toindex[i];
   }
   return success();
 }
--- a/awkward-cpp/src/cpu-kernels/awkward_Index_nones_as_index.cpp
+++ b/awkward-cpp/src/cpu-kernels/awkward_Index_nones_as_index.cpp
@@ -9,14 +9,14 @@ ERROR awkward_Index_nones_as_index(
   T* toindex,
   int64_t length) {
   int64_t n_non_null = 0;
-  // Assuming that `toindex` comprises of contiguous integers, and is zero-based
+  // Assuming that `toindex` comprises of unique, contiguous integers (or -1), and is zero-based
   // Compute the number of non-null values to determine our starting index
   for (int64_t i = 0; i < length; i++) {
     if (toindex[i] != -1) {
         n_non_null++;
     }
   }
-  //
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
Patching is good. The definition should also be updated here: https://github.com/scikit-hep/awkward/blob/8927adbe99b514f7ffd85c4980740a8a608e5b11/kernel-specification.yml#L4501-L4510

agoose77:
> The definition should also be updated here: 

Eek, nearly forgot that. Thanks.

PR #2770, opened 2023-10-25
fix: handle unhashable behaviour type

### LINKED ISSUE #2762: Numba's `njit` doesn't support non-default behaviors
### Version of Awkward Array

main

```python
import numba
import awkward as ak

SOME_ATTRS = {"FOO": "BAR"}
builder = ak.ArrayBuilder(behavior=SOME_ATTRS)


@numba.njit
def func(array):
    return array


assert func(builder).behavior is SOME_ATTRS

```

ianna:
fixes #2762


jpivarski (commented):
This was already dealt with in

https://github.com/scikit-hep/awkward/blob/17aabdecdbbbc7c9eb6fae6e834ab51e607386f7/src/awkward/_connect/numba/arrayview.py#L343-L354

Perhaps ArrayBuilder could use the same mechanism as ArrayView?

Also, @agoose77 was going to send an example of a `behavior` that can't be serialized due to a lambda function. I hope we don't have to pull in cloudpickle just for that. But maybe Numba's serialization already does something clever: we should try it and find out.

agoose77:
Here's a non-pickleable array:
```python
import pickle

import numpy as np
import numba as nb
import pytest

import awkward as ak


def make_add_xyr():
    def add_xyr(left, right):
        x = left.x + right.x
        y = left.y + right.y
        return ak.zip(
            {
                "x": x,
                "y": y,
                "r": np.sqrt(x**2 + y**2),
            },
            with_name="xyr",
        )

    return add_xyr


behavior = {(np.add, "xyr", "xyr"): make_add_xyr()}


a = ak.Array([{"x": 3, "y": 4, "r": 5}], with_name="xyr", behavior=behavior)
b = ak.Array([{"x": 3, "y": 4, "r": 5}], with_name="xyr", behavior=behavior)

builder = ak.ArrayBuilder(behavior=behavior)


with pytest.raises(AttributeError):
    behavior_pkl = pickle.dumps(behavior)


@nb.njit
def make_ab(builder):
    builder.begin_record("xyz")
    builder.field("x").integer(3)
    builder.field("y").integer(4)
    builder.field("z").integer(3)
    builder.end_record()

    builder.begin_record("xyz")
    builder.field("x").integer(3)
    builder.field("y").integer(4)
    builder.field("z").integer(3)
    builder.end_record()
    return builder


result = make_ab(builder).snapshot()

print(result)

```

We discussed in the meeting perhaps stashing a reference in the Python layer, and retrieving it later on? Just in case we forgot this (though I doubt we did).

src/awkward/_connect/numba/builder.py
```diff
@@ -74,8 +74,16 @@ def box_ArrayBuilder(arraybuildertype, arraybuilderval, c):
...
+    serializable2dict_obj = c.pyapi.unserialize(
+        c.pyapi.serialize_object(ak._connect.numba.arrayview.serializable2dict)
+    )
+    behavior2_obj = c.pyapi.unserialize(
+        c.pyapi.serialize_object(
+            ak._connect.numba.arrayview.dict2serializable(arraybuildertype.behavior)
+        )
+    )
+    behavior_obj = c.pyapi.call_function_objargs(
+        serializable2dict_obj, (behavior2_obj,)
```
ianna, line 86: @jpivarski - Thanks for the reminder! This is indeed a correct way to serialize a dict :-)

tests/test_2770_serialize_and_deserialize_behaviour_for_numba.py
```diff
@@ -0,0 +1,40 @@
...
+                },
+                with_name="xyr",
+            )
+
+        return add_xyr
+
+    behavior = {(np.add, "xyr", "xyr"): make_add_xyr()}
+
+    builder = ak.ArrayBuilder(behavior=behavior)
+    assert func(builder).behavior is behavior
```
ianna, line 40: @agoose77 - Thanks for the test! We are back to `is`!

ianna (commented):
@agoose77 - I'm done with the PR. Please, check and modify if needed. For example, I put both tests in one - it could be split. Thanks!

ianna:
> @agoose77 - I'm done with the PR. Please, check and modify if needed. For example, I put both tests in one - it could be split. Thanks!

Nope, not quite finished with it :-(

tests/test_2770_serialize_and_deserialize_behaviour_for_numba.py
```diff
@@ -0,0 +1,97 @@
...
+    # Compare the unique identifiers of the lambda functions
+    # Note: Lambda functions have different identities
+    lambda_behavior_out = next(iter(behavior_out.values()))
+    lambda_behavior = next(iter(behavior.values()))
+
+    if id(lambda_behavior_out) == id(lambda_behavior):
+        print("Lambda functions have the same identity")
+    else:
+        print("Lambda functions have different identities")
+
```
ianna: @agoose77 - I'm not sure how to compare that the behaviour passed through is the same. Here are some checks.

tests/test_2770_serialize_and_deserialize_behaviour_for_numba.py
```diff
@@ -0,0 +1,97 @@
...
+        builder.field("x").integer(3)
+        builder.field("y").integer(4)
+        builder.field("z").integer(3)
+        builder.end_record()
+        return builder
+
+    result = make_ab(builder).snapshot()
+
+    print(result)
+    assert result.behavior == make_ab(builder).behavior
```
ianna: @agoose77 - perhaps, this check is sufficient?

ianna (commented):
@jpivarski - I see that the lambda functions have different identity:
```
{(<ufunc 'add'>, 'xyr', 'xyr'): <function test_ArrayBuilder_non_picklable_behavior.<locals>.make_add_xyr.<locals>.add_xyr at 0x12395de40>}
{(<ufunc 'add'>, 'xyr', 'xyr'): <function test_ArrayBuilder_non_picklable_behavior.<locals>.make_add_xyr.<locals>.add_xyr at 0x1242d25c0>}
```

tests/test_2770_serialize_and_deserialize_behaviour_for_numba.py
agoose77: I'll push to your branch :)

agoose77:
I've loosened the test to just ensure that the `.keys()` agree. The other test ensures that the behavior with non-serialised lambdas is equal, which I think is sufficient.

PR #2771, opened 2023-10-25
feat: add helper methods to `utils.h`

ManasviGoyal:
I have added the helper functions to `utils.h` as we discussed.

I am not sure about the comments we want to add regarding not recommending it.

### DIFF
--- a/header-only/layout-builder/awkward/utils.h
+++ b/header-only/layout-builder/awkward/utils.h
@@ -270,6 +270,47 @@ namespace awkward {
     visit_impl<sizeof...(CONTENTs)>::visit(contents, index, fun);
   }
 
+  /// @brief Helper function to retrieve the names of the buffers.
+  ///
+  /// Only to to used with C interface. Not recommended as it can cause a mismatch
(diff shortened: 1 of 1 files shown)


header-only/layout-builder/awkward/utils.h
```diff
@@ -270,6 +271,47 @@ namespace awkward {
     visit_impl<sizeof...(CONTENTs)>::visit(contents, index, fun);
   }
 
+  /// @brief Helper function to retrieve the names of the buffers.
+  ///
+  /// Only to to used with C interface. Not recommended as it can cause a mismatch
```
ianna: I'd say: "Note: use with caution, beware of a potential mismatch between retrieved values!"

ianna (approved):
@ManasviGoyal - thanks, looks good! As we have discussed, it is a good place for the helper functions - the details of the comments do not matter that much. You could say "for expert use" or something alike.
It's good to go!

ianna:
@jpivarski - I'm merging it so that we could test a build procedure from the master branch.

PR #2774, opened 2023-10-25
fix: add `truediv` support for `unknown_length`

douglasdavis:
Comes from working on https://github.com/dask-contrib/dask-awkward/issues/396, I found that unknown_length objects cannot be `truediv`ed!

`Scalar / Scalar` operations in dask-awkward are unoptimizable because the optimization only graph eventually hits `unkown_length / unknown_length` if the `Scalar` comes from a `dak.num(..., axis=0)`

### DIFF
--- a/src/awkward/_nplikes/shape.py
+++ b/src/awkward/_nplikes/shape.py
@@ -37,6 +37,15 @@ class _UnknownLength(PrivateSingleton):
     __rmul__ = __mul__
     __imul__ = __mul__
 
+    def __truediv__(self, other) -> Self | NotImplemented:
+        if isinstance(other, int) or other is self:
+            return self
+        else:
+            return NotImplemented
+
+    __rtruediv__ = __truediv__
+    __itruediv__ = __truediv__
+
     def __floordiv__(self, other) -> Self | NotImplemented:
         if isinstance(other, int) or other is self:
             return self


agoose77:
@jpivarski would benefit from your thoughts here — my suggestion is that `ak.num` should return the "index" type rather than `ShapeItem` type. This means we'd convert the `UnknownLength` to a `TypeTracerArray(shape=(), dtype=np.int64)`. Do you think that's appropriate? i.e. at the "high level", we get array objects, whilst `Content.length` is still an `unknown_length`.

jpivarski:
Do you mean for `axis=0`? (Because for any other `axis`, the return value is an `ak.Array`.)

The return type on real data is a Python `int`, but that should probably be an `np.int64` instead:

```python
>>> type(ak.num(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]), axis=0))
<class 'int'>   # should be <class 'numpy.int64'>
```

Then, like a scalar returned by reducers with `axis=None`, it would stay within the system of Awkward Arrays, Records, and nplike zero-dimensional arrays:

```python
>>> type(ak.count(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]), axis=None))
<class 'numpy.int64'>
```

which can be important for CuPy.

That aligns exactly with what you're asking for (or, I'm just describing it in a different way):

```python
>>> ak.count(ak.to_backend(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]), "typetracer"), axis=None)
TypeTracerArray(dtype('int64'), shape=())
```

agoose77:
Closed in favour of #2785

PR #2775, opened 2023-10-26
fix: drop lengths recursively

### LINKED ISSUE #2764: Typetracer can't slice with `(int, str, int)`

agoose77:
As discussed in https://github.com/scikit-hep/awkward/pull/2765#issuecomment-1780771826, it doesn't make sense for us to retain interior (buffer) lengths when the outer length is not known. If we think about layouts as being built in a `from_buffers` style, we *must* know parent lengths to know child lengths. 

Fixes #2764

### DIFF
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -311,7 +311,7 @@ class BitMaskedArray(Content):
         tt = TypeTracer.instance()
         return BitMaskedArray(
             self._mask.to_nplike(tt),
-            self._content._to_typetracer(False),
+            self._content._to_typetracer(forget_length),
             self._valid_when,
             unknown_length if forget_length else self.length,
             self._lsb_order,
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -234,7 +234,7 @@ class ByteMaskedArray(Content):
         mask = self._mask.to_nplike(tt)
         return ByteMaskedArray(
             mask.forget_length() if forget_length else mask,
-            self._content._to_typetracer(False),
+            self._content._to_typetracer(forget_length),
             self._valid_when,
             parameters=self._parameters,
         )
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -224,7 +224,7 @@ class IndexedArray(Content):
         index = self._index.to_nplike(TypeTracer.instance())
         return IndexedArray(
             index.forget_length() if forget_length else index,
-            self._content._to_typetracer(False),
+            self._content._to_typetracer(forget_length),
             parameters=self._parameters,
         )
 
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -213,7 +213,7 @@ class IndexedOptionArray(Content):
         index = self._index.to_nplike(TypeTracer.instance())
         return IndexedOptionArray(
             index.forget_length() if forget_length else index,
-            self._content._to_typetracer(False),
+            self._content._to_typetracer(forget_length),
             parameters=self._parameters,
         )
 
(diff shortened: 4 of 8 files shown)


jpivarski (approved):
Yep, we'll try it and see what happens. 👉🙄👈

Could this one PR, alone, be put in a pre-release separately from all other PRs? Like, after releasing version 2.4.7, merge this into main and release version 2.5.0rc1. That way, if there are any issues in downstream analysis code, we'd be able to test this change in isolation by swapping releases (instead of building manually).

grst:
My pre-release tests in scirpy are failing with awkward 2.5.0rc0: https://github.com/scverse/scirpy/actions/runs/6722925605/job/18271885510

I still need to investigate in more detail what's going on, but maybe you can hold off releasing this until then. 

The offending line appears to be
```python
ak.argsort(ak.fill_none(airr[k][idx], default), stable=True, axis=-1, ascending=False)
```
```pytb
 E           ValueError: cannot argsort an irreducible UnionArray
E           
E           This error occurred while calling
E           
E               ak.argsort(
E                   <Array [['CAGGGSGTYKYIF'], ...] type='5 * var * union[string, 0 * c...'>
E                   stable = True
E                   axis = -1
E                   ascending = False
E               )

/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/site-packages/awkward/_errors.py:82: ValueError
```

agoose77:
@grst that was actually introduced in our now-yanked version 2.4.7. This PR (https://github.com/scikit-hep/awkward/pull/2793) should fix it. Thanks for testing our pre release!

PR #2777, opened 2023-10-26
chore: move to Ruff-format

henryiii:
This moves to Ruff-format. Most changes are better (and the one that's worse is being fixed, see https://github.com/astral-sh/ruff/issues/8226).

I've also removed a few checks that were redundant (pyupgrade) or slow and not very helpful / probably redundant (check debug statements).

An uncached run of black took 4.63s. With caching it's 0.4s. Ruff-format doesn't have caching yet AFAIK, but it takes 0.06 seconds.

Dropping pyupgrade saves 2 more seconds. Debug statements saves another 0.5 seconds.

Total time is now under 4 seconds for me. It used to be 6.5-10 seconds.

See https://github.com/scientific-python/cookie/pull/299.

### DIFF
--- a/src/awkward/_typing.py
+++ b/src/awkward/_typing.py
@@ -53,7 +53,7 @@ else:
     )
 
 
-JSONSerializable: TypeAlias = (
-    "str | int | float | bool | None | list | tuple | JSONMapping"
-)
+JSONSerializable: (
+    TypeAlias
+) = "str | int | float | bool | None | list | tuple | JSONMapping"
 JSONMapping: TypeAlias = "dict[str, JSONSerializable]"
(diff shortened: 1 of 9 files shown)


src/awkward/_typing.py
```diff
@@ -53,7 +53,7 @@
     )
 
 
-JSONSerializable: TypeAlias = (
-    "str | int | float | bool | None | list | tuple | JSONMapping"
-)
+JSONSerializable: (
+    TypeAlias
+) = "str | int | float | bool | None | list | tuple | JSONMapping"
```
henryiii: FYI, next release of Ruff should fix this, and it sounds like it's coming out pretty soon (maybe todayish), so we could wait.

PR #2779, opened 2023-10-27
chore: remove `ak.to_categorical`

agoose77:
This follows our planned removal in 2.5.0


jpivarski (approved):
Yes, it's [on the list](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20), so it's time for it to go!

PR #2783, opened 2023-10-28
ci: fix oldest-supported-dependencies test workflow

### LINKED ISSUE #2784: NumPy 1.18 is not properly supported
### Version of Awkward Array

main

NumPy 1.18 is very old, so few people were actually testing this.

We were not aware of the problem until we introduced features that silently depended upon "NumPy>=1.24", which is much newer; downstream projects (coffea) tested against `1.18 < numpy < 1.24` and raised the alarm.

agoose77:
It turns out that our CI wasn't properly testing the package combinations that we'd asked it to — there were subsequent install steps that overwrote the intended variations. This means that #2767 was not properly tested against older versions of NumPy, leading to a bug.

Although it's possible to use multiple install steps, it best reflects our intentions if we use a single install command.

In future, we probably just want to have a single `requirements.txt` file, and a separate `constraints.txt` that pins these dependency versions on our lower-bound configuration. However, we don't currently test *all* dependencies at the lower bound. This PR only runs core tests (NumPy + PyArrow) in the 3.8 tests.

Fixes #2784

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -80,20 +80,15 @@ jobs:
           python -m pip install build
           python -m build -w awkward-cpp
 
-      - name: Install awkward-cpp
-        run: python -m pip install -v @(get-childitem -path awkward-cpp/dist/*.whl)
-
-      - name: Build & install awkward
-        run: python -m pip install -v .
+      - name: Install awkward, awkward-cpp, and dependencies
+        run: >-
+          python -m pip install --only-binary "numpy,pandas,pyarrow,numexpr,numexpr"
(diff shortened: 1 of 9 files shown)


agoose77:
414f69a should break the CI, such that 6c708c6 fixes it!

jpivarski (approved):
This PR fixes both the test infrastructure and also the bug for old versions of NumPy. I strongly approve of the use of a `NUMPY_HAS_NEP_50` instead of finding a solution that works for both old and new NumPy, so that we know what to remove when our minimum NumPy does get to be 1.24 or above.

I think this is ready to merge!

PR #2785, opened 2023-10-30
fix: `ak.num` should always return a useful (non-unknown length) type

agoose77:
Replaces #2774 by ensuring the return type of `ak.num` is always a valid "index type" (something that kernels produce).

It might be that this is an integer rather than an `np.ndarray` (or `cp.ndarray`, etc); this is up to the nplike. Right now, nplikes don't promote Python integers to 0D arrays.


tests/test_2785_ak_num_typetracer_axis_0.py
```diff
@@ -0,0 +1,15 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+
+import awkward as ak
+from awkward._nplikes.typetracer import TypeTracerArray
+
+
```
jpivarski, line 7: For consistency, the `axis=0` return value should be a `np.int64`, like `ak.count` and `ak.count_nonzero`. Then the TypeTracer is specifying the same type that the non-TypeTracer array returns.

```suggestion
import numpy as np
import awkward as ak
from awkward._nplikes.typetracer import TypeTracerArray


def test_nontypetracer_length():
    array = ak.Array([0.0, 1.1, 2.2, 3.3])
    assert isinstance(ak.num(array, axis=0), np.integer)


def test_analogy_with_count_nonzero():
    array = ak.Array([0.0, 1.1, 2.2, 3.3])
    assert isinstance(ak.count_nonzero(array, axis=None), np.integer)
    assert ak.count_nonzero(array, axis=None).dtype == np.dtype(np.int64)


```

jpivarski (approved):
Just one fix, and then it's ready to merge (i.e. no need for a second review).

tests/test_2785_ak_num_typetracer_axis_0.py
agoose77: I relaxed my view on this since the meeting - permitting `int | ArrayLike` as return values. If we do that, then `ak.num` doesn't need to allocate the scalar on the GPU for e.g. CuPy. That's not super concerning; it's already needing to do a round-trip to pull from the GPU.

A separate point to note; typetracer returns 0D arrays (as does CuPy), whilst NumPy kernels return `np.generic` objects. As such, I'm feeling as though we don't need to worry too much about returning actual int types. Do you agree?

jpivarski: If the TypeTracer says that it's a zero-dimensional array, not a Python `int`, then the non-TypeTracer should honor that promise. Also, there's the consistency with reducers and `__getitem__`, which all return NumPy numbers, rather than Python numbers.

Objects with `np.generic` type are zero-dimensional arrays.

The fact that there's a distinction between NumPy numbers and Python numbers has caused problems in the past. Now there's a nice way to test them generically with `numbers.Integral` and `numbers.Real`, but `isinstance(·, int)` can still be a surprise. We can't do anything about the fact that there's a perhaps-unwanted distinction between NumPy numbers and Python numbers, but at least we can be consistent in always returning one or always returning the other.

`ak.num` returned `int` because it was one of the very early implementations (maybe the first). This `axis=0` case was forgotten about. It would be good to get this into line with the rest, which switched from Python numbers to NumPy numbers in the Awkward v1 → v2 transition.

agoose77: > Objects with np.generic type are zero-dimensional arrays.

At a fundamental level, these are different types. But, it sounds like you consider the important distinction to be Python integer vs non-Python "int-like", which includes both 0D arrays and (generic) scalars? 

I was hoping for a `shape_item_as_index` implementation because that mirrors the reverse direction; mapping from index values (int, np.int64, unknown-scalar, etc) to shape items (int, unknown-length). Nevertheless, we can easily add a secondary `int → np.int64` translation if we want!

agoose77:
@jpivarski to confirm — forcibly returning `ArrayLike` values from `ak.num` means that it's not possible to use `ak.num` to return concrete lengths for typetracers with known shapes. I assume you're OK with that, but I want to surface before we merge.

jpivarski:
> @jpivarski to confirm — forcibly returning `ArrayLike` values from `ak.num` means that it's not possible to use `ak.num` to return concrete lengths for typetracers with known shapes. I assume you're OK with that, but I want to surface before we merge.

I don't see anything wrong with that—internally in the codebase, we should be using `.length` and users have `len` if they want to make a distinction between getting the length eagerly at DAG-building time (if the TypeTracer has a known shape) and getting the length as delayed data until it's taken up by a Dask worker.

In fact, if I had to make that distinction, I would expect the Awkward function to give me delayed data and the Python function to give me eager data, because Awkward knows about Dask and Python doesn't.

I guess we're losing a case of "give me the eager, numerical length if you have it and a delayed token if you don't." But if I wanted that case, I wouldn't have guessed that `ak.num` would provide it (and it would be a tricky thing to explain in the documentation—likely undiscoverable, anyway). If I needed this case, I'd write a `try`-`except` around `len`.

PR #2789, opened 2023-10-31
chore: run mypy in CI

agoose77:
This PR enables `mypy` checks on a subset of the codebase and runs it as part of CI.
Right now, we don't publicly declare type-hints via `py.typed`, so these are only used internally.

Some things don't type super well; the `Sentinel` singleton doesn't seem very easy to handle. The closest we could get I *think* is to use e.g. 
```python
import enum
from typing import Literal


class Unset(enum.Enum):
    TOKEN = enum.auto()


def func_with_defaults(x: int, y: int | Literal[Unset.TOKEN]):
    ...
```

But for now, we'll just ignore this part of the typing.

We're not running in strict mode, so this only checks places where we declare types.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -66,3 +66,10 @@ repos:
   hooks:
     - id: check-github-workflows
       args: ["--verbose"]
+
+- repo: https://github.com/pre-commit/mirrors-mypy
+  rev: v1.6.1
+  hooks:
+    - id: mypy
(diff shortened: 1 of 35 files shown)


agoose77:
[65fc760](https://github.com/scikit-hep/awkward/pull/2789/commits/65fc760b2529b4a5749041c2d9d0e45628010158) fixes an existing bug, but a new assertion caught it.

PR #2792, opened 2023-11-01
fix: don't preserve unexpected option for is_in

agoose77:
PyArrow 14 started returning an option type, which follows from [the new `null_matching_behavior` option](https://github.com/apache/arrow/pull/36739).

@jpivarski I'm wondering about the `_drop_option_preserving_form` helper function that we use to erase options. Should we only check the bitmasks being erased in our test suite? i.e., at production, we don't assert that each entry is 0, but rather assume that the operation doesn't do this — the mask is just PyArrow adding a mask node?

### DIFF
--- a/src/awkward/operations/str/__init__.py
+++ b/src/awkward/operations/str/__init__.py
@@ -84,7 +84,7 @@ from awkward.operations.str.akstr_trim_whitespace import *
 from awkward.operations.str.akstr_upper import *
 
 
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
This is why Arrow's major version is increasing so rapidly: they change their API a lot!

Yes, this is good to merge, if it makes our string functions insensitive to this change in Arrow 14.

PR #2793, opened 2023-11-01
fix: support scalar strings in `fill_none`

agoose77:
This regression was introduced by #2763. This PR ensures that strings are treated as scalars in `fill_none`.

### DIFF
--- a/src/awkward/operations/ak_fill_none.py
+++ b/src/awkward/operations/ak_fill_none.py
@@ -89,7 +89,8 @@ def _impl(array, value, axis, highlevel, behavior):
     elif isinstance(valuelayout, ak.contents.Content):
         valuelayout = valuelayout[np.newaxis, ...]
     else:
-        # Now we know we've got Python scalars/strings, we can promote without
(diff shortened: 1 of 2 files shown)


src/awkward/operations/ak_fill_none.py
```diff
@@ -81,19 +81,23 @@ def _impl(array, value, axis, highlevel, behavior):
...
         primitive_policy="pass-through",
+        string_policy="pass-through",
     )
 
     if isinstance(valuelayout, ak.record.Record):
         valuelayout = valuelayout.array[valuelayout.at : valuelayout.at + 1]
     elif isinstance(valuelayout, ak.contents.Content):
         valuelayout = valuelayout[np.newaxis, ...]
     else:
+        # Now we know we've got Python scalars/strings, we can promote without
```
agoose77:
```suggestion
        # Now that we know `valuelayout` isn't a low-level type, we must have scalars
        # Thus, we can safely promote these scalars to a layout without
```

jpivarski (approved):
Thanks for fixing this! It can be merged, which unblocks those releases.

PR #2795, opened 2023-11-03
refactor: introduce parametrised nplike types

agoose77:
The existing type hints for `nplike` were watered down to support a single `ArrayLike` protocol. This prevents us from imposing the constraint that a CuPy array can only be handled by a CuPy nplike. 

This PR restores the original parametrised-nplike protocol, so that it's possible to declare an nplike as operating upon fixed array types. Unfortunately, mypy doesn't support higher-kinded-types, which are required to compose generic types. This means that we can't e.g. have 
```python
class NumpyArray(Generic[Backend]):
    @property
    def backend(self) -> Backend:
        ...
```
and so on. There are higher-kinded-type works related to mypy, e.g. https://sobolevn.me/2020/10/higher-kinded-types-in-python, but they don't look like a good fit for our needs; they have some drawbacks and impose constraints on the runtime logic.

So, this PR won't have huge sweeping impact, but it will allow us to explicitly type an nplike as a typetracer nplike, e.g.

```python
def is_typetracer_nplike(nplike: NumpyLike) -> TypeGuard[TypeTracer]:
    return not nplike.known_data


if is_typetracer_nplike(nplike):
    print(nplike.zeros(4).report)
```

such that we can use fewer assertions.


agoose77:
This is really just typing, so we should be safe to merge.

PR #2799, opened 2023-11-06
fix: support arbitrary ufuncs with respect to NEP-50

agoose77:
This PR takes into consideration the failure of Numba ufuncs in our NEP-50 aware machinery. It's unfortunate that this has taken a subsequent PR, but in retrospect we should have duck-typed this from the beginning.

Now, we just ask the ufunc object if it supports NEP-50 via `resolve_dtypes`. In future, we should gracefully support Numba's implementation, assuming that it implements this in a similar way to NumPy, which it ought to.

Our test suite will fail once Numba >=0.59.0 is available, because we can't know for certain what kind of exception it will throw. Thus, we'll catch that at the time and implement the correct exception handler in pytest.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -4,9 +4,6 @@ from __future__ import annotations
 import math
 from functools import lru_cache
 
-import numpy
-import packaging.version
-
 from awkward._nplikes.numpy_like import (
     ArrayLike,
     IndexType,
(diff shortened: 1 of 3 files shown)


agoose77:
Hmm, test fails on Windows. Let me perhaps choose some different types to avoid the 32-bit aspect to this. I'll use a typed `np.generic` scalar ~`np.int16`

PR #2801, opened 2023-11-06
chore: drop `isort` config, require annotation imports

agoose77:
This PR updates our linting rules to always require annotation module imports. This was inspired by the default in https://github.com/scientific-python/cookie

We also update the tests. I would probably prefer to allow the tests to skip this import, but it doesn't seem possible to do this at the moment, so let's just make life easy and impose the same rules for all source files.


agoose77:
@henryiii have you come across a mechanism for scoping `isort.required-imports` to particular files?

henryiii:
You can disable I101 for subsets of files. Sadly, that drops everything, not just the isort addition. There will be a way to do it with PEP 723 if that gets implemented per file. You can also put a `.ruff.toml` in a subdirectory, I think.

jpivarski (approved):
If we're going to be touching all of these files, anyway, let's clean up some long-standing format issues.

* GitHub website (dropping the "-1.0")
* uniform placement of the LICENSE line
* uniform vertical spacing in the first few lines
* uniform placement of `__all__` relative to imports.

Since the new `from __future__ import annotations` goes first, I put the `__all__` after all imports. There's usually a few more assignments in that position, so it fits in, but it's always separated from them by at least a linefeed.

This includes touching the C++ code, since those files had inconsistent LICENSE messages, too. If you want to avoid triggering a new awkward-cpp right now, we can revert the last commit on the whole awkward-cpp directory.

Otherwise, I think this is ready to be merged.

agoose77:
Although I'm loathed to touch the C++ again, we really want to avoid these things living for a long time. As such, I'm just going to merge! Thanks all!

PR #2805, opened 2023-11-07
refactor: drop use of `behavior` in `recursively_apply`

agoose77:
This prepares the way to drop this argument in `recursively_apply`

### DIFF
--- a/src/awkward/_do.py
+++ b/src/awkward/_do.py
@@ -35,7 +35,6 @@ def recursively_apply(
     if isinstance(layout, Content):
         return layout._recursively_apply(
             action,
-            behavior,
             1,
             copy.copy(depth_context),
             lateral_context,
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -33,8 +33,9 @@ from awkward.contents.bytemaskedarray import ByteMaskedArray
 from awkward.contents.content import (
     ApplyActionOptions,
     Content,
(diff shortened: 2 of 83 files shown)


jpivarski (approved):
When string-broadcasting was hard-coded, no longer derived from a `behavior`, we no longer needed to pass behaviors into `broadcast_and_apply`, and I suppose it was being passed into `recursively_apply` for symmetry.

If the tests pass after simply removing it like this, then I suppose nothing was using it. I would have expected it to be passed to the `action` function, but apparently it's not.

Actually, why are these changes only in `ak.operations.str.*`? We use `recursively_apply` in a lot of places, not just here. Why don't they all have `behavior=behavior` push-downs to remove? Was this never really a part of the `recursively_apply` interface and their use in string functions was a mistake?

Anyway, I'm in favor of removing `behavior` from `recursively_apply`. It doesn't even mean we can't use it in `recursively_apply`'s `action` functions, since the `action` function can be defined in such a way as to close over that variable. So, please do make the end-state of this to be such that `behavior` is removed from the `recursively_apply` interface, whether that involves changes beyond `ak.operations.str.*` or not.

agoose77:
> Actually, why are these changes only in ak.operations.str.*? We use recursively_apply in a lot of places, not just here. Why don't they all have behavior=behavior push-downs to remove? Was this never really a part of the recursively_apply interface and their use in string functions was a mistake?

I was originally going to just start with `ak.str`, and follow up, as this PR was cherry-picked from another. However, let's just do it all together.

PR #2806, opened 2023-11-08
feat: expose attrs in typetracer

agoose77:
This PR exposes the new `attrs` mapping argument in the `ak.typetracer` public L2 API.

### DIFF
--- a/src/awkward/typetracer.py
+++ b/src/awkward/typetracer.py
@@ -2,12 +2,13 @@
 
 from __future__ import annotations
 
+from collections.abc import Callable, Mapping
+
 import awkward.forms
 from awkward._backends.typetracer import TypeTracerBackend
-from awkward._behavior import behavior_of
 from awkward._do import touch_data as _touch_data
 from awkward._errors import deprecate
-from awkward._layout import wrap_layout
+from awkward._layout import HighLevelContext, wrap_layout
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Sounds good!

jpivarski:
Adding a `HighLevelContext` to `touch_data` and `_length_0_1_if_typetracer` causes `to_layout` to be called on the array, which falls back to `to_iter`, which fails because it's a typetracer array. These functions _only_ make sense if the argument is already a typetracer array, so this effectively breaks them. It's responsible for https://github.com/scikit-hep/awkward/issues/3077#issuecomment-2050043714 (which is an issue about a regression, and the last version they checked is, indeed, before this PR).

The reason you added `HighLevelContext` here was to propagate `attrs`. Is there a way to propagate `attrs` without something that eventually goes to `to_layout`? I could do it the pedestrian way of just checking for the things that need to be propagated, `attrs` and `behaviors`, and propagating them. I would be in favor of that for these exceptional functions—low-level functions that need to propagate high-level state—but if there's a utility function for it already, I ought to use that.

Is there a utility function for what I'm trying to do? If not, I'll happily do it the pedestrian way.

jpivarski:
This conversation is continuing at https://github.com/scikit-hep/awkward/issues/3077#issuecomment-2050068263.

PR #2807, opened 2023-11-08
feat!: drop `forget_length` in `typetracer_with_report`

agoose77:
This PR follows the deprecation schedule to remove this argument.


agoose77:
As this is a deprecation PR, we've already agreed upon it's contents, so I'm happy to merge without Jim's input.

PR #2813, opened 2023-11-10
chore: only build _ext and kernels

agoose77:
Right now, we install three libraries: `_ext`, `libawkward`, `libawkward-cpu-kernels`, and build some additional static variants.

We didn't change this when splitting out awkward-cpp, because I didn't want to rock the boat any more. Since becoming more well versed with CMake's nuances, I'm confident that we can remove the leftover baggage. 

This PR keeps the separate `libawkward` and `_ext` libraries (because it doesn't appear trivial to merge them), but drops the static libraries.

### DIFF
--- a/awkward-cpp/CMakeLists.txt
+++ b/awkward-cpp/CMakeLists.txt
@@ -25,15 +25,11 @@ Please check https://github.com/scikit-hep/awkward#installation-for-developers t
 endif()
 
 # Setup the RPATH for built libraries
-if(CMAKE_SYSTEM_NAME STREQUAL Emscripten)
-  message(STATUS "Using Emscripten")
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
-  set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
-elseif(APPLE)
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
+if(APPLE)
   set(CMAKE_INSTALL_RPATH "@loader_path")
 else()
-  set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
+  set(CMAKE_INSTALL_RPATH "\$ORIGIN")
 endif()
 
 # Three tiers: [cpu-kernels (extern "C" interface), cuda-kernels (extern "C" interface)],
--- a/awkward-cpp/CMakeLists.txt
+++ b/awkward-cpp/CMakeLists.txt
@@ -54,31 +54,21 @@ target_include_directories(awkward-parent INTERFACE rapidjson/include)
 add_subdirectory(header-only EXCLUDE_FROM_ALL)
 target_link_libraries(awkward-parent INTERFACE awkward::growable-buffer)
 
-# First tier: cpu-kernels (object files, static library, and dynamic library).
-add_library(awkward-cpu-kernels-objects OBJECT ${CPU_KERNEL_SOURCES})
-set_target_properties(awkward-cpu-kernels-objects PROPERTIES POSITION_INDEPENDENT_CODE ON)
-target_link_libraries(awkward-cpu-kernels-objects PUBLIC awkward-parent)
+# First tier: cpu-kernels
+add_library(awkward-cpu-kernels SHARED ${CPU_KERNEL_SOURCES})
+target_link_libraries(awkward-cpu-kernels PUBLIC awkward-parent)
 set_target_properties(
-  awkward-cpu-kernels-objects
+  awkward-cpu-kernels
   PROPERTIES CXX_VISIBILITY_PRESET hidden
              VISIBILITY_INLINES_HIDDEN ON
              CXX_EXTENSIONS NO)
 
-add_library(awkward-cpu-kernels-static STATIC $<TARGET_OBJECTS:awkward-cpu-kernels-objects>)
-set_property(TARGET awkward-cpu-kernels-static PROPERTY POSITION_INDEPENDENT_CODE ON)
-target_link_libraries(awkward-cpu-kernels-static PUBLIC awkward-parent)
-
-add_library(awkward-cpu-kernels SHARED $<TARGET_OBJECTS:awkward-cpu-kernels-objects>)
-target_link_libraries(awkward-cpu-kernels PUBLIC awkward-parent)
-
-# Second tier: libawkward (object files, static library, and dynamic library).
-add_library(awkward-objects OBJECT ${LIBAWKWARD_SOURCES})
-set_target_properties(awkward-objects PROPERTIES POSITION_INDEPENDENT_CODE 1)
-target_compile_definitions(awkward-objects PRIVATE LIBAWKWARD_EXPORT_SYMBOL=EXPORT_SYMBOL)
+# Second tier: libawkward
+add_library(awkward SHARED ${LIBAWKWARD_SOURCES})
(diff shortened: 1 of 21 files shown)


awkward-cpp/CMakeLists.txt
```diff
@@ -54,31 +55,24 @@ target_include_directories(awkward-parent INTERFACE rapidjson/include)
...
-add_library(awkward-cpu-kernels SHARED $<TARGET_OBJECTS:awkward-cpu-kernels-objects>)
-target_link_libraries(awkward-cpu-kernels PUBLIC awkward-parent)
+# Second tier: python
+find_package(Python REQUIRED COMPONENTS Interpreter Development)
+find_package(pybind11 CONFIG REQUIRED)
 
-# Second tier: libawkward (object files, static library, and dynamic library).
-add_library(awkward-objects OBJECT ${LIBAWKWARD_SOURCES})
-set_target_properties(awkward-objects PROPERTIES POSITION_INDEPENDENT_CODE 1)
-target_compile_definitions(awkward-objects PRIVATE LIBAWKWARD_EXPORT_SYMBOL=EXPORT_SYMBOL)
```
agoose77, line 77: I don't think we need to deal with `LIBAWKWARD_EXPORT_SYMBOL` any more; that was a solution to a problem with exporting template functions when linking against awkward.

awkward-cpp/CMakeLists.txt
```diff
@@ -25,15 +25,11 @@ Please check https://github.com/scikit-hep/awkward#installation-for-developers t
...
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
-  set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
-elseif(APPLE)
-  set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
+set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
+if(APPLE)
   set(CMAKE_INSTALL_RPATH "@loader_path")
 else()
-  set(CMAKE_BUILD_RPATH_USE_ORIGIN TRUE)
+  set(CMAKE_INSTALL_RPATH "\$ORIGIN")
```
agoose77, line 32: Previously I'd been under the impression that `CMAKE_BUILD_RPATH_USE_ORIGIN` _sets_ the `RPATH` to `$ORIGIN`. But, it appears that instead this variable just makes it possible to use relative paths with `$ORIGIN`. As such, we need to set `CMAKE_INSTALL_RPATH` in addition to enabling `$ORIGIN` on platforms for which it is supported.

agoose77: N.B. This was caught by auditwheel, which complained that `_ext`'s new link to `libawkward` couldn't be resolved. In the past, `_ext` was statically linked against awkward, so we didn't have to resolve the `RPATH`.

agoose77:
@jpivarski this PR fails on Windows. The error message in the CI indicates that the copy-assignment constructor for `GrowableBuffer` is being called by `std::vector`, but this constructor is implicitly deleted if one defines a move-assignment constructor. 

I'm not sure why it's suddenly an issue. It's possible that the CI toolchain bumped MSVC at the same time that I wrote this PR, but I doubt it; in this PR I also added an EXPORT_SYMBOL for the `FromJsonObjectSchema` class, which has `std::vector<GrowableBuffer<T>>` members. That seems like causative.

This raises the following questions for me:
1. Why does GCC not complain if MSVC does?
2. Why does the standard not imply that the `V` in `std::vector<V>` needs to be copy-assignable (and copy-constructible?)? It used to in C++11, but _seems_ to relax the rule for C++17+ (see "Template parameters" in https://en.cppreference.com/w/cpp/container/vector).
3. Is it reasonable to add a copy-assignment constructor to `GrowableBuffer`? It seems to me like we really don't want that.

I'll keep digging.

agoose77:
Well, that (05ee4a2) worked! I'm not entirely clear on why this became an error once I exported this class, but it seems that the emission of the default copy-assignment constructor leads to the attempted copy-assignment of the `GrowableBuffer`. 

Clearly, we don't want that, so [05ee4a2](https://github.com/scikit-hep/awkward/pull/2813/commits/05ee4a25151a224074e5335db4b2094475b06b47) deletes the top-level copy-assignment (rather than implementing it)

awkward-cpp/include/awkward/io/json.h
```diff
@@ -61,6 +61,12 @@ namespace awkward {
              int64_t initial,
              double resize);
 
+    // Delete copy constructor
+    FromJsonObjectSchema(const FromJsonObjectSchema&) = delete;
+
+    // Delete copy-assignment constructor
+    FromJsonObjectSchema& operator=(FromJsonObjectSchema&) = delete;
```
jpivarski, line 68: I'm going to betray some ignorance of modern C++.

This `FromJsonObjectSchema` object has a lot of private members, and many of them are `std::vectors`.

https://github.com/scikit-hep/awkward/blob/05ee4a25151a224074e5335db4b2094475b06b47/awkward-cpp/include/awkward/io/json.h#L353-L373

If we set the destructor to `= delete`, the object _does_ delete all of its private members when the `FromJsonObjectSchema` gets deleted, right? I want to be sure that this code change is not introducing a memory leak.

@ianna, do you know if this is the case?

jpivarski (approved):
Reducing the number of shared libraries to two (one is the Python module and the other is `dlopened` through `ctypes`) is ideal.

However, this PR doesn't change [awkward-cpp/src/awkward_cpp/cpu_kernels.py](https://github.com/scikit-hep/awkward/blob/main/awkward-cpp/src/awkward_cpp/cpu_kernels.py) to look for kernels in libawkward.so. It's odd that it passes tests: how does the kernel dispatch find the CPU kernels?

Are the tests picking up a cached version of awkward-cpp?

awkward-cpp/include/awkward/io/json.h
jpivarski: Yikes: https://stackoverflow.com/a/40746247/1623645

> Why mark a destructor as delete?
>
> To prevent the destructor from being invoked, of course ;)

I'm pretty sure that this use of `= delete` is causing a memory leak. It might be hard to notice because the depth of the objects created by `FromJsonObjectSchema` is limited by the depth of the JSON schema, though every time a JSON file is read with `schema=...`, another one of these would be created.

agoose77: > Reducing the number of shared libraries to two (one is the Python module and the other is dlopened through ctypes) is ideal.

This PR no-longer does this. It seems like to merge the Python module and shared libraries is not trivial, so I just removed the static libraries instead, and thus changed the linking between `libawkward` and `_ext` to dynamic.

agoose77: > If we set the destructor to = delete, the object does delete all of its private members when the FromJsonObjectSchema gets deleted, right? I want to be sure that this code change is not introducing a memory leak.

I may have made a mistake, but this line 
```
    // Delete copy constructor
    FromJsonObjectSchema(const FromJsonObjectSchema&) = delete;

    // Delete copy-assignment constructor
    FromJsonObjectSchema& operator=(FromJsonObjectSchema&) = delete;
```

Just erases the copy (assignment) _constructors_ so that the default copy (assignment) constructor is not used.

jpivarski: Ah! I thought this was deleting the destructors. Deleting the copy constructors is fine. (I don't care if it would copy all of the data upon assignment—I doubt we ever assign a `FromJsonObjectSchema` instance, anyway.)

I must have hallucinated a tilde at the beginning of one of these lines. Since these aren't destructors, it's fine.

This PR can be merged!

PR #2815, opened 2023-11-11
fix: support scalar returns from `firsts`

agoose77:
I noticed that we expect `ak.firsts` to return scalar values, which was broken in a recent refactoring PR.


jpivarski (approved):
`axis=0` is a case I hadn't thought about, but I guess you're right that it should return a scalar (`np.int64`).

agoose77:
> `axis=0` is a case I hadn't thought about, but I guess you're right that it should return a scalar (`np.int64`).

You and me both!

PR #2816, opened 2023-11-13
feat: export more of typetracer

agoose77:
Now that `dask-awkward` is stabilising, it's clear that we need to export some more of our internals. In this PR, I add a constructor for unknown scalars, but not for arrays. This is a very restrictive approach, but I think we want to keep things as locked down as possible until we have a clearer sense of what's needed besides unknown scalars.


jpivarski (approved):
Great! It's ready to merge.

PR #2818, opened 2023-11-14
feat: julia helper functions

ianna:
Julia helper functions to check the environment, versions, and initialize it.

The `AwkwardArray.jl` backend is loaded as follows. Note, if you don't have a Python shared library, you need to pass `compiled_modules=False` to Julia:
```python
In [2]: M = init_julia(julia_kwargs={"compiled_modules": False})

In [3]: A = _load_backend(M)

In [4]: A
Out[4]: <module 'julia.AwkwardArray' (<julia.core.JuliaModuleLoader object at 0x111a2f730>)>

```

M is a Julia's `Main`. A is a `'julia.AwkwardArray'` module:

```python
In [8]: A.PrimitiveArray([1, 2, 3, 4, 5])
Out[8]: [1, 2, 3, 4, 5]

In [9]: jl_array = A.PrimitiveArray([1, 2, 3, 4, 5])

In [10]: jl_array
Out[10]: [1, 2, 3, 4, 5]

In [11]: type(jl_array)
Out[11]: list
```

TODO:
- [ ] Update CI to install PyJulia (probably extend the new cppyy job)


src/awkward/_connect/julia/utils.py
```diff
@@ -0,0 +1,257 @@
...
+        juliainfo = _load_juliainfo()
+    current_version = (
+        juliainfo.version_major,
+        juliainfo.version_minor,
+        juliainfo.version_patch,
+    )
+    return current_version >= version
+
+
+def _check_for_conflicting_libraries():  # pragma: no cover
```
ianna, line 51: I'm not sure if we want to check this. I've run into some inconsistencies between `PythonCall` and `PyCall`. It could be just my environment: the installation order or outdated versions.

src/awkward/_connect/julia/utils.py
```diff
@@ -0,0 +1,258 @@
...
+    else:
+        is_shared = False
+        processed_julia_project = Path(julia_project)
+    return processed_julia_project, is_shared
+
+
+def _import_error():
+    return """
+    Required dependencies are not installed or built.  Run the following command in your terminal:
+        python3 -m awkward install
```
ianna, line 78: `pip install`?

ianna (commented):
@agoose77 - please, have a look. Thanks!

src/awkward/_connect/julia/utils.py
agoose77: Is this error encountered if `pyjulia` is not properly installed? I found https://github.com/JuliaPy/pyjulia/blob/69b734f117e489904a07d20f46c8f96b56577779/src/julia/pytestplugin.py#L71-75 which suggests a message of
```
PyCall is not installed or built.  Run the following code in Python REPL:

    >>> import julia
    >>> julia.install()

See:
    https://pyjulia.readthedocs.io/en/latest/installation.html
```

src/awkward/_connect/julia/utils.py
agoose77: I think we should be happy to enforce a lower bound; with our "special" dependencies we can be stricter, and that makes our lives easier. I've simplified this a bit, so now 1.9.0 is definitely a lower bound.

src/awkward/_connect/julia/utils.py
ianna: No, it's to install Julia dependencies for awkward, e.g. a specific Julia project -`AwkwardArray.jl` or check if it is installed.

ianna (commented):
@agoose77 - I like it! Thanks!

src/awkward/_connect/julia/utils.py
```diff
@@ -158,12 +156,12 @@ def init_julia(julia_project=None, quiet=False, julia_kwargs=None, return_aux=Fa
             )
 
     if julia_initialized and julia_activated_env != processed_julia_project:
-        Main.eval("using Pkg")
+        julia_main.eval("using Pkg")
```
ianna, line 159: I don't know if it's better to use the same name in Python as we would in Julia. Would it make the code more consistent and readable if for people familiar with Julia conventions?

agoose77: I think that's a good point. Because this code is internal, I'm less worried about that than I am about following our internal style guidelines. Are you happy with that reasoning?

ianna: yes, thanks!

agoose77:
@ianna is this waiting on me? :)

ianna:
> @ianna is this waiting on me? :)

No. Thanks for asking.

When AwkwardArray.jl is released and registered I'll update this PR with a more Julia-like recipe:
```python
from julia import Pkg
Pkg.add("AwkwardArray")
```

ianna:
closing it in favour of #2896

ianna (commented):
@jpivarski - the PR was closed in favour of https://github.com/scikit-hep/awkward/pull/2896. The branch can be deleted.

PR #2819, opened 2023-11-15
fix: include `_kernel_signatures.py` in package

agoose77:
I'm creating a Docker container to run our GPU tests using systemd, with the intention that errors can be reported as GitHub issues.

In doing so, I noticed that we weren't including the CUDA kernel signatures in our built packages, so non-development installs weren't getting these. Another problem associated with not testing our installs :(

This PR *also* fixes an API breakage in the `traceback` module that is not super-clear from the docs — the base image I'm using from NVidia uses Python 3.8 which surfaced our misusage in the CUDA tests.


agoose77:
I fully tested this on a private repo, but a subtle configuration difference meant that when switching to `scikit-hep/awkward` for committing the service files to this branch, the service restarted on exit. This has lead to many (now closed) issues being filed here. Sorry! I don't have the necessary privileges to delete them, apparently.

jpivarski (approved):
I haven't personally tested this (I don't currently have Docker), but it doesn't affect mainline Awkward and can be fixed later if it needs to be generalized beyond the one computer where it was tested (@agoose77's).

The pyproject.toml is changed, but these are corrections (including `/src/awkward/_connect/cuda/_kernel_signatures.py`).

Go ahead and merge!

PR #2837, opened 2023-11-16
fix: ufunc handling of attrs

agoose77:
I noticed that ufuncs in dask-awkward were not propagating attrs. It turns out that the bug originates here — odd that our tests didn't catch this, but perhaps it was a trivial oversight.

### DIFF
--- a/tests/test_2837_ufunc_attrs_behavior.py
+++ b/tests/test_2837_ufunc_attrs_behavior.py
@@ -1,0 +1,28 @@
+from __future__ import annotations
+
+import awkward as ak
+
+
+def test():
+    x = ak.Array([1, 2, 3], behavior={"foo": "BAR"}, attrs={"hello": "world"})
+    y = ak.Array([8, 0, 9], behavior={"do": "re"}, attrs={"mi": "fa"})
+    z = x + y
+    assert z.attrs == {"hello": "world", "mi": "fa"}
+    assert z.behavior == {"foo": "BAR", "do": "re"}
+
+
+def test_unary():
+    x = ak.Array([1, 2, 3], behavior={"foo": "BAR"}, attrs={"hello": "world"})
+    y = -x
+    assert y.attrs is x.attrs
+    assert x.behavior is y.behavior
+
+
+def test_two_return():
+    x = ak.Array([1, 2, 3], behavior={"foo": "BAR"}, attrs={"hello": "world"})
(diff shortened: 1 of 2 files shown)


tests/test_2837_ufunc_attrs_behavior.py
```diff
@@ -0,0 +1,28 @@
...
+    z = x + y
+    assert z.attrs == {"hello": "world", "mi": "fa"}
+    assert z.behavior == {"foo": "BAR", "do": "re"}
+
+
+def test_unary():
+    x = ak.Array([1, 2, 3], behavior={"foo": "BAR"}, attrs={"hello": "world"})
+    y = -x
+    assert y.attrs is x.attrs
+    assert x.behavior is y.behavior
```
jpivarski, line 18:
```suggestion
    assert y.attrs == x.attrs
    assert x.behavior == y.behavior
```

tests/test_2837_ufunc_attrs_behavior.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_two_return():
+    x = ak.Array([1, 2, 3], behavior={"foo": "BAR"}, attrs={"hello": "world"})
+    y, y_ret = divmod(x, 2)
+    assert y.attrs is y_ret.attrs
+    assert y.attrs is x.attrs
+
+    assert y.behavior is y_ret.behavior
+    assert y.behavior is x.behavior
```
jpivarski, line 28:
```suggestion
    assert y.attrs == y_ret.attrs
    assert y.attrs == x.attrs

    assert y.behavior == y_ret.behavior
    assert y.behavior == x.behavior
```

jpivarski (approved):
Although it may be true that `attrs` and `behavior` are passed on by reference, do we really want to pin the tests to that detail?

Otherwise, looks great, and merge when you're satisfied with it.

tests/test_2837_ufunc_attrs_behavior.py
agoose77: I think for now we _do_ want to do this, so that we catch places that we're unnecessarily copying metadata. If we change this in future it means fixing up these tests, but at least it won't silently regress.

jpivarski: Okay, I get it, then.

PR #2841, opened 2023-11-20
refactor: split "meta" into `Meta`

agoose77:
@jpivarski and I have discussed this on-and-off for over a year now.

Presently, `ak.contents.Content` subclasses re-use code from `ak.forms.Form` subclasses using explicit descriptor lookups. As such, `ak.forms.Form` serves as the implementation for many metadata methods. As we look to grow the set of metadata-only methods, it would be nice to formalise this hierarchy. Benefits include:
- easier type checking
- lower overhead (right now, e.g. `Content.parameter` involves an additional attribute lookup and function call)

As it now seems possible that we'll want an `enforce_form` function, I would like for the ability to implement the `_mergeable_next` protocol at the "meta" level, and doing this properly feels more readable (and easier to reason about) to me; we're not fighting the type system by implementing this via inheritance.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -226,10 +226,11 @@ module = [
     'awkward._nplikes.*',
     'awkward._behavior.*',
     'awkward._backends.*',
-    'awkward.forms.*',
-    'awkward.types.*',
+    'awkward._meta.*',
     'awkward._errors',
     'awkward._dispatch',
+    'awkward.forms.*',
+    'awkward.types.*',
     'awkward.index',
 ]
 ignore_errors = false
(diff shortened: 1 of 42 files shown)


agoose77:
I don't think we can / should group `Type` under `Meta`; types form a distinct group to forms and layouts.

jpivarski (approved):
Yes, we have been talking about this for a long time.

Sure, let's go ahead with this, but only if it isn't disruptive. From a user's or downstream developer's perspective, the Form and Content classes should appear to be as they have been (i.e. a pure refactor).

And yes, Types don't fit into this scheme.

PR #2851, opened 2023-11-28
ci: fix path to header only tests

zonca:
I think the header only tests were not executing due to a change in the path

### DIFF
--- a/.github/workflows/header-only-test.yml
+++ b/.github/workflows/header-only-test.yml
@@ -32,7 +32,7 @@ jobs:
         import os
         import pathlib
         import subprocess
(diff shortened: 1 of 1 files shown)


zonca:
In a previous CI runs like https://github.com/scikit-hep/awkward/actions/runs/7010054914/job/19069748258

image

after the fix it prints the name of the tests https://github.com/zonca/awkward/actions/runs/7024250732/job/19112557365?pr=2:

image

ianna (approved):
@zonca - thanks for fixing it! The header-only tests run in CI.

agoose77:
Whoops! Good catch @zonca, thanks!

PR #2852, opened 2023-11-29
refactor: split `ak._do` into `meta` and `content`

agoose77:
This PR:
1. Splits `Content._mergeable_next` and `ak._do.mergeable` into `Meta._mergeable_next` and `ak._do.meta.mergeable`.
2. Adds `dtype` to `NumpyMeta`
3. Drops use of `Generic` in type hints for `Meta` classes

(3) is required because of the lack of higher-kinded types. This means that our type hints will only let us prove that `NumpyArray` is a `NumpyMeta` in some contexts, e.g. via `ak._meta.meta.is_numpy()`

### DIFF
--- a/src/awkward/__init__.py
+++ b/src/awkward/__init__.py
@@ -16,7 +16,8 @@ import awkward.types
 import awkward.forms
 
 # internal
-import awkward._do
+import awkward._do.content
+import awkward._do.meta
 import awkward._slicing
 import awkward._broadcasting
 import awkward._reducers
--- a/src/awkward/_connect/jax/trees.py
+++ b/src/awkward/_connect/jax/trees.py
@@ -9,6 +9,7 @@ from awkward import contents, highlevel, record
 from awkward._backends.backend import Backend
 from awkward._backends.jax import JaxBackend
 from awkward._behavior import behavior_of
+from awkward._do.content import recursively_apply
 from awkward._layout import wrap_layout
 from awkward._nplikes.jax import Jax
 from awkward._nplikes.numpy import Numpy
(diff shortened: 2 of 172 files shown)


jpivarski (dismissed):
I see that this splits the `ak._do` module into two submodules for `Content` and the superclasses of both `Content` classes and `Form` classes.

`Meta` replaces the TypeVar `T` in many places; I suppose that's because `T` used to represent a `Content` or a `Form`, which had no common superclass, and now it can all be OOP-style, rather than template-style.

The merge* functions have moved. That's fine.

On the whole, there's a lot of differences, but it seems to be just (1) moving things and (2) merging the `Content` and `Form` hierarchies. Does this PR _complete_ the merging of those hierarchies, or is there another after this?

Also, what will the hierarchy look like after merging is complete? Is it like this?

```mermaid
flowchart LR
    C(Content)
    F(Form)
    M(Meta)
    ListArray --> C
    NumpyArray --> C
    ListForm --> F
    NumpyForm --> F
    ListMeta --> M
    NumpyMeta --> M
    ListArray --> ListMeta
    NumpyArray --> NumpyMeta
    ListForm --> ListMeta
    NumpyForm --> NumpyMeta
```

agoose77:
> Meta replaces the TypeVar T in many places; I suppose that's because T used to represent a Content or a Form, which had no common superclass, and now it can all be OOP-style, rather than template-style.

The `T` was used when I was trying to preserve the idea that `ListOffsetArray.content` is a `Content`, not just a `Meta`. But this ends up needing higher kinded types, which mypy does not yet express.

agoose77:
> On the whole, there's a lot of differences, but it seems to be just (1) moving things and (2) merging the Content and Form hierarchies. Does this PR complete the merging of those hierarchies, or is there another after this?

Yes, I think so. We may be able to promote some Content functions to Meta, but on balance I think its complete.

PR #2854, opened 2023-11-29
docs: link to execute JupyterLite full screen

zonca:
I could not find a way to run JupyterLite at https://awkward-array.org/doc/main/getting-started/try-awkward-array.html#try-it full screen, if not it is too small to execute comfortably in my opinion.

Following https://jupyterlite-sphinx.readthedocs.io/en/latest/full.html, I think the easiest way is to provide a link to https://awkward-array.org/doc/main/lite/retro/notebooks/?path=what-is-an-awkward-array.ipynb

### DIFF
--- a/docs/getting-started/demo/what-is-an-awkward-array.ipynb
+++ b/docs/getting-started/demo/what-is-an-awkward-array.ipynb
@@ -2,6 +2,14 @@
  "cells": [
   {
    "cell_type": "markdown",
+   "metadata": {
+    "deletable": false,
+    "editable": false
+   },
+   "source": "<center>\n<button data-commandLinker-command=\"application:toggle-zen\" href=\"#\">Toggle fullscreen</i></button>\n</center>"
+  },
+  {
+   "cell_type": "markdown",
    "id": "53a4d322",
    "metadata": {
     "deletable": false,
--- a/docs/getting-started/demo/what-is-an-awkward-array.ipynb
+++ b/docs/getting-started/demo/what-is-an-awkward-array.ipynb
@@ -6,7 +6,7 @@
     "deletable": false,
     "editable": false
    },
-   "source": "<center>\n<button data-commandLinker-command=\"application:toggle-zen\" href=\"#\">Toggle fullscreen</i></button>\n</center>"
+   "source": "<center>\n<button data-commandLinker-command=\"application:toggle-zen\" href=\"#\">Toggle fullscreen</i> <i class=\"fa-solid fa-expand\"></button>\n</center>"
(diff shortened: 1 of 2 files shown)


jpivarski:
I'd be in favor of that. I've been wondering if we can eliminate the third column (table of contents for the current page) because it's irrelevant for this page. In general, the problem is that the whole JupyterLite environment is fitting within an area designed for text, but this page should be exceptional.

If you find a way to add a full-screen button, that's great! If you take another approach and make this page custom, maybe by defining a new HTML `layout` for it and having that `layout` apply only to this page somehow, that would be another way to do it. @agoose77 is the author and already-but-soon-to-be-even-more expert in JupyterBook, so he might have some suggestions and opinions.

It's valuable for the JupyterLite element to not (always) be the entire page, since a user/reader of documentation will want to navigate to other parts of the documentation.

This "Try it!" page differs from NumPy's (https://numpy.org/, scroll down) so that it can guide the user with an example. But that means that there's no place for generic exploration, as NumPy has. But as you can see, NumPy's has less JupyterLite decoration around the edges, at the expense of not having cells that interleave input and output.

agoose77:
Hi @zonca, thanks for opening this PR!

There _is_ a UI option to open the JupyterLab environment:
image

Or "zen-mode" fullscreen:
image


However, it's _not_ self-evident (I had to hunt around for this myself the other day). Another approach to making this full-screen is to use a commandlink button in the notebook that enters "zen mode":
```html
<center>
<button data-commandLinker-command="application:toggle-zen" href="#">Go fullscreen! <i class="fa-solid fa-expand"></i></button>
</center>
```

If you create a Markdown cell at the top of the notebook and run this, what do you think?
image

zonca:
I agree those are hard to find, I think the button at the top of the notebook is best,
I'll test and mark PR ready when it looks good.

zonca:
ok @agoose77 I named it "Toggle" and I removed the icon because the icon only works within sphinx, if someone downloads the notebook or does "Open JupyterLab", the icon breaks.

http://preview.awkward-array.org.s3-website.us-east-1.amazonaws.com/jupyterlite_fullscreen/getting-started/try-awkward-array.html

agoose77:
I don't think we need to worry about the icon breaking if someone downloads the notebook; the notebook as-is only works in pyodide due to the use of piplite. With that in mind, could we restore the icon? I think it helps confer the purpose of the button.

agoose77 (approved):

agoose77:
Thanks @zonca! This is a useful change :)

PR #2857, opened 2023-11-29
fix: only compare scalars in full_like

agoose77:
Currently we follow a ufunc branch in `full_like`, which can be avoided. This branch may fail for non-castable fill-value dtypes e.g. datetime64.


agoose77:
This is simple and blocking dask-awkward CI; I'll merge.

Issue #2853, opened 2023-11-29
Categoricals should merge subject to particular rules


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

- Categoricals should merge for strings, but error for non strings?
- We _can_ add an extension for merging non-string categoricals.

jpivarski:
This is going to be hard to understand in 12 months!

Here's what I remember of our conversation:

* merging `X` with `X` results in `X` (naturally)
* merging `categorical[X]` with `X` results in `X` (I think this is least surprise, and has the least-bad performance consequences: the second `X` could be a million distinct strings)
* merging `categorical[X]` with `categorical[X]` results in `categorical[X]`, but the look-up categories have to be combined into a single look-up, without duplicates, which changes the `indexes` of at least one of the two.

The third case can be handled for the special case in which `X = string` or `X = bytestring`, but not in general, not unless the user passes some way of determining whether two instances of `X` are equal, or a way of ensuring that a set of instances of `X` has no duplicates (a new `ak.behavior`?).

For now, the third case should just go to `NotImplementedError`, so that it's a reminder that something better needs to be done there (and maybe point to this issue). In the planned future, merging-strings should eventually be implemented. The general case is in the unplanned future—we'll only do it if someone encounters this need. (So there may be a `NotImplementedError` case in the code forever, and that's fine.)

agoose77:
Thanks @jpivarski — haven't got round to writing up my notes yet! Looks like you've saved me the challenge :)

agoose77:
Just to add some additional context: we discussed making support for merging categoricals be a behavior-overload. It would probably make sense for this to only be defined for `RecordArray`, meaning that awkward should implement support for merging categoricals of the various primitive types that we support.

PR #2860, opened 2023-11-30
feat: add `enforce_concatenated_form`

agoose77:
> [!TIP]
> This PR makes the following changes:
> 1. Merging categoricals is an error, see #2853 
> 2. `_mergemany` no longer enforces a backend
> 3. More bugs in merging are fixed
> 4. New `ak.operations.ak_concatenate.enforce_concatenated_form` for enforcing a concatenated form.

`dask-awkward` needs to be able to compute a concatenation operation _but_ without actually concatenating buffers. @jpivarski and I discussed the various ways we could go about this:

1. Concatenate some typetracer arrays and use `enforce_type` to produce partitions of the correct _type_.
2. Concatenate each partition with a length-zero array of the final type.
3. Concatenate some typetracer arrays and introduce a new `enforce_form` to produce partitions of the correct _form_.
4. Add a special path to `_mergemany` that doesn't touch buffers if they have known length-zero.

Each of these solutions has drawbacks. In particular, modifying existing code to satisfy two constraints (perform an optimal merge in convenient amounts of code, and have a length-zero path) is more difficult than having a dedicated function that only handles the "don't touch buffers" case. As such, this PR instead adds a new function that acts like `enforce_form` _but_ under the conditions that the `form` was build from `ak.concatenate((layout, ...), axis=0)` where `layout` is the content being enforced to `form`. This allows us to make many simplifying assumptions according to the implementation of `_mergemany`.


src/awkward/contents/content.py
```diff
@@ -762,41 +761,23 @@ def _merging_strategy(
...
-                x if isinstance(x.backend.nplike, TypeTracer) else x.to_typetracer()
-                for x in tail
-            ]
-
-        return (head, tail)
+
+        if any(x.backend.nplike.known_data for x in head + tail) and not all(
+            x.backend.nplike.known_data for x in head + tail
+        ):
+            raise RuntimeError
```
agoose77: This can eventually be removed, but we'll leave it in for safety for now. Callers should be enforcing to a common backend.

src/awkward/contents/indexedarray.py
```diff
@@ -517,32 +514,38 @@ def _merging_strategy(self, others):
         head = [self]
         tail = []
 
-        i = 0
-        while i < len(others):
-            other = others[i]
+        it_others = iter(others)
+        for other in it_others:
```
agoose77, line 518: These (and subsequent) lines just simplify the logic

src/awkward/contents/indexedarray.py
```diff
@@ -663,9 +666,24 @@ def _mergemany(self, others: Sequence[Content]) -> Content:
                 contentlength_so_far += array.length
                 length_so_far += array.length
 
+                # Categoricals may only survive if all contents are categorical
+                if (
+                    parameters is not None
+                    and parameters.get("__array__") == "categorical"
+                ):
+                    parameters = {**parameters}
+                    del parameters["__array__"]
```
agoose77, line 675: This handles the case of merging content X against `categorical[Y]`, whereby `Y` is merged against `X`. In this case, we need to recognise that the outer content (`categorical[Y | X]`) is no longer categorical.

src/awkward/contents/listarray.py
```diff
@@ -1143,6 +1145,11 @@ def _mergemany(self, others: Sequence[Content]) -> Content:
         length_so_far = 0
 
         for array in head:
+            # We need contiguous content, so let's just convert to RegularArray
+            # immediately.
+            if array.is_numpy:
+                array = array.to_RegularArray()
```
agoose77, line 1151: A missing case of `NumpyArray` against `ListArray`.

src/awkward/contents/unmaskedarray.py
```diff
@@ -127,7 +126,7 @@ def simplified(cls, content, *, parameters=None):
                 parameters=parameters_union(content._parameters, parameters),
             )
         elif content.is_indexed or content.is_option:
-            return content.copy(
+            return content.to_IndexedOptionArray64().copy(
```
agoose77, line 129: When simplifying `UnmaskedArray`, we want to convert the content to an `IndexedOptionArray`, and update its parameters with our own.

src/awkward/contents/content.py
jpivarski: This could be left in forever as an assertion. (I hope I correctly inverted the logic: `not (any(...) and not all(...))` → `not any(...) or all(...)`.)

```suggestion
        assert not any(x.backend.nplike.known_data for x in head + tail) or all(
            x.backend.nplike.known_data for x in head + tail
        )
```

I can't believe I was paranoid enough to do this, but here's proof:

```python
>>> def f1(a, b):
...     return not(any(a) and not all(b))
... 
>>> def f2(a, b):
...     return not any(a) or all(b)
... 
>>> def make_a():
...     for i in range(2**4):
...         yield [x != "0" for x in format(i, "#06b")[-4:]]
... 
>>> from itertools import product
>>> for a, b in product(make_a(), make_a()):
...     assert f1(a, b) == f2(a, b)
... 
```

src/awkward/contents/indexedarray.py
jpivarski: That's good.

src/awkward/contents/indexedoptionarray.py
```diff
@@ -650,35 +647,38 @@ def _merging_strategy(self, others):
...
+        if (
+            other.is_indexed
+            and other.parameter("__array__")
+            == self.parameter("__array__")
+            == "categorical"
+        ):
+            raise NotImplementedError(
+                "merging categorical arrays is currently not implemented. "
+                "Use `ak.enforce_type` to drop the categorical type and use general merging."
+            )
```
jpivarski, line 680: _(I moved this comment because this is the relevant part of the code.)_

I was just talking with @lobis, and he has a use-case for `merge(categorical[X], categorical[X])` coming up soon (~month timescale). I'll be thinking about how to do this.

In the very first case (Luis's), the two categorical arrays have the same contents and no reindexing is necessary. We could kick this can down the road by only raising `NotImplementedError` when the contents of `categorical[X]` and `categorical[X]` are different. (And we only need to consider the case of `X = string`, since we don't know how to check equality for other types.)

jpivarski (approved):
Nice! I noticed all of the typetracer tests that were added to tests/test_0449_merge_many_arrays_in_one_pass.py, too.

We talked about this in our meeting, and the code looks good to me, so this is ready to merge.

src/awkward/contents/content.py
agoose77: Ha, I think binary ops are a perfect example of where one person's readable is not anothers! I'm happy to make the change, and it looks correct.

src/awkward/contents/indexedoptionarray.py
agoose77: I suggest let's keep this as-is, but be thinking about next steps in anticipation for @lobis' work? Out-of-interest, which categoricals are we thinking about in that context? strings?

lobis: https://github.com/lobis/geant4-python-application

The goal is to make a python interface for a generic Geant4 application which supplies Geant4 data as awkward arrays.

The structure of this data would mirror Geant4's Event > Track > Step hierarchy (the user would recieve an awkward array of events). Some of these fields are categorical strings: `track.particle`, `track.step.process`. `track.step.volume`, etc.

How it's currently laid out (very preliminary, @jpivarski and I already discussed some improvements to the awkward code). each Geant4 thread would produce an awkward array that would then be concatenated to the final array (this concatenation most likely taking place in python code).

> In the very first case (Luis's), the two categorical arrays have the same contents and no reindexing is necessary. We could kick this can down the road by only raising NotImplementedError when the contents of categorical[X] and categorical[X] are different. (And we only need to consider the case of X = string, since we don't know how to check equality for other types.)

I'm not sure if I'm following @jpivarski here. The categorical arrays may not have the same elements (most of them would be the same): some events may contain a very unlikely process that is not present in another batch of events, an unlikely volume to hit, etc. Their indexing order could also be different as they could appear in different order.

jpivarski: I was thinking of the physics processes, which I assume are enums in C++. You're right that there are also volumes, whose names are runtime-determined. Those could have different indexes that would have to get merged.

lobis: > I was thinking of the physics processes, which I assume are enums in C++. You're right that there are also volumes, whose names are runtime-determined. Those could have different indexes that would have to get merged.

Physics process are also strings, there is no id-assignment (meaning they could get indexed differently). The only thing as far as I am aware that coul be considered an enum are the particles (PDG particle ID). In principle the full list of every categorical value should be available at the beginning of the simulation (list all available physics processes, particles, volumes, etc).

PR #2862, opened 2023-12-01
feat: add `Form.is_equal_to`

agoose77:
We already have `Type.is_equal_to` and `Content.is_equal_to`. This PR extends this to `Form`, and fixes some bugs in the existing implementations.

### DIFF
--- a/src/awkward/types/recordtype.py
+++ b/src/awkward/types/recordtype.py
@@ -4,7 +4,6 @@ from __future__ import annotations
 
 import json
 from collections.abc import Iterable, Mapping
-from itertools import permutations
 
 import awkward as ak
 import awkward._prettyprint
@@ -212,16 +211,11 @@ class RecordType(Type):
             if set(self._fields) != set(other._fields):
                 return False
 
-            self_contents = [self.content(f) for f in self._fields]
-            other_contents = [other.content(f) for f in other._fields]
-
-            return any(
-                all(
-                    this._is_equal_to(that, all_parameters)
-                    for this, that in zip(self_contents, contents)
-                )
-                for contents in permutations(other_contents)
+            return all(
+                content._is_equal_to(other.content(field), all_parameters)
+                for field, content in zip(self._fields, self._contents)
             )
+
         # Mixed
         else:
             return False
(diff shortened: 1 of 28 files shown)


src/awkward/types/recordtype.py
```diff
@@ -212,16 +211,11 @@ def _is_equal_to(self, other: Any, all_parameters: bool) -> bool:
...
-            return any(
-                all(
-                    this._is_equal_to(that, all_parameters)
-                    for this, that in zip(self_contents, contents)
-                )
-                for contents in permutations(other_contents)
+            return all(
+                content._is_equal_to(other.content(field), all_parameters)
+                for field, content in zip(self._fields, self._contents)
             )
```
agoose77, line 217: This is wrong!

jpivarski: I don't see what's wrong about it, but I also don't know the larger context. Is this a note for yourself?

jpivarski (approved):
Right, so `is_equal_to` is a named function because it has a lot of custom arguments for controlling what kind of equality the caller needs. (For instance, "Should parameters count? Which parameters? How about form keys?")

I see that `__eq__` is identical to `is_equal_to` and only defined at the level of the abstract class (`Form`). That's good.

I see that you've also consolidated some repeated logic in `Content._is_equal_to` into a helper function.

This all looks good to me! I'll let you decide when to merge it, since you're saying that something is wrong below:

src/awkward/types/recordtype.py
agoose77: The original logic permuted the contents, but as we have named fields (or positions), we should not do this. Otherwise, `{x: int64, y: complex128}` and `{x: complex128, y: int64}` would be considered equal!

I brought in permutations after working on `enforce_type` and my mind must have been full of rubbish!

jpivarski: Yes,

1. `{x: int64, y: complex128} == {x: complex128, y: int64}` is bad and
2. `{x: int64, y: complex128} == {y: complex128, x: int64}` is good.

Although it doesn't look to me like that's what the code above is doing. It's getting a correlated `field, content` pair from `self` and comparing this `content` with `other.content(field)`, which ought to be case (2) above. As long as some code before this also ensures that `set(self.fields) == set(other.fields)`, that is.

agoose77: Oh, I see where the confusion lies. I am observing that the previous code (that I wrote) that uses permutations is wrong. This PR fixes it. Sorry for the misdirect!

PR #2866, opened 2023-12-01
fix: `Array.__getitem__` should set `attrs` on result

agoose77:
The assignment of `attrs` to derived arrays was missing in highlevel. This fixes it, and adds more tests!


jpivarski (approved):
Right, I can see how those were missed while you were focusing on `ak.operations.*`. Good catch!

Ready to merge.

agoose77:
> Right, I can see how those were missed while you were focusing on `ak.operations.*`. Good catch!
> 
> Ready to merge.

That's _exactly_ what happened!

PR #2869, opened 2023-12-04
ci: simplify test workflow

agoose77:
I've been wanting to do this for a while, and seeing @jpivarski's CI in the integration test repository inspired me to actually do it.

This PR merges our Linux, Windows, and MacOS test jobs into a single branching job.


agoose77:
Hmm, looks like something's not linking on clang 14 between _ext and libawkward.

jpivarski (approved):
This is a nice clean-up! The three platforms (Windows, MacOS, and Linux) were originally three separate workflows because I had problems with exceptions for individual platforms becoming spaghetti code in the past, but these three have remained fairly similar, so there's an advantage to consolidating them into one sequence with a matrix.

Before I merge this, I'll need to change the set of required tests from old names to new names, and then this would need to be merged into every open PR. There are a lot of open PRs.

@ianna, @zonca, @ManasviGoyal, @zbilodea, @lobis, are you ready to either merge your open PRs or merge with `main` once this goes into `main`?

agoose77:
@jpivarski at one point it wasn't obvious to me that this would be worth doing. But, at this juncture our use of shell commands is simple enough that it works cross-platform!

agoose77:
@jpivarski as this is blocking some other PRs, are you comfortable with me merging it and then merging into https://github.com/scikit-hep/awkward/pull/2902 https://github.com/scikit-hep/awkward/pull/2880 https://github.com/scikit-hep/awkward/pull/2896 https://github.com/scikit-hep/awkward/pull/2898 https://github.com/scikit-hep/awkward/pull/2899 https://github.com/scikit-hep/awkward/pull/2902?

jpivarski:
I suppose so. Do you have permissions to change the set of required tests?

agoose77:
n.b. I rebased the merge commit on main to remove the change to the C++, which was a temporary attempt at fixing `macos-latest`

PR #2870, opened 2023-12-05
ci: compile header only tests in Debug mode

zonca:
The header only tests were compiled in `Release` mode,
therefore making all `assert` statements ineffective.

See for example https://github.com/zonca/awkward/pull/4
where I add an error in the unit test but tests pass anyway

### DIFF
--- a/.github/workflows/header-only-test.yml
+++ b/.github/workflows/header-only-test.yml
@@ -24,7 +24,7 @@ jobs:
 
     - name: Run CMake
       run: |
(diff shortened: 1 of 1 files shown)


zonca:
after this fix, the test where I added an error actually fails:
https://github.com/zonca/awkward/pull/5

ianna (approved):
@zonca - yes, indeed, the assertions are compiled out in a release build. Thanks for catching it!

agoose77:
Thanks @zonca - this was a good oversight to catch!

PR #2871, opened 2023-12-06
ci: deploy branch preview docs only on main repo

zonca:
Forked repositories normally do not have secrets correctly configured,
so the deploy preview of the docs always fail, making PRs on forks always failing.
Of course this won't work if you have cases where you have forked repos with secrets.

See
https://github.com/orgs/community/discussions/26409#discussioncomment-3251822

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -350,11 +350,11 @@ jobs:
         path: docs/_build/.jupyter_cache
 
   branch-preview:
+    # We can only deploy for PRs on host repo
+    if: github.repository_owner == 'scikit-hep'
     runs-on: ubuntu-22.04
     needs: [build-docs]
     name: Deploy Branch Preview
-    # We can only deploy for PRs on host repo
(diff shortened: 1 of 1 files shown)


zonca:
apologies just noticed there was already a `if` statement on that action,
however it does not seem to work:

https://github.com/zonca/awkward/pull/6

zonca:
hopefully this one instead is working:

https://github.com/zonca/awkward/pull/7

agoose77:
We need to think about this - we don't want to leak any secrets to forked PRs. When I looked last, I felt we couldn't use any of the mechanisms for opening permissions without compromising access keys. It might be somewhat viable, I'll take a look this week!

agoose77:
Thanks for this PR! I'm going to close it, because there are security concerns with deploying docs from third-party repositories; we don't have any way of ensuring that executable code run during the docs is well-behaved, or that the built assets that are deployed aren't e.g. embedding a cryptominer. 

The only way I think we can implement this securely (without requiring people to constantly be vigilant) is to implement a one-off execution for a fork. I don't think there's a convenient way to do that, unfortunately.

PR #2875, opened 2023-12-07
refactor: remove dead code in generate-tests.py

ManasviGoyal:


ianna (approved):
@ManasviGoyal - great! Indeed these functions are not used.

PR #2876, opened 2023-12-07
refactor: remove obsolete cuda_kernels

ManasviGoyal:
Removed the following `cuda_kernels` which had a corresponding `cpu-kernels` in `v1` but not in `v2`:

1. awkward_ByteMaskedArray_getitem_carry.cu
2. awkward_ByteMaskedArray_mask.cu
3. awkward_Index_to_Index64
4. awkward_IndexedArray_getitem_nextcarry_outindex_mask
5. awkward_IndexedArray_mask
6. awkward_ListOffsetArray_reduce_global_startstop_64
7. awkward_NumpyArray_contiguous_init
8. awkward_NumpyArray_contiguous_next
9. awkward_NumpyArray_fill_tobool
10. awkward_NumpyArray_getitem_boolean_nonzero
11. awkward_NumpyArray_getitem_next_array
12. awkward_NumpyArray_getitem_next_array_advanced
13. awkward_NumpyArray_getitem_next_at
14. awkward_NumpyArray_getitem_next_range
15. awkward_NumpyArray_getitem_next_range_advanced
16. awkward_carry_arange
17. awkward_combinations
18. awkward_content_reduce_zeroparents_64
19. awkward_index_carry_nocheck
20. awkward_regularize_arrayslice
21. awkward_zero_mask

This one was in none of the versions:

22. awkward_index_carry


ManasviGoyal:
@jpivarski @ianna 
These are two more that had `cpu-kernels` in `v1` and don't have one in `v2`

- awkward_reduce_argmax_bool_64
- awkward_reduce_argmin_bool_64

I see that they have a different if condition. But since their `cpu-kernels` were dropped, are they still needed?

jpivarski (approved):
This is good to do as a first step, so that you don't accidentally spend time on obsolete kennels. The number of removed kennels doesn't surprise me—going from v1 to v2, it became possible to use NumPy in places where we previously needed custom kernels. For the GPU, CuPy takes that role (assuming that CuPy is a good replacement for NumPy in those places, which it very likely is, and if not, a shim in its nplike would fix it).

I'd say you can merge when ready. You only need one review, not two, although if @ianna has anything to say about this, please do!

ManasviGoyal:
> This is good to do as a first step, so that you don't accidentally spend time on obsolete kennels. The number of removed kennels doesn't surprise me—going from v1 to v2, it became possible to use NumPy in places where we previously needed custom kernels. For the GPU, CuPy takes that role (assuming that CuPy is a good replacement for NumPy in those places, which it very likely is, and if not, a shim in its nplike would fix it).
> 
@jpivarski These two should be removed too right?
- awkward_reduce_argmax_bool_64
- awkward_reduce_argmin_bool_64

jpivarski:
If you don't see them in v2, then yes.

ianna (approved):
@ManasviGoyal - looks good to me! please, merge it when you are done with it. Thanks!

PR #2879, opened 2023-12-08
fix: register Numba in test

### LINKED ISSUE #2878: `test_numba_ufunc_legacy` fails with `TypeError: return type must be specified for object mode`
### Version of Awkward Array

2.5.1rc1

The `tests/test_2799_numba_ufunc_resolution.py::test_numba_ufunc_legacy` test fails with `TypeError: return type must be specified for object mode` 

```python
% python
>>> import numpy as np
>>> np.__version__
'1.24.3'
>>> import numba
>>> numba.__version__
'0.57.0'
>>> import awkward as ak
>>> ak.__version__
'2.5.1rc1'
```

agoose77:
This PR recognises that Numba isn't being registered for a specific test, and fixes it!

### DIFF
--- a/tests/test_2799_numba_ufunc_resolution.py
+++ b/tests/test_2799_numba_ufunc_resolution.py
@@ -15,6 +15,9 @@ NUMBA_HAS_NEP_50 = packaging.version.parse(
 ) >= packaging.version.Version("0.59.0")
 
 
+ak.numba.register_and_check()
+
+
 @pytest.mark.skipif(not NUMBA_HAS_NEP_50, reason="Numba does not have NEP-50 support")
 def test_numba_ufunc_nep_50():
     raise NotImplementedError


agoose77:
@jpivarski I simplified this to only address a missing `ak.numba.register_and_check()`.

ianna (changes requested):
@agoose77 - I think we also need to check on the `numba` version **greater or equal to** `0.58` to avoid specifying the types for vectorise. Perhaps, adding a warning in case the version does not support it yet?

jpivarski (approved):
Looks good.

About the minimum Numba version: it's true that `@nb.vectorize` without explicit arguments won't work before a certain Numba version, but that could be described as something missing from Numba, rather than something missing from Awkward.

In addition, `@nb.vectorize` is not a very popular function. This plot comes from static analysis of all the non-fork repos that `import numba` (or `from numba import`):

image

(If Numba is imported as another name, like `nb`, this includes `nb.vectorize`, and the `nb.jit` and `nb.njit` have been folded such that `nb.jit` with `nopython=True` counts as `nb.njit` only.)

So it doesn't seem necessary to me for Awkward to require a much newer Numba version just for `vectorize`. Since the Numba dependency is not managed by pip/conda (it's an optional dependency), dealing with mismatched version numbers is more annoying, as it's something the user must do manually.

jpivarski:
All of the tests pass. @ianna, if you agree, you can do the merge. If not, let me know your reasons and we'll talk!

ianna (approved):
Thanks, @jpivarski! Yes, I agree, let's merge it!

PR #2880, opened 2023-12-08
feat: add some simpler `cuda_kernels`

ManasviGoyal:
Added some `cuda_kernels`

- awkward_ByteMaskedArray_reduce_next_nonlocal_nextshifts_fromshifts_64
- awkward_IndexedArray_fill
- awkward_ListArray_fill
- awkward_IndexedArray_numnull_unique
- awkward_RegularArray_rpad_and_clip_axis1
- awkward_index_rpad_and_clip_axis0
- awkward_UnionArray_simplify
- awkward_UnionArray_simplify_one
- awkward_UnionArray_fillindex
- awkward_UnionArray_fillindex_count
- awkward_UnionArray_filltags
- awkward_UnionArray_filltags_const
- awkward_NumpyArray_fill


jpivarski (approved):
These look good, but they need tests (user-level tests in `tests-cuda`).

Do you know which `ak.*` functions would test them? One way to find out is to search the codebase for a kernel name and put in an exception before the kernel runs, then run everything in the `tests` directory to see what fails on the CPU, then just copy that test into `tests-cuda`. If you have a choice, you'd want to find one that doesn't trigger other kernels that don't yet exist in CUDA.

Another way to search is to do the same in the `tests-cuda` directory with an exception on one of the kernels that's similar to the kernel whose test you want to add. The tests in the `tests-cuda` directory are already designed around not touching kernels that don't yet exist. (Eventually, everything will need to work, but you want to get there in small steps.)

These can be merged and their tests can come in another PR, or you can put the tests in this PR for better bookkeeping (so that you can be sure that you got them all; the diff makes a good checklist).

src/awkward/_connect/cuda/cuda_kernels/awkward_ListArray_fill.cu
```diff
@@ -0,0 +1,22 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
+
+template <typename T, typename C, typename U, typename V>
+__global__ void
+awkward_ListArray_fill(T* tostarts,
+                       int64_t tostartsoffset,
```
ianna, line 6: I guess, it's an input, not an output variable. I'd call it `length_so_far` as in the code that uses it. It is the same as `tostopsoffset`.

src/awkward/_connect/cuda/cuda_kernels/awkward_ListArray_fill.cu
```diff
@@ -0,0 +1,22 @@
...
+                       int64_t tostopsoffset,
+                       const U* fromstarts,
+                       const V* fromstops,
+                       int64_t length,
+                       int64_t base,
+                       uint64_t invocation_index,
+                       uint64_t* err_code) {
+  if (err_code[0] == NO_ERROR) {
+    int64_t thread_id = blockIdx.x * blockDim.x + threadIdx.x;
+    if (thread_id < length) {
```
ianna, line 17: I guess, the `length` here is the length of the array.

ianna (commented):
@ManasviGoyal and @jpivarski - I'm ok with the one-to-one copy of the CPU to GPU kernels. However, our code readability has improved a lot (thanks to @agoose77 and @ioanaif :-) I was wandering if we should make the kernels code more readable (e.g self-documented) by using similar to the code base variable names?

jpivarski:
> I was wandering if we should make the kernels code more readable (e.g self-documented) by using similar to the code base variable names?

I'd prefer to not refactor the kernel names and the names of their arguments because

* they're buried deeply, call it "L5" in our public-private scheme
* refactoring will take time and probably introduce some bugs
* right now, there's a big job to do—adding the new kernels—and I don't believe that this job would be easier if the kernel names and arguments were more in line with how they're used in the codebase. The problem of writing kernels is almost perfectly factored out of how the kernels are used in the codebase. "Almost" refers to the fact that the choice of good test cases would come from how they're intended to be used. (In fact, that criterion was already used to define most of the `unit-tests` in the kernel-test-data.json.)

jpivarski:
This PR doesn't affect awkward-cpp, so it does not need to go before @agoose77's upcoming release of awkward-cpp and awkward.

ianna (approved):
No objections to merge it :-)

PR #2882, opened 2023-12-08
chore: use tomli

henryiii:
Little bit of modernization.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -255,7 +255,7 @@ extend-exclude = [
 
 [tool.ruff.lint]
 extend-select = [
-    "B", "B904",   # flake8-bugbear
+    "B",           # flake8-bugbear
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -255,7 +255,7 @@ extend-exclude = [
 
 [tool.ruff.lint]
 extend-select = [
```
agoose77, line 257: @henryiii how did you know to remove this rule? B904? Curious as to your rationale.

henryiii: It doesn't do anything, B includes B904.

henryiii: (unlike Flake8, which needed B9)

agoose77: Gah, I misread the file and didn't see the wide selector (B). Thanks!

PR #2877, opened 2023-12-08
fix: errors in existing `cuda_kernels`

ManasviGoyal:
Some minor fixes in some of the `cuda_kernels`.

PR #2886, opened 2023-12-11
fix: failing tests after #2877

ManasviGoyal:
Just noticed that a few changes in #2877 are making the cuda tests fail. So I have fixed them.

### DIFF
--- a/src/awkward/_connect/cuda/cuda_kernels/awkward_ByteMaskedArray_getitem_nextcarry.cu
+++ b/src/awkward/_connect/cuda/cuda_kernels/awkward_ByteMaskedArray_getitem_nextcarry.cu
@@ -15,8 +15,8 @@ template <typename T, typename C>
 __global__ void
 awkward_ByteMaskedArray_getitem_nextcarry_a(T* tocarry,
                                             const C* mask,
-                                            bool validwhen,
                                             int64_t length,
+                                            bool validwhen,
                                             int64_t* scan_in_array,
                                             uint64_t invocation_index,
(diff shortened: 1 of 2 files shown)


ianna:
@ManasviGoyal - could you, please, run the linter and fix the warnings? Thanks!

ManasviGoyal:
> @ManasviGoyal - could you, please, run the linter and fix the warnings? Thanks!

I think we need to update the `py.version` to `3.8` in `pyproject.toml` to support `typing.final`. Should I go ahead and go it?

ianna:
> > @ManasviGoyal - could you, please, run the linter and fix the warnings? Thanks!
> 
> I think we need to update the `py.version` to `3.8` in `pyproject.toml` to support `typing.final`. Should I go ahead and go it?

why is suddenly failing? what was changed?

agoose77:
@ManasviGoyal is right that we should bump the PyLint Python version. I'd missed that when we last updated our Python bounds. I'm not sure why PyLint suddenly started noticing, though. I think we can safely assume that it's a strictness update on their part being pulled in via CI. No need to dig into it, though - we clearly should correct this on our side.

agoose77:
@ManasviGoyal don't worry about pylint here - I'll fix it in a new PR.

ManasviGoyal:
> @ManasviGoyal don't worry about pylint here - I'll fix it in a new PR.

Thanks!

ManasviGoyal:
@ianna I think this can be squash and merged now, right?

PR #2893, opened 2023-12-12
chore: only copy non-test headers

agoose77:
Whilst checking our wheel to ensure that the necessary headers are included (c.f. #2892), I noticed that an extra header file has snuck into the archive.

This PR ensures we only flatten the `header-only/*/awkward/*` heirarchy.

### DIFF
--- a/dev/copy-cpp-headers.py
+++ b/dev/copy-cpp-headers.py
@@ -20,7 +20,7 @@ if __name__ == "__main__":
     if connect_path.exists():
         shutil.rmtree(connect_path)
     connect_path.mkdir(parents=True)
-    for path in header_only_path.rglob("*.h"):
+    for path in header_only_path.rglob("*/awkward/*.h"):
         dest_path = connect_path / path.name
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
If it was flattening the hierarchy of all `*.h` files, I don't see how awkward-cpp would have been compiling. The `#include` statements use relative subdirectories. For example,

https://github.com/scikit-hep/awkward/blob/ae5923e957406f1c0af3728abafde9dd9cde9097/awkward-cpp/src/libawkward/io/json.cpp#L16-L19

Or was this just an issue in the header files included in the sdist?

Anyway, it's a good thing to fix; please do merge.

agoose77:
@jpivarski these are headers in the header-only part of `awkward` rather than `awkward-cpp`. For our header-only files, each component lives under `awkward/XXX.h`, so `foo/awkward/FOO.h` and `bar/awkward/BAR.h` can be flattened under the same `awkward/` prefix

PR #2896, opened 2023-12-13
feat: configure Julia version and required packages

ianna:
[The Python module JuliaCall](https://juliapy.github.io/PythonCall.jl/stable/juliacall/#The-Python-module-JuliaCall) can be installed as follows:
```
pip install juliacall
```
Here is how it works (see [JuliaPkg](https://github.com/JuliaPy/PyJuliaPkg) for more details):

```python
 % python
>>> from juliacall import Main as jl
[juliapkg] Locating Julia ^1.9
[juliapkg] Using Julia 1.9.2 at /usr/local/bin/julia
[juliapkg] Using Julia project at /Users/yana/anaconda3/julia_env
[juliapkg] Installing packages:
           julia> import Pkg
           julia> Pkg.add([Pkg.PackageSpec(name="AwkwardArray", uuid="7d259134-7f60-4bf1-aa00-7452e11bde56"), Pkg.PackageSpec(name="PythonCall", uuid="6099a3de-0909-46bc-b1f4-468b9a2dfc0d")])
           julia> Pkg.resolve()
           julia> Pkg.precompile()
   Resolving package versions...
    Updating `~/anaconda3/julia_env/Project.toml`
  [7d259134] + AwkwardArray v0.1.2
  [6099a3de] + PythonCall v0.9.15
    Updating `~/anaconda3/julia_env/Manifest.toml`
  [7d259134] + AwkwardArray v0.1.2
  [992eb4ea] + CondaPkg v0.2.22
...
```


jpivarski (approved):
Does this need to be added to a MANIFEST or something to be available to users who install awkward from the sdist or from a wheel? We want this to work for users who install through pip and conda, not just by checking out GitHub.

Also, I suppose the version number will need to be updated after every AwkwardArray.jl release. Can it be set to "latest"? If not, that's okay. It will just become part of our normal workflow to increase this number _after_ an AwkwardArray.jl release becomes public.

Otherwise, this is good! We should do whatever is necessary to streamline the interface between Python and Julia.

agoose77:
@ianna from the README, it looks like we want this file to exist under `site-packages/awkward/juliapkg.json`. Is that so? I'd be keen to work on this, it might take some thinking about w.r.t editable installs etc.

ianna:
> @ianna from the README, it looks like we want this file to exist under `site-packages/awkward/juliapkg.json`. Is that so? I'd be keen to work on this, it might take some thinking about w.r.t editable installs etc.

@agoose77 - Thank you! It would be great if you could find time :-)

agoose77:
I've modified the build configuration. We should explicitly check that editable installs perform this rewrite (they should, though; hatch uses the wheel target).

ianna (commented):
@agoose77 and @jpivarski - I think, this is good to be merged. Thanks!

jpivarski:
Okay, I'll add this to the set that are being tested and will be included in awkward 2.5.2. Thanks!

PR #2898, opened 2023-12-13
feat: add to_parquet_dataset function

zbilodea:


jpivarski (approved):
In Awkward 2.x, we use `fsspec`[^1] instead of `os` routines to get filenames and directory listings, since they can be remote, with URI schemes like `https://`, `root://`, or `ssh://`. I can see how this implementation follows Awkward 1.x (but is much better, being a separate function now).

You could merge this implementation as-is and add a feature-request Issue to expand it to fsspec. We could consider the local-filesystem implementation a first working version and remote-filesystems to be an improvement.

Cc: @martindurant

[^1]: In the next minor release, [2.6.0 on February 1](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20), fsspec will become a strict dependency.

src/awkward/operations/ak_to_parquet_dataset.py
```diff
@@ -0,0 +1,115 @@
+from __future__ import annotations
+
+__all__ = ("to_parquet_dataset",)
+
+
+def to_parquet_dataset(directory, filenames=None, filename_extension=".parquet"):
```
martindurant: Essentially the same as fastparquet.merge?

src/awkward/operations/ak_to_parquet_dataset.py
```diff
@@ -0,0 +1,115 @@
...
+    The `_common_metadata` contains the schema that all files share. (If the files
+    have different schemas, this function raises an exception.)
+
+    The `_metadata` contains row-group metadata used to seek to specific row-groups
+    within the multi-file dataset.
+    """
+
+    # Implementation
+
+    from os import fsdecode, path
```
martindurant: Why restrict to local paths?

src/awkward/operations/ak_to_parquet_dataset.py
```diff
@@ -0,0 +1,115 @@
...
+    )
+    pyarrow_parquet.write_metadata(schema, path.join(directory, "_common_metadata"))
+    pyarrow_parquet.write_metadata(
+        schema,
+        path.join(directory, "_metadata"),
+        metadata_collector=metadata_collector,
+    )
+
+
+def _regularize_path(path):
```
martindurant: This looks like fsspec work

src/awkward/operations/ak_to_parquet_dataset.py
```diff
@@ -0,0 +1,115 @@
...
+def _common_parquet_schema(pq, filenames, relpaths):
+    assert len(filenames) != 0
+
+    schema = None
+    metadata_collector = []
+    for filename, relpath in zip(filenames, relpaths):
+        if schema is None:
+            schema = pq.ParquetFile(filename).schema_arrow
+            first_filename = filename
+        elif not schema.equals(pq.ParquetFile(filename).schema_arrow):
```
martindurant: Whilst not cleanly defined, parquet does in principle allow schema evolution, but perhaps not in conjunction with (_common)_metadata

martindurant (commented):
I have commented on this before: _metadata is still used by fastparquet, but not many other frameworks. Some frameworks no longer even provide the option to create it. This is because the global file becomes tricky in the place of many datafiles, and can be slow to read too (which must happen in the client even on a distributed system). Any directory of parquet files will be seen by fastparquet, (py)arrow, dask, spark ... as a dataset.

(edit: I see @jpivarski already addressed some of the "let's use fsspec" comment).

src/awkward/operations/ak_to_parquet_dataset.py
jpivarski: Addressed above. @zbilodea started with the Awkward 1.x implementation as a model, which was strictly local filesystem. That's something we want to expand here, but I think it can be done in two steps, with this PR being the first step.

Actually, that's Zoë's choice: Zoë, if you want to continue working on this PR to add fsspec now, that's fine.

jpivarski:
> _metadata is still used by fastparquet, but not many other frameworks. Some frameworks no longer even provide the option to create it.

But it's not a problem if we create it, right? We're also making the _common_metadata. Maybe the choice can be a boolean function argument.

src/awkward/operations/ak_to_parquet_dataset.py
zbilodea: Okay! I think I will continue add fsspec now then

martindurant:
I would add to the general discussion here (perhaps add to docstring), the real benefits of consolidated metadata are:
- single read to get the metadatas of all files, which can decrease latency for many small reads (not an issue when distributing the work)
- allows for up-front filtering out of row-groups based on column chunk statistics

martindurant:
> it's not a problem if we create it, right?

It should never be a problem - some frameworks might just ignore it

zbilodea:
@jpivarski Having a strange problem here - the fsspec implementation worked with S3, but now there's two tests failing because of this:

_schema = pyarrow_parquet.ParquetFile(filepath, filesystem=fs).schema_arrow
Throwing: TypeError: __init__() got an unexpected keyword argument 'filesystem'_

It's only happening with two out of four tests (there are no obvious similarities), and 'filesystem' is definitely a keyword argument as shown in the [docs...] (https://arrow.apache.org/docs/python/generated/pyarrow.parquet.ParquetFile.html#:~:text=most%20Parquet%20files.-,filesystem,-FileSystem%2C%20default%20None) 
I'll keep trying to figure it out

martindurant:
@zbilodea , the "minimal" Ci run has pyarrow 7.0.0 (2 yr old), so I'm not too surprised if the API has changed in that time.

martindurant:
Here is the v7 doc page: https://arrow.apache.org/docs/7.0/python/generated/pyarrow.parquet.ParquetFile.html , and indeed I don't see filesystem= . The m=previous method would be to open the file and pass that:

```
with fsspec.open(...) as f:
     pa.ParquetFile(f, ...)
```

jpivarski:
What's the status of this PR, @zbilodea?

zbilodea:
I made all the changes, tested it with S3 and it worked. All set as far as I can tell!

jpivarski:
I think this is done but we lost track of it before merging.

If it's done, @zbilodea, please merge it!

PR #2899, opened 2023-12-14
feat: add string builder

lobis:
Add a string builder to the header-only library to make building arrays of strings easier.


header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -388,6 +388,23 @@ namespace awkward {
       size_t id_;
     };
 
+      /// @class String
+      template<class PRIMITIVE>
+      class String : public ListOffset<PRIMITIVE, Numpy<uint8_t>> {
+      public:
+          String() : ListOffset<PRIMITIVE, Numpy<uint8_t>>() {
+      this->set_parameters(R"""("__array__": "string")""");
```
ianna, line 399: @lobis - please, make sure to run `clang-format`

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -388,6 +388,23 @@ namespace awkward {
...
+      this->content().set_parameters(R"""("__array__": "char")""");
+  }
+
+    void append_string(const std::string& value) {
+        this->begin_list();
+        for (const auto c: value) {
```
ianna:
```suggestion
        for (const auto& c: value) {
```

ianna (changes requested):
@lobis - looks great! I don't think `clang-format` was run. Could you please run it by hand? Thanks!

lobis:
By the way I noticed that there is not `clang-format` in the pre-commit

> @lobis - looks great! I don't think `clang-format` was run. Could you please run it by hand? Thanks!

I manually formatted the file (no clang-format).

I wasn't able to run clang-format without causing a huge diff so I created a separate PR for this (https://github.com/scikit-hep/awkward/pull/2902).

ianna (approved):
@lobis - Thanks, it looks good to me.

jpivarski (approved):
Great, thanks! I approve it as well.

I'm pretty sure this will automatically pass all of the test (unless there's a C++ compilation error or something).

Is it ready to merge, assuming that it does pass tests?

lobis:
> Great, thanks! I approve it as well.
> 
> I'm pretty sure this will automatically pass all of the test (unless there's a C++ compilation error or something).
> 
> Is it ready to merge, assuming that it does pass tests?

Ready!

jpivarski:
@all-contributors please add @lobis for code

Issue #1842, opened 2022-10-27
Introduce clang-format in pre-commit


jpivarski:
(Almost) exactly following https://github.com/scikit-hep/fastjet/pull/125.

The first time we do this will be a big diff in the C++ files, but being in pre-commit will prevent trivial differences in C++ which might make our awkward-core testing infrastructure mandate a new release.

This is a low priority; nice to have before the 2.0.0 release (so that 2.0.0 onward has minimal C++ diffs), and it should be after the awkward/awkward-core split: PR #1778.

jpivarski:
The first attempt at this was #1865.

But it should be reapplied to post-v2 code; it should be mostly LLVM style with a few exceptions, most of which were discussed in the comments on #1865.

There's also some configuration to help it find our *.h, *.cpp files, and *.cu (and *.cuh?), which are in a few strange places.

jpivarski:
Note that #2902 was an attempt (now closed because it's stale). This is still a good thing to do, but it would have to be a concerted effort, since we only want to do it once.

PR #35, opened 2019-12-18
[WIP] Test clang-tidy with selected checks

ianna:
Apply fixes from **clang-tidy** with **-checks**='-*,
   boost-use-to-string,
   misc-string-compare,
   misc-uniqueptr-reset-release,
   modernize-deprecated-headers,
   modernize-make-shared,
   modernize-use-bool-literals,
   modernize-use-equals-delete,
   modernize-use-nullptr,
   modernize-use-override,
   performance-unnecessary-copy-initialization,
   readability-container-size-empty,
   readability-redundant-string-cstr,
   readability-static-definition-in-anonymous-namespace,
   readability-uniqueptr-delete-release'

https://clang.llvm.org/extra/clang-tidy/

@jpivarski - please, review and comment

PR #34, opened 2019-12-17
[WIP] Test clang-format

ianna:
C and C++ files automatically formatted with clang-format code. Please, see 
https://electronjs.org/docs/development/clang-format for more details.

@jpivarski - you can also integrate clang-format directly into Atom.

PR #2902, opened 2023-12-14
ci: add clang-format to pre-commit configuration

lobis:


lobis:
Wow the pre-commit bot did a really big diff, is the clang-format being ran correctly and it wasn't formatted before or what is going on? I'm having troubles running this `.clang-format` file myself so can't really tell.

lobis:
Not sure why this is such a big diff @agoose77 , perhaps you want to make some changes?

agoose77:
This was something that @ianna and @jpivarski were discussing a while ago. I don't recall if we ever reached a conclusion on the style and went as far as enforcing it. Perhaps they can chime in?

awkward-cpp/include/awkward/common.h
```diff
@@ -49,44 +52,43 @@
...
-  const int64_t  kMaxInt64  = 9223372036854775806;   // 2**63 - 2: see below
-  const int64_t  kSliceNone = kMaxInt64 + 1;         // for Slice::none()
-  const int64_t  kMaxLevels =                  48;
+const int8_t kMaxInt8 = 127;                    // 2**7  - 1
+const uint8_t kMaxUInt8 = 255;                  // 2**8  - 1
+const int32_t kMaxInt32 = 2147483647;           // 2**31 - 1
+const uint32_t kMaxUInt32 = 4294967295;         // 2**32 - 1
+const int64_t kMaxInt64 = 9223372036854775806;  // 2**63 - 2: see below
+const int64_t kSliceNone = kMaxInt64 + 1;       // for Slice::none()
+const int64_t kMaxLevels = 48;
```
jpivarski, line 68: I liked the right-formatting of these numbers, but it's impressive that it exactly maintained the alignment of the comments (which is more important, anyway).

awkward-cpp/src/cpu-kernels/awkward_BitMaskedArray_to_IndexedOptionArray.cpp
```diff
@@ -1,148 +1,130 @@
...
-  const uint8_t* frombitmask,
-  int64_t bitmasklength,
-  bool validwhen,
-  bool lsb_order) {
+ERROR
+awkward_BitMaskedArray_to_IndexedOptionArray(T* toindex,
+                                             const uint8_t* frombitmask,
+                                             int64_t bitmasklength,
+                                             bool validwhen,
+                                             bool lsb_order) {
```
jpivarski, line 16: This kind of indentation is bad for limited horizontal space. Also, there's a reason black chose to put all of the arguments on a different line from the function name: with this formatting, changing the length of the name would change the indentation of all of the arguments, which is a bigger, more noisy change.

If it's possible to ask for black-style indentation here, that would be an improvement.

awkward-cpp/src/cpu-kernels/awkward_BitMaskedArray_to_IndexedOptionArray.cpp
```diff
@@ -1,148 +1,130 @@
...
+    for (int64_t i = 0; i < bitmasklength; i++) {
       uint8_t byte = frombitmask[i];
       if ((byte & ((uint8_t)1)) == validwhen) {
-        toindex[i*8 + 0] = i*8 + 0;
-      }
-      else {
-        toindex[i*8 + 0] = -1;
+        toindex[i * 8 + 0] = i * 8 + 0;
+      } else {
+        toindex[i * 8 + 0] = -1;
```
jpivarski, line 23: I disagree with `} else {`, but not strongly.

awkward-cpp/src/cpu-kernels/awkward_ByteMaskedArray_getitem_nextcarry_outindex.cpp
```diff
@@ -1,39 +1,38 @@
 // BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
 
-#define FILENAME(line) FILENAME_FOR_EXCEPTIONS_C("src/cpu-kernels/awkward_ByteMaskedArray_getitem_nextcarry_outindex.cpp", line)
+#define FILENAME(line)                                          \
+  FILENAME_FOR_EXCEPTIONS_C(                                    \
+      "src/cpu-kernels/"                                        \
+      "awkward_ByteMaskedArray_getitem_nextcarry_outindex.cpp", \
+      line)
```
jpivarski, line 7: These lines were easier to edit and check in bulk when they were one-liners, but it could be hard to explain to the formatter that only macros are allowed to violate the line width rule.

Before this PR, awkward-cpp/src/cpu-kernels/awkward_ListOffsetArray_reduce_nonlocal_outstartsstops_64.cpp already had a multi-line implementation. That was the only file, so the pattern was already broken (automated scans that assumed that it's one line would have failed to catch that kernel).

So I guess this is fine. At least it's enforced.

jpivarski:
@ianna and I had discussed it in #1842, #1865, #34, and #35.

This does have a lot of diffs, mostly in whitespace. We can merge a PR like this, but only once, after conclusively deciding on a set of style options.

Clang-format/clang-tidy doesn't come with a set of opinions, the way that black does, but there must be some de facto standards, like LLVM Style. Which is the most popular one?

jpivarski:
The issue, #1842, is still open, and we should try again someday. This (closed) PR will be helpful for whoever tries to do it in the future.

PR #2909, opened 2023-12-19
fix: include header-only files in sdist

### LINKED ISSUE #2897: Header-only libraries are not included in SDist archives
### Version of Awkward Array

main

This is really frustrating. It turns out that our header-only libraries have not been included in our sdists, which in turn means that they have not ended up in our conda archives.

agoose77:
Fixes #2897 by ensuring that our SDist includes the header-only files. This is necessary both because semantically the SDist should represent a snapshot of the repo (in our case a generated snapshot), and because it's possible to build wheels from SDists, which is not possible if files are missing.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -74,3 +74,8 @@ repos:
       files: src
       additional_dependencies:
         - numpy>=1.24
+
+- repo: https://github.com/abravalheri/validate-pyproject
+  rev: v0.15
+  hooks:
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
~~Okay, good—this is another way to get header-only code, though I would think that users ordinarily wouldn't get it in a Python package; they'd use the CMakeLists fetch or explicitly download it...~~

Nope: the motivation for this is to make it possible to compile from the sdist. Full steam ahead!

Issue #2760, opened 2023-10-17
`ak.softmax` with `axis=0` fails


agoose77:
### Version of Awkward Array

main

### Description and code to reproduce

I suspect this is a general bug with reducers implemented through multiple-step operations, i.e. not just `axis=0`.

The problem arises when reducing a ragged array e.g.
```python
import awkward as ak

assert isinstance(
    ak.softmax([[1, 2, 3, 4], [5], [10]], axis=0, highlevel=True), ak.Array
)
```

jpivarski:
I want to fix this issue because pytest shows xfailed tests in red, and I keep thinking that it's an unexpected error, rather than an expected error.

The commentary on #2918 says that this might be similar to an error that was fixed in `ak.std` and `ak.var`. In that error, we noticed that only reduction along `axis=-1` (or `axis=None`, trivially) kept an array broadcastable with its original self; other `axis` values change the shape in ways that are no longer broadcastable. It could be _accidentally_ broadcastable, and that's even worse: it would be the wrong broadcasting for the operation.

The underlying issue is the same as in #2918, but the fix can't be the same. For `ak.std` and `ak.var`, we could replace $E[(x - E(x))^2]$ with $E[x^2] - E[x]^2$. The first expression has the broadcasting problem, but the second doesn't. Both expectation values sum along the same axis, producing the same shape independently, but $x - E[x]$ has to broadcast the original shape, $x$, with a reduced shape, $E[x]$.

However, softmax is $\exp(x)/\Sigma\exp(x)$, so the original shape, $\exp(x)$, has to be broadcasted against the reduced shape, $\Sigma\exp(x)$ in the division operation. I don't think this expression can be rewritten in a way that avoids that.

For the specific test case in question,

```python
>>> x = ak.Array([[1, 2, 3, 4], [5], [10]])
```

the `axis=-1` case that passes is

```python
>>> np.exp(x) / ak.sum(np.exp(x), axis=1)
<Array [[0.0321, 0.0871, 0.237, 0.644], [1], [1]] type='3 * var * float64'>
```

and the `axis=0` case that fails is

```python
>>> np.exp(x) / ak.sum(np.exp(x), axis=0)
...
ValueError: cannot broadcast RegularArray of size 3 with RegularArray of size 4 in divide

This error occurred while calling

    numpy.divide.__call__(
        <Array [[2.72, 7.39, 20.1, 54.6], ...] type='3 * var * float64'>
[...]

PR #2918, opened 2024-01-02
fix: ak.std and ak.var for axis != -1.

jpivarski:
Although $\frac{1}{n}\sum_i (x_i - \mbox{mean}(x))^2$ is more precise than $(\frac{1}{n}\sum_i x_i^2) - \mbox{mean}(x)$, the latter can only be broadcasted when `axis=-1`. This had been failing without a useful error message; now it works (with less precision).

### DIFF
--- a/src/awkward/operations/ak_var.py
+++ b/src/awkward/operations/ak_var.py
@@ -188,16 +188,6 @@ def _impl(x, weight, ddof, axis, keepdims, mask_identity, highlevel, behavior, a
     weight = ctx.wrap(weight_layout, allow_other=True)
 
     with np.errstate(invalid="ignore", divide="ignore"):
-        xmean = ak.operations.ak_mean._impl(
-            x,
-            weight,
-            axis,
-            keepdims=True,
-            mask_identity=True,
-            highlevel=True,
-            behavior=ctx.behavior,
-            attrs=ctx.attrs,
-        )
         if weight is None:
             sumw = ak.operations.ak_count._impl(
                 x,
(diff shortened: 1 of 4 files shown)


agoose77:
This is probably the same issue as #2760, so we should look at `softmax` too. I'd be tempted to keep the `axis=-1` path, and add the new logic as an `axis != -1` case in order to improve precision of the result in what I suspect would be the most common case.

jpivarski:
Note to self:

```
      def test_dtype_deprecated():
          form = ak.forms.EmptyForm()
  >       with pytest.warns(
              DeprecationWarning,
              match=r"the `dtype` parameter in EmptyForm\.to_NumpyForm is deprecated",
          ):
  E       DeprecationWarning: In version 2.4.0, this will be an error.
  E       To raise these warnings as errors (and get stack traces to find out where they're called), run
  E           import warnings
  E           warnings.filterwarnings("error", module="awkward.*")
  E       after the first `import awkward` or use `@pytest.mark.filterwarnings("error:::awkward.*")` in pytest.
  E       Issue: from_dtype conversion of temporal units to generic `datetime64` and `timedelta64` types is deprecated, pass `time_units_as_parameter=False` to disable this warning..
  
  form       = EmptyForm()
  next_form  = NumpyForm('int64')
  
  /project/tests/test_2503_deprecate_to_numpyform.py:13: DeprecationWarning
```

jpivarski:
> This is probably the same issue as #2760, so we should look at `softmax` too. I'd be tempted to keep the `axis=-1` path, and add the new logic as an `axis != -1` case in order to improve precision of the result in what I suspect would be the most common case.

I noticed this in association with softmax when I had to turn a test from xfail to normal. I haven't looked at softmax yet, but it's likely the same issue of trying to compute it by calling the other semi-reducer functions [^1]. For `axis=-1`, calling another semi-reducer, even with the right `axis`, puts it into a shape that isn't broadcastable with the original. Computing things from associate reducers only removes the need for any broadcasting.

About the precision, it would be a bad surprise for different `axis` values to have different precisions due to different algorithms. This is as much something to think about as the Kahan summation (which would turn even `ak.sum` into a "semi-reducer.")

[^1]: by which I mean functions that do reduce a dimension, but do not satisfy associativity.

PR #1514, opened 2022-06-22
Update type-parser for v2

jpivarski:
A more proper fix than #1513. (See continuing issues in scikit-hep/uproot5/pull/620.)

Issue #2920, opened 2024-01-03
All deprecations should be finalized now; no deprecations are (currently) pending


jpivarski:
### Version of Awkward Array

HEAD

### Description and code to reproduce

```bash
% fgrep -r 'deprecate(' src/awkward | fgrep -v 'def deprecate'
src/awkward/forms/emptyform.py:            deprecate(
src/awkward/forms/form.py:            deprecate(
src/awkward/forms/form.py:            deprecate(
src/awkward/forms/numpyform.py:        deprecate(
src/awkward/contents/numpyarray.py:        deprecate(
src/awkward/contents/emptyarray.py:        deprecate(
```

jpivarski:
Nope—here's one for the future: `np.asarray(layout)` (for `EmptyArray` and `NumpyArray`) will be disabled in Awkward 2.6.0, but that hasn't happened yet.

https://github.com/scikit-hep/awkward/blob/d466d7047fb4b867fb07051b7154d81a5c627348/src/awkward/contents/emptyarray.py#L113-L118

https://github.com/scikit-hep/awkward/blob/d466d7047fb4b867fb07051b7154d81a5c627348/src/awkward/contents/numpyarray.py#L169-L174

PR #2926, opened 2024-01-09
chore: restore files that generated the type-parser

jpivarski:
This is how [src/awkward/types/_awkward_datashape_parser.py](https://github.com/scikit-hep/awkward/blob/main/src/awkward/types/_awkward_datashape_parser.py) was made.

These files were taken from 4ba1dbb8712fa9a00e86385e04548d646047cffc (#1514), the last commit before they disappeared in the v1 → v2 transition. I _think_ this was their final state—I _think_ they weren't developed any further. (Am I missing anything, @reikdas?)

These should remain in the dev directory in case we ever need to make changes to the grammar or update the generated file to a new Lark version or a new Python version.

For reference, the auto-generated file was made according to the procedure documented in

* https://lark-parser.readthedocs.io/en/latest/examples/standalone/json_parser_main.html
* https://lark-parser.readthedocs.io/en/latest/tools.html
* http://blog.erezsh.com/create-a-stand-alone-lalr1-parser-in-python/

Since this PR only touches files in the dev directory, I'm going to merge it without review.


agoose77:
Huh, this was around at some point. Didn't realise we lost them! Good catch.

ianna:
> This is how [src/awkward/types/_awkward_datashape_parser.py](https://github.com/scikit-hep/awkward/blob/main/src/awkward/types/_awkward_datashape_parser.py) was made.
> 
> These files were taken from [4ba1dbb](https://github.com/scikit-hep/awkward/commit/4ba1dbb8712fa9a00e86385e04548d646047cffc) (#1514), the last commit before they disappeared in the v1 → v2 transition. I _think_ this was their final state—I _think_ they weren't developed any further. (Am I missing anything, @reikdas?)
> 
> These should remain in the dev directory in case we ever need to make changes to the grammar or update the generated file to a new Lark version or a new Python version.
> 
> For reference, the auto-generated file was made according to the procedure documented in
> 
> * https://lark-parser.readthedocs.io/en/latest/examples/standalone/json_parser_main.html
> * https://lark-parser.readthedocs.io/en/latest/tools.html
> * http://blog.erezsh.com/create-a-stand-alone-lalr1-parser-in-python/
> 
> Since this PR only touches files in the dev directory, I'm going to merge it without review.

@jpivarski - it looks like there were tests in `awkward/studies/type-parser/parser.py` if I look into it for a time of PR1522 (my local copy).

jpivarski:
@ianna was right (in a now-deleted comment). These were saved, but they were in the `studies/` directory, rather than `dev/`. The files I included here are from an intermediate state between [studies/type-parser](https://github.com/scikit-hep/awkward/tree/main/studies/type-parser) and [studies/type-parser-2](https://github.com/scikit-hep/awkward/tree/main/studies/type-parser-2), so I put these files into [studies/type-parser-1.5](https://github.com/scikit-hep/awkward/tree/decfa0903e11884af9fbc644dab9248560499dac/studies/type-parser-1.5).

These diffs are not too big:

```bash
diff dev/typeparser/parser.py studies/type-parser/parser.py
diff dev/typeparser/type-grammar.lark studies/type-parser/grammar.lark
```

I can confirm that [studies/type-parser-2](https://github.com/scikit-hep/awkward/tree/main/studies/type-parser-2) (added in #1514) is the most recent; it added our current [src/awkward/types/_awkward_datashape_parser.py](https://github.com/scikit-hep/awkward/blob/main/src/awkward/types/_awkward_datashape_parser.py) and there have been [no substantial changes to it since then](https://github.com/scikit-hep/awkward/commits/main/src/awkward/_v2/types/_awkward_datashape_parser.py).

jpivarski:
One test is failing because of #2920. I'll get to that later.

jpivarski:
The other tests aren't running because this PR only puts data into the studies/ directory, so it's sufficiently tested to merge.

PR #2929, opened 2024-01-10
fix: cartesian nested record

### LINKED ISSUE #2928: ak.cartesian(..., nested=True) results in behaviour expected from nested=False
### Version of Awkward Array

2.5.1

In awkward 2 the behaviour from ak.cartesian is unexpected and deviant from behaviour in v1.

```
>>> one = ak.zip({"pt": [[1, 2], [1]], "eta": [[1.1, 2.1], [1]]})
>>> two = ak.zip({"pt": [[2.1], [2.21, 2.22]], "eta": [[2.11], [2.221, 2.222]]})
>>> ak.cartesian({"foo": one, "bar": two}, nested=False)["foo"].show()
[[{pt: 1, eta: 1.1}, {pt: 2, eta: 2.1}],
 [{pt: 1, eta: 1}, {pt: 1, eta: 1}]]
ak.cartesian({"foo": one, "bar": two}, nested=True)["foo"].show()
[[{pt: 1, eta: 1.1}, {pt: 2, eta: 2.1}],
 [{pt: 1, eta: 1}, {pt: 1, eta: 1}]]
```

In v1 (1.9.0) the behaviour was:
```
>>> for x in ak.cartesian({"foo": one, "bar": two}, nested=True)["foo"]:
>>>     print(x)
[[{pt: 1, eta: 1.1}], [{pt: 2, eta: 2.1}]]
[[{pt: 1, eta: 1}, {pt: 1, eta: 1}]]
>>> for x in ak.cartesian({"foo": one, "bar": two}, nested=False)["foo"]:
>>>     print(x)
[{pt: 1, eta: 1.1}, {pt: 2, eta: 2.1}]
[{pt: 1, eta: 1}, {pt: 1, eta: 1}]
```

Am I misinterpreting what `ak.cartesian(..., nested=True)` should do or is this a bug?

Thanks for considering this comment!

agoose77:
Fixes #2928, by normalising string field names to integers.

This code could be further refactored; it's becoming clear that our `recursively_apply` abstraction would benefit from better support for smarter continuations (i.e. re-entrant with a different callable). That can be tackled later, though.

### DIFF
--- a/src/awkward/operations/ak_cartesian.py
+++ b/src/awkward/operations/ak_cartesian.py
@@ -312,7 +312,7 @@ def _impl(arrays, axis, nested, parameters, with_name, highlevel, behavior, attr
 
         result = ak.contents.RecordArray(outs, fields, parameters=parameters)
         for i in range(len(array_layouts))[::-1]:
-            if i in nested:
+            if i in nested_as_index:
                 result = ak.contents.RegularArray(result, layouts[i + 1].length, 0)
 
     else:
--- a/src/awkward/operations/ak_cartesian.py
+++ b/src/awkward/operations/ak_cartesian.py
@@ -290,8 +290,11 @@ def _impl(arrays, axis, nested, parameters, with_name, highlevel, behavior, attr
 
     backend = next((layout.backend for layout in layouts), cpu)
     if posaxis == 0:
+        # Translate nested field names to nested field indices
         if fields is not None:
-            nested = [i for i, name in enumerate(fields) if name in nested]
+            nested_as_index = [i for i, name in enumerate(fields) if name in nested]
+        else:
+            nested_as_index = nested
 
         indexes = [
             ak.index.Index64(backend.index_nplike.reshape(x, (-1,)))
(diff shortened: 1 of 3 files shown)


jpivarski (approved):
I've been scanning over the code and I don't see how it changes things. Previously, the name `nested` was reused, which is bad practice—naming the possibly-new list `nested_as_index` is a good idea. But I didn't see any further uses of `nested` that were previously getting clobbered.

Is it a function-scoping issue? Python has a gotcha in which functions defined in a loop that close over a variable that changes in the loop don't end up with the value of the variable at the time when the function was defined, but at the time when the function is called—that can be tricky. But there aren't any loops here.

I don't see how it does it, but it evidently does make a difference. Some of the tests are failing: are they stale expected values, derived based on the old, wrong behavior?

agoose77:
@jpivarski I introduced a new bug with _this_ PR — I changed tack during fixing of the bug and must have stepped too far back in the editor history. I had thought that such a simple fix would not break anything else, so I didn't touch the local testing and let the CI handle it! That shows me!

As for the change itself, the issue was indeed that we're using `nested` to mean too many things. Now, it's clear that we test indices against `nested_as_index`, which ensures the cause of the original bug (test for axis name in list of integers) can't happen.

PR #2935, opened 2024-01-11
fix: cast `bool_` to `int8` for positional reducers

### LINKED ISSUE #2934: `ak.argmax` and `ak.argmin`: support for Boolean arrays

`ak.argmax` and `ak.argmin` do not support Boolean arrays unless they are first cast to integer. Meanwhile, `np.argmax` and `np.argmin` can operate on Boolean arrays directly:

```python
import awkward as ak
import numpy as np

assert ak.__version__ == '2.5.1'
assert np.__version__ == '1.24.4'

np_bool_array = np.array([[1, 2, 3], [3, 2, 1]]) > 2
assert np.argmax(np_bool_array, axis=1).tolist() == [2, 0]
assert np.argmin(np_bool_array, axis=1).tolist() == [0, 1]

ak_bool_array = ak.Array([[1, 2, 3], [4, 3, 2, 1]]) > 2

try:
    ak.argmax(ak_bool_array, axis=1)
except KeyError:
    print("this doesn't work :(")

try:
    ak.argmax(np_bool_array, axis=1)
except KeyError:
    print("this doesn't work either :(")

# casting to `int` works:
ak_int_array = ak.values_astype(ak_bool_array, int)
assert ak.argmax(ak_int_array, axis=1).tolist() == [2, 0]
assert ak.argmin(ak_int_array, axis=1).tolist() == [0, 2]
```

agoose77:
Fixes #2934

### DIFF
--- a/src/awkward/_reducers.py
+++ b/src/awkward/_reducers.py
@@ -129,6 +129,14 @@ class ArgMin(KernelReducer):
     preferred_dtype: Final = np.int64
     needs_position: Final = True
 
+    @classmethod
+    def _dtype_for_kernel(cls, dtype: DTypeLike) -> DTypeLike:
+        dtype = np.dtype(dtype)
+        if dtype == np.bool_:
+            return np.dtype(np.int8)
+        else:
+            return super()._dtype_for_kernel(dtype)
+
     def apply(
         self,
         array: ak.contents.NumpyArray,
@@ -182,6 +190,14 @@ class ArgMax(KernelReducer):
     name: Final = "argmax"
     preferred_dtype: Final = np.int64
     needs_position: Final = True
+
+    @classmethod
+    def _dtype_for_kernel(cls, dtype: DTypeLike) -> DTypeLike:
+        dtype = np.dtype(dtype)
+        if dtype == np.bool_:
+            return np.dtype(np.int8)
+        else:
+            return super()._dtype_for_kernel(dtype)
 
     def apply(
         self,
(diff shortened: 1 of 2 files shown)


src/awkward/_reducers.py
```diff
@@ -129,6 +129,14 @@ class ArgMin(KernelReducer):
     preferred_dtype: Final = np.int64
     needs_position: Final = True
 
+    @classmethod
+    def _dtype_for_kernel(cls, dtype: DTypeLike) -> DTypeLike:
+        dtype = np.dtype(dtype)
+        if dtype == np.bool_:
+            return np.dtype(np.int8)
```
ianna, line 136: @agoose77 - from Python documentation: "[bool](https://docs.python.org/3/library/functions.html#bool) is a subclass of [int](https://docs.python.org/3/library/functions.html#int) (see [Numeric Types — int, float, complex](https://docs.python.org/3/library/stdtypes.html#typesnumeric)). In many numeric contexts, False and True behave like the integers 0 and 1, respectively. However, relying on this is discouraged; explicitly convert using [int()](https://docs.python.org/3/library/functions.html#int) instead." 
Viewing a `bool` as an `int8` is correct!

ianna (approved):
@agoose77 - thanks, it looks great!

from Python documentation: "[bool](https://docs.python.org/3/library/functions.html#bool) is a subclass of [int](https://docs.python.org/3/library/functions.html#int) (see [Numeric Types — int, float, complex](https://docs.python.org/3/library/stdtypes.html#typesnumeric)). In many numeric contexts, False and True behave like the integers 0 and 1, respectively. However, relying on this is discouraged; explicitly convert using [int()](https://docs.python.org/3/library/functions.html#int) instead."
Viewing a bool as an int8 is correct!

PR #2940, opened 2024-01-12
ci: fix name of artifacts in deploy-cpp

agoose77:
This was done on-the-fly, so it might have mistakes. I'll give it another look before merging.

The changes to upload-artifact include breaking merging of same-name artifacts (the default behaviour if artifacts are unnamed). This PR adds explicit names, and downloads artifacts by a pattern.

### DIFF
--- a/.github/workflows/deploy-cpp.yml
+++ b/.github/workflows/deploy-cpp.yml
@@ -48,7 +48,7 @@ jobs:
     - name: Check metadata
       run: pipx run twine check awkward-cpp/dist/*
 
-    - uses: actions/upload-artifact@v3
+    - uses: actions/upload-artifact@v4
       with:
         path: awkward-cpp/dist/*.tar.gz
 
@@ -109,7 +109,7 @@ jobs:
         )
 
     - name: Upload wheels
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         path: wheelhouse/*.whl
 
@@ -158,7 +158,7 @@ jobs:
         )
 
     - name: Upload wheels
-      uses: actions/upload-artifact@v3
+      uses: actions/upload-artifact@v4
       with:
         path: wheelhouse/*.whl
 
@@ -172,7 +172,7 @@ jobs:
       name: "pypi"
       url: "https://pypi.org/project/awkward-cpp/"
     steps:
-    - uses: actions/download-artifact@v3
+    - uses: actions/download-artifact@v4
       with:
         name: artifact
         path: dist
--- a/.github/workflows/deploy-cpp.yml
+++ b/.github/workflows/deploy-cpp.yml
@@ -50,6 +50,7 @@ jobs:
 
     - uses: actions/upload-artifact@v4
       with:
+        name: awkward-sdist
         path: awkward-cpp/dist/*.tar.gz
 
   build_wheels:
@@ -111,6 +112,7 @@ jobs:
     - name: Upload wheels
       uses: actions/upload-artifact@v4
       with:
+        name: awkward-wheels-${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.build }}
         path: wheelhouse/*.whl
 
 
@@ -160,6 +162,7 @@ jobs:
     - name: Upload wheels
       uses: actions/upload-artifact@v4
       with:
+        name: awkward-wheels-${{ matrix.arch }}-py${{ matrix.python }}
         path: wheelhouse/*.whl
 
   upload_all:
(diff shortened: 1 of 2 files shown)


jpivarski:
```
3081 Segmentation fault      (core dumped) python -m pytest -vv -rs tests --cov=awkward --cov-report=term --cov-report=xml
```

is an intermittent segfault in PyPy that we're monitoring now. I restarted the jobs and it will probably pass. (Obviously, this isn't a good situation for anyone using PyPy; we need to figure out what's causing the segfault, at least because there's a slight chance it's also a silent undefined behavior in CPython.)

jpivarski (approved):
Thanks! (This is un-reverting c13c091eb2d438a8de4792ebd190cbab1f8137ec.)

This will probably pass tests before the awkward-cpp deployment is done; I'll merge it to try out the updated deploy.yml when deploying awkward 2.5.2. It looks like the most substantial change is in deploy-cpp.yml, though.

If (after all of the new-release dust settles) I run deploy-cpp.yml _without_ "deploy to PyPI", will it be a sufficient test of this fix? If so, I'll do that, so that it won't be a surprise the next time we need an awkward-cpp.

jpivarski:
When https://github.com/scikit-hep/awkward/actions/runs/7505629752 is done, I'll merge this.

jpivarski:
Hmm. [That didn't do it](https://github.com/scikit-hep/awkward/actions/runs/7506174989/job/20437001303) because there's an asterisk in the name, where I think you wanted it in the pattern only:

```
Run actions/upload-artifact@v4
  with:
    name: awkward-wheels-ubuntu-latest-auto64-pp*
    path: wheelhouse/*.whl
    if-no-files-found: warn
    compression-level: 6
  env:
    SOURCE_DATE_EPOCH: 1705076015
    pythonLocation: /opt/hostedtoolcache/Python/3.11.7/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.7/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.7/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.7/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.7/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.7/x64/lib
With the provided path, there will be 3 files uploaded
Error: The artifact name is not valid: awkward-wheels-ubuntu-latest-auto64-pp*. Contains the following character:  Asterisk *
          
Invalid characters include:  Double quote ", Colon :, Less than <, Greater than >, Vertical bar |, Asterisk *, Question mark ?, Carriage return \r, Line feed \n, Backslash \, Forward slash /
          
These characters are not allowed in the artifact name due to limitations with certain file systems such as NTFS. To maintain file system agnostic behavior, these characters are intentionally not allowed to prevent potential problems with downloads on different file systems.
```

jpivarski:
From the name `awkward-wheels-ubuntu-latest-auto64-pp*`, it must be this one:

https://github.com/scikit-hep/awkward/blob/a67c0ed4dfa2832c403337fbe18cf88561dcf12f/.github/workflows/deploy-cpp.yml#L112-L116

It's picking up this `"pp*"` as the build name:

https://github.com/scikit-hep/awkward/blob/a67c0ed4dfa2832c403337fbe18cf88561dcf12f/.github/workflows/deploy-cpp.yml#L66

rather than the file or directory that it matches to.

But this should have the same multiplicity as the number of CI tasks, not the number of files in the directory that the wildcard matches to, right? So in that case, it could be fixed by (a) stripping off the `"*"` using string manipulation or (b) not using a wildcard in the definition of the `matrix.build` variable, but adding a wildcard after its use in defining this environment variable:

https://github.com/scikit-hep/awkward/blob/a67c0ed4dfa2832c403337fbe18cf88561dcf12f/.github/workflows/deploy-cpp.yml#L95-L97

(Both values for `matrix.build` have a wildcard in the same place.)

jpivarski:
I'll implement that in a non-PR commit, and then test it in deploy-cpp.yml now: 327b1bf1fa50ec74414ea191db88903e49a15bfc.

jpivarski:
So far, I think that's working (https://github.com/scikit-hep/awkward/actions/runs/7506459981), but I'll let it run all the way through to be sure.

**Edit:** It's done, and it's fine.

agoose77:
Thanks Jim! I noticed the failure on my phone, and I appreciate you fixing it first! Rebuilding the `CIBW_BUILD` variable looks fine to me — easy to change this down the road.

PR #2950, opened 2024-01-16
fix: prepare for NumPy 2

agoose77:
To build this PR I ran the ruff linter, and found a case of a removal in 2.0. I would not be surprised if there are usages that the linter can't find, so we'll only be confident once the preview wheels are out.


agoose77:
partly replaced by #3010

PR #2956, opened 2024-01-17
test: don't fail fast (so pypy doesn't stop the tests)

jpivarski:


ManasviGoyal (approved):
All CPU tests are running now. Looks good!

agoose77:
@jpivarski originally we skipped the tests if awkward-cpp had not changed. Is there a reason that this pr changes that?

jpivarski:
Hmmm. Maybe we'll go back to that. I see the logic that if the YAML specification doesn't change and the CPU implementations (in awkward-cpp) don't change, there's no reason to re-run their tests. While @ManasviGoyal is fixing the dev/generate-tests.py, the tests do change, so they need to run in CI. (Ideally, the `if` condition should check for changes in both awkward-cpp and this dev script.)

It's also likely to undergo a lot of changes in the near future as she adds cuda-kernel tests. Naturally, that motivates some fixes to the testing framework itself.

agoose77:
Oh, I hadn't realised those tests were changing. We can perhaps extend the ci to handle that!

PR #2958, opened 2024-01-18
chore: add missing contributors

henryiii:
This is a list of contributors missing, based on `yarn all-contributors check`.


jpivarski:
Thanks! But unfortunately, I've already made this obsolete. Also, I noticed that not everybody contributed `code`. I have to look up their contributions anyway, so I can continue doing it manually.

PR #2967, opened 2024-01-19
feat: decorator to validate and annotate dependencies

### LINKED ISSUE #2952: Dependencies of `from_parquet` should be included or made explicit in docs
### Which documentation?

Other (please explain)?

When setting up a fresh installation of `awkward@2.5.2` (and earlier) and trying to run

```python
import awkward as ak

_ = ak.from_parquet("foo.parquet")
```

the dependencies `pyarrow`, `fsspec` and `pandas` have to be installed by hand first for this minimal example to work. As far as I see this is not in the documentation. If this is the desired behaviour to keep the footprint of `awkward` small if the user does not want to use this function, I suggest to add a section to the docs explaining the necessary dependencies.
If this is not the desired behaviour it would be convenient to add the mentioned dependencies in the `pyproject.toml`.

Thanks for considering!

### LINKED ISSUE #2953: Add new argument to `high_level_function` decorator that specifies dependencies

- Add declarative dependency specification at `high_level_function`
  - Declare versions & dependency group
  - Annotate docstring with this information
  - Check package versions once

agoose77:
Closes #2953 

This PR adds a new `dependencies` argument to `high_level_function`. This let's us fail at runtime if dependency specifications are not met, and also to annotate our documentation to inform the user ahead-of-time.

We will still need to keep these values in sync with our new `pyproject.toml` groups, but we can make a helper for that if needs be.

Examples:
image

image

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -42,6 +42,7 @@ run_cuda_notebooks = os.environ.get("DOCS_RUN_CUDA", False)
 extensions = [
     "sphinx_copybutton",
     "sphinx_design",
+    "sphinx_togglebutton",
     "sphinx_external_toc",
     "sphinx.ext.intersphinx",
     "myst_nb",
--- a/docs/prepare_docstrings.py
+++ b/docs/prepare_docstrings.py
@@ -395,55 +395,55 @@ for filename in glob.glob("../src/awkward/**/*.py", recursive=True):
         if isinstance(toplevel, ast.FunctionDef):
             dofunction(link, linelink, shortname, toplevel.name, toplevel)
 
-#
-# def test_signature_pos_or_kw():
-#     mod = ast.parse(
-#         """
-# def func(x, z):
-#     ...
-# """
-#     )
-#     node = mod.body[0]
-#     assert dosig(node) == "x, z"
-#
-#
-# def test_signature_kwarg():
-#     mod = ast.parse(
-#         """
-# def func(x, z=None):
-#     ...
-# """
-#     )
-#     node = mod.body[0]
-#     assert dosig(node) == "x, z=None"
-#
-#
-# def test_signature_vararg():
-#     mod = ast.parse(
-#         """
-# def func(x, *y, z=None):
-#     ...
-# """
-#     )
-#     node = mod.body[0]
-#     assert dosig(node) == "x, *y, z=None"
-#
-#
-# def test_signature_kwonly():
-#     mod = ast.parse(
-#         """
-# def func(x, *, y, z=None):
-#     ...
-# """
-#     )
-#     node = mod.body[0]
-#     assert dosig(node) == "x, *, y, z=None"
-#
-#
-# test_signature_pos_or_kw()
-# test_signature_kwarg()
-# test_signature_vararg()
-# test_signature_kwonly()
+
+def test_signature_pos_or_kw():
+    mod = ast.parse(
+        """
+def func(x, z):
+    ...
+"""
+    )
+    node = mod.body[0]
+    assert dosig(node) == "x, z"
+
+
+def test_signature_kwarg():
+    mod = ast.parse(
+        """
+def func(x, z=None):
+    ...
+"""
+    )
+    node = mod.body[0]
+    assert dosig(node) == "x, z=None"
+
+
+def test_signature_vararg():
+    mod = ast.parse(
+        """
+def func(x, *y, z=None):
+    ...
+"""
+    )
+    node = mod.body[0]
+    assert dosig(node) == "x, *y, z=None"
+
+
+def test_signature_kwonly():
+    mod = ast.parse(
+        """
+def func(x, *, y, z=None):
+    ...
+"""
+    )
+    node = mod.body[0]
+    assert dosig(node) == "x, *, y, z=None"
+
+
+test_signature_pos_or_kw()
+test_signature_kwarg()
+test_signature_vararg()
+test_signature_kwonly()
 
 
 toctree_path = reference_path / "toctree.txt"
--- a/docs/prepare_docstrings.py
+++ b/docs/prepare_docstrings.py
@@ -330,6 +330,8 @@ def dofunction(link, linelink, shortname, name, astfcn):
 
         outfile.write(
             ".. note::\n"
+            "   :class: dropdown\n"
+            "\n"
             "   This function requires the following optional third-party libraries:\n"
             "\n"
         )
@@ -339,7 +341,7 @@ def dofunction(link, linelink, shortname, name, astfcn):
             "\n\n"
             f"   If you use pip, you can install these packages with "
             f"``python -m pip install {install_dependencies_string}``.\n"
-            "    Otherwise, if you use Conda, install the corresponding packages "
+            "   Otherwise, if you use Conda, install the corresponding packages "
             "for the correct versions. "
         )
         if extras:
--- a/docs/prepare_docstrings.py
+++ b/docs/prepare_docstrings.py
@@ -2,13 +2,13 @@
 
 from __future__ import annotations
 
-import re
 import ast
 import glob
 import io
-import subprocess
-import pathlib
 import os
+import pathlib
+import re
+import subprocess
 import warnings
 
 import sphinx.ext.napoleon
@@ -20,7 +20,7 @@ output_path = reference_path / "generated"
 output_path.mkdir(exist_ok=True)
 
 latest_commit = (
-    subprocess.run(["git", "rev-parse", "HEAD"], stdout=subprocess.PIPE)
+    subprocess.run(["git", "rev-parse", "HEAD"], stdout=subprocess.PIPE, check=False)
     .stdout.decode("utf-8")
     .strip()
 )
@@ -48,16 +48,16 @@ def tostr(node):
         return tostr(node.value) + "." + node.attr
 
     elif isinstance(node, ast.Subscript):
-        return "{0}[{1}]".format(tostr(node.value), tostr(node.slice))
+        return f"{tostr(node.value)}[{tostr(node.slice)}]"
 
     elif isinstance(node, ast.Slice):
         start = "" if node.lower is None else tostr(node.lower)
         stop = "" if node.upper is None else tostr(node.upper)
         step = "" if node.step is None else tostr(node.step)
         if step == "":
-            return "{0}:{1}".format(start, stop)
+            return f"{start}:{stop}"
         else:
-            return "{0}:{1}:{2}".format(start, stop, step)
+            return f"{start}:{stop}:{step}"
 
     elif isinstance(node, ast.Call):
         return "{0}({1})".format(
@@ -78,10 +78,7 @@ def tostr(node):
 
     elif isinstance(node, ast.Dict):
         return "{{{0}}}".format(
-            ", ".join(
-                "{0}: {1}".format(tostr(x), tostr(y))
-                for x, y in zip(node.keys, node.values)
-            )
+            ", ".join(f"{tostr(x)}: {tostr(y)}" for x, y in zip(node.keys, node.values))
         )
 
     elif isinstance(node, ast.Lambda):
@@ -165,7 +162,7 @@ def dodoc(docstring, qualname, names):
     out = re.sub(r"<<<([^>]*)>>>", r"`\1`_", out)
     out = re.sub(r"#(ak\.[A-Za-z0-9_\.]*[A-Za-z0-9_])", r":py:obj:`\1`", out)
     for x in names:
-        out = out.replace("#" + x, ":py:meth:`{1} <{0}.{1}>`".format(qualname, x))
+        out = out.replace("#" + x, f":py:meth:`{x} <{qualname}.{x}>`")
     out = re.sub(r"\[([^\]]*)\]\(([^\)]*)\)", r"`\1 <\2>`__", out)
     out = str(sphinx.ext.napoleon.GoogleDocstring(out, config))
     out = re.sub(
@@ -212,8 +209,8 @@ def doclass(link, linelink, shortname, name, astcls):
     outfile = io.StringIO()
     outfile.write(qualname + "\n" + "-" * len(qualname) + "\n\n")
     outfile.write(f".. py:module: {qualname}\n\n")
-    outfile.write("Defined in {0}{1}.\n\n".format(link, linelink))
-    outfile.write(".. py:class:: {0}({1})\n\n".format(qualname, dosig(init)))
+    outfile.write(f"Defined in {link}{linelink}.\n\n")
+    outfile.write(f".. py:class:: {qualname}({dosig(init)})\n\n")
 
     docstring = ast.get_docstring(astcls)
     if docstring is not None:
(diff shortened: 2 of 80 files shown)


jpivarski (commented):
This is really nice!

But what is meant by "extras" in this case? Is it `pip install awkward[extras]`? That would likely be obscure to most users.

I think it's useful for both the documentation and the exception message to include sample pip-install and conda-install text that can be copy-pasted into a terminal to actually install the necessary packages. There are situations in which I'll deliberately do something wrong if I know that the error message has what I need to do in a copy-pasteable form—it beats typing it manually or remembering its spelling.

I don't think any of these libraries will need a mapping between PyPI name and conda name, but if such a thing becomes necessary in the future, it can be a dict. (Uproot optionally requires `xxhash`, which is `python-xxhash` in conda-forge for disambiguation.)

----------------

I made a new environment without the optional dependencies to test it out.

```python
>>> ak.to_arrow(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
```

```
ImportError: to use to_arrow, you must install pyarrow:

    pip install pyarrow

or

    conda install -c conda-forge pyarrow


    ak.to_arrow(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
    )
```

```python
>>> ak.to_jax(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
```

```
ImportError: This function has the following dependency requirements that are not met by your current environment:

    * jax — you do not have this package
    * jaxlib — you do not have this package



    ak.to_jax(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
    )
```

```python
>>> ak.to_parquet("/tmp/whatever.parquet", ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
```

```
ImportError: This function has the following dependency requirements that are not met by your current environment:

    * pyarrow>=7.0.0 — you do not have this package
    * fsspec — you do not have this package

You can fix this error by installing these packages directly, or install awkward with all of the following extras:

    * arrow

    ak.to_parquet(
        '/tmp/whatever.parquet'
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
    )
```

```python
>>> ak.to_backend(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]), "cuda")
```

```
ModuleNotFoundError: to use Awkward Arrays with CUDA, you must install cupy:

    pip install cupy

or

    conda install -c conda-forge cupy


    ak.to_backend(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
        'cuda'
    )
```

It might be better to exclude the "This error occurred while calling" for failures due to missing libraries, so that the message about the missing libraries is last. Otherwise, someone might spend a lot of time looking at the arguments they passed to the function, looking for errors, but the more relevant message is right above.

I also tried installing the wrong version of Arrow, to see what the message would look like:

```python
>>> ak.to_arrow(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
```

```
ImportError: pyarrow 7.0.0 or later required for to_arrow

    ak.to_arrow(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
    )
```

It doesn't seem to be reaching

```python
f"    * {req} — you have {ver} installed"
```

for that case.

agoose77:
@jpivarski I did another pass! I can't reproduce your arrow bug, perhaps you could interrogate what `importlib.metadata.version` shows for `pyarrow`?

Let me know what you think of the docs deployment.

agoose77:
I've added `pyarrow` as a dependency for all string functions. We can go finer grained now than just "pyarrow>=7.0.0", but this is a start.

I've made the docs `note` directive collapsed by default.

agoose77:
I noticed that the docs environment isn't solving properly. I suspect that the dependencies are not pinned in Sphinx. The proper fix is to update our sphinx usage, but for now let's try just pinning everything hard.

Also, I need to update the decorators of more functions as-per your recent comment in the issue tracker.

agoose77:
The new errors are caused by pathways that don't technically require e.g. fsspec hitting this new decorator.

As I see it, we have two options: 
- enforce dependencies at the function scope (e.g. there is no way to call `to_json` without needing fsspec)
- set dependencies at the function scope, enforce requirement in branches (see below)

```python

@high_level_function(name="ak.to_json", dependencies=["fsspec>1.0.0"])
def to_json(...):
    if FOO:
        # Ensures that we get fsspec>1.0.0
        fsspec = require("fsspec")
    else:
        # no fsspec
        pass

```

@jpivarski any preferences?

jpivarski:
> As I see it, we have two options:
> 
> 1. enforce dependencies at the function scope (e.g. there is no way to call `to_json` without needing fsspec)
> 2. set dependencies at the function scope, enforce requirement in branches (see below)

Option 2. The documentation can say, without qualification, that `to_json` needs fsspec to run and if given a remote URL, `to_json` can say, without qualification, that `to_json` needs fsspec to run, but if given a local file, `to_json` can just do it whether it has fsspec or not. It's a little lie, but a lie in a convenient direction. Explaining the truth in detail would only cloud the issue: this situation is similar to the "easier to ask forgiveness than permission" situation... sort of.

jpivarski:
Will the packaging tests pass if we update the branch? (Not including pypy; that one is fixed by just running it multiple times, but it's not required.)

agoose77:
@jpivarski probably not. I imagine it's because the dependencies are now enforced at the function scope, not at the branch level. The changes we've agreed on (2) will fix this!

jpivarski:
In an environment that has all of these things installed (i.e. my JAX works, I can make a JAX array), I get

```python
>>> ak.to_jax(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
```

```
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 461, in wrapper
    return func(*args, **kwargs)
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 211, in dispatch
    on_dispatch_internal()
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 175, in validate_runtime_dependencies
    raise exception
ImportError: This function has the following dependency requirements that are not met by your current environment:

    * jaxlib — you have 0.4.23.dev20231222 installed

If you use pip, you can install these packages with `python -m pip install jaxlib`.
Otherwise, if you use Conda, install the corresponding packages for the correct versions. 

  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 460, in wrapper
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 85, in __exit__
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 95, in handle_exception
ImportError: This function has the following dependency requirements that are not met by your current environment:

    * jaxlib — you have 0.4.23.dev20231222 installed

If you use pip, you can install these packages with `python -m pip install jaxlib`.
Otherwise, if you use Conda, install the corresponding packages for the correct versions. 

    ak.to_jax(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
    )
```

In ak_to_jax.py, there's no version requirement on JAX. I have noticed that the CUDA part of my JAX installation has broken:

```python
>>> import jax
>>> a = jax.numpy.arange(10)
An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
>>> 
```

That looks like a print statement, not a warning, so I don't think Awkward is detecting it.

Here's my JAX version and source:

```bash
% mamba list jax
# packages in environment at /home/jpivarski/mambaforge:
#
# Name                    Version                   Build  Channel
jax                       0.4.23             pyhd8ed1ab_0    conda-forge
jaxlib                    0.4.23          cpu_py310h9ed8a0c_0    conda-forge
```

(That is, my JAX is not 0.4.23.dev20231222 but 0.4.23.)

All of the other optional library checks seem to be working fine. (I'm going to test an environment that _doesn't_ have them installed now.)

jpivarski:
In an environment without dependencies, this is what I'm seeing:

```python
>>> ak.to_arrow(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]))
```

```
  File "/home/jpivarski/mambaforge/envs/testy/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 411, in from_name
    return next(iter(cls.discover(name=name)))
StopIteration

During handling of the above exception, another exception occurred:

  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 74, in iter_missing_dependencies
    _version = importlib_metadata.version(requirement.name)
  File "/home/jpivarski/mambaforge/envs/testy/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 924, in version
    return distribution(distribution_name).version
  File "/home/jpivarski/mambaforge/envs/testy/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 897, in distribution
    return Distribution.from_name(distribution_name)
  File "/home/jpivarski/mambaforge/envs/testy/lib/python3.10/site-packages/importlib_metadata/__init__.py", line 413, in from_name
    raise PackageNotFoundError(name)
importlib_metadata.PackageNotFoundError: No package metadata was found for pyarrow

During handling of the above exception, another exception occurred:

  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 461, in wrapper
    return func(*args, **kwargs)
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 211, in dispatch
    on_dispatch_internal()
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 171, in validate_runtime_dependencies
    exception = build_runtime_dependency_validation_error(dependency_spec)
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 115, in build_runtime_dependency_validation_error
    extra_missing_dependencies = [
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 77, in iter_missing_dependencies
    mod = importlib.import_module(requirement.name)
  File "/home/jpivarski/mambaforge/envs/testy/lib/python3.10/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'pyarrow'

  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 460, in wrapper
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 85, in __exit__
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 95, in handle_exception
ModuleNotFoundError: No module named 'pyarrow'

    ak.to_arrow(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
    )
```

I.e. no message about how this function requires pyarrow and how to install it.

Same for JAX: no message about the fact that `ak.to_jax` requires it and how to install JAX.

But the one for CuPy works:

```python
>>> ak.to_backend(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]), "cuda")
```

```
...

  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 460, in wrapper
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 85, in __exit__
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 95, in handle_exception
ModuleNotFoundError: to use Awkward Arrays with CUDA, you must install cupy:

    pip install cupy

or

    conda install -c conda-forge cupy


    ak.to_backend(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
        'cuda'
    )
```

Also, I noticed that this is following Option 1: calling `ak.to_json` (to write a local file) without fsspec raises an error. Interestingly, that error doesn't indicate that fsspec is required and how to install it, so that might be helpful in figuring out which path this is taking.

```python
>>> ak.to_json(ak.Array([[1.1, 2.2, 3.3], [], [4.4, 5.5]]), "/tmp/whatever.json")
```

```
...

  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 460, in wrapper
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 85, in __exit__
  File "/home/jpivarski/irishep/awkward/src/awkward/_errors.py", line 95, in handle_exception
ModuleNotFoundError: No module named 'fsspec'

    ak.to_json(
        <Array [[1.1, 2.2, 3.3], [], [4.4, 5.5]] type='3 * var * float64'>
        '/tmp/whatever.json'
    )
```

It doesn't even get to the point of checking the arguments; I can use no arguments and get this error message; it's checking before anything else.

agoose77:
I've just pushed a WIP that tries this new approach — functions establish a dependency context, which is evaluated when the function implementation calls `import_required_module`.

I've not yet finished (or fixed the docs component to reflect the new declaration style).

Some points:
1. I'm not sure whether the repeated dependency specs are good or bad. We could write some of this in our pyproject.toml for _all_ functions.
   - I don't think this is worth it — we'd still want to declare _which_ of these dependencies each function uses. Inspecting for `import_required_module` would be too fragile imo.
2. Now we're using a `requires` decorator to enable multiple argument passing. I don't know if I prefer
   ```python
   @high_level_function(
       dependencies=[
           requires("fsspec", group="arrow"),
           requires("pyarrow>=7.0.0", group="arrow", module_name="arrow"),
       ]
   )
   ```
   or
   ```python
   @requires("pyarrow>=7.0.0", group="arrow", module_name="arrow")
   @requires("fsspec", group="arrow")
   @high_level_function()
   ```

jpivarski:
I'm confused about why the interface needs all of this structure. A high-level function (possibly) needs libraries `x`, `y`, and `z` to run; these are the libraries it might encounter in some if-branch of the implementation, but not necessarily. (Also, we only list non-strict dependencies. fsspec will [become a strict dependency in 2.6.0](https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20), which is due on February 1.)

If there are dependencies, it tries to run the high-level function, catches any ImportError (including ModuleNotFoundError), and re-raises it with a note about how to install the set of dependencies.

In principle, the PyPI or conda install name of the package might be different from the import name of the package, but we haven't encountered any like that. If there is such a distinction, then it would be global: the install-import mapping does not depend on which high-level function we're running, so it can be a private dict somewhere.

That's my understanding of the problem; the solution so far is a lot more complicated than that.

agoose77:
> If there are dependencies, it tries to run the high-level function, catches any ImportError (including ModuleNotFoundError), and re-raises it with a note about how to install the set of dependencies.

The short answer is versions matter. We also want to error if the user has e.g. `cppyy` but the wrong version. Rather than having two sets of dependency information scattered through the codebase, the intention here is that we define it in a single place for each function.

> In principle, the PyPI or conda install name of the package might be different from the import name of the package, but we haven't encountered any like that. If there is such a distinction, then it would be global: the install-import mapping does not depend on which high-level function we're running, so it can be a private dict somewhere.

We need to embed this information somewhere: already, `ROOT` isn't on PyPI, and has a different (lower-case) name. Arrow is "pyarrow" rather than "arrow". I chose to embed this in a single place so that we don't have to handle this information in multiple related places. Instead, it's just defined many times. This makes the docs generator simpler (stateless), too. 

We could define the "distribution-level" metadata somewhere such that most definitions can use the short-hand, i.e. a table mapping distribution to conda package and python import name (we only need distribution → python name for the case that it's not installed, otherwise `importlib.metadata` has that table). Maybe this table would also define the base version, e.g.
```python
@requires_global("arrow", "fsspec", group="arrow")
@high_level_function(requires)
```

There's more to think about here — I'll come back to it :)

I feel that there's an opportunity here to clean up / centralise our "extras" handling (including how `ak._connect` works), but we don't want too much complexity.

agoose77:
Let's see if: 
1. we can move to extras as a mean of defining dependencies for function classes (e.g. `str`/`arrow`, `dataframe`, etc.)
2. we can use "scoped extras" to define per-function dependencies e.g. if one func needs `pyarrow>13`
3. we can tell users the `pip` command
4. we can detect the users' conda binary, and _suggest_ a possible command -- with appropriate warning about packages
5. we can tell users a detailed info of the distributions required
6. we can list the PyPI page on the documentation
7. extras are scoped to the function, not branch

The `requirement()` function can be simple; import errors just tell users to install extra / meet dependencies. Version errors can be scoped to the package itself.

agoose77:
Let's close this - I don't have bandwidth for it right now, and it's adding to the noise :)

PR #2977, opened 2024-01-22
fix: use `std::tuple_element_t` as an `OptionType` type alias

ianna:
`std::tuple_element_t` was missing on some of the CI nodes. Check if this has changed. Plus, this PR should trigger all cpp tests to be run.

### DIFF
--- a/header-only/builder-options/awkward/BuilderOptions.h
+++ b/header-only/builder-options/awkward/BuilderOptions.h
@@ -22,15 +22,8 @@ namespace awkward {
 
     using OptionsPack = typename std::tuple<OPTIONS...>;
 
-    // FIXME:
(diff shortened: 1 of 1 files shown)


ianna (commented):
@jpivarski - I think, this is ok to remove the FIXME - since we've moved to C++17 by default. I do not see any CI jobs failing, but it is possible that it is because we do not run the cpp tests on all architectures.

jpivarski (approved):
Yes, `std::tuple_element_t` is [since C++14](https://en.cppreference.com/w/cpp/utility/tuple_element) and we require C++17:

https://github.com/scikit-hep/awkward/blob/4571af99aa2247d8550904ea042b52f977ca605c/awkward-cpp/CMakeLists.txt#L44

So please go ahead and merge this when you see fit. Thanks!

PR #2979, opened 2024-01-22
feat: add to_parquet_row_groups

zbilodea:


src/awkward/operations/ak_to_parquet.py
```diff
@@ -196,16 +252,28 @@ def to_parquet(
         primitive_policy="error",
         string_policy="as-characters",
     )
```
jpivarski:
```python
layout = ak.operations.ak_to_layout._impl(data, ...)
```

will convert `data` into an Awkward Array, and Awkward Arrays are eagerly in memory all at once. `ak.to_parquet_row_groups` is supposed to take an iterator as `data`, so this will iterate over it, pulling it all into memory, which is not what you want.

In the case of `ak.to_parquet_row_groups`, you'll want to do a Python for loop over `data` and call `ak_to_layout` on each item individually.

src/awkward/operations/ak_to_parquet.py
```diff
@@ -178,6 +178,62 @@ def to_parquet(
...
+        parquet_page_version,
+        parquet_metadata_statistics,
+        parquet_dictionary_encoding,
+        parquet_byte_stream_split,
+        parquet_coerce_timestamps,
+        parquet_old_int96_timestamps,
+        parquet_compliant_nested,  # https://issues.apache.org/jira/browse/ARROW-16348
+        parquet_extra_options,
+        storage_options,
+        iter=False,
```
jpivarski: `iter` is a builtin Python function name, one that you might want to use in the implementation (if it doesn't fit well into a single `for` loop). By using it as the boolean back-door, you lose access to that function. (You can get it back with `__builtins__.iter`, but that's deliberately making things difficult.)

```python
>>> def generator():
...     for _ in range(5):
...         print("making another one")
...         yield "another one"
... 
>>> it = iter(generator())
>>> next(it)
making another one
'another one'
>>> next(it)
making another one
'another one'
>>> next(it)
making another one
'another one'
>>> next(it)
making another one
'another one'
>>> next(it)
making another one
'another one'
>>> next(it)
StopIteration
```

jpivarski (changes requested):
You might not need to define the `batch_iterator` function. (Is it a copy of what was in Awkward 1.x? This is a different context.) You're already iterating over batches in ak_to_parquet.py lines 408‒435. I haven't looked closely enough at it to point out whether something is superfluous, but it looks like it might be. It's worth focusing on that a bit more.

I wish there was a way to write a test that would check to see that all the batches are not loaded into memory at once. Maybe you could try it on a manual test, outside of CI, with a dataset that is too large to fit in memory at once. Putting (temporary!) print statements in some places would also let you check the temporal order of things.

src/awkward/operations/ak_to_parquet.py
zbilodea: Okay! Now calling to_layout for each iteration - I printed each new layout to check that they only have data from one array and it looks good, but I'll test it on a large dataset to be sure

zbilodea: I made a test and was able to write file larger than my laptop's memory!

Screenshot 2024-01-29 at 11 54 11

jpivarski: Also, the output file is compressed, whereas the data in memory were not compressed.

src/awkward/operations/ak_to_parquet.py
```diff
@@ -178,35 +178,114 @@ def to_parquet(
...
+            string_to32,
+            bytestring_to32,
+            emptyarray_to,
+            categorical_as_dictionary,
+            extensionarray,
+            count_nulls,
+        )
+    else:
+        layout = ak.operations.ak_to_layout._impl(
+            next(data),
```
jpivarski: Python makes a distinction between an iterator (something that you can call `next` on to get another item) and an iterable (something that you can call `iter` on to get an iterator). The interface should work for iterables, so `iter(data)` will be needed at some point.

jpivarski (approved):
This is good in that it works, but the code is a bit repetitive and the control flow is hard to follow. I'll just make a few edits after this review: don't merge until I'm done.

jpivarski:
This refactoring reduces three repetitions of `layout = ...` followed by `table = ...` into one by defining a temporary function, and it simplifies the `while ... try ... except StopIteration` logic with a `for` loop. It also adds guards to prevent users from making some likely mistakes.

The `for` loop takes advantage of the distinction between iterables (such as lists, array, and iterators) and iterators: when the first items in an iterator have been consumed with `next`, a `for` loop over the iterator does not include the already-consumed items—Python does not go back to the beginning.

```python
>>> some_list = ["one", "two", "three", "four", "five", "six", "seven", "eight", "nine", "ten"]
>>> next(some_list)
TypeError: 'list' object is not an iterator
```

```python
>>> iterator = iter(some_list)
>>> next(iterator)
'one'
>>> next(iterator)
'two'
>>> next(iterator)
'three'
>>> for item in iterator:
...     print(item)
... 
four
five
six
seven
eight
nine
ten
```

but

```python
>>> for item in some_list:
...     print(item)
... 
one
two
three
four
five
six
seven
eight
nine
ten
```

I think this PR is ready for merging, after doing an "Update branch." Look it over and decide if you want to make any counter-edits. If so, ping me again for another review. If not, go ahead and merge it.

src/awkward/operations/ak_to_parquet.py
```diff
@@ -415,38 +431,24 @@ def parquet_columns(specifier, only=None):
...
+
+                # this `data` is an iterator (not iterable), starting AFTER the first array
+                # a `for` loop implicitly calls `next` and stops at `StopIteration`
+                for item in data:
+                    layout, table = get_layout_and_table(item)
+                    writer.write_table(table, row_group_size=row_group_size)
+
+        finally:
+            # ensure that the ParquetWriter is closed for iterative and non-iterative cases
+            writer.close()
```
jpivarski: This is the main thing that I wanted to simplify, for future maintainability.

PR #2981, opened 2024-01-22
fix: delete `header-only/tests/test_1542-array-builder.cpp`

### LINKED ISSUE #1637: ci: run c++ tests in an integration build

There are four c++ tests in `tests-cpp` that are built, but not run. They require C++17.
```[tasklist]
### Tasks
- [ ] https://github.com/scikit-hep/awkward/pull/2978
- [ ] https://github.com/scikit-hep/awkward/pull/2976
- [ ] https://github.com/scikit-hep/awkward/pull/2977
```

ianna:
as discussed in https://github.com/scikit-hep/awkward/issues/1637

fixes #1637


jpivarski (approved):
Great, thanks!

Since this is not going to need any edits, I'll merge it now.

jpivarski:
That pypy segfault seems to be more frequent now...

PR #2986, opened 2024-01-25
docs: update dependencies

### LINKED ISSUE #2975: Fix docs toolchain
Right now we'll pin it, but we should upgrade to sphinx 7.

agoose77:
Closes #2975

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -3,7 +3,7 @@ jupyterlite-pyodide-kernel==0.1.0
 jupyterlite[piplite]==0.1.0
 jupyterlite-sphinx==0.8.0
 
-sphinx>=4.5.0,<5.0.0
+sphinx~=7.0
 PyYAML
 black
 pycparser
@@ -12,7 +12,8 @@ sphinx-copybutton
 sphinx-design
 sphinx-sitemap
 pydata-sphinx-theme
-myst-nb
+myst-nb~=1.0
+myst-parser~=2.0
 sphinx-external-toc
 ipyleaflet
 
@@ -27,6 +28,5 @@ h5py
 matplotlib
 awkward
 uproot
-uproot3
 jax>=0.2.7;python_version>="3.6" and sys_platform != "win32"
 jaxlib>=0.1.57;python_version>="3.6" and sys_platform != "win32"


docs/requirements.txt
```diff
@@ -1,13 +1,15 @@
...
-jupyterlite-pyodide-kernel==0.1.0
-jupyterlite[piplite]==0.1.0
-jupyterlite-sphinx==0.8.0
+# This kernel requires a specific pyodide version (see pyodide-build)
+jupyterlite-pyodide-kernel==0.2.*
+jupyterlite[piplite]==0.2.*
+jupyterlite-sphinx==0.11.*
 
-sphinx>=4.5.0,<5.0.0
+sphinx==7.*
```
jpivarski: Pin to a version? The [admonitions](https://iscinumpy.dev/post/bound-version-constraints/) against `<` must apply to `=` as well[^1]; is there a reason to think that Sphinx 8 would be bad?

[^1]: If they don't, then I don't understand the issue.

docs/requirements-wasm.txt
```diff
@@ -1,2 +1,2 @@
-pyodide-build==0.23.3
-pydantic<2
+# This determines the pyodide version
+pyodide-build==0.24.1
```
jpivarski: I can be convinced that an exact version of Pyodide is needed, since extensions are vendored into it.

jpivarski (approved):
I'm in favor of any docs-tech updates that you think are appropriate.

docs/requirements.txt
```diff
@@ -1,13 +1,15 @@
 # JupyterLite (kernel requires specific pyodide version)
-jupyterlite-pyodide-kernel==0.1.0
-jupyterlite[piplite]==0.1.0
-jupyterlite-sphinx==0.8.0
+# This kernel requires a specific pyodide version (see pyodide-build)
+jupyterlite-pyodide-kernel==0.2.*
+jupyterlite[piplite]==0.2.*
+jupyterlite-sphinx==0.11.*
```
jpivarski: Same argument as for Pyodide.

docs/requirements.txt
jpivarski: > Right now we'll pin it, but we should upgrade to sphinx 7.

Okay!

agoose77: For application-like uses, we can be very restrictive e.g. pin exactly. I'm not in favour of going _that_ far; we just want our docs to break less often.

docs/requirements.txt
agoose77: My assumption is that jupyterlite-pyodide-kernel will receive a minor bump if the pyodide version changes.

agoose77:
Looks like pyodide broke the emscripten toolchain bundling in this version, but jupyterlite doesn't have a kernel that supports the newer version yet. I'll update this PR to upgrade only Sphinx

agoose77:
This needs more work; I'm still recovering my development environment, but will finalise this PR soon.

jpivarski:
I decided to postpone it for after 2.6.0. I merged what was mergeable, and have now started the awkward-cpp build.

agoose77:
I've changed this PR to ignore the jupyterlite changes, and just touch the sphinx parts. Let's see what happens.

PR #2995, opened 2024-01-30
fix: `ak.merge_union_of_records` error message for cases that can't use `axis < 0`

### LINKED ISSUE #2994: ak.merge_union_of_records cannot merge two NanoAOD records?
### Version of Awkward Array

2.5.2

```python3
import awkward as ak
import uproot

file1 = uproot.dask("./61B08DB4-292E-734A-B082-E56243B5365A.root:Events", ak_add_doc=True, open_files=False)
file2 = uproot.dask("./E0416E7B-3364-D64D-910A-E83496BCD1EF.root:Events", ak_add_doc=True, open_files=False)

lza_1 = file1._meta.layout.form.length_zero_array()
lza_2 = file2._meta.layout.form.length_zero_array()

merged = ak.merge_union_of_records(
    ak.concatenate([lza_1, lza_2])
)
```

results in:
```            
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_dispatch.py", line 62, in dispatch
    next(gen_or_result)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_merge_union_of_records.py", line 60, in merge_union_of_records
    return _impl(array, axis, highlevel, behavior, attrs)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_merge_union_of_records.py", line 268, in _impl
    out = ak._do.recursively_apply(layout, apply)
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_do.py", line 36, in recursively_apply
    return layout._recursively_apply(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/contents/unionarray.py", line 1578, in _recursively_apply
    result = action(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/operations/ak_merge_union_of_records.py", line 167, in apply
    if depth < posaxis + 1 and layout.is_leaf:
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'

  File "minimal_repro_merged_union.py", line 10, in <module>
    merged = ak.merge_union_of_records(
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_dispatch.py", line 70, in dispatch
    return gen_or_result
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_errors.py", line 85, in __exit__
  File "/Users/lgray/miniforge3/envs/coffea-dev/lib/python3.8/site-packages/awkward/_errors.py", line 95, in handle_exception
TypeError: unsupported operand type(s) for +: 'NoneType' and 'int'

    ak.merge_union_of_records(
        <Array [] type='0 * union[struct[{run: uint32[parameters={"__doc__"...'>
    )
```

I have sent the requisite files to @jpivarski since they're internal CMS data.

jpivarski:
`maybe_posaxis` can return `None` if the `layout`'s depth branches, and that was unguarded in `ak.merge_union_of_records`. (It might be unguarded in other functions; I didn't check them all.)

Maybe the default `axis` for this function should not have been `-1`, but I don't think it's necessary to change that. The new error message suggests changing the `axis`, which is the fix for #2994.

@agoose77, this would be easy to review. I'm asking you just in case it makes you remember something that you think we should also check. Thanks!

### DIFF
--- a/src/awkward/operations/ak_merge_union_of_records.py
+++ b/src/awkward/operations/ak_merge_union_of_records.py
@@ -164,6 +164,12 @@ def _impl(array, axis, highlevel, behavior, attrs):
 
     def apply(layout, depth, backend, **kwargs):
         posaxis = maybe_posaxis(layout, axis, depth)
+        if posaxis is None:
+            mindepth, maxdepth = layout.minmax_depth
+            raise TypeError(
+                f"cannot merge_union_of_records with axis={axis} if the array has multiple levels of nesting depth; this array has depths from {mindepth} to {maxdepth} (inclusive); try setting axis to a non-negative value"
+            )
+
         if depth < posaxis + 1 and layout.is_leaf:
             raise AxisError(f"axis={axis} exceeds the depth of this array ({depth})")
         elif depth == posaxis + 1 and layout.is_union:


agoose77 (approved):
It's my feeling that we need to be more formal about branching in our `ak.XXX` functions; in this case, it would seem reasonable just to proceed past the branch point, i.e. the point at which `maybe_posaxis` resolves to an integer. However, this is a good starting point that doesn't require much work.

Do you agree? :)

jpivarski:
In other cases, yes: we keep going until `maybe_posaxis` returns non-`None` (if ever). Can that work here? This function can only merge top-level records; if there are nested records, that's not what it would be merging.

I can see the trouble with handling it piecemeal like this, though.

PR #2997, opened 2024-02-01
feat!: __array__ is no longer allowed on NumpyArray and EmptyArray

jpivarski:
As per https://github.com/scikit-hep/awkward/wiki#api-breaking-changes-after-20 for 2.6.0, we now longer allow `__array__` to be called on NumpyArray and EmptyArray.


agoose77:
Nice! This is failing in CI because numba 0.59 has been released, bringing in a feature that we couldn't test yet (the test throws `NotImplementedError`). I'll take a look.

jpivarski:
#2998 isn't in `main` yet, but when it is, merging `main` into here will fix the tests.

agoose77 (approved):

jpivarski:
Thanks!

PR #2998, opened 2024-02-01
test: update test_2799 for Numba 0.59.0

jpivarski:
Numba 0.59.0 is now available (on PyPI), so test_2799 was failing because its first line was

```python
raise NotImplementedError
```

I think you were waiting on information about how Numba would be dealing with this issue when 0.59.0 comes out. Your expectation that `flattened.dtype` would be `int8` doesn't seem to be right. Experimentally, it returns `int64`. Even for pure NumPy (no Awkward Array):

```python
>>> import numba
>>> import numpy as np
>>> @numba.vectorize(nopython=True)
... def add(x, y):
...     return x + y
... 
>>> array = np.array([1, 2, 3, 4], np.int8)
>>> result = add(array, np.int16(np.iinfo(np.int8).max + 1))
>>> result.dtype
dtype('int64')
```

I've updated the test with this correction (what I believe to be a correction).

#2996 and #2997 both need it to be merged, and they need to be merged to release Awkward 2.6.0. (In fact, any subsequent runs of the CI tests will need this, since they get Numba through PyPI and that's 0.59.0 now.)

Two other-library updates came out on the same day, both of which complicate our tests! (pytest 8.0.0 came out, which breaks pytest-asyncio, and Numba 0.59.0 came out.)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
         include:
           - python-version: '3.9'
             python-architecture: x86
-            runs-on: windows-2019
+            runs-on: windows-latest
             dependencies-kind: full
           - python-version: '3.8'
             python-architecture: x64
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,7 +51,7 @@ jobs:
         include:
           - python-version: '3.9'
             python-architecture: x86
-            runs-on: windows-latest
+            runs-on: windows-2019
             dependencies-kind: full
           - python-version: '3.8'
             python-architecture: x64
--- a/requirements-test-full.txt
+++ b/requirements-test-full.txt
@@ -1,6 +1,6 @@
 fsspec;sys_platform != "win32"
 jax[cpu]>=0.2.15;sys_platform != "win32" and python_version < "3.12"
-numba>=0.50.0,!=0.58.0rc1;python_version < "3.12"
+numba>=0.50.0;sys_platform != "win32" and python_version < "3.12"
 numexpr>=2.7; python_version < "3.12"
 pandas>=0.24.0;sys_platform != "win32" and python_version < "3.12"
 pyarrow>=7.0.0;sys_platform != "win32" and python_version < "3.12"
--- a/tests/test_2799_numba_ufunc_resolution.py
+++ b/tests/test_2799_numba_ufunc_resolution.py
@@ -22,8 +22,6 @@ ak.numba.register_and_check()
 
 @pytest.mark.skipif(not NUMBA_HAS_NEP_50, reason="Numba does not have NEP-50 support")
 def test_numba_ufunc_nep_50():
-    raise NotImplementedError
-
     @numba.vectorize(nopython=True)
     def add(x, y):
         return x + y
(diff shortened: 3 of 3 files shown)


jpivarski:
This is just the sort of thing that comes out of nowhere and stops a release:

```
  C:\hostedtoolcache\windows\Python\3.9.13\x86\python.exe C:\Users\runneradmin\AppData\Local\Temp\pip-install-921_u8fz\llvmlite_bdff54372838458392dbf1151245b0b5\ffi\build.py
  CMake Warning (dev) at CMakeLists.txt:3 (project):
    cmake_minimum_required() should be called prior to this top-level project()
    call.  Please see the cmake-commands(7) manual for usage documentation of
    both commands.
  This warning is for project developers.  Use -Wno-dev to suppress it.

  CMake Error at CMakeLists.txt:3 (project):
    Generator

      Visual Studio 16 2019

    could not find any instance of Visual Studio.
```

Is it saying that we need just one more thing in CMakeLists.txt? If so, why did this never come up before, and why only on one version of one build?

Why, o why?

jpivarski:
In fact, `cmake_minimum_required()` _is_ called prior to the top-level `project()`!

https://github.com/scikit-hep/awkward/blob/b0322b568fd2ef647822d0d707df3ee0b7456c7b/awkward-cpp/CMakeLists.txt#L1-L9

agoose77:
@jpivarski no, that's a symptom that the build environment on Windows is not working -- we need to discover the C++ compilers to be able to proceed.

The OS is `windows-latest` and the arch is 32-bit -- my guess is that the windows image has updated and doesn't include VS compilers for 32-bit.

I'm spitballin' as they say, but I don't think it's necessarily a problem. We can check to see whether the windows image has changed, for example.

jpivarski:
`windows-latest` is `windows-2022`, which doesn't sound like a recent update. Maybe that page is old. Perhaps we can explicitly set the Windows runner to `windows-2019` or `windows-2022` (on the assumption that the [runner-images](https://github.com/actions/runner-images) page is old).

jpivarski:
I'll try `windows-2019`.

agoose77:
OK, maybe not that then. Hmm!

agoose77:
OK, it seems unlikely that the image has changed. I note that this fails _later_ when trying to build llvmlite. I suspect this is because Numba no longer releases wheels for this platform - I know that they've just released a new version. But, I think that would be a red herring - we should see the same error in `windows-latest`. Let me see if cmake has changed.

agoose77:
This is ultimately being caused by pip trying to build an sdist of llvmlite because PyPI doesn't have a 32-bit wheel. I am content not to get to the bottom of why this happens, but I suspect the traceback is caused by llvmlite's internal detection logic (https://github.com/numba/llvmlite/blob/f22420ad768a31dd15ae45d551bb4e73907a27fa/ffi/build.py#L56). This follows from the fact that an awkward-cpp build succeeds in the same job (although it might use a different CMake from PyPI, that's irrelevant).

Off-hand I can't remember @jpivarski when we plan to drop 32-bit windows. It won't be hard to keep it running for now; we just need disable numba on win (32bit)

jpivarski:
That's a good idea—dropping Numba on the 32-bit Windows test. I'll do that.

The Windows 32-bit test was historically very valuable in helping us eliminate cases in which we wrongly assumed that `intp is int64` in our C++ code. There's no reason to test that with Numba, but it's good to keep testing it as long as we _might_ edit the C++ code and Windows 32-bit is still a common enough platform to consider targeting.

You've also answered the "why now?" question: because Numba released 0.59.0 so recently.

agoose77 (approved):

Issue #2987, opened 2024-01-25
CUDA kernels that are implemented but not optimal


jpivarski:
This is primarily for record-keeping, so that we don't forget about CUDA kernels that should be revisited _someday_. To be in this list, a kernel must be implemented correctly (in `main` or an impending PR), but have some reason to be rewritten. The list is to help us stick to the policy that existence is the first priority and optimization is second, without the temptation to go down a rabbit-hole of optimizing every kernel before moving on to the next one.

- [x] [awkward_ListArray_min_range](https://awkward-array.org/doc/main/reference/generated/kernels.html#awkward-listarray-min-range) is a reduce-to-scalar algorithm, but it's implemented with `atomicMin` instead of tree-reduction.

Variable-length inner loop:
- [ ] awkward_IndexedArray_ranges_next_64
- [ ] awkward_IndexedArray_ranges_carry_next_64
- [ ] awkward_ListArray_getitem_jagged_numvalid
- [ ] awkward_ListArray_getitem_next_range_spreadadvanced
- [ ] awkward_ListArray_broadcast_tooffsets
- [ ] awkward_ListArray_localindex
- [ ] awkward_ListOffsetArray_drop_none_indexes
- [ ] awkward_ListOffsetArray_reduce_local_nextparents_64
- [ ] awkward_ListArray_rpad_axis1
- [ ] awkward_ListOffsetArray_rpad_axis1
- [ ] awkward_ListArray_combinations_length
- [ ] awkward_NumpyArray_pad_zero_to_length
- [ ] awkward_NumpyArray_rearrange_shifted
- [ ] awkward_UnionArray_flatten_combine
- [ ] awkward_UnionArray_nestedfill_tags_index

PR #3003, opened 2024-02-05
feat: add variable length loop kernels

ManasviGoyal:
Added the following kernels:

1. awkward_IndexedArray_ranges_next_64*
2. awkward_IndexedArray_ranges_carry_next_64*
3. awkward_ListArray_getitem_jagged_descend
4. awkward_ListArray_getitem_jagged_numvalid*
5. awkward_ListArray_getitem_next_range_spreadadvanced*
6. awkward_ListOffsetArray_toRegularArray
7. awkward_ListOffsetArray_rpad_length_axis1
8. awkward_ListArray_broadcast_tooffsets*
9. awkward_ListArray_localindex*
10. awkward_ListOffsetArray_drop_none_indexes*
11. awkward_ListOffsetArray_reduce_local_nextparents_64*
12. awkward_ListOffsetArray_reduce_nonlocal_maxcount_offsetscopy_64
13. awkward_UnionArray_regular_index_getsize

*variable-length inner loop (need to look for a way to optimize, added them to #2987)

Also, fixed awkward_ListArray_compact_offsets. Now we can generate tests for kernels where out arguments are also in arguments.


ManasviGoyal:
@jpivarski there are a few more kernels that would fall in the same category but I think it would be better to do that in a new PR.

dev/generate-tests.py
```diff
@@ -228,8 +228,10 @@ def getdtypes(args):
         if "List" in typename:
             count = typename.count("List")
             typename = gettypename(typename)
-            if typename == "bool" or typename == "float":
+            if typename == "bool":
                 typename = typename + "_"
+            if typename == "float":
+                typename = typename + "32"
```
jpivarski, line 234: That must have been because of some past misunderstanding. Thanks for fixing this.

jpivarski (approved):
Looks great! I also checked it out and tested it on my GPU. It's nice to see only 2 xfails left in `tests-cuda`.

Feel free to merge at any time.

Also, you're right to break them up in separate PRs like this.

PR #3006, opened 2024-02-05
build: set a minimum fsspec version (see #3002)

### LINKED ISSUE #3002: fsspec needs a version limit in
### Version of Awkward Array

2.6.0

I had a very old version of fsspec installed when I installed awkward. Upon import awkward I got:

```
File "/path/to/lib/python3.9/site-packages/awkward/operations/ak_from_parquet.py", line 5, in <module>
    import fsspec.parquet
ModuleNotFoundError: No module named 'fsspec.parquet'
```

I think fsspec.parquet has only been available since late 2021. The [requirements](https://github.com/scikit-hep/awkward/blob/main/requirements-test-minimal.txt) [lists](https://github.com/scikit-hep/awkward/blob/main/requirements-test-full.txt) should specify recent versions of fsspec that work.

jpivarski:


jpivarski:
I'm going to merge this without review; it's a minimum version of fsspec. I want to get it into a new version quickly.

Issue #2971, opened 2024-01-19
Intermittent segfault in pypy


jpivarski:
### Version of Awkward Array

HEAD

### Description and code to reproduce

We now have a test that's tracking it, but that test fails (with the segfault) occasionally, and we eventually need to do something about it.

If we knew that the segfault is _only_ in pypy, then I'm partially inclined to just declare that we don't support pypy. (How important is it? Who uses it?) _But_, it's not uncommon for a segfault that only occurs on one platform to be an indication of a bug across all platforms, but it "accidentally works" on the others because it lands within C++'s undefined behavior. So as long as we don't know why the segfault occurs, it's more important than just supporting pypy.

PR #3010, opened 2024-02-07
chore: fix ruff warnings

agoose77:
This PR fixes some ruff warnings that have appeared. Some are general, some are NumPy.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -505,7 +505,7 @@ def _scalar_type_of(obj) -> DType:
     if is_unknown_scalar(obj):
         return obj.dtype
     else:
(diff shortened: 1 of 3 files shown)


agoose77:
@ManasviGoyal this should hopefully appease ruff :)

agoose77:
@jpivarski I set this to automerge, and it succeeded despite the pypy test failure. I assume that means we explicitly don't want this to be a required test?

Have you seen that pypy segfault before? I don't think this PR did anything that would introduce it, so I'm wondering if it's flaky.

jpivarski:
It's #2971, a known issue. The reason it's not a required test is because of the segfault. I want to keep running the test, but not require it, to keep in mind that it's an unsolved problem.

PR #3017, opened 2024-02-11
feat: add __dask_tokenize__ hook

lgray:
This is related to https://github.com/dask/dask/pull/10918

In particular with https://github.com/dask/dask/pull/10883 even those changes don't fix everything.

This at least makes it so that only shape is touched in the present error and with future changes.

This rest is that `array.layout.form` touches shape.

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -348,6 +348,9 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
     _histogram_module_ = awkward._connect.hist
 
+    def __dask_tokenize__(self):
+        return self.layout.form.to_json(), self.behavior, self.attrs
+    
     def _update_class(self):
         self._numbaview = None
         self.__class__ = get_array_class(self._layout, self._behavior)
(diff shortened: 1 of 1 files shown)


src/awkward/highlevel.py
```diff
@@ -348,6 +348,9 @@ def __init_subclass__(cls, **kwargs):
 
     _histogram_module_ = awkward._connect.hist
 
+    def __dask_tokenize__(self):
+        return self.layout.form.to_json(), self.behavior, self.attrs
```
agoose77: I don't think this will work for our needs --- `__dask_tokenize__` should deterministically represent the object. In this case, any `ak.Array` with the same form would be considered equal, which is not correct.

lgray:
Just to link to the wider discussion: https://github.com/dask-contrib/dask-awkward/issues/468

agoose77:
@lgray is this still a direction worth pursuing?

lgray:
Yes it is still useful in case people are sticking awkward arrays directly in the graph.

agoose77:
Then I *think* we can merge this. Are there any tests in dask-awkward?

lgray:
Since the operation I mentioned is something you do *outside* of dask awkward there are no tests in dask awkward for it.

agoose77:
@lgray hmm, fair. We can add tests here to check this, then :) Leave it to me ...

jpivarski:
This was accepted, but somehow missed getting merged. I'll merge it now.

PR #3026, opened 2024-02-16
fix: throw more meaningful error message for non-NumPy buffers

agoose77:
Improves dask-contrib/dask-awkward#472


jpivarski (approved):
Absolutely; thanks!

I'm merging it now.

PR #3027, opened 2024-02-16
chore: figuring out how to implement ak.from_cudf

jpivarski:
For now, this is just tinkering in the studies directory to try to figure stuff out.

### DIFF
--- a/studies/cudf-to-awkward.py
+++ b/studies/cudf-to-awkward.py
@@ -5,6 +5,7 @@ import numpy
 
 import awkward as ak
 from awkward._backends.numpy import NumpyBackend
+from awkward._backends.cupy import CupyBackend
 
 
 ######################### stripped-down copy of src/awkward/_connect/pyarrow.py
@@ -71,13 +72,13 @@ def popbuffers_finalize(out, array, validbits, generate_bitmasks, fix_offsets=Tr
         )
 
 
-def popbuffers(paarray, storage_type, buffers, generate_bitmasks):
+def popbuffers(paarray, arrow_type, buffers, generate_bitmasks):
     ### Beginning of the big if-elif-elif chain!
 
-    if isinstance(storage_type, pyarrow.lib.DictionaryType):
+    if isinstance(arrow_type, pyarrow.lib.DictionaryType):
         masked_index = popbuffers(
             paarray.indices,
-            storage_type.index_type,
+            arrow_type.index_type,
             buffers,
             generate_bitmasks,
         )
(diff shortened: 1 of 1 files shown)


jpivarski:
@martindurant, I've started on this approach, in which I'm implementing

https://github.com/scikit-hep/awkward/blob/bac7ab8020479b92c73303e571618de0119b6ffe/studies/cudf-to-awkward.py#L350-L480

to be like

https://github.com/scikit-hep/awkward/blob/bac7ab8020479b92c73303e571618de0119b6ffe/studies/cudf-to-awkward.py#L48-L347

substituting CuDF accessors (`base_data`, `base_mask`, `base_children`[^1]) for pyarrow accessors (`buffers` in a particular order) and NumPy casting (`np.frombuffer)` for CuPy casting (`cp.asarray`). The tree of CuDF `ColumnBase` nodes is a safer interface than the list of pyarrow `Buffer` objects because you can't accidentally be on the wrong tree node; it's less sensitive to how many `None` values should be associated with a particular node, etc.

In

https://github.com/scikit-hep/awkward/blob/bac7ab8020479b92c73303e571618de0119b6ffe/studies/cudf-to-awkward.py#L525-L543

you can see that I've started testing some nested lists and missing values. (We're not checking expected types because Arrow → Awkward has an ambiguity about top-level option-types.)

I can continue with this, though not right now. We still don't know how multiple children in a StructArray or multiple children _and_ multiple Indexes (tags & index) in a UnionArray are laid out, but I think we've seen enough to know that CuDF exposing `Series._data` would be sufficient for our needs.

[^1]: For CuDF, `base_children` includes what we call Indexes as well as the subtrees, with the Indexes presented as `NumericColumn`. CuDF is not making a distinction between what we call Index and what we call Content.

martindurant:
Let me know if you want me to try anything here. Of course you know the code much better.

I feel like somewhere in https://github.com/rapidsai/cudf/blob/branch-24.04/python/cudf/cudf/core/column/column.py is a way to make a series (column) out of buffers.

jpivarski:
I think it's straightforward from this point. I'll just have to do it.

If you have an alternative idea, please go ahead and try it. Also, none of this touches `ak.to_cudf`, so if you can find out how to build a CuDF Column, that will help. (In Awkward, the `to_arrow` functions are implemented as methods on all of the Content node types—that direction is much easier because of the direction of the projection.)

jpivarski:
I finished off the implementation. Since this PR doesn't touch the codebase, I'll merge it without review.

Some comments on CuDF's implementation of Arrow:

* We have to go down a different rabbithole to find the indexes and dictionary of a categorical variable than we do with pyarrow (not surprising). The index is not necessarily 32-bit though: for small numbers of categories, it can be 8-bit (surprising; I think that goes against the Arrow spec). I couldn't include a test of categoricals because `IndexedOptionArray.simplified`'s kernel has not been implemented (though I tested it offline by swapping `IndexedOptionArray` for `IndexedArray`, which would be wrong with nullable data, but it's enough of a test to be confident in the implementation.)
* CuDF has not implemented Arrow's `FixedSizeListType` or `FixedSizeBinaryType`. I included a very plausible implementation. (The new data, the fixed size, comes from the `arrow_type`, which we already have.)
* CuDF has not implemented Arrow's binary (as opposed to Unicode string) type, but I included a very plausible implementation (because it's just like the Unicode string implementation).
* CuDF's boolean arrays are implemented as bytes, rather than bits. (That's better when you want to apply masks, but it's against the Arrow spec.)
* I tested date-time and time-delta types, but I don't have a way to do timezone-aware date-times because Awkward inherits NumPy's timezone-naivety.
* CuDF does have decimal types, but I can't test them (just as for Arrow) because NumPy doesn't have them.
* I don't know what the `IntervalColumn` is. Interval types are new to me.

PR #3028, opened 2024-02-19
fix: proper typetracer array slicing in `BitMaskedArray`

agoose77:
A long-running design choice was to avoid handling unknown-lengths as indices in `TypeTracerArray`; the caller (usually `Content` methods) would normalise to unknown-scalars (0D arrays).

This PR fixes a place where this assumption is not properly ensured, which was surfaced during testing of dask-awkward.

In the long run, I'd be in favour of relaxing this assumption and requiring that `TypeTracerArray` properly normalise unknown lengths to unknown scalars, but that's out of scope for this fix.

### DIFF
--- a/src/awkward/contents/bitmaskedarray.py
+++ b/src/awkward/contents/bitmaskedarray.py
@@ -471,7 +471,9 @@ class BitMaskedArray(BitMaskedMeta[Content], Content):
                 self._lsb_order,
             )
         )
-        return bytemask.data[: self._length].view(np.bool_)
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Got it—that makes sense! I followed this through to its definition; the wrapper function is ensuring that the `unknown_length` singleton becomes a TypeTracerArray scalar, so that it can be used as a slice endpoint.

This is good to merge, and it looks done, so I'll merge it myself.

PR #3030, opened 2024-02-21
fix: performance of ufunc resolution for non-nominal signatures

agoose77:
@lgray observed that `find_ufunc` is a significant bottleneck in coffea-dask workflows. Whilst we can optimise this further by changing our internal representation of behaviors (i.e. `__getitem__` registers overloads in high-performance lookup structures), for now I suspect that we can improve the performance of the common-path. @jpivarski notes that the behaviour changed in this PR was never intended to be supported, so it represents a correctness improvement regardless.

The case that this PR improves is

`np.add(foo, bar, ...)` where neither `foo` nor `bar` are Awkward Arrays with names (e.g. `__record__`). This represents _most_ ufunc operations given that even ufunc overloads of records often delegate to Awkwards' ufunc mechanism on arrays of numbers.

### DIFF
--- a/src/awkward/_behavior.py
+++ b/src/awkward/_behavior.py
@@ -100,29 +100,32 @@ def find_ufunc_generic(
 
 
 def find_ufunc(behavior: Mapping | None, signature: tuple) -> UfuncLike | None:
-    if all(s is not None for s in signature):
-        behavior = overlay_behavior(behavior)
+    if any(s is None for s in signature):
+        return None
 
-        # Special case all strings or hashable types.
+    behavior = overlay_behavior(behavior)
+
+    # Try and fast-path the lookup
+    try:
+        return behavior[signature]
+    except KeyError:
+        # We didn't find an exact overload, and we won't find any!
         if all(isinstance(x, str) for x in signature):
-            return behavior.get(signature)
-        else:
-            for key, custom in behavior.items():
-                if (
-                    isinstance(key, tuple)
-                    and len(key) == len(signature)
-                    and key[0] == signature[0]
-                    and all(
-                        k == s
-                        or (
-                            isinstance(k, type)
-                            and isinstance(s, type)
-                            and issubclass(s, k)
-                        )
-                        for k, s in zip(key[1:], signature[1:])
-                    )
-                ):
-                    return custom
+            return None
+
+    # Fall back on linear search (first-wins)
+    for key, custom in behavior.items():
+        if (
+            isinstance(key, tuple)
+            and len(key) == len(signature)
+            and key[0] == signature[0]
+            and all(
+                k == s
+                or (isinstance(k, type) and isinstance(s, type) and issubclass(s, k))
+                for k, s in zip(key[1:], signature[1:])
+            )
+        ):
+            return custom
     return None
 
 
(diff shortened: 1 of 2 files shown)


lgray:
This corresponds to a ~20% improvement in execution time (O(1s) out of 5s) building a single task graph in the workflow in question, so this is a rather significant improvement. 

Keeping in mind the task graph gets built 180 times or so for the full analysis.

src/awkward/_behavior.py
```diff
@@ -100,29 +100,32 @@ def find_ufunc_generic(
...
+    if any(s is None for s in signature):
+        return None
+
+    behavior = overlay_behavior(behavior)
 
-        # Special case all strings or hashable types.
+    # Try and fast-path the lookup
+    try:
+        return behavior[signature]
+    except KeyError:
```
jpivarski, line 111: If a signature consists entirely of hashable things like the ufunc itself and strings matching a set of names in parameters (`__array__`, `__record__`, or `__list__`), then it gets fast-pathed and we get to use dict-lookup.

Good! :heavy_check_mark:

src/awkward/_behavior.py
```diff
@@ -100,29 +100,32 @@ def find_ufunc_generic(
...
-                        or (
-                            isinstance(k, type)
-                            and isinstance(s, type)
-                            and issubclass(s, k)
-                        )
-                        for k, s in zip(key[1:], signature[1:])
-                    )
-                ):
-                    return custom
+            return None
```
jpivarski, line 114: If a signature (apart from the ufunc itself—or does the signature-part start _after_ the ufunc?) is all strings and we didn't find it with a hash-lookup, we won't find it by iterating over `behavior.keys()`, so don't even bother.

Good! :heavy_check_mark:

src/awkward/_behavior.py
```diff
@@ -100,29 +100,32 @@ def find_ufunc_generic(
 
 
 def find_ufunc(behavior: Mapping | None, signature: tuple) -> UfuncLike | None:
-    if all(s is not None for s in signature):
-        behavior = overlay_behavior(behavior)
+    if any(s is None for s in signature):
+        return None
```
jpivarski, line 104: If a signature contains any `None` value, give up on it.

I don't remember a reason for a signature to have a `None` value. Does this ever happen? Why? What does it mean?

src/awkward/_connect/numpy.py
```diff
@@ -207,23 +207,29 @@ def _array_ufunc_adjust_apply(
...
             else:
                 signature.append(None)
         else:
             signature.append(type(x))
 
-    return tuple(signature)
+    if has_seen_nominal_type:
+        return (ufunc, *signature)
+    else:
+        return None
```
jpivarski, line 232: A signature must have at least one nominal type in it (at least one string). We should not be allowing overrides of functions based on types that are not named by parameters (`__array__`, `__record__`, or `__list__`).

We talked about this on our call, and I think it's a good rule. (I hadn't, previously, intended the alternative: overriding ufunc behaviors when there are no named types in the Awkward Array.)

Good! :heavy_check_mark:

jpivarski (approved):
I think this is ready to be merged. I've check-marked that all of the rules I'd expect are included in the new implementation.

Thanks!

src/awkward/_behavior.py
agoose77: `None` is the signature type given to an Awkward `Content` that has no names and is not a `NumpyArray`.

PR #3038, opened 2024-03-04
chore: keeping CONTRIBUTING.md up to date (take 1)

jpivarski:


ManasviGoyal:
@jpivarski I think we should also mention the unit-tests (`-explicit`) for fine grained testing of all three types of kernels.

jpivarski:
That's a good idea. Please just edit the file in this PR. It can be a new section, after basic testing, and it doesn't have to be detailed at all—just what kernel-testing is for and how to do it.

jpivarski:
To keep this from getting out of date, I'm going to merge it now. If we need to change CONTRIBUTING.md again, we'll make a "take 2" PR.

PR #3051, opened 2024-03-13
feat: Add to_cudf

martindurant:
Not too much here yet, just started writing out the template to let others get involved


src/awkward/contents/numpyarray.py
```diff
@@ -5,6 +5,8 @@
 import copy
 from collections.abc import Mapping, MutableMapping, Sequence
 
+import cupy
```
martindurant: to remove, should only be in method

martindurant:
I think we intended to come back to this PR and implement string conversion? At the moment, you have the following behaviour:
```
s = cudf.Series([["he", "hi"], [], ["ho"]])

In [6]: import awkward_pandas.cudf

In [7]: s.ak
Out[7]: <awkward_pandas.cudf.CudfAwkwardAccessor at 0x7ff1cb7a2d40>

In [8]: s.ak.to_output()  # lost string-ness
Out[8]: 
0    [[104, 101], [104, 105]]
1                          []
2                [[104, 111]]
dtype: list

In [9]: s.ak.array  # cupy-backed ak array
Out[9]: <Array [['he', 'hi'], [], ['ho']] type='3 * var * ?string'>

In [10]: ak.to_cupy(s.ak.array)
KeyError                                  Traceback (most recent call last)
 ...

File /floppy/code/awkward/src/awkward/_backends/cupy.py:39, in CupyBackend.__getitem__(self, index)
     38 _cuda_kernels = cuda.initialize_cuda_kernels(cupy)
---> 39 func = _cuda_kernels[index]
     40 if func is not None:

KeyError: ('awkward_NumpyArray_prepare_utf8_to_utf32_padded', <class 'numpy.uint8'>, <class 'numpy.int32'>, <class 'numpy.int64'>)
```

I'm not entirely sure why the two routes are different, but either way we do need this to work.

This PR is pinned to release 2.6.3, so it's also slipped behind the main branch, and we could do with some plan to complete it.

src/awkward/_errors.py
```diff
@@ -75,6 +75,7 @@ def __enter__(self):
             self._slate.__dict__["__primary_context__"] = self
 
     def __exit__(self, exception_type, exception_value, traceback):
+        return
```
martindurant:
```suggestion
```

martindurant:
@jpivarski , probably plenty of work left here and classes not covered, but I thought you would like to have a look, now that I have at least a few tests.

martindurant:
Missing: IndexedArray; does this map to categorical?

jpivarski:
IndexedArray is only mapped to Arrow's categorical ("dictionary") if `"__array__": "categorical"`. Otherwise, the IndexedArray gets [projected](https://awkward-array.org/doc/main/reference/generated/ak.contents.IndexedArray.html#ak.contents.IndexedArray.project). The reason is that Arrow's categorical segfaults if the contents are not unique, and Awkward only guarantees uniqueness if requested, for instance if the array was constructed via [ak.str.to_categorical](https://awkward-array.org/doc/main/reference/generated/ak.str.to_categorical.html).

martindurant:
@jpivarski , this PR is still languishing here. What do you think is needed to get it released, at least in experimental form?

src/awkward/contents/numpyarray.py
jpivarski: This was removed, right?

jpivarski (approved):
I see that the `import cupy` was removed from top-level imports. (Anyway, tests wouldn't pass without that.) This all looks good—I don't know why I didn't see it when I was scanning through the open PRs for any that had been waiting.

I also just tested tests-cuda on my GPU, and everything passes.

So I'll merge it!

src/awkward/contents/numpyarray.py
jpivarski: Yes, it was.

martindurant: Correct, has `from awkward._nplikes.cupy import Cupy`

tacaswell:
This reverted commit `rapidjson` is on which breaks building with gcc 14 (xref https://github.com/scikit-hep/awkward/pull/3110).

I could not quickly tell if this was intentional or accidental from the commits in this PR.

tacaswell:
https://github.com/scikit-hep/awkward/pull/3249 to fix it.

jpivarski:
It was accidental. Thanks!

Issue #2936, opened 2024-01-11
Test against NumPy 2.0


jpivarski:
Ruff has tools for this, and there are NumPy 2.0 prereleases (or there will be soon).

jpivarski:
https://odapt.readthedocs.io/en/latest/

jpivarski:
https://numpy.org/devdocs/release/2.0.0-notes.html

https://pythonspeed.com/articles/numpy-2/

jpivarski:
NumPy 2.0's [release notes](https://numpy.org/devdocs/release/2.0.0-notes.html) serve as a migration guide, and their umbrella issue, numpy/numpy/issues/24300, is a good source of news.

jpivarski:
The development wheels are here: https://anaconda.org/scientific-python-nightly-wheels/numpy!

rgommers:
Some potential concern was raised at https://github.com/pybind/pybind11/pull/5050#issuecomment-1987351644 about Awkward Array using structured dtypes, which are a potential hiccup in `pybind11`'s support for numpy 2.0. It's unclear if that risk will materialize. Have you done any testing with numpy 2.0 nightlies already @jpivarski?

agoose77:
Hmm, I'm not sure what @henryiii is exactly saying in pybind/pybind11#5050. AFAICR, we don't inherently use any C structs via NumPy structured dtypes. But, perhaps I'm forgetting. @jpivarski might recall.

I believe I've tested locally against a NumPy 2.0 pre/dev-release, and didn't spot anything.

henryiii:
The change only landed in NumPy quite recently from what I understand. If you didn’t have to use a pybind11 branch, it shouldn’t have worked, I believe (regardless of if any custom dtypes are used). It might be that all structured types are handled by the forth layer?

Anyway, I’ll try to test with boost-histogram, which uses structured dtyoes for the weighed and mean storages.

rgommers:
> The change only landed in NumPy quite recently from what I understand.

Indeed, last Thursday or so.

> If you didn’t have to use a pybind11 branch, it shouldn’t have worked, I believe (regardless of if any custom dtypes are used)

`pybind11` 2.11.1 works just fine for SciPy and Matplotlib right now - it just emits verbose warnings about `pybind11` digging into private NumPy internals that have moved, but that doesn't break the build (at least not yet - it will at some point).

jpivarski:
[...]

PR #3052, opened 2024-03-14
docs: add a link to documentation of AwkwardArray in Julia

ianna:
AwkwardArray in Julia documentation is created by Documenter.jl.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -101,7 +101,7 @@ Spotted a typo in the documentation? Want to add to the codebase? The contributi
 :link: https://juliahep.github.io/AwkwardArray.jl/dev/
 :class-card: admonition warning
 
-Looking for the documentation for Awkward Array in Julia? Click this card.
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -93,6 +93,15 @@ Spotted a typo in the documentation? Want to add to the codebase? The contributi
 :class-card: admonition warning
 
 {fas}`history` Need the documentation for version 1 of Awkward Array? Click this card.
+
+:::
+
+:::{grid-item-card}
```
ianna, line 99: The card looks fine - as expected.

ianna (commented):
@jpivarski - I'm not sure why the packaging test on ubuntu fails.

agoose77:
@ianna looks like we're pulling in a NumPy beta, which is not yet supported!

jpivarski:
For now, we can ignore that test failure (or prevent NumPy 2.0.0b1 from being installed, only to have to reverse it later?), particularly for this PR, which is a documentation PR.

@tcawlfield is looking into #2936 right now. Hopefully, the changes needed are small and can be confined to nplike. The first issue, `numpy.AxisError` moving to some other path, would be easy to fix. Before making major changes, like putting an upper cap on NumPy version, we need to get a sense of how many issues there are and how long it will take.

jpivarski:
Note to self: I want to find a three-dots inline Julia icon, as a replacement for the

```
{fas}`history`
```

If neither of us can find such a thing, we can call this PR done.

ianna:
I only found an ellipsis icon among `{fas}`

jpivarski:
Ha! It worked!

image

I'd say it's ready to merge now. The tests won't all run, since this PR doesn't touch code.

PR #3053, opened 2024-03-15
feat: adding support for numpy.real, imag, round, angle, 

### LINKED ISSUE #2917: np.real, np.imag, np.round, np.around can't take an Awkward Array as input
### Version of Awkward Array

HEAD

This is more like a strange "gotcha" of NumPy, but `np.real` and `np.imag` are not ufuncs, unlike, for example, np.conjugate.

```python
>>> np.real
<function real at 0x7e73131d8930>
>>> np.imag
<function imag at 0x7e73131d8b30>
>>> np.conjugate
<ufunc 'conjugate'>
```

Therefore, it would require some special handling to make it work in Awkward, but not much. Here's what happens now:

```python
>>> np.real(ak.Array([[1+0.1j, 2+0.2j, 3+0.3j], [], [4+0.4j, 5+0.5j]]))
  File "/home/jpivarski/irishep/awkward/src/awkward/highlevel.py", line 1527, in __array_function__
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 109, in array_function
    rectilinear_args = tuple(_to_rectilinear(x, backend) for x in args)
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 109, in <genexpr>
    rectilinear_args = tuple(_to_rectilinear(x, backend) for x in args)
  File "/home/jpivarski/irishep/awkward/src/awkward/_connect/numpy.py", line 78, in _to_rectilinear
    return layout.to_backend(backend).to_backend_array(allow_missing=True)
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/content.py", line 1021, in to_backend_array
    return self._to_backend_array(allow_missing, backend)
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/listoffsetarray.py", line 2078, in _to_backend_array
    return self.to_RegularArray()._to_backend_array(allow_missing, backend)
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/listoffsetarray.py", line 288, in to_RegularArray
    self._backend.maybe_kernel_error(
  File "/home/jpivarski/irishep/awkward/src/awkward/_backends/backend.py", line 67, in maybe_kernel_error
    raise ValueError(self.format_kernel_error(error))
ValueError: cannot convert to RegularArray because subarray lengths are not regular (in compiled code: https://github.com/scikit-hep/awkward/blob/awkward-cpp-26/awkward-cpp/src/cpu-kernels/awkward_ListOffsetArray_toRegularArray.cpp#L22)
```

It's trying to auto-convert the Awkward Array into a NumPy array to use the function, and it can't because this one is ragged.

tcawlfield:
Closes #2917

This issue does not mention round explicitly. and it does mention around which I have not duplicated yet.
I could additionally implement `numpy.fix` and `real_if_close`.

There is an error in my attempt to support the `out=<array>` optional argument -- see discussion.


tcawlfield:
Are there automatic code formatters and formatting checks? I guess I'll find out! Oh, I did find out and there are indeed.

ianna:
> Are there automatic code formatters and formatting checks? I guess I'll find out! Oh, I did find out and there are indeed.

Indeed, they are. You can run `pre-commit run --all` and fix the warnings before pushing then you will not need to pull the automatically updated branch from remote.

jpivarski:
We don't yet have automatic code formatting on C++, but ~~black~~ ruff auto-formats Python. (And some other things in pre-commit format YAML and TOML.) We've talked about formatting C++ (#34, #1865, #2902), but never converged.

I agree with @ianna that it's more convenient to run pre-commit locally, since formatting changes can automatically write new commits to the PR, and then if you commit on your local clone, you've got a conflict to manage.

src/awkward/_nplikes/typetracer.py
```diff
@@ -81,6 +81,14 @@ def _emptyarray(x):
         return numpy.empty(0, numpy.array(x).dtype)
 
 
+_real_parts_of = {
+    # Functions real, imag, and angle return real arrays when given complex arguments.
+    np.dtype("csingle"): np.dtype("single"),
+    np.dtype("cdouble"): np.dtype("double"),
+    np.dtype("clongdouble"): np.dtype("longdouble"),
+}
```
jpivarski: This works, although it reproduces some logic that exists in NumPy, and our main concern is to make Awkward do the same thing that NumPy does, whatever that is. Also, this uses some non-standard names, like `"single"`, `"double"`, and `"longdouble"`; in this codebase, we would always say `np.float32`, `np.float64`, and `np.float128` (if it exists; [it's platform-dependent](https://github.com/scikit-hep/awkward/blob/dfc9690537386ab690c7e7de1baa8e6b041632be/src/awkward/types/numpytype.py#L87-L88)).

Here's a way to get NumPy to tell us how it convert a complex dtype `dt` into the corresponding real dtype:

```python
np.real(np.zeros(0, dtype=dt)).dtype
```

src/awkward/_nplikes/typetracer.py
```diff
@@ -1438,6 +1446,36 @@ def isnan(self, x: TypeTracerArray) -> TypeTracerArray:
...
+        return TypeTracerArray._new(_real_parts_of.get(x.dtype, x.dtype), shape=x.shape)
+
+    def angle(self, x: TypeTracerArray, deg: bool = False) -> TypeTracerArray:
+        assert isinstance(x, TypeTracerArray)
+        try_touch_data(x)
+        if np.issubdtype(x.dtype, np.integer) or np.issubdtype(x.dtype, np.bool_):
+            # numpy's angle() casts integer-like values to float (platform-dependent).
+            result_type = np.dtype(numpy.float64)
+        else:
+            result_type = _real_parts_of.get(x.dtype, x.dtype)
```
jpivarski: Similar to the above, it looks like you're trying to encode type conversion rules that NumPy already implements. If this is platform-dependent, it's one that NumPy will likely change in NumPy 2.0. But you can use the same trick for `np.angle` as `np.real`.

src/awkward/operations/ak_real_imag_angle.py
```diff
@@ -0,0 +1,133 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
```
jpivarski: Even if it looks wasteful, let's keep it as one Python file per high-level function (so, 3 files in this case).

jpivarski (changes requested):
This implements `real`, `imag`, `angle`, and `round` as NEP-18 overloads, since they're not ufuncs like `np.floor`:

```python
>>> np.real
<function real at 0x70646be60d70>
>>> np.imag
<function imag at 0x70646be613f0>
>>> np.angle
<function angle at 0x7063e9b4ac30>
>>> np.round
<function round at 0x70646bfd3eb0>
>>> np.floor
<ufunc 'floor'>
```

Overall, it looks good; I just have a few inline requested changes.

src/awkward/operations/ak_real_imag_angle.py
agoose77: Ooh, I missed this on mobile!

jpivarski:
Let me know when you want me to re-review this!

jpivarski (approved):

jpivarski:
Great, thanks! This one is done and I'll merge it.

jpivarski:
@all-contributors please add @tcawlfield for code

PR #3054, opened 2024-03-15
perf: add a flag to turn off isinstance in `TypeTracerArray._new`

### LINKED ISSUE #3036: awkward._nplikes.typetracer.TypeTracerArray._new needs further optimization, it is hot code
### Version of Awkward Array

2.6.1

The present code is below:
```python3
    @classmethod
    def _new(
        cls,
        dtype: DType,
        shape: tuple[ShapeItem, ...],
        form_key: str | None = None,
        report: TypeTracerReport | None = None,
    ):
        self = super().__new__(cls)
        self.form_key = form_key
        self.report = report

        if not isinstance(shape, tuple):
            raise TypeError("typetracer shape must be a tuple")
        if not all(isinstance(x, int) or x is unknown_length for x in shape):
            raise TypeError("typetracer shape must be integers or unknown-length")
        if not isinstance(dtype, np.dtype):
            raise TypeError("typetracer dtype must be an instance of np.dtype")
        self._shape = shape
        self._dtype = dtype

        return self
```

in particular:
```python3
        if not isinstance(shape, tuple):
            raise TypeError("typetracer shape must be a tuple")
        if not all(isinstance(x, int) or x is unknown_length for x in shape):
            raise TypeError("typetracer shape must be integers or unknown-length")
        if not isinstance(dtype, np.dtype):
            raise TypeError("typetracer dtype must be an instance of np.dtype")
```
is quite expensive at the scale of thousands of awkward array kernel calls and in normal, non-development, operation none of these if statements ever evaluate to True. They are relatively expensive to evaluate when there is no data processing, see for example typetracing in dask-awkward.

We know already that commenting out these lines brings significant performance improvements.
A reasonable alternative to that would be wrapping this code in an if block.

```python3
    @classmethod
    def _new(
        cls,
        dtype: DType,
        shape: tuple[ShapeItem, ...],
        form_key: str | None = None,
        report: TypeTracerReport | None = None,
    ):
        self = super().__new__(cls)
        self.form_key = form_key
        self.report = report

        if __development_build__:
            if not isinstance(shape, tuple):
                raise TypeError("typetracer shape must be a tuple")
            if not all(isinstance(x, int) or x is unknown_length for x in shape):
                raise TypeError("typetracer shape must be integers or unknown-length")
            if not isinstance(dtype, np.dtype):
                raise TypeError("typetracer dtype must be an instance of np.dtype")
        self._shape = shape
        self._dtype = dtype

        return self
```

Where `__development_build__` is just a stand in name. I think the only requirements would be that the parameter to the if statement is a simple bool, no lookups (etc.), as this is known to be pretty hot code.

Savings are O(1s) for a complex HEP analysis when building a single few-thousand layer task graph. This corresponds to many minutes saved in processing for full-scale analysis.

@jpivarski @agoose77 @martindurant

jpivarski:
The runtime type-checks in `TypeTracerArray._new` can now be turned off by setting

```python
awkward.typetracer.TypeTracerArray.runtime_typechecks = False
```

which is a developer-public (but user-hidden) global interface.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -164,6 +164,8 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
     _dtype: numpy.dtype
     _shape: tuple[ShapeItem, ...]
 
+    runtime_typechecks = True
+
     def __new__(cls, *args, **kwargs):
         raise TypeError(
             "internal_error: the `TypeTracer` nplike's `TypeTracerArray` object should never be directly instantiated"
@@ -185,12 +187,13 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
         self.form_key = form_key
         self.report = report
 
-        if not isinstance(shape, tuple):
-            raise TypeError("typetracer shape must be a tuple")
-        if not all(isinstance(x, int) or x is unknown_length for x in shape):
-            raise TypeError("typetracer shape must be integers or unknown-length")
-        if not isinstance(dtype, np.dtype):
-            raise TypeError("typetracer dtype must be an instance of np.dtype")
+        if cls.runtime_typechecks:
+            if not isinstance(shape, tuple):
+                raise TypeError("typetracer shape must be a tuple")
+            if not all(isinstance(x, int) or x is unknown_length for x in shape):
+                raise TypeError("typetracer shape must be integers or unknown-length")
+            if not isinstance(dtype, np.dtype):
+                raise TypeError("typetracer dtype must be an instance of np.dtype")
         self._shape = shape
         self._dtype = dtype
 


jpivarski:
I forgot to say, cc: @martindurant and @lgray.

martindurant:
Please ping me when it's in. In dak, I can make the context manager right away, and it will have no downside: before this change it will simply have no effect.

jpivarski:
Does this really need to be a context manager in dask-awkward, @martindurant? Because it is now a public (though hidden) interface and Coffea can just

```python
awkward.typetracer.TypeTracerArray.runtime_typechecks = False
```

Being a context manager suggests that one would want to turn it off only for limited times, but I think Coffea wants to turn it off globally and only revisit that if there are any errors, right, @lgray?

martindurant:
With this, you have both possibilities. I would say that, generally, a library shouldn't change any state in another library except in limited scopes.

jpivarski:
Oh, that's right—Coffea might not be the only user of Awkward in a Python process. I was thinking of this as a global, set-it-once thing, like `vector.register_awkward()`, but that only adds behaviors to `ak.behavior` (which, in principle, might overshadow another library's, but in practice likely doesn't). This setting doesn't append state, it replaces state, and I can see how that would go wrong.

PR #3058, opened 2024-03-20
docs: new Try-It page based on plain Pyodide

jpivarski:
With the exception of the new `<script>`, this is a copy of https://pyodide.org/en/stable/console.html. Maybe there will be a way to link that in, maybe open it in an iframe and inject the new `<script>`.

The idea is for the "Try It!" button to open a new window or tab with a console that someone can use to try code in one window while reading the documentation in another.


jpivarski:
Does this work?

jpivarski:
@agoose77, this replaces the Try-It page with a console that has Awkward loaded, like we talked about in one of our Zoom calls. I'd like to make it open in a new tab, so that it's easy to access the console independently of the documentation, but there doesn't seem to be a way to do this for a MyST image target. Alternatively, it could reveal a hidden iframe on click (also don't know how to do that), but then you'd lose it when navigating around the documentation. At some point, the user would have to open two browser tabs or windows, anyway.

jpivarski:
I should also reintroduce that tutorial that you wrote as an ordinary page, too. It had good material. Or maybe I could modify the new try-it.html to be left and right side-by-side with the tutorial and the console, like on https://numpy.org/.

jpivarski:
It now opens a new tab and has a little arrow to indicate that this is the case. It's good!

agoose77:
Jim thanks for pushing on this, it fell unfortunately during my move else I'd have had more eyes on it.

The big change here is that we no longer deploy our WASM builds in a testable way for PRs. Maybe a longer-term development feature would be to build the WASM wheel separately to docs deployment for a pyodide-console deployment. We could even make that opt-in (triggering it via a label) to reduce our CI overhead on average.

jpivarski:
I was thinking about that as a follow-up PR. But on the other hand, testing the WASM compilation with every commits is as good of an idea as testing compilation on the other platforms. Maybe I should leave the infrastructure in just for that. 

In general, I've been thinking about doing a sprint to finally fill in all of the documentation.

PR #3059, opened 2024-03-21
feat: add boolean cases for CPU kernels

ManasviGoyal:
- Removed the following CPU Kernels and replaced their implementation with Numpy/Cupy -
  - awkward_unique
  - awkward_unique_copy
  - awkward_NumpyArray_fill

- Added boolean cases for same kernels
- Fixed and added some CUDA kernels
- Tests!


ManasviGoyal:
@jpivarski I have added a new CPU kernel but the kernel is not found and the tests here are failing. I am not sure what I am missing here.

ManasviGoyal:
@jpivarski I had removed `awkward_unique` and `awkward_unique_copy` and replaced it with `nplike.unique_values` which makes the CUDA implementation simple too. But I see that it is failing for ubuntu minimal test as the `equal_nan` argument requires `numpy >= 1.19.0`. Should I go back to using the kernels and add a separate case for the CUDA kernels or would it be possible to bump up the minimum version?

jpivarski:
As usual, replacing kernels with NumPy/CuPy equivalents is a good thing to do, and I don't want to return to the custom kernel now that this mostly works. 

One error, with the BitMaskedArray, might be because the whole `content` is being checked for uniqueness, rather than just the part up to the BitMaskedArray's `length`.

For the missing `equal_nan` argument in old versions of NumPy, you can implement it in the `nplike` layer, similar to the way that `byteorder` has been added to `packbits` for old NumPy versions. In the case of `equal_nan`, you would just need to check `isnan` for floating point types and do a `logical_and` of equality and `isnan`. It's not easy to check NumPy to see whether an argument exists (other than by version), but I don't think it would be bad to use this custom logic on all floating point types.

jpivarski (approved):
I verify that the CUDA tests all pass. All that's left is to fix the old-NumPy case (`np.unique` not having an `equal_nan` argument), and then this will be ready to merge.

jpivarski:
This looks like it's done. If so, please merge it. Thanks!

Issue #3056, opened 2024-03-19
`to_numpy` for option-type of record type fails because `_to_backend_array` is missing a case for structured arrays


jpivarski:
### Version of Awkward Array

HEAD

### Description and code to reproduce

This if-elif needs one more elif:

https://github.com/scikit-hep/awkward/blob/dfc9690537386ab690c7e7de1baa8e6b041632be/src/awkward/contents/indexedoptionarray.py#L1593-L1624

For NumPy structured arrays, the `dtype` is a list of fields, and we should make dummy data for every field in that list. That probably means that the whole if-elif needs to be factorized out into another function, so that the list case can recursively (1 level deep!) call into it for each of its fields.

jpivarski:
@zonca, a minimal reproducer without Kaitai would be great; it can be used for test-driven development and the test that will eventually be needed for pytest.

jpivarski:
```python
>>> import awkward as ak
>>> ak.Array([{"x": 1, "y": 1.1}, None])
<Array [{x: 1, y: 1.1}, None] type='2 * ?{x: int64, y: float64}'>
>>> ak.Array([{"x": 1, "y": 1.1}, None]).to_numpy()
```

```
Traceback (most recent call last):
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 62, in dispatch
    next(gen_or_result)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_to_numpy.py", line 48, in to_numpy
    return _impl(array, allow_missing)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_to_numpy.py", line 60, in _impl
    return numpy_layout.to_backend_array(allow_missing=allow_missing)
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/content.py", line 1024, in to_backend_array
    return self._to_backend_array(allow_missing, backend)
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/indexedoptionarray.py", line 1624, in _to_backend_array
    raise AssertionError(f"unrecognized dtype: {content.dtype}")
AssertionError: unrecognized dtype: [('x', '<i8'), ('y', '<f8')]

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jpivarski/irishep/awkward/src/awkward/highlevel.py", line 502, in to_numpy
[...]

PR #3060, opened 2024-03-22
fix: Support structured arrays in option type to_numpy

zonca:
- [test] implement unit test for `to_numpy` on option-type of record type
- [fix] handle structured arrays for option-type

See https://github.com/scikit-hep/awkward/issues/3056

### DIFF
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -1594,26 +1594,15 @@ class IndexedOptionArray(IndexedOptionMeta[Content], Content):
 
                 data[index_nplike.logical_not(mask0)] = content
 
-                if np.issubdtype(content.dtype, np.bool_):
-                    data[mask0] = False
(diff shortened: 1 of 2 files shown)


ManasviGoyal:
@zonca Instead of adding the test to an existing test file, you should add a new test according to the contributing guidelines.

```
tests/test_XXXX-yyyy.py
where XXXX is either the number of the issue your pull request fixes or the number of the pull request and yyyy is descriptive text
```

zonca:
> @zonca Instead of adding the test to an existing test file, you should add a new test according to the contributing guidelines.
> 
> ```
> tests/test_XXXX-yyyy.py
> where XXXX is either the number of the issue your pull request fixes or the number of the pull request and yyyy is descriptive text
> ```

Thanks @ManasviGoyal, fixed that

jpivarski (approved):
That does it, thank you! Since this appears to be a complete solution, with test, and I don't think you're adding any more to it, I'm going to merge it now.

PR #3064, opened 2024-03-27
feat: Allow interoperability with numpy 2.x

tcawlfield:
We would like to allow awkward to work correctly with numpy 2.x.

PR #3065, opened 2024-03-28
test: cuda integration tests

ManasviGoyal:


jpivarski:
I want you to be aware that I removed two kernel-test-data.json tests in 2bcb015a2f21496d485ec88c3e65a8a55618fd83 (https://github.com/scikit-hep/awkward/pull/3064#issuecomment-2030104802) because they were failing—that's probably important. Does the `status` flag turn on and off tests without having to delete them? (It's too bad JSON can't be commented out the way that YAML can.)

src/awkward/_connect/cuda/cuda_kernels/awkward_UnionArray_regular_index_getsize.cu
```diff
@@ -18,6 +20,6 @@ awkward_UnionArray_regular_index_getsize(
     uint64_t invocation_index,
     uint64_t* err_code) {
   if (err_code[0] == NO_ERROR) {
-    *size = length > 0 && *size > 0 ? *size + 1 : 1;
+    *size = *size + 1;
```
jpivarski, line 23: I checked this against [the specification](https://awkward-array.org/doc/main/reference/generated/kernels.html#awkward-unionarray-regular-index-getsize) to verify that there's no condition on whether `*size` should be incremented. It looks good!

tests-cuda/test_2922b_new_cuda_kernels.py
```diff
@@ -1444,68 +1443,6 @@ def test_0193_is_none_axis_parameter():
...
-        3,
-    )
-    cuda_a = ak.to_backend(a, "cuda", highlevel=False)
-
-    assert len(cuda_a) == 2
-    cuda_a = cuda_a.maybe_to_NumpyArray()
-    assert isinstance(
-        cuda_a[1,],
-        ak.contents.numpyarray.NumpyArray,
-    )
```
jpivarski, line 1506: This is a big test to remove. What happened? Is it something that can be temporarily skipped or xfailed, or is the whole idea misguided? (Or is this test actually just moved to another file?

jpivarski (approved):
I ran all the new tests, and they all work on my GPU.

```
====================== test session starts =======================
platform linux -- Python 3.10.14, pytest-8.1.1, pluggy-1.4.0
Matplotlib: 3.8.3
Freetype: 2.12.1
rootdir: /home/jpivarski/irishep/awkward
configfile: pyproject.toml
plugins: forked-1.6.0, typeguard-4.2.1, xdist-3.5.0, reverse-1.7.0, mpl-0.17.0, mock-3.14.0, timeout-2.2.0, cov-5.0.0, anyio-4.3.0
collected 463 items                                              

tests-cuda/test_1276_cuda_num.py .........                 [  1%]
tests-cuda/test_1276_cuda_transfers.py ................    [  5%]
tests-cuda/test_1276_cupy_interop.py .                     [  5%]
tests-cuda/test_1276_from_cupy.py .....                    [  6%]
tests-cuda/test_1300_same_for_numba_cuda.py .............. [  9%]
.........                                                  [ 11%]
tests-cuda/test_1381_check_errors.py .                     [ 11%]
tests-cuda/test_1809_array_cuda_jit.py ..............      [ 14%]
tests-cuda/test_2922a_new_cuda_kernels.py ................ [ 18%]
......................................................     [ 30%]
tests-cuda/test_2922b_new_cuda_kernels.py ................ [ 33%]
.............                                              [ 36%]
tests-cuda/test_3065a_cuda_kernels.py .................... [ 40%]
.......................................................... [ 53%]
.......................................................... [ 65%]
.......................................................... [ 78%]
.........................................                  [ 87%]
tests-cuda/test_3065b_cuda_kernels.py .................... [ 91%]
....                                                       [ 92%]
tests-cuda/test_3065c_cuda_kernels.py .................... [ 96%]
................                                           [100%]

====================== 463 passed in 12.90s ======================
```

(as well as all the kernel unit tests). I have some questions below, but no complaints. Go ahead and merge this PR whenever you're ready!

tests-cuda/test_2922b_new_cuda_kernels.py
ManasviGoyal: I am not sure exactly but this seemed to cause a segfault as it was causing the subsequent tests to fail. I have a list of such tests for which I need to figure out what if the error is in some kernels or someplace else. I can add it with the next batch of tests.

ManasviGoyal:
> I want you to be aware that I removed two kernel-test-data.json tests in [2bcb015](https://github.com/scikit-hep/awkward/commit/2bcb015a2f21496d485ec88c3e65a8a55618fd83) ([#3064 (comment)](https://github.com/scikit-hep/awkward/pull/3064#issuecomment-2030104802)) because they were failing—that's probably important. Does the `status` flag turn on and off tests without having to delete them? (It's too bad JSON can't be commented out the way that YAML can.)

@jpivarski The status flag turns off all the tests for a particular kernel. We can't turn off off a specific test with that. 

I saw your comment about that earlier. I checked and they were failing only for Python kernels. Were there any errors in the cuda tests too on your GPU?

Due to the changes to promotion in Numpy 2.x, it gives `OverflowError: Python integer 256 out of bounds for uint8`. It needs to be explicitly promoted now. I will create a issue for this and fix it separately.

src/awkward/_connect/cuda/cuda_kernels/awkward_UnionArray_regular_index_getsize.cu
ManasviGoyal: This kernel was causing illegal memory access error for some edge cases, so I just fixed that.

jpivarski:
> Were there any errors in the cuda tests too on your GPU?

Those tests are included in this branch, right? If so, then I have tested them on the GPU and they work.

> Due to the changes to promotion in Numpy 2.x, it gives `OverflowError: Python integer 256 out of bounds for uint8`. It needs to be explicitly promoted now. I will create a issue for this and fix it separately.

That's very likely; we only saw this error when upgrading to NumPy 2.0. Since I had to get a version of Awkward out with NumPy 2.0 support, I had to drop the tests, but I didn't want them to get forgotten about.

tests-cuda/test_2922b_new_cuda_kernels.py
jpivarski: Okay, if you're getting back to it, then that's fine.

ManasviGoyal:
> Those tests are included in this branch, right? If so, then I have tested them on the GPU and they work.

No, I meant the tests that were removed. As for me those tests were failing only for the Python kernel and not CPU and CUDA kernels.

> That's very likely; we only saw this error when upgrading to NumPy 2.0. Since I had to get a version of Awkward out with NumPy 2.0 support, I had to drop the tests, but I didn't want them to get forgotten about.

I have opened up a issue #3071 and self- assigned it. I will fix it in the coming week!

PR #3068, opened 2024-04-01
fix: Windows and 32-bit Linux default to 32-bit integers in sums and prods UNLESS it's NumPy 2.0

jpivarski:
Caught in this test: https://github.com/scikit-hep/awkward/actions/runs/8511994785/job/23312715586

Although we could check all platforms with

```python
numpy._module.sum([given_dtype(0)]).dtype.type
```

I want to limit that to only the unusual platforms (Windows and 32-bit Linux) because the others are a quick lookup, and I don't want to slow down metadata-checking any more.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,6 +65,14 @@ jobs:
             python-architecture: x64
             runs-on: ubuntu-latest
             dependencies-kind: numpy2
+          - python-version: '3.11'
+            python-architecture: x64
+            runs-on: macos-11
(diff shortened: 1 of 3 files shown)


jpivarski:
This is a good hint that we should test NumPy 2.0 on all platforms. NumPy used to have a 32-bit/64-bit split on Windows only, but now its dtypes are consistent across all platforms. However, that means that we need to be ready for it, too.

tcawlfield (approved):
LGTM

PR #3072, opened 2024-04-03
fix: restrict ak.softmax to axis=-1, the only axis for which the calculation is correct right now

### LINKED ISSUE #2760: `ak.softmax` with `axis=0` fails

jpivarski:


tcawlfield (approved):
This looks correct to me.

PR #3073, opened 2024-04-04
fix: Numpy 2.0 out of bounds error for BitMaskedArray Py kernels

### LINKED ISSUE #3071: Numpy 2.x causes out of bounds error for BitMaskedArray kernels
### Version of Awkward Array

2.6.3

Due to the changes to promotion in Numpy 2.x, some tests for BitMaskedArray Python kernels gives:
`OverflowError: Python integer 464 out of bounds for uint8`

It needs to be explicitly promoted now.

ManasviGoyal:
- fixes #3071 
- added additional tests

### DIFF
--- a/kernel-specification.yml
+++ b/kernel-specification.yml
@@ -35,19 +35,19 @@ kernels:
               for i in range(bitmasklength):
                   byte = frombitmask[i]
                   tobytemask[(i * 8) + 0] = ((byte & uint8(128)) != 0) != validwhen
-                  byte <<= 1
+                  byte = (byte << 1) & 0xFF
                   tobytemask[(i * 8) + 1] = ((byte & uint8(128)) != 0) != validwhen
-                  byte <<= 1
+                  byte = (byte << 1) & 0xFF
                   tobytemask[(i * 8) + 2] = ((byte & uint8(128)) != 0) != validwhen
-                  byte <<= 1
+                  byte = (byte << 1) & 0xFF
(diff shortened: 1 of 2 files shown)


kernel-specification.yml
```diff
@@ -35,19 +35,19 @@ kernels:
               for i in range(bitmasklength):
                   byte = frombitmask[i]
                   tobytemask[(i * 8) + 0] = ((byte & uint8(128)) != 0) != validwhen
-                  byte <<= 1
+                  byte = (byte << 1) & 0xFF
```
jpivarski, line 38: That makes sense: a hardware `<<` instruction would clamp to the integer size automatically (because there's nowhere else for the bits to go), but Python's software `<<` instructions are implemented in arbitrary-precision arithmetic, so the "byte" can hold a value bigger than 255.

I'm sure that this was intended as a convenience, but it means we need to be extra careful when applying any bitwise operation (`<<`, `>>`, `&`, `|`, `~`) to a Python integer. Results of bitwise operations can depend on bit width (consider `~`, especially) but the notion of "bit width" can be nebulous if Python's arbitrarily increasing it to fit the integer value.

In normal Python code, I try to avoid this by casting integers as NumPy integers with a specific width, such as `np.uint8` or `np.int64`. In the kernel specification code, however, we should use only basic Python so what you have here is ideal.

jpivarski (approved):
Great! We talked about it in our Awkward-Uproot meeting. Also, having more tests that before is even better.

This is now mergeable, and it looks clear to me that the PR is done (you don't have any more edits), so I'll just merge it. Thanks!

Issue #3077, opened 2024-04-11
ATLAS `ElementLinks` used to work, NOT anymore


ekourlit:
### Version of Awkward Array

2.6.2

### Description and code to reproduce

I'm writing this on behalf of a team within the ATLAS Collaboration who tries to read ATLAS-specific data (PHYSLITE) with `uproot` and `awkward`. The problem is related to `ElementLinks`.

We use `coffea` and the `PHYSLITESchema` to load and structure our data in awkward arrays. `ElementLinks` are also treated within the `PHYSLITESchema`. This is something that it **used to work fine** in the past with:
```
coffea version: 2023.7.0rc1.dev83+g52950d1
uproot version: 5.0.13
awkward version: 2.4.4
```
but it is **not** working anymore with:
```
coffea version: 2024.3.0
uproot version: 5.3.2
awkward version: 2.6.2
```

When we try to retrieve the linked collection from the parent one we get the following error:
```
Traceback (most recent call last):
  File "/venv/lib/python3.9/site-packages/awkward/_dispatch.py", line 56, in dispatch
    raise NotImplementedError
NotImplementedError

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/ekourlit/failing_example.py", line 67, in <module>
    clusters = events.Electrons.caloClusters
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 1535, in __getattr__
    return self.map_partitions(
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 1588, in map_partitions
    return map_partitions(func, self, *args, traverse=traverse, **kwargs)
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 2107, in map_partitions
    return _map_partitions(
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 1943, in _map_partitions
    meta = map_meta(fn, *args, **kwargs)
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 2467, in map_meta
    meta = fn(*to_meta(deps))
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 1903, in __call__
    return self.fn(*args, **kwargs)
  File "/venv/lib/python3.9/site-packages/dask_awkward/lib/core.py", line 2575, in __call__
    return getattr(coll, self.attr)
[...]

PR #3079, opened 2024-04-11
fix: refuse dak.Array in hidden-but-public-functions

jpivarski:
This is solution (3) from https://github.com/scikit-hep/awkward/issues/3077#issuecomment-2050099114, but I made it a "refuse if dask-awkward," rather than "convert if dask-awkward." They're hidden-but-public (L2) functions, so they can be restrictive. If dask-awkward ever does write overloads for them, those overloads can be a way to avoid this behavior (so we're not making decisions _for_ dask-awkward).

To fix the ATLAS problem, Coffea will need to make an update, to be more careful about `dak.Array` versus `dak.Array._meta`.

### DIFF
--- a/src/awkward/operations/ak_backend.py
+++ b/src/awkward/operations/ak_backend.py
@@ -14,7 +14,7 @@ def backend(*arrays):
     Args:
         arrays: Array-like data (anything #ak.to_layout recognizes).
 
-    Returns the names of the backend used by `arrays`. This name may be
+    Returns the name of the backend used by `arrays`. This name may be
 
     * `"cpu"` for arrays backed by NumPy;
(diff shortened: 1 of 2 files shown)


jpivarski:
My main question: are there other places in the code like this, which should have a similar dak-check?

I decided to go with `type(x).__name__.startswith("dask_awkward.")`, rather than `hasattr(x, "_meta")`, since there may be other reasons why a non-dask-awkward array has an attribute named `_meta`. If class definitions get moved in dask-awkward, then this check should really be

```python
type(x).__name__ == "dask_awkward" or type(x).__name__.startswith("dask_awkward.")
```

and maybe it should be a utility function.

If there's already a utility function that I just don't know about, please point me to it!

PR #2179, opened 2023-01-31
fix: rework parameter merging rules

agoose77:
## TL;DR

This PR fixes #2173 by two main changes: make `EmptyArray` a true identity (by ignoring parameters), and consolidate the parameter merging rules.

- Fixes #2173 
- Collapses `_mergeable` and `_mergeable_next` into `_mergeable_next`
- Replace `merge_parameters` with `_parameters_intersect` and `_parameters_union`
- Changes index/option merging rules:
   - option/index — option/index: merge option parameters, merge content parameters
   - option/index — non-option/non-index: keep option parameters, merge content parameters
- Changes union merging rules: 
  - Now always take intersection of merged content parameters
- Changes to union simplification logic:
  - Take union of outermost union parameters and inner union parameters
- Make the base `Content` and `Form` initialiser arguments keyword-only (non-breaking unless you are subclassing)
- Deprecates parameters on empty/unknown types/arrays/forms
- Treats `EmptyArray` as entirely invisible to merging
 

## TL;
This PR is a tentative direction for resolving #2173 by making two big changes:

### Consider `EmptyArray` as truly "empty"
The issue is caused by the fact that `EmptyArray` often has no parameters, which can occur for layouts with only-options, e.g. from Parquet (?) or `ArrayBuilder`. If we try and merge parameters with such layouts, the `merge_equal` rule will lead to the resulting parameter dictionary being empty. This means that we lose `__array__`.

A more permissive rule would fix #2173 is to *only* merge the non-array/record parameters of `EmptyArray`. Such a rule would preserve e.g. strings, but fail to preserve any other parameters whenever a bare `EmptyArray` is encountered. 

I prefer the blanket "EmptyArray-is-empty" approach, but I don't have a good feel for whether that's too restrictive. 

### Index/Option merging rules

In https://github.com/scikit-hep/awkward/discussions/2180 I raise the idea that we need to tighten our parameter rules. This PR makes a particular choice to treat (option, non-option) as merging only the contents. This will make it more obvious that our current logic for `__array__` doesn't look past the top-level node.

### Union merging/simplification

The simple logic followed in this PR is
- `simplified` → union of parameters in simplification
- `merge` → intersection of parameters

Changes to union merging/simplification now reflect this.

PR #3085, opened 2024-04-19
fix: support conversion to arrow and back with non-option Unknown type

### LINKED ISSUE #2340: Awkward allows non-nullable unknown type, but Arrow doesn't
In trying to construct a minimal reproducer, I find that they directly forbid a null type from being non-nullable:

```python
>>> import pyarrow as pa
>>> pa.field("name", pa.null(), nullable=False)
  File "pyarrow/types.pxi", line 2266, in pyarrow.lib.field
ValueError: A null type field may not be non-nullable
```

In our code, we must be getting at it some other way that bypasses this check, but it's clear what their intentions are. We _do_ allow an `UnknownType` to not be inside an `OptionType`, so the problem is that we have a broader type system and a direct conversion makes something that Arrow doesn't consider legal.

So the right way to fix this is to wrap our `EmptyArray` inside `UnmaskedArray` on conversion to Arrow, but include enough metadata in the `ExtensionType` that when we convert it back, we know to remove the option-type, so that it's round-trip preserved. I'll make this an issue.

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward/issues/2337#issuecomment-1482892635_

tcawlfield:
When we convert an Awkward array of type Unknown, this becomes
an arrow array (even prior to this) with zero length, nullable-null type,
but with Awkward metadata indicating no mask (non-option-type).
I believe in this case we can be sure that the original Awkward array was an EmptyArray.


src/awkward/_connect/pyarrow.py
```diff
@@ -610,6 +614,16 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
         validbits = buffers.pop(0)
         assert storage_type.num_fields == 0
 
+        if (
+            awkwardarrow_type
```
jpivarski: I don't like to use Python's truthiness; if we're checking for whether an object is `None`, I'd rather be explicit about that (and not get burned by empty lists or whatever):

```suggestion
            awkwardarrow_type is not None
```

src/awkward/_connect/pyarrow.py
```diff
@@ -610,6 +614,16 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
...
+            and awkwardarrow_type._is_nonnullable_nulltype
+            and len(paarray) == 0
```
jpivarski: This is an unnecessary check because `_is_nonnullable_nulltype` can _only_ ever apply to EmptyArrays. By having this check here, it suggests that this might sometimes not be true, which could send a future maintainer down a rabbit-hole.

```suggestion
```

src/awkward/_connect/pyarrow.py
```diff
@@ -903,6 +925,9 @@ def is_revertable(akarray):
 
 
 def remove_optiontype(akarray):
+    if not is_revertable(akarray):
+        # In the case of the Unknown type, we get an EmptyArray at this point.
+        return akarray
```
jpivarski: Instead of having a special case here, the

```python
            return ak.contents.EmptyArray()
```

on new line 625 could include a `__pyarrow_original` just like all of the other nodes:

```python
            out = ak.contents.EmptyArray()
            out.__pyarrow_original = None
            return
```

That would keep the changes more self-contained to the parts that are new. This `remove_optiontype` doesn't have much to do with the new functionality.

src/awkward/contents/emptyarray.py
```diff
@@ -83,8 +83,7 @@ def __getitem__(self, where):
     """
 
     def __init__(self, *, parameters=None, backend=None):
-        if not (parameters is None or len(parameters) == 0):
-            raise TypeError(f"{type(self).__name__} cannot contain parameters")
+        self._enforce_parameters(parameters, type(self).__name__)
```
jpivarski: I see that this is refactoring PR #2179 into a utility function. That's not related to this PR and I think it's a bit noisy to have a test that would be repeated only 3 times to be factored out like this. A future maintainer would be led to wonder if anything else (taking `type(self).__name__`) would be enforced to have no parameters (they wouldn't; this is the only node type that might do that).

src/awkward/contents/emptyarray.py
```diff
@@ -97,8 +96,7 @@ def copy(
         parameters=UNSET,
         backend=UNSET,
     ):
-        if not (parameters is UNSET or parameters is None or len(parameters) == 0):
-            raise TypeError(f"{type(self).__name__} cannot contain parameters")
+        self._enforce_parameters(parameters, type(self).__name__)
```
jpivarski: Following on the above.

src/awkward/contents/emptyarray.py
```diff
@@ -109,10 +107,16 @@ def __copy__(self):
     def __deepcopy__(self, memo):
         return self.copy()
 
+    @classmethod
+    def _enforce_parameters(cls, parameters, instance_name):
+        if not (parameters is UNSET or parameters is None or len(parameters) == 0):
+            raise TypeError(
+                f"{instance_name} does not support parameters (given {parameters})"
+            )
+
```
jpivarski: Following on the above.

src/awkward/contents/emptyarray.py
```diff
@@ -109,10 +107,16 @@ def __copy__(self):
...
     @classmethod
     def simplified(cls, *, parameters=None, backend=None):
-        if not (parameters is None or len(parameters) == 0):
-            raise TypeError(f"{cls.__name__} cannot contain parameters")
+        cls._enforce_parameters(parameters, cls.__name__)
```
jpivarski: Following on the above.

src/awkward/_meta/meta.py
```diff
@@ -118,3 +118,7 @@ def dimension_optiontype(self) -> bool:
 
     def copy(self, *, parameters: JSONMapping | None | Sentinel = UNSET) -> Self:
         raise NotImplementedError
+
+    @classmethod
+    def _arrow_needs_option_type(cls):
+        return cls.is_option
```
jpivarski: This should happen in `Content`, rather than `Meta`.

`Meta` is a superclass of `Content` and `Form`, and you're only overriding this in `EmptyArray` (a subclass of `Content`), not `EmptyForm` (a subclass of `Form`).

src/awkward/contents/emptyarray.py
```diff
@@ -364,11 +368,12 @@ def _to_arrow(
...
-                    options["extensionarray"],
-                    options["record_is_scalar"],
-                    mask_node,
-                    self,
+                    storage_type=pyarrow.null(),
+                    use_extensionarray=options["extensionarray"],
+                    record_is_scalar=options["record_is_scalar"],
+                    mask=mask_node,
+                    node=self,
+                    is_nonnullable_nulltype=True,
```
jpivarski: You don't want this to be `True` for every EmptyArray—you want it to be `True` for those EmptyArrays that are artificially wrapped in UnmaskedArray. If the EmptyArray had already been nested within some other option type, you want `is_nonnullable_nulltype=False`, so that you don't turn what _should_ be `?unknown` into `unknown` when reading back from Arrow.

jpivarski (changes requested):
The only really important issue is on emptyarray.py new line 376: you're setting `is_nonnullable_nulltype=True` for every EmptyArray.

If it's originally a non-optional `unknowntype`, you want to wrap it in UnmaskedArray and set `is_nonnullable_nulltype=True` so that when  it gets read back from Arrow, the fake UnmaskedArray gets taken off again.

If it's originally a `?unknowntype` (the `?` is option-type), you want to _not_ wrap it in UnmaskedArray (it's already in an option-tyep) and set `is_nonnullable_nulltype=True` so that when it gets read back from Arrow, nothing gets taken off again.

To ensure that this behavior gets tested, you should have some originally `?unknowntype` examples. You're currently testing

* `[]` of `unknowntype`
* `[[], [], []]` of `var * unknowntype`

Add to that

* `[]` of `?unknowntype`: you can make it with `ak.Array([None])[0:0]`
* `[[], [], []]` of `var * ?unknowntype`: you can make it with `ak.Array([[], [], [], [None]])[0:3]`.

The rest of my comments are minor coding style preferences.

src/awkward/_connect/pyarrow.py
tcawlfield: Ahh, yes, after addressing your first suggestion (_is_nonnullable_nulltype is only True if it's not option-enclosed), this check became redundant.

src/awkward/contents/emptyarray.py
tcawlfield: That's a fair point. I'll try to back that change out. I added `_enforce_parameters` earlier when I was adding "is_empty_array" to node_parameters, at which point _enforce_parameters was complicated enough to not want to just copy+paste. Then when removing that stuff, lazy-me felt "Eh, this can be argued to be an improvement" even if only barely.

I really don't feel strongly about this one way or another, but the counter-argument to your point is to adopt a kind of policy that any code changes that are thought to be cleanup or simplification are encouraged to happen surrounding any code that one revisits and studies in making a PR. That's a loose policy that sacrifices succinct PRs for gains in code polish -- allowing changes that are too minor to want to formalize into an issue and PR.

Whether or not that sounds reasonable, is what I've done here specifically an improvement or is it noise? It's not obvious to me either way, and I won't argue for it.

src/awkward/_meta/meta.py
tcawlfield: I put this in Content originally, but then realized that the is_option property is defined in Meta, not Content. So it seemed more correct to put `return cls.is_option` in the same place that I could see `is_option`. Also, *is* Meta a superclass of Content? Oh! Yes it is but I thought differently after reading:
```
class EmptyArray(EmptyMeta, Content):
```
But for full disclosure, while I can guess at what `Content` means, I really don't know what `Meta` is for. I feel like I'm still stumbling in the dark rather than coding with a full understanding. I really don't like doing that, but I suppose that for now it's necessary.

src/awkward/contents/emptyarray.py
tcawlfield: You're absolutely right. I'm changing this to `mask_node is None`. Yet again I wish I had a rock-solid understanding of what I'm doing here, but this *feels* right. If a mask is provided, then you're inside an option type. Unless maybe you're in an UnmaskedArray? Argh! I'm not unit-testing yet.

src/awkward/_connect/pyarrow.py
tcawlfield: I did ... something here although it feels a little scary. Curious what you think!

While I did not really like my `if not is_revertable` line, in a very limited way it seems reasonable: the very next line assumes akarray has a __pyarrow_original property even though that's not a property of Content. Things that prevent crashes seem wise, *if* there is a clearly correct way to handle that case. It felt like maybe I was doing just that.

Still, with my addition of `revertable(out, out)` (again, cringe?) this check seems okay to remove.

src/awkward/_meta/meta.py
tcawlfield: Okay, this now lives in Content. :smile:

jpivarski (approved):
This is done! I checked the code carefully and I tested a few manually constructed arrays:

```python
>>> import numpy as np
>>> import awkward as ak
>>> from awkward.contents import EmptyArray, UnmaskedArray, ListOffsetArray
>>> from awkward.index import Index64
>>>
>>> ak.from_arrow(ak.to_arrow(EmptyArray(), extensionarray=False), highlevel=False)
<IndexedOptionArray len='0'>
    <index><Index dtype='int64' len='0'>[]</Index></index>
    <content><EmptyArray len='0'/></content>
</IndexedOptionArray>
>>>
>>> ak.from_arrow(ak.to_arrow(EmptyArray(), extensionarray=True), highlevel=False)
<EmptyArray len='0'/>
```

That's good: with `extensionarray=False`, it can't round-trip: the EmptyArray comes back as an option-type (IndexedOptionArray) EmptyArray, but that's because it has nowhere to store the metadata. Meanwhile, the `extensionarray=True` case _does_ round-trip: it's using that metadata.

Same thing inside of a list:

```python
>>> ak.from_arrow(ak.to_arrow(ListOffsetArray(Index64(np.array([0])), EmptyArray()), extensionarray=False), highlevel=False)
<ListOffsetArray len='0'>
    <offsets><Index dtype='int64' len='1'>[0]</Index></offsets>
    <content><IndexedOptionArray len='0'>
        <index><Index dtype='int64' len='0'>
            []
        </Index></index>
        <content><EmptyArray len='0'/></content>
    </IndexedOptionArray></content>
</ListOffsetArray>
>>>
>>> ak.from_arrow(ak.to_arrow(ListOffsetArray(Index64(np.array([0])), EmptyArray()), extensionarray=True), highlevel=False)
<ListOffsetArray len='0'>
    <offsets><Index dtype='int64' len='1'>[0]</Index></offsets>
    <content><EmptyArray len='0'/></content>
</ListOffsetArray>
```

Without an `extensionarray`, it can't store the metadata to know that the EmptyArray originally wasn't wrapped in any option-type node, but with an `extensionarray`, it knows that and reproduces the original type.

Now if we wrap the EmptyArray in an option-type (UnmaskedArray), it always comes back with the option-type, regardless of the `extensionarray` setting.

```python
>>> ak.from_arrow(ak.to_arrow(UnmaskedArray(EmptyArray()), extensionarray=False), highlevel=False)
<IndexedOptionArray len='0'>
    <index><Index dtype='int64' len='0'>[]</Index></index>
    <content><EmptyArray len='0'/></content>
</IndexedOptionArray>
>>>
>>> ak.from_arrow(ak.to_arrow(UnmaskedArray(EmptyArray()), extensionarray=True), highlevel=False)
<IndexedOptionArray len='0'>
    <index><Index dtype='int64' len='0'>[]</Index></index>
    <content><EmptyArray len='0'/></content>
</IndexedOptionArray>
>>>
>>> ak.from_arrow(ak.to_arrow(ListOffsetArray(Index64(np.array([0])), UnmaskedArray(EmptyArray())), extensionarray=False), highlevel=False)
<ListOffsetArray len='0'>
    <offsets><Index dtype='int64' len='1'>[0]</Index></offsets>
    <content><IndexedOptionArray len='0'>
        <index><Index dtype='int64' len='0'>
            []
        </Index></index>
        <content><EmptyArray len='0'/></content>
    </IndexedOptionArray></content>
</ListOffsetArray>
>>>
>>> ak.from_arrow(ak.to_arrow(ListOffsetArray(Index64(np.array([0])), UnmaskedArray(EmptyArray())), extensionarray=True), highlevel=False)
<ListOffsetArray len='0'>
    <offsets><Index dtype='int64' len='1'>[0]</Index></offsets>
    <content><IndexedOptionArray len='0'>
        <index><Index dtype='int64' len='0'>
            []
        </Index></index>
        <content><EmptyArray len='0'/></content>
    </IndexedOptionArray></content>
</ListOffsetArray>
```

The unit tests test these cases as well, but not explicitly with UnmaskedArray. (Constructing it from Python lists makes the option-type with IndexedOptionArray, but the code isn't hard-coded to any particular node type, just the option-ness.)

So I think this is ready to merge!

src/awkward/contents/emptyarray.py
```diff
@@ -364,11 +364,12 @@ def _to_arrow(
...
-                    options["extensionarray"],
-                    options["record_is_scalar"],
-                    mask_node,
-                    self,
+                    storage_type=pyarrow.null(),
+                    use_extensionarray=options["extensionarray"],
+                    record_is_scalar=options["record_is_scalar"],
+                    mask=mask_node,
+                    node=self,
+                    is_nonnullable_nulltype=mask_node is None,
```
jpivarski, line 372: This was the key point. It's fixed now, so I'll resolve the conversation about it.

src/awkward/_connect/pyarrow.py
jpivarski: `revertable(out, out)` is fine: it's revertable to itself. When/if the reverting process comes along, it will get replaced by itself (i.e. nothing happens).

The value here is that now we have trees in which either every node has a `__pyarrow_original` or every nod does not have a `__pyarrow_original`, and so the code that removes these things doesn't have to be special-cased.

This is just a matter of keeping like with like.

PR #3087, opened 2024-04-19
fix: distributed rdataframe is not yet support

### LINKED ISSUE #3061: from_rdataframe does not work for distributed rdataframe
### Version of Awkward Array

2.6.2

For distributed RDataFrames, e.g. with dask or spark backends, some operations are not exposed including `GetColumnType` which is used by the `ak.from_rdataframe` method.

Reproducer:
```
import ROOT
import awkward as ak
from dask.distributed import LocalCluster, Client

RDataFrame = ROOT.RDF.Experimental.Distributed.Dask.RDataFrame

cluster = LocalCluster(n_workers=4, threads_per_worker=1, processes=True)
client = Client(cluster)
rdf = RDataFrame(1000, daskclient=client).Define("x", "1")

array = ak.from_rdataframe(rdf, columns=['x'])
```

which gives the error:

```
AttributeError: 'Define' object has no attribute 'GetColumnType'

    ak.from_rdataframe(
        NodeProxy-instance
        columns = ['x']
    )
```

One way to solve this will be to access the `GetColumnType` method internally (this is not ideal, an [issue ](https://github.com/root-project/root/issues/15039)is submitted to ROOT github):
```
rdf._headnode._localdf.GetColumnType
```

ianna:
add a clear message

### DIFF
--- a/src/awkward/_connect/rdataframe/from_rdataframe.py
+++ b/src/awkward/_connect/rdataframe/from_rdataframe.py
@@ -65,6 +65,9 @@ cppyy.include("rdataframe/jagged_builders.h")
 def from_rdataframe(
     data_frame, columns, highlevel, behavior, with_name, offsets_type, keep_order
 ):
+    if hasattr(data_frame, "proxied_node"):
+        raise NotImplementedError("Distributed RDataFrame is not yet supported")
+
     def cpp_builder_type(depth, data_type):
         if depth == 1:
             return f"awkward::LayoutBuilder::Numpy<{data_type}>>"


ianna:
@jpivarski - is someone looking into pypy segfaults? Thanks

jpivarski:
> @jpivarski - is someone looking into pypy segfaults? Thanks

Not at the moment. If you have a chance to look into #2971, that would be great.

ianna (commented):
@jpivarski - as discussed yesterday we are making it clear that distributed RDataFrame is not supported yet.

jpivarski (approved):
If only distributed RDataFrames have the `proxied_node` attribute, this is a good way to catch it. It sounds like that might be true, but would checking the data type be more robust? E.g.

```python
if type(data_frame).__name__ == "DistributedRDataFrame":
```

or something similar?

Either way, this is good to merge. I'll let you decide how to handle it.

PR #3091, opened 2024-04-23
fix: is_valid checks in header-only library

zonca:
The `is_valid` methods implemented for `Indexed`, `IndexedOption` and `Union` are not performing the right checks, see details by @jpivarski in https://github.com/det-lab/kaitai_struct_awkward_runtime/pull/39#issuecomment-2073035561

Implemented tests that confirm these bugs, will implement later a fix.


jpivarski:
Let me know if you have a time-estimate for this.

I noticed that there hasn't been an Awkward release in a while, and we'd need a new awkward-cpp release before that happens. ([I just checked](https://github.com/scikit-hep/awkward/actions/runs/8835968808).) Building awkward-cpp takes a while (about 4 hours), and it uses up PyPI space, so I'd rather do it once than twice.

This PR will be touching awkward-cpp, so if it will be done in a few days, then I'll plan on building awkward-cpp when this PR is done. If it's going to be longer than a few days, I'll do awkward-cpp now.

Thanks!

zonca:
@jpivarski I plan to get this done by Tuesday at the latest

jpivarski:
Sounds good to me! I'll wait.

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -1151,6 +1151,14 @@ namespace awkward {
         return content_;
       }
 
+      /// @brief Inserts an explicit value in the `index` buffer and
+      /// returns the reference to the builder content.
+      BUILDER&
+      append_index(size_t i) noexcept {
+        index_.append(i);
+        return content_;
+      }
```
zonca, line 1167: I think this is a necessary method for categorical

header-only/tests/test_3091-layout-builder-is-valid.cpp
```diff
@@ -0,0 +1,116 @@
+// BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
```
zonca, line 1: I built tests for Indexed and IndexedOption, unfortunately I couldn't figure out how to build a unit test for `Union` that made sense, so I am just relying on `is_valid` calls in the layout builder test file. I checked in the debugger and it seems to do what I expected.

header-only/layout-builder/awkward/LayoutBuilder.h
jpivarski: Okay!

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -1211,17 +1219,23 @@ namespace awkward {
...
-              << content_.length() << " but index has length " << index_.length()
-              << "\n";
-          error.append(out.str());
 
-          return false;
-        } else {
-          return content_.is_valid(error);
+        std::unique_ptr<PRIMITIVE[]> ptr(new PRIMITIVE[index_.length()]);
+        index_.concatenate(ptr.get());
+        // check that each element of index_ is >= 0 and < content_.length()
```
jpivarski: This would also be possible without the $\mathcal{O}(n)$ check by keeping track of the maximum index value as you go along, instead of checking all the values at the end.

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -1411,17 +1434,22 @@ namespace awkward {
...
       is_valid(std::string& error) const noexcept {
-        if (content_.length() != last_valid_ + 1) {
-          std::stringstream out;
-          out << "IndexedOption node" << id_ << " has content length "
-              << content_.length() << " but last valid index is " << last_valid_
-              << "\n";
-          error.append(out.str());
+        std::unique_ptr<PRIMITIVE[]> ptr(new PRIMITIVE[index_.length()]);
+        index_.concatenate(ptr.get());
+        // check that each element of index_ is  < content_.length()
```
jpivarski: Same here.

header-only/tests/test_3091-layout-builder-is-valid.cpp
jpivarski: Okay, I think that's there aren't any uses of Union in Kaitai yet, so this can be something to be reevaluated later, when there's a use case.

jpivarski (approved):
This is good, but you can avoid a loop over the whole array at the end of processing by keeping track of the maximum index while processing. If the `is_valid` is only called once, however, then it doesn't matter all that much. 

We can merge this in its current state, but I'll let you decide if you want to do the thing with accumulating a maximum index first.

header-only/layout-builder/awkward/LayoutBuilder.h
zonca: is there an alternative to `concatenate` that I can use to achieve this?

zonca: or do you mean modify `append_index` to keep track of max index as a private member of the class?

jpivarski: I wasn't able to select the code that I wanted on my phone, so I commented on the first line. I meant this whole block:

```c++
        for (size_t i = 0; i < index_.length(); i++) {
          if (ptr.get()[i] < 0 ||
              ptr.get()[i] >= content_.length()) {
            std::stringstream out;
            out << "Indexed node" << id_ << " has index " << ptr.get()[i]
                << " at position " << i << " but content has length "
                << content_.length() << "\n";
            error.append(out.str());

            return false;
          }
```

At the time when `is_valid` is called, this runs through the whole `index_` to see if any of its values are `>= content_.length()`. That's a $\mathcal{O}(n)$ test (albeit a compiled one, so not terrible). Instead of that extra run through memory addresses that have probably been evicted from CPU cache and would have to be loaded again, this class could have kept a

```c++
  int64_t max_index_;
```

member that is always equal to the largest `index_` that has been seen so far. So, everywhere that `index_` is updated (only a few member functions do that), check to see if the new `index_` is greater than `max_index_`, and if it is, update `max_index_`. Then when you come to the `is_valid` function, you only need to verify that

```c++
  max_index_ < content_.length()
```

which is a $\mathcal{O}(1)$ check. We've traded a little memory (8 bytes) for some CPU time (running through all $n$ values in `index_`).

Both IndexedArray and IndexedOptionArray can benefit from this, both taking on a single new 8 byte member, and UnionArray can as well, though it needs to take on an 8 byte member for every one of its `contents_` vectors.

If `is_valid` is run frequently, like once per `append`, then this should definitely be fixed. (Once per `append`, which is $\mathcal{O}(n)$, would make the whole process $\mathcal{O}(n^2)$.) But I think `is_valid` is run only once at the end of array-building, in which case it doesn't matter a lot. It's one unnecessary pass over the data, but Awkward Arrays in Python do a lot of passes over the data.

zonca: `is_valid` is called just once after loading the data:

https://gist.github.com/zonca/8045dc6294e2b9ff4e053a339a3a77fb#file-simple_enum-cpp-L115-L116

however, is `concatenate` making an entire copy of the data? If so, it is probably better to implement `max_index_` to avoid memory overhead.

jpivarski: Yes, `concatenate` does copy the entire data. The data are filled in fixed-size panels, which allows them to grow indefinitely, but Awkward Array needs buffers to be (a) contiguous and (b) owned by Python, so that's why we have a concatenate-while-copying step.

Earlier, we had the data grow as `std::vectors`, which reallocates logarithmically many times (because it allocates a bigger buffer each time), but since we had to copy the data to make it owned by NumPy arrays _anyway_, we gave up on trying to keep them contiguous while filling. That's how we ended up with our current pattern.

header-only/layout-builder/awkward/LayoutBuilder.h
```diff
@@ -1151,6 +1153,17 @@ namespace awkward {
...
+      /// @brief Inserts an explicit value in the `index` buffer and
+      /// returns the reference to the builder content.
+      BUILDER&
+      append_index(size_t i) noexcept {
+        index_.append(i);
+        if (i > max_index_) {
+          max_index_ = i;
+        }
+        return content_;
+      }
```
zonca: @jpivarski I have implemented tracking max_index_ for Indexed, however I am not checking that the index is >= 0.
The problem is that if I add the check in `append_index` I should then add a `error` argument to store the error message?

header-only/layout-builder/awkward/LayoutBuilder.h
zonca: ok, implemented

header-only/layout-builder/awkward/LayoutBuilder.h
zonca: implemented here as well

header-only/layout-builder/awkward/LayoutBuilder.h
jpivarski: I think the `>= 0` check could even be ignored. The main reason for the `is_valid` checks is to make sure that non-local constraints are satisfied: it's easy to make one node locally correct and another node locally correct and not notice that they won't work together, so these `is_valid` checks prevent that kind of mistake.

Passing a negative value to an IndexedArray's `index` is a local, more obvious kind of mistake. A non-negative check could be applied immediately when the `index_` is appended to in the first place. If that function is a `noexcept`, then it could set the running `max_index_` to `MAXINT`, so that it would be caught here.

(Thanks for pinging me! Go ahead and do that whenever I've been slow to respond.)

zonca: ok @jpivarski, implemented that, used `UINTMAX_MAX` for max value of a `size_t`

zonca: should I check for that value in `is_valid` and explain the reason for the error?

jpivarski: Yes, a custom error message for `UINTMAX_MAX`, as opposed to any other `max_index_ >= content_.length()`, would help a user because of its different cause.

(I don't know why I missed this 2 days ago. I'm going to keep this tab open, along with my email, Slacks, and all that, so that I check it in my regular rotation.)

zonca: @jpivarski ok, implemented this and rebased on main

jpivarski:
Okay, thanks! It looks done to me; I'll merge it.

PR #3093, opened 2024-04-24
fix: PyPy segfault may be related to using out-of-scope array

jpivarski:
There might be a few other places where pointers-to-pointers are not keeping the pointed-to arrays in scope, like this one.


jpivarski:
No, that wasn't the _only_ problem.

jpivarski:
But this needed to be fixed, so I'm going to merge it anyway.

PR #3112, opened 2024-05-13
fix: skip failing test when jax is not installed

ianna:


ianna (commented):
@jpivarski - the test checked only if `numba` is present, but not `jax`.

jpivarski (approved):
This is definitely necessary. Thanks for catching it!

I'll also just merge it because it's a simple enough PR that I know you're not adding to it right now.

PR #3114, opened 2024-05-13
fix: skip tests that require pyarrow when it is not installed

ianna:


ManasviGoyal (approved):
Looks like I missed adding `importorskip` in some tests. Thank you for catching it @ianna. Please go ahead and merge :)

PR #3115, opened 2024-05-13
fix: generate array view

### LINKED ISSUE #3105: typing ak.Array for numba.cuda.jit signature
### Version of Awkward Array

2.6.2

Hey guys, I followed a hint from the discussion in [#696](https://github.com/scikit-hep/awkward/discussions/696#discussion-2571850) to type `ak.Array` for numba signatures. So I tried something like

```python
import awkward as ak
import numba as nb
from numba import types

cpu_arr_type = ak.Array([[[0, 1], [2, 3]], [[4, 5]]], backend='cpu').numba_type

@nb.njit(types.void(cpu_arr_type))
def cpu_kernel(arr):
   do_something_with_arr
```
and this works like a charm.

However, I'm interested in the same case but with a cuda kernel. So I tried what appeared more natural to do:
```python
gpu_arr_type = ak.Array([[[0, 1], [2, 3]], [[4, 5]]], backend='cuda').numba_type

@nb.cuda.jit(types.void(gpu_arr_type), extensions=[ak.numba.cuda])
def cuda_kernel(arr):
   do_something_with_arr
```
This time, I get the error:
```python
self = <awkward._connect.numba.arrayview_cuda.ArrayViewArgHandler object at 0x784afbc13fa0>
ty = ak.ArrayView(ak.ListArrayType(array(int64, 1d, C), ak.ListArrayType(array(int64, 1d, C), ak.NumpyArrayType(array(int64, 1d, C), {}), {}), {}), None, ())
val = <Array [[[4, 1], [2, -1]], [...], [[4, 0]]] type='3 * var * var * int64'>
stream = 0, retr = []

    def prepare_args(self, ty, val, stream, retr):
        if isinstance(val, ak.Array):
            if isinstance(val.layout.backend, CupyBackend):
                # Use uint64 for pos, start, stop, the array pointers values, and the pylookup value
                tys = numba.types.UniTuple(numba.types.uint64, 5)
    
>               start = val._numbaview.start
E               AttributeError: 'NoneType' object has no attribute 'start'

.../site-packages/awkward/_connect/numba/arrayview_cuda.py:21: AttributeError
```
How should this latter case be correctly treated? Note that, without typing, the thing works as expected:
```python
@nb.cuda.jit(extensions=[ak.numba.cuda])
def cuda_kernel_no_typing(arr):
   do_something_with_arr
```
However, I'm interested in `ak.Array`s with the 3D layout of integers (as above) and would like to take advantage of numba's eager compilation. I'm passing the `arr` for testing as
```python
backend = 'cpu'  # or 'cuda'
arr = ak.to_backend(
    ak.Array([
        [[4, 1], [2, -1]],
        [[0, -1], [1, 1], [3, -1]],
        [[4, 0]]
    ]),
    backend
)
```
Any help is appreciated!

ianna:
fixes #3105

### DIFF
--- a/src/awkward/_connect/numba/arrayview_cuda.py
+++ b/src/awkward/_connect/numba/arrayview_cuda.py
@@ -17,11 +17,13 @@ class ArrayViewArgHandler:
             if isinstance(val.layout.backend, CupyBackend):
                 # Use uint64 for pos, start, stop, the array pointers values, and the pylookup value
                 tys = numba.types.UniTuple(numba.types.uint64, 5)
-
-                start = val._numbaview.start
-                stop = val._numbaview.stop
-                pos = val._numbaview.pos
-                arrayptrs = val._numbaview.lookup.arrayptrs.data.ptr
+                
+                view = val._numbaview
+                
+                start = view.start
+                stop = view.stop
+                pos = view.pos
+                arrayptrs = view.lookup.arrayptrs.data.ptr
                 pylookup = 0
 
                 return tys, (pos, start, stop, arrayptrs, pylookup)
--- a/src/awkward/_connect/numba/arrayview_cuda.py
+++ b/src/awkward/_connect/numba/arrayview_cuda.py
@@ -20,7 +20,7 @@ class ArrayViewArgHandler:
                         f"the array type: {val.numba_type} does not match "
                         f"the kernel signature type: {ty}"
                     )
-                
+
                 # Use uint64 for pos, start, stop, the array pointers values, and the pylookup value
                 tys = numba.types.UniTuple(numba.types.uint64, 5)
 
--- a/src/awkward/_connect/numba/arrayview_cuda.py
+++ b/src/awkward/_connect/numba/arrayview_cuda.py
@@ -17,9 +17,9 @@ class ArrayViewArgHandler:
             if isinstance(val.layout.backend, CupyBackend):
                 # Use uint64 for pos, start, stop, the array pointers values, and the pylookup value
                 tys = numba.types.UniTuple(numba.types.uint64, 5)
-                
+
                 view = val._numbaview
-                
+
                 start = view.start
                 stop = view.stop
                 pos = view.pos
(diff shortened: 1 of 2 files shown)


src/awkward/_connect/numba/arrayview_cuda.py
```diff
@@ -18,10 +18,16 @@ def prepare_args(self, ty, val, stream, retr):
                 # Use uint64 for pos, start, stop, the array pointers values, and the pylookup value
                 tys = numba.types.UniTuple(numba.types.uint64, 5)
 
-                start = val._numbaview.start
-                stop = val._numbaview.stop
-                pos = val._numbaview.pos
-                arrayptrs = val._numbaview.lookup.arrayptrs.data.ptr
+                if val._numbaview is None:
```
ianna: @jpivarski - when a type is given as a parameter to a numba-cuda kernel an array view is not generated. It is generated first time when an array `numba_type` attribute is accessed.

ianna (commented):
@jpivarski - I think, this is fixes the issue. However, now I'm thinking what if the numba_type of the array does not match the numba type of the kernel?

src/awkward/_connect/numba/arrayview_cuda.py
ianna: ok, I think we need to check the type here!

jpivarski: The assumption was that Numba would need to know the `numba_type` in order to start compiling, so accessing the `numba_type` property makes the necessary ArrayView.

If `prepare_args` is a code path that doesn't always invoke `numba_type`, then manually invoking it like this sounds like a good solution.

jpivarski (approved):
Looks good to me! You can merge when ready.

ianna (commented):
@jpivarski - it's strange that the CPU tests on Mac are segfaulting. I'll check it locally.

ianna:
@jpivarski - all the tests pass. Shall I merge it? Thanks.

jpivarski:
Yes! In fact, I can do it, since you've just signed-off on it.

PR #3119, opened 2024-05-14
fix: prevent exponential memory growth in UnionArray

### LINKED ISSUE #3118: Exponential memory growth in UnionArray broadcasting
### Version of Awkward Array

2.6.4

Hey!

To jump right in: I have recently switched from using awkward v1.10.3, to the newest version v2.6.4. I have, however, been running into some very weird issues when adding fields to union arrays. 

So a bit of backstory: 
I am working with electrons and muons, I then try and find dilepton pairs (e-e, mu-mu) and finally I am adding extra information about these dileptons (e.g. angle of dileptons to the missing energy of the event). For some reason, however, after switching to the newer version of awkward I am suddenly running into issues. Specifically, I am running into issues when I add the new fields to the dilepton array. For example, if I try and add new fields like this:
`
dileptons[("Dilepton", "mass")] = <Some array>
`
it will eventually crash from using too much memory, but this issue does not happen when I use awkward v1.10.3. Furthermore, even adding a single field without crashing seems to take longer in v2.6.4 than v1.10.3.

Here is a short snippet that hopefully highlights the issue that I am having. However, instead of having much larger arrays, to show what I mean I am using much smaller ones, and so I have to repeatedly apply the buggy line before the memory gets so big it crashes. In my actual code I am not just repeatedly trying to add the same field!

```
import awkward as ak
import vector
vector.register_awkward()

# Array of electrons
electrons = ak.Array(
        [[],
        [{'pt': 80.5, 'eta': -1.11, 'phi': 1.46, 'mass': 0.0236, "WP90" : True}, {'pt': 25.9, 'eta': -0.954, 'phi': 1.71, 'mass': 0.00993, "WP90" : True}],
        [{'pt': 40.8, 'eta': 0.449, 'phi': 1.88, 'mass': 0.00444, "WP90" : True}, {'pt': 14.7, 'eta': 0.787, 'phi': 3.04, 'mass': 0.00324, "WP90" : True}],
        [{'pt': 55.3, 'eta': 0.449, 'phi': -1.16, 'mass': -0.0147, "WP90" : True}, {'pt': 40.2, 'eta': 0.664, 'phi': -1.95, 'mass': 0.0129, "WP90" : True}]]
)
muons = ak.Array(
        [[{'pt': 41.2, 'eta': -2.24, 'phi': -0.45, 'mass': 0.106}, {'pt': 39.3, 'eta': -1.65, 'phi': -1.01, 'mass': 0.106, "tightID" : True}],
        [],
        [],
        []]
)

# Convert to Momentum4D vectors so that we can do vector operations e.g. add as 4-vectors
muons_4V = ak.Array(muons, with_name = "Momentum4D")
elecs_4V = ak.Array(electrons, with_name = "Momentum4D")

# Find e-e and mu-mu pairs
mm_pairs = ak.combinations(muons_4V, 2, fields = ["LeadLepton", "SubleadLepton"])
ee_pairs = ak.combinations(elecs_4V, 2, fields = ["LeadLepton", "SubleadLepton"])

# Concatenate these together
dileptons = ak.concatenate([ee_pairs, mm_pairs], axis = 1)

# Add the 4-vectors to get the 4-vector of the dilepton
dileptons["Dilepton"] = dileptons.LeadLepton + dileptons.SubleadLepton 


# 

mass = dileptons.Dilepton.mass

# This eventually crashes! It also goes very slow before it crashes!
for i in range(50):
    dileptons[("Dilepton", "mass")] = mass
```
I should note that this doesn't crash if I just ran once:
```
dileptons[("Dilepton", "mass")] = mass
```
It seems to be the creating of the field that retains something which causes memory to build up after repeated calls and then it crashes.

If I run the above code with the older awkward version then it doesn't crash. 
If I do this in a jupyter notebook as well, I can't even run 
```
%%timeit
dileptons[("Dilepton", "mass")] = mass
```
because this ends up crashing before it finishes.

Finally, I should add why I specifically mentioned union arrays.
As you can see in the electron and muon fields, the electron array contains the field "WP90" which is not in muons, and the muon array contains the field "tightID" which is not in the electron array. 
Therefore when you concatenate the ee_pairs and mm_pairs you get a union array:
```
[[{LeadLepton: {pt: 41.2, eta: -2.24, ...}, SubleadLepton: {...}, ...}],
 [{LeadLepton: {pt: 80.5, eta: -1.11, ...}, SubleadLepton: {...}, ...}],
 [{LeadLepton: {pt: 40.8, eta: 0.449, ...}, SubleadLepton: {...}, ...}],
 [{LeadLepton: {pt: 55.3, eta: 0.449, ...}, SubleadLepton: {...}, ...}]]
------------------------------------------------------------------------
type: 4 * var * union[
    {
        LeadLepton: Momentum4D[
            pt: float64,
            eta: float64,
            phi: float64,
            mass: float64,
            WP90: bool
        ],
        SubleadLepton: Momentum4D[
            pt: float64,
            eta: float64,
            phi: float64,
            mass: float64,
            WP90: bool
        ],
        Dilepton: Momentum4D[
            rho: float64,
            phi: float64,
            eta: float64,
            tau: float64
        ]
    },
    {
        LeadLepton: Momentum4D[
            pt: float64,
            eta: float64,
            phi: float64,
            mass: float64,
            tightID: ?bool
        ],
        SubleadLepton: Momentum4D[
            pt: float64,
            eta: float64,
            phi: float64,
            mass: float64,
            tightID: ?bool
        ],
        Dilepton: Momentum4D[
            rho: float64,
            phi: float64,
            eta: float64,
            tau: float64
        ]
    }
]
```
If I now take out both the fields "WP90", "tightID" from the electron and muons arrays, respectively, I no longer get a union array (I guess because all the fields match):
```
[[{LeadLepton: {pt: 41.2, eta: -2.24, ...}, SubleadLepton: {...}, ...}],
 [{LeadLepton: {pt: 80.5, eta: -1.11, ...}, SubleadLepton: {...}, ...}],
 [{LeadLepton: {pt: 40.8, eta: 0.449, ...}, SubleadLepton: {...}, ...}],
 [{LeadLepton: {pt: 55.3, eta: 0.449, ...}, SubleadLepton: {...}, ...}]]
------------------------------------------------------------------------
type: 4 * var * {
    LeadLepton: Momentum4D[
        pt: float64,
        eta: float64,
        phi: float64,
        mass: float64
    ],
    SubleadLepton: Momentum4D[
        pt: float64,
        eta: float64,
        phi: float64,
        mass: float64
    ],
    Dilepton: Momentum4D[
        rho: float64,
        phi: float64,
        eta: float64,
        tau: float64
    ]
}
```
Now if I run the code above, but with the "WP90", "tightID" removed, it no longer crashes! So I am guessing that it is something to do with the union bit? Also not sure exactly why it becomes a union because as far as I can tell the fields "WP90", "tightID" aren't even accessible through the dilepton array?

Not exactly sure what is going on here, but would be interesting to hear your thoughts!

Thanks!

jpivarski:
I decided to just make `UnionArray.simplified` not allow lazy carries. It should always return correct results and it can only have a performance impact on wide RecordArrays that go through `UnionArray.simplified` (perhaps via `ak.concatenate`).

The new test would fail for the old code: for 5 loops, the NumpyArray length gets up to 32, and we assert `<= 2`.

I only had to change one test, which was hard-coded to a particular Form. It was for testing `allow_noncanonical_form=True` in `ak.from_buffers`. I think that feature is still being tested with the updated test.

@agoose77, am I missing anything here? Would you do anything differently?

### DIFF
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -431,7 +431,7 @@ class UnionArray(UnionMeta[Content], Content):
             ]
 
         if len(contents) == 1:
-            next = contents[0]._carry(index, True)
+            next = contents[0]._carry(index, False)
             return next.copy(parameters=parameters_union(next._parameters, parameters))
 
         else:
@@ -702,7 +702,7 @@ class UnionArray(UnionMeta[Content], Content):
         nextcarry = ak.index.Index64(
             tmpcarry.data[: lenout[0]], nplike=self._backend.index_nplike
         )
-        return self._contents[index]._carry(nextcarry, True)
+        return self._contents[index]._carry(nextcarry, False)
 
     @staticmethod
     def regular_index(
--- a/tests/test_2713_from_buffers_allow_noncanonical.py
+++ b/tests/test_2713_from_buffers_allow_noncanonical.py
@@ -120,9 +120,10 @@ def test_union_simplification():
     projected = ak.from_buffers(
         projected_form, length, container, allow_noncanonical_form=True
     )
+
     assert projected.layout.form.to_dict(verbose=False) == {
-        "class": "IndexedArray",
-        "index": "i64",
-        "content": {"class": "RecordArray", "fields": ["x"], "contents": ["int64"]},
+        "class": "RecordArray",
+        "fields": ["x"],
+        "contents": ["int64"],
     }
     assert ak.almost_equal(array[["x"]], projected)
--- a/tests/test_3118_prevent_exponential_memory_growth_in_unionarray.py
+++ b/tests/test_3118_prevent_exponential_memory_growth_in_unionarray.py
@@ -1,0 +1,34 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
+
+from __future__ import annotations
+
+import awkward as ak
+
+
+def test():
+    one_a = ak.Array([{"x": 1, "y": 2}], with_name="T")
+    one_b = ak.Array([{"x": 1, "y": 2}], with_name="T")
+    two_a = ak.Array([{"x": 1, "z": 3}], with_name="T")
+    two_b = ak.Array([{"x": 1, "z": 3}], with_name="T")
+    three = ak.Array([{"x": 4}, {"x": 4}], with_name="T")
+
+    first = ak.zip({"a": one_a, "b": one_b})
+    second = ak.zip({"a": two_a, "b": two_b})
+
+    cat = ak.concatenate([first, second], axis=0)
+
+    cat["another"] = three
+
+    def check(layout):
+        if hasattr(layout, "contents"):
+            for x in layout.contents:
+                check(x)
+        elif hasattr(layout, "content"):
+            check(layout.content)
+        else:
+            assert layout.length <= 2
+
+    for _ in range(5):
+        check(cat.layout)
+
+        cat["another", "w"] = three.x


ianna (commented):
@jpivarski - it looks like the tests on Mac are segfaulting and not only for this PR. I'm checking locally.

agoose77:
@jpivarski I agree with your reasoning. I haven't given much more thought to alternative solutions, but we can always revisit this in future!

ianna:
> @jpivarski - it looks like the tests on Mac are segfaulting and not only for this PR. I'm checking locally.

I can confirm that I can reproduce the segfault locally on MacOS 11.6 with `pyarrow 16.1.0` (released yesterday). It happens in `tests/test_0080_flatpandas_multiindex_rows_and_columns.py`

Changing back to `pyarrow 16.0.0` fixes the issue.

jpivarski:
Although we could hold back `pyarrow<16.1.0`, that would have to be temporary; we'd have to remove the constraint after pyarrow gets fixed. We can only expect pyarrow to get fixed if the error is reported to them, so I tried to do that, but #3122.

ianna:
> Although we could hold back `pyarrow<16.1.0`, that would have to be temporary; we'd have to remove the constraint after pyarrow gets fixed. We can only expect pyarrow to get fixed if the error is reported to them, so I tried to do that, but #3122.

I uninstalled both `awkward` and `awkward-cpp` and have a minimum reproducer of a segfault with:
 ```python
% python
>>> import pyarrow
zsh: segmentation fault  python

```

The `test_pyarrow.py` has only one line:
`import pyarrow`

```zsh
 % lldb python test_pyarrow.py 
(lldb) target create "python"
Current executable set to '/Users/yana/anaconda3/bin/python' (x86_64).
(lldb) settings set -- target.run-args  "test_pyarrow.py"
(lldb) run
Process 34536 launched: '/Users/yana/anaconda3/bin/python' (x86_64)
Process 34536 stopped
* thread #1, queue = 'com.apple.main-thread', stop reason = EXC_BAD_ACCESS (code=EXC_I386_GPFLT)
    frame #0: 0x0000000129c8685e libarrow.1601.dylib`malloc_conf_init_helper + 302
libarrow.1601.dylib`malloc_conf_init_helper:
->  0x129c8685e <+302>: movzbl (%rbx), %eax
    0x129c86861 <+305>: testb  %al, %al
    0x129c86863 <+307>: je     0x129c8936e               ; <+11326>
    0x129c86869 <+313>: movq   %rbx, %rcx
(lldb) 

```

jpivarski:
That reproducer must depend on the environment because it didn't segfault immediately on import for me. I also noticed in the output that this is x86-64. My laptop is ARM (M2). That's likely relevant. 

Could you send as much information as you can to the Arrow developers? Since we also need a short-term solution, we can put a version cap on Arrow in our tests for now. There are a few pull requests that need it.

ianna:
> That reproducer must depend on the environment because it didn't segfault immediately on import for me. I also noticed in the output that this is x86-64. My laptop is ARM (M2). That's likely relevant.
> 
> Could you send as much information as you can to the Arrow developers? Since we also need a short-term solution, we can put a version cap on Arrow in our tests for now. There are a few pull requests that need it.

I've opened an issue https://github.com/apache/arrow/issues/41696

jpivarski:
Awesome! It looks like they're dealing with it right away.

lgray:
This PR introduces massive data inefficiencies for HEP analyses that involve combinations of multiple flavors of leptons, or generally gluing multiple tables together. We should reconsider. I would propose we revert this PR for now, there doesn't seem to be an issue this fixes, but rather one that might happen. I appreciate that, but we should be more careful and reconcile this with the needs of the primary user base.

src/awkward/contents/unionarray.py
```diff
@@ -702,7 +702,7 @@ def project(self, index):
         nextcarry = ak.index.Index64(
             tmpcarry.data[: lenout[0]], nplike=self._backend.index_nplike
         )
-        return self._contents[index]._carry(nextcarry, True)
+        return self._contents[index]._carry(nextcarry, False)
```
jpivarski, line 705: Replaced lazy projection with eager projection here.

PR #3120, opened 2024-05-14
fix: Header-only fix is_valid for empty Indexed or IndexedOption

zonca:
Noticed that the implementation of `is_valid` in #3901
did not handle correctly empty objects.
In that case `max_index_` has the initial value of 0
the length of content is also 0.
Therefore `is_valid` fails because they are equal.

Added 2 test cases for this to the same file given they are related.

- test: test empty IndexedOption builders are valid
- test: suppress printout to avoid thinking tests fail
- fix: append_index() should call append_index(i)
- fix: specific case of is_valid to catch empty objects
- test: separate test case for empty Indexed and IndexedOption

### DIFF
--- a/header-only/layout-builder/awkward/LayoutBuilder.h
+++ b/header-only/layout-builder/awkward/LayoutBuilder.h
@@ -1235,6 +1235,7 @@ namespace awkward {
           out << "Indexed node" << id_ << " has a negative index\n";
           error.append(out.str());
           return false;
+        } else if ((content_.length() == 0) && (max_index_ == 0)) { // catches empty object
         } else if (max_index_ >= content_.length()) {
           std::stringstream out;
           out << "Indexed node" << id_ << " has index " << max_index_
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
This is a good fix—thanks for catching it! 

I'll merge this PR as soon as we get the MacOS bug fixed. I think we'll get that done tomorrow, even if we have to put an upper limit on the pyarrow version we accept.

ManasviGoyal (approved):

ManasviGoyal:
@jpivarski Since the MacOS bug is fixed, this can be merged, right?
@zonca Once there is an awkward release with this fix, we should bump the minimum version of awkward in `kaitai_struct_awkward_runtime` and make a release (or maybe once you are done with the documentation).

jpivarski:
Yes, the MacOS bug was the only thing holding it back. I'll merge it now.

PR #3124, opened 2024-05-17
chore: pin pyarrow to 16.0.0

ianna:
until the issue https://github.com/apache/arrow/issues/41696 is fixed we cannot use 16.1.0


jpivarski (approved):
Thanks! I'll merge this right away so that it can be merged into the other PRs, to unblock the tests.

It looks like the Internet doesn't know how to make a requirements.txt that distinguishes between Intel and Apple Silicon MacOS: https://stackoverflow.com/q/71223782/1623645

PR #3127, opened 2024-05-22
feat: 2772 pyarrow doesnt permit selective reading with extensionarray

### LINKED ISSUE #2772: PyArrow doesn't permit selective reading with ExtensionArray
### Version of Awkward Array

main`

We should stash metadata in the parquet schema, rather than using extensionarray.
We still need `ExtensionArray` for `to_arrow` because `pa.array` doesn't otherwise have a mechanism for stashing metadata.

tcawlfield:
Adding a module, `awkward._connect.pyarrow_table_conv`, to convert pyarrow tables with extensionarray types to ones without, storing metadata necessary to reconstruct the extensionarray types  into the table-level metadata.

This allows reading select columns, as the tables in newly-written Parquet files will have native column types together with this metadata needed to convert back again on read.


src/awkward/_connect/pyarrow_table_conv.py
```diff
@@ -0,0 +1,250 @@
...
+
+from .pyarrow import (
+    AwkwardArrowArray,
+    AwkwardArrowType,
+    import_pyarrow,
+    to_awkwardarrow_storage_types,
+)
+
+pyarrow = import_pyarrow("pyarrow_table_conv")
+AWKWARD_INFO_KEY = b"ak_extn_array_info"  # metadata field in Table schema
```
jpivarski:
```suggestion
AWKWARD_INFO_KEY = b"awkward_array_metadata"  # metadata field in Table schema
```

src/awkward/_connect/pyarrow_table_conv.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def convert_awkward_arrow_table_to_native(aatable: pyarrow.Table) -> pyarrow.Table:
+    """
+    aatable: A pyarrow Table created with extensionarray=True
+    returns: A pyarrow Table without extensionsarrays, but
+      with 'awkward_info' in the schema's metadata that can be used to
```
jpivarski: Do you mean

```suggestion
      with 'awkward_array_metadata' in the schema's metadata that can be used to
```

Is this documentation referring to the value of `AWKWARD_INFO_KEY`?

agoose77:
Super excited to see this work @tcawlfield!

src/awkward/_connect/pyarrow_table_conv.py
```diff
@@ -0,0 +1,250 @@
...
+        # return pyarrow.map_(storage_type.index_type, fields[0])
+        raise NotImplementedError("pyarrow MapType is not supported by Awkward")
+    if isinstance(storage_type, pyarrow.lib.StructType):
+        return pyarrow.struct(fields)
+    if isinstance(storage_type, pyarrow.lib.UnionType):
+        return pyarrow.union(fields, storage_type.mode, storage_type.type_codes)
+    if isinstance(storage_type, pyarrow.lib.DataType) and storage_type.num_fields == 0:
+        # Catch-all for primitive types, nulltype, string types, FixedSizeBinaryType
+        return storage_type
+    raise NotImplementedError(f"Type {typ} is not handled for conversion.")
```
jpivarski: I prefer this kind of structure (a wanna-be `match`-`case`) as

```python
if ...:
    return ...
elif ...:
    return ...
elif ...:
    return ...
elif ...:
    return ...
else:
    raise AssertionError  # because it's a programming error, not expected future work
```

Following up on what we were talking about on Slack, the `elif` prevents scopes from mixing by control flow leaving one of the `if` blocks by not having a `return`. (I know, all of these have `return` statements, but if one gets commented out someday for some reason, these independent `if` statements don't prevent control flow from getting into one of the other ones.)

Some linters complain that `elif` is not necessary because of the `return`, but I turn those messages off because they're encouraging more dangerous code.

jpivarski (commented):
It's looking good! The next step would be to make `ak.to_arrow_table` and `ak.from_arrow_table` (and, by extension, `ak.to_parquet` and `ak.from_parquet`) do this automatically. Then a lot of the existing tests in

* tests/test_1125_to_arrow_from_arrow.py
* tests/test_1154_arrow_tables_should_preserve_parameters.py
* tests/test_1294_to_and_from_parquet.py
* tests/test_1453_write_single_records_to_parquet.py
* tests/test_2340_unknown_type_to_arrow_and_back.py
* tests/test_2772_parquet_extn_array_metadata.py

would verify that the metadata is being fully preserved and arrays are being properly reconstructed.

I tried this:

```python
>>> arrow_table = ak.to_arrow_table(   # tuples need metadata
...     ak.Array([[(1, 1.1), (2, 2.2)], [], [(3, 3.3)]]),
...     extensionarray=True,
... )
>>> ak._connect.pyarrow_table_conv.convert_awkward_arrow_table_to_native(
...     arrow_table
... )
```

because this function takes a `pa.Table` and returns a `pa.Table`, but it didn't work:

```
  File "/Users/jpivarski/irishep/awkward/src/awkward/_connect/pyarrow_table_conv.py", line 37, in convert_awkward_arrow_table_to_native
    return replace_schema(aatable, new_schema)
  File "/Users/jpivarski/irishep/awkward/src/awkward/_connect/pyarrow_table_conv.py", line 182, in replace_schema
    for col, new_field in zip(batch.itercolumns(), new_schema):
AttributeError: 'pyarrow.lib.RecordBatch' object has no attribute 'itercolumns'. Did you mean: 'num_columns'?
```

(I wanted to see what the collected metadata looks like.) Maybe I'm pulling the wrong function, though. (This whole new submodule consists of internal functions, so they don't have to have an obvious way to use them. They can be eclectic.)

This is looking good, but the real test will be when it's applied to all Awkward ↔ Arrow Table conversions, which will verify that the full conversions are lossless and correct.

src/awkward/_connect/pyarrow_table_conv.py
tcawlfield: Oh, yes it is. I shot myself in the proverbial foot.
Can I put ``f""" ... `{AWKWARD_INFO_KEY}` ... """`` in the docstring?
Looks like that should work. Less footgun.

tcawlfield:
> It's looking good! The next step would be to make `ak.to_arrow_table` and `ak.from_arrow_table` (and, by extension, `ak.to_parquet` and `ak.from_parquet`) do this automatically. Then a lot of the existing tests in
> 
> * tests/test_1125_to_arrow_from_arrow.py
> * tests/test_1154_arrow_tables_should_preserve_parameters.py
> * tests/test_1294_to_and_from_parquet.py
> * tests/test_1453_write_single_records_to_parquet.py
> * tests/test_2340_unknown_type_to_arrow_and_back.py
> * tests/test_2772_parquet_extn_array_metadata.py
> 
> would verify that the metadata is being fully preserved and arrays are being properly reconstructed.
> 
> I tried this:
> 
> ```python
> >>> arrow_table = ak.to_arrow_table(   # tuples need metadata
> ...     ak.Array([[(1, 1.1), (2, 2.2)], [], [(3, 3.3)]]),
> ...     extensionarray=True,
> ... )
> >>> ak._connect.pyarrow_table_conv.convert_awkward_arrow_table_to_native(
> ...     arrow_table
> ... )
> ```
> 
> because this function takes a `pa.Table` and returns a `pa.Table`, but it didn't work:
> 
> ```
> Traceback (most recent call last):
>   File "<stdin>", line 1, in <module>
>   File "/Users/jpivarski/irishep/awkward/src/awkward/_connect/pyarrow_table_conv.py", line 37, in convert_awkward_arrow_table_to_native
>     return replace_schema(aatable, new_schema)
>   File "/Users/jpivarski/irishep/awkward/src/awkward/_connect/pyarrow_table_conv.py", line 182, in replace_schema
>     for col, new_field in zip(batch.itercolumns(), new_schema):
> AttributeError: 'pyarrow.lib.RecordBatch' object has no attribute 'itercolumns'. Did you mean: 'num_columns'?
> ```
> 
> (I wanted to see what the collected metadata looks like.) Maybe I'm pulling the wrong function, though. (This whole new submodule consists of internal functions, so they don't have to have an obvious way to use them. They can be eclectic.)
> 
> This is looking good, but the real test will be when it's applied to all Awkward ↔ Arrow Table conversions, which will verify that the full conversions are lossless and correct.

This should have worked. It works for me. Perhaps you found the first issue w/rt differing pyarrow versions? Or maybe I had failed to push local commits yesterday? I pushed some commits just now -- curious what this does for you now.

src/awkward/_connect/pyarrow_table_conv.py
tcawlfield: I agree with you here. I think I wrote it this way out of habit or maybe fear that ruff would change it to this anyway. But it was fine with the if-elif chain. I left the final `raise NotImplementedError` out of the chain because that's more bulletproof. If you ever make it out of that that chain without returning something, fail.

src/awkward/_connect/pyarrow_table_conv.py
tcawlfield: It did not work, and Ruff caught the would-be-error thankfully. I just corrected it manually.

tcawlfield:
I'm marking this as ready for review, even though there's failures on Python 3.8 still. I'll look into what it takes to reproduce this, but so far I'm hitting errors like:
```
AttributeError("'pyarrow.lib.LargeListType' object has no attribute 'field'")
```
that come [specifically from here](https://github.com/scikit-hep/awkward/blob/2772-pyarrow-doesnt-permit-selective-reading-with-extensionarray/src/awkward/_connect/pyarrow/table_conv.py#L142).

I'm having trouble installing all of requirements-test-minimal.txt in a Python 3.11 environment, although I got pyarrow 7.0.0 installed. And indeed LargeListArray in that version has a `value_field` property but not a `field()` method. I can probably special-case that with hasattr, but without reproducing the whole environment I'm not sure what would fail next. So I'll pause here and wait for any advice you might offer, and in the meantime try to get this working with an old Python somewhere.

You'll quickly notice that I finally broke down and converted `._connect.pyarrow` into a package, with my new code as a submodule, `.table_conv`. I did a very minimal restructuring of the rest of pyarrow.py -- top-level `__init__.py` just handles safe importing with/without pyarrow and creates the same namespace of symbols regardless, so that other code could import anything here without having to get too fancy. The extension classes went into a separate submodule because that part was easy to chop off and it felt reasonable to do so. Everything else went into a grab-bag, `.../pyarrow/conversions.py`. (I'm happy to rename this.) I felt this was a safe place to draw the line between minor and major rearranging. My primary motivation was to pull the new table-conversion functions, convert_awkward_arrow_table_to_native and convert_native_arrow_table_to_awkward, up into `..._connect.pyarrow` in a way that kept the new code separate, avoided circular imports, and worked well with/without pyarrow.

jpivarski:
> And indeed LargeListArray in that version has a `value_field` property but not a `field()` method. I can probably special-case that with hasattr, but without reproducing the whole environment I'm not sure what would fail next.

Actually, I remember that issue with LargeListArray. I might have even reported it, but it's probably from back when Arrow used JIRA instead of GitHub and I'd have trouble finding it. Special-casing with `hasattr` is definitely fine, and it's an isolated thing. There's no reason to expect a deluge of issues afterward.

Is the difficulty of making an environment with pyarrow 7 related to conda versus pip venv (because pyarrow has a binary dependency)? Making environments with the minimal and maximal dependencies works with conda (mamba).

Reorganizing all of the pyarrow connections into a directory is a good idea. Thanks!

tcawlfield:
Okay I think this is all ready for review now, with tests passing!

> Is the difficulty of making an environment with pyarrow 7 related to conda versus pip venv (because pyarrow has a binary dependency)? Making environments with the minimal and maximal dependencies works with conda (mamba).

My issue with pyarrow 7.0.0 is resolved. I'm not sure why I had trouble installing the older requirements earlier. It was a compilation error of some sort, but I was able to grab old pyarrow but latest versions of the other dependencies in our requirements-minimal.txt (something like that anyway) and work out all the changes needed. I left some earlier code commented-out with the rationale that we might in the future drop support for older pyarrow versions, and can clean up and/or improve the code. Some people don't like checking in commented-out code. "That's what git is for!" And this is true but I have things commented-out that might be worth reviewing in the future, or might also make intentions more clear. I do try to keep this to a bare minimum though.

src/awkward/_connect/pyarrow/__init__.py
```diff
@@ -0,0 +1,124 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
```
jpivarski, line 1: I like the fact that the code in this `__init__.py` is minimal; it's just about importing various parts of pyarrow.

src/awkward/_connect/pyarrow/conversions.py
```diff
@@ -1,202 +1,22 @@
-# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
-
 from __future__ import annotations
```
jpivarski:
```suggestion
# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE

from __future__ import annotations
```

src/awkward/_connect/pyarrow/extn_types.py
```diff
@@ -0,0 +1,139 @@
+"""
```
jpivarski, line 3:
```suggestion
# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE

"""
```

src/awkward/_connect/pyarrow/table_conv.py
```diff
@@ -0,0 +1,273 @@
+from __future__ import annotations
```
jpivarski, line 1:
```suggestion
# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE

from __future__ import annotations
```

tests/test_2772_parquet_extn_array_metadata.py
```diff
@@ -0,0 +1,206 @@
+# ruff: noqa: E402
```
jpivarski, line 2:
```suggestion
# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE

# ruff: noqa: E402
```

tests/test_2772_parquet_extn_array_metadata.py
```diff
@@ -0,0 +1,206 @@
...
+    # if not isinstance(akarray, ak.Array):
+    #     akarray = ak.Array(akarray)  # A possible issue but this provides .type.show()
+    # ak_type_str_orig = io.StringIO()
+    # ak_type_str_rtrp = io.StringIO()
+    # akarray.type.show(stream=ak_type_str_orig)
+    # rt_array.type.show(stream=ak_type_str_rtrp)
+    # if ak_type_str_orig.getvalue() != ak_type_str_rtrp.getvalue():
+    #     print("  Original type:", ak_type_str_orig.getvalue())
+    #     print("  Rnd-trip type:", ak_type_str_rtrp.getvalue())
+    # assert ak_type_str_orig.getvalue() == ak_type_str_rtrp.getvalue()
```
jpivarski: Along with the argument of "this is what git is for" (though it's sometimes hard to find old code in git, especially if it only exists in a PR branch like this), there's also, "this is what `@pytest.mark.skip` is for." You could include this failing test with the `skip` decorator and put the comment about why it's failing in the `reason`.

Then there's the question of "`@pytest.mark.skip`" versus "`@pytest.mark.xfail`". We do expect this to fail, but I don't like the way that pytest shows you the verbose, red stacktrace for tests that are expected to fail—it becomes hard to find the output of tests that are not expected to fail. Or, at least, if `xfail` is going to be used, it should be for tests that will be fixed in the short term, rather than the long term. So I think this one should be `skip`, since I get the impression that it won't be fixed in the short term.

jpivarski (approved):
This is looking great! I think this is ready to merge, if updating to `main` is uneventful (i.e. it merges cleanly and all tests continue to pass).

src/awkward/_connect/pyarrow/__init__.py
tcawlfield: Thanks, this arrangement should allow for other code to do `from awkward._connect.pyarrow import ...` without too much anxiety, if the author wants.

tests/test_2772_parquet_extn_array_metadata.py
tcawlfield: I opened issue 3131 and attached a stand-alone unit test there. Your discretion how to resolve this one. It's not clear to me that my expectations are even correct.

PR #3113, opened 2024-05-13
fix: wait for Jitify performing a one-time only warm-up

ianna:
fixes https://github.com/scikit-hep/awkward/issues/3111

PR #3130, opened 2024-05-27
fix: use grid-stride looping for kernels with variable-length loops

ManasviGoyal:


ManasviGoyal:
@jpivarski Since `awkward_ListArray_getitem_next_range` kernel is needed for use, I think this can be merged if the everything works fine. I can make a new PR to work on other kernels.

jpivarski (commented):
So far, I haven't been able to test it because of thousands of

```
E   cupy._util.PerformanceWarning: Jitify is performing a one-time only warm-up to populate the persistent cache, this may take a few seconds and will be improved in a future release...
```

warnings. #3113 was supposed to fix this, but it apparently isn't. I'll have to follow up again later.

ManasviGoyal:
@jpivarski This seems to be an issue with the more recent versions of `cupy`. Maybe testing on a older version `12.x.x` can be done so that we don't have to wait for fix.

jpivarski (approved):
I think CuPy 13 added a persistent cache to reduce the time spent compiling after the first compilation. In fact, the message is only presented the first time it encounters a kernel, so that must be what's going on. The messages might be helpful to end-users, but our test framework is set to consider warnings as errors. I added a filter to ignore these warnings when running the tests.

I think this PR is ready to be merged! I made an edit, but not to your code, so I don't think you'll have any counter-edits. I'll merge it as soon as the (CPU) tests pass.

Of course, I ran all of the GPU tests, and they all pass.

PR #3126, opened 2024-05-21
ci: Add GitHub artifact attestations to package distribution

matthewfeickert:
* Add generation of GitHub artifact attestations to built sdist and wheel before upload. c.f.:
   - https://github.blog/2024-05-02-introducing-artifact-attestations-now-in-public-beta/
   - https://docs.github.com/en/actions/security-guides/using-artifact-attestations-to-establish-provenance-for-builds
* Add verification of artifact attestation before publishing awkward to PyPI using the 'gh attestation verify' CLI API, added in v2.49.0.
   - c.f. https://github.com/cli/cli/releases/tag/v2.49.0

As an example c.f. https://github.com/scikit-hep/pyhf/pull/2473.

PR #3135, opened 2024-05-28
ci: Ensure attestations for awkward-cpp sdist and wheels

matthewfeickert:
* awkward-cpp's sdists are named `awkward-cpp-*.tar.gz` while the wheels are named `awkward_cpp-*.whl`, so to ensure that both the sdist and wheels get attestations use a wildcard to capture both `-` and `_`.
* Amends PR https://github.com/scikit-hep/awkward/pull/3126.

### DIFF
--- a/.github/workflows/deploy-cpp.yml
+++ b/.github/workflows/deploy-cpp.yml
@@ -37,6 +37,6 @@ jobs:
     - name: Generate artifact attestation for sdist and wheel
       uses: actions/attest-build-provenance@173725a1209d09b31f9d30a3890cf2757ebbff0d # v1.1.2
       with:
-        subject-path: "dist/awkward-cpp-*"
+        subject-path: "dist/awkward*cpp-*"
 
     - uses: pypa/gh-action-pypi-publish@v1.8.14


.github/workflows/deploy-cpp.yml
```diff
@@ -37,6 +37,6 @@ jobs:
     - name: Generate artifact attestation for sdist and wheel
       uses: actions/attest-build-provenance@173725a1209d09b31f9d30a3890cf2757ebbff0d # v1.1.2
       with:
-        subject-path: "dist/awkward-cpp-*"
+        subject-path: "dist/awkward*cpp-*"
```
matthewfeickert, line 40: c.f. https://github.com/scikit-hep/awkward/pull/3126#issuecomment-2136079341

jpivarski: This is fine. Nothing else will be named in such a way that starts with `awkward` and ends with `cpp`.

<p>Maybe <tt>awkward-robocpp</tt>? 
</p>

jpivarski (approved):

PR #3140, opened 2024-06-04
feat: add slicing CUDA kernels

ManasviGoyal:


jpivarski (approved):
This passes all tests on my GPU.

It's also (almost) purely adding code, rather than removing any.

The new kernels look good to me.

Go ahead and merge it!

PR #3141, opened 2024-06-05
feat: add some misc CUDA kernels

ManasviGoyal:


jpivarski (approved):
This is good! (Add additions, and they look like good implementations. Why was tests-cuda/test_3086_cuda_flatten.py removed, though?)

I tested it on my GPU and it's ready to merge.

ManasviGoyal:
> This is good! (Add additions, and they look like good implementations. Why was tests-cuda/test_3086_cuda_flatten.py removed, though?)
> 
> I tested it on my GPU and it's ready to merge.

I moticed that I added a few tests multiple times. So I just removed the repetitive ones and moved the rest to other tests files. So the `flatten` tests are still there but moved to different test files as per the functions they were testing like `concatenate`, `slicing` etc.

PR #3146, opened 2024-06-10
ci: faster wheel builds

henryiii:
Faster wheel builds.


jpivarski (approved):
If this passes tests with the updated merge point, I'll take it.

Is it likely to conflict (or maybe even help) with #3164?

henryiii:
> Is it likely to conflict (or maybe even help) with https://github.com/scikit-hep/awkward/pull/3164?

It'll reduce the time it takes to find out if changes fail or pass. :)

PR #3156, opened 2024-06-18
feat: 3154 from_parquet should be able to read partial columns

### LINKED ISSUE #3154: ak.from_parquet should be able to read partial columns

Pyarrow's parquet reader, which Awkward uses, can read not only selected columns but also sub-fields of selected columns, at least if that column holds a StructArray.

Example:
```
ak_tbl = ak.Array(
    {
        "a": [
            {"lbl": "item 1", "idx": 11, "ids": [1, 2, 3]},
            {"lbl": "item 2", "idx": 12, "ids": [51, 52]},
            {"lbl": "item 3", "idx": 13, "ids": [61, 62, 63, 64]},
        ],
    }
)
ak.to_parquet(ak_tbl, "some.parquet")
selection = ak.from_parquet("some.parquet", columns=["a.ids"])
```

tcawlfield:
This should fix a missing feature (or fix a bug, depending on your POV) from 2772.

Also taking the opportunity to clean up test_2772 a tiny little bit. This completes a misunderstanding of mine from issue 3131 (which can be closed or rejected).


tests/test_2772_parquet_extn_array_metadata.py
```diff
@@ -137,6 +138,22 @@ def test_array_conversions(akarray, as_dict):
...
+        assert rt_array.type == akarray_high.type
+
+        ak_type_str_orig = io.StringIO()
+        ak_type_str_rtrp = io.StringIO()
+        akarray_high.type.show(stream=ak_type_str_orig)
+        rt_array.type.show(stream=ak_type_str_rtrp)
+        if ak_type_str_orig.getvalue() != ak_type_str_rtrp.getvalue():
+            print("  Original type:", ak_type_str_orig.getvalue())
+            print("  Rnd-trip type:", ak_type_str_rtrp.getvalue())
+        assert ak_type_str_orig.getvalue() == ak_type_str_rtrp.getvalue()
```
jpivarski, line 155: This depends on us not changing the way that types are represented as strings. Well, it depends on not introducing any differences between the original type and the round-trip type, which is unlikely. There must be a better way to compare types with your chosen granularity (e.g. preserving the distinction between `X` and `Categorical[X]`).

tests/test_3154_parquet_subcolumn_select.py
```diff
@@ -0,0 +1,41 @@
...
+    ak.to_parquet(ak_tbl, parquet_file)
+
+    selection = ak.from_parquet(parquet_file, columns=["a.ids", "b"])
+    assert selection["a"].to_list() == [
+        {"ids": [1, 2, 3]},
+        {"ids": [51, 52]},
+        {"ids": [61, 62, 63, 64]},
+    ]
+    assert selection["b"].to_list() == ak_tbl["b"].to_list()
+    # return ak_tbl, selection
```
jpivarski: pytest tests don't return anything

jpivarski (approved):
This looks like a good test of partial read-back. This is going through the new metadata, right?

tests/test_2772_parquet_extn_array_metadata.py
tcawlfield: I think this ends up testing the parameters for the nested types, which doesn't happen automatically. I was stealing the approach from [tests/test_1440_start_v2_to_parquet.py](https://github.com/scikit-hep/awkward/blob/e9dd2694cf5c6ccb1717358fe926816d8ecbfb9c/tests/test_1440_start_v2_to_parquet.py#L32-L39).

tests/test_3154_parquet_subcolumn_select.py
tcawlfield: True, I should have removed that. Yes, I'll remove it and check in again.

This let me do: `python -i tests/test_3154_parquet_subcolumn_select.py` and `ak_tbl, selection = test_...()` so I could explore the results a little more easily. I was trying to select parts of column "b" which never worked. I think sub-column select is limited to burrowing into StructArrays. It's not particularly well documented within pyarrow, which requires a lot of experimentation.

PR #3158, opened 2024-06-19
fix: dtype of np.sum boolean arrays

ianna:


ianna:
Closing it as duplicate. The issue is fixed in https://github.com/scikit-hep/awkward/pull/3159

PR #3161, opened 2024-06-20
fix: ak.from_numpy should fail on zero-dimensional arrays.

### LINKED ISSUE #1057: ak.Array and ak.from_numpy should not accept zero-dimensional arrays
... because they're not iterable. This was raised in an issue by @bfis:

> #### In case that there is supposed to be support for shapeless arrays, this illustrates the inconsistency:
> ```python
> a = np.array(0) # value: array(0)
> b = ak.Array(a) # value: <Array [0] type='1 * int64'>
> c = ak.to_numpy(b) # value: array([0])
> assert a.shape == c.shape # this fails
> ```

Elsewhere, we consider NumPy scalars (e.g. `np.float32(3.14)`) to be equivalent to zero-dimensional arrays (e.g. `np.array(3.14, np.float32)`), but this is interpreting the zero-dimensional array as a length-1 array.

`ak.Array` and `ak.from_numpy` do not accept scalars:

```python
>>> ak.Array(np.float32(3.14))
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward/highlevel.py", line 254, in __init__
    layout = ak.operations.convert.from_iter(
  File "/home/jpivarski/miniconda3/lib/python3.8/site-packages/awkward/operations/convert.py", line 885, in from_iter
    for x in iterable:
TypeError: 'numpy.float32' object is not iterable
```

So they should not accept zero-dimensional arrays, for consistency. In other words, not this:

```python
>>> ak.Array(np.array(3.14, np.float32))
<Array [3.14] type='1 * float32'>
```

_Originally posted by @jpivarski in https://github.com/scikit-hep/awkward-1.0/issues/1055#issuecomment-896017089_, with @bfis's response in https://github.com/scikit-hep/awkward-1.0/issues/1055#issuecomment-896027116

tcawlfield:
At this point, the first tests, `ak.Array()`, was already passing without modifications.
But the second test was not -- `ak.from_numpy()`.
Fix is in `_layout.py`. No policy options are passed to `_layout.from_arraylib()`.

### DIFF
--- a/src/awkward/_layout.py
+++ b/src/awkward/_layout.py
@@ -341,6 +341,11 @@ def from_arraylib(array, regulararray, recordarray):
     if array.dtype == np.dtype("O"):
         raise TypeError("Awkward Array does not support arrays with object dtypes.")
 
+    if array.ndim == 0:
+        raise TypeError(
+            "Encountered a scalar (ndarray), but scalar conversion is not allowed."
+        )
+
     if isinstance(array, numpy.ma.MaskedArray):
         mask = numpy.ma.getmask(array)
         array = numpy.ma.getdata(array)
--- a/src/awkward/_layout.py
+++ b/src/awkward/_layout.py
@@ -250,7 +250,12 @@ def maybe_highlevel_to_lowlevel(obj):
         return obj
 
 
-def from_arraylib(array, regulararray, recordarray):
+def from_arraylib(
+    array,
+    regulararray,
+    recordarray,
+    primitive_policy: Literal["error", "promote", "pass-through"] = "promote",
+):
     from awkward.contents import (
         ByteMaskedArray,
         ListArray,
@@ -341,9 +346,9 @@ def from_arraylib(array, regulararray, recordarray):
     if array.dtype == np.dtype("O"):
         raise TypeError("Awkward Array does not support arrays with object dtypes.")
 
-    if array.ndim == 0:
+    if primitive_policy == "error" and array.ndim == 0:
         raise TypeError(
-            "Encountered a scalar (ndarray), but scalar conversion is not allowed."
+            f"Encountered a scalar ({type(array).__name__}), but scalar conversion/promotion is disabled"
         )
 
     if isinstance(array, numpy.ma.MaskedArray):
--- a/src/awkward/operations/ak_from_cupy.py
+++ b/src/awkward/operations/ak_from_cupy.py
@@ -9,7 +9,15 @@ __all__ = ("from_cupy",)
 
 
 @high_level_function()
-def from_cupy(array, *, regulararray=False, highlevel=True, behavior=None, attrs=None):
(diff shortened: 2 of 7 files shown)


tcawlfield:
Ugh this was too naive. Breaks a bunch of other tests.

tcawlfield:
> Ugh this was too naive. Breaks a bunch of other tests.

Okay, I fixed the problems by adding a `primitive_policy` kwarg to `_layout.from_arraylib` and upstream functions.
I'm not sure now why all the CI tests are failing. They pass on my machine, and the reasons seem to relate to C++ issues outside the scope of my change here.

This change ended up being noisier than I would have liked, but I don't see any other way to get the desired policy.

src/awkward/_layout.py
```diff
@@ -250,7 +250,12 @@ def maybe_highlevel_to_lowlevel(obj):
         return obj
 
 
-def from_arraylib(array, regulararray, recordarray):
+def from_arraylib(
+    array,
+    regulararray,
+    recordarray,
+    primitive_policy: Literal["error", "promote", "pass-through"] = "promote",
```
jpivarski, line 257: Why is the default `primitive_policy` to raise an error for some backends and to promote for other backends?

jpivarski (commented):
If `from_arraylib` were a public function, I'd want to think about the consequences of adding a `primitive_policy` argument. As an internal function, we can always reconsider it's interface at a later date, and this interface looks good enough for now. But the defaults seem to be inconsistent, and that would make it hard to think about.

Understanding the test failures will be important, since they're telling us that behavior is changing—probably not in ways that we want.

src/awkward/_layout.py
tcawlfield: Looking through the code, there are quite a lot of functions taking a primitive_policy optional argument, and the defaults vary between "promote", "error", and "pass-through." In _layout we have `HighLevelContext.unwrap` with `primitive_policy="promote"` in addition to the one I added this argument to, `from_arraylib`.

There are quite a lot of ways into from_arraylib. One common path is the high-level function [from_numpy](https://github.com/scikit-hep/awkward/blob/03f6169841ccd92714dd904e76e8531887dbe3c3/src/awkward/operations/ak_from_numpy.py#L55). But these already had a primitive_policy option which was being specified higher up the call stack, for example in [ak_fill_none](https://github.com/scikit-hep/awkward/blob/03f6169841ccd92714dd904e76e8531887dbe3c3/src/awkward/operations/ak_fill_none.py#L77-L84).

This argument allows us to add `primitive_policy="errer"` in, specifically, the [from_numpy](src/awkward/operations/ak_from_numpy.py) function. (Edit: Yes this is the same function called indirectly above, but when invoked by the user, our new default policy becomes to *not* promote zero-dim arrays to be length-1 arrays holding that scalar value.

PR #3164, opened 2024-06-24
chore: update matrix to macOS-latest

ianna:
The macOS-11 environment is deprecated and will be removed on June 28th, 2024. Currently, all tests on MacOS-11 are cancelled:
```
Run Tests (macos-11, 3.8, x64, full)
This is a scheduled macOS-11 brownout. The macOS-11 environment is deprecated and will be removed on June 28th, 2024.
Run Tests (macos-11, 3.11, x64, full)
GitHub Actions has encountered an internal error when running your job.
Run Tests (macos-11, 3.9, x64, full)
GitHub Actions has encountered an internal error when running your job.
Run Tests (macos-11, 3.10, x64, full)
GitHub Actions has encountered an internal error when running your job.
Run Tests (macos-11, 3.12, x64, full)
This is a scheduled macOS-11 brownout. The macOS-11 environment is deprecated and will be removed on June 28th, 2024.
Run Tests (macos-11, 3.12, x64, full)
GitHub Actions has encountered an internal error when running your job.
```

PR #3167, opened 2024-06-25
fix: use classic linker with AppleClang 15+

henryiii:
Checking #3164 with macos-13.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -256,3 +256,13 @@ jobs:
 
       - name: Test
         run: python -m pytest -vv -rs tests -k cppyy
+
+
+  pass-tests:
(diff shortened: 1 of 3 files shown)


ianna (approved):
@henryiii - thanks so much! I see that all the tests pass!

jpivarski:
Does this fix everything? If so, thank you @henryiii!!!

Also, it would mean that a lot of test-PRs can be cleaned up, and done-but-waiting PRs can be updated and merged.

henryiii:
Kind of. It fixes things as long as Apple doesn’t remove the legacy option. :)

We should probably try macOS 14. And eventually figure out why we can’t link with the new linker.

PR #3168, opened 2024-06-25
fix: unexpected behaviour of akwhere with arrays containing nones

### LINKED ISSUE #3098: unexpected behaviour of ak.where with arrays containing Nones
### Version of Awkward Array

1.10.5

I ran into this weird behaviour of ak.where when used on arrays of mixed datatype (not sure about the terminology, but I mean arrays potentially containing None values, as indicated by the question mark).

Basically, the function breaks down when the boolean condition is of mixed datatype and the choices contain a None value.
In that case, the output will be None, even when the boolean condition doesn't contain any Nones at all and the None value shouldn't have been selected. For the choice arrays, the problem only arises when one of them contains a None indeed. For the boolean arrays, it is sufficient that the datatype is mixed.
See the code below to reproduce this behaviour. 

```
mixed_type_cond = ak.Array([[True], [None]])[0]
pure_type_cond = ak.Array([True])
none_alternative = ak.Array([None])
zero_alternative = ak.Array([0])
mixed_zero_alternative = ak.Array([[0], [None]])[0]

ak.where(mixed_type_cond, 1, none_alternative)
>>> <Array [None] type='1 * ?int64'>
ak.where(mixed_type_cond, 1, zero_alternative)
>>> <Array [1] type='1 * ?int64'>
ak.where(mixed_type_cond, 1, mixed_zero_alternative)
>>> <Array [1] type='1 * ?int64'>

ak.where(pure_type_cond, 1, none_alternative)
>>> <Array [1] type='1 * ?int64'>
ak.where(pure_type_cond, 1, zero_alternative)
>>> <Array [1] type='1 * ?int64'>
ak.where(pure_type_cond, 1, mixed_zero_alternative)
>>> <Array [1] type='1 * ?int64'>

ak.where(~mixed_type_cond, none_alternative, 1)
>>> <Array [None] type='1 * ?int64'>
ak.where(~mixed_type_cond, zero_alternative, 1)
>>> <Array [1] type='1 * ?int64'>
ak.where(~mixed_type_cond, mixed_zero_alternative, 1)
>>> <Array [1] type='1 * ?int64'>
```

tcawlfield:
There is a problem with ak.which (the three-argument version) where if both (1) a conditional array element is optional yet True/False, and (2) the array element NOT selected by the condition is optional and None, the resulting element will be None regardless of the value that ought to have been selected.


src/awkward/operations/ak_where.py
```diff
@@ -90,6 +90,8 @@ def _impl3(condition, x, y, mergebool, highlevel, behavior, attrs):
     def action(inputs, backend, **kwargs):
         x, y, condition = inputs
         if isinstance(condition, ak.contents.NumpyArray):
+            # print(f"ak_where action: we're a go! {condition =}")
+            # print(f"{x =}\n{y =}")
```
jpivarski:
```suggestion
```

src/awkward/operations/ak_where.py
```diff
@@ -103,13 +105,15 @@ def action(inputs, backend, **kwargs):
                         backend.nplike.shape_item_as_index(tags.length),
                     )
                 )
+                # print(f"x coerced into {x =}")
```
jpivarski:
```suggestion
```

src/awkward/operations/ak_where.py
```diff
@@ -103,13 +105,15 @@ def action(inputs, backend, **kwargs):
...
             if not isinstance(y, ak.contents.Content):
                 y = ak.contents.NumpyArray(
                     backend.nplike.repeat(
                         backend.nplike.asarray(y),
                         backend.nplike.shape_item_as_index(tags.length),
                     )
                 )
+                # print(f"y coerced into {y =}")
```
jpivarski:
```suggestion
```

src/awkward/operations/ak_where.py
```diff
@@ -119,8 +123,11 @@ def action(inputs, backend, **kwargs):
                 ),
             )
         else:
+            # print(f"ak_where action: we're no-go. {condition =}")
```
jpivarski:
```suggestion
```

jpivarski (approved):
This is good! We talked about the refactoring of how `broadcast_any_option_akwhere` is dispatched, and I see that it has been updated.

The print statements, even though they're commented out, should be removed; just accept all the suggestions below.

Also, update this to `main` and re-run the tests. Then if everything passes, it can be merged.

Actually, regarding that: in our other PRs, the MacOS tests have been failing because GitHub removed MacOS 11 and the newer versions of MacOS have been compiling/linking code that can't be loaded. (Talk to @ianna if you have any ideas.) Is your PR really passing all of those tests? If it fails to pass after you've updated, maybe that's a hint as to what's going wrong: it might be subtly in one of the new GPU kernels, or something that was touched to add those GPU kernels (MacOS does not compile the GPU code itself, but it might have been affected in some unexpected way).

If, after updating to `main`, you see errors, then this PR will be approved for merging, but can't be merged until the MacOS errors are fixed. @ianna and I have talked, and we're considering temporarily turning off the tests that fail on MacOS so that PRs aren't blocked, but we can't make a release with that because it will break Uproot. (Uproot depends on the code that's failing to load.)

PR #3176, opened 2024-07-04
fix: out of bounds index access issue in argmin and argmax complex in CPU kernel

ManasviGoyal:


ManasviGoyal:
@jpivarski @ianna I added the unit tests tests for `awkward_reduce_argmin_complex` and `awkward_reduce_argmin_complex` and noticed that there was an out of bounds index access in the respective CPU kernels which gives a segmentation fault for the unit tests. Since these kernels were not tested earlier in the generated tests it might have been overlooked. There are integration tests already in `tests` directory that use these kernels which makes me suspect that they might be the reason for the pypy segfault (or was that fixed)?

jpivarski (approved):
Good catch! This looks like it's more than fixing an out-of-bounds issue; the result wasn't right with the old indexing, was it? Are there any examples of tests that would fail numerically but not with an out of bounds index? (It's not necessary to produce one; I'm just asking.)

ManasviGoyal:
> Good catch! This looks like it's more than fixing an out-of-bounds issue; the result wasn't right with the old indexing, was it? Are there any examples of tests that would fail numerically but not with an out of bounds index? (It's not necessary to produce one; I'm just asking.)

For the already implemented integration tests and the unit tests I added, the results were actually correct numerically but for the unit tests it shows a segfault on running the tests in `tests-cpu-kernels-explicit` multiple times.

ManasviGoyal:
@ianna This one only changes the CPU kernels and since the CI tests pass, I think it can be merged. Thanks!

ianna (approved):
@ManasviGoyal - Good catch! Thanks! I'll merge it when the tests pass. Thanks!

PR #3177, opened 2024-07-04
fix: refine the implementation of copy_behaviors

Saransh-cpp:


Saransh-cpp:
For discussion:

> How should I interpret the fact this is in the `_util` module? Is it private? Is it ok to use in coffea?

_Originally posted by @nsmith- in https://github.com/scikit-hep/awkward/issues/3137#issuecomment-2217807448_

jpivarski (approved):
Great, thanks! I remember that this was the solution we discussed. It changes the way `copy_behaviors` works, but it's a new enough function (and you're the only user), so that backward incompatibility is not going to be an issue.

This PR is ready to be merged, as-is. When you're done, go ahead and merge it!

Saransh-cpp:
Thanks for the review!

PR #3182, opened 2024-07-14
fix: skip avro reading tests on mac arm64

ianna:
This is not a bugfix, just to check that there are no other issues with moving to mac-latest.


ianna:
no need to skip the tests now

PR #3189, opened 2024-07-29
fix: requesting layout.form should not touch shape

jpivarski:
@martindurant

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -321,6 +321,12 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
         return self._shape
 
     @property
+    def inner_shape(self) -> tuple[ShapeItem, ...]:
+        if len(self._shape) > 1:
+            self.touch_shape()
+        return self._shape[1:]
+
+    @property
     def form_key(self) -> str | None:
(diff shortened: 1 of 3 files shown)


martindurant:
Is awkward main still compatible with the last awkward-cpp, so that I can install this simply without recompiling?

martindurant:
I was also going to comment "wow, so many lines", but I see it's all tests for the different classes.

jpivarski:
> Is awkward main still compatible with the last awkward-cpp, so that I can install this simply without recompiling?

I think so. I'm testing against `main` here: https://github.com/scikit-hep/awkward/actions/runs/10163112284

And testing against `pivarski/requesting_form_should_not_touch_shape` here: https://github.com/scikit-hep/awkward/actions/runs/10163115961

Oh, apparently not. Let me see if I can get a new awkward-cpp into PyPI today.

jpivarski:
A new version of awkward-cpp is available. You can just pip-install it.

martindurant:
This seems to work as intended.
I still don't know how to solve the concat case, though - we can talk about that separately.

agoose77:
I like this! It makes sense that we don't need any shape-_content_ information to know that the inner_shape is `()` -- the dimension is a form-level thing.

Issue #2947, opened 2024-01-15
Array.to_list fails when length is a numpy integer in 2.5.2


ivirshup:
### Version of Awkward Array

2.5.2

### Description and code to reproduce

```python
import awkward as ak, numpy as np

awk = ak.Array(np.ones((7, 0)))
form, length, container = ak.to_buffers(ak.to_packed(awk))
awk_from_buf = ak.from_buffers(form, np.int64(length), container)

awk_from_buf.to_list()
```

```pytb
TypeError                                 Traceback (most recent call last)
Cell In[1], line 7
      4 form, length, container = ak.to_buffers(ak.to_packed(awk))
      5 awk_from_buf = ak.from_buffers(form, np.int64(length), container)
----> 7 awk_from_buf.to_list()

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/awkward/highlevel.py:496, in Array.to_list(self)
    492 def to_list(self):
    493     """
    494     Converts this Array into Python objects; same as #ak.to_list.
    495     """
--> 496     return self._layout.to_list(self._behavior)

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/awkward/contents/content.py:1086, in Content.to_list(self, behavior)
   1085 def to_list(self, behavior: dict | None = None) -> list:
-> 1086     return self.to_packed()._to_list(behavior, None)

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/awkward/contents/regulararray.py:1464, in RegularArray.to_packed(self, recursive)
   1462 index_nplike = self._backend.index_nplike
   1463 length = self._length * self._size
-> 1464 content = self._content[: index_nplike.shape_item_as_index(length)]
   1466 return RegularArray(
   1467     content.to_packed(True) if recursive else content,
   1468     self._size,
   1469     self._length,
   1470     parameters=self._parameters,
   1471 )

File ~/miniforge3/envs/scanpy-dev/lib/python3.11/site-packages/awkward/_nplikes/array_module.py:323, in ArrayModuleNumpyLike.shape_item_as_index(self, x1)
    321     return x1
    322 else:
--> 323     raise TypeError(f"expected None or int type, received {x1}")

TypeError: expected None or int type, received 0
```

Notably, this doesn't throw an error in 2.5.1.

The issue may be a bit more specific than the title, but we're running into this on the anndata test suite with some of our round-tripping IO tests.

[...]

PR #3191, opened 2024-07-30
fix: proposal for #2947

maxgalli:
Test against np.integer as well in ArrayModuleNumpyLike.shape_item_as_index such that it does not break if ak.from_buffers is built with a numpy integer instead of an integer.
I have the feeling that this might be an "easy fix" and that maybe the fix should be implemented at a deeper level - please let me know if this is the case.

### DIFF
--- a/src/awkward/operations/ak_from_buffers.py
+++ b/src/awkward/operations/ak_from_buffers.py
@@ -136,6 +136,9 @@ def _impl(
     elif isinstance(form, dict):
         form = ak.forms.from_dict(form)
 
+    if isinstance(length, np.integer):
+        length = int(length)
+
     if not (is_integer(length) and length >= 0):
         raise TypeError("'length' argument must be a non-negative integer")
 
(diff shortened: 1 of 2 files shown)


agoose77:
@maxgalli thanks for this PR! I think you're right that we'd prefer to tackle this slightly differently; presently, our internal representation of shape items is `int` or our unknown length types. So, we'd prefer to cast the NumPy types in the entry-points to our internals, e.g. `ak.from_buffers`.

maxgalli:
> @maxgalli thanks for this PR! I think you're right that we'd prefer to tackle this slightly differently; presently, our internal representation of shape items is `int` or our unknown length types. So, we'd prefer to cast the NumPy types in the entry-points to our internals, e.g. `ak.from_buffers`.

thanks @agoose77, I changed the PR accordingly to perform the casr in `ak.from_buffers`. Are there other functions in which this might be needed?

agoose77:
At this stage, any function that accepts a length e.g. `ak.unflatten`, `ak.from_buffers`, and maybe some of the functions like `ak.pad_none` etc. I think it's reasonable to wait for bug reports there, rather than guessing :)

agoose77:
This LGTM, and I feel confident in merging despite being out of the regular development cycle. @jpivarski I'm notifying you that I'm merging despite feeling comfortable to do so!

agoose77:
@maxgalli thanks for this useful PR! <3

jpivarski:
@all-contributors please add @maxgalli for code

Issue #3185, opened 2024-07-17
PR 3119 introduced overtouching


ikrommyd:
### Version of Awkward Array

`>=2.6.5`

### Description and code to reproduce

Reproducer:
```python
import awkward as ak
import numpy as np

from coffea.nanoevents import NanoEventsFactory
import dask_awkward as dak
import hist.dask as hda

def test_necessary_columns():
    events, report = NanoEventsFactory.from_root(
        {
            "https://github.com/CoffeaTeam/coffea/raw/master/tests/samples/nano_dy.root": "Events"
        },
        metadata={"dataset": "Test"},
        uproot_options={"allow_read_errors_with_report": True},
    ).events()

    events["Electron", "pdgId"] = -11 * events.Electron.charge
    events["Muon", "pdgId"] = -13 * events.Muon.charge
    events["leptons"] = ak.concatenate(
        [events.Electron, events.Muon],
        axis=1,
    )
    events = events[ak.num(events.leptons) >= 3]
    pair = ak.argcombinations(events.leptons, 2, fields=["l1", "l2"])
    pair = pair[(events.leptons[pair.l1].pdgId == -events.leptons[pair.l2].pdgId)]

    pair = pair[
        ak.singletons(
            ak.argmin(
                abs((events.leptons[pair.l1] + events.leptons[pair.l2]).mass - 91.2),
                axis=1,
            )
        )
    ]

    events = events[ak.num(pair) > 0]
    pair = pair[ak.num(pair) > 0][:, 0]

    l3 = ak.local_index(events.leptons)
    l3 = l3[(l3 != pair.l1) & (l3 != pair.l2)]
    l3 = l3[ak.argmax(events.leptons[l3].pt, axis=1, keepdims=True)]
    l3 = events.leptons[l3][:, 0]

    mt = np.sqrt(2 * l3.pt * events.MET.pt * (1 - np.cos(events.MET.delta_phi(l3))))
    q8_hist = (
        hda.Hist.new.Reg(100, 0, 200, name="mt", label="$\ell$-MET transverse mass [GeV]")
        .Double()
        .fill(mt)
    )
    print(dak.necessary_columns(q8_hist))
    if len(str(dak.necessary_columns(q8_hist))) > 1000:
        return 1
    else:
        return 0

exit(test_necessary_columns())
```
The bug is introduced by https://github.com/scikit-hep/awkward/pull/3119

My `git bisect` report:
```
[...]

PR #3193, opened 2024-07-31
fix: issue #3118 and #3185

pfackeldey:
After a longer discussion with @jpivarski and @maxgalli we came up with a solution for issue #3118.

This PR addresses issue #3118 and  by trimming the `Index*Arrays` safely before concatenation and thus avoiding memory growth of the underlying `NumpyArrays`.

In addition, it allows to re-enable laziness (disabled recently in https://github.com/scikit-hep/awkward/pull/3119) for `UnionArray.simplified` and thus avoids touching too many columns as described in https://github.com/scikit-hep/awkward/issues/3185 and https://github.com/scikit-hep/awkward/pull/3119#issuecomment-2234227019.

Best, Peter

### DIFF
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -1180,10 +1180,11 @@ class IndexedArray(IndexedMeta[Content], Content):
         )
 
     def _trim(self) -> Self:
-        if self._index.length == 0:
+        nplike = self._backend.index_nplike
+
+        if not nplike.known_data or self._index.length == 0:
             return self
 
-        nplike = self._backend.index_nplike
         idx_buf = nplike.asarray(self._index.data, copy=True)
         min_idx = nplike.min(idx_buf)
         max_idx = nplike.max(idx_buf)
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -626,7 +626,7 @@ class IndexedArray(IndexedMeta[Content], Content):
             if isinstance(
                 array, (ak.contents.IndexedOptionArray, ak.contents.IndexedArray)
             ):
-                array = array._trim() # see: #3185 and #3119
+                array = array._trim()  # see: #3185 and #3119
                 parameters = parameters_intersect(parameters, array._parameters)
 
                 contents.append(array.content)
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -626,6 +626,7 @@ class IndexedArray(IndexedMeta[Content], Content):
             if isinstance(
                 array, (ak.contents.IndexedOptionArray, ak.contents.IndexedArray)
             ):
+                array = array._trim()
                 parameters = parameters_intersect(parameters, array._parameters)
 
                 contents.append(array.content)
@@ -1177,3 +1178,16 @@ class IndexedArray(IndexedMeta[Content], Content):
                 other.content, index_dtype, numpyarray, all_parameters
             )
         )
+
+    def _trim(self) -> Self:
+        if self._index.length == 0:
+            return self
+
+        min_idx = self._backend.index_nplike.min(self._index.data)
+        max_idx = self._backend.index_nplike.max(self._index.data)
+
+        index = Index(self._index.data - min_idx)
+
+        # left and right trim
+        content = self._content._getitem_range(min_idx, max_idx + 1)
+        return IndexedArray(index, content, parameters=self._parameters)
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -626,7 +626,7 @@ class IndexedArray(IndexedMeta[Content], Content):
             if isinstance(
                 array, (ak.contents.IndexedOptionArray, ak.contents.IndexedArray)
             ):
-                array = array._trim()
+                array = array._trim() # see: #3185 and #3119
                 parameters = parameters_intersect(parameters, array._parameters)
 
                 contents.append(array.content)
@@ -1183,10 +1183,12 @@ class IndexedArray(IndexedMeta[Content], Content):
         if self._index.length == 0:
             return self
 
-        min_idx = self._backend.index_nplike.min(self._index.data)
-        max_idx = self._backend.index_nplike.max(self._index.data)
-
-        index = Index(self._index.data - min_idx)
+        nplike = self._backend.index_nplike
+        idx_buf = nplike.asarray(self._index.data, copy=True)
+        min_idx = nplike.min(idx_buf)
+        max_idx = nplike.max(idx_buf)
+        idx_buf -= min_idx
+        index = Index(idx_buf)
 
         # left and right trim
         content = self._content._getitem_range(min_idx, max_idx + 1)
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -1769,12 +1769,18 @@ class IndexedOptionArray(IndexedOptionMeta[Content], Content):
         )
 
     def _trim(self) -> Self:
-        if self._index.length == 0:
+        nplike = self._backend.index_nplike
+
+        if not nplike.known_data or self._index.length == 0:
             return self
 
-        nplike = self._backend.index_nplike
         idx_buf = nplike.asarray(self._index.data, copy=True)
         only_positive = idx_buf >= 0
+
+        # no positive index at all
+        if not nplike.any(only_positive):
+            return self
+
         min_idx = nplike.min(idx_buf[only_positive])
         max_idx = nplike.max(idx_buf[only_positive])
         idx_buf[only_positive] -= min_idx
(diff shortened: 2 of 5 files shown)


src/awkward/contents/indexedarray.py
```diff
@@ -1177,3 +1178,18 @@ def _is_equal_to(
                 other.content, index_dtype, numpyarray, all_parameters
             )
         )
+
+    def _trim(self) -> Self:
+        if self._index.length == 0:
+            return self
```
jpivarski: Maybe also `return self` if ~~the `self._index.length` is `awkward._nplikes.shape.unknown_length`. (To check that with `isinstance`, the type is `awkward._nplikes.shape.UnknownLength`.)~~ if the `self._nplike.known_data` (robustly checks for typetracers, not just typetracers with unknown lengths).

Since this function never changes the Type or Form of an array, it's legal to skip the implementation for typetracers (which can have unknown length). I think there's a chance that a typetracer wouldn't successfully make it through the code below.

src/awkward/contents/indexedoptionarray.py
```diff
@@ -1767,6 +1768,22 @@ def _is_equal_to(
             )
         )
 
+    def _trim(self) -> Self:
+        if self._index.length == 0:
+            return self
```
jpivarski: Same as above.

jpivarski (approved):
To my eyes, this looks like a good solution, though a lot of tests are failing.

The original PR #3119 had to change a test expectation: [see this](https://github.com/scikit-hep/awkward/pull/3119/files#diff-c51c04703de7914227ddb63829e91761168d9a150fddb46c1f5e2f6f8eab0e59), which explains one of the test failures, but it looks like there are more.

Another might be related to the `== 0` check with typetracer arrays; I made a suggestion below.

Making `UnionArray.project` non-lazy shouldn't require many tests to change, since toggling it in the other direction only upset one test. We should only see the same one complaining again when we toggle it back the other way. Unless `UnionArray.project` and `UnionArray.simplified` need to be toggled in tandem?

src/awkward/contents/indexedoptionarray.py
```diff
@@ -1767,6 +1768,22 @@ def _is_equal_to(
...
+
+        nplike = self._backend.index_nplike
+        idx_buf = nplike.asarray(self._index.data, copy=True)
+        only_positive = idx_buf >= 0
+        min_idx = nplike.min(idx_buf[only_positive])
+        max_idx = nplike.max(idx_buf[only_positive])
```
agoose77, line 1785: @jpivarski should we create a kernel to perform `min_and_max`?

src/awkward/contents/indexedarray.py
agoose77: @jpivarski this _would_ need to touch the index for typetracer, and the `self._content._getitem_range` might also need to know some kind of length information, right? I'd vote in favour of avoiding special-casing typetracer for that reason.

agoose77: i.e.
```python
if self._index.length is not unknown_length and self._index.length == 0:
    return self
```

I _think_ we prefer that kind of check to testing `known_data`, because we might sometimes know buffer lengths during typetracing. Or, maybe not. But it's definitely more exact.

src/awkward/contents/indexedoptionarray.py
jpivarski: To avoid two loops over the data? Maybe that can be a later improvement.

Hmmm. https://stackoverflow.com/q/12200580/1623645

Although that would be an improvement, I don't think it should hold up this PR. Also notice that there are other $\mathcal{O}(n)$ array-traversals in the same function; this won't increase the runtime by a large _fraction_.

src/awkward/contents/indexedarray.py
jpivarski: I made the above suggestion because I thought it might be related to the test failures. If it's not possible for a typetracer to pass through the trimming code, then we'd need a guard to make it work.

But I shouldn't get ahead of myself: the first thing to do is to figure out what is responsible for the test failures, and the second thing is to make the smallest change that addresses the problem case, such as the

```python
if self._index.length is not unknown_length and self._index.length == 0:
    return self
```

that you suggestion. (`unknown_length` can't be `== 0`, right? Why does it need to be spelled out?)

This trimming operation should touch the length of the IndexedArray or IndexedOptionArray. I can't imagine that increasing the overall amount of data touched, though: the length of the IndexedArray or IndexedOptionArray must have been needed already. (Nothing in the `content` needs to be touched because a range-slice is just a view.)

src/awkward/contents/numpyarray.py
```diff
@@ -188,9 +188,14 @@ def _raw(self, nplike=None):
...
     def _form_with_key(self, getkey: Callable[[Content], str | None]) -> NumpyForm:
+        if hasattr(self._data, "inner_shape"):
+            inner_shape = self._data.inner_shape
+        else:
+            inner_shape = self._data.shape[1:]
+
         return self.form_cls(
             ak.types.numpytype.dtype_to_primitive(self._data.dtype),
-            self._data.shape[1:],
+            inner_shape,
```
jpivarski: I thought that this had already been established in a recent PR... #3189.

I don't know why GitHub is showing it as a diff relative to `main`; it's already in `main`. Anyway, it's the right thing to do.

src/awkward/operations/ak_from_buffers.py
```diff
@@ -136,6 +136,9 @@ def _impl(
     elif isinstance(form, dict):
         form = ak.forms.from_dict(form)
 
+    if isinstance(length, np.integer):
+        length = int(length)
```
jpivarski: If the goal is to ensure that any integer-like `length` is cast as an integer, use the more generic check, `numbers.Integral`.

```suggestion
    if isinstance(length, numbers.Integral):
        length = int(length)
```

(If `numbers` needs to be imported to do this, then that change is needed as well.) I could have added `and not isinstance(length, int)`, but the `int` function is idempotent.

By the way, what does this change have to do with the original issue? I'm seeing a few irrelevant diffs, like we're not looking at the difference between your branch and `main`, but that's what GitHub says it's comparing.

In fact, this is in `main`:

https://github.com/scikit-hep/awkward/blob/39541dbf38452d1e0a3dbf2a3097e2dbd68b3521/src/awkward/operations/ak_from_buffers.py#L139-L140

Ah: @maxgalli added it earlier today and @agoose77 approved and merged it. If I had looked at it closely, I personally would have made the above change, and we can still make it, but it would be a new PR.

More importantly, it looks like GitHub is showing differences with respect to the time when @pfackeldey forked the repo. But `wants to merge 9 commits into scikit-hep:main from pfackeldey:fix/issue_3118_prevent_exponential_memory_growth_in_unionarray` is supposed to show diffs with respect to the current `scikit-hep:main`. I think we're looking at a GitHub bug.

jpivarski (approved):
This looks good and ready to merge; the tests @pfackeldey added resolves the `known_data` versus `unknown_length` question that @agoose77 and I were discussing. Also, it's nice that it completely reverts PR #3119 to the state that the code has been in for years. The probability of another unforeseen consequence is smaller than it was with PR #3119.

I noted a few surprising diffs below and even made a (minor) suggested change in one that was actually implemented in PR #3191 earlier today. There's nothing left to do for them, unless @maxgalli wants to make the minor suggested change as a new PR. You can click "resolved" and merge the PR.

src/awkward/operations/ak_from_buffers.py
pfackeldey: Sorry, I messed something up syncing my fork with the latest additions in awkward by you and @maxgalli . It should be solved now.

src/awkward/contents/indexedarray.py
agoose77: Ah, I'm following! Ignore me...

Yes, iirc we have guards on unknown_length to catch erroneous comparisons. Filtering out the identity means thar we explicitly handle it.

lgray:
@ikrommyd could you give this fix a try?

ikrommyd:
Seems to be working. Removes overtouching from both the original `q8_hist` reproducer and the your more minimal one.
Naturally doing this:
```diff
diff --git a/src/awkward/contents/unionarray.py b/src/awkward/contents/unionarray.py
index bb3b6ca9..04cb1f54 100644
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -702,7 +702,7 @@ class UnionArray(UnionMeta[Content], Content):
         nextcarry = ak.index.Index64(
             tmpcarry.data[: lenout[0]], nplike=self._backend.index_nplike
         )
-        return self._contents[index]._carry(nextcarry, True)
+        return self._contents[index]._carry(nextcarry, False)

     @staticmethod
     def regular_index(
```
introduces overtouching again as expected.
I believe https://github.com/scikit-hep/awkward/issues/3185 is to be closed.
Thanks for the fix y'all and if everything is good, I'd be happy to see this in a release soon.

jpivarski:
@all-contributors please add @pfackeldey for code

PR #3195, opened 2024-08-01
fix: change np.integer to numbers.Integral

maxgalli:
Implement change suggested [here](https://github.com/scikit-hep/awkward/pull/3193#discussion_r1699185832)

### DIFF
--- a/src/awkward/operations/ak_from_buffers.py
+++ b/src/awkward/operations/ak_from_buffers.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import math
-import numbers
 
 import awkward as ak
 from awkward._backends.dispatch import regularize_backend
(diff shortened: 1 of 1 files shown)


jpivarski (approved):
I just noticed several lines later that we have a utility function, `is_integer`, which you can use instead of the `numbers` library. (Sorry!)

Instead of

```python
    if isinstance(length, numbers.Integral):
        length = int(length)

    if not (is_integer(length) and length >= 0):
        raise TypeError("'length' argument must be a non-negative integer")
```

do

```python
    if not (is_integer(length) and length >= 0):
        raise TypeError("'length' argument must be a non-negative integer")
    length = int(length)
```

(and then this becomes an extremely small PR). When that's done, go ahead and merge. Thanks!

Issue #3188, opened 2024-07-26
avro reader crash


martindurant:
### Version of Awkward Array

ak 2.6.5
ak-cpp 34

### Description and code to reproduce

I tried to read the avro file at this location https://github.com/Teradata/kylo/blob/master/samples/sample-data/avro/userdata5.avro (downloaded to local) using `ak.from_avro_file` and got `segmentation fault`. The file loads fine with fastavro.

macOS 13.6.7, conda install of python 3.10

jpivarski:
Although there's not much demand (that I know of) for the Avro reader, it's a canary in the coalmine for AwkwardForth and therefore Uproot. If the Avro is not handled correctly, it can fail with a Forth runtime error, but it should not be segfaulting under any circumstances, so this is serious. 

I'll try to reproduce it sometime when I have access to my main computer again and see if I can narrow in on the point of failure.

martindurant:
In addition, from_avro_file fails if passed a file-like rather than a str filename. On [this line](https://github.com/scikit-hep/awkward/blob/main/src/awkward/operations/ak_from_avro_file.py#L61), the attribute should be `.outcontents` (as in the branch above, L50) not `.outarr`.

jpivarski:
Has PR #3167 solved this issue?

martindurant:
I didn't see that PR. I can test.

martindurant:
Updating awkward to 55a9a432 (current main) still crashes. Is there a chance of any cached state from a previous run?
awkward-cpp == 36

jpivarski:
> Is there a chance of any cached state from a previous run?

I'm not sure. GitHub Actions recompiles awkward-cpp in each test, so we _shouldn't_ cache from previous runs. (@agoose77, did the test ever have a cache like that?)

Meanwhile, is this an awkward-cpp-only reproducer?

```bash
python -c 'from awkward_cpp.lib._ext import ForthMachine32 ; vm = ForthMachine32("1 2 3 4 5"); vm.run(); print(vm.stack)'
```

I'm thinking that we should bisect awkward-cpp itself, removing parts of it until there isn't a segfault and then see which part makes the difference. Awkward relies on many parts of awkward-cpp, but awkward-cpp itself is mostly a bag of functions and classes; it should be easy to remove large parts of it and still be able to rely on the above test.

martindurant:
[...]

PR #3209, opened 2024-08-09
fix: adjust checks in ForthMachine to prevent segfault when num_items is negative

ariostas:
This PR addresses bug that was found in #3188. For some particular file, the `num_items` variable ended up being negative and it was causing a segfault when trying to write some data. I adjusted and added some checks to prevent the segfault and make it crash gracefully. Nevertheless, it does not address the issue with reading the file in #3188.

I checked with the Compiler Explorer and both of these options produce the same machine code, so I just used the more expressive one.
```cpp
if (num_items < 0) num_items = 0;
num_items *= (num_items >= 0)
```

### DIFF
--- a/awkward-cpp/src/libawkward/forth/ForthMachine.cpp
+++ b/awkward-cpp/src/libawkward/forth/ForthMachine.cpp
@@ -3075,7 +3075,7 @@ namespace awkward {
             uint64_t tmp;
             uint8_t tmpbyte;
 
-            if (items_remaining != 0) {
+            if (items_remaining > 0) {
               tmpbyte = input->read_byte(current_error_);
               if (current_error_ != util::ForthError::none) {
                 return;
@@ -3087,7 +3087,7 @@ namespace awkward {
               }
               data = tmp;
             }
-            while (items_remaining != 0) {
+            while (items_remaining > 0) {
               if (bits_wnd_r >= 8) {
                 bits_wnd_r -= 8;
                 bits_wnd_l -= 8;
@@ -3230,6 +3230,7 @@ namespace awkward {
                 break;                                                         \
               }
 
+            if (num_items < 0) num_items = 0;
             switch (format) {
               case READ_BOOL:    WRITE_DIRECTLY(bool, bool)
               case READ_INT8:    WRITE_DIRECTLY(int8_t, int8)
@@ -3311,6 +3312,7 @@ namespace awkward {
                 break;                                                         \
               }
 
+            if (num_items < 0) num_items = 0;
             switch (format) {
               case READ_BOOL:    WRITE_TO_STACK(bool)
               case READ_INT8:    WRITE_TO_STACK(int8_t)
--- a/docs/reference/awkwardforth.rst
+++ b/docs/reference/awkwardforth.rst
@@ -1219,6 +1219,20 @@ and
 
 but the second is faster because it involves two Forth instructions and one ``memcpy``.
 
+If the number of items to read is negative then it is interpreted as zero.
+
+.. code-block:: python
+
+    >>> vm = ForthMachine32("""
+    ... input x
+    ... output y float32
+    ... 
+    ... -1000000 x #d-> y
+    ... """)
+    >>> vm.run({"x": np.arange(1000000) * 1.1})
+    >>> np.asarray(vm["y"])
+    array([], dtype=float32)
+
 Type codes
 """"""""""
 
--- a/tests/test_3209_awkwardforth_read_negative_number_of_items.py
+++ b/tests/test_3209_awkwardforth_read_negative_number_of_items.py
@@ -1,0 +1,45 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
+
+from __future__ import annotations
+
+import numpy as np
+
+import awkward as ak
+
+
+def test_read_negative_number_of_items():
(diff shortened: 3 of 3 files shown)


ariostas:
Maybe to make it more complete it should set `current_error_` to some (possibly new) type of error?

jpivarski:
> Maybe to make it more complete it should set `current_error_` to some (possibly new) type of error?

That was my first thought, since this is another thing that could go wrong. However, for some AwkwardForth [type codes](https://awkward-array.org/doc/main/reference/awkwardforth.html#input-read), it has effectively been taking the maximum of the stack-read variable and zero for a while now. Turning this into an error would make code development in AwkwardForth more developer-friendly, but most of the development that (I believe) will ever be done in this language has already been done. Its existence was motivated by accelerating Uproot TTree-reading, and as we move to RNTuple, it becomes irrelevant. In principle, it's useful for other non-columnar data formats whose types are static through the dataset but not known before reading (i.e. the type is specified in a file-bound schema). Avro is one example of that, Protobuf, and Thrift are others, maybe also Flatbuffers, but there hasn't been high demand for getting these formats into Awkward Arrays. AwkwardForth is not the best choice for fully dynamic formats (JSON, BSON, etc.) or columnar formats (Parquet), or if a JIT-compiler is available. It's fairly niche—great for the Uproot TTree use-case—but that's done.

So I'd be happy to turn the accidental interpretation of negative `num_items` as zero into an explicit rule in the language. It can be documented as a short sentence somewhere in here:

https://github.com/scikit-hep/awkward/blob/af232f545a01d06dd269d5ed16df38ec4469b30e/docs/reference/awkwardforth.rst?plain=1#L1183-L1220

For tests (for this PR), how about the following? I ran these on the old/broken AwkwardForth on Linux.

Positive `num_items`; no issues:

```python
>>> vm = awkward.forth.ForthMachine32("input source 5 source #q-> stack")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.stack
[1, 2, 3, 4, 5]
```

and

```python
>>> vm = awkward.forth.ForthMachine32("input source output sink float64 5 source #q-> sink")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.output("sink")
array([1., 2., 3., 4., 5.])
```

Negative `num_items`; some issues (but no segfault on Linux):

```python
>>> vm = awkward.forth.ForthMachine32("input source -5 source #q-> stack")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.stack
[]
```

and

```python
>>> vm = awkward.forth.ForthMachine32("input source output sink float64 -5 source #q-> sink")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.output("sink")
ValueError: negative dimensions are not allowed
```

Maybe these would segfault on MacOS, or maybe the segfault only happens deeper in a more complex example. But anyway, with the fix currently implemented in this PR, the first would return the same thing and the second would also return an empty array.

Actually, maybe it would be better to test a negative `num_items` followed by a positive `num_items` (or vice-versa), which are showing bad behavior on Linux. (Keep in mind that I'm not running in this PR.)

```python
>>> vm = awkward.forth.ForthMachine32("input source -5 source #q-> stack 5 source #q-> stack")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.stack
[1068248464, 3, -1170705504, 0, 49]
```

With this PR, the output should be `[1, 2, 3, 4, 5]` because the first `-5 source #q-> stack` will now be interpreted as a no-op.

```python
>>> vm = awkward.forth.ForthMachine32("input source output sink float64 -5 source #q-> sink 5 source #q-> sink")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.output("sink")
array([], dtype=float64)
```

This should also be `[1, 2, 3, 4, 5]` for the same reason.

This is clearly doing bad things in Linux because a few steps _after_ these steps, I got a core dump.

In a new process, I tested the positive `num_items` followed by negative `num_items`, and these should also both return `[1, 2, 3, 4, 5]`, which they do not (without this PR):

```python
>>> vm = awkward.forth.ForthMachine32("input source 5 source #q-> stack -5 source #q-> stack")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.stack
[1, 2, 3, 4, 5]
```

and

```python
>>> vm = awkward.forth.ForthMachine32("input source output sink float64 5 source #q-> sink -5 source #q-> sink")
>>> vm.run({"source": np.array([1, 2, 3, 4, 5])})
>>> vm.output("sink")
array([], dtype=float64)
```

ariostas:
Thanks so much, Jim. I added the rule to the documentation and the tests you suggested. I couldn't find if the `awkward-cpp` tests should go somewhere else, so let me know if I should move them.

jpivarski (approved):
This all looks good and I'd say it's ready to merge!

I don't know which awkward-cpp tests you mean, but almost all of the awkward-cpp testing is done in Python through Awkward, anyway. (The major exception is the kernel-testing, but that's unrelated to AwkwardForth.) These tests are definitely sufficient.

ariostas:
Hmm I'll have to check why some Windows tests are failing.

ariostas:
Welp, it's the second time this week where I run into issues by not explicitly specifying `dtype`. But now it should all be good to go.

ianna (approved):
Looks great! Thanks!

jpivarski:
@all-contributors please add @ariostas for code

PR #3215, opened 2024-08-14
feat: Add ak.array_equal

### LINKED ISSUE #1105: ak.array_equal to override np.array_equal

tcawlfield:
Adding `ak.operations.array_equal`, based on `numpy.array_equal`

This uses `...ak_almost_equal._impl()` which now has more parameters to
suit some particular needs of array_equal. Even so, `ak.almost_equal()` remains
unchanged operationally.

Note that there was broken code in `src/awkward/_nplikes/array_module.py`:
```
both_nan = self._module.logical_and(x1 == np.nan, x2 == np.nan)
```
The expressions like `x1 == np.nan` is all False even when x1 contains NaN values.
I'm hoping that my fix works for older versions of numpy. It works for numpy 1.26.4 and 2.0.1.

### DIFF
--- a/src/awkward/operations/ak_almost_equal.py
+++ b/src/awkward/operations/ak_almost_equal.py
@@ -106,7 +106,7 @@ def _impl(
 
     def visitor(left, right) -> bool:
         # Most firstly, check same_content_types before any transformations
-        if same_content_types and left.__class__ != right.__class__:
+        if same_content_types and left.__class__ is right.__class__:
             return False
 
         # First, erase indexed types!
--- a/src/awkward/operations/ak_almost_equal.py
+++ b/src/awkward/operations/ak_almost_equal.py
@@ -60,20 +60,23 @@ def almost_equal(
         dtype_exact=dtype_exact,
         check_parameters=check_parameters,
         check_regular=check_regular,
+        exact_eq=False,
+        same_content_types=False,
+        equal_nan=False,
     )
 
 
 def _impl(
     left,
     right,
-    rtol: float = 1e-5,
-    atol: float = 1e-8,
-    dtype_exact: bool = True,
-    check_parameters: bool = True,
-    check_regular: bool = True,
-    exact_eq: bool = False,
-    same_content_types: bool = False,
-    equal_nan: bool = False,
+    rtol: float,
+    atol: float,
+    dtype_exact: bool,
+    check_parameters: bool,
+    check_regular: bool,
+    exact_eq: bool,
+    same_content_types: bool,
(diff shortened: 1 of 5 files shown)


src/awkward/operations/ak_almost_equal.py
```diff
@@ -52,6 +52,29 @@ def almost_equal(
...
+
+
+def _impl(
+    left,
+    right,
+    rtol: float = 1e-5,
+    atol: float = 1e-8,
+    dtype_exact: bool = True,
+    check_parameters: bool = True,
+    check_regular: bool = True,
```
jpivarski:
```suggestion
    rtol: float,
    atol: float,
    dtype_exact: bool,
    check_parameters: bool,
    check_regular: bool,
```

Duplicating the user-facing default arguments with internal default arguments runs the risk of letting them diverge, since there would appear to be two sources of truth.

src/awkward/operations/ak_almost_equal.py
```diff
@@ -82,6 +105,10 @@ def packed_list_content(layout):
         return layout.content[layout.offsets[0] : layout.offsets[-1]]
 
     def visitor(left, right) -> bool:
+        # Most firstly, check same_content_types before any transformations
+        if same_content_types and left.__class__ != right.__class__:
```
jpivarski:
```suggestion
        if same_content_types and left.__class__ is right.__class__:
```

I'm surprised that a Python linter didn't complain about that, but class objects should be checked for reference equality, since they ought to be singletons (like `None`). In fact, a linter might even complain about the mere fact that you're checking class equality, rather than `issubclass(left.__class__, right.__class__)`, though you really do want class equality, not inheritance.

jpivarski (approved):
I like the way that the implementation reuses code between `ak.array_equal` and `ak.almost_equal`. This is a good approach; all that's needed is to fix the test errors and a few suggested changes below.

src/awkward/operations/ak_almost_equal.py
tcawlfield: Good point. I fixed these and used consistent parameter ordering.

PR #3217, opened 2024-08-15
ci: add 3.13 wheels

henryiii:
Adding 3.13 (non-free-threaded at the moment).

### DIFF
--- a/.github/workflows/build-wheels.yml
+++ b/.github/workflows/build-wheels.yml
@@ -105,7 +105,7 @@ jobs:
     - name: Prepare build files
       run: pipx run nox -s prepare
 
-    - uses: pypa/cibuildwheel@v2.19
+    - uses: pypa/cibuildwheel@v2.20
       env:
         CIBW_BUILD: "${{ matrix.build }}*"
         CIBW_ARCHS: ${{ matrix.arch }}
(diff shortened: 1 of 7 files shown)


henryiii:
```
  E       AssertionError: Regex pattern did not match.
  E        Regex: "Can't pickle"
  E        Input: "Can't get local object 'test_serialise_with_nonserialisable_attrs.<locals>.<lambda>'"
```

lgray:
@henryiii "cp313", "cp313t" are now available and numpy is also (as of a few hours ago) producing 3.13t wheels as well. Worth another try now!

henryiii:
Passing.

lgray:
Guessing because of cramjam workaround we presently cannot test 3.13t?

henryiii:
3.13 works with the workaround. Either cramjam happens to work with it or we don’t use enough of it to be affected by any breakage.

3.13t probably not.

henryiii:
Ah, yes, you have the t.

lgray:
Where do you see it in the tests? Or are you saying I'm asking for it and we don't have it? If the latter, I guess we just wait a bit for setup-python to make a release.

henryiii:
We aren't building 3.13t yet in cibuildwheel, so it's not built or tested. I'm assuming there's no chance this workaround works on 3.13t; I'm not even sure PyO3 main supports 3.13t yet.

For PyO3 (current, not old version), see https://github.com/Quansight-Labs/free-threaded-compatibility/issues/53.

lgray:
FYI https://github.com/milesgranger/cramjam/releases/tag/v2.8.4-rc3

jpivarski (approved):

jpivarski:
All the tests that this PR branch runs are now required, and nothing that it doesn't run is required. Now I'll update and auto-merge.

PR #3222, opened 2024-08-21
feat: provide a way for `Form.select_columns` to distinguish structural dots from dots in the names of record fields

### LINKED ISSUE #3088: `Form.select_columns` needs a way to distinguish structural dots from dots in the names of record fields
### Version of Awkward Array

HEAD

Suppose you have an array like

```python
>>> array = ak.Array([{"outer": {"crazy.dot": {"inner_a": 12.3, "inner_b": 123}}}])
>>> array.show(type=True)
type: 1 * {
    outer: {
        "crazy.dot": {
            inner_a: float64,
            inner_b: int64
        }
    }
}
[{outer: {'crazy.dot': {inner_a: 12.3, inner_b: 123}}}]
```

If you want to Form-select `inner_a` and not `inner_b`, you can't:

```python
>>> array.layout.form.select_columns("outer.crazy.dot.inner_a")
RecordForm([], [])
```

When this problem is encountered in Spark, they use backticks. To do the same, we'd want

```python
>>> array.layout.form.select_columns("outer.`crazy.dot`.inner_a")
RecordForm([], [])
```

to result in

```python
RecordForm([RecordForm([RecordForm([NumpyForm('float64')], ['inner_a'])], ['crazy.dot'])], ['outer'])
```

which means a bit more parsing of the specifier.

tcawlfield:
Addressing the need to select field names with literal dots in them.

This is a change to `Form.select_columns()`. This change allows a user to supply a `specifier` as a list of lists of field (or column) names.

### DIFF
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -437,6 +437,19 @@ class Form(Meta):
     def select_columns(
         self, specifier, expand_braces=True, *, prune_unions_and_records: bool = True
     ):
+        """
+        select_columns returns a new Form with only columns and sub-columns selected.
+        Returns an empty Form if no columns matched the specifier(s).
+
+        `specifier` can be a `str | Iterator[str | list[str] | tuple[str]]`.
+        Strings may include shell-globbing-style wildcards "*" and "?".
+        If `expand_braces` is `True` (the default), strings may include alternatives in braces.
+        For example, `["a.{b,c}.d"]` is equivalent to `["a.b.d", "a.c.d"]`.
+        Glob-style matching would also suit this single-character instance: `"a.[bc].d"`.
+        If specifier is a list which contains a list/tuple, that inner list will be interpreted as
+        column and subcolumn specifiers. They *may* contain wildcards, but "." will not be
+        interpreted as a `<field>.<subfield>` pattern.
+        """
         if isinstance(specifier, str):
             specifier = {specifier}
 
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -441,7 +441,7 @@ class Form(Meta):
         select_columns returns a new Form with only columns and sub-columns selected.
         Returns an empty Form if no columns matched the specifier(s).
 
-        `specifier` can be a `str | Iterator[str | list[str] | tuple[str]]`.
+        `specifier` can be a `str | Iterable[str | Iterable[str]]`.
         Strings may include shell-globbing-style wildcards "*" and "?".
         If `expand_braces` is `True` (the default), strings may include alternatives in braces.
         For example, `["a.{b,c}.d"]` is equivalent to `["a.b.d", "a.c.d"]`.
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
This looks great!

The one error doesn't _seem_ to be related to this, and it's persistent enough to survive re-running the worker. (It looked to me like an installation/packaging issue, so that's why I tried re-running.)

```
E       AssertionError: Regex pattern did not match.
E        Regex: "Can't pickle"
E        Input: "Can't get local object 'test_serialise_with_nonserialisable_attrs.<locals>.<lambda>'"
```

The second error message does seem to be related to not being able to pickle; lambda functions generally aren't pickleable. But somehow the "Can't pickle" message isn't there.

src/awkward/forms/form.py
```diff
@@ -437,29 +437,59 @@ def columns(self, list_indicator=None, column_prefix=()):
...
 
         # Only take unique specifiers
         for item in specifier:
-            if not isinstance(item, str):
+            if isinstance(item, str):
+                if item == "":
+                    raise ValueError(
+                        "A column-selection specifier cannot be an empty string"
+                    )
+            elif isinstance(item, (tuple, list)):
```
jpivarski:
```suggestion
            elif isinstance(item, Iterable):
```

and

```python
from collections.abc import Iterable
```

The danger here is that `str` is also `Iterable`, but it's okay if you check for `Iterable` after checking for `str`.

This way, if someone has some other type, like a `set`, it will still work. It's not great that `dict` is also interpreted as an iterable (of its keys only), but at least that's Python convention and expected.

src/awkward/forms/form.py
```diff
@@ -437,29 +437,59 @@ def columns(self, list_indicator=None, column_prefix=()):
...
         if isinstance(specifier, str):
             specifier = {specifier}
 
         # Only take unique specifiers
         for item in specifier:
-            if not isinstance(item, str):
+            if isinstance(item, str):
+                if item == "":
+                    raise ValueError(
+                        "A column-selection specifier cannot be an empty string"
```
jpivarski: There's a convention for error messages to be lowercase without ending punctuation.

```suggestion
                        "a column-selection specifier cannot be an empty string"
```

src/awkward/forms/form.py
```diff
@@ -437,29 +437,59 @@ def columns(self, list_indicator=None, column_prefix=()):
...
+                    )
+            elif isinstance(item, (tuple, list)):
+                for field in item:
+                    if not isinstance(field, str):
+                        raise ValueError("A sub-column specifier must be a string")
```
jpivarski:
```suggestion
                        raise ValueError("a column specifier must be a string or an iterable of strings")
```

src/awkward/forms/form.py
```diff
@@ -437,29 +437,59 @@ def columns(self, list_indicator=None, column_prefix=()):
...
+            else:
                 raise TypeError(
-                    "a column-selection specifier must be a list of non-empty strings"
-                )
-            if not item:
-                raise ValueError(
-                    "a column-selection specifier must be a list of non-empty strings"
+                    "A column-selection specifier must be a list of strings, lists, or tuples"
```
jpivarski:
```suggestion
                    "a column specifier must be a string or an iterable of strings"
```

src/awkward/operations/ak_from_parquet.py
```diff
@@ -32,10 +32,11 @@ def from_parquet(
     Args:
         path (str): Local filename or remote URL, passed to fsspec for resolution.
             May contain glob patterns.
-        columns (None, str, or list of str): Glob pattern(s) with bash-like curly
+        columns (None, str, or list of (str or list of str)): Glob pattern(s) including bash-like curly
```
jpivarski:
```suggestion
        columns (None, str, or list of (str or iterable of str)): Glob pattern(s) including bash-like curly
```

jpivarski (approved):
Selecting everything that matches, when there's ambiguity, is a good choice.

Thanks for the detailed docstring on column selector syntax!

I just have a few comments on the error messages themselves, and broadening the argument to accept any iterable, not just lists and tuples. (It's not unusual for a Python user to not even know the exact type of collection they're using, just that it's a collection. For instance, it could be an array, a set, an iterator from `some_dict.values()` (which is not a list), etc.)

jpivarski:
Also, I've updated the branch, so be sure to `git pull` if you're editing a local copy (instead of just patching in small fixes through the web interface).

PR #3225, opened 2024-08-23
fix: add `ak.array_equal` to NEP18 overrides, documentation, and add one more test

tcawlfield:
* Now overrides numpy.array_equal
* Adding link to toctree
* Unit test checks numpy override

### DIFF
--- a/docs/reference/toctree.txt
+++ b/docs/reference/toctree.txt
@@ -299,11 +299,12 @@
     generated/ak.backend
 
 .. toctree::
-    :caption: Approximation
+    :caption: Approximation and comparison
 
     generated/ak.round
     generated/ak.isclose
     generated/ak.almost_equal
+    generated/ak.array_equal
 
 .. toctree::
     :caption: NumPy compatibility
--- a/src/awkward/operations/ak_almost_equal.py
+++ b/src/awkward/operations/ak_almost_equal.py
@@ -2,10 +2,11 @@
 
 from __future__ import annotations
 
-from awkward._backends.dispatch import backend_of_obj
+from awkward._backends.dispatch import backend_of
 from awkward._backends.numpy import NumpyBackend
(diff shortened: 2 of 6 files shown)


jpivarski (approved):
This is good; I agree that the heading `:caption:` should be "approximation and comparison".

The first set of test failures are due to the fact that Awkward only supports native-endian numbers, and you have some tests that make big-endian unsigned int32. I don't know what the second issue is.

Once those are fixed, this is mergeable!

tcawlfield:
> The first set of test failures are due to the fact that Awkward only supports native-endian numbers, and you have some tests that make big-endian unsigned int32. I don't know what the second issue is.

Oh, a regression failure! I'm seeing the first of these errors on my own system once I run *all* tests. Hmm. The failure [is here](https://github.com/scikit-hep/awkward/blob/f9304c92721c774dbb424ca7002b36ad8d48a114/tests/test_2305_nep_18_lazy_conversion.py#L14).

`almost_equal` is [getting the layout](https://github.com/scikit-hep/awkward/blob/769ee4032b5eaa66ee7586daa6f11494235c5607/src/awkward/operations/ak_almost_equal.py#L91-L92) of both inputs sooner than it really can in the case that one array is a pure numpy array. Maybe I need to add some earlier tests of that condition -- if one (or both I suppose) array is raw numpy (or cupy or jax etc), then do a simpler set of tests. I could also try/except and ... no, that won't work. Or maybe only run `to_layout(...)` on Array instances, and only `to_packed()` on particular layout classes?

I'll work on this.

jpivarski (approved):
This makes sense (the update in ec93fd2d9c76da5c5cb4cf6bfb9e5e82af3d784b): "same backend" is a broader criterion than someone is likely to be asking about when they ask whether two arrays are equal. Having it raise an error, like all of the other functions, is good for consistency.

PR #3228, opened 2024-09-02
fix: cuda max reducer bugfix and more tests

### LINKED ISSUE #3219: CUDA `ak.max` result on `cuda` backend doesn't agree with baseline `cpu`
### Version of Awkward Array

master branch (2.6.7)

It seems like switching to a recommended random generator revealed a bug. To reproduce:

```python
import numpy as np
import cupy as cp
import awkward as ak
rng = np.random.default_rng(seed=42)
array = rng.integers(3000, size=3000)
content = ak.contents.NumpyArray(array)
cuda_content = ak.to_backend(content, "cuda", highlevel=False)
```
one can check that:
```python
>>> np.max(array)
2998
>>> ak.max(content)
2998
>>> ak.max(cuda_content)
array(2997)
>>> len(cuda_content)
3000
>>> len(content)
3000
>>> ak.argmax(content)
2261
>>> cuda_content[2261]
array(2998)
```

ianna:
fixes https://github.com/scikit-hep/awkward/issues/3219

### DIFF
--- a/src/awkward/_connect/cuda/cuda_kernels/awkward_reduce_max.cu
+++ b/src/awkward/_connect/cuda/cuda_kernels/awkward_reduce_max.cu
@@ -60,7 +60,7 @@ awkward_reduce_max_b(
         T val = identity;
 
         if (idx >= stride && thread_id < lenparents && parents[thread_id] == parents[thread_id - stride]) {
-          val = temp[idx - stride];
+          val = temp[thread_id - stride];
         }
         __syncthreads();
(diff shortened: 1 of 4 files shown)


ianna (commented):
@jpivarski - please, feel free to merge it if the test pass. This PR only fixes a max reducer bug. The test that fails on the first run issue will be a separate PR. Thanks!

jpivarski (approved):
Looks good! The tests pass, too. I ought to check it on my GPU, but it will be easier to do that next week. (It can wait that long.)

Meanwhile, I'll merge this now. Thanks!

PR #3234, opened 2024-09-09
fix: empty arrays in ak.to_parquet with extensionarray=True

jpivarski:
A corner-case of #3127, but an important one.


ikrommyd:
Fixes https://github.com/dask-contrib/dask-awkward/issues/541

Tested this and it solves the problem in eager awkward and also solves the reproducer at https://github.com/dask-contrib/dask-awkward/issues/541. Happy with merging and honeslty, I wouldn't say no to an awkward release if you aren't waiting for anything.

jpivarski:
To do a release, I'll need to take a look at the open PRs to see if anything else is waiting. We'll be having our first Awkward-Uproot meeting in a long time this Thursday; we should get a release by then (or someone will remind me of a good reason to not make a release).

Issue #3214, opened 2024-08-14
CUDA test failure: `tests-cuda/test_3149_complex_reducers.py::test_block_boundary_prod_complex13`, but only the _first time_ it is run


jpivarski:
### Version of Awkward Array

HEAD

### Description and code to reproduce

The first time I ran `pytest tests-cuda` on a new system, I got one test failure:

```
=============================================================== test session starts ================================================================
platform linux -- Python 3.11.9, pytest-8.3.2, pluggy-1.5.0
Matplotlib: 3.9.1
Freetype: 2.12.1
rootdir: /home/jpivarski/irishep/awkward
configfile: pyproject.toml
plugins: forked-1.6.0, timeout-2.3.1, reverse-1.7.0, mpl-0.17.0, anyio-4.4.0, mock-3.14.0, cov-5.0.0, xdist-3.6.1
collected 694 items                                                                                                                                

tests-cuda/test_1276_cuda_num.py .........                                                                                                   [  1%]
tests-cuda/test_1276_cuda_transfers.py ................                                                                                      [  3%]
tests-cuda/test_1276_cupy_interop.py .                                                                                                       [  3%]
tests-cuda/test_1276_from_cupy.py .....                                                                                                      [  4%]
tests-cuda/test_1300_same_for_numba_cuda.py .......................                                                                          [  7%]
tests-cuda/test_1381_check_errors.py .                                                                                                       [  7%]
tests-cuda/test_1809_array_cuda_jit.py ..............                                                                                        [  9%]
tests-cuda/test_2327_array_interface.py .                                                                                                    [ 10%]
tests-cuda/test_2649_dlpack_support.py .                                                                                                     [ 10%]
tests-cuda/test_2922a_new_cuda_kernels.py ......................................................................                             [ 20%]
tests-cuda/test_2922b_new_cuda_kernels.py .............................                                                                      [ 24%]
tests-cuda/test_3065a_cuda_kernels.py ...................................................................................................... [ 39%]
.....................................................................................................................................        [ 58%]
tests-cuda/test_3065b_cuda_kernels.py ........................                                                                               [ 61%]
tests-cuda/test_3065c_cuda_kernels.py ...................................................                                                    [ 69%]
tests-cuda/test_3086_cuda_concatenate.py ....................                                                                                [ 72%]
tests-cuda/test_3115_array_typed_cuda_jit.py .                                                                                               [ 72%]
[...]

PR #3235, opened 2024-09-12
fix: GPU complex reducer prod for empty lists

ianna:
fixes issue #3214

The test failure is due to a mismatch between the expected and actual results in the assertion `cpt.assert_allclose`. The issue arises when comparing the results of the `ak.prod `operation on arrays converted to the CUDA backend (`cuda_depth1`) and the original CPU backend (`depth1`). Specifically, there's a discrepancy in the third element of the arrays (which is an `EmptyArray`) being compared: `0j` versus (`1.4641 - 0j`).

The product of no element is defined as 1.

### DIFF
--- a/src/awkward/_connect/cuda/cuda_kernels/awkward_reduce_prod_complex.cu
+++ b/src/awkward/_connect/cuda/cuda_kernels/awkward_reduce_prod_complex.cu
@@ -59,8 +59,8 @@ awkward_reduce_prod_complex_b(
 
     if (thread_id < lenparents) {
       for (int64_t stride = 1; stride < blockDim.x; stride *= 2) {
-        T real = (T)1.0f;
-        T imag = (T)0.0f;
+        T real = (T)1;
+        T imag = (T)0;
         if (idx >= stride && thread_id < lenparents && parents[thread_id] == parents[thread_id - stride]) {
           real = temp[(idx - stride) * 2];
           imag = temp[(idx - stride) * 2 + 1];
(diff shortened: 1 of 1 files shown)


jpivarski (commented):
I don't know why this makes a difference. As I understand it, `T` is `float` when the kernel is applied to complex64 and `T` is `double` when the kernel is applied to complex128. C-style casting should do a conversion-cast to the appropriate type, regardless of whether the constant was initially `int` or `float` (and it should do it at compile-time).

```c++
root [0] (float)1.0f
(float) 1.00000f
root [1] (double)1.0f
(double) 1.0000000
root [2] (float)1
(float) 1.00000f
root [3] (double)1
(double) 1.0000000
```

But if it does make a difference, what test demonstrates that? Just adding a test that fails for the old code and passes for the new code would be convincing enough.

ianna:
> I don't know why this makes a difference. As I understand it, `T` is `float` when the kernel is applied to complex64 and `T` is `double` when the kernel is applied to complex128. C-style casting should do a conversion-cast to the appropriate type, regardless of whether the constant was initially `int` or `float` (and it should do it at compile-time).
> 
> ```c++
> root [0] (float)1.0f
> (float) 1.00000f
> root [1] (double)1.0f
> (double) 1.0000000
> root [2] (float)1
> (float) 1.00000f
> root [3] (double)1
> (double) 1.0000000
> ```
> 
> But if it does make a difference, what test demonstrates that? Just adding a test that fails for the old code and passes for the new code would be convincing enough.

Please, check the test that was reported in #3214 - it does not fail for me anymore. However, I was thinking along the same lines: why it makes a difference? I did some more digging. It turns out that we have the same or similar issue with `ak.prod` on CPU: https://github.com/scikit-hep/awkward/issues/3236.

jpivarski (approved):
(In our meeting, @ianna pointed out that one of the tests was failing without this fix. So we do have a test that is sensitive to it.)

All tests-cuda pass on my GPU as well, so now we've double-tested it.

I'll merge this because it was passing all CI tests before the set of required tests was partially updated. _Partial_ because the new tests in #3217 haven't been run in a while, so they're not in the drop-down menu to add them until I run them again. That was taking a while, so I came over here to take care of this PR, without the other one being done first.

Issue #2596, opened 2023-07-27
Add support for annotated axis names


agoose77:
### Description of new feature

At the PyHEP.dev workshop, it was observed that the axis parameter has the potential to harm readability of an analysis. This follows from the idea that the semantic interpretation of the `axis` parameter requires knowledge of the array structure, which is normally a strong function of the execution history.

The notes from the workshop are below:
 
> 
> Present: Angus, Alex, Ioana, Matthew, Ianna, Jim, Tal, Remco, Jonas, Mason, Clemens, 
> 
> - We should be able to label / name axes for readability.
>     - Awkward Array can/should add support for named axes, c.f. XArray with dimension names
>     - No need to add labels because they're much more specific to plotting
>     - New function e.g. `array = ak.with_axis_name(array, 0, "events")` to permit `ak.sum(array, axis="events")`
> - The `flatten(array[argmax(..., keepdims=True)])` pattern is complex to understand, despite being a *pattern*
>     - We could add an accessor that allows us to force ragged indexing. This permits to have a single-index accessor if needs be, that directly consumes the result of `argmax(..., keepdims=False)`, e.g. `array.at[...]`. We could also have a similar-yet-different `array.select[...]` that accepts a `keepdims=True` positional reducer result, and flattens the result afterwards (we know that `keepdims=True` produces regular dimensions, which can be identified statically). 
> - Interest in removing repeated array names `event.foo.x[event.foo.y > 2]` → `event.foo.x[.y > 2]` (or better)
>     - Could introduce a `this` proxy that refers to array being sliced.
> 
> - Tools:
>     - Xarray
>         - Example from user guide: https://docs.xarray.dev/en/stable/user-guide/indexing.html#
>     - Nathan: [JAX also has some (experimental) support for named axes and parallelism](https://jax.readthedocs.io/en/latest/notebooks/xmap_tutorial.html#introducing-and-eliminating-named-axes)
> - Examples
>     - [Alex example from AGC](https://gist.github.com/alexander-held/a9ff4928fe33e8e0c09dd27fbdfd24d9)
>     - [Gordon fantasy rewrite](https://gist.github.com/gordonwatts/87d29b9e1dd13f0958968cd194e7b929) of Alex cell's 3.

My current thinking on this, following from the workshop discussion, is that we should add a mechanism for labelling axes with a name, such that this name can later be used in place of the integer value.

```python
import awkward as ak

array = ak.Array([
	[{'pt': 0.0, ...}],
    ...
])

[...]

PR #3238, opened 2024-09-12
feat: named axis for `ak.Array`

pfackeldey:
# Proposal for named axis

This PR addresses https://github.com/scikit-hep/awkward/issues/2596.

References for other named axis implementations:
- [hist](https://hist.readthedocs.io/en/latest/)
- [haliax](https://github.com/stanford-crfm/haliax)
- [Tensor Considered Harmful](https://nlp.seas.harvard.edu/NamedTensor.html)
- [PyTorch Named Tensors](https://pytorch.org/docs/stable/name_inference.html#name-inference-reference-doc)
- [Penzai Named Axis](https://penzai.readthedocs.io/en/stable/notebooks/named_axes.html)
- [xarray Named Axis](https://docs.xarray.dev/en/stable/user-guide/indexing.html#)

## Motivation

As argumented at [PyHEP.dev 2023](https://indico.cern.ch/event/1234156/) and by the Harvard NLP group in their ["Tensor Considered Harmful"](https://nlp.seas.harvard.edu/NamedTensor.html) write-up, named axis can be a powerful tool to make code more readable and less error-prone.

### `ak.Array` with named axis

Named axis are implemented through a mapping from *named axis* to *positional axis*.
*named axis* are hashables (currently restricted to strings), except for integers (and None) as they are reserved for *positional axis*.

```python
import typing

AxisName: typing.Alias = typing.Hashable
```

By default an `ak.Array` uses positional axis, but named axis can be added to the array in the following ways:

```python
import awkward as ak

# tuple:
#   positional axis: (0, 1)
#   named axis: {"events": 0, "jets": 1}
array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis=("events", "jets"))

# dict:
#   positional axis: (0, 1)
#   named axis:  {"events": 0, "jets": 1}
array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis={"events": 0, "jets": 1})

# the dict interface allows to name single axis, also negative positional axis
array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis={"jets": -1})

# attach axis naming to an existing array
array = ak.Array([[1, 2], [3], [], [4, 5, 6]])
array = ak.with_named_axis(array, ("events", "jets"))
# or
array = ak.with_named_axis(array, {"events": 0, "jets": 1})
```

The `named_axis` argument of the constructor of an `ak.Array` is a either tuple of `AxisName` or a dict of `AxisName` to integers.
It is stored in the `.attrs` attribute of the array with a reserved key `"__named_axis__"` of type `dict[AxisName, int]`.
The two types of axis can be accessed through the `named_axis` and `positional_axis` property (always represented as a tuple):

```python
import awkward as ak

array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis=("events", "jets"))
array.named_axis
>>>  {"events": 0, "jets": 1}
array.positional_axis
>>> (0, 1)

array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis={"jets": -1})
array.named_axis
>>> {"jets": -1}
```

### Named axis in high-level functions

Named axis can be used by all high-level functions, e.g. `ak.sum`, `ak.max`, etc.:

```python
import awkward as ak

array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis=("events", "jets"))

# sum over the "jets" axis
sum_jets = ak.sum(array, axis="jets")
>>> ak.Array([3, 3, 0, 15])
sum_jets.named_axis
>>> {"events": 0}

# the `keepdims=True` argument keeps the named axis
sum_jets = ak.sum(array, axis="jets", keepdims=True)
>>> ak.Array([[3], [3], [], [15]])
sum_jets.named_axis
>>> {"events": 0, "jets": 1}
```

There are different scenarios how named axis are propagated to the resulting array:

1. Named axis are unchanged e.g. `ak.sum(array, axis="jets", keepdims=True)` or `array ** 2`.
2. Named axis are removed from the resulting array e.g. through reduction `ak.sum(array, axis="jets")`.
3. Named axis are unified e.g. in binary operations of two `ak.Array` or broadcasting:

  ```python
  import awkward as ak

  array1 = ak.Array([[1, 2], [3, 4]], named_axis=("In", None))
  array2 = ak.Array([[5, 6], [7, 8]], named_axis=(None, "Out"))

  (array1 + array2).named_axis
  >>> {"In": 0, "Out": 1}
  ```
  Here, checks for matching named axis are performed, the rules are:
  
  ```python
  ak.Array([1], named_axis=("foo",)) + ak.Array([1], named_axis=("foo",))    # OK
  ak.Array([1], named_axis=("foo",)) + ak.Array([1], named_axis=(None,))     # OK
  ak.Array([1], named_axis=("foo",)) + ak.Array([1], named_axis=("bar",))    # raise Exception
  ``` 

### Named axis in indexing

In addition, named axis can be used to select data:

```python
import awkward as ak

array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis=("events", "jets"))

# select the first event
array[{"events": 0, "jets": slice(None)}]
>>> <Array [1, 2] jets:0 type='2 * int64'>

# select the first jet of each event
array[{"events": slice(None), "jets": slice(0, 1)}]
>>> <Array [[1], [3], [], [4]] events:0,jets:1 type='4 * var * int64'>

# mixed positional & named indexing
array[{0: slice(None), "jets": slice(0, 1)}]
>>> <Array [[1], [3], [], [4]] events:0,jets:1 type='4 * var * int64'>
```

For synthatic sugar use `np.s_`:

```python
import awkward as ak

array = ak.Array([[1, 2], [3], [], [4, 5, 6]], named_axis=("events", "jets"))

# select the first jet of each event
array[{"events": np.s_[...], "jets": np.s_[0:1]}]
>>> <Array [[1], [3], [], [4]] events:0,jets:1 type='4 * var * int64'>

# or mixed with positional axis
array[{0: np.s_[...], "jets": np.s_[0:1]}]
>>> <Array [[1], [3], [], [4]] events:0,jets:1 type='4 * var * int64'>
```

---

This PR has to touch a lot of code and needs to add custom named axis propagation to each high-level operation. Thus, this PR is currently in draft mode.

Looking forward to ideas, thoughts, feedback on this effort!


pfackeldey:
# Progress

## general

- [x] documentation for named axis

## broadcasting

- [x] NumPy-style (right broadcasting)
- [x] Awkward-style (left broadcasting) 

## slicing

- [x] positional axis only (e.g. `array[0]`)
- [x] named axis only (e.g. `array[{"events": 0}]`)
- [x] mixed positional and named axis (e.g. `array[{0: 0, "jets": 0}]`)

## Unary and binary operations

- [x] unary operations (e.g. `array ** 2`)
- [x] binary operations (e.g. `array1 + array2`)


## high-level functions

New:
- [x] `ak.with_named_axis`
- [x] `ak.without_named_axis`

Can be used with named axis:
- [x] `ak.all`
- [x] `ak.almost_equal`
- [x] `ak.angle`
- [x] `ak.any`
- [x] `ak.argcartesian`
- [x] `ak.argcombinations`
- [x] `ak.argmax`
- [x] `ak.argmin`
- [x] `ak.argsort`
- [x] `ak.array_equal`
- [x] `ak.backend`
- [x] `ak.broadcast_arrays`
- [x] `ak.broadcast_fields`
- [x] `ak.cartesian`
- [x] `ak.categories`
- [x] `ak.combinations`
- [x] `ak.concatenate`
- [x] `ak.copy`
- [x] `ak.corr` (not yet supported)
- [x] `ak.count`
- [x] `ak.count_nonzero`
- [x] `ak.covar` (not yet supported)
- [x] `ak.drop_none`
- [x] `ak.enforce_type`
- [x] `ak.fill_none`
- [x] `ak.firsts`
- [x] `ak.flatten`
- [x] `ak.imag`
- [x] `ak.is_none`
- [x] `ak.local_index`
- [x] `ak.mask`
- [x] `ak.max`
- [x] `ak.mean`
- [x] `ak.min`
- [x] `ak.moment`
- [x] `ak.nan_to_none`
- [x] `ak.nan_to_num`
- [x] `ak.num`
- [x] `ak.ones_like`
- [x] `ak.pad_none`
- [x] `ak.prod`
- [x] `ak.ptp`
- [x] `ak.ravel`
- [x] `ak.real`
- [x] `ak.round`
- [x] `ak.run_lengths`
- [x] `ak.singletons`
- [x] `ak.softmax`
- [x] `ak.sort`
- [x] `ak.std`
- [x] `ak.strings_astype`
- [x] `ak.sum`
- [x] `ak.to_packed`
- [x] `ak.unflatten`
- [x] `ak.values_astype`
- [x] `ak.var`
- [x] `ak.where`
- [x] `ak.with_field`
- [x] `ak.with_name`
- [x] `ak.with_parameter`
- [x] `ak.without_parameters`
- [x] `ak.zeros_like`
- [x] `ak.zip`


## Independent of named axis: improvements / bugs found that are fixed by this PR aswell:
- various typos in doc strings
- Indexing could have multiple `...` in certain cases, this is prohibited in NumPy (https://github.com/scikit-hep/awkward/pull/3238/commits/a705981fa4b446cb5017b9e4f422acfdb48c1fa2)
- keepdims argument in `ak.corr` and `ak.covar` was wrong for the mean calculation (https://github.com/scikit-hep/awkward/pull/3238/commits/00669a3c3cc3680077f0f711721da05d6d33827c)
- avoid touching shape unnecessarily often when accessing `.purelist_depth`, `.minmax_depth`, and `.branch_depth` (typetracers) through `self.inner_shape` property of `Numpy{Meta|Array}` (https://github.com/scikit-hep/awkward/pull/3238/commits/1af4376075f6eeda9f2ebbbd3b97fb9a88b1518a)

jpivarski:
And all the data types that can be passed into square brackets with `__getitem__`.

agoose77:
@pfackeldey how is this PR going? Did you manage to solve all the blockers?

jpivarski:
I'll let @pfackeldey give the details, but named axes are looking very good and are nearly done. (They're in that point where the progress bar is a hair's width from 100% but seems to spend extra time there, due to the 80/20 rule.)

I just wanted to give a brief explanation as to why it's not included in the Awkward 2.6.9 release: we want Python 3.13.0 wheels to be ready for tomorrow.

Wait a minute—[it was released today](https://docs.python.org/3.13/whatsnew/3.13.html). I wasn't expecting that. All the more reason, then.

pfackeldey:
Hi @agoose77,
yes, it is actually (almost) ready for review.
It just fails for win32 tests that I'm sure is unrelated to named axes: I added a test that unveiled a problem in broadcasting for win32 environments, and I'm very sure this is going wrong also for arrays without named axes. I marked this test with `pytest.xfail` but the test still doesn't pass? I don't understand why...
I'll merge main and satisfy the pylint test, then it should be ready for review :+1:

pfackeldey:
The PR is ready for review @agoose77 and @jpivarski. If you know why the windows tests are failing, please let me know if you have an idea... (I thought `xfail` would mark them as expected to fail and thus the test would pass?)

I'm sorry that the PR got so huge, but the largest part are tests related to named axes and translating named axes to positional axes in each high-level function. 
I'd be especially happy for your review of named axes propagation in indexing and broadcasting operations as they are the most complicated ones.

agoose77:
@pfackeldey the `xfail` is failing because it's ... succeeding (`xpass`) :rofl:

jpivarski:
I'd like to see the rendered documentation, but the Deploy Branch Preview jobs are getting skipped. I don't know why: [they were running PRs relatively recently](https://github.com/scikit-hep/awkward/actions/runs/11150224326).

Meanwhile, I'm still going over the rest of the PR.

**Edit:** It's because this is a fork, not a branch. That's okay; the markdown looks good; we'll see the rendered version after merging.

docs/user-guide/how-to-array-properties-named-axis.md
```diff
@@ -0,0 +1,304 @@
...
+| `ak.prod(..., axis=int, keepdims=True)`             | `keep all`         |
+| `ak.ravel`                                          | `remove all`       |
+| `ak.singletons`                                     | `add one`          |
+| `ak.sum(..., axis=None)`                            | `remove all`       |
+| `ak.sum(..., axis=int, keepdims=False)`             | `remove one`       |
+| `ak.sum(..., axis=int, keepdims=True)`              | `keep all`         |
+| `ak.unflatten`                                      | `remove all`       |
+| `ak.where`                                          | `unify`, `add one` |
+| `ak.with_field`                                     | `unify`, `add one` |
+| `ak.zip`                                            | `unify`, `add one` |
```
jpivarski, line 304: This is the only place where this table appears, right? I guess this is a good place for it, and there should be only one—and there should be reminders to fix the table if changes are ever made in the code.

src/awkward/_namedaxis.py
```diff
@@ -0,0 +1,785 @@
...
+        'x:0, y:1, z:2'
+        >>> _prettify_named_axes({"x": 0, "y": 1, "$": 2})
+        'x:0, y:1, "$":2'
+    """
+
+    def _prettify(ax: AxisName) -> str:
+        repr_ax = str(ax)
+        if re.match("[A-Za-z_][A-Za-z_0-9]*", repr_ax):
+            return repr_ax
+        return json.dumps(repr_ax)
```
jpivarski, line 69: I should someday make the value field names also use double-quotes like this (`json.dumps`, rather than `repr`). Then values, named axes, and type strings would be consistent in their quoting.

src/awkward/highlevel.py
```diff
@@ -326,9 +338,20 @@ def __init__(
         if behavior is not None and not isinstance(behavior, Mapping):
             raise TypeError("behavior must be None or a mapping")
 
-        if attrs is not None and not isinstance(attrs, Mapping):
+        if attrs is not None and not isinstance(attrs, dict):
```
jpivarski: Wait—why? We like to be general in what we expect from users (though if given a generic Mapping, we'd convert it into a dict).

src/awkward/highlevel.py
```diff
@@ -2224,12 +2333,18 @@ def show(
...
             self.type.show(stream=tmp)
-            out = "type: " + tmp.getvalue() + valuestr
-        else:
-            out = valuestr
+            out += "type: " + tmp.getvalue()
+        if axes:
+            if self.named_axis:
+                out += f"axes: {_prettify_named_axes(self.named_axis)}\n"
+            else:
+                out += "axes: " + ", ".join(map(str, self.positional_axis)) + "\n"
```
jpivarski: I'd rather

```suggestion
        if axes and self.named_axis:   # by the way, this is using Truthiness
            out += f"axes: {_prettify_named_axes(self.named_axis)}\n"
```

so if there are no named axes, no space is taken to describe them. (The `self.positional_axis` is just "0, 1, 2, 3, ...", right?)

Also, is `axes=True` in Jupyter?

And another thing, let's make user-facing words consistent: the problem is that it can be hard to guess whether I should say "axis" or "axes". Maybe this argument can be `named_axis: bool`, which would make it more obvious why you wouldn't see any axes when `named_axis=True` and there aren't any _named_ axes. Bottom line: can we refer to these things as "`named_axis`" (singular) everywhere (in all _public_ arguments, methods, properties, ...)?

src/awkward/operations/ak_local_index.py
```diff
@@ -88,8 +93,35 @@ def local_index(array, axis=-1, *, highlevel=True, behavior=None, attrs=None):
...
+
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis):
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: This looks like more of an assertion. It's a good idea to only use exceptions like TypeError and ValueError for user (or dependent library) mistakes and AssertionError and RuntimeError for mistakes in Awkward itself.

jpivarski:
```suggestion
    assert is_integer(axis), f"'axis' must be an integer by now, not {axis!r}"
```

src/awkward/operations/ak_all.py
```diff
@@ -67,9 +73,29 @@ def all(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_any.py
```diff
@@ -67,9 +73,29 @@ def any(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_argmax.py
```diff
@@ -132,9 +138,29 @@ def nanargmax(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_argmin.py
```diff
@@ -129,10 +135,29 @@ def nanargmin(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_argsort.py
```diff
@@ -70,11 +74,25 @@ def argsort(
...
+
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis):
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_count.py
```diff
@@ -109,9 +115,29 @@ def count(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_count_nonzero.py
```diff
@@ -68,7 +74,29 @@ def count_nonzero(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_drop_none.py
```diff
@@ -65,10 +69,19 @@ def _drop_none_if_list(layout):
...
 
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_fill_none.py
```diff
@@ -84,6 +86,16 @@ def _impl(array, value, axis, highlevel, behavior, attrs):
...
 
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_firsts.py
```diff
@@ -58,10 +63,23 @@ def firsts(array, axis=1, *, highlevel=True, behavior=None, attrs=None):
...
+    #   use strategy "remove one" (see: awkward._namedaxis)
+    out_named_axis = _remove_named_axis(
+        named_axis=named_axis,
+        axis=axis,
+        total=layout.minmax_depth[1],
+    )
+
     if not is_integer(axis):
-        raise TypeError(f"'axis' must be an integer, not {axis!r}")
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_is_none.py
```diff
@@ -41,12 +46,22 @@ def is_none(array, axis=0, *, highlevel=True, behavior=None, attrs=None):
...
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
     if not is_integer(axis):
-        raise TypeError(f"'axis' must be an integer, not {axis!r}")
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_max.py
```diff
@@ -142,9 +148,29 @@ def nanmax(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_mean.py
```diff
@@ -191,6 +195,17 @@ def _impl(x, weight, axis, keepdims, mask_identity, highlevel, behavior, attrs):
...
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    if named_axis := _get_named_axis(ctx):
+        if _is_valid_named_axis(axis):
+            # Step 1: Normalize named axis to positional axis
+            axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_min.py
```diff
@@ -142,9 +148,29 @@ def nanmin(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_num.py
```diff
@@ -83,13 +96,28 @@ def num(array, axis=1, *, highlevel=True, behavior=None, attrs=None):
...
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+    # Step 2: propagate named axis from input to output,
+    #   use strategy "keep one" (see: awkward._namedaxis)
+    out_named_axis = _keep_named_axis(named_axis, axis)
+
     if not is_integer(axis):
-        raise TypeError(f"'axis' must be an integer, not {axis!r}")
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_pad_none.py
```diff
@@ -113,9 +117,18 @@ def pad_none(
...
         layout = ctx.unwrap(array, allow_record=False, primitive_policy="error")
+
+    axis = regularize_axis(axis)
+
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis):
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_prod.py
```diff
@@ -119,9 +125,29 @@ def nanprod(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_ptp.py
```diff
@@ -83,10 +88,19 @@ def ptp(
...
 
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_singletons.py
```diff
@@ -56,12 +61,24 @@ def singletons(array, axis=0, *, highlevel=True, behavior=None, attrs=None):
...
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
     if not is_integer(axis):
-        raise TypeError(f"'axis' must be an integer, not {axis!r}")
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_sort.py
```diff
@@ -59,11 +63,25 @@ def sort(
...
+
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis):
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_std.py
```diff
@@ -182,6 +185,16 @@ def _impl(x, weight, ddof, axis, keepdims, mask_identity, highlevel, behavior, a
...
 
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_sum.py
```diff
@@ -269,9 +275,29 @@ def nansum(
...
+    out_named_axis = _keep_named_axis(named_axis, None)
+    if not keepdims:
+        out_named_axis = _remove_named_axis(
+            named_axis=out_named_axis,
+            axis=axis,
+            total=layout.minmax_depth[1],
+        )
+
+    if not is_integer(axis) and axis is not None:
+        raise TypeError(f"'axis' must be an integer or None by now, not {axis!r}")
```
jpivarski: Same here.

src/awkward/operations/ak_unflatten.py
```diff
@@ -105,6 +107,16 @@ def _impl(array, counts, axis, highlevel, behavior, attrs):
...
 
+    axis = regularize_axis(axis)
+
+    # Handle named axis
+    named_axis = _get_named_axis(ctx)
+    # Step 1: Normalize named axis to positional axis
+    axis = _named_axis_to_positional_axis(named_axis, axis)
+
+    if not is_integer(axis):
+        raise TypeError(f"'axis' must be an integer by now, not {axis!r}")
```
jpivarski: Same here.

jpivarski: I didn't realize how many of those there would be!

jpivarski (approved):
It's hard to review such a large PR, but we went over it at length during development. I think we iterated much more quickly because of that, and so I don't have much to say here, on the PR itself. The spot-check issues below are all minor. Still, we should check in again before the final merger, since everybody working on the codebase needs to be ready to incorporate the update when it comes.

jpivarski:
Oh, and I should have mentioned that this is very high quality code (type annotations, docstrings, comments)! Thank you!

pfackeldey:
Thank you very much for your review @jpivarski! I'll add your suggestions soon :+1:

src/awkward/highlevel.py
```diff
@@ -357,7 +380,7 @@ def _update_class(self):
         self.__class__ = get_array_class(self._layout, self._behavior)
 
     @property
-    def attrs(self) -> Mapping[str, Any]:
+    def attrs(self) -> Mapping | AttrsNamedAxisMapping:
```
agoose77, line 383: What we want to be able to express here is a mapping that has a few known value types, and the rest all `Mapping[T, Any]`. But, Python doesn't let us express that, so I think we need to relax this type hint, and cast at the consumer side. I.e.,
```python
named_axis = tp.cast(AxisMapping, array.attrs)
```

I recognise this will probably change a lot of code!

agoose77: What we want to be able to express here is a mapping that has a few known value types, and the rest all `Mapping[T, Any]`. But, Python doesn't let us express that, so I think we need to relax this type hint, and cast at the consumer side. I.e.,
```python
named_axis = tp.cast(AxisMapping, array.attrs)
```

I recognise this will probably change a lot of code!

src/awkward/highlevel.py
```diff
@@ -1062,13 +1094,30 @@ def __getitem__(self, where):
...
             )
 
+            if NamedAxis.mapping:
+                return ak.operations.ak_with_named_axis._impl(
+                    out,
+                    named_axis=NamedAxis.mapping,
+                    highlevel=True,
+                    behavior=None,
+                    attrs=None,
+                )
```
agoose77: I haven't thought about this deeply, but I think it would be better to defer the first `wrap_layout` until the `else` branch of `if NamedAxis.mapping`. Then, `with_named_axis` doesn't need to unwrap the high-level array (and, thus, you'll need to pass in the `attrs` and `behavior` arguments to `with_named_axis`.

src/awkward/highlevel.py
```diff
@@ -1309,6 +1358,20 @@ def _repr(self, limit_cols):
...
+        # if it's too long we cut it off until there is enough space for in total 20 chars including ",..."
+        if len(",".join(axisstr)) > 20:
+            while len(",".join(axisstr)) > 20 - len(",..."):
+                axisstr.pop(-1)
+            axisstr.append("...")
+        axisstr = ",".join(axisstr)
+        if axisstr:
+            axisstr = f" {axisstr}"
+        # subtract the reserved space from the limit_cols
+        limit_cols -= len(axisstr)
```
agoose77: I'm mildly in favour of making this elision logic part of `_prettify_named_axes`, so that the implementation details of the string don't leak out to this part of the code.

src/awkward/highlevel.py
```diff
@@ -1365,12 +1429,18 @@ def show(
...
-            out = "type: " + tmp.getvalue() + valuestr
-        else:
-            out = valuestr
+            out += "type: " + tmp.getvalue()
+        if axes:
+            if self.named_axis:
+                out += f"axes: {_prettify_named_axes(self.named_axis)}\n"
+            else:
+                out += "axes: " + ", ".join(map(str, self.positional_axis)) + "\n"
+        out += valuestr
```
agoose77: Prefer lifting the `StringIO` object here to avoid allocating small strings:
```python
out_io = io.StringIO()
if type:
    out_io.write("type: ")
    self.type.show(stream=out_io)
if axes and self.named_axis:
    out_io.write("axes: ")
    out_io.write(_prettify_named_axes(self.named_axis))
    out_io.write("\n")
out_io.write(valuestr)    
```

It won't be a big performance change either way.

src/awkward/highlevel.py
```diff
@@ -1762,6 +1833,20 @@ def __init__(
         if behavior is not None and not isinstance(behavior, Mapping):
             raise TypeError("behavior must be None or mapping")
 
+        if attrs is not None and not isinstance(attrs, dict):
```
agoose77: Same as @jpivarski's point above :)

src/awkward/_namedaxis.py
```diff
@@ -0,0 +1,795 @@
...
+        {"x": 0, "y": 1, "z": 2}
+        >>> _get_named_axis({"other_key": "other_value"})
+        {}
+    """
+    if isinstance(ctx, MaybeSupportsNamedAxis):
+        return _get_named_axis(ctx.attrs)
+    elif isinstance(ctx, tp.Mapping) and NAMED_AXIS_KEY in ctx:
+        return dict(ctx[NAMED_AXIS_KEY])
+    else:
+        return {}
```
agoose77: I'm tempted to say that this function is doing too much; it can be simplified to 
```python
_get_named_axis(array.attrs)
```

It's not a big deal either way, by my instinct is that the high-level callers know whether they're passing in `ctx` or `ctx.attrs`, so they can avoid hitting this if branch altogether.

src/awkward/_namedaxis.py
```diff
@@ -0,0 +1,795 @@
...
+    Examples:
+        >>> _neg2pos_axis(-1, 3)
+        2
+        >>> _neg2pos_axis(1, 3)
+        1
+    """
+    if axis < 0:
+        return total + axis
+    return axis
+
```
agoose77: Is it worth connecting this with `maybe_posaxis`?

src/awkward/_namedaxis.py
```diff
@@ -0,0 +1,795 @@
...
+# These are the strategies to handle named axis for the
+# output array when performing operations along an axis.
+# See studies/named_axis.md#named-axis-in-high-level-functions and
+# https://pytorch.org/docs/stable/name_inference.html.
+#
+# The possible strategies are:
+# - "keep all" (_keep_named_axis(..., None)): Keep all named axes in the output array, e.g.: `ak.drop_none`
+# - "keep one" (_keep_named_axis(..., int)): Keep one named axes in the output array, e.g.: `ak.firsts`
+# - "keep up to" (_keep_named_axis_up_to(..., int)): Keep all named axes upto a certain positional axis in the output array, e.g.: `ak.local_index`
```
agoose77:
```suggestion
# - "keep up to" (_keep_named_axis_up_to(..., int)): Keep all named axes up to a certain positional axis in the output array, e.g.: `ak.local_index`
```

src/awkward/_namedaxis.py
```diff
@@ -0,0 +1,795 @@
...
+        AxisMapping: The new named axis after keeping the specified axis.
+
+    Examples:
+        >>> _keep_named_axis({"x": 0, "y": 1, "z": 2}, 1)
+        {"y": 0}
+        >>> _keep_named_axis({"x": 0, "y": 1, "z": 2}, None)
+        {"x": 0, "y": 1, "z": 2}
+    """
+    if axis is None:
+        return dict(named_axis)
```
agoose77: I'm sure that this copy was deliberate, but I wonder if copying is always needed when it's not required by this function? Is there a way of reducing the copies without making the caller code more complex?

src/awkward/_namedaxis.py
```diff
@@ -0,0 +1,795 @@
...
+                return pos + direction
+            # negative axis and position greater than the removed/added (negative) axis, but smaller than 0
+            # -> keep position
+            else:
+                return pos
+
+    out = dict(named_axis)
+    for k, v in out.items():
+        out[k] = _adjust(v, axis, direction)
+    return out
```
agoose77:
```suggestion
    return {k: _adjust(v, axis, direction) for k, v in named_axis.items()}
```

agoose77 (commented):
A small review -- I'll pop back tomorrow.

src/awkward/highlevel.py
pfackeldey: `typing.Mapping` does not support `__setitem__` which I need in order to set named axis. I think the correct type annotation would then be `typing.MutableMapping` instead of `dict`?

src/awkward/operations/ak_local_index.py
pfackeldey: These checks exist like this already in awkward-array/main: https://github.com/scikit-hep/awkward/blob/main/src/awkward/operations/ak_is_none.py#L48-L49
I just added them consistently to every high-level operation.

In principle this is the error a user sees when they pass in an axis that is neither a valid `named_axis` or an `int` (or sometimes `None`) - so it should be a user mistake.

I thought of it as 3 possibilities:
1. it is a valid `named_axis` (a `str`): translate it to `int` if possible else raise a meaningful error for the user
2. it is `int` (or sometimes `None`): just pass through
3. it is something else, e.g. a `dict` (why would someone ever do this, but it is not prohibited): raise the `TypeError` as above

src/awkward/highlevel.py
pfackeldey: I'll try this out - less unwrapping + wrapping makes sense. This was a little fiddly part in the code where I wanted to make sure that the return type stay the same compared to awkward-array/main. But I'm sure your suggestion would not alter the return type, so great!

src/awkward/highlevel.py
pfackeldey: Where do I find the Jupyter setting?

src/awkward/_namedaxis.py
pfackeldey: This is a bit different to `maybe_posaxis` because it does not account for branching. Hm, but we could use `_neg2pos_axis` within `maybe_posaxis`... is this what you're suggesting?

src/awkward/highlevel.py
pfackeldey: Sure, I commented under Jims comment already. I resolve this, and we can discuss it in the other discussion thread.

src/awkward/highlevel.py
pfackeldey: Should we maybe in general type hint `attrs` with:
```python
import typing as tp

class AttrsMapping(tp.MutableMapping):
    @tp.overload
    def __getitem__(self, key: NAMED_AXIS_KEY) -> AxisMapping: ...
```
?
That way we just have `MutableMapping` for `.attrs` that knows about the type of `AxisMapping` is we extract it from `.attrs`. 
But maybe we're over-engineering here and just use `MutableMapping` always for `.attrs`...

src/awkward/_namedaxis.py
pfackeldey: Yes, but I wanted it to also work with objects (e.g. `HighLevelContexts` or `np.arrays`) that don't necessarily have a `.attrs` attribute without me worrying about it. In principle this is similar to `attrs_of_obj`, but this one does not work with `HighLevelContext` (and I couldn't make it work properly with it).

src/awkward/highlevel.py
pfackeldey: I switched it to `tp.MutableMapping` :+1:

PR #3242, opened 2024-09-13
test: add simple slice test to cuda

martindurant:
Shows failure in the function I added and probably all following ones.


martindurant:
cc @Ianna

PR #3254, opened 2024-09-23
perf: avoid inflating UnmaskedArrays in broadcasting when you can

jpivarski:
@martindurant

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -701,6 +701,36 @@ def apply_step(
                 for x, p in zip(outcontent, parameters)
             )
 
+    def broadcast_any_option_all_UnmaskedArray():
+        nextinputs = []
+        nextparameters = []
+        for x in inputs:
+            if isinstance(x, UnmaskedArray):
+                nextinputs.append(x.content)
+                nextparameters.append(x._parameters)
+            elif isinstance(x, Content):
+                nextinputs.append(x)
+                nextparameters.append(x._parameters)
+            else:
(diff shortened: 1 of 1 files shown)


martindurant:
Thanks, @jpivarski .

Have you considered implementing something like asv for ak, to make sure that new changes don't accidentally hurt performance?

jpivarski:
Airspeed Velocity? Maybe, although we'd have to define some large samples and a suite of computations to test. We have a huge number of tests, since they were added as features were added and bugs were fixed, but adding a suite of performance tests would be starting from scratch because _all_ of the functionality tests are way too small to be meaningful for performance.

This particular PR, however, won't make anything slower. There's a special case that will get faster. The only cost here is the added complexity for maintenance. I thought this one was worth the cost because it cleanly splits out as a special case and it wouldn't be a very rare special case: conversions from Arrow are likely to introduce UnmaskedArrays (because Arrow nodes are option-type by default, but often don't actually have missing values).

martindurant:
You mean when mask is None? Makes sense. 

About asv, agree that it would be quite an undertaking, maybe a good project for a student, for a select few workflows.

pfackeldey (approved):
Looks good to me.

If anything you may want to add a dedicated test with a comment to this commit. That may make future maintenance a bit easier.

jpivarski:
I couldn't think of a test because it doesn't affect functionality, just performance. And the array I was using to test performance is too big to save in scikit-hep-testdata.

Thanks for checking! I think I'll merge this now.

PR #3259, opened 2024-09-27
feat: to/from PyTorch Tensor

maxymnaumchyk:


src/awkward/operations/ak_from_torch.py
```diff
@@ -0,0 +1,64 @@
...
+    device = "cuda" if array.is_cuda else "cpu"
+
+    # convert tensors to cupy if they are on cuda
+    if device == "cuda":
+        try:
+            from awkward._nplikes.cupy import Cupy
+
+            cp = Cupy.instance()
+        except (ModuleNotFoundError, ImportError) as err:
+            raise err
```
jpivarski: Catching an exception and then immediately/always re-raising it is the same as not catching it:

```suggestion
        from awkward._nplikes.cupy import Cupy

        cp = Cupy.instance()
```

src/awkward/operations/ak_to_torch.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def _impl(array):
+    try:
+        import torch
+    except ImportError as err:
+        raise ImportError(
+            """to use ak.to_torch, you must install 'torch' package with:
+
+        pip install torch or conda install pytorch"""
```
jpivarski: This makes it possible to copy-paste the line from the error message onto the terminal. (On most terminals, triple-click selects the whole line, and leading spaces are irrelevant to shell commands.)

```suggestion
        pip install torch

or

        conda install pytorch"""
```

src/awkward/operations/ak_to_torch.py
```diff
@@ -0,0 +1,67 @@
...
+            "Only arrays containing regular-length lists (# *) of numbers can be converted into a PyTorch Tensor"
+        ) from err
+
+    # check if cupy or numpy
+    if isinstance(np_array, np.ndarray):
+        # convert numpy to a torch tensor
+        tensor = torch.from_numpy(np_array).to(device)
+    else:
+        # cupy -> torch tensor
+        tensor = torch.as_tensor(np_array, device=device)
```
jpivarski: The above involves a copy (I tried it). This doesn't:

```suggestion
        tensor = torch.utils.dlpack.from_dlpack(np_array.toDlpack())
```

src/awkward/operations/ak_to_torch.py
```diff
@@ -0,0 +1,67 @@
...
+        np_array = array.to_backend_array(allow_missing=False)
+    except ValueError as err:
+        raise TypeError(
+            "Only arrays containing regular-length lists (# *) of numbers can be converted into a PyTorch Tensor"
+        ) from err
+
+    # check if cupy or numpy
+    if isinstance(np_array, np.ndarray):
+        # convert numpy to a torch tensor
+        tensor = torch.from_numpy(np_array).to(device)
```
jpivarski: If it's already a NumPy array, it's already on the `"cpu"` device.

```suggestion
        tensor = torch.from_numpy(np_array)
```

src/awkward/operations/ak_to_torch.py
```diff
@@ -0,0 +1,67 @@
...
+
+    # get the device array is on
+    device = ak.backend(array)
+
+    # convert to numpy or cupy if `array` on gpu
+    try:
+        np_array = array.to_backend_array(allow_missing=False)
+    except ValueError as err:
+        raise TypeError(
+            "Only arrays containing regular-length lists (# *) of numbers can be converted into a PyTorch Tensor"
```
jpivarski: The `to_backend_array` accepts arrays whose type allows for variable-length lists if those lists actually have equal-length. Here's a clarification of the error message:

```suggestion
            "Only arrays containing equal-length lists of numbers can be converted into a PyTorch Tensor"
```

src/awkward/operations/ak_to_torch.py
```diff
@@ -0,0 +1,67 @@
...
+
+        pip install torch or conda install pytorch"""
+        ) from err
+
+    # useful function that handles all possible input arrays
+    array = ak.to_layout(array, allow_record=False)
+
+    # get the device array is on
+    device = ak.backend(array)
+
```
jpivarski, line 54: With the changes below, you don't need `device` anymore.

```suggestion
```

src/awkward/operations/ak_to_torch.py
```diff
@@ -0,0 +1,67 @@
...
+    # convert to numpy or cupy if `array` on gpu
+    try:
+        np_array = array.to_backend_array(allow_missing=False)
```
jpivarski: `np_array` has a misleading name (but I can't "suggest change" for that because it affects multiple lines).

Also, you might not have known about it, but there are more backend types than `"cpu"` and `"cuda"`. I think it would be reasonable for this function to raise an error if the `array` is not one of these two. (So maybe there is a reason to get the `device`, after all!) The other _currently existing_ backends are `"typetracer"` for Dask-delayed data and `"jax"` for autodiff (which might be replaced). `ak.to_torch` is an inherently eager function, so it can't be used with Dask and raising an error would be a reasonable thing to do.

jpivarski (changes requested):
Looks good! I just have a few suggestions.

src/awkward/operations/ak_to_torch.py
maxymnaumchyk: Thanks for catching this! I should've tested it more carefully.

jpivarski (approved):

ianna (approved):
Looks good to me!

PR #3261, opened 2024-09-30
docs: pybind11 demo project should have NumPy own the data

jpivarski:
The original demo _does_ link the memory-release to when the `py::capsule` does out of scope in Python, but [there are issues](https://indico.cern.ch/event/855454/contributions/4605044/) with crossing ownership of arrays between Python and C++.

The (nontrivial) back-and-forth API for LayoutBuilder snapshots was explicitly intended to allow Python to fully own the data, so I'm updating the example to make it do that.

@HavryliukAY, I'm approaching this as I would always approach a problem like this, doing it in small steps that can be individually tested. I don't ordinarily make a git-commit for each step, but I'll be doing that in this PR to show what this looks like. Between each commit, I do

```bash
pip uninstall demo
pip install .
python -c 'import demo; print(repr(demo.create_demo_array()))'
```

to see what I'm doing. In the first commit, "step 0", it returns `None`.

### DIFF
--- a/header-only/examples/pybind11/demo.cpp
+++ b/header-only/examples/pybind11/demo.cpp
@@ -51,9 +51,16 @@ py::object snapshot_builder(const T &builder) {
     for (auto name_nbytes : names_nbytes) {
       int nbytes = name_nbytes.second;
 
-      py::object array = np.attr("empty")(nbytes, dtype_u1);
+      py::object array = np.attr("ones")(nbytes, dtype_u1);
 
-      py::module::import("builtins").attr("print")(array);
+      size_t pointer = py::cast<size_t>(array.attr("ctypes").attr("data"));
+      void* raw_data = (void*)pointer;
+
+      int c0 = ((unsigned char*)raw_data)[0];
+      int c1 = ((unsigned char*)raw_data)[1];
+      int c2 = ((unsigned char*)raw_data)[2];
+
+      std::cout << "pointer = " << pointer << " raw data = " << c0 << " " << c1 << " " << c2 << std::endl;
 
     }
 
--- a/header-only/examples/pybind11/demo.cpp
+++ b/header-only/examples/pybind11/demo.cpp
@@ -62,32 +62,8 @@ py::object snapshot_builder(const T &builder) {
     // Write non-contiguous contents to memory.
     builder.to_buffers(cpp_container);
 
-    py::module::import("builtins").attr("print")(py_container);
-
-    // ak.attr("from_buffers")(py_container);
-
-    // // Build Python dictionary containing arrays
-    // // dtypes not important here as long as they match the underlying buffer
-    // // as Awkward Array calls `frombuffer` to convert to the correct type
-    // py::dict container;
-    // for (auto it: buffers) {
-
-    //     py::capsule free_when_done(it.second, [](void *data) {
-    //         uint8_t *dataPtr = reinterpret_cast<uint8_t *>(data);
-    //         delete[] dataPtr;
-    //     });
-
-    //     uint8_t *data = reinterpret_cast<uint8_t *>(it.second);
-    //     container[py::str(it.first)] = py::array_t<uint8_t>(
-    //             {names_nbytes[it.first]},
-    //             {sizeof(uint8_t)},
-    //             data,
-    //             free_when_done
-    //     );
-    // }
-    // return from_buffers(builder.form(), builder.length(), container);
-
-  return py::none();
+    // Build Python dictionary containing arrays.
+    return ak.attr("from_buffers")(builder.form(), builder.length(), py_container);
 }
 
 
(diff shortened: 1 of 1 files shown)


jpivarski:
I did a quick test with

```c++
auto np = py::module::import("not_a_real_module_name");
```

_without_ checking for errors, and pybind11 was nice enough to raise a ModuleNotFoundError for me—no seg-faults! So, based on this, I can `py::module::import` NumPy and Awkward Array without checking to see if they exist, since the standard error-handling is all I need.

jpivarski:
```python
nbytes = 24
nbytes = 32
nbytes = 24
None
```

jpivarski:
```python
[177  65 227   8  71  87   0   0   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0]
[248 248 148   8  71  87   0   0   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0   0   0   0   0   0   0   0   0]
[177  65 227   8  71  87   0   0   0   0   0   0   0   0   0   0   0   0
   0   0   0   0   0   0]
```

(The exact values in the allocated arrays will depend on what random memory is located at that address.)

jpivarski:
```python
pointer = 94383525626576 raw data = 1 1 1
pointer = 94383523939584 raw data = 1 1 1
pointer = 94383525626576 raw data = 1 1 1
```

The pointer positions are run-dependent, but since I temporarily changed `np.empty` to `np.ones`, we should expect all of the bytes to be equal to `1`. This test didn't seg-fault because I saw in a previous step that all of these buffers would be allocated with at least 3 bytes...

jpivarski:
```python
{'node1-data': array([ 86, 159, 205, 171, 248,  90,   0,   0,   0,   0,   0,   0,   0,
         0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0], dtype=uint8), 'node2-offsets': array([ 94, 194, 220, 171, 248,  90,   0,   0,   0,   0,   0,   0,   0,
         0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,   0,
         0,   0,   0,   0,   0,   0], dtype=uint8), 'node3-data': array([221,  65, 208, 175,   5,   0,   0,   0,   0,   0,   0,   0,   0,
         0,   0,   0,  32,   0,   0,   0,   0,   0,   0,   0], dtype=uint8)}
```

jpivarski:
```python
{'node1-data': array([154, 153, 153, 153, 153, 153, 241,  63, 154, 153, 153, 153, 153,
       153,   1,  64, 102, 102, 102, 102, 102, 102,  10,  64], dtype=uint8), 'node2-offsets': array([0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0,
       0, 0, 6, 0, 0, 0, 0, 0, 0, 0], dtype=uint8), 'node3-data': array([1, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 3, 0,
       0, 0], dtype=uint8)}
```

These values are _not_ random data:

```python
>>> from numpy import array, uint8
>>> container = {'node1-data': array([154, 153, 153, 153, 153, 153, 241,  63, 154, 153, 153, 153, 153,
...        153,   1,  64, 102, 102, 102, 102, 102, 102,  10,  64], dtype=uint8), 'node2-offsets': array([0, 0, 0, 0, 0, 0, 0, 0, 1, 0, 0, 0, 0, 0, 0, 0, 3, 0, 0, 0, 0, 0,
...        0, 0, 6, 0, 0, 0, 0, 0, 0, 0], dtype=uint8), 'node3-data': array([1, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 1, 0, 0, 0, 2, 0, 0, 0, 3, 0,
...        0, 0], dtype=uint8)}
>>> container["node1-data"].view("f8")
array([1.1, 2.2, 3.3])
>>> container["node2-offsets"].view("i8")
array([0, 1, 3, 6])
>>> container["node3-data"].view("i4")
array([1, 1, 2, 1, 2, 3], dtype=int32)
```

It's the data we want!

jpivarski:
The above commit is the last step, which builds the array from `ak.from_buffers` and returns it, instead of returning `None`.

```bash
python -c 'import demo; demo.create_demo_array().show(type=True)'
```

prints

```python
type: 3 * {
    one: float64,
    two: var * int32
}
[{one: 1.1, two: [1]},
 {one: 2.2, two: [1, 2]},
 {one: 3.3, two: [1, 2, 3]}]
```

jpivarski:
TODO: something similar should be done for the Cython example.

https://github.com/scikit-hep/awkward/blob/e946646c6315ff9bde0c8a2255ee6f7fc6a31539/header-only/examples/cython/demo_impl.cpp#L40-L44

It should be even easier in Cython, since you can call Python directly.

PR #3283, opened 2024-10-11
chore: add download stats

ianna:
I thought, it would be nice to show it. See https://www.pepy.tech/projects/awkward

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@
 [![Documentation](https://img.shields.io/badge/docs-online-success)](https://awkward-array.org/)
 [![Gitter](https://img.shields.io/badge/chat-online-success)](https://gitter.im/Scikit-HEP/awkward-array)
 
+[![Downloads](https://static.pepy.tech/badge/awkward)](https://pepy.tech/project/awkward)
+[![Downloads](https://static.pepy.tech/badge/awkward/month)](https://pepy.tech/project/awkward)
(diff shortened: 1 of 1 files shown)


jpivarski:
Yeah, I don't know if we should show those numbers because they're significantly inflated by distributed jobs that run `pip install` as the first line. You can see this by separating MacOS and Windows downloads from Linux, the only handle that the [PyPI dataset](https://cloud.google.com/blog/topics/developers-practitioners/analyzing-python-package-downloads-bigquery) provides for this. (It's much more common for distributed jobs to run on Linux.)

[image](https://github.com/user-attachments/files/17344526/pip-installs-2024.pdf)

The Linux downloads are more than 10× MacOS and Windows; they need to be down-scaled with a separate (right) axis to show them both on the same plot. But also, the Linux downloads are more volatile, since they depend on when very large jobs run, rather than smoother patterns like when people decide to try out the software and how frequently they update their versions. The downloads that are tracked by [the site you pointed to](https://www.pepy.tech/projects/awkward) are very spiky, and I also don't know how to reconcile its much larger numbers with what I found in Google BigQuery.

I think download statistics are only interesting when they're measuring human choices—that is, when there's a one-to-one relationship between a `pip install` and a happy customer, rather than "I think I'll run with 100k jobs today, rather than 10k" (which is also a human choice, but not what I mean). So I'd rather not do this. We might find ourselves in a position of having to defend these numbers, and I don't think I can do that.

PR #3288, opened 2024-10-31
fix: test `from_raggedtensor` on GPU

ianna:
make sure the backend of the reference array is the same as the result's


jpivarski (approved):
Yes, this is an important fix! I'll merge it.

In this git-branch, `pytest tests-cuda` still runs. :heavy_check_mark:

PR #3292, opened 2024-11-06
feat: to/from TensorFlow Tensor

### LINKED ISSUE #3267: Add interoperability between Awkward Array and ML libraries
This issue is a checklist of different functions that convert Awkward Arrays to ML libraries and backwards.

### TensorFlow library

Functions for rectangular arrays
- [x] ak.from_tensorflow
- [x] ak.to_tensorflow
#3292

Functions for irregular arrays
- [x] ak.from_raggedtensor
- [x] ak.to_raggedtensor
#3210
- [x] add cuda support for `ak.to_raggedtensor` and `ak.from_raggedtensor`
#3263

### PyTorch library

Functions for rectangular arrays
- [x] ak.from_torch
- [x] ak.to_torch
#3259

For irregular arrays, a number of functions will be created, for the most flexible and convenient use. Here are the issues opened for this task:
* #3256
* #3257
* #3265
And a useful comment by Jim: https://github.com/scikit-hep/awkward/pull/3246#pullrequestreview-2328831856

maxymnaumchyk:


maxymnaumchyk:
Note: as discussed [here](https://github.com/tensorflow/tensorflow/issues/33254#issuecomment-542379165) when converting a TensorFlow Tensor into Numpy array with `tensor.numpy()` it creates a copy since Numpy is mutable and TensorFlow Tensor is not.

When converting Awkward Array to TensorFlow (_while on CPU_), this function: `tf.convert_to_tensor(numpy_array)` also makes a copy. I suppose for the same reason.

jpivarski (approved):
This looks perfect! Thank you again! This looks complete, and if you agree, I'll merge it into `main`.

On copying data from TensorFlow to NumPy: it might not be_technically_ necessary, since NumPy arrays have flags and can be marked as not writable. Thus, a NumPy array could zero-copy view a TensorFlow tensor in RAM (not GPU memory), _if_ TensorFlow has stable pointers. Some systems, such as Java, copy data to different locations to defragment their memory use, which causes simple pointers (static integers) to get out of date. If TensorFlow does that, then it has to copy when converting to NumPy, even with NumPy's "writable" flag.

But this is probably out of our hands, since we have to rely on what functions TensorFlow gives us. They might have stable pointers that could be viewed as a non-writable NumPy array and yet just don't give us a function to do that—all we could do is ask them to add such a thing to their library (which we won't, because it's not such a big deal).

jpivarski:
(If you agree and it's ready to be merged, write a comment, rather than an emoji, since I get emails for comments but not emojis. Thanks!)

jpivarski:
It looks done to me. I'm going to merge it. Thanks!

PR #3297, opened 2024-11-08
build: Update awkward-cpp required Python to v3.9+

matthewfeickert:
* Update the minimum required Python to v3.9 for awkward-cpp, matching awkward v2.7.0.
    - c.f. https://github.com/scikit-hep/awkward/releases/tag/v2.7.0

### DIFF
--- a/tests/test_3210_to_raggedtensor_from_raggedtensor.py
+++ b/tests/test_3210_to_raggedtensor_from_raggedtensor.py
@@ -111,6 +111,10 @@ def test_convert_from_raggedtensor():
         result1.content.data == ak.to_backend(np_array1, result1.backend).layout.data
     ).all()
     assert (result1.offsets.data == [0, 2, 3, 3, 5]).all()
+    assert (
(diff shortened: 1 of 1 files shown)


matthewfeickert:
This is ready for review.

matthewfeickert:
Note to self: Though this is broken on PyPI (`awkward-cpp` `v41` being Python 3.8+ instead of Python 3.9+) this shouldn't matter in practice ever as nothing should depend on `awkward-cpp` except for `awkward` which pins versions of `awkward-cpp`. So there should never be an issue of a misinstall where a Python 3.8 installs `awkward-cpp` `v41` and breaks.

matthewfeickert:
This got fixed in e0f5febbfe23df85f33ae5133dbb4ec3853f3892 which got merged into `main`

https://github.com/scikit-hep/awkward/blob/6c47e1d665f026eb95777e6e09af760506eb4165/awkward-cpp/pyproject.toml#L47

so on rebased this closed itself.

PR #3298, opened 2024-11-08
fix: test from_raggedtensor on CUDA

ianna:
Use a consistent back-end for comparison.


ianna (commented):
I don't know how I missed it :-)

jpivarski (approved):
This looks like a good test and I will merge it. Thanks!

PR #3301, opened 2024-11-11
fix: ak.argcombinations should allow negative axis

### LINKED ISSUE #3300: Argcombinations not taking negative axis values
### Version of Awkward Array

2.7.0

ak.argcombinations is not accepting negative axis values. It seems to be an issue of inconsistent documentation, as the [api reference](https://awkward-array.org/doc/main/reference/generated/ak.argcombinations.html#ak.argcombinations) says negative axis values should be accepted but this value error is coded into the function class.
`    if axis < 0:
        raise ValueError("the 'axis' for argcombinations must be non-negative")
    else:`
  Is there a reason for combinations to accept negative axis numbers while argcombinations does not?

jpivarski:


pfackeldey:
@jpivarski looks good to me! 
I wonder: do you know why this was forbidden in the first place?

jpivarski:
I think it's just always been an oversight.

* `ak.argcombinations` and `ak.combinations` were ported over from the Awkward 1 C++ in #1307, and `ak.argcombinations` has the `axis >= 0` restriction there, while `ak.combinations` doesn't. `ak.argcombinations` was written in terms of `ak.combinations` and `ak.localindex`, as it is now.
* `ak.argchoose` and `ak.choose` (the old names) were orinally written in Awkward 1 in #160, and `ak.argchoose` has the `axis >= 0` restriction there, while `ak.choose` doesn't. `ak.argchoose` was written in terms of `ak.combinations` and `localindex`, as it is now.
* `argchoose` and `choose` didn't have an `axis` argument in [Awkward 0](https://github.com/scikit-hep/awkward-0.x/blob/dd885bef15814f588b58944d2505296df4aaae0e/awkward0/array/jagged.py#L1128-L1242). (`choose` was written in terms of `argchoose`, by applying the `argchoose` result as an integer slice, so that's different.)

When I was adding `axis` arguments to everything in the Awkward 0 → 1 transition, I must have thought, "There's no way this is well-defined (or implementable) if it has an `axis` argument" for `argchoose` but then had a different thought in `choose`. I probably never looked at them side by side like this. It is _not_ the case that they originally both had the restriction and it was lifted from `choose`. So, it's not a maintenance inconsistency; it's a development inconsistency.

PR #3299, opened 2024-11-10
fix: use None when ndims can't be inferred from a layout-like obj

pfackeldey:
Fixes: https://github.com/scikit-hep/vector/issues/524

PR #3304, opened 2024-11-13
fix: restrict `named_axis` inferring to `ak.Arrays/ak.Records/ak.HighLevelContexts` by default

pfackeldey:
This PR is a follow-up to PR https://github.com/scikit-hep/awkward/pull/3299. The `named_axis` mapping is now only inferred from awkward-constructs (`ak.Array`, `ak.Record`, and `HighLevelContext`) by default. This avoids likely unwanted clashes with objects that implements the `named_axis` mapping interface (for whatever reason) but are not meant to be used this way. I expect this to happen in very rare corner-cases. This PR covers them now. An example is given below:

Old behavior:
```python
from awkward._namedaxis import _get_named_axis, NAMED_AXIS_KEY


class some:
  @property
  def attrs(self):
    return {NAMED_AXIS_KEY: {"x": 0}}
    
_get_named_axis(some())
>> {"x": 0}
```

New behavior:
```python
from awkward._namedaxis import _get_named_axis, NAMED_AXIS_KEY


class some:
  @property
  def attrs(self):
    return {NAMED_AXIS_KEY: {"x": 0}}
    
_get_named_axis(some())
>> {}
 
# old behavior can be enforced by:
_get_named_axis(some(), allow_any=True)
>> {"x": 0} 
```
 
If we want to infer `named_axis` from any types again somewhere in the future, the `allow_any` argument will allow us to do so.

### DIFF
--- a/src/awkward/_broadcasting.py
+++ b/src/awkward/_broadcasting.py
@@ -463,7 +463,7 @@ def apply_step(
                         named_axis = _add_named_axis(named_axis, depth, ndim)
                         depth_context[NAMED_AXIS_KEY][i] = (
                             _unify_named_axis(named_axis, seen_named_axis),
-                            ndim + 1,
+                            ndim + 1 if ndim is not None else ndim,
                         )
                         if o.is_leaf:
                             _export_named_axis_from_depth_to_lateral(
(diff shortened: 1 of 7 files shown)


jpivarski (approved):
Great! (We talked about this elsewhere, maybe Slack.) Was there an issue opened about this by @veprbl? Does this PR close that issue?

pfackeldey:
No, there's no open issue. Restricting this behavior came up in a discussion between us regarding the vector issue https://github.com/scikit-hep/vector/issues/524, but it is not _directly_ related to it. This issue has been closed by https://github.com/scikit-hep/awkward/pull/3299.

jpivarski:
Oh, I see. Well, you can merge this PR when you're ready (probably now).

PR #3314, opened 2024-11-22
fix: sorting of record arrays without fields (`ak.zip(tuple(...))`)

### LINKED ISSUE #1361: `ak.sort` and `ak.argsort` should refuse to sort RecordArrays (raise a TypeError)
### Version of Awkward Array

2e8ce817

```python
import awkward as ak
import numpy as np
from awkward._v2.tmp_for_testing import v1_to_v2, v2_to_v1

# Check non branching cases
a = ak.layout.NumpyArray(
    np.array([1, 50, 100, 200, 500, 1000])
)

b = ak.layout.ListOffsetArray64(
    ak.layout.Index64(np.array([0, 2, 4, 6])),
    a
)

assert (ak.to_list(a.sort(-1, False, False)) == ak.to_list(
    v2_to_v1(v1_to_v2(a).sort(-1, False, False))))

assert (ak.to_list(b.sort(-1, False, False)) == ak.to_list(
    v2_to_v1(v1_to_v2(b).sort(-1, False, False))))

# Check branching cases
this = ak.layout.RecordArray(
    [a, a[::-1]],
    None
)
assert ak.to_list(this.sort(-1, False, False)) == ak.to_list(
    v2_to_v1(v1_to_v2(this).sort(-1, False, False)))

that = ak.layout.RecordArray(
    [a[:3], b],
    None
)
assert ak.to_list(that.sort(-1, False, False)) == ak.to_list(
    v2_to_v1(v1_to_v2(that).sort(-1, False, False)))

```

pfackeldey:
Fixes https://github.com/scikit-hep/awkward/issues/1361#issuecomment-1901389735.

It does not affect `ak.argsort`, because that does not even have an implementation for record arrays in the first place - it will raise a `NotImplementedError` as for now.

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -829,7 +829,7 @@ class RecordArray(RecordMeta[Content], Content):
         raise NotImplementedError
 
     def _sort_next(self, negaxis, starts, parents, outlength, ascending, stable):
-        if self._fields is None or len(self._fields) == 0:
+        if len(self.fields) == 0:
             return ak.contents.NumpyArray(
                 self._backend.nplike.instance().empty(0, dtype=np.int64),
                 parameters=None,
--- a/tests/test_1316_sort_recordarrays.py
+++ b/tests/test_1316_sort_recordarrays.py
@@ -1,0 +1,23 @@
+from __future__ import annotations
+
+import numpy as np
+
+import awkward as ak
+
+
+def test_sort_record():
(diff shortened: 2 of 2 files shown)


jpivarski (approved):
I didn't realize that we support sorting on RecordArrays at all! At one point in the past, we decided that only numerical data should be sortable, but I just tested it (in records with named fields) and it does work in a reasonable way: lexicographically. Supporting the tuple case is therefore a logical completion.

Feel free to merge!

PR #3316, opened 2024-11-22
fix: concatenation of empty records

### LINKED ISSUE #3258: Can't concatenate arrays of records with no fields
### Version of Awkward Array

HEAD

This line:

https://github.com/scikit-hep/awkward/blob/e946646c6315ff9bde0c8a2255ee6f7fc6a31539/src/awkward/_broadcasting.py#L471

doesn't handle the possibility that the RecordArray has no fields (`len(contents) == 0`).

Very, very few functions actually broadcast through fields; the ufuncs raise an error if they haven't been overloaded (and if they have been overloaded, the use the overload instead), `ak.broadcast_arrays` stops at records by a deliberate decision:

https://github.com/scikit-hep/awkward/blob/e946646c6315ff9bde0c8a2255ee6f7fc6a31539/src/awkward/operations/ak_broadcast_arrays.py#L240

In our test suite, the only examples that reach the record-broadcasting code (`broadcast_any_record`, past the part that errors-out when `options["allow_records"]` is false) are `ak.transform` and `ak.concatenate`.

I tried to trigger this error and got a different one:

```python
ak.concatenate([ak.Array([{}, {}, {}]), ak.Array([{}, {}])], axis=0)
```

```
  File "/home/jpivarski/irishep/awkward/src/awkward/_dispatch.py", line 64, in dispatch
    next(gen_or_result)
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_concatenate.py", line 64, in concatenate
    return _impl(arrays, axis, mergebool, highlevel, behavior, attrs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_concatenate.py", line 160, in _impl
    contents = [ak._do.mergemany(b) for b in batches]
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/operations/ak_concatenate.py", line 160, in <listcomp>
    contents = [ak._do.mergemany(b) for b in batches]
                ^^^^^^^^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/_do.py", line 218, in mergemany
    return contents[0]._mergemany(contents[1:])
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/recordarray.py", line 760, in _mergemany
    next = RecordArray(
           ^^^^^^^^^^^^
  File "/home/jpivarski/irishep/awkward/src/awkward/contents/recordarray.py", line 162, in __init__
    raise TypeError(
TypeError: RecordArray 'length' must be a non-negative integer or None, not awkward._util.UNSET

    ak.concatenate(
        [<Array [{}, {}, {}] type='3 * {}'>, <Array [{}, {}] type='2 * {}'>]
        axis = 0
    )
```

It's not a high priority because it's such a weird case, but this really should result in

```python
ak.Array([{}, {}, {}, {}, {}], type='5 * {}'>
```

pfackeldey:
Fixes https://github.com/scikit-hep/awkward/issues/3258.

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -752,10 +752,13 @@ class RecordArray(RecordMeta[Content], Content):
             ):
                 minlength = merged.length
 
-        if minlength is unknown_length:
+        # `not for_each_field`: is a corner-case when all
+        # the arrays are empty and have no fields either.
+        if minlength is unknown_length or not for_each_field:
+            from operator import attrgetter
+
             minlength = self.length
-            for x in others:
-                minlength += x.length
+            minlength += sum(map(attrgetter("length"), others))
 
         next = RecordArray(
             nextcontents,
--- a/tests/test_3258_concat_empty_records.py
+++ b/tests/test_3258_concat_empty_records.py
@@ -1,0 +1,10 @@
+from __future__ import annotations
+
+import awkward as ak
+
+
+def test_concat_empty_records():
+    a = ak.Array([{}, {}, {}])
+    b = ak.Array([{}, {}])
+    c = ak.concatenate([a, b], axis=0)
+    assert c.to_list() == [{}, {}, {}, {}, {}]


jpivarski (approved):
Thanks! I think this is ready to go!

PR #3326, opened 2024-12-04
ci: use macos-latest as macos-12 was sunset

lgray:


lgray:
@jpivarski altering ci in a separate issue (looks like it needs a setting changed in the repository configuration as well for passed tests)

lgray:
Closing in favor of #3326.

PR #3325, opened 2024-12-04
fix: handle type tracer correctly in indexed option array flattening

lgray:
Fixes #3324

@pfackeldey if my little patch breaks things please give it a try. I'll try to cook up a test in the mean time.

PR #3327, opened 2024-12-05
chore: upgrade macos-12 -> macos-13

jpivarski:
It was downgraded in 65928c144d88c07381a7b4aaf4b926e255b1a8ed because of a problem that I hope was temporary. The reason this is a problem now is because macos-12 was removed on December 3 (https://github.com/actions/runner-images/issues/10721). We first saw this as a blocker in #3325. I hope that PR #3167 lets us use newer MacOS with the classic linker, which is what I think the problem was.


jpivarski:
"Build awkward-cpp: macos-13" got past the part where it dlopens the awkward-cpp so file, so I think it will be fine.

PR #3328, opened 2024-12-05
fix: Upstream API changes in cuDF 24.10

jpivarski:
This is exactly the same as PR #3309, except that it doesn't touch awkward-cpp/rapidjson. It supersedes that PR.


jpivarski:
I enabled auto-merge because I know it will take a while. @martindurant, if you see any issues, just shout and I'll stop the presses. (This will likely make it into a release today or tomorrow.)

martindurant:
Looks correct

jpivarski:
The only failure is in failing to build WASM (no Ninja: command not found), which is very unrelated to this PR and I made an issue (#3329), so I'm just going to merge this.

PR #3338, opened 2024-12-10
fix: `datetime.datetime.utcfromtimestamp` deprecation warning

### LINKED ISSUE #3337: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version.
### Version of Awkward Array

2.6.9

`Python 3.12.2`

The update is needed for 

https://github.com/scikit-hep/awkward/blob/8f69dbf6eb84f518150f47a81f9605642bbff837/dev/generate-kernel-signatures.py#L146

https://github.com/scikit-hep/awkward/blob/8f69dbf6eb84f518150f47a81f9605642bbff837/dev/generate-tests.py#L22

to timezone-aware objects:

```python
/awkward/dev/generate-tests.py:22: DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
  build_date = datetime.datetime.utcfromtimestamp(
```

ianna:
Fixes issue #3337

### DIFF
--- a/dev/generate-kernel-signatures.py
+++ b/dev/generate-kernel-signatures.py
@@ -143,9 +143,16 @@ cuda_kernels_impl = [
 
 
 def reproducible_datetime():
-    build_date = datetime.datetime.utcfromtimestamp(
-        int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
-    )
+    import sys
+
+    timestamp = int(os.environ.get("SOURCE_DATE_EPOCH", time.time()))
+
+    if sys.version_info >= (3, 11):
+        build_date = datetime.datetime.fromtimestamp(timestamp, tz=datetime.UTC)
+    else:
+        build_date = datetime.datetime.utcfromtimestamp(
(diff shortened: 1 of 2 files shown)


jpivarski (approved):
Thanks for updating `utcfromtimestamp`, for keeping ahead of deprecations. I'm not sure why the environment variable `SOURCE_DATE_EPOCH` (not defined in my terminal...) gets precedence over `time.time()`, which I think should always work. If this were in the main codebase, I'd be more concerned because it would be introducing a dependence on the environment variable state, rather than anything in Python, but I think it's fine for the developer scripts.

The failing tests are all coming from JAX. I started running a test on the `main` branch to see if it's something that changed in JAX: https://github.com/scikit-hep/awkward/actions/runs/12259902669 All of these test were passing just a few days ago, so it might be due to a new version of JAX.

PR #3333, opened 2024-12-06
chore: remove WASM building in docs.yml

jpivarski:

PR #3340, opened 2024-12-12
fix: current failing jax tests for `jax>=0.4.36`

pfackeldey:
This PR fixes the current failing JAX tests since jax version 0.4.36. It adds a flag to restore old (jax<0.4.36) tracing behavior, see more at: https://jax.readthedocs.io/en/latest/changelog.html#jax-0-4-36-dec-5-2024

At some point we likely want to address this properly, but for now this should make our jax tests pass again.

### DIFF
--- a/tests/test_1447_jax_autodiff_slices_ufuncs.py
+++ b/tests/test_1447_jax_autodiff_slices_ufuncs.py
@@ -4,13 +4,15 @@ from __future__ import annotations
 
 import numpy as np
 import pytest
+from packaging.version import parse as parse_version
 
 import awkward as ak
 
 jax = pytest.importorskip("jax")
 jax.config.update("jax_platform_name", "cpu")
 jax.config.update("jax_enable_x64", True)
-jax.config.update("jax_data_dependent_tracing_fallback", True)
+if parse_version(jax.__version__) >= parse_version("0.4.36"):
+    jax.config.update("jax_data_dependent_tracing_fallback", True)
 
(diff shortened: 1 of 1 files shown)


ianna (approved):
@pfackeldey - Thanks for prompt fix! Looks good to me.

jpivarski (approved):
Great, this fixes the tests! But it's a good example of how hard it is to work with JAX: `jax.config.update` is not in our control. It only takes effect if it is the first thing that's called after importing JAX. After that time, it's silently ignored!

jpivarski:
About the WASM failure: that one would be fixed by #3333. In fact, we have a cross-dependence and can't merge either one first without errors. I'll force-merge #3333 and then you can pick that up here, then we'll merge this one and the other PRs can pick up on that.

martindurant:
(#3333 is now merged, if you were waiting)

henryiii:
The jobs in https://github.com/scikit-hep/awkward/pull/3333 are still required, so non-admins can no longer merge. Why was that job removed?

jpivarski:
The WASM-building job in docs.yml was removed because it was building a WASM version of awkward-cpp for the documentation, but the documentation no longer includes an embedded JupyterLite iframe. (Instead, the "try it" button points to a raw-Pyodide page that gets awkward-cpp from the web.)

I don't know how #3333 passed if that's still required. I don't remember it waiting for that reason. But I'll remove the requirement now.

PR #3343, opened 2024-12-15
fix: stacktraces for custom exception handling

pfackeldey:
This PR fixes stack traces for awkward's custom error handling. It allows to use debuggers like "pdb" that try to go through the stack traces again, before they've been evicted from the stack (or wrongly ordered?) (I'm not sure what causes that exactly). This PR does not affect the custom decoration of errors.

### DIFF
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -2,7 +2,6 @@
 
 from __future__ import annotations
 
-import builtins
 import sys
 import threading
 import warnings
--- a/src/awkward/_errors.py
+++ b/src/awkward/_errors.py
@@ -51,11 +51,6 @@ class PartialFunction:
         return self.func(*self.args, **self.kwargs)
 
 
-class KeyError(builtins.KeyError):
-    def __str__(self):
-        return super(Exception, self).__str__()
-
-
 class ErrorContext:
     # Any other threads should get a completely independent _slate.
     _slate = threading.local()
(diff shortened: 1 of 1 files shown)


pfackeldey:
Could you have a look @henryiii on this PR? I remember you had some ideas about it in one of the recent awkward meetings

src/awkward/_errors.py
```diff
@@ -75,50 +69,33 @@ def __enter__(self):
...
-
     def decorate_exception(self, cls: type[E], exception: E) -> Exception:
-        if sys.version_info >= (3, 11, 0, "final"):
-            if issubclass(cls, (NotImplementedError, AssertionError)):
-                exception.add_note(
-                    "\n\nSee if this has been reported at https://github.com/scikit-hep/awkward/issues"
-                )
+        def _add_note(exception: E, note: str) -> E:
+            if sys.version_info >= (3, 11, 0, "final"):
+                exception.add_note(note)
```
jpivarski: Although the pre-PR version of the code had this explicit version number check, I wonder if the check can be more specific and less dependent on external knowledge (the fact that this is the relevant version number and it includes the word "final") with

```suggestion
            if hasattr(exception, "add_note"):
                exception.add_note(note)
```

That is, it seems like we want to start using this method as soon as it exists... unless it existed for a few versions in a broken state that we don't want to use it.

This is just a thought—not a requirement for this PR.

jpivarski (approved):
This looks good! (Minor comment below.)

The JAX test failures should be dealt with elsewhere and in one place, perhaps as a single PR that fixes or removes them and then gets merged into all of these pending PRs.

pfackeldey:
Great! The jax test fix is in #3340 already :)

src/awkward/_errors.py
henryiii: I'd recommend dropping the "final" (which doesn't work anyway), but this is better because users can see _why_ `add_note` might be missing, and static tools like Ruff can automatically remove the if statement when the min version is set to 3.11.

henryiii: Though if this is generic, I think the `exceptiongroup` backport includes `add_note`. So if there's a chance it could be from there, you could keep the change.

henryiii:
I don't think it's related, but there's a lot of annotations like:

```
jax.errors.UnexpectedTracerError: Encountered an unexpected tracer. A function transformed by JAX had a side effect, allowing for a reference to an intermediate value with type float64[3] wrapped in a JVPTracer to escape the scope of the transformation.
JAX transformations require that functions explicitly return their outputs, and disallow saving intermediate values to global state.
To catch the leak earlier, try setting the environment variable JAX_CHECK_TRACER_LEAKS or using the `jax.checking_leaks` context manager.
See https://jax.readthedocs.io/en/latest/errors.html#jax.errors.UnexpectedTracerError
```

being produced.

Edit: Ah, I see Jim mentioned this. Edit: And Peter mentioned a fix. I'm behind. :)

henryiii (approved):

PR #3344, opened 2024-12-16
fix: control attrs better as described in issue #3277

### LINKED ISSUE #3277: We should control the type of `ak.Array.attrs`, and it should be immutable
Right now,we just take the `attrs` that a user gives us and pass it around, allowing it to be mutated arbitrarily.

```python
>>> import awkward as ak
>>> class MyDict(dict): pass
... 
>>> attrs = MyDict(one=1)

>>> array1 = ak.Array([1, 2, 3], attrs=attrs)
>>> type(array1.attrs)
<class '__main__.MyDict'>
>>> array1.attrs["two"] = 2

>>> array2 = array1 * 1.1
>>> type(array2.attrs)
<class '__main__.MyDict'>
>>> array2.attrs["three"] = 3

>>> array2.attrs
{'one': 1, 'two': 2, 'three': 3}
>>> array1.attrs
{'one': 1, 'two': 2, 'three': 3}
>>> attrs
{'one': 1, 'two': 2, 'three': 3}
```

This is not how Awkward Array things usually work, usually we

* make a copy to unlink "our" version of the data from the user-provided collection
* regularize the type to simplify our internal code by being able to make assumptions about its type
* prefer to limit mutability; the `ak.Array` object is mutable, but so far, none of the attributes hanging off of the `ak.Array` objects have been mutable

This is all just a matter of expectations and different users will have different expectations, so this issue is labeled "policy" and we can brainstorm in the comments. However, I think the current state of affairs becomes a problem when people serialize/deserialize through Arrow/Parquet, since serialization

* copies data, unlinking it from the original
* can't have unrecognized types like `MyDict`
* is essentially immutable

Here's what I think it should do instead. I think that, given an arbitrary `attrs` Mapping, we should copy it into a frozendict-like type that we control, pass that frozendict from arrays to derived arrays, and _replace_ it when modified by `__setitem__`, similar to the way that `ak.Array.__setitem__` replaces its layout, rather than changing its layout in-place. (Because layout nodes are highly shared among arrays; the `ak.Array` outer shell is not.)

Our `attrs` type would be something like

```python
class Attrs:
    def __init__(self, parent: ak.Array, data: Mapping[str, Any]):
        self._parent = parent
        self._data = dict(data)

    def __getitem__(self, key: str):
        return self._data[key]

    def __setitem__(self, key: str, value: Any):
        self._parent._attrs = Attrs(self._parent, self._data | {key: value})
```

The effect of this is that it would still allow

```python
>>> array1.attrs["one"] = 1
>>> array1.attrs
{"one": 1}
```

and `attrs` would be passed to derived arrays

```python
>>> array2 = array1 * 1.1
>>> array2.attrs
{"one": 1}
```

but influence would not be propagated backward:

```python
>>> array2.attrs["two"] = 2
>>> array2.attrs
{"one": 1, "two": 2}
>>> array1.attrs
{"one": 1}
```

This is what I think Awkward Arrays should do, but it's open for discussion as to whether it's too weird for Python. It fits in better with what serialized data does: if you save an array to a file and open it in another process and start making changes, it absolutely won't affect the original array—this ensures the same semantics for derived arrays regardless of whether they're saved to files.

Side-note: `parameters` should do this, too. This issue would be resolved by ensuring that `attrs` and `parameters` both have this semantics, or by ensuring that `parameters` are strictly immutable.

```python
>>> layout1 = ak.contents.NumpyArray(np.array([1, 2, 3]), parameters={"one": 1})
>>> layout1.parameters
{'one': 1}
>>> layout2 = layout1[[2, 0, 0, 1]]
>>> layout2.parameters["two"] = 2
>>> layout2.parameters
{'one': 1, 'two': 2}
>>> layout1.parameters
{'one': 1, 'two': 2}
```

A note on performance: frozendicts don't need to be copied deeply. They can be passed around by reference. If the Attrs class is implemented with a reference to a parent, then this is the only thing that needs to change when deriving one array from another.

A note on memory management: the way that I described it above, with a reference to parent, introduces a cyclic reference (because the `ak.Array` has to have a reference to the Attrs). I think this cycle can be broken with two types; one is a short-lived proxy.

A note on deprecation: this would change semantics in two things that have already been "in the wild," both `attrs` and `parameters`. But there's a natural place to put the deprecation warning—in the `__setitem__` call. This can go through the normal two-version deprecation cycle.

If this seems to be too big of a change, here's a more conservative suggestion: keep the value/reference semantics that we have now, but at least ensure that the internal representation is a dict. We could refuse to accept a general Mapping or MutableMapping and only accept simple dicts, since we know we'll be able to reconstitute such a thing when deserializing from a file. (This is the opposite of what I told @pfackeldey in #3238.)

pfackeldey:
This changes the behavior of `attrs` as described in https://github.com/scikit-hep/awkward/issues/3277. This PR closes https://github.com/scikit-hep/awkward/issues/3277.

An example that works with this PR is:
```python
import awkward as ak

arr = ak.Array([1])
arr.attrs["foo"] = "bar"

arr2 = ak.copy(arr)
assert arr2.attrs == arr.attrs

arr2.attrs["foo"] = "baz"
assert arr2.attrs != arr.attrs
```

Because of this change (and that we have now a `Attrs` class as proposed in https://github.com/scikit-hep/awkward/issues/3277) I had to change `is` to `==` in some of the tests.

### DIFF
--- a/src/awkward/_attrs.py
+++ b/src/awkward/_attrs.py
@@ -43,7 +43,7 @@ def attrs_of(*arrays, attrs: Mapping | None = None) -> Mapping:
 
 
 def without_transient_attrs(attrs: dict[str, Any]) -> JSONMapping:
-    return {k: v for k, v in attrs.items() if not k.startswith("@")}
+    return {k: v for k, v in attrs.items() if not (isinstance(k, str) and k.startswith("@"))}
 
 
 class Attrs(Mapping):
--- a/src/awkward/_attrs.py
+++ b/src/awkward/_attrs.py
@@ -1,6 +1,7 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward-1.0/blob/main/LICENSE
 from __future__ import annotations
 
+import weakref
 from collections.abc import Mapping
 from types import MappingProxyType
 
@@ -47,14 +48,18 @@ def without_transient_attrs(attrs: dict[str, Any]) -> JSONMapping:
 
 class Attrs(Mapping):
     def __init__(self, ref, data: Mapping[str, Any]):
-        self._ref = ref
+        self._ref = weakref.ref(ref)
         self._data = _freeze_attrs(data)
 
     def __getitem__(self, key: str):
         return self._data[key]
 
     def __setitem__(self, key: str, value: Any):
-        self._ref._attrs = _unfreeze_attrs(self._data) | {key: value}
+        ref = self._ref()
+        if ref is None:
+            msg = "The reference array has been deleted. If you still need to set attributes, convert this 'Attrs' instance to a dict with '.to_dict()'."
+            raise ValueError(msg)
+        ref._attrs = _unfreeze_attrs(self._data) | {key: value}
 
     def __iter__(self):
         return iter(self._data)
@@ -63,7 +68,10 @@ class Attrs(Mapping):
         return len(self._data)
 
     def __repr__(self):
-        return repr(_unfreeze_attrs(self._data))
+        return f"Attrs({_unfreeze_attrs(self._data)!r})"
+
+    def to_dict(self):
+        return _unfreeze_attrs(self._data)
 
 
 def _freeze_attrs(attrs: Mapping[str, Any]) -> Mapping[str, Any]:
--- a/src/awkward/_attrs.py
+++ b/src/awkward/_attrs.py
@@ -43,7 +43,9 @@ def attrs_of(*arrays, attrs: Mapping | None = None) -> Mapping:
 
 
 def without_transient_attrs(attrs: dict[str, Any]) -> JSONMapping:
-    return {k: v for k, v in attrs.items() if not (isinstance(k, str) and k.startswith("@"))}
+    return {
+        k: v for k, v in attrs.items() if not (isinstance(k, str) and k.startswith("@"))
+    }
 
 
 class Attrs(Mapping):
--- a/src/awkward/_attrs.py
+++ b/src/awkward/_attrs.py
@@ -2,6 +2,7 @@
 from __future__ import annotations
 
 from collections.abc import Mapping
+from types import MappingProxyType
 
 from awkward._typing import Any, JSONMapping
 
@@ -42,3 +43,32 @@ def attrs_of(*arrays, attrs: Mapping | None = None) -> Mapping:
 
 def without_transient_attrs(attrs: dict[str, Any]) -> JSONMapping:
     return {k: v for k, v in attrs.items() if not k.startswith("@")}
+
+
+class Attrs(Mapping):
+    def __init__(self, ref, data: Mapping[str, Any]):
+        self._ref = ref
+        self._data = _freeze_attrs(data)
+
+    def __getitem__(self, key: str):
+        return self._data[key]
+
+    def __setitem__(self, key: str, value: Any):
+        self._ref._attrs = _unfreeze_attrs(self._data) | {key: value}
+
+    def __iter__(self):
+        return iter(self._data)
+
+    def __len__(self):
+        return len(self._data)
+
+    def __repr__(self):
+        return repr(_unfreeze_attrs(self._data))
+
+
+def _freeze_attrs(attrs: Mapping[str, Any]) -> Mapping[str, Any]:
+    return MappingProxyType(attrs)
+
+
+def _unfreeze_attrs(attrs: Mapping[str, Any]) -> dict[str, Any]:
+    return dict(attrs)
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -381,7 +381,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
     @property
     def attrs(self) -> Attrs:
         """
-        The mutable mapping containing top-level metadata, which is serialised
+         The mapping containing top-level metadata, which is serialised
         with the array during pickling.
 
         Keys prefixed with `@` are identified as "transient" attributes
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -18,7 +18,7 @@ from awkward_cpp.lib import _ext
 
 import awkward as ak
 import awkward._connect.hist
-from awkward._attrs import attrs_of, without_transient_attrs
+from awkward._attrs import Attrs, attrs_of, without_transient_attrs
 from awkward._backends.dispatch import register_backend_lookup_factory
 from awkward._backends.numpy import NumpyBackend
 from awkward._behavior import behavior_of, get_array_class, get_record_class
@@ -42,7 +42,7 @@ from awkward._pickle import (
     unpickle_record_schema_1,
 )
 from awkward._regularize import is_non_string_like_iterable
-from awkward._typing import Any, MutableMapping, TypeVar
+from awkward._typing import Any, TypeVar
 from awkward._util import STDOUT
 from awkward.prettyprint import Formatter
 from awkward.prettyprint import valuestr as prettyprint_valuestr
@@ -337,7 +337,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         if behavior is not None and not isinstance(behavior, Mapping):
             raise TypeError("behavior must be None or a mapping")
 
-        if attrs is not None and not isinstance(attrs, MutableMapping):
+        if attrs is not None and not isinstance(attrs, Mapping):
             raise TypeError("attrs must be None or a mapping")
 
         if named_axis:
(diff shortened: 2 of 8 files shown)


jpivarski (approved):
This is both a nice API design and it's the first good use of weak references that I've seen. These are intended to be short-lived objects that need to refer back to the `ak.Array` that made them, and creating a cyclic reference (putting the array into a list of things that won't be deleted until garbage collection) just for that doesn't make sense.

In fact, there's one more proxy like this: the `ak.Array.mask` property makes a proxy like this, for applying a slice as `ak.mask` instead of `__getitem__`. The fact that masking creates cyclic references might be why some physics analyses are using more memory than others, in a way that would surely be surprising/unexpected.

jpivarski:
Oh, I forgot to mention: I don't see anything here that enforces the attribute type or immutability. I like the mutability in this object, though: seeing how this can be used, I'm convinced that

```python
my_array.attrs["something"] = 123
```

is a good thing. About the type: this PR enforces the key type to be strings, but it doesn't enforce the value type to be anything in particular, right? Persistent attributes would eventually need to be JSON or pickleable to be saved in files, perhaps Parquet files or pickles, but the transient attributes don't need to be restricted. There might be good reason to let a transient attribute be an unpickleable lambda function.

pfackeldey:
I can have a look at `ak.Array.mask` aswell with weakref in a different PR.

> this PR enforces the key type to be strings, but it doesn't enforce the value type to be anything in particular, right? 

No, the types have never been enforced, it's just type annotations. With and without this PR the types can be anything that a Mapping would accept (key: Hashable, value: Any). 

I don't know how the serialization of attrs works. Shouldn't the serialization step just try to serialize attrs, and if it can't drop a warning and save the arrays without attrs? (I'd expect something like this as a behavior)

edit: from a brief look at the `ak.to_parquet` serialization method I don't see any serialization of `.attrs` right now, it seems to serialize on the layout level, which doesn't have `.attrs`. The serialization of `.attrs` is only defined on `__reduce_ex__`, so for pickle. Are you sure that ever worked for parquet?

jpivarski:
I mentioned immutability and controlling the type because that's what the issue (#3277) was saying, although we can certainly change our minds on that.

Ah! "Controlling the type" meant ensuring that it's a dict. It doesn't say anything about other interpretations:

* controlling the key type: I think we always want these to be strings (as it is for `parameters`)
* controlling persistent value types: maybe—at least to raise an exception when trying to serialize it if it's not serializable (JSON only or fallback to pickle?)
* controlling transient value types: no. That's too strict.

I'd rather not raise a _warning_ when failing to serialize. This is a general problem with warnings in Python: user A of package B that depends on package C that depends on Awkward will see the warning and (1) not know what it's for, since it's something internal to B or C, and (2) not get a stack trace to report. It could be an exception or silently skip writing the attribute, but there are more problems with silently skipping.

I just checked `parameters` ([here](https://github.com/scikit-hep/awkward/blob/3725585ca2f8cb8038f325ca94b1451b061e5d15/src/awkward/contents/content.py#L136-L173)): it doesn't do any runtime guarantee of the key type, either. Okay, at least that's consistent. Our policy is that the key types should be strings, but that's only enforced at the type annotation level.

For persistent value types, if it's not serializable, I guess it raises an exception now, right?

Yes, both before

```python
>>> a = ak.Array([1, 2, 3])
>>> a.attrs["yikes"] = lambda x: x
>>> pickle.dumps(a)
_pickle.PicklingError: Can't pickle <function <lambda> at 0x752e88cd9a80>: attribute lookup <lambda> on __main__ failed
```

and after

```python
>>> a = ak.Array([1, 2, 3])
>>> a.attrs["yikes"] = lambda x: x
>>> pickle.dumps(a)
_pickle.PicklingError: Can't pickle <function <lambda> at 0x791809fd5a80>: attribute lookup <lambda> on __main__ failed
```

this PR. I think everything above is in a good state and this PR is ready to merge.

pfackeldey:
Yes, exactly, that's the policy right now. Maybe it does make sense to enforce at least the key type to be a string because otherwise this is a bad error:
```python
a = ak.Array([1, 2, 3])
a.attrs[1] = lambda x: x
pickle.dumps(a)
# > AttributeError: 'int' object has no attribute 'startswith'
```
because the pickling step checks for transient keys first.

I have an idea:
We don't enforce the key type to be `str`, but instead define transient keys only as strings; i.e. 'if you're not using a string, it can't be transient'. This would be a simple change in [this function](https://github.com/scikit-hep/awkward/blob/main/src/awkward/_attrs.py#L44) that would additionally check for `isinstance(key, str)`.

What do you think about this?

jpivarski:
> We don't enforce the key type to be `str`, but instead define transient keys only as strings; i.e. 'if you're not using a string, it can't be transient'.

That makes sense to me. The current rule for identifying a transient is that it starts with an "@"; the new rule would be that it's a string and starts with an "@" (which was kind-of implicit in the current rule).

We do want people to only make `parameter` and `attr` keys that are strings. The type annotations should be enough to assert that this is the API. (In the past, I would have made this a runtime check.)

src/awkward/highlevel.py
```diff
@@ -379,7 +379,7 @@ def _update_class(self):
         self.__class__ = get_array_class(self._layout, self._behavior)
 
     @property
-    def attrs(self) -> Mapping:
+    def attrs(self) -> Attrs:
         """
         The mutable mapping containing top-level metadata, which is serialised
```
agoose77:
```suggestion
         The mapping containing top-level metadata, which is serialised
```

src/awkward/_attrs.py
```diff
@@ -41,4 +43,42 @@ def attrs_of(*arrays, attrs: Mapping | None = None) -> Mapping:
 
 
 def without_transient_attrs(attrs: dict[str, Any]) -> JSONMapping:
-    return {k: v for k, v in attrs.items() if not k.startswith("@")}
+    return {
+        k: v for k, v in attrs.items() if not (isinstance(k, str) and k.startswith("@"))
```
agoose77, line 47: I think we should simplify this and only allow strings. Attributes are nominally strong types, and by enforcing that we do not need to branch in every routine that handles attributes. Furthermore we don't then have to answer the question of what string-like objects should do (I.e non string subclasses).

We can possibly relax that down the road if there is a compelling case.

pfackeldey: I can enforce that all keys of `Attrs` are strings in the `Attrs` constructor. What do you think of this @jpivarski and @agoose77 ?
In principle I like it how it is now: users can store _anything_ in `.attrs` but not necessarily serialize it - pickle would complain properly in that case.

pfackeldey: I'm merging it as is, my thought process about this is that `.attrs` might well be used just for in-memory processing to keep track of _any_ metadata, it's not necessarily only used for serialization - so let's not restrict it preemptively for this (we would need to restrict values then aswell). If one pickles the array, `.attrs` obeys pickling rules and it will give reasonable errors to the user the same way as manually pickling python objects would. 
Also: it was possible to store anything until now, so restricting this might surprise some users? 

Within the awkward codebase we ensure using strings as keys through type annotations.

Hope this is fine for you, if not, I'm happy to address this in another PR :+1:

agoose77: @pfackeldey it was originally intended that attrs should only contain string keys -- e.g.
```
>>> import awkward
>>> import awkward as ak
>>> import pickle
>>> x = pickle.dumps(ak.Array([], attrs={1: 2}))
  File "/home/angus/Git/awkward/src/awkward/highlevel.py", line 1644, in __reduce_ex__
    attrs = without_transient_attrs(self._attrs)
  File "/home/angus/Git/awkward/src/awkward/_attrs.py", line 44, in without_transient_attrs
    return {k: v for k, v in attrs.items() if not k.startswith("@")}
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/angus/Git/awkward/src/awkward/_attrs.py", line 44, in <dictcomp>
    return {k: v for k, v in attrs.items() if not k.startswith("@")}
                                                  ^^^^^^^^^^^^
AttributeError: 'int' object has no attribute 'startswith'
```

If we change that behaviour, we won't easily be able to undo it in future, where it will become breaking.

pfackeldey: I didn't change that behavior, before this PR it was possible to put non-string keys into `.attrs`. In that case it would fail with your error, with this PR I let it fall through to pickle that would either store it if it can or raise an exception that its not pickable (with the correct pickle error message). And that would be more understandable than `AttributeError: 'int' object has no attribute 'startswith'`, no?

pfackeldey: Oh... I see what you mean: we restricted it to be string keys only for pickling, got you! Ok, I'll make sure we only allow string keys - but only for the pickling step! I assume this is what you meant?

agoose77: To be super clear, the original intention of `attrs` is that it should hold string keys, and I think we should write our code as though that is true. I was replying to 
>  it's not necessarily only used for serialization - so let's not restrict it preemptively for this (we would need to restrict values then aswell).

Where I think we _should_ restrict preemptively given the above policy. If people really want to stash state in `attrs`, then they should have a custom namespace e.g. `attrs['my-namespace'] = {}`, I think!

PR #3347, opened 2024-12-16
perf: make the ak.Array.mask weakly dependent on the array itself

pfackeldey:
This PR makes the reference array (that is to be masked) weakly referenced in `ak.Array.mask` as suggested by @jpivarski in https://github.com/scikit-hep/awkward/pull/3344#pullrequestreview-2506946032

### DIFF
--- a/docs/user-guide/how-to-create-missing.md
+++ b/docs/user-guide/how-to-create-missing.md
@@ -265,6 +265,6 @@ def faster_example():
 
 data, mask = faster_example()
 
-array = ak.Array(data).mask[mask]
+array = ak.mask(data, mask)
 array
 ```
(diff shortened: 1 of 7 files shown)


pfackeldey:
It looks like there are weakref/GC differences for CPython and PyPy - interesting: https://doc.pypy.org/en/latest/cpython_differences.html#differences-related-to-garbage-collection-strategies (that's why the new test fails for PyPy currently).

jpivarski (approved):
This looks great!

You're right that PyPy objects have a different lifetime. Unlike CPython, everything in PyPy waits for the garbage collector; reference count = 0 is not used to immediately delete an object.

PR #3348, opened 2024-12-17
feat: add pretty nbytes repr to .show and jupyter repr

pfackeldey:
This is a little PR that adds (optionally, default is False) a new line to `.show(nbytes=True)` and the jupyter repr of high-level ak.Arrays and ak.Records. It's for the lazy (like me) who don't want to do the calculation in their head...

Preview (`.show()`):

Screenshot 2024-12-16 at 8 40 21 PM

Preview (Jupyter repr):
Screenshot 2024-12-16 at 8 44 03 PM

What do you think?

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1387,6 +1387,54 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
         return f"<{pytype}{valuestr}{axisstr} type={typestr}>"
 
+    def _show_rows(
+        self,
+        limit_rows=20,
+        limit_cols=80,
+        type=False,
+        named_axis=False,
+        nbytes=False,
+        backend=False,
+        *,
+        formatter=None,
+        precision=3,
+    ) -> list[str]:
+        rows = []
+        formatter_impl = Formatter(formatter, precision=precision)
+
+        valuestr = prettyprint_valuestr(
+            self, limit_rows, limit_cols, formatter=formatter_impl
+        )
+        rows.append(valuestr)
+
+        if type:
+            out_io = io.StringIO()
+            out_io.write("type: ")
+            self.type.show(stream=out_io)
+            rows.append(out_io.getvalue())
+
+        # other info
+        if named_axis and self.named_axis:
+            out_io = io.StringIO()
+            out_io.write("axes: ")
+            out_io.write(
+                _prettify_named_axes(self.named_axis, delimiter=", ", maxlen=None)
+            )
+            rows.append(out_io.getvalue())
+        if nbytes:
+            out_io = io.StringIO()
+            out_io.write(f"nbytes: {bytes_repr(self.nbytes)}")
+            rows.append(out_io.getvalue())
+        if backend:
+            out_io = io.StringIO()
+            out_io.write(f"backend: {self.layout.backend.name}")
+            rows.append(out_io.getvalue())
+
+        # make sure the type is always the second row, don't move it
+        if type:
+            assert rows[1].startswith("type: ")
+        return rows
+
     def show(
         self,
         limit_rows=20,
@@ -1394,6 +1442,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         type=False,
         named_axis=False,
         nbytes=False,
+        backend=False,
         stream=STDOUT,
         *,
         formatter=None,
@@ -1405,9 +1454,14 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
             limit_cols (int): Maximum number of columns (characters wide).
             type (bool): If True, print the type as well. (Doesn't count toward number
                 of rows/lines limit.)
+            named_axis (bool): If True, print the named axis as well. (Doesn't count toward number
+                of rows/lines limit.)
+            nbytes (bool): If True, print the number of bytes as well. (Doesn't count toward number
+                of rows/lines limit.)
+            nbytes (bool): If True, print the backend of the array as well. (Doesn't count toward number
+                of rows/lines limit.)
             stream (object with a ``write(str)`` method or None): Stream to write the
                 output to. If None, return a string instead of writing to a stream.
-
             formatter (Mapping or None): Mapping of types/type-classes to string formatters.
                 If None, use the default formatter.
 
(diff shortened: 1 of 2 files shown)


agoose77:
Love this improvement! There's the usual caveat note about `nbytes` changing depending upon how you share an array, but that's true already!

pfackeldey:
Yes :+1:, also it does not take into account the python objects/structure and whatever is in `attrs`, but I'd argue that this is only reasonable - the array(s) (should) make up 99% of consumed memory.

jpivarski:
This is very nice! But can we make it say "`nbytes:`" instead of "`size:`" (even if it's converted into units like "`MB`")? That would do two things: it would provide a hint as to how to access this number programmatically and it would avoid a confusion in which NumPy uses "size" to mean the number of items (`size * itemsize = nbytes`).

Also, are you making the right distinction between "MB" (megabytes) and "MiB" (mebibytes)? One system uses factors of 1000 and the other uses factors of 1024.

pfackeldey:
I chose `size` initially because of the 4 letters that align visually nicely with `axes` and `type` for the repr (this may not be a good argument though). I can change it to `nbytes` :+1:
 
 > Also, are you making the right distinction between "MB" (megabytes) and "MiB" (mebibytes)? One system uses factors of 1000 and the other uses factors of 1024.

That should be correct, I'm using factors of `1000` on `ak.Array.nbytes` for formatting the unit, which should then be in "MB" and not "MiB", right?

pfackeldey:
Oh btw, one thing that I noticed is that the typetracer backend reports `nbytes` as if it is a numpy array. I can see why we want to mimic numpy arrays in every aspect with the typetracer backend, but wouldn't it make sense to set `nbytes` for typetracers explicitly to `0` ? This is just a minor design choice...

jpivarski:
https://simple.wikipedia.org/wiki/Mebibyte (I'm in a meeting)

pfackeldey:
as discussed I changed:
- "KB" -> "kB"
- "size:" -> "nbytes:"
- added `backend` as an argument 
- sorted the rows (without `array` and `type`) in increasing/decreasing order for `.show`/`jupyter-repr`

Example:
```python
In [1]: import awkward as ak; import numpy as np

In [2]: array = ak.with_named_axis(
   ...:     ak.zip({
   ...:         "one": np.full((1234567,), 1.0),
   ...:         "two": np.full((1234567,), 2.0),
   ...:         "three": np.full((1234567,), 3.0),
   ...:     }),
   ...:     named_axis=("some",),
   ...: )

In [3]: array.show(type=True, named_axis=True, nbytes=True, backend=True)
type: 1234567 * {
    one: float64,
    two: float64,
    three: float64
}
axes: some:0
nbytes: 29.6 MB
backend: cpu
[{one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 {one: 1, two: 2, three: 3},
 ...,
```

For `.show(...)` the order is: 
(1) `type` (2) ascending order of `axes`, `nbytes`, `backend` (based on the `<prefix>:` length) (3) `array`

For the Jupyter repr the order is:
(1) `array` (2) `------` (3) descending order of `axes`, `nbytes`, `backend` (based on the `<prefix>:` length) (4) `type`

src/awkward/highlevel.py
```diff
@@ -1387,12 +1387,62 @@ def _repr(self, limit_cols):
 
         return f"<{pytype}{valuestr}{axisstr} type={typestr}>"
 
+    def _show_rows(
```
jpivarski: How different are the two `_show_rows` functions? Is one under `Array` and the other under `Record`?

src/awkward/highlevel.py
```diff
@@ -1387,12 +1387,62 @@ def _repr(self, limit_cols):
...
+            )
+            rows.append(out_io.getvalue())
+        if nbytes:
+            out_io = io.StringIO()
+            out_io.write(f"nbytes: {bytes_repr(self.nbytes)}")
+            rows.append(out_io.getvalue())
+        if backend:
+            out_io = io.StringIO()
+            out_io.write(f"backend: {self.layout.backend.name}")
+            rows.append(out_io.getvalue())
```
jpivarski: I understand why `type` uses a `io.StringIO`, because that's the way to get a string out of its `show` method (unless passing `stream=None` does it, but if so, it's because `stream=None` is implemented as "return a string, rather than modifying a stream").

But the named axes, nbytes, and backend don't need to, and it seems wasteful to create a `io.StringIO`, write one line to it, and convert that to a string. Not necessarily wasteful in computational resources (I don't know how `io.StringIO` is implemented; it might be fine), but it's a roundabout dance that brings in extra concepts without an obvious benefit.

Is it for symmetry with the `type` case? Symmetry arguments can make sense, but this one seems like a bit of a stretch.

src/awkward/highlevel.py
```diff
@@ -1404,9 +1454,14 @@ def show(
             limit_cols (int): Maximum number of columns (characters wide).
             type (bool): If True, print the type as well. (Doesn't count toward number
                 of rows/lines limit.)
+            named_axis (bool): If True, print the named axis as well. (Doesn't count toward number
+                of rows/lines limit.)
+            nbytes (bool): If True, print the number of bytes as well. (Doesn't count toward number
+                of rows/lines limit.)
+            nbytes (bool): If True, print the backend of the array as well. (Doesn't count toward number
+                of rows/lines limit.)
```
jpivarski: Eventually, someday, `all: bool` would set all of these to `True`.

jpivarski: Also, the argument is spelled `named_axis` here but is presented as `axes` in the text output.

Yet one more thing: these are all boolean arguments, but they're allowed to be positional (they're before the `*`). As an API design, it's safer to force users to type the keyword names of contiguous boolean arguments.

jpivarski (approved):

src/awkward/highlevel.py
pfackeldey: they're exactly the same, in fact many are the same between `Array` and `Record`, e.g. `show` and many properties such as `nbytes`, `layout`, `behavior`, `named_axis`, `positional_axis`, and also some functions such as `to_list()`. They mainly differ in their docstring though.

src/awkward/highlevel.py
pfackeldey: That's right, I can change that, the roundtrip is unnecessary. :+1:

src/awkward/highlevel.py
pfackeldey: > Eventually, someday, all: bool would set all of these to True.

"Someday" will be (hopefully) tomorrow when I add `all: bool` :+1:

> Also, the argument is spelled named_axis here but is presented as axes in the text output.

Good point, I'll change `axes:` to `named axis:`.

> Yet one more thing: these are all boolean arguments, but they're allowed to be positional (they're before the *). As an API design, it's safer to force users to type the keyword names of contiguous boolean arguments.

Sounds good, will change that :+1:

src/awkward/highlevel.py
pfackeldey: if we would have a `ak.highlevel.HighLevelArray` base class these things could live there. It would maybe also simplify checks like:
`isinstance(array, (ak.Array, ak.Record))` -> `isinstance(array, ak.highlevel.HighLevelArray)`

pfackeldey: I moved `self._show_rows` into a helper function to avoid code duplication :+1:

src/awkward/highlevel.py
pfackeldey: all comments addressed

src/awkward/highlevel.py
pfackeldey: all comments addressed

pfackeldey:
@jpivarski could you have one last look at this PR? thanks 🙏

jpivarski (approved):
It looks great! I see that the code duplication is gone (now in `highlevel_array_show_rows`) and it has an `all` argument. I think the PR is ready to merge.

PR #3350, opened 2024-12-18
fix: enforce string-only keys in `.attrs`

pfackeldey:
As @agoose77 suggested (in https://github.com/scikit-hep/awkward/pull/3344) we should restrict keys of `.attrs` to str only.


jpivarski (approved):
Looks good! At first, I was wondering why we need a helper function just to do a runtime type check, but now I see that it's to use in a list comprehension and `|` union (instead of a more pedestrian sequence of statements). But that's fine.

This is ready to merge!

PR #3353, opened 2024-12-18
feat: improve error message when indexing placeholder arrays

pfackeldey:
This PR improves error messages of PlaceholderArrays by (1) keeping track of the field path to the placeholder (only for record arrays) and by (2) improving the error message text itself.

In the context of e.g.: https://github.com/scikit-hep/coffea/issues/1231

The new error message looks like:
```python
TypeError: PlaceholderArray supports only trivial slices, not ndarray

Awkward-array attempted to access a field 'Jet.area', but it has been excluded during a pre-run phase (possibly by Dask). If this was supposed to happen automatically (e.g. you're using Dask), please report it to the developers at: https://github.com/scikit-hep/awkward/issues

    numpy.add.__call__(
        <JetArray [[Jet, ..., Jet], ...] type='40 * var * ?Jet[area: float3...'>
        <CandidateArray [{x: 0, y: 0, z: 0, ...}, ...] type='40 * Candidate...'>
    )
```
Compared to 
```python
TypeError: PlaceholderArray supports only trivial slices, not ndarray

    numpy.add.__call__(
        repr-raised-TypeError
        <CandidateArray [{x: 0, y: 0, z: 0, ...}, ...] type='40 * Candidate...'>
    )
```
(`repr-raised-TypeError` was already fixed by https://github.com/scikit-hep/awkward/pull/3342)

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -306,7 +306,7 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
     ) -> ArrayLikeT | PlaceholderArray:
         if isinstance(x, PlaceholderArray):
             next_shape = self._compute_compatible_shape(shape, x.shape)
-            return PlaceholderArray(self, next_shape, x.dtype)
+            return PlaceholderArray(self, next_shape, x.dtype, x._field_path)
 
         if copy is None:
             return self._module.reshape(x, shape)
--- a/src/awkward/_nplikes/placeholder.py
+++ b/src/awkward/_nplikes/placeholder.py
@@ -17,10 +17,21 @@ if TYPE_CHECKING:
 
 
 class PlaceholderArray(ArrayLike):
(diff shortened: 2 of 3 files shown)


jpivarski (approved):
This is an improvement, particularly the fact that you can now specify the field where the PlaceholderArray is. We talked on Slack about also adding hints about how to fix the problem, but you're still developing that function.

I think this is ready to be merged!

pfackeldey:
> We talked on Slack about also adding hints about how to fix the problem, but you're still developing that function.

I'll update the error message once this function is there. :+1:

I'll add (for documentation purposes) why reporting this field path will be beneficial in the future as a comment here: 
This new function (that's in development) would fix it (by explicitly force-touching the missing field). However, the root cause should still be addressed if a user hits this error. Hitting this error is essentially always some sort of 'under-touching' of e.g. dask-awkward's column optimization - that may happen in any down-stream library or in awkward-array itself. Adding the field path will tell us at least what field should've been touched.

PR #3354, opened 2024-12-19
chore: use `_util.native_to_byteorder` function in `ak.from_buffers`

pfackeldey:
Directly use an existing utility function for byte ordering. Stumbled over this while working on #3353.

### DIFF
--- a/src/awkward/operations/ak_from_buffers.py
+++ b/src/awkward/operations/ak_from_buffers.py
@@ -186,10 +186,7 @@ def _from_buffer(
         return array[:count]
     else:
         array = nplike.frombuffer(buffer, dtype=dtype, count=count)
-        if byteorder != ak._util.native_byteorder:
-            return array.byteswap(inplace=False)
-        else:
-            return array
+        return ak._util.native_to_byteorder(array, byteorder)
 
 
 def _reconstitute(


jpivarski (approved):
It's true that Awkward Arrays (data and indexes) must always be native-endian, and it's also true that the buffers passed to `ak.from_buffers` may be big or little-endian, depending on the `byteorder` argument:

https://github.com/scikit-hep/awkward/blob/8285871a75db2dffebae1df0e031ca2aa6959f2d/src/awkward/operations/ak_from_buffers.py#L56-L58

and this is forced to be `"<"` when we pickle arrays as part of the (unwritten?) format specification:

https://github.com/scikit-hep/awkward/blob/fb245f13057595c8134f9c687cecd2623a9e7aee/src/awkward/highlevel.py#L1690-L1723

Digging into this, the `dtype` passed to `_from_buffer` (a private function) always comes from `primitive_to_dtype` or `index_to_dtype`. The `primitive_to_dtype` is always defined to be native endian (good):

https://github.com/scikit-hep/awkward/blob/fb245f13057595c8134f9c687cecd2623a9e7aee/src/awkward/types/numpytype.py#L67-L83

But `index_to_dtype` is always defined to be little-endian (probably bad):

https://github.com/scikit-hep/awkward/blob/8285871a75db2dffebae1df0e031ca2aa6959f2d/src/awkward/forms/form.py#L375-L381

The latter might be a mistake that I introduced in #2660 and it has never mattered because all machines are little-endian machines.

And now I realize that this PR wasn't about changing the byteswap behavior at all—it was just to use a utility function ([native_to_byteorder](https://github.com/scikit-hep/awkward/blob/fb245f13057595c8134f9c687cecd2623a9e7aee/src/awkward/_util.py#L46-L60)) instead of reimplementing those four lines here.

But my digging into it might have revealed a mistake. What do you think about `index_to_dtype`? It should be native-endian, not little-endian, right? I'm 90% sure of it (but this is always a problem because you can't test it to verify).

On its own, this PR is fine to merge.

agoose77:
This is not a bug fix, right? 

I'm tempted to rename the utility function, as it doesn't coerced to a byte order but rather just swaps.

pfackeldey:
> it was just to use a utility function ([native_to_byteorder](https://github.com/scikit-hep/awkward/blob/fb245f13057595c8134f9c687cecd2623a9e7aee/src/awkward/_util.py#L46-L60)) instead of reimplementing those four lines here.

yes It was only about this 😅 I'd leave it like that for now, because I'm too unfamiliar with litte/big endian as I would feel confident to correctly implement what you discovered. Hope this is fine 👍

jpivarski:
That's correct—this PR is not a bug-fix. But while I was thinking about it incorrectly, I probably discovered an unrelated bug.

The name of the utility function seems right to me, but I'm rather neutral on it. It swaps the byte order if the given buffers have non-native order.

PR #3355, opened 2024-12-19
fix: hash of `HashableDict`

pfackeldey:
Python's builtin hash function takes only 1 argument, not 2.


jpivarski (approved):
Yes, this is correct. I wonder how it never came up before? This bug would be hit as soon as any `HashableDict` is ever created.

pfackeldey:
Yeah, I don't know, I stumbled over this because my editor complained.

PR #3359, opened 2025-01-06
perf: add caches to reduce python overhead

pfackeldey:
This PR adds some caches to reduce python overhead. This does not affect any "number crunching" with arrays. Thus, the main benefit is for typetracing although I noticed a slight speed difference for the cpu backend aswell. 

I tested this with a NanoAOD MC sample of 100k events and the trijet mass calculation from the AGC (ref: https://github.com/iris-hep/idap-200gbps/blob/main/agc-coffea-2024.ipynb, cell 2).

I measured the runtime, peak memory consumption, accumulated memory consumption (total memory), and number of allocations applying the trijet mass calculation to a `coffea.NanoEvents` record array and got the following:

### CPU backend
|         | Runtime | Peak memory | Total memory | # Allocations |
|---------|---------|-------------|------------|---------------|
| main    |   29.3 ms ± 434 μs      |   7.025 MB          | 37.769 MB |     3691         |
| this PR (49f99f2c7fd6bf731d90feccd7b24e420446dbf8) |   25.6 ms ± 260 μs     |     7.025 MB        |  36.659 MB |     2975        |
| this PR (dc89223f5c68e3dcb488b1e9d298088f3bbb6011) |   24.8 ms ± 167 μs     |     7.025 MB        |  36.538 MB |     2915        |
| **this PR (a6f2c8f805bbea0d036dab02384b06403608646f, final)** |  **25.1 ms ± 126 μs**     |     **7.025 MB**        | **36.390 MB** |     **2633**        |

### TypeTracer backend
|         | Runtime | Peak memory | Total memory | # Allocations |
|---------|---------|-------------|---|------------|
| main    |   30.3 ms ± 145 μs      |   17.405 KB      | 2.609 MB    |      2882         |
| this PR (49f99f2c7fd6bf731d90feccd7b24e420446dbf8) |   25.3 ms ± 136 μs    |     17.405 KB     | 1.495 MB   |       2164        |
| this PR (dc89223f5c68e3dcb488b1e9d298088f3bbb6011) |   25.2 ms ± 135 μs     |     17.405 KB        |  1.375 MB |     2104        |
| **this PR (a6f2c8f805bbea0d036dab02384b06403608646f, final)** |   **24.4 ms ± 195 μs**     |     **16.336 KB**        |  **1.224MB** |     **1816**        |

I think this is a noticeable speedup (~25% faster in this example) for typetracing, which ultimately improves the user experience with dask-awkward.

### DIFF
--- a/src/awkward/_meta/numpymeta.py
+++ b/src/awkward/_meta/numpymeta.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from functools import cached_property
+
 from awkward._meta.meta import Meta
 from awkward._nplikes.shape import ShapeItem
 from awkward._typing import JSONSerializable
@@ -31,13 +33,13 @@ class NumpyMeta(Meta):
     def is_identity_like(self) -> bool:
         return False
 
-    @property
-    def minmax_depth(self) -> tuple[int, int]:
+    @cached_property
+    def minmax_depth(self) -> tuple[int, int]:  # type: ignore[override]
         depth = len(self.inner_shape) + 1
         return (depth, depth)
 
-    @property
-    def branch_depth(self) -> tuple[bool, int]:
+    @cached_property
+    def branch_depth(self) -> tuple[bool, int]:  # type: ignore[override]
         return (False, len(self.inner_shape) + 1)
 
     @property
--- a/src/awkward/_meta/recordmeta.py
+++ b/src/awkward/_meta/recordmeta.py
@@ -45,12 +45,14 @@ class RecordMeta(Meta, Generic[T]):
     def minmax_depth(self) -> tuple[int, int]:  # type: ignore[override]
         if len(self._contents) == 0:
             return (1, 1)
-        mins, maxs = [], []
-        for content in self._contents:
-            mindepth, maxdepth = content.minmax_depth
-            mins.append(mindepth)
-            maxs.append(maxdepth)
-        return (min(mins), max(maxs))
+        mindepth, maxdepth = self._contents[0].minmax_depth
+        for content in self._contents[1:]:
+            mindepth_, maxdepth_ = content.minmax_depth
+            if mindepth_ < mindepth:
+                mindepth = mindepth_
+            if maxdepth_ > maxdepth:
+                maxdepth = maxdepth_
+        return (mindepth, maxdepth)
 
     @cached_property
     def branch_depth(self) -> tuple[bool, int]:  # type: ignore[override]
--- a/src/awkward/_meta/recordmeta.py
+++ b/src/awkward/_meta/recordmeta.py
@@ -2,6 +2,8 @@
 
 from __future__ import annotations
 
+from functools import cached_property
+
 from awkward._meta.meta import Meta
 from awkward._regularize import is_integer
 from awkward._typing import Generic, JSONSerializable, TypeVar
@@ -39,8 +41,8 @@ class RecordMeta(Meta, Generic[T]):
     def is_identity_like(self) -> bool:
         return False
 
-    @property
-    def minmax_depth(self) -> tuple[int, int]:
+    @cached_property
+    def minmax_depth(self) -> tuple[int, int]:  # type: ignore[override]
         if len(self._contents) == 0:
             return (1, 1)
         mins, maxs = [], []
@@ -50,8 +52,8 @@ class RecordMeta(Meta, Generic[T]):
             maxs.append(maxdepth)
         return (min(mins), max(maxs))
 
-    @property
-    def branch_depth(self) -> tuple[bool, int]:
+    @cached_property
+    def branch_depth(self) -> tuple[bool, int]:  # type: ignore[override]
         if len(self._contents) == 0:
             return False, 1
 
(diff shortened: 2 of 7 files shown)


pfackeldey:
Oh there seems to be something new with mypy and the typetracer. I'm pretty sure this is unrelated to this PR though...

pfackeldey:
I'm done optimizing. Looking forward to your review & feedback @jpivarski. I structured the table such that you can see how much each commit brings to the table.

ianna (approved):
@pfackeldey - It looks impressive. I think, we should definitely consider merging it. Firstly,  we do not support Python versions older than 3.8. Secondly, the arrays are immutable. 

However, I wonder why both memory and speed improve (not one at the expense of another?). Also, is there a risk of stale cache? Thanks!

pfackeldey:
Hi @ianna,
Total memory is the cumulative sum of all memory allocated during the process: less allocations lead to less cumulated memory consumption. The speedup comes of course from hitting caches and not doing python things. The cache does exist a certain time and has some memory, but it doesn't increase the peak memory consumption. I think this is because it doesn't matter there if it is allocated on the fly or already there in the cache. There might be times where the cache is existing, but not needed, this would increase slightly the "instantaneous" memory consumption. I'd argue this doesn't matter too much because a) the caches are pretty small and b) the peak memory is not increasing (this is the most important metric).

Concerning the cache lifetimes and cache sizes:

### `@cached_property`

These caches are bound to the instance and thus are deleted when the instance is deleted. They only cache a tuple of two integers, which is just a few bytes, or a list of strings (fields), also just a few bytes. This shouldn't be any problem.

### `@lru_cache`

These caches live until the python process is closed (or you actively clear them). They cache function arguments with their hash (just a python int) and the function's return value(s). It's important to add that this is a LRU cache with a max size of 128 cache entries. Here, the return values can be slightly larger:
- `_broadcast_shapes`: tuple of ints -> O(bytes)
- `primitive_to_dtype`: numpy dtype -> O(bytes)
-  `dtype_to_primitive`: str -> O(bytes)
- `from_json`: ak.forms.Form -> can range from O(bytes) to O(kilo bytes), e.g. `coffea.NanoEvents` form (for MC) is ~580 KB (if serialized as a string), the form itself is a lot smaller - but I might not measure it correctly. This cache is usually hit in `ak.from_iter` with the ArrayBuilder. I can see rare scenarios  (i.e. _many different large_ `ak.forms.Forms`) where this cache can become a bit large (O(MB)), but I don't have a good feeling about how realistic this is. In these scenarios it might actually be very beneficial for the runtime to have this cache, because such large forms might take some time to recursively iterate over... (see: https://github.com/scikit-hep/awkward/blob/main/src/awkward/forms/form.py#L54-L202)
If you want, I can remove this one, I have no strong opinions about it. (If I recall correctly this cache did not significantly contribute to the speedups shown in the tables from above)

Concerning stale caches: I don't think it can happen that the caches are not updated correctly. The cache keys are based on the hashes of the function arguments and they are well-defined and unique (e.g. tuples of ints). So I'd argue there's nothing to worry about stale caches, unless there's a bug in `functools` or how cpython calculates the hashes of builtin data types, which is something we can rely on.

pfackeldey:
One thing I want to mention here:
I found it quite interesting that even for 100k events (typical chunk size of a batch job) the trijetmass calculation of the AGC runs at similar speed for the typetracer or with actual data... Has anyone else ever experienced this? Or is this snippet of the trijetmass calculation an "unlucky" scenario for the runtime with typetracer backend?

pfackeldey:
And a side note for future reference:
I tested adding `__slots__` to all meta, index, and content classes, but I saw no measurable speedup.

martindurant:
> I tested adding __slots__ to all meta, index, and content classes, but I saw no measurable speedup.

It's not faster, but it should save memory across many instances. It amounts to storing a dict per instance without slots, and a tuple per instance with - so you save the storage of the keys.

pfackeldey:
> > I tested adding **slots** to all meta, index, and content classes, but I saw no measurable speedup.
> 
> It's not faster, but it should save memory across many instances. It amounts to storing a dict per instance without slots, and a tuple per instance with - so you save the storage of the keys.

That's right! But attribute access is faster with `__slots__`, and since we access metadata (e.g. `minmax_depth`, `purelist_depth`, ...) very frequently I was hoping to see some speedup, but it wasn't measurable. 

E.g.:
```python
In [1]: class Foo:
   ...:     def __init__(self):
   ...:         self.x = 1
   ...:

In [2]: class FooSlots:
   ...:     __slots__ = ("x",)
   ...:
   ...:     def __init__(self):
   ...:         self.x = 1
   ...:

In [3]: foo = Foo()

In [4]: fooslots = FooSlots()

In [5]: %timeit foo.x
5.19 ns ± 0.0676 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)

In [6]: %timeit fooslots.x
4.89 ns ± 0.0126 ns per loop (mean ± std. dev. of 7 runs, 100,000,000 loops each)
```

martindurant:
```
In [3]: sys.getsizeof(foo)
Out[3]: 48

In [4]: sys.getsizeof(fooslots)
Out[4]: 40
```

pfackeldey:
> ```
> In [3]: sys.getsizeof(foo)
> Out[3]: 48
> 
> In [4]: sys.getsizeof(fooslots)
> Out[4]: 40
> ```

I'll give `__slots__` another try and see if it matters for memory. :+1:

edit: didn't improve noticeable

PR #3360, opened 2025-01-07
chore: remove duplicate code

pfackeldey:
These two properties are already defined in the RecordMeta base class.


ianna (approved):
@pfackeldey - Great! Thanks!

PR #3361, opened 2025-01-07
fix: nbytes repr for typetracers with nbytes=unknown_length

pfackeldey:
This was an oversight from https://github.com/scikit-hep/awkward/pull/3348

The `nbytes` property of a typetracer can be the `unknown_length` sentinel, which doesn't tell us anything about the number of bytes of the array. Thus, we can't add it to the repr.

### DIFF
--- a/src/awkward/prettyprint.py
+++ b/src/awkward/prettyprint.py
@@ -11,6 +11,7 @@ import awkward as ak
 from awkward._layout import wrap_layout
 from awkward._namedaxis import _prettify_named_axes
 from awkward._nplikes.numpy import Numpy, NumpyMetadata
+from awkward._nplikes.shape import unknown_length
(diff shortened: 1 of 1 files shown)


agoose77:
@pfackeldey given that we're allowing reprs to be "non-touching" (in the sense that we don't error if a repr happens after column optimisation, where there are placeholders), I think this can probably say "nbytes: unknown".

pfackeldey:
> @pfackeldey given that we're allowing reprs to be "non-touching" (in the sense that we don't error if a repr happens after column optimisation, where there are placeholders), I think this can probably say "nbytes: unknown".

That's a very good point! I'll update the PR accordingly 👍

ianna (approved):
@pfackeldey - Thanks for fixing it promptly! Looks good to me. Please, merge it if you finished with it. Thanks!

PR #3368, opened 2025-01-14
fix: `_getitem_at_placeholder` checks

pfackeldey:
Some implementations of `_getitem_at_placeholder` have been wrong, e.g. `Index` classes can't be instances of `PlaceholderArrays` - only their `.data` can be.

Also a `ListoffsetArray` should return true if either its offsets or its contents contain PlaceholderArrays. Now that I think about it, that may be wrong from my side and the check should be only for the offsets (as originally intended). Let me know if this is the case, then I update this PR.

### DIFF
--- a/src/awkward/contents/indexedarray.py
+++ b/src/awkward/contents/indexedarray.py
@@ -289,7 +289,7 @@ class IndexedArray(IndexedMeta[Content], Content):
         return self._content._getitem_range(0, 0)
 
     def _is_getitem_at_placeholder(self) -> bool:
-        if isinstance(self._index, PlaceholderArray):
+        if isinstance(self._index.data, PlaceholderArray):
             return True
         return self._content._is_getitem_at_placeholder()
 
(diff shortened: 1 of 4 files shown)


jpivarski (approved):
I agree 100%—only an Index's `data` can be a Placeholder, not an Index. Might this have been responsible for some spurious cases of trying to access PlaceholderArrays at runtime? (I haven't followed the logic through, but it seems like it might be related.)

pfackeldey:
> I agree 100%—only an Index's `data` can be a Placeholder, not an Index. Might this have been responsible for some spurious cases of trying to access PlaceholderArrays at runtime? (I haven't followed the logic through, but it seems like it might be related.)

Unfortunately not likely, unless a user manages to make a `repr` call be part of the dask graph with dask-awkward - which I doubt. Executing the DAG is non-interactive and there's not really a way to interact with the "materialized" awkward Array during a `.compute()` phase. `_getitem_at_placeholder()` is only used for constructing the `repr` of an awkward Array that may contain `PlaceholderArrays` as far as I understand. 

Without this PR the `repr` of awkward Arrays that contain `ListoffsetArrays` (and some others that have `Index` instances on them) would lead to an error; I stumbled over this while playing around with https://github.com/dask-contrib/dask-awkward/pull/565.

ianna (approved):
@pfackeldey - Great, thanks! Please, merge it if you finished with it. Thanks!

PR #3369, opened 2025-01-15
docs: add a performance tip to `ak.transform` docstring

pfackeldey:
As discussed in https://github.com/scikit-hep/vector/pull/554.

### DIFF
--- a/src/awkward/operations/ak_transform.py
+++ b/src/awkward/operations/ak_transform.py
@@ -426,7 +426,7 @@ def transform(
         The output arrays will not be given parameters.
 
 
-    Performance Tipp
+    Performance Tip
     ================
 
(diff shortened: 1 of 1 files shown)


src/awkward/operations/ak_transform.py
```diff
@@ -425,6 +425,40 @@ def transform(
     `"none"`
         The output arrays will not be given parameters.
 
+
+    Performance Tipp
```
agoose77:
```suggestion
    Performance Tip
```

pfackeldey: Thanks! Good to have feedback from a native speaker 👍

ianna (approved):
@pfackeldey - Thanks! Looks good to me.

PR #3371, opened 2025-01-16
feat: Attempt to add awkward's own low-level DAG at the buffer level

pfackeldey:
This is WIP, don't merge!

### DIFF
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -13,6 +13,7 @@ from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpy_like import NumpyMetadata
 from awkward._nplikes.typetracer import try_touch_data
 from awkward._typing import Protocol, TypeAlias
+from awkward.tracing.typetracer import trace_typetracer_method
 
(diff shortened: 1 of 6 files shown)


pfackeldey:
This attempt has several problems:
1. lifetimes of typetracer arrays and _actual_ arrays are different -> we can't connect the DAG nodes based on the `id` of a typetracer array at runtime with actual data.
2. we'd have to either convert this custom DAG to dask and make sure it will never be run distributed (it's way too large!), or have to write our own interpreter that executes operations one by one. This latter approach has the problem of the need to have a dedicated memory management to delete intermediate arrays once they are not needed for the output anymore.
3. we're skipping any metadata python code with this approach, which can be problematic if one wants to repeat the same computation, but e.g. the high-level Array's `.attrs` would be modified differently. This is a conceptual problem.

The typetracer interface is currently very much built to recognize the data that it will need, not to track the operations that are performed on each typetracer array - unfortunately this is not trivial to add. This is also a _big_ difference to e.g. JAX's tracing.

That's why I'm going to close this PR. It'll be in the history of closed PRs for future reference.

PR #3372, opened 2025-01-16
fix: resulting __setitem__ layout type for ak.Record

### LINKED ISSUE #3370: ak.Record breaks after adding a field
### Version of Awkward Array

2.7.2

I noticed I get issues with an ak.Record when assigning a new field, like so:

```
>>> import awkward as ak
>>> x = ak.Record({'hello':'world'})
>>> x['test'] = 'test'
>>> print(x)
  File ".../awkward/highlevel.py", line 2266, in __str__
    return prettyprint_valuestr(self, 1, 80)
  File ".../awkward/prettyprint.py", line 356, in valuestr
    _, strs = valuestr_horiz(data, limit_cols, formatter)
  File ".../awkward/prettyprint.py", line 213, in valuestr_horiz
    fields = data.fields
  File ".../awkward/highlevel.py", line 2200, in __getattr__
    return super().__getattribute__(where)
  File ".../awkward/highlevel.py", line 2024, in fields
    return self._layout.array.fields
AttributeError: 'RecordArray' object has no attribute 'array'
```

I would expect this to work without errors, and a new field inserted into the Record. It does work with `ak.with_field`, however this seems to convert the type to an ak.Array (which I would like to avoid).

Let me know if I can help further, I just found this library but love it so far!

pfackeldey:
Fixes: https://github.com/scikit-hep/awkward/issues/3370

`ak.Record`'s `.layout` property is expected to be an instance of `ak.record.Record`. `ak.with_field` returns a `ak.content.RecordArray` though. A `ak.record.Record`'s `.array` property points to this content class, thus we need to set `.array` and not `.layout` to the return value of `ak.with_field`.

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -2144,8 +2144,8 @@ class Record(NDArrayOperatorsMixin):
             ):
                 raise TypeError("only fields may be assigned in-place (by field name)")
 
-            self._layout = ak.operations.ak_with_field._impl(
-                self._layout,
+            self._layout._array = ak.operations.ak_with_field._impl(
+                self._layout.array,
                 what,
                 where,
                 highlevel=False,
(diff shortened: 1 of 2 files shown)


ianna (approved):
@pfackeldey - thanks for fixing the bug so quickly!

PR #3374, opened 2025-01-17
refactor: propagate attrs & behavior with dict-like `ak.Array` input data

### LINKED ISSUE #3346: Named axes in nested awkward arrays

image
Axis names are lost when an awkward array with named axes is nested inside another awkward array. (Example attached)

Can awkward array preserve the named axes?

pfackeldey:
Currently `ak.Array` (with `dict` input data) and `ak.zip` differ in the following ways:
1. `ak.zip` does broadcasting, `ak.Array` does not (it errors instead)
2. `ak.zip` does propagate `attrs` and `behavior` from all (nested) arrays in the given dictionary, `ak.Array` does not

For `1.` this PR improves the error message for `ak.Array` if it encounters different length inputs to give a hint to try `ak.zip` instead (if broadcasting is a suitable solution).

The second difference does not exist anymore with this PR, `ak.Array` will now also propagate `attrs` and `behavior` in the same way `ak.zip` does. I think it is a bit unintuitive that `ak.zip` and `ak.Array` currently behave different for propagating `attrs` and `behavior`.

I'll add a `policy` label, because it might be good to hear some input/feedback from @ianna, @jpivarski and @agoose77 about this change. 

(Closes: https://github.com/scikit-hep/awkward/issues/3346)

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -302,19 +302,22 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
 
         elif isinstance(data, dict):
             fields = []
-            contents = []
+            _arrays = []
             length = None
             for k, v in data.items():
                 fields.append(k)
-                contents.append(Array(v).layout)
+                _arrays.append(Array(v))
                 if length is None:
-                    length = contents[-1].length
-                elif length != contents[-1].length:
+                    length = _arrays[-1].layout.length
+                elif length != _arrays[-1].layout.length:
                     raise ValueError(
                         "dict of arrays in ak.Array constructor must have arrays "
-                        f"of equal length ({length} vs {contents[-1].length})"
+                        f"of equal length ({length} vs {_arrays[-1].layout.length}). "
+                        "For automatic broadcasting use 'ak.zip' instead. "
                     )
-            layout = ak.contents.RecordArray(contents, fields)
+            layout = ak.contents.RecordArray([arr.layout for arr in _arrays], fields)
+            attrs = attrs_of(*_arrays, attrs=attrs)
+            behavior = behavior_of(*_arrays, behavior=behavior)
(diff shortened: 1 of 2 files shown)


src/awkward/highlevel.py
```diff
@@ -302,19 +302,22 @@ def __init__(
...
+                elif length != _arrays[-1].layout.length:
                     raise ValueError(
                         "dict of arrays in ak.Array constructor must have arrays "
-                        f"of equal length ({length} vs {contents[-1].length})"
+                        f"of equal length ({length} vs {_arrays[-1].layout.length}). "
+                        "For automatic broadcasting use 'ak.zip' instead. "
                     )
-            layout = ak.contents.RecordArray(contents, fields)
+            layout = ak.contents.RecordArray([arr.layout for arr in _arrays], fields)
+            attrs = attrs_of(*_arrays, attrs=attrs)
```
ianna, line 319: @pfackeldey - would this overwrite the `attrs` and `behavior` parameters of `__init__` if they are not `None`?

ianna (commented):
Thanks! I think, a test would be nice to have along with these changes :-)

src/awkward/highlevel.py
pfackeldey: Yes, but only if `_arrays` would have non-`None` `attrs` and `behavior` defined on them. This is what I mean with propagation of metadata. `ak.zip` does this aswell. I hope that makes sense?

pfackeldey: whatever is provided in the `__init__` for attrs and behavior will take precedence. In case they're `None` it will try to propagate them from the given input arrays. (And that's also how `ak.zip` works.)

ianna (approved):
@pfackeldey - Thanks! please, merge it.

PR #3378, opened 2025-01-23
fix: placeholder array repr with extra info

pfackeldey:
This PR fixes the repr of PlaceholderArrays for good by moving the `array_str` into the array module (nplike).

Currently the repr could be printed if no extra information is present, however, when a `parameter` exists (which is typically the case for e.g. `coffea.NanoEvents` that includes the `__doc__` of a TBranch as a parameter) the repr was broken.

This should also be a more cleaner approach as it avoids runtime type dependent if conditions in certain parts of the codebase.

### DIFF
--- a/src/awkward/contents/numpyarray.py
+++ b/src/awkward/contents/numpyarray.py
@@ -258,12 +258,9 @@ class NumpyArray(NumpyMeta, Content):
 
         extra = self._repr_extra(indent + "    ")
 
-        if isinstance(self._data, (TypeTracerArray, PlaceholderArray)):
-            arraystr_lines = ["[## ... ##]"]
-        else:
-            arraystr_lines = self._backend.nplike.array_str(
-                self._data, max_line_width=30
-            ).split("\n")
+        arraystr_lines = self._backend.nplike.array_str(
+            self._data, max_line_width=30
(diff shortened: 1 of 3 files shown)


src/awkward/contents/numpyarray.py
```diff
@@ -258,12 +258,9 @@ def _repr(self, indent, pre, post):
 
         extra = self._repr_extra(indent + "    ")
 
-        if isinstance(self._data, (TypeTracerArray, PlaceholderArray)):
```
ianna, line 261: Why removing `TypeTracerArray` is needed?

ianna (commented):
@pfackeldey - I'm a bit confused about `TypeTracerArray`. Do we support it?

src/awkward/contents/numpyarray.py
pfackeldey: The typetracer `nplike` does define this already: https://github.com/scikit-hep/awkward/blob/main/src/awkward/_nplikes/typetracer.py#L1638-L1648
This if-condition was a redundant implementation for `TypeTracerArray`.

ianna (approved):
@pfackeldey - Thanks! I'm enabling auto-merge.

PR #3364, opened 2025-01-09
feat: implement virtual arrays

ikrommyd:
# Implement Virtual Buffers in Awkward Array  

This PR introduces the `VirtualArray` class, which enables virtual buffers at the lowest level of an Awkward Array's structure. Awkward Arrays are fundamentally tree-like, with data stored in 1D array buffers (e.g., NumPy, CuPy, or JAX). The `VirtualArray` class acts as a virtual buffer, allowing deferred materialization of data when needed.

## Key Features  
- **Lazy Data Generation**: A `VirtualArray` requires a known `dtype` and `shape` upon instantiation, along with a generator function responsible for producing the actual buffer data. This generator function is designed for use cases where materialization is expensive, such as disk reads.  
- **NPLike Consistency**: Each `VirtualArray` is tied to a specific `nplike` (NumPy or CuPy). Switching between different `nplike` backends is not allowed unless the array is first materialized.  
- **Optimized Materialization**: Some operations, such as trivial slicing, can be performed without triggering materialization, preserving the virtual nature of the array. Other operations that inherently require data access will trigger materialization on demand.  

## Adjustments to the Codebase  
- **Array Operations**: Updates to `array_module` ensure that certain operations remain virtual when possible while others materialize data when necessary.  
- **Kernel Execution**: Awkward Array kernels now enforce materialization before processing virtual arrays.  
- **Conversion Functions**: Functions of the form `ak.to_*` are updated to ensure that virtual arrays are materialized before conversion.  
- **Layout Enhancements**: Additional helper methods are introduced in each layout to efficiently check whether an array contains any virtual buffers.  

This should be reviewed in combination with https://github.com/ikrommyd/awkward/pull/1 which runs the whole awkward ci by creating a virtual array inside every `NumpyArray` and every `Index` and materializing them on the spot and also in combination with this branch: https://github.com/ikrommyd/awkward/tree/test-virtual-arrays-without-materialize where most of the tests pass without even needing to materialize inside the `NumpyArray`. You can compare this branch with the branch from https://github.com/ikrommyd/awkward/pull/1 for the changes.

PR #3379, opened 2025-01-23
fix: set allow_lazy to True to avoid touching potential placeholders

### LINKED ISSUE #1231: `PlaceholderArray supports only trivial slices, not ndarray` hit with vector addition after field addition to Jets
This code:
```python
import awkward as ak
import dask
from coffea.nanoevents import NanoEventsFactory

jet_field = "Jet"
delta_r_val = 0.3

events = NanoEventsFactory.from_root(
    {"../coffea/tests/samples/nano_dy.root": "Events"},
    delayed=True,
).events()

events_with_jet = events[ak.num(events[jet_field]) > 0]
events_with_jet[jet_field, "Delta_r_bb"] = events_with_jet[jet_field].metric_table(events_with_jet.GenPart)
nearest_jet = events_with_jet[jet_field][ak.argmin(events_with_jet[jet_field].Delta_r_bb, axis=1)]
dask.compute(nearest_jet + events_with_jet.Muon.sum(axis=1))[0]
```
fails with
```
TypeError: PlaceholderArray supports only trivial slices, not ndarray

    numpy.add.__call__(
        repr-raised-TypeError
        <CandidateArray [{x: 0, y: 0, z: 0, ...}, ...] type='40 * Candidate...'>
    )
```
switching to `delayed=False` works fine. Also not masking events and doing `events_with_jet = events` or not attaching `Delta_r_bb` to the Jets and jeeping it as a separate variable solve the problem.

pfackeldey:
(Finally) fixes: https://github.com/scikit-hep/coffea/issues/1231. This issue has causing me multiple headaches.

This PR is a follow-up of https://github.com/scikit-hep/awkward/pull/2524 for `IndexedOptionArrays`. The problem here is that a `RecordArray` may be wrapped in a `IndexedOptionArray` after a `ak.with_field/__setitem__` operation. This has been fixed for `IndexedArrays` in https://github.com/scikit-hep/awkward/pull/2524, but not for option types.

By setting `allow_lazy=True` in the `_carry` function we don't apply a slice to each content of a `RecordArray`, which causes errors for `PlaceholderArrays`. Instead this wraps them in an `IndexedArray` and thus follows the fix provided in https://github.com/scikit-hep/awkward/pull/2524 by "pushing" the `IndexedArray` down into each field of the `RecordArray` in the next recursion during broadcasting.

Adding the same "pushing" fix to `IndexedOptionArrays` led to multiple issues, which is why I gave up on it. If you have an idea how to implement this, I can give it another try. I'm not 100% sure if my/this solution is what we want.

(PS: I _think_, setting `allow_lazy=True` here should also benefit in the scope of the current `VirtualArray` implementation (https://github.com/scikit-hep/awkward/pull/3364) as it delays materialization of slicing `RecordArrays`. The same is likely true for `IndexedArrays`?)

### DIFF
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -588,7 +588,7 @@ class IndexedOptionArray(IndexedOptionMeta[Content], Content):
                 )
             )
 
-            return self._content._carry(nextcarry, False)
+            return self._content._carry(nextcarry, True)
 
     def _offsets_and_flattened(self, axis: int, depth: int) -> tuple[Index, Content]:
         posaxis = maybe_posaxis(self, axis, depth)


pfackeldey:
(sorry for pinging the three of you, but I feel only partly convinced that this is the correct fix. Given that this has been fixed in https://github.com/scikit-hep/awkward/pull/2524 I asked for a review from all participating parties of this PR.)

ianna (approved):
@pfackeldey - Thanks! I agree with your arguments. Also, if it fixes coffea, we shoud get it out asap. Thanks!

agoose77:
The obvious consequence of this change is that operations that expect `project()` to return a particular type no longer have this guarantee. I'd want to check uses of `project()` internally to confirm the impact. 

I don't *think* the performance for non-dask cases will be hugely affected, which is nice; we're already creating the index buffer that is being used, so it's not exactly more memory. It does mean that `carry`ing can have multiple times, though, if successive operations operate on this layout. In any case, I think a trend towards laziness is probably a good move.

Jim will have thoughts; we have tweaked these defaults for other node types before.

jpivarski:
The application of `allow_lazy` in the various places where `carry` is called can lead to infinite recursion errors if you're not careful. If this were an `IndexedArray.project` passing `allow_lazy=True`, it would definitely cause infinite recursion because the two-level IndexedArray + X structure wouldn't get unpacked to X, it would remain IndexedArray + X, and the broadcasting code that's trying to unpack it would probably call it over and over and never get anywhere. But this is `IndexedOptionArray.project`, so IndexedOptionArray + X first gets unpacked to IndexedArray + X, and then another step unpacks it to just X. That's apparently fine because the tests passed.

@agoose77, I don't think that there's any code that expects the output of `project` to have a particular layout/form, just a particular type (the same type for an IndexedArray or the type without "option" for an IndexedOptionArray). That is, unless you count the issue I described in the previous paragraph, in which broadcasting code expects to make a set of `inputs` simpler (i.e. "unpacked") in each descending step. We _do_ assume that calling `project` doesn't give you exactly the same layout/form you had before, but a simpler one. However, that's preserved by this code change.

This is motivated by correctness, not performance, right? Getting these `allow_lazy` arguments right was tricky and had long-range consequences (though they've always been revealed by the tests, and while Uproot is often losing tests due to data servers going down, Awkward Array isn't). If the motivation is performance, or at least if it's anything less than a factor of 5 or 10 or so, I'd be nervous about accepting it right away. If it's breaking code with PlaceholderArrays, then that's a good argument to take this step.

jpivarski (approved):
Oh, the above was a review. I expressed a conditional, but I'm going to press "approve" here.

pfackeldey:
> This is motivated by correctness, not performance, right?

This PR is purely motivated by correctness. (I don't have a good feeling what implications this would have for performance.)
If you have a different idea how to achieve this correctness I can also give it a try, but I couldn't come up with anything so far.

pfackeldey:
As this PR touches a critical part of awkward's internals, I run in addition (locally) the test suites of `vector`, `dask-awkward`, and `coffea` against this branch and got:

- vector: ✅ 
- dask-awkward: ✅ 
- coffea: ⛔ 

coffea needs https://github.com/scikit-hep/awkward/pull/3383 and #3384 to go in, and then a little tweak in one of coffea's test to adopt to the new `.mask` usage (PR is here: https://github.com/scikit-hep/coffea/pull/1253) with weak-refs as introduced by https://github.com/scikit-hep/awkward/pull/3347. Then coffea's tests pass aswell.

_This_ PR is not responsible for the failing test in coffea, so I think we can go ahead and merge this one.

PR #3380, opened 2025-01-23
chore: make the 'field_path' argument of placeholder arrays more robust

pfackeldey:
This improves the robustness of the `field_path` argument of a `PlaceholderArray`, by 
1. ensuring it's a tuple (this makes sure that explicit usage of this feature has the correct type, e.g. when used in dask-awkward in the future)
2. ensuring all elements are strings before they're joined (otherwise this causes issues if e.g. `None` or an `int` is encountered)

### DIFF
--- a/src/awkward/_nplikes/placeholder.py
+++ b/src/awkward/_nplikes/placeholder.py
@@ -22,13 +22,13 @@ class PlaceholderArray(ArrayLike):
         nplike: NumpyLike,
         shape: tuple[ShapeItem, ...],
         dtype: DType,
-        field_path: tuple[str, ...] = (),
+        field_path: tuple = (),
     ):
         self._nplike = nplike
         self._shape = shape
         self._dtype = np.dtype(dtype)
         if not isinstance(field_path, tuple):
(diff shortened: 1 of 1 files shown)


src/awkward/_nplikes/placeholder.py
```diff
@@ -22,16 +22,18 @@ def __init__(
...
         self._shape = shape
         self._dtype = np.dtype(dtype)
+        if not isinstance(field_path, tuple):
+            raise TypeError(f"field_path must be a tuple, not {field_path}")
         self._field_path = field_path
 
     @property
     def field_path(self) -> str:
-        return ".".join(self._field_path)
+        return ".".join(map(str, self._field_path))
```
agoose77, line 36: Given that this is an internal type, I'd be more in favour of fixing the call-sites, or adding an assert in the constructor, to ensure that there are not Nones here?

pfackeldey: That's a good argument. It will be used likely only once in dask-awkward, and I can make sure that it will be correct there. I'll close this PR then :+1:

PR #3381, opened 2025-01-23
ci: use native arm for wheels

henryiii:
Faster, simpler.


ianna (approved):
@henryiii  - fantastic! Thank you!

PR #3383, opened 2025-01-23
refactor: simplify attrs & fix its pickability (for e.g. dask.distributed)

pfackeldey:
This PR makes `.attrs` properly pickable again by getting rid of the weak reference (without loosing the intended behavior introduced in https://github.com/scikit-hep/awkward/pull/3344).

Without this PR we'll break any usage with dask.distributed as it needs to pickle all objects (including `.attrs`) for scheduling.
Thus, we should get this PR in before a new awkward-array release.

(and also I think it's much cleaner now :-) )


ianna (approved):
@pfackeldey - Thanks, it looks great! Indeed, the code is cleaner.

PR #3384, opened 2025-01-23
fix: remove unsafe lru cache

pfackeldey:
It turns out (from one coffea test) that adding this cache was unsafe -> this PR removes it.


ianna (approved):
@pfackeldey - thanks for fixing it! I'll enable auto merge.

PR #3386, opened 2025-01-24
chore: check if 32bit right broadcasting test pass again

pfackeldey:
This PR is only opened to see if the 32bit fix in https://github.com/scikit-hep/awkward/pull/3270 is related in any way to the marked test failure for 32bit windows for right broadcasting.

~If the tests pass now, I'll update this PR, so please don't merge!~

PR  https://github.com/scikit-hep/awkward/pull/3270 fixed it!

### DIFF
--- a/tests/test_2596_named_axis.py
+++ b/tests/test_2596_named_axis.py
@@ -373,11 +373,6 @@ def test_negative_named_axis_indexing():
     )
 
 
-# @pytest.mark.xfail(
-#     sys.platform == "win32",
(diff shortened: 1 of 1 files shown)


henryiii (approved):
Passes tests, so seems fine to me! A little surprised this would have affected windows, since we left the windows part alone, I thought.

pfackeldey:
Hm, it was the only thing that we changed related to 32bit I _think_. Anyway, the test passes again, so it seems to work :tada:

ianna (approved):
@pfackeldey - Thanks! The tests pass. Please merge it. Thank you.

PR #3387, opened 2025-01-24
chore: reducing warnings

henryiii:
Reducing a few more warnings:

* Updated rapidjson to the latest commit, which removes all the "deprecated in C++17" warnings. There hasn't been a release of rapidjson in many, many years.
* Dropped a useless variable. If it was supposed to be used, then that should probably be fixed instead.
* Made it possible to compile directly, without scikit-build-core. Makes it easier to run clang-tidy, for example.

There are a bunch of signed -> unsigned conversions happening in one file which I did not fix yet.


ianna (approved):
@henryiii - looks good to me. Thanks!

Issue #3305, opened 2024-11-14
Intermittent failures in deploying the documentation preview


jpivarski:
For example, https://github.com/scikit-hep/awkward/actions/runs/11783736057/job/32997893500?pr=3302

```
Run aws-actions/configure-aws-credentials@v4
  with:
    aws-region: eu-west-[2](https://github.com/scikit-hep/awkward/actions/runs/11783736057/job/32997893500?pr=3302#step:3:2)
    role-to-assume: arn:aws:iam:::role/
    audience: sts.amazonaws.com
  env:
    X86_64_PYTHON_VERSION: [3](https://github.com/scikit-hep/awkward/actions/runs/11783736057/job/32997893500?pr=3302#step:3:3).11.0
    SOURCE_DATE_EPOCH: 1668811211
    S3_BUCKET: preview.awkward-array.org
    DEPLOY_URL: http://preview.awkward-array.org.s3-website.us-east-1.amazonaws.com
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Assuming role with OIDC
Error: Could not assume role with OIDC: 1 validation error detected: Value 'arn:aws:iam:::role/' at 'roleArn' failed to satisfy constraint: Member must have length greater than or equal to [20](https://github.com/scikit-hep/awkward/actions/runs/11783736057/job/32997893500?pr=3302#step:3:21)
```

But this doesn't always happen. @henryiii asked why we're not using ReadTheDocs—we no longer have build times that exceed ReadTheDocs's 10 minutes (it's more like 1 minute and we're not adding new compiled code). @agoose77, what was the motivation? (I think this was set up _after_ you made awkward-cpp builds fast.)

agoose77:
@jpivarski we don't use RTD because build & execution of tutorials takes a while. Our builds are usually fast on GHA because we have caching of the C++.

I haven't done a clean-cache build + clean-cache execution, but I assume it's still somewhat slow.

I don't know what causes that error, but I think it's the case that OIDC fails to issue a token sometimes. I haven't looked yet. Will try to take a peek tomorrow?

jpivarski:
Ah, it's those notebooks! Yes, they do take a while to run from the Jupytext to produce all their outputs.

[...]

PR #3391, opened 2025-01-31
test: debug AWS role ARN

ianna:


ianna:
I forget, it's intermittent and mentioned here: https://github.com/scikit-hep/awkward/issues/3305

PR #3396, opened 2025-02-07
fix: mark awkward as freedthread safe with pybind11

### LINKED ISSUE #3395: awkward-cpp needs to declare that it can be run freethreaded!

Right now when running in python 3.13t:

```
(coffea-freethreaded) lgray@MacBookPro dask-awkward % python
Python 3.13.1 experimental free-threading build | packaged by conda-forge | (main, Jan 13 2025, 09:43:07) [Clang 18.1.8 ] on darwin
>>> import awkward
<frozen importlib._bootstrap>:488: RuntimeWarning: The global interpreter lock (GIL) has been enabled to load module 'awkward_cpp.lib._ext', which has not declared that it can run safely without the GIL. To override this behavior and keep the GIL disabled (at your own risk), run with PYTHON_GIL=0 or -Xgil=0.
```

It would be nice to both verify that we can declare this and then also enable this functionality so it's easier for users to get ripping in freethreaded python.

lgray:
Fixes: #3395 

Need to create some tests that demonstrate this is, in fact, the case.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,7 +72,11 @@ jobs:
           - python-version: '3.13t'
             python-architecture: x64
             runs-on: ubuntu-latest
-            dependencies-kind: full
+            dependencies-kind: nogil
+          - python-version: '3.13t'
+            python-architecture: x64
+            runs-on: windows-latest
+            dependencies-kind: nogil
 
     runs-on: ${{ matrix.runs-on }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,7 +72,7 @@ jobs:
           - python-version: '3.13t'
             python-architecture: x64
             runs-on: ubuntu-latest
-            dependencies-kind: minimal
+            dependencies-kind: full
 
     runs-on: ${{ matrix.runs-on }}
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,6 +39,7 @@ jobs:
           - ubuntu-latest
           - macos-13
         python-version:
+          - '3.13t'
           - '3.13'
           - '3.12'
           - '3.11'
@@ -81,7 +82,7 @@ jobs:
           submodules: true
 
       - name: 'Python ${{ matrix.python-version }}'
-        uses: actions/setup-python@v5
+        uses: Quansight-Labs/setup-python@v5
         with:
           python-version: '${{ matrix.python-version }}'
           architecture: '${{ matrix.python-architecture }}'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -39,7 +39,6 @@ jobs:
           - ubuntu-latest
           - macos-13
         python-version:
-          - '3.13t'
           - '3.13'
           - '3.12'
           - '3.11'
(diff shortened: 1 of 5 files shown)


lgray:
My understanding of the C component of awkward array is that it is already reentrant and carries no state.
So I think this is safe to do without adding tests aside from the usual tests in python 3.13t (which is soon to be available through actions).

I am thinking about the python side of awkward and there are not many situations where you'd be manipulating array metadata in a reentrant way, no structure we can make is large enough to need it. Typically multithreaded work on awkward arrays is going to be at the numpy level.

The only place where there might be some issues is with typetracers, but this is a fairly ill-defined worry. 

My gut feeling is that we don't really need to do too much to merge this.

lgray:
Ah right this will be blocked by cramjam freethreading then.

c.f.: https://github.com/milesgranger/cramjam/pull/200

.github/workflows/test.yml
```diff
@@ -81,7 +82,7 @@ jobs:
           submodules: true
 
       - name: 'Python ${{ matrix.python-version }}'
-        uses: actions/setup-python@v5
+        uses: Quansight-Labs/setup-python@v5
```
ianna, line 92: @lgray - do we need any Quansight-specific modifications?

ianna (commented):
@ariostas - could you, please, check the failing workflows. Is there anything we can do on our side? Thanks!

.github/workflows/test.yml
```diff
@@ -39,6 +39,7 @@ jobs:
           - ubuntu-latest
           - macos-13
         python-version:
+          - '3.13t'
```
ariostas: Instead of adding 3.13t here, we could include
```
- python-version: '3.13t'
  python-architecture: x64
  runs-on: ubuntu-latest
  dependencies-kind: minimal
```
in the matrix for now. The minimal test dependencies don't include Uproot, so it should work fine.

.github/workflows/test.yml
ariostas: The "official" action doesn't support 3.13t yet, but they say that it's coming soon.

ariostas:
Oh, now I'm realizing that minimal probably meant "oldest supported versions" instead of "the minimal set to run tests"

lgray:
We probably want this to run over all platforms for 3.13t as well since linux and windows have different threading models.

ariostas:
I can't think of a simpler option than to add another requirements file for the free-threaded tests for now. Apart from that, there's a couple more things that need to be done.

1. The `pipx run nox -s prepare -- --headers --signatures --tests` step fails because `pyyaml` doesn't have full support for free-threading (but the fix is easy).
2. We need to make sure that 3.13t wheels are built and distributed by adding this to the `awkward-cpp` `pyproject.toml`.
```
[tool.cibuildwheel]
free-threaded-support = true
```

I ran the tests locally and there are some that fail, but it seems that it's because free-threaded PyArrow is broken.

ariostas:
I added a few more commits to fix the things I mentioned in my last comment. Hopefully everything passes. (Feel free to remove or modify them)

There are a couple of unrelated things that I noticed.

1. We are not fully testing things in 3.13 since there are some caps that can probably be relaxed now.
https://github.com/scikit-hep/awkward/blob/b82566d90bf32e6b78588b3b5e55c4f9ff1414ae/requirements-test-full.txt#L1-L10

2. We are not testing for macos-arm64. There should still be tests that run when building the wheels, but it might be good to also test that in the normal test workflow.

ianna (approved):
@ariostas - looks great! Thank you! All test pass.

PR #3397, opened 2025-02-07
fix: structure removal order for unions

### LINKED ISSUE #3180: fill_none followed by ravel reorders arrays
### Version of Awkward Array

2.6.6

When I apply `fill_none` and then `ravel` (or `flatten`), the resulting array has been reordered. In the following snippet, I expect the result to be `ak.Array([False,False,True])`, but instead it is `ak.Array([False,True,False])`.

```python
import awkward as ak

x = ak.Array([None,[False],[True]])
ak.ravel(ak.fill_none(x,False,axis=0))
```
returns
```
[False,
 True,
 False]
--------------
type: 3 * bool
```

Apologies if this is expected, and is instead my own misunderstanding. Thank you!

pfackeldey:
Fixes: #3180

### DIFF
--- a/src/awkward/_nplikes/numpy.py
+++ b/src/awkward/_nplikes/numpy.py
@@ -36,10 +36,6 @@ class Numpy(ArrayModuleNumpyLike["NDArray"]):
     def ndarray(self):
         return self._module.ndarray
 
-    @property
-    def atleast_1d(self):
-        return self._module.atleast_1d
-
     @classmethod
     def is_own_array_type(cls, type_: type) -> bool:
         """
--- a/src/awkward/_nplikes/numpy.py
+++ b/src/awkward/_nplikes/numpy.py
@@ -36,6 +36,10 @@ class Numpy(ArrayModuleNumpyLike["NDArray"]):
     def ndarray(self):
         return self._module.ndarray
 
+    @property
+    def atleast_1d(self):
+        return self._module.atleast_1d
+
     @classmethod
     def is_own_array_type(cls, type_: type) -> bool:
         """
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1574,7 +1574,8 @@ class UnionArray(UnionMeta[Content], Content):
         self, backend: Backend, options: RemoveStructureOptions
     ) -> list[Content]:
         out = []
-        for i in range(len(self._contents)):
+        _, unique_index, *_ = numpy.unique_all(self._tags.data)
+        for i in self._tags.data[numpy.sort(unique_index)]:
             index = self._index[self._tags.data == i]
             out.extend(
                 self._contents[i]
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1574,8 +1574,8 @@ class UnionArray(UnionMeta[Content], Content):
         self, backend: Backend, options: RemoveStructureOptions
     ) -> list[Content]:
         out = []
-        _, unique_index, *_ = numpy.unique_all(self._tags.data)
-        for i in self._tags.data[numpy.sort(unique_index)]:
+        _, unique_index, *_ = self._backend.index_nplike.unique_all(self._tags.data)
+        for i in self._tags.data[self._backend.index_nplike.sort(unique_index)]:
             index = self._index[self._tags.data == i]
             out.extend(
                 self._contents[i]
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1574,14 +1574,25 @@ class UnionArray(UnionMeta[Content], Content):
         self, backend: Backend, options: RemoveStructureOptions
     ) -> list[Content]:
         out = []
-        _, unique_index, *_ = self._backend.index_nplike.unique_all(self._tags.data)
-        for i in self._tags.data[self._backend.index_nplike.sort(unique_index)]:
-            index = self._index[self._tags.data == i]
-            out.extend(
-                self._contents[i]
-                ._carry(index, False)
+
+        tags = numpy.atleast_1d(self._tags.raw(numpy))
+        index = numpy.atleast_1d(self._index.raw(numpy))
+
+        for i in range(len(tags)):
+            tag = tags[i]
+            idx_value = index[i]
+
+            content = (
+                self._contents[tag]
+                ._carry(ak.index.Index(numpy.atleast_1d(idx_value)), False)
                 ._remove_structure(backend, options)
             )
+
+            if isinstance(content, list):
+                out.extend(content)
+            else:
+                out.append(content)
+
         return out
 
     def _recursively_apply(
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1575,24 +1575,22 @@ class UnionArray(UnionMeta[Content], Content):
     ) -> list[Content]:
         out = []
 
-        tags = numpy.atleast_1d(self._tags.raw(numpy))
-        index = numpy.atleast_1d(self._index.raw(numpy))
+        # typetracer
+        if not self._backend.nplike.known_data:
+            self._touch_data(recursive=False)
+            # just flatten, ignore order, tags, index
+            for c in self._contents:
+                out.extend(c._remove_structure(backend, options))
+            return out
 
-        for i in range(len(tags)):
-            tag = tags[i]
-            idx_value = index[i]
-
+        # backends with concrete data
+        for i in range(self._tags.length):
             content = (
-                self._contents[tag]
-                ._carry(ak.index.Index(numpy.atleast_1d(idx_value)), False)
+                self._contents[self._tags.data[i]]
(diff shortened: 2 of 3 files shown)


ianna (approved):
@pfackeldey - Great! Thanks for providing the fix so quickly! Indeed, iterating over the unique index fixes the issue.

src/awkward/contents/unionarray.py
```diff
@@ -1574,7 +1574,8 @@ def _remove_structure(
         self, backend: Backend, options: RemoveStructureOptions
     ) -> list[Content]:
         out = []
-        for i in range(len(self._contents)):
+        _, unique_index, *_ = numpy.unique_all(self._tags.data)
```
ianna: @pfackeldey - should it be Numpy-like? To cover the `cupy` as a backend?

pfackeldey: yes, this is to cover all backends (also typetracer).

pfackeldey:
> @pfackeldey - Great! Thanks for providing the fix so quickly! Indeed, iterating over the unique index fixes the issue.

this PR does however make the assumption that the number of unique tags corresponds to the `len(self._contents)`, which I think should be always true. Could you confirm this is true? If yes, then I'm confident that this is the right fix :+1:

src/awkward/contents/unionarray.py
pfackeldey: oh, wait, you're right. I'm using the global `numpy` nplike. Let me have a look again to use the nplike of the content class... 🤦

agoose77:
@pfackeldey just reading the comments and not the diff -- `len(contents)` may not equal `unique(tags)` if a content is not indexed. I think we have code-paths that can break this invariant, e.g.
```python
import awkward as ak
import numpy as np
x = ak.Array([1, "hi", 2, "bye"])
y = x[[0,2]]
assert np.unique(y.layout.tags).size == len(y.layout.contents)
```

pfackeldey:
> @pfackeldey just reading the comments and not the diff -- `len(contents)` may not equal `unique(tags)` if a content is not indexed. I think we have code-paths that can break this invariant, e.g.
> 
> ```python
> import awkward as ak
> import numpy as np
> x = ak.Array([1, "hi", 2, "bye"])
> y = x[[0,2]]
> assert np.unique(y.layout.tags).size == len(y.layout.contents)
> ```

Ah I see. In this case this implementation is still working correct though:
```python
# with this PR
x = ak.Array([1, "hi", 2, "bye"])
y = x[[0,2]]
ak.ravel(y)
# <Array [1, 2] type='2 * int64'>
```
on main this actually fails, which I would consider wrong aswell (so this PR might fix another bug aswell?):
```python
x = ak.Array([1, "hi", 2, "bye"])
y = x[[0,2]]
ak.ravel(y)
# ... > AssertionError: cannot merge NumpyArray with ListOffsetArray
```
The loop over the unique ordered tags will pick (in order) the contents where tags point to. This is always less or equal to the length of contents. Do you see any problem with that @agoose77 ?

agoose77:
@pfackeldey the original code has two bugs (which are the same kind of bug)! This PR fixes one, but not the other I think.

The intention of `remove_structure` is to allow the caller to obtain a collection of in-order zero-copy (where possible) 1D buffers such that (approximately) `A_ij -> B_k` (in Einstein notation). That's only partially possible here. I think of unions a bit like a flattened matrix, i.e. `B_k` where `B_k1` and `B_k2` may come from different `content`s. As such, a contiguous array of `B_k` will require copying each union item into a contiguous buffer.

For trivial-unions (e.g. a union of integer and bool), the proper way to flatten the layout in an order-preserving way is to write a kernel that fills a new buffer with `result = [contents[tags[i]][i] for i in self.length]`. But, for non-trivial unions that's a lot harder to figure out. For each content `contents[j]`, there may be `N` associated values that end up in the structureless result.

Moreover, as we are dealing with a union, to remove the structure probably means we want to simplify a union over the results such that the union disappears and the result is simplified to a common type.

So, maybe the proper way to do this is to build a new union via `.simplified` over `[content[i]._carry(...)._remove_structure() ...]` and throw an error if the union doesn't vanish.

But, I also feel compelled to say -- `_remove_structure` does a few different things. It might be that there are some contexts where we don't care about the order invariant in the final result (e.g. non-positional reducers). In that case, perhaps we want to accept a flag that lets us take a fast path involving less `carry`ing. Although, even then we would need to ensure that the type of the result is correct (i.e. figure out the result type and cast each `content[i]._carry(...)._remove_structure() ...]` to that type).

ianna:
@pfackeldey and @agoose77 - please, check when you have time. I'm sure, it's not the most efficient implementation - I just want to make sure it is correct and I'm not missing anything. Thanks!

src/awkward/_nplikes/numpy.py
```diff
@@ -36,6 +36,10 @@ def char(self):
     def ndarray(self):
         return self._module.ndarray
 
+    @property
+    def atleast_1d(self):
+        return self._module.atleast_1d
```
pfackeldey: I think this should rather live on the [`ArrayModuleNumpyLike`](https://github.com/scikit-hep/awkward/blob/main/src/awkward/_nplikes/array_module.py#L41) as a method.

src/awkward/contents/unionarray.py
```diff
@@ -1574,13 +1574,25 @@ def _remove_structure(
         self, backend: Backend, options: RemoveStructureOptions
     ) -> list[Content]:
         out = []
-        for i in range(len(self._contents)):
-            index = self._index[self._tags.data == i]
-            out.extend(
-                self._contents[i]
-                ._carry(index, False)
+
+        tags = numpy.atleast_1d(self._tags.raw(numpy))
```
pfackeldey: `self._tags.raw` will try to convert a typetracer to a concrete array, which is not possible. Also, `numpy.atleast_1d` should rather be `self.backend.index_nplike.atleast_1d` and then have a corresponding implementation for typetracers.

src/awkward/contents/unionarray.py
```diff
@@ -1574,13 +1574,25 @@ def _remove_structure(
...
+        index = numpy.atleast_1d(self._index.raw(numpy))
+
+        for i in range(len(tags)):
+            tag = tags[i]
+            idx_value = index[i]
+
+            content = (
+                self._contents[tag]
```
pfackeldey: Hm, I don't think this can ever work with typetracers, because `tag` will be a `unknown_scalar`.

pfackeldey (commented):
I think this is the right direction :+1:, but the code won't be compatible with typetracers as is right now. I'm not sure though how to fix this properly - I'll give this some thoughts...

We should make sure that the following works (this is on main):
```python
import awkward as ak

a = ak.Array([[1, 2, 3], 4, [5, 6], 7, 8], backend="typetracer")
ak.ravel(a)
# <Array-typetracer [...] type='## * int64'>
```

src/awkward/contents/unionarray.py
ianna: it also mean that we cannot do 

`a.tolist()` on a typetracer array:

```python
  File "/Users/yana/Projects/PR3397/awkward/src/awkward/highlevel.py", line 535, in tolist
    return self.to_list()
           ^^^^^^^^^^^^^^
  File "/Users/yana/Projects/PR3397/awkward/src/awkward/highlevel.py", line 541, in to_list
    return self._layout.to_list(self._behavior)
  File "/Users/yana/Projects/PR3397/awkward/src/awkward/contents/content.py", line 1183, in to_list
    return self.to_packed()._to_list(behavior, None)
           ^^^^^^^^^^^^^^^^
  File "/Users/yana/Projects/PR3397/awkward/src/awkward/contents/unionarray.py", line 1671, in to_packed
    if len(contents[tag]) > num_tag:
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/yana/Projects/PR3397/awkward/src/awkward/_nplikes/shape.py", line 76, in __gt__
    raise TypeError("cannot order unknown lengths")
TypeError: cannot order unknown lengths
```

ianna:
@pfackeldey - https://github.com/scikit-hep/awkward/pull/3409 should fix the failing tests. It will be auto-merged soon. This PR will need to be re-based.

ianna:
@pfackeldey - thanks! I'm merging it. We'll do the profiling and then performance improvements in a separate PR.

src/awkward/contents/unionarray.py
```diff
@@ -1574,13 +1574,23 @@ def _remove_structure(
...
+            return out
+
+        # backends with concrete data
+        for i in range(self._tags.length):
+            content = (
+                self._contents[self._tags.data[i]]
+                ._carry(ak.index.Index(self._index.data[i]), False)
                 ._remove_structure(backend, options)
             )
+            out.extend(content)
```
agoose77, line 1593: @pfackeldey this is a correct solution as far as structure goes, although allocates repeatedly according to the tags which is not ideal. 

Would you be able to leave a `# FIXME: ` note here (probably just commit to `main`) that we can optimise this by building a `UnionArray` over the original naive loop to re-order the result? That might not be trivial to do, but we should definitely remark upon it!

PR #3399, opened 2025-02-11
docs: add `ak.zip_no_broadcast` and `ak.with/without_named_axis` to public API docs

ikrommyd:
I just randomly noticed today that `ak.zip_no_broadcast` and `ak.with/without_named_axis` are not part of the public API docs for the users. This PR adds them.


pfackeldey (approved):
Thanks!

ianna (approved):
@ikrommyd - Thanks for spotting and fixing it!

PR #3400, opened 2025-02-13
fix: typo/bug in `to_nplike`

ikrommyd:
`to_nplike` is the function itself. `nplike` is the target nplike.


pfackeldey (approved):
good catch!

agoose77 (approved):
Haha, what an amusing error! Nice catch!

PR #3405, opened 2025-02-24
fix: jax backend tracer leaks

pfackeldey:
This fixes https://github.com/scikit-hep/awkward/pull/3340 properly. 

The issue was: we were using the `layout` as part of the tree definition of a `PyTree`. The `layout` holds a reference to _all_ arrays, also the ones that will be replace by tracers by Jax. To fix this, the (un-)flattening can not make use of `layouts` anymore, instead I'm using all other metadata to (de-)construct an awkward array. We had these tracer leaks all the time, just the newer Jax version (from 0.4.36 onwards) uncovered it.

In addition, `ak.form.Form` is much closer to a `PyTree`-definition than a layout is. 

This was really a hard bug to identify, tracer leaks are no fun!

### DIFF
--- a/src/awkward/_connect/jax/reducers.py
+++ b/src/awkward/_connect/jax/reducers.py
@@ -59,7 +59,7 @@ class ArgMin(JAXReducer):
         shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
-        raise RuntimeError("Cannot differentiate through argmin")
+        raise NotImplementedError()
 
 
 @overloads(_reducers.ArgMax)
@@ -85,7 +85,7 @@ class ArgMax(JAXReducer):
         shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
-        raise RuntimeError("Cannot differentiate through argmax")
+        raise NotImplementedError()
 
 
 @overloads(_reducers.Count)
@@ -146,7 +146,7 @@ class CountNonzero(JAXReducer):
         shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
-        raise RuntimeError("Cannot differentiate through count_nonzero")
+        raise NotImplementedError()
 
 
 @overloads(_reducers.Sum)
--- a/src/awkward/_connect/jax/trees.py
+++ b/src/awkward/_connect/jax/trees.py
@@ -8,12 +8,15 @@ import awkward as ak
 from awkward import contents, highlevel, record
 from awkward._backends.jax import JaxBackend
 from awkward._layout import HighLevelContext
+from awkward._nplikes.numpy_like import NumpyMetadata
 from awkward._typing import Generic, TypeVar, Union
 from awkward.contents import Content
 
 T = TypeVar(
     "T", bound=Union[highlevel.Array, highlevel.Record, contents.Content, record.Record]
 )
+
+np = NumpyMetadata.instance()
 
 
 def split_buffers(buffers: dict) -> tuple[dict, dict]:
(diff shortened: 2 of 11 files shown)


src/awkward/_connect/jax/reducers.py
```diff
@@ -59,7 +59,7 @@ def apply(
         shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
-        raise RuntimeError("Cannot differentiate through argmin")
+        raise NotImplementedError()
```
ianna, line 62: I think, the function argmin (or argmax) is not differentiable because it involves a discrete selection rather than a smooth operation.

src/awkward/_connect/jax/reducers.py
```diff
@@ -85,7 +85,7 @@ def apply(
         shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
-        raise RuntimeError("Cannot differentiate through argmax")
+        raise NotImplementedError()
```
ianna, line 88: same here. I'd leave the runtime error.

src/awkward/_connect/jax/reducers.py
```diff
@@ -146,7 +146,7 @@ def apply(
         shifts: ak.index.Index | None,
         outlength: ShapeItem,
     ) -> ak.contents.NumpyArray:
-        raise RuntimeError("Cannot differentiate through count_nonzero")
+        raise NotImplementedError()
```
ianna, line 149: and this one as well?

src/awkward/_connect/jax/trees.py
```diff
@@ -6,162 +6,114 @@
...
-            self._layout, list(buffers), backend=JaxBackend.instance()
+    def unflatten(self, data_buffers: tuple) -> T:
+        # reconstitute data buffers
+        data_buffers = dict(zip(self._data_buffer_keys, data_buffers))
+
+        # combine data buffers with other buffers
+        buffers = {**self._other_buffers, **data_buffers}
+
+        # reconstruct layout
+        layout = ak.operations.from_buffers(
```
agoose77, line 104: This is a nice change!

src/awkward/_connect/jax/trees.py
```diff
@@ -6,162 +6,114 @@
...
+            highlevel=self._highlevel,
+            allow_other=True,
         )
 
     def __eq__(self, other: AuxData) -> bool:
-        return self.layout.is_equal_to(
-            other.layout, index_dtype=False, numpyarray=False
-        )
+        # should this test for more?
+        return self._form == other._form and self._length == other._length
```
agoose77: Perhaps use `form.is_equal_to`, and pass `all_parameters=True` and `form_key=True`?

src/awkward/_connect/jax/trees.py
```diff
@@ -6,162 +6,114 @@
...
+
+        # reconstruct layout
+        layout = ak.operations.from_buffers(
+            self._form, self._length, buffers, backend="jax"
         )
-        return wrap_layout(
-            layout, behavior=self._behavior, highlevel=self._is_highlevel
+
+        # wrap layout
+        return self._ctx.wrap(
```
agoose77, line 113: Do we need to handle `attrs` too?

src/awkward/_connect/jax/trees.py
pfackeldey: good point! :+1:

src/awkward/_connect/jax/trees.py
pfackeldey: this is automatically handled by `HighLevelContext`, isn't it?

ianna (approved):
@pfackeldey - Looks good to me! Please check some minor comments. Thanks!

Saransh-cpp (commented):
Thanks for fixing and cleaning this up, @pfackeldey! Looks good to me :)

src/awkward/_connect/jax/reducers.py
pfackeldey: yes, but it's also not implemented. This error is hit not only for differentiating through it, it's hit when this reducer is called in any case:
```python
import awkward as ak
ak.jax.register_and_check()

x = ak.Array([[1.0], [2.0], [3.0]], backend="jax")

ak.argmin(x)
>> ... RuntimeError: Cannot differentiate through argmin
```
This is wrong, it's just not implemented (I'm not doing any autograd here). That's why I changed it

src/awkward/_connect/jax/trees.py
pfackeldey: works :+1: (`jax.tree.map` triggers flattening and unflattening)

```python
import jax
import awkward as ak

ak.jax.register_and_check()

x = ak.Array([[1.0], [2.0], [3.0]], backend="jax")
x.attrs["foo"] = "bar"

jax.tree.map(jax.numpy.sqrt, x).attrs
# Attrs({'foo': 'bar'})
```

ianna (approved):
@pfackeldey - Great! Thanks! Please, merge it when you are done.

PR #3410, opened 2025-03-04
fix: maintain tupleness of RecordForms when using _select_columns and _prune_columns

ariostas:
Currently, if you take some form that contains a tuple `RecordForm` and apply `select_columns` it is converted into a non-tuple one by adding field names "1", "2", "3", ... . This PR fixes that, since it doesn't really make sense to do so.

### DIFF
--- a/src/awkward/forms/recordform.py
+++ b/src/awkward/forms/recordform.py
@@ -127,7 +127,9 @@ class RecordForm(RecordMeta[Form], Form):
         if not fields and is_inside_record_or_union:
             return None
         else:
-            return self.copy(contents=contents, fields=fields)
+            return self.copy(
+                contents=contents, fields=None if self.is_tuple else fields
+            )
 
     def _select_columns(self, match_specifier: _SpecifierMatcher) -> Self:
         contents = []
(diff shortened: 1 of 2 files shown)


ianna (commented):
@ariostas - I wonder if it is a feature? Perhaps it is intentional, for example, to keep track of which content index is selected from the original array?

```python
(records, records_tuple) = (
    ak.contents.RecordArray(
        [
            ak.contents.NumpyArray(
                np.array([1.1, 2.2, 3.3]), parameters={"foo": "inner1"}
            ),
            ak.contents.ListOffsetArray(
                ak.index.Index32(np.array([0, 3, 3, 5], dtype=np.int32)),
                ak.contents.NumpyArray(
                    np.array([1.1, 2.2, 3.3, 4.4, 5.5]), parameters={"foo": "inner2"}
                ),
            ),
        ],
        None if is_tuple else ["x", "y"],
        parameters={"foo": "outer"},
    )
    for is_tuple in (False, True)
)
```

Could you, please, add a test?

ariostas:
This is ready for review now

ianna (approved):
@ariostas - Looks great! Thanks. All tests pass, I'm merging it. Thanks!

Issue #3402, opened 2025-02-20
awkward-array doesn't work with pyarrow `19.0.0`


pfackeldey:
### Version of Awkward Array

2.7.4

### Description and code to reproduce

awkward-array doesn't work with pyarrow `19.0.0`, however, it does work with `18` and earlier.

Running the test-suite with version `19`, yields:
```shell
FAILED tests/test_1125_to_arrow_from_arrow.py::test_recordarray[False-False] - pyarrow.lib.ArrowInvalid: Column 'x' is declared non-nullable but contains nulls
FAILED tests/test_1125_to_arrow_from_arrow.py::test_recordarray[True-False] - pyarrow.lib.ArrowInvalid: Column 'x' is declared non-nullable but contains nulls
FAILED tests/test_1125_to_arrow_from_arrow.py::test_recordarray[True-True] - pyarrow.lib.ArrowInvalid: Column '0' is declared non-nullable but contains nulls
FAILED tests/test_1294_to_and_from_parquet.py::test_recordarray[False-through_parquet-True] - pyarrow.lib.ArrowInvalid: Column '0' is declared non-nullable but contains nulls
FAILED tests/test_1294_to_and_from_parquet.py::test_recordarray[True-through_parquet-True] - pyarrow.lib.ArrowInvalid: Column '0' is declared non-nullable but contains nulls
FAILED tests/test_1440_start_v2_to_parquet.py::test_recordarray[True-True] - RuntimeError: file metadata is only available after writer close
```

We should make sure that awkward's dependencies are pinned correctly, and also update the pyarrow version dependency for the test-suite here: https://github.com/scikit-hep/awkward/blob/main/requirements-test-full.txt#L6

bnavigator:
Can confirm the same error on openSUSE with apache-arrow 19.0.1

Arrow reference: https://github.com/apache/arrow/issues/41667 https://github.com/apache/arrow/pull/44921

ianna:
@martindurant - I need some help here. 

The PR #3416 fixes the issue by removing the declaration that the columns are `non-nullable` that contains `nulls`. However the column type changes as well. I'm not sure if that is what we want and if not, how to avoid it. Thanks!

```python
    def parquet_round_trip(
        akarray, extensionarray, tmp_path, categorical_as_dictionary=False
    ):
        filename = os.path.join(tmp_path, "whatever.parquet")
        ak.to_parquet(
            akarray,
            filename,
            extensionarray=extensionarray,
[...]

PR #3416, opened 2025-03-11
fix: pyarrow 19.0 support

ianna:
also enable `datetime` tests

### DIFF
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -1095,8 +1095,7 @@ class RecordArray(RecordMeta[Content], Content):
                 pyarrow.field(
                     self.index_to_field(i),
                     values[i].type,
-                    (mask_node is not None and mask_node._arrow_needs_option_type())
-                    or x._arrow_needs_option_type(),
+                    x._arrow_needs_option_type(),
                 )
                 for i, x in enumerate(self._contents)
             ]
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -1092,8 +1092,13 @@ class RecordArray(RecordMeta[Content], Content):
 
         types = pyarrow.struct(
             [
-                pyarrow.field(self.index_to_field(i), values[i].type).with_nullable(
-                    x._arrow_needs_option_type()
+                pyarrow.field(
+                    self.index_to_field(i),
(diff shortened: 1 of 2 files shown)


src/awkward/contents/recordarray.py
```diff
@@ -1092,8 +1092,13 @@ def _to_arrow(
...
             [
-                pyarrow.field(self.index_to_field(i), values[i].type).with_nullable(
-                    x._arrow_needs_option_type()
+                pyarrow.field(
+                    self.index_to_field(i),
+                    values[i].type,
+                    ak._connect.pyarrow.to_null_count(
+                        validbytes, options["count_nulls"]
+                    )
+                    or x._arrow_needs_option_type(),
```
ianna: @martindurant - I'm not sure if this is a correct way. It definitely fixes the error reported in #3402, but it affects typing. I don't know if we need to preserve the type or if it is even possible?

ianna (commented):
@martindurant - could you please have a look? Thanks!

src/awkward/contents/recordarray.py
martindurant: I'm not sure what _arrow_needs_option_type does. I seem to recall that when going in the arrow->awkward direction, you would get Unmasked types in the case that the type was nullable, but the NULLs mask is None. Maybe this is the same.

ianna: `_arrow_needs_option_type()` returns `is_option` meta. It is `True` for all contents that allow `null`s: `IndexedOptionArray`, `BitMaskedArray`, `ByteMaskedArray`, and `UnmaskedArray`.

src/awkward/contents/recordarray.py
```diff
@@ -1092,8 +1092,11 @@ def _to_arrow(
 
         types = pyarrow.struct(
             [
-                pyarrow.field(self.index_to_field(i), values[i].type).with_nullable(
-                    x._arrow_needs_option_type()
+                pyarrow.field(
+                    self.index_to_field(i),
+                    values[i].type,
+                    (mask_node is not None and mask_node._arrow_needs_option_type())
+                    or x._arrow_needs_option_type(),
```
ianna:
```suggestion
                    x._arrow_needs_option_type(),
```

PR #3419, opened 2025-03-18
fix: index with int in unionarray.py

ianna:
fix` TypeError:` list indices must be integers or slices, not `ndarray` for `cuda` backend.

### DIFF
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1586,7 +1586,7 @@ class UnionArray(UnionMeta[Content], Content):
         # backends with concrete data
         for i in range(self._tags.length):
             content = (
-                self._contents[self._tags.data[i]]
+                self._contents[int(self._tags.data[i])]
                 ._carry(ak.index.Index(self._index.data[i]), False)
                 ._remove_structure(backend, options)
             )
--- a/src/awkward/contents/unionarray.py
+++ b/src/awkward/contents/unionarray.py
@@ -1586,7 +1586,7 @@ class UnionArray(UnionMeta[Content], Content):
         # backends with concrete data
         for i in range(self._tags.length):
             content = (
-                self._contents[int(self._tags.data[i])]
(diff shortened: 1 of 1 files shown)


src/awkward/contents/unionarray.py
```diff
@@ -1586,7 +1586,7 @@ def _remove_structure(
         # backends with concrete data
         for i in range(self._tags.length):
             content = (
-                self._contents[self._tags.data[i]]
+                self._contents[int(self._tags.data[i])]
```
pfackeldey: Looks good to me, I was just curious if it should maybe be `self._tags.data[i].item()` instead (or is this the same?)?

pfackeldey (approved):

src/awkward/contents/unionarray.py
ianna: Thanks, yes, I think, it's the same result. I'll check what we do in other places.

pfackeldey: I saw e.g.: https://github.com/scikit-hep/awkward/blob/58eea9bc541e0f672468a846f80c0c980a90aa8d/src/awkward/_nplikes/cupy.py#L75 which uses `.item()` and thought that may be the preferred way for CuPy?

pfackeldey: Oh, wait, actually I'm confused, shouldn't this do that already automatically: https://github.com/scikit-hep/awkward/blob/58eea9bc541e0f672468a846f80c0c980a90aa8d/src/awkward/index.py#L236 ?

pfackeldey: Could you try `self._tags[i]` instead of `int(self._tags.data[i])` ?

ianna:
```suggestion
                self._contents[self._tags[i]]
```

ianna: Thanks! Done.

PR #3423, opened 2025-03-18
fix: `ruff` linting

ianna:
Remove unnecessary casts reported by `ruff` linting,
disable `ruff` liniting for `src/awkward/types/_awkward_datashape_parser.py`.


ianna (commented):
reported by `ruff`. Merging it to re-base next pr.

PR #3424, opened 2025-03-18
docs: fix `array_equal` reference guide

ianna:


ianna (commented):
Thanks @ikrommyd for spotting it!

PR #3425, opened 2025-03-18
fix: `ak.almost_equal` and `ak.array_equal` should check that jagged arrays have the same offsets

### LINKED ISSUE #3426: `array_equal`/`array_almost_equal` does not take into an account nested layouts
Hi awkward. Thanks for this great codebase, and I'm sorry if I'm asking a question which has been answered elsewhere. I looked around and couldn't find anything.

Do you understand why `array_equal` gives `True` in this case? The nested arrays are different.

```python
a = ak.Array([
    [ [], [1] ],
    [ [], [0] ]
])
b = ak.Array([
    [ [], [1], [] ],
    [ [0] ]
])
print(ak.array_equal(a, b))
```

`ak.__version__ = 2.7.4`

_Originally posted by @alexandertuna in https://github.com/scikit-hep/awkward/discussions/3418_

ikrommyd:
Fixes https://github.com/scikit-hep/awkward/issues/3426
Addresses https://github.com/scikit-hep/awkward/discussions/3418

Jagged arrays can have the same content but different offsets. `ak.almost_equal` and `ak.array_equal` do not check for that and only check whether the contents are equal. Example:
```python
In [15]: a = ak.Array([[1, 2], [3]])

In [16]: b = ak.Array([[1], [2, 3]])

In [17]: ak.almost_equal(a, b)
Out[17]: True

In [18]: ak.array_equal(a,b)
Out[18]: True
```
This isn't proper behavior in my opinion.

### DIFF
--- a/src/awkward/operations/ak_almost_equal.py
+++ b/src/awkward/operations/ak_almost_equal.py
@@ -162,8 +162,8 @@ def _impl(
         # List-list
         elif left.is_list and right.is_list:
             # Check that indexes are equal
-            left_index = left.to_ListOffsetArray64(False).offsets
-            right_index = right.to_ListOffsetArray64(False).offsets
+            left_index = left.to_ListOffsetArray64(True).offsets
+            right_index = right.to_ListOffsetArray64(True).offsets
             if not backend.index_nplike.array_equal(left_index, right_index):
                 return False
             # Mixed regular-var
--- a/src/awkward/operations/ak_almost_equal.py
+++ b/src/awkward/operations/ak_almost_equal.py
@@ -161,6 +161,11 @@ def _impl(
             return (left.size == right.size) and visitor(left.content, right.content)
         # List-list
         elif left.is_list and right.is_list:
+            # Check that indexes are equal
+            left_index = left.to_ListOffsetArray64(False).offsets
+            right_index = right.to_ListOffsetArray64(False).offsets
+            if not backend.index_nplike.array_equal(left_index, right_index):
+                return False
             # Mixed regular-var
             if left.is_regular and not right.is_regular:
                 return (not check_regular) and visitor(
--- a/tests/test_1105_ak_aray_equal.py
+++ b/tests/test_1105_ak_aray_equal.py
@@ -52,6 +52,16 @@ def test_array_equal_on_listoffsets():
     b = ak.Array([[[], [1], []], [[0]]])
     assert not ak.array_equal(a, b)
 
+    # same outer index, different inner index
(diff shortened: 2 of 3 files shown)


agoose77:
I think this is a bug, yes — we should likely convert to a list with `start=0`

src/awkward/operations/ak_almost_equal.py
```diff
@@ -161,6 +161,11 @@ def visitor(left, right) -> bool:
             return (left.size == right.size) and visitor(left.content, right.content)
         # List-list
         elif left.is_list and right.is_list:
+            # Check that indexes are equal
+            left_index = left.to_ListOffsetArray64(True).offsets
+            right_index = right.to_ListOffsetArray64(True).offsets
+            if not backend.index_nplike.array_equal(left_index, right_index):
+                return False
```
ianna, line 168: @ikrommyd - what about the nested `ListOffsetArrays` offsets?

ianna (commented):
@ikrommyd - Thanks for looking into it! Could you, please, add more tests for the `ListOffsetArrays` of `ListOffsetArrays` of `ListOffsetArrays` etc. Where offsets are the same at the top level, but different in the nested ones? Thanks.

src/awkward/operations/ak_almost_equal.py
ikrommyd: The function is recursive. It will be called on the contents again. If the contents are list-like, they have their own offsets and this will be checked.

ianna: Thanks! I've added a couple of more tests.

ianna (approved):
@ikrommyd - great, thanks! The tests pass, I'll merge.

src/awkward/operations/ak_almost_equal.py
ikrommyd: Oh I'm sorry. I was gonna do them once I woke up. I just added them to the other test as well (mainly for obsessive reasons 🤣). Merge whenever ready!

ikrommyd:
@all-contributors please add @ikrommyd for code

src/awkward/operations/ak_almost_equal.py
ianna: It's me, sorry. I assumed you are done because the PR was marked as "ready for review". Please, use "draft" mode to avoid misunderstanding in the future. Thanks!

ikrommyd: It's alright. I just didn't expect you to do it when you said "please add more tests" 😃. Noted for the future.

PR #3432, opened 2025-03-19
fix: an `Index`'s `data` should be passed into `array_equal`, not the `Index` itself

ikrommyd:
A subtle bug was introduced in https://github.com/scikit-hep/awkward/pull/3425. `array_equal` should see the `Index`'s `data`, not the `Index` itself. The ci didn't catch it but the offline cuda tests did because they request for the `.shape` property which doesn't exist for an `Index`.


ianna (approved):
@ikrommyd - thanks for fixing it!

PR #3434, opened 2025-03-20
feat: add `tobytes` to `VirtualArray`

ikrommyd:
Discovered via a failing test in `coffea`, we need it to make `ak._util.tobytes` work.

### DIFF
--- a/src/awkward/_nplikes/virtual.py
+++ b/src/awkward/_nplikes/virtual.py
@@ -191,6 +191,12 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
             lambda: self.materialize().byteswap(inplace=inplace),
         )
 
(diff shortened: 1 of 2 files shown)


pfackeldey:
works! thanks for the quick fix! that resolves https://github.com/scikit-hep/coffea/actions/runs/13957896388/job/39073801392?pr=1277 :)

ikrommyd:
@pfackeldey Only implementing `tobytes` would also work right? Numpy calls `tostring` deprecated since 1.19 and prompts the user to use `tobytes`. Cupy doesn't even implement it at all.

ianna (approved):
@ikrommyd - Thanks! Looks good to me.

pfackeldey (approved):
Great! Please go ahead and squash-merge @ikrommyd :)

PR #3439, opened 2025-03-30
docs: minor fixes to JAX guide + add VJP example

Saransh-cpp:
- explicitly tell users that `jax.jit` won't work on reducers
- `jax.grad` is supported
- add an example for `jax.vjp` (reverse mode AD)

### DIFF
--- a/docs/user-guide/how-to-specialize-differentiate-jax.md
+++ b/docs/user-guide/how-to-specialize-differentiate-jax.md
@@ -14,7 +14,7 @@ kernelspec:
 Differentiation using JAX
 =========================
 
-JAX, amongst other things, is a powerful tool for computing derivatives of native Python and NumPy code. Awkward Array implements support for the {func}`jax.jvp` and {func}`jax.vjp` JAX functions for computing forward/reverse-mode Jacobian-vector/vector-Jacobian products of functions that operate upon Awkard Arrays. Only a subset of Awkward Array operations can be differentiated through, including:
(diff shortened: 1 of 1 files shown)


ianna (approved):
@Saransh-cpp - thanks for the update! If you’re done with this, please merge it. Thanks

docs/user-guide/how-to-specialize-differentiate-jax.md
```diff
@@ -45,7 +45,8 @@ def reverse_sum(array):
     return ak.sum(array[::-1], axis=0)
 ```
 
-We can then create an array with which to evaluate `reverse_sum`. The `backend` argument ensures that we build an Awkward Array that is backed by {class}`jaxlib.xla_extension.DeviceArray` buffers, which power JAX's automatic differentiation and JIT compiling features.
+We can then create an array with which to evaluate `reverse_sum`. The `backend` argument ensures that we build an Awkward Array that is backed by {class}`jax.Array` ({class}`jaxlib.xla_extension.ArrayImpl`) buffers, which power JAX's automatic differentiation and JIT compiling features. However, Awkward Array's JAX backend does not support JIT compilation on reducers as XLA requires array sizes to not be
+dependent on data values at compile-time.
```
Saransh-cpp:
```suggestion
We can then create an array with which to evaluate `reverse_sum`. The `backend` argument ensures that we build an Awkward Array that is backed by {class}`jax.Array` ({class}`jaxlib.xla_extension.ArrayImpl`) buffers, which power JAX's automatic differentiation and JIT compiling features. However, Awkward Array's JAX backend does not support JIT compilation on reducers as XLA requires array sizes to not be dependent on data values at compile-time.
```

PR #3440, opened 2025-03-31
feat: Add a parquet uuid calculation

NJManganelli:
Calculate a uuid from parquet metadata, utilizing detailed info of the first and last row_groups plus the col_counts of all row_groups of the file or dataset. At the column-page level, parquet should have a checksum AFAIK, but an approximate calculation that would catch differences in numbers of rows, row groups, columns, compression, etc. that deterministically uses two row groups should be sufficient for the equivalent of what coffea does with root files (which is flag them for changes to recalculate the form, steps, etc.).

https://github.com/scikit-hep/coffea/blob/master/src/coffea/dataset_tools/preprocess.py#L46-L48

Also, the ParquetMetadata namedtuple doesn't appear to be used, at least in this file that's touched. Given there's an extra line to handle not changing the length of returned tuple to try and avoid breaking compatibility with outside users, maybe this should be deprecated and the namedtuple should be used instead?

### DIFF
--- a/tests/test_3440_calculate_parquet_uuid.py
+++ b/tests/test_3440_calculate_parquet_uuid.py
@@ -21,7 +21,10 @@ input = SAMPLES_DIR / "nullable-record-primitives.parquet"
 
 def test_parquet_uuid():
     meta = metadata_from_parquet(input)
-    assert meta["uuid"] == "582dabdb8c87bfa17bc930676ed26b8d4ab22a900f92357751dc380c41acb593"
+    assert (
+        meta["uuid"]
+        == "582dabdb8c87bfa17bc930676ed26b8d4ab22a900f92357751dc380c41acb593"
+    )
 
 
 @pytest.mark.parametrize("calculate_uuid", [True, False])
--- a/tests/test_3440_calculate_parquet_uuid.py
+++ b/tests/test_3440_calculate_parquet_uuid.py
@@ -8,6 +8,8 @@ import pytest
 
 import awkward
 
+pytest.importorskip("pyarrow.parquet")
+
 metadata_from_parquet_submodule = pytest.importorskip(
     "awkward.operations.ak_metadata_from_parquet"
 )
--- a/tests/test_3440_calculate_parquet_uuid.py
+++ b/tests/test_3440_calculate_parquet_uuid.py
@@ -23,7 +23,7 @@ def test_parquet_uuid():
     meta = metadata_from_parquet(input)
     assert (
         meta["uuid"]
-        == "37f505d95053fdd40fbc572b13fbc054d3b737ff85890990204c7e9cded13f87"
+        == "DEBUG"
     )
 
 
--- a/tests/test_3440_calculate_parquet_uuid.py
+++ b/tests/test_3440_calculate_parquet_uuid.py
@@ -1,0 +1,56 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
+
+from __future__ import annotations
+
+import pathlib
+
+import pytest
+
+import awkward
+
+metadata_from_parquet_submodule = pytest.importorskip(
+    "awkward.operations.ak_metadata_from_parquet"
+)
+metadata_from_parquet = metadata_from_parquet_submodule.metadata_from_parquet
+
+SAMPLES_DIR = pathlib.Path(__file__).parent / "samples"
+input = SAMPLES_DIR / "nullable-record-primitives.parquet"
+
+
+def test_parquet_uuid():
+    meta = metadata_from_parquet(input)
+    assert (
+        meta["uuid"]
+        == "93fd596534251b1ac750f359d9d55418b02bcddcc79cd5ab1e3d9735941fbcae"
+    )
+
+
+@pytest.mark.parametrize("calculate_uuid", [True, False])
+def test_return_tuple_with_or_without_uuid(calculate_uuid):
+    results = awkward.operations.ak_from_parquet.metadata(
+        input,
+        {},
+        None,
+        None,
+        False,
+        True,
+        calculate_uuid=calculate_uuid,
+    )
+    if calculate_uuid:
+        assert len(results) == 8, "Expected 8 items in the result tuple"
+        (
+            parquet_columns,
+            subform,
+            actual_paths,
+            fs,
+            subrg,
+            col_counts,
+            metadata,
+            uuid,
+        ) = results
+        assert uuid is not None, "UUID should be present when calculate_uuid is True"
+    else:
+        assert len(results) == 7, "Expected 7 items in the result tuple"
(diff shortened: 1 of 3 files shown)


ianna:
@NJManganelli - what is the status of this PR? Are you still working on it? Thanks!

NJManganelli:
Hi @ianna I'll add a test, then I think it'll be ready from my side

NJManganelli:
Not without a performance penalty, but if it needs to be optimized, we could figure out a smarter but still sufficient calculation (I'd like to ensure that any changes in compression, columns, rows is captured). It's also possible that I didn't explore enough some checksum information that's supposed to be available (but these I think were at the page level or something, and just the loop over all those seems like it would be much worse than this)

Without uuid:
```bash
python -m timeit -n 1000 "import test_3440_calculate_parquet_uuid; test_3440_calculate_parquet_uuid.test_parquet_uuid()"
1000 loops, best of 5: 83 usec per loop
```
With uuid:
```
tests git:(parquet_uuid) ✗ python -m timeit -n 1000 "import test_3440_calculate_parquet_uuid; test_3440_calculate_parquet_uuid.test_parquet_uuid()"
1000 loops, best of 5: 209 usec per loop
```

NJManganelli:
Rebased for my own sanity, and marked ready (presuming all the tests are going to pass, will fix if otherwise)

ianna (changes requested):
@NJManganelli - it looks like uuids do not match:

```python
______________________________ test_parquet_uuid _______________________________

    def test_parquet_uuid():
        meta = metadata_from_parquet(input)
>       assert (
            meta["uuid"]
            == "93fd596534251b1ac750f359d9d55418b02bcddcc79cd5ab1e3d9735941fbcae"
        )
E       AssertionError: assert 'adc236484e10384ad680a1ae2ce1fc675f6f9f0a84a02c651ed91ad97fba41c0' == '93fd596534251b1ac750f359d9d55418b02bcddcc79cd5ab1e3d9735941fbcae'
E         
E         - 93fd596534251b1ac750f359d9d55418b02bcddcc79cd5ab1e3d9735941fbcae
E         + adc236484e10384ad680a1ae2ce1fc675f6f9f0a84a02c651ed91ad97fba41c0

meta       = {'col_counts': [5],
 'columns': ['u1', 'u4', 'u8', 'f4', 'f8', 'raw', 'utf8'],
 'form': RecordForm([BitMaskedForm('u8', NumpyForm('bool'), True, True), BitMaskedForm('u8', NumpyForm('int32'), True, True), BitMaskedForm('u8', NumpyForm('int64'), True, True), BitMaskedForm('u8', NumpyForm('float32'), True, True), BitMaskedForm('u8', NumpyForm('float64'), True, True), BitMaskedForm('u8', ListOffsetForm('i32', NumpyForm('uint8', parameters={'__array__': 'byte'}), parameters={'__array__': 'bytestring'}), True, True), BitMaskedForm('u8', ListOffsetForm('i32', NumpyForm('uint8', parameters={'__array__': 'char'}), parameters={'__array__': 'string'}), True, True)], ['u1', 'u4', 'u8', 'f4', 'f8', 'raw', 'utf8']),
 'fs': <fsspec.implementations.local.LocalFileSystem object at 0x7ff63efc9340>,
 'num_row_groups': 1,
 'num_rows': 5,
 'paths': ['/home/runner/work/awkward/awkward/tests/samples/nullable-record-primitives.parquet'],
 'uuid': 'adc236484e10384ad680a1ae2ce1fc675f6f9f0a84a02c651ed91ad97fba41c0'}

tests/test_3440_calculate_parquet_uuid.py:22: AssertionError
```

NJManganelli:
Aye, looks like this will need to be more selective of what goes into the hash, from that printout. I’ll have a look when I am back from holidays

NJManganelli:
@ianna I'm trying a more selective set of key-value pairs, hoping it'll be more stable, but "it works on my machine" the same as the previous one, so need to see what the tests say I think

ianna:
@all-contributors please add @NJManganelli  for code

ianna:
strange:
```python
ImportError: to use ak.from_parquet, you must install pyarrow:

    pip install pyarrow

or

    conda install -c conda-forge pyarrow
```

tests/test_3440_calculate_parquet_uuid.py
```diff
@@ -0,0 +1,57 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
+
+from __future__ import annotations
+
+import pathlib
+
+import pytest
+
+import awkward
+
```
ariostas, line 10: You should add `pytest.importorskip("pyarrow.parquet")` somewhere here since not all workflows install it.

ariostas: Also, maybe it's worth bumping this to 18 and including 3.13 to see what happens.

https://github.com/scikit-hep/awkward/blob/a5c9510c081dbbf236f4ed1271a25a79841ed70a/requirements-test-full.txt#L6

ianna: > Also, maybe it's worth bumping this to 18 and including 3.13 to see what happens.
> 
> https://github.com/scikit-hep/awkward/blob/a5c9510c081dbbf236f4ed1271a25a79841ed70a/requirements-test-full.txt#L6

I hope to be able to remove the pinned version before the release.

NJManganelli: I had hoped "metadata_from_parquet_submodule = pytest.importorskip(
    "awkward.operations.ak_metadata_from_parquet"
)" would implicitly accomplish this because it imports pyarrow. Okay, adding the explicit importskip

NJManganelli:
Nope, it is not actually producing a stable hash, so it's still not a uuid. I don't know if the dictionary keys might not be sorting the same every time, either because of parquet data ingestion or something else, if some of the values are not stable, or something else entirely.

How should I iterate on this? temporarily add in a print out of all the intermediate info that goes into the hash and find the difference when a test fails?

NJManganelli:
@ianna could tests be rerun with these debug commits? I couldn't think of another way to discern what is os-dependent or not deterministic, i always get the same hash on my system

NJManganelli:
Found two problems, in the "columns" key a field called "statistics" stores extra info like min, max, etc. distinct_counts is None or 0 in the minimal/full ubuntu install. the sorting columns is also missing in one. Let's see if that's everything or not, and if so I'll remove the two debug commits / update the test hash

NJManganelli:
Alright, 17th time is the charm, as they say

NJManganelli:
Thanks, @ikrommyd 
Passed this time, if we later discover any skew that maybe pops up with pyarrow changes, I’ll address it then.

NJManganelli:
@ianna a anything else needed?

ianna (approved):
@NJManganelli - great! Thanks. I’m merging it.

PR #3446, opened 2025-04-07
fix: remove mixed numpy<->jax arrays in Jax backend

pfackeldey:
This PR makes the jax backend _only_ use jax arrays (no more mixing with numpy arrays). This was the only motivation to split `nplike` and `index_nplike` so far, which is why `index_nplike` is now entirely dropped. 

In addition, this PR enables (fancier) slicing with the jax backend (I had to introduce Jax's set/getitem syntax where needed), such as:
```python
import awkward as ak

ak.jax.register_and_check()

array = ak.Array({"foo": [[1], [], [3]]}, backend="jax")
array.foo[array.foo>2]
>> <Array [[], [], [3]] type='3 * var * int64'> 

# and
array[array.foo>2]
>> <Array [{foo: []}, {...}, {foo: [3]}] type='3 * {foo: var * int64}'>

# or
array = ak.Array([[1.0, 2.0, 3.0], [], [4.0, 5.0]], backend="jax")
array[array>2.0]
>> <Array [[3.0], [], [4.0, 5.0]] type='3 * var * float64'>
```
None of these cuts are possible with the current jax backend.

### DIFF
--- a/src/awkward/_backends/jax.py
+++ b/src/awkward/_backends/jax.py
@@ -22,20 +22,13 @@ class JaxBackend(Backend):
     name: Final[str] = "jax"
 
     _jax: Jax
-    _numpy: Numpy
 
     @property
     def nplike(self) -> Jax:
         return self._jax
 
-    @property
-    def index_nplike(self) -> Numpy:
-        # we need to ensure Numpy like here for our awkward-cpp kernels (they don't work with Jax arrays)
-        return self._numpy
-
     def __init__(self):
         self._jax = Jax.instance()
-        self._numpy = Numpy.instance()
 
     def __getitem__(self, index: KernelKeyType) -> JaxKernel:
         # JAX uses Awkward's C++ kernels for index-only operations
--- a/src/awkward/contents/indexedoptionarray.py
+++ b/src/awkward/contents/indexedoptionarray.py
@@ -10,6 +10,7 @@ from awkward._backends.backend import Backend
 from awkward._layout import maybe_posaxis
 from awkward._meta.indexedoptionmeta import IndexedOptionMeta
 from awkward._nplikes.array_like import ArrayLike
+from awkward._nplikes.jax import Jax
 from awkward._nplikes.numpy import Numpy
 from awkward._nplikes.numpy_like import IndexType, NumpyMetadata
 from awkward._nplikes.placeholder import PlaceholderArray
@@ -1722,11 +1723,21 @@ class IndexedOptionArray(IndexedOptionMeta[Content], Content):
             is_none = original_index < 0
             num_none = nplike.count_nonzero(is_none)
             new_index = nplike.empty(self._index.length, dtype=self._index.dtype)
-            new_index[is_none] = -1
-            new_index[~is_none] = nplike.arange(
-                nplike.shape_item_as_index(new_index.size - num_none),
-                dtype=self._index.dtype,
-            )
+            if isinstance(nplike, Jax):
+                num_none = num_none.item()
+                new_index = new_index.at[is_none].set(-1)
+                new_index = new_index.at[~is_none].set(
+                    nplike.arange(
+                        nplike.shape_item_as_index(new_index.size - num_none),
+                        dtype=self._index.dtype,
+                    )
+                )
+            else:
+                new_index[is_none] = -1
+                new_index[~is_none] = nplike.arange(
+                    nplike.shape_item_as_index(new_index.size - num_none),
+                    dtype=self._index.dtype,
+                )
             projected = self.project()
             return ak.contents.IndexedOptionArray(
                 ak.index.Index(new_index, nplike=self._backend.nplike),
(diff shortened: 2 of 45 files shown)


agoose77:
FYI the split nplikes was also thought to be useful for mixed-device arrays iirc. There's some issue about it somewhere I think, or slack message :/

pfackeldey:
> FYI the split nplikes was also thought to be useful for mixed-device arrays iirc. There's some issue about it somewhere I think, or slack message :/

Oh, I didn't know! I understood from @jpivarski that it was introduced because of the mix of array types in the Jax backend.

Also: I didn't even know we support mixed device arrays? Is this an awkward array where parts live on CPU and parts on GPU? I don't think this is possible right now, is it? Afaik you can only change the backend for a whole array. The current implementation would also only allow to mix nplikes for indices and data, I'm not sure if that is really useful?

pfackeldey:
I don't know why this one numba test is failing? I don't see how it's related to this PR?

ikrommyd:
> I don't know why this one numba test is failing? I don't see how it's related to this PR?

It’s also failing on https://github.com/scikit-hep/awkward/pull/3445

ianna:
Hmm... it looks like a memory leak - we are still holding a ref while deleting:
```python
>       assert sys.getrefcount(array) == 4
E       AssertionError: assert 5 == 4
E        +  where 5 = <built-in function getrefcount>(<Array [1, 2, 3] type='3 * int64'>)
E        +    where <built-in function getrefcount> = sys.getrefcount

array      = <Array [1, 2, 3] type='3 * int64'>
f2         = CPUDispatcher(<function test_refcount.<locals>.f2 at 0x7f5d6d75e980>)
```

ianna (commented):
JAX will automatically handle broadcasting and device placement when performing operations on arrays from different devices:

```python
import jax
import jax.numpy as jnp

cpu_array = jnp.ones((3, 3))
gpu_array = jax.device_put(jnp.ones((3, 3)), device=jax.devices('gpu')[0]) 

result = cpu_array + gpu_array

print(result)
```

src/awkward/_backends/jax.py
```diff
@@ -22,20 +22,13 @@ class JaxBackend(Backend):
...
     _jax: Jax
-    _numpy: Numpy
 
     @property
     def nplike(self) -> Jax:
         return self._jax
 
-    @property
-    def index_nplike(self) -> Numpy:
-        # we need to ensure Numpy like here for our awkward-cpp kernels (they don't work with Jax arrays)
```
ianna, line 33: It looks like we do need it, now that we have more tests some of them are failing with a `TypeError`:
```python
tests/test_1490_jax_reducers_combinations.py::test_reducer[prod1-None] PASSED [ 48%]
Error: test_reducer[ptp-0]

TypeError: JAX arrays are immutable and do not support in-place item assignment. Instead of x[idx] = y, use x = x.at[idx].set(y) or another .at[] method: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html

    ak.to_list(
        <Array [6.0, 6.0, 6.0] type='3 * ?float64'>
    )
tests/test_1490_jax_reducers_combinations.py::test_reducer[ptp-0] FAILED [ 48%]
Error: test_reducer[ptp-1]

TypeError: JAX arrays are immutable and do not support in-place item assignment. Instead of x[idx] = y, use x = x.at[idx].set(y) or another .at[] method: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html

    ak.to_list(
        <Array [2.0, 2.0, 2.0] type='3 * ?float64'>
    )
```

ianna (changes requested):
@pfackeldey - what do you think are our options here? Thanks.

src/awkward/_backends/jax.py
pfackeldey: These are additional place where we have to use the `setitem` syntax of JAX. Weirdly, they passed for me locally, I'll try to look into it 👍

src/awkward/contents/indexedoptionarray.py
```diff
@@ -1737,19 +1717,19 @@ def to_packed(self, recursive: bool = True) -> Self:
...
         else:
-            index_nplike = self._backend.index_nplike
+            nplike = self._backend.nplike
             original_index = self._index.data
             is_none = original_index < 0
-            num_none = index_nplike.count_nonzero(is_none)
-            new_index = index_nplike.empty(self._index.length, dtype=self._index.dtype)
+            num_none = nplike.count_nonzero(is_none)
+            new_index = nplike.empty(self._index.length, dtype=self._index.dtype)
             new_index[is_none] = -1
```
ianna:
```suggestion
            new_index.at[is_none].set(-1)
```

ianna (changes requested):
@pfackeldey - this is what JAX suggests :-)

src/awkward/contents/indexedoptionarray.py
ianna: there might be some other places as well...

pfackeldey: yeah, I just pushed the fix :)

Saransh-cpp (approved):
Thanks, @pfackeldey! This looks amazing!

PR #3447, opened 2025-04-08
fix: highlevel function dispatch type hint

### LINKED ISSUE #3401: Broken type hints for functions decorated with `high_level_function`
### Version of Awkward Array

2.7.4

When using functions like `ak.ones_like`, `ak.almost_equal`, the static type checkers (for reference I tested with pylance in vscode and pycharm) are not able to properly resolve the signatures of those functions. Instead, they will always give a `(...) -> Any`. It is likely because in the type hint for [`high_level_function`](https://github.com/scikit-hep/awkward/blob/e2f61af73e076fcb902fed939d69c79ea981a3b6/src/awkward/_dispatch.py#L12-L19) the parameters are omitted and the `TypeVar("T")` is not binded to the decorator. A simple solution is to a `ParamSpec` in the following way:
```python
from awkward._typing import Any, ParamSpec, TypeAlias, TypeVar

P = ParamSpec("P") # declare a ParamSpec
T = TypeVar("T")
DispatcherType: TypeAlias = "Callable[P, Generator[Collection[Any], None, T]]" # add P
HighLevelType: TypeAlias = "Callable[P, T]" # add P

def high_level_function(
    module: str = "ak", name: str | None = None
) -> Callable[[DispatcherType[P, T]], HighLevelType[P, T]]: # bind P and T to the decorator
```

pfackeldey:
(Hopefully) closes https://github.com/scikit-hep/awkward/issues/3401

Can you give it a try @chuyuanliu ?

### DIFF
--- a/src/awkward/_dispatch.py
+++ b/src/awkward/_dispatch.py
@@ -7,16 +7,17 @@ from functools import wraps
 from inspect import isgenerator
 
 from awkward._errors import OperationErrorContext
-from awkward._typing import Any, TypeAlias, TypeVar
+from awkward._typing import Any, ParamSpec, TypeAlias, TypeVar
 
+P = ParamSpec("P")
 T = TypeVar("T")
-DispatcherType: TypeAlias = "Callable[..., Generator[Collection[Any], None, T]]"
-HighLevelType: TypeAlias = "Callable[..., T]"
+DispatcherType: TypeAlias = "Callable[P, Generator[Collection[Any], None, T]]"
+HighLevelType: TypeAlias = "Callable[P, T]"
 
 
 def high_level_function(
     module: str = "ak", name: str | None = None
-) -> Callable[[DispatcherType], HighLevelType]:
+) -> Callable[[DispatcherType[P, T]], HighLevelType[P, T]]:
     """Decorate a high-level function such that it may be overloaded by third-party array objects"""
 
     def capture_func(func: DispatcherType) -> HighLevelType:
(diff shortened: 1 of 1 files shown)


chuyuanliu:
I tested both commits with pylance. The first one 207d9eda89fc96a1600772921690857fb3109741 works fine. But after the change https://github.com/scikit-hep/awkward/commit/849d5385c62fb72ac1a587cd9d831dbfca23eab5#diff-c903c22679aaf17d03f02f4fcedb5004bb2e146053f50bcd1ffc7eeead27d5c0R20 is introduced, the typehint disappear again, as the P, T are no longer bound to that function. If the second commit is to fix the E1126 in pylint, I came across this issue https://github.com/pylint-dev/pylint/issues/9908.

pfackeldey:
> I tested both commits with pylance. The first one [207d9ed](https://github.com/scikit-hep/awkward/commit/207d9eda89fc96a1600772921690857fb3109741) works fine. But after the change [849d538#diff-c903c22679aaf17d03f02f4fcedb5004bb2e146053f50bcd1ffc7eeead27d5c0R20](https://github.com/scikit-hep/awkward/commit/849d5385c62fb72ac1a587cd9d831dbfca23eab5#diff-c903c22679aaf17d03f02f4fcedb5004bb2e146053f50bcd1ffc7eeead27d5c0R20) is introduced, the typehint disappear again, as the P, T are no longer bound to that function. If the second commit is to fix the E1126 in pylint, I came across this issue [pylint-dev/pylint#9908](https://github.com/pylint-dev/pylint/issues/9908).

Thank for the link!
Ahh, oops, I'll fix it in a moment 😅

chuyuanliu:
f9ec16328f9fe222ea745ed9570f3b043f1e8873 looks good to me.

agoose77:
It's really nice to see type-hint fixes going on. A lot of the type hints we have (or at least, had) were advisory, and not always consistent. Over time, we should look to reduce the number of exempt files in `pyproject.toml`!

ianna (approved):
@pfackeldey - Looks good! Thanks!

PR #3448, opened 2025-04-08
fix: force gc in numba test

ianna:


ianna (commented):
It seems strange that gc fixes the issue. Have we introduced a cyclic dependency? Is GC more lazy in 3.11?

PR #3449, opened 2025-04-09
feat: add segmented argmin and argmax

ianna:
also enable tests for more reducers

### DIFF
--- a/src/awkward/_connect/jax/reducers.py
+++ b/src/awkward/_connect/jax/reducers.py
@@ -37,6 +37,34 @@ class JAXReducer(Reducer):
         raise NotImplementedError
 
 
+def segment_argmin(data, segment_ids, num_segments=None):
+    """
+    Applies a segmented argmin-style reduction.
+
+    Parameters:
+        data: jax.numpy.ndarray — the values to reduce.
+        segment_ids: same shape as data — indicates segment groupings.
+        num_segments: number of segments (optional).
+
+    Returns:
+        jax.numpy.ndarray — indices of min within each segment.
+    """
+    if num_segments is None:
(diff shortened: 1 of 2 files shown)


src/awkward/_connect/jax/reducers.py
```diff
@@ -59,7 +88,39 @@ def apply(
...
+
+    data = jax.numpy.asarray(data)
+    segment_ids = jax.numpy.asarray(segment_ids)
+
+    # Create a 2D array where each row corresponds to one segment
+    segmented = jax.vmap(
+        lambda i: jax.numpy.where(segment_ids == i, data, -jax.numpy.inf)
+    )(jax.numpy.arange(num_segments))
+
+    # Apply the argmin along axis 1 (within each segment)
```
pfackeldey:
```suggestion
    # Apply the argmax along axis 1 (within each segment)
```

pfackeldey (approved):
This looks great! :tada: 

These were the two missing reducers afaik for the jax backend.

I suspect that this does not scale nicely for many segments (also memory wise due to the `vmap`), but I don't see a way around it?

ianna:
> This looks great! 🎉
> 
> These were the two missing reducers afaik for the jax backend.
> 
> I suspect that this does not scale nicely for many segments (also memory wise due to the `vmap`), but I don't see a way around it?

Thanks! It was inspired by [the deleted comment](https://github.com/scikit-hep/awkward/pull/3446/files#diff-58be18bab1634e18e76f7c91d879d9bb8ee7a72304b8863bc78920f6ca2ca93aL33) in your PR. I just wanted to be sure that all reducers are tested. There are a few more that are "not implemented", but I might leave it for a summer student :-)

PR #3450, opened 2025-04-10
feat: change argmin and argmax algorithms to handle big data

ianna:
The new algorithms work better on big data - tested up to 100 million. Although the old algorithm is faster it runs out of memory with the data over a few million values.


ianna (commented):
@Saransh-cpp - please have a look when you fave time. I'll merge it in the meantime.

Saransh-cpp (commented):
This looks good, @ianna! Thanks for writing these custom segment functions!

PR #3452, opened 2025-04-11
feat: implement segment count_nonzero reducer

ianna:


pfackeldey (approved):
nice! 🎉 
(is there somewhere a list of kernels that are missing for the jax backend?)

ianna:
> nice! 🎉 (is there somewhere a list of kernels that are missing for the jax backend?)

Thanks! This PR completes the JAX reducers which were marked as `NotImplemented`. I don't have such a list, but was planning to open a discussion for that.

PR #3453, opened 2025-04-11
fix: better support for pyright

henryiii:
Currently this is relying on a side effect of import to fill in the namespace; by actually placing the modules in the local namespace, type checkers like pyright will work.


ikrommyd:
The issue still persists on the internals like `ak._util.kSliceNone` which this PR didn't change. Maybe it should?

ianna (approved):
@henryiii - Thanks! It looks great. Please merge it if you finished with it.

PR #3455, opened 2025-04-11
fix: setitem in ak firsts with jax backend

### LINKED ISSUE #3454: `ak.firsts` doesn't work with the jax backend
This is introduced by https://github.com/scikit-hep/awkward/pull/3446. With the previous commit it works.
```py
import awkward as ak
ak.jax.register_and_check()

array = ak.Array([[1.1], [2.2], [], [3.3], [], [], [4.4], [5.5]], backend="jax")
ak.firsts(array)
```
will raise 
```
File ~/Dropbox/work/coffea_dev/awkward/src/awkward/operations/ak_firsts.py:109, in _impl.<locals>.action(layout, depth, backend, **kwargs)
    106     stops = layout.stops.data
    108     empties = starts == stops
--> 109     index[empties] = -1
    111     return ak.contents.IndexedOptionArray.simplified(
    112         ak.index.Index64(index), layout._content
    113     )
    115 elif layout.is_leaf:

File ~/miniforge3/envs/ak-uproot/lib/python3.12/site-packages/jax/_src/numpy/array_methods.py:592, in _unimplemented_setitem(self, i, x)
    588 def _unimplemented_setitem(self, i, x):
    589   msg = ("JAX arrays are immutable and do not support in-place item assignment."
    590          " Instead of x[idx] = y, use x = x.at[idx].set(y) or another .at[] method:"
    591          " https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html")
--> 592   raise TypeError(msg.format(type(self)))

TypeError: JAX arrays are immutable and do not support in-place item assignment. Instead of x[idx] = y, use x = x.at[idx].set(y) or another .at[] method: https://jax.readthedocs.io/en/latest/_autosummary/jax.numpy.ndarray.at.html

    ak.firsts(
        <Array [[1.1], [2.2], [], ..., [4.4], [5.5]] type='8 * var * float64'>
    )
```

pfackeldey:
Fixes https://github.com/scikit-hep/awkward/issues/3454

### DIFF
--- a/src/awkward/operations/ak_firsts.py
+++ b/src/awkward/operations/ak_firsts.py
@@ -10,6 +10,7 @@ from awkward._namedaxis import (
     _named_axis_to_positional_axis,
     _remove_named_axis,
 )
+from awkward._nplikes.jax import Jax
 from awkward._nplikes.numpy_like import NumpyMetadata
 from awkward._nplikes.shape import unknown_length
 from awkward._regularize import regularize_axis
@@ -106,7 +107,10 @@ def _impl(array, axis, highlevel, behavior, attrs):
                 stops = layout.stops.data
 
                 empties = starts == stops
-                index[empties] = -1
+                if isinstance(layout.backend.nplike, Jax):
+                    index = index.at[empties].set(-1)
+                else:
+                    index[empties] = -1
 
                 return ak.contents.IndexedOptionArray.simplified(
                     ak.index.Index64(index), layout._content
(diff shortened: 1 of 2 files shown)


ikrommyd:
@pfackeldey is it possible to make sure that there are no other places when `.at`s are missing. Is it "easy" to check?

ianna:
> @pfackeldey is it possible to make sure that there are no other places when `.at`s are missing. Is it "easy" to check?

I think, coverage should give us a clue if anything is not tested.

ianna (approved):
@pfackeldey - Thanks for fixing it!

Issue #3456, opened 2025-04-11
`ak.min/max/argmin/argmax` do not work on `axis=1` for the `jax` backend if the last element of an input `ListOffsetArray` is empty


ikrommyd:
Example:
```py
import awkward as ak
ak.jax.register_and_check()

ak.min(ak.Array([[1, 2], []], backend="cpu"), axis=1) #works
ak.min(ak.Array([[1, 2], []], backend="jax"), axis=1) #errors
```
All `ak.min/max/argmin/argmax` are going to give you 
```
File ~/Dropbox/work/coffea_dev/awkward/src/awkward/contents/listoffsetarray.py:1618, in ListOffsetArray._reduce_next(self, reducer, negaxis, starts, shifts, parents, outlength, mask, keepdims, behavior)
   1615 trimmed = self._content[self.offsets[0] : self.offsets[-1]]
   1616 nextstarts = self.offsets[:-1]
-> 1618 outcontent = trimmed._reduce_next(
   1619     reducer,
   1620     negaxis,
   1621     nextstarts,
   1622     shifts,
   1623     nextparents,
   1624     globalstarts_length,
   1625     mask,
   1626     keepdims,
   1627     behavior,
   1628 )
   1630 outoffsets = Index64.empty(outlength + 1, nplike)
   1631 assert outoffsets.nplike is nplike and parents.nplike is nplike

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/contents/numpyarray.py:1155, in NumpyArray._reduce_next(self, reducer, negaxis, starts, shifts, parents, outlength, mask, keepdims, behavior)
   1139     assert (
   1140         outmask.nplike is self._backend.nplike
   1141         and parents.nplike is self._backend.nplike
   1142     )
   1143     self._backend.maybe_kernel_error(
   1144         self._backend[
   1145             "awkward_NumpyArray_reduce_mask_ByteMaskedArray_64",
   (...)   1153         )
   1154     )
-> 1155     out = ak.contents.ByteMaskedArray(outmask, out, False, parameters=None)
   1157 if keepdims:
   1158     out = ak.contents.RegularArray(out, 1, self.length, parameters=None)

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/contents/bytemaskedarray.py:140, in ByteMaskedArray.__init__(self, mask, content, valid_when, parameters)
    131     raise TypeError(
[...]

PR #3451, opened 2025-04-11
feat: enable jax backend for virtual arrays

ikrommyd:
Title explains it. This is a PR we'll need to revert if the jax backend gets dropped one day.

PR #3457, opened 2025-04-12
fix: replace some `__setitem__` calls with `.at[where].set(value)` for the jax backend

ikrommyd:
Expands  upon https://github.com/scikit-hep/awkward/pull/3455 and adds all the fixes I had to add to get the tests added in https://github.com/scikit-hep/awkward/pull/3451 working locally.
This + fixing https://github.com/scikit-hep/awkward/issues/3456 should get the ci there to pass.


ikrommyd:
This still needs tests for sure. I think we should probably merge them with the tests from https://github.com/scikit-hep/awkward/pull/3455

tests/test_3454_jax_ak_firsts.py
```diff
@@ -1,18 +0,0 @@
...
-
-
-def test_jax_ak_firsts():
-    ak.jax.register_and_check()
-
-    jax_array = ak.Array([[1.1], [2.2], [], [3.3], [], [], [4.4], [5.5]], backend="jax")
-    jax_firsts = ak.firsts(jax_array)
-    cpu_array = ak.to_backend(jax_array, "cpu")
-    cpu_firsts = ak.firsts(cpu_array)
-    assert jax_firsts.to_list() == cpu_firsts.to_list()
```
ianna: @ikrommyd - please, restore the test. Our policy is to add a test with each PR and modify one if there is a bug in the test. Thanks!

src/awkward/operations/ak_unflatten.py
```diff
@@ -153,8 +154,16 @@ def _impl(array, counts, axis, highlevel, behavior, attrs):
...
+        if isinstance(maybe_counts_layout.backend.nplike, Jax):
+            current_offsets = current_offsets.at[0].set(0)
+            current_offsets = current_offsets.at[1:].set(
+                maybe_counts_layout.backend.nplike.cumsum(counts)
+            )
+        else:
+            current_offsets[0] = 0
+            maybe_counts_layout.backend.nplike.cumsum(
+                counts, maybe_out=current_offsets[1:]
+            )
```
ianna, line 166: @pfackeldey - I wonder if we should try to avoid this boilerplate code. Perhaps, something like:
```python
class BackendArraySetter:
    def __init__(self, array, backend):
        self.array = array
        self.backend = backend
        self.nplike = backend.nplike

    def __setitem__(self, index, value):
        if hasattr(self.nplike, "at"):  # JAX-like
            self.array = self.array.at[index].set(value)
        elif hasattr(self.array, "__setitem__"):  # NumPy, CuPy, etc.
            self.array[index] = value
        elif hasattr(self.nplike, "scatter"):  # PyTorch-like (custom handling)
            # You can define custom scatter logic here if needed
            raise NotImplementedError("Add scatter logic for torch here.")
        else:
            raise TypeError(f"Unsupported nplike type: {type(self.nplike)}")

    def result(self):
        return self.array
```

ianna (changes requested):
@ikrommyd - Thanks for chasing and promptly fixing the bug! If you think #3455 is incomplete, I can revert it to avoid removing the test. Thanks!

tests/test_3454_jax_ak_firsts.py
ikrommyd: Hi @ianna, my goal was just to have one test for all the similar fixes instead of two so I just copied the test from https://github.com/scikit-hep/awkward/pull/3455 to the new one here. The code changes from that PR are needed though. Do you think that the best approach is to revert that one and put everything here so that we have it grouped together? If so, please go ahead reverting that one and we can add the changes in this one. https://github.com/scikit-hep/awkward/pull/3455 was just the first one I found that Peter fixed very promptly.

src/awkward/operations/ak_unflatten.py
ikrommyd: I like that one! I found it kinda annoying having all those if statements everywhere. Even a `ak._util.setitem` function (not a class) or something similar would do. It may be best to wait a little bit for that refactoring in case we find some more places where this is needed that my testing didn't cover.

ikrommyd: OHHH now I like this even more cause it's gonna let us override https://github.com/cupy/cupy/issues/9089 which is causing an issue with virtual arrays in the cuda backend.

pfackeldey: The idea is good. I think it should be a method on each nplike, not in `ak._util` or a setter class.

tests/test_3454_jax_ak_firsts.py
ianna: @ikrommyd - please, add the changes to [ak.firsts](https://github.com/scikit-hep/awkward/pull/3460/files#diff-3f269762210dee401f483a6d433c4f932036cc61ebf0d98c7043e0c668151cfdL12-L115) here. Thanks!

src/awkward/operations/ak_unflatten.py
ianna: Thanks. I have opened an issue for that. We de want to factor it out, but we do not want to allow mutable arrays.

ianna (approved):
@ikrommyd - Great! Thanks for fixing it!

PR #3460, opened 2025-04-12
revert: "fix: setitem in ak firsts with jax backend"

ianna:
Reverts scikit-hep/awkward#3455


ianna (commented):
reverting this one not to duplicate #3457 that is more complete.

PR #3464, opened 2025-04-14
fix: jax reducers returning incorrect output values or lengths

### LINKED ISSUE #3456: `ak.min/max/argmin/argmax` do not work on `axis=1` for the `jax` backend if the last element of an input `ListOffsetArray` is empty

### LINKED ISSUE #3462: All jax backend reducers give wrong length output if the last element of the input jagge arrays is empty
All the reducers ignore the last element as we're not properly assigning the right `num_segments` and that causes them to ignore the last element which is empty
```py
In [5]: import awkward as ak
   ...: ak.jax.register_and_check()
   ...:
   ...: x = ak.sum(ak.Array([[1, 2], [], [6,4,5], []], backend="cpu"), axis=1)
   ...: y = ak.sum(ak.Array([[1, 2], [], [6,4,5], []], backend="jax"), axis=1)

In [6]: x
Out[6]: <Array [3, 0, 15, 0] type='4 * int64'>

In [7]: y
Out[7]: <Array [3, 0, 15] type='3 * int64'>
```
Similar results are observed with every reducer.

### LINKED ISSUE #3463: `ak.argmin/argmax` give incorrect results in the jax backend
```py
In [3]: import awkward as ak
   ...: ak.jax.register_and_check()
   ...:
   ...: x = ak.argmin(ak.Array([[1, 2], [], [6,4,5], [7]], backend="cpu"), axis=1)
   ...: y = ak.argmin(ak.Array([[1, 2], [], [6,4,5], [7]], backend="jax"), axis=1)

In [4]: x
Out[4]: <Array [0, None, 1, 0] type='4 * ?int64'>

In [5]: y
Out[5]: <Array [0, None, 3, 5] type='4 * ?int64'>
```
My understand is that this happens because positional corrections aren't applied like the cpu backend does: https://github.com/scikit-hep/awkward/blob/c93da2f8bf88074068a50a1e5fe52c72857bb767/src/awkward/_reducers.py#L186 even though the output of the reduce kernel is correct.

### LINKED ISSUE #3465: `ak.any` yields True for empty rows in float arrays with the jax backend
I'm actually not sure if this is a bug. I actually don't know if  `ak.any` should work at all in such cases. I'm just noticing and inconsistent behavior between the cpu and jax backends even though ideally `ak.any` should be called on boolean arrays.

To reproduce
```py
In [1]: import awkward as ak

In [2]: ak.jax.register_and_check()

In [3]: ak.any(ak.Array([[1], [1, 2], [], [3, 4], [7]], backend="cpu"), axis=1)
Out[3]: <Array [True, True, False, True, True] type='5 * bool'>

In [4]: ak.any(ak.Array([[1], [1, 2], [], [3, 4], [7]], backend="jax"), axis=1)
WARNING:2025-04-13 19:20:37,179:jax._src.xla_bridge:967: An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
Out[4]: <Array [True, True, True, True, True] type='5 * bool'>
```
It also yields incorrect length
```py
In [1]: import awkward as ak

In [2]: ak.jax.register_and_check()

In [3]: ak.any(ak.Array([[], [1, 2], [], [3, 4], []], backend="cpu"), axis=1)
Out[3]: <Array [False, True, False, True, False] type='5 * bool'>

In [4]: ak.any(ak.Array([[], [1, 2], [], [3, 4], []], backend="jax"), axis=1)
WARNING:2025-04-13 19:17:46,219:jax._src.xla_bridge:967: An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
Out[4]: <Array [True, True, True, True] type='4 * bool'>
```
but the current status of https://github.com/scikit-hep/awkward/pull/3464 fixes that and only gives us incorrect values:
```py
In [1]: import awkward as ak

In [2]: ak.jax.register_and_check()

In [3]: ak.any(ak.Array([[], [1, 2], [], [3, 4], []], backend="cpu"), axis=1)
Out[3]: <Array [False, True, False, True, False] type='5 * bool'>

In [4]: ak.any(ak.Array([[], [1, 2], [], [3, 4], []], backend="jax"), axis=1)
WARNING:2025-04-13 19:21:42,367:jax._src.xla_bridge:967: An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
Out[4]: <Array [True, True, True, True, True] type='5 * bool'>
```

ikrommyd:
Needs more work and I'd appreciate any help @ianna @pfackeldey.
I'm adding a test that tests all the reducers via parametrization.

Definitely needs https://github.com/scikit-hep/awkward/pull/3457

This PR should fix https://github.com/scikit-hep/awkward/issues/3456, https://github.com/scikit-hep/awkward/issues/3462, https://github.com/scikit-hep/awkward/issues/3463 and https://github.com/scikit-hep/awkward/issues/3465

### DIFF
--- a/src/awkward/_connect/jax/reducers.py
+++ b/src/awkward/_connect/jax/reducers.py
@@ -154,7 +154,7 @@ class ArgMax(JAXReducer):
 @overloads(_reducers.Count)
 class Count(JAXReducer):
     name: Final = "count"
-    preferred_dtype: Final = np.float64
+    preferred_dtype: Final = np.int64
     needs_position: Final = False
 
     @classmethod
@@ -177,6 +177,7 @@ class Count(JAXReducer):
         assert isinstance(array, ak.contents.NumpyArray)
         result = jax.numpy.ones_like(array.data, dtype=array.dtype)
         result = jax.ops.segment_sum(result, parents.data, outlength)
+        result = jax.numpy.asarray(result, dtype=self.preferred_dtype)
 
         if np.issubdtype(array.dtype, np.complexfloating):
             return ak.contents.NumpyArray(
@@ -208,7 +209,7 @@ def segment_count_nonzero(data, segment_ids, num_segments):
 @overloads(_reducers.CountNonzero)
 class CountNonzero(JAXReducer):
     name: Final = "count_nonzero"
-    preferred_dtype: Final = np.float64
+    preferred_dtype: Final = np.int64
     needs_position: Final = False
 
     @classmethod
--- a/src/awkward/_connect/jax/reducers.py
+++ b/src/awkward/_connect/jax/reducers.py
@@ -37,6 +37,69 @@ class JAXReducer(Reducer):
         raise NotImplementedError
 
 
+def awkward_JAXArray_reduce_adjust_starts_64(toptr, outlength, parents, starts):
+    if outlength == 0 or parents.size == 0:
+        return toptr
+    sub_toptr = toptr[:outlength]
+    identity = jax.numpy.astype(jax.numpy.iinfo(jax.numpy.int64).max, toptr.dtype)
+    valid = sub_toptr != identity
+    safe_sub_toptr = jax.numpy.where(valid, sub_toptr, 0)
+    safe_sub_toptr_int = jax.numpy.astype(safe_sub_toptr, jax.numpy.int64)
+    parent_indices = parents[safe_sub_toptr_int]
+    adjustments = starts[jax.numpy.astype(parent_indices, jax.numpy.int64)]
+    updated = jax.numpy.where(valid, sub_toptr - adjustments, sub_toptr)
+    return jax.numpy.concatenate([updated, toptr[outlength:]])
+
+
+def awkward_JAXArray_reduce_adjust_starts_shifts_64(
+    toptr, outlength, parents, starts, shifts
+):
+    if outlength == 0 or parents.size == 0:
+        return toptr
+    sub_toptr = toptr[:outlength]
+    identity = jax.numpy.astype(jax.numpy.iinfo(jax.numpy.int64).max, toptr.dtype)
+    valid = sub_toptr != identity
+    safe_sub_toptr = jax.numpy.where(valid, sub_toptr, 0)
+    safe_sub_toptr_int = jax.numpy.astype(safe_sub_toptr, jax.numpy.int64)
+    parent_indices = parents[safe_sub_toptr_int]
+    delta = (
+        shifts[safe_sub_toptr_int]
+        - starts[jax.numpy.astype(parent_indices, jax.numpy.int64)]
+    )
+    updated = jax.numpy.where(valid, sub_toptr + delta, sub_toptr)
+    return jax.numpy.concatenate([updated, toptr[outlength:]])
+
+
+def apply_positional_corrections(
+    reduced: ak.contents.NumpyArray,
+    parents: ak.index.Index,
+    starts: ak.index.Index,
+    shifts: ak.index.Index | None,
+) -> ak._nplikes.ArrayLike:
+    if shifts is None:
+        assert (
+            parents.nplike is reduced.backend.nplike
+            and starts.nplike is reduced.backend.nplike
+        )
+        return awkward_JAXArray_reduce_adjust_starts_64(
+            reduced.data, reduced.length, parents.data, starts.data
+        )
+
+    else:
+        assert (
+            parents.nplike is reduced.backend.nplike
+            and starts.nplike is reduced.backend.nplike
+            and shifts.nplike is reduced.backend.nplike
+        )
+        return awkward_JAXArray_reduce_adjust_starts_shifts_64(
+            reduced.data,
+            reduced.length,
+            parents.data,
+            starts.data,
+            shifts.data,
+        )
+
+
 def segment_argmin(data, segment_ids, num_segments):
     """
     Applies a segmented argmin-style reduction.
@@ -68,7 +131,7 @@ def segment_argmin(data, segment_ids, num_segments):
 class ArgMin(JAXReducer):
     name: Final = "argmin"
     needs_position: Final = True
-    preferred_dtype: Final = np.int64
+    preferred_dtype: Final = np.float64
 
     @classmethod
     def from_kernel_reducer(cls, reducer: Reducer) -> Self:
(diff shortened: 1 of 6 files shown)


ikrommyd:
I'm currently skipping the argmin argmax tests because of https://github.com/scikit-hep/awkward/issues/3463 but that should change.

ikrommyd:
https://github.com/scikit-hep/awkward/blob/c93da2f8bf88074068a50a1e5fe52c72857bb767/src/awkward/_connect/jax/reducers.py#L296-L299

So we can't take the product of an array with negative numbers? The logarithm will just NaN the output

ikrommyd:
> https://github.com/scikit-hep/awkward/blob/c93da2f8bf88074068a50a1e5fe52c72857bb767/src/awkward/_connect/jax/reducers.py#L296-L299
> 
> So we can't take the product of an array with negative numbers? The logarithm will just NaN the output

I may have fixed that in https://github.com/scikit-hep/awkward/pull/3464/commits/9874810d9f1686e3f4a3402ca522be74fe601334. I still haven't made `ak.any` to behave like the cpu case when segments have zeros though.

ianna:
@ikrommyd - impressive! There are only 5 tests that fails: 3 for `ak.any`, 1 for `ak.count`, and 1 for `ak.sum`. It looks like the latter is failing with a boolean `dtype`. Perhaps, return should make an int?

src/awkward/_connect/jax/reducers.py
```diff
@@ -261,7 +328,7 @@ def apply(
         if array.dtype.kind == "M":
             raise TypeError(f"cannot compute the sum (ak.sum) of {array.dtype!r}")
```
ianna, line 327: @ikrommyd - if we want to allow sum of boolean, I think, we should view the data as integers here:

```python
    if array.dtype == np.bool_:
        data = array.data.astype(jax.numpy.int32)
    else:
        data = array.data

```

ianna (commented):
Thanks for looking into it!

src/awkward/_connect/jax/reducers.py
```diff
@@ -68,7 +131,7 @@ def segment_argmin(data, segment_ids):
 class ArgMin(JAXReducer):
     name: Final = "argmin"
     needs_position: Final = True
-    preferred_dtype: Final = np.int64
+    preferred_dtype: Final = np.float64
```
ianna: `argmin` returns the index (i.e. position) of the minimum value. Indices are always integers, not floating-point numbers.

src/awkward/_connect/jax/reducers.py
```diff
@@ -125,7 +191,7 @@ def segment_argmax(data, segment_ids):
 class ArgMax(JAXReducer):
     name: Final = "argmax"
     needs_position: Final = True
-    preferred_dtype: Final = np.int64
+    preferred_dtype: Final = np.float64
```
ianna: the same argument here - indices are always integers, not floating-point numbers.

ikrommyd:
So I made ci pass but there are a few things that are to be done for sure
1. The code needs refactoring, I don't like how it looks at all, it's very hacky
2. The reducer tests should definitely test more input array cases
3. We need to make sure I'm not breaking something that should work but is untested due to the smaller amount of tests for the jax backend

ianna (commented):
@ikrommyd - just some minor comments. I agree, JAX backend needs to be thoroughly tested. Perhaps, our fellow could take over? @pfackeldey - when do we discuss his project? Thanks.

ianna (approved):
@ikrommyd - Great! Thanks! The tests pass, I'm merging it now.

PR #3469, opened 2025-04-14
fix: single int slicing for named axis

pfackeldey:
Fixes single int slicing with named axis, and improves robustness of named axis attachment in case all named axes are gone.

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -533,7 +533,7 @@ class Content(Meta):
         if is_integer_like(where):
             # propagate named_axis to output
             named_axis.mapping = _remove_named_axis(
-                named_axis.mapping, where, self.purelist_depth
+                named_axis.mapping, 0, self.purelist_depth
             )
             return self._getitem_at(ak._slicing.normalize_integer_like(where))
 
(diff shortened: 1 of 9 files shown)


pfackeldey:
This fixes now in addition primitive promotion rules for the typetracer:
now this works:
```python
import awkward as ak
import numpy as np

ak.to_layout(ak._nplikes.typetracer.TypeTracerArray._new(np.dtype(np.int64), ()), primitive_policy="pass-through")
# TypeTracerArray(dtype('int64'), shape=())

ak.to_layout(ak._nplikes.typetracer.TypeTracerArray._new(np.dtype(np.int64), ()), primitive_policy="promote")
# <NumpyArray dtype='int64' len='1'>[## ... ##]</NumpyArray>

ak.to_layout(ak._nplikes.typetracer.TypeTracerArray._new(np.dtype(np.int64), ()), primitive_policy="error")
# TypeError: Encountered a scalar (TypeTracerArray), but scalar conversion/promotion is disabled
```
on main the 0-dim typetracer array would've been always promoted to a length 1 layout - it would ignore whatever `primitive_policy` is set.

ianna (approved):
@pfackeldey - Looks good to me! Thanks! Please, go ahead and merge it.

PR #3471, opened 2025-04-15
fix: add `ptr` implementation and fix `__setitem__` for `Index` and `tocarryraw` to make `ak.(arg)combinations` work for the jax backend

### LINKED ISSUE #3467: `ak.(arg)combinations` doesn't work for the jax backend
`ak.(arg)combinations` errors because even though the kernel is there, `.ptr` is not available for the jax backend. Is it possible to do something about that?
```
File ~/Dropbox/work/coffea_dev/awkward/src/awkward/_do.py:142, in combinations(layout, n, replacement, axis, fields, parameters)
    140     if len(recordlookup) != n:
    141         raise ValueError("if provided, the length of 'fields' must be 'n'")
--> 142 return layout._combinations(n, replacement, recordlookup, parameters, axis, 1)

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/contents/listoffsetarray.py:1360, in ListOffsetArray._combinations(self, n, replacement, recordlookup, parameters, axis, depth)
   1358 posaxis = maybe_posaxis(self, axis, depth)
   1359 if posaxis is not None and posaxis + 1 == depth:
-> 1360     return self._combinations_axis0(n, replacement, recordlookup, parameters)
   1361 elif posaxis is not None and posaxis + 1 == depth + 1:
   1362     if (
   1363         self.parameter("__array__") == "string"
   1364         or self.parameter("__array__") == "bytestring"
   1365     ):

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/contents/content.py:953, in Content._combinations_axis0(self, n, replacement, recordlookup, parameters)
    951     tocarry.append(ptr)
    952     if self._backend.nplike.known_data:
--> 953         tocarryraw[i] = ptr.ptr
    955 toindex = Index64.empty(n, self._backend.nplike, dtype=np.int64)
    956 fromindex = Index64.empty(n, self._backend.nplike, dtype=np.int64)

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/index.py:146, in Index.ptr(self)
    144     return 0
    145 else:
--> 146     raise NotImplementedError(
    147         f"this function hasn't been implemented for the {type(self._nplike).__name__} backend"
    148     )

NotImplementedError: this function hasn't been implemented for the Jax backend
```

ikrommyd:
Fixes https://github.com/scikit-hep/awkward/issues/3467

While writing this, I found a couple of more places where we had to use the `.at[where].set(value)` syntax for jax and also I noticed that in a new places `tocarryraw` is being instantiated as `ak.index.Index.empty(n, dtype=np.intp, nplike=self._backend.nplike)` while in others as `self._backend.nplike.empty(n, dtype=np.intp)`  for no apparent reason.
I decided to make it consistent and use an `ak.index.Index` everywhere cause that makes `__setitem__` calls on the index work instead of having to add an `if` statement to do `.at[where].set(value)` everywhere for `tocarryraw`.

### DIFF
--- a/src/awkward/contents/content.py
+++ b/src/awkward/contents/content.py
@@ -940,7 +940,7 @@ class Content(Meta):
                     combinationslen = combinationslen * (size - j + 1)
                     combinationslen = combinationslen // j
 
-        tocarryraw = self._backend.nplike.empty(n, dtype=np.intp)
+        tocarryraw = ak.index.Index.empty(n, dtype=np.intp, nplike=self._backend.nplike)
         tocarry = []
         for i in range(n):
             ptr = Index64.empty(
@@ -966,7 +966,7 @@ class Content(Meta):
                 toindex.data.dtype.type,
                 fromindex.data.dtype.type,
             ](
-                tocarryraw,
+                tocarryraw.data,
                 toindex.data,
                 fromindex.data,
                 n,
--- a/src/awkward/contents/regulararray.py
+++ b/src/awkward/contents/regulararray.py
@@ -912,7 +912,7 @@ class RegularArray(RegularMeta[Content], Content):
                     combinationslen = combinationslen // j
 
             totallen = combinationslen * self._length
-            tocarryraw = nplike.empty(n, dtype=np.intp)
+            tocarryraw = ak.index.Index.empty(n, dtype=np.intp, nplike=nplike)
             tocarry = []
             for i in range(n):
                 ptr = ak.index.Index64.empty(
@@ -936,7 +936,7 @@ class RegularArray(RegularMeta[Content], Content):
                         toindex.data.dtype.type,
                         fromindex.data.dtype.type,
                     ](
-                        tocarryraw,
+                        tocarryraw.data,
                         toindex.data,
                         fromindex.data,
                         n,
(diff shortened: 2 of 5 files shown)


ianna (commented):
@ikrommyd - Great! Thanks for fixing it. Though, it seems to me that this PR is a bugfix rather than a feature :-)

ikrommyd:
> @ikrommyd - Great! Thanks for fixing it. Though, it seems to me that this PR is a bugfix rather than a feature :-)

You're right! I changed the title. Feel free to merge if ci passes after merging main to it.

I originally thought that supporting combinations for the jax backend was a "feature" but then I realised that the fact they they weren't working is only because the `Index` is a `jnp.ndarray` after https://github.com/scikit-hep/awkward/pull/3446 and it used to work before it since `ptr` was implemented fine for `np.ndarray`

ikrommyd:
@all-contributors please add @krommyd for test

ikrommyd:
@all-contributors please add @ikrommyd for test

ianna (approved):
@ikrommyd - Looks great! Thanks. I'm merging it.

PR #3475, opened 2025-04-21
feat: support `unknown_length` for virtual arrays in order to read without any materialization

ikrommyd:
This PR introduces `unknown_length` support for virtual arrays.

Because multiple arrays can have the same shape (coming from the same offsets for example) and because the awkward codebase wants us to know shapes/length very often, the most consistent way to add `unknown_length` support is via introducing a separate shape generator for virtual arrays that returns a shape tuple when called. This is in order to be able to generate the shape of something without generating its data.

We make the distinction between private `._shape` and `._length` versus public `.shape` and `.length` properties.
The public ones materialize the shape while the private ones don't. We extend this logic to the layouts and for the layouts that define a private `self._length` property as a function of the content, we instantiate that with `unknown_length` in the virtual array case in the `__init__` method and we actually calculate it the first time `.length` of that layout is called in order to be able to instantiate layouts without materializing shapes. We also avoid materializing shapes for the `__repr__` of the layouts.
To make our life easier, we introduce two helper utils `maybe_shape_of` and `maybe_length_of`.

Finally, we make the necessary changes in `from_buffers` in order to be able to to construct the proper data and shape generators to pass down to the `VirtualArray` buffers.

This has been tested through coffea as well with the ADL benchmarks, the coffea processors example and the AGC where we observe no materialization when reading with nanoevents and proper materialization of exactly the right buffers when running the analyses snippets.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -66,7 +66,7 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             return obj
         if isinstance(obj, VirtualArray):
             if obj.is_materialized:
-                obj = obj.materialize()
+                obj = obj.materialize_data()
             else:
                 if obj.dtype == dtype or dtype is None:
                     return obj
@@ -75,7 +75,7 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
                         obj.nplike,
                         obj.shape,
                         dtype,
-                        lambda: self.asarray(obj.materialize(), dtype=dtype),
+                        lambda: self.asarray(obj.materialize_data(), dtype=dtype),
                     )
         if copy:
             return self._module.array(obj, dtype=dtype, copy=True)
@@ -96,13 +96,13 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             return x
         elif isinstance(x, VirtualArray):
             if x.is_materialized:
-                return self._module.ascontiguousarray(x.materialize())
+                return self._module.ascontiguousarray(x.materialize_data())
             else:
                 return VirtualArray(
                     x.nplike,
                     x.shape,
                     x.dtype,
-                    lambda: self._module.ascontiguousarray(x.materialize()),
+                    lambda: self._module.ascontiguousarray(x.materialize_data()),
                 )
         else:
             return self._module.ascontiguousarray(x)
@@ -354,10 +354,10 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
                     self,
                     next_shape,
                     x.dtype,
-                    lambda: self.reshape(x.materialize(), next_shape),  # type: ignore[union-attr]
+                    lambda: self.reshape(x.materialize_data(), next_shape),  # type: ignore[union-attr]
                 )
             else:
-                x = x.materialize()  # type: ignore[assignment]
+                x = x.materialize_data()  # type: ignore[assignment]
 
         if copy is None:
             return self._module.reshape(x, shape)
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -66,16 +66,17 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             return obj
         if isinstance(obj, VirtualArray):
             if obj.is_materialized:
-                obj = obj.materialize_data()
+                obj = obj.materialize()
             else:
                 if obj.dtype == dtype or dtype is None:
                     return obj
                 else:
                     return VirtualArray(
-                        obj.nplike,
-                        obj.shape,
+                        obj._nplike,
+                        obj._shape,
                         dtype,
-                        lambda: self.asarray(obj.materialize_data(), dtype=dtype),
+                        lambda: self.asarray(obj.materialize(), dtype=dtype),
+                        lambda: obj.shape,
                     )
         if copy:
             return self._module.array(obj, dtype=dtype, copy=True)
@@ -96,13 +97,14 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             return x
         elif isinstance(x, VirtualArray):
             if x.is_materialized:
-                return self._module.ascontiguousarray(x.materialize_data())
+                return self.ascontiguousarray(x.materialize())  #  type: ignore[arg-type]
             else:
                 return VirtualArray(
-                    x.nplike,
-                    x.shape,
-                    x.dtype,
-                    lambda: self._module.ascontiguousarray(x.materialize_data()),
+                    x._nplike,
+                    x._shape,
+                    x._dtype,
+                    lambda: self.ascontiguousarray(x.materialize()),  # type: ignore[arg-type]
+                    lambda: x.shape,
                 )
         else:
             return self._module.ascontiguousarray(x)
@@ -354,10 +356,11 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
                     self,
                     next_shape,
                     x.dtype,
-                    lambda: self.reshape(x.materialize_data(), next_shape),  # type: ignore[union-attr]
+                    lambda: self.reshape(x.materialize(), next_shape),  # type: ignore[union-attr]
+                    lambda: next_shape,
                 )
             else:
-                x = x.materialize_data()  # type: ignore[assignment]
+                x = x.materialize()  # type: ignore[assignment]
 
         if copy is None:
             return self._module.reshape(x, shape)
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -794,9 +794,7 @@ class TypeTracer(NumpyLike[TypeTracerArray]):
                         "asarray was called with copy=False for an array of a different dtype"
                     )
                 else:
-                    return TypeTracerArray._new(
-                        obj.dtype, ak._util.non_materializing_shape_of(obj)
-                    )
+                    return TypeTracerArray._new(obj.dtype, ak._util.maybe_shape_of(obj))
             # Python objects
             elif isinstance(obj, (Number, bool)):
                 as_array = numpy.asarray(obj)
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -276,7 +276,7 @@ class TypeTracerArray(NDArrayOperatorsMixin, ArrayLike):
         if cls.runtime_typechecks:
             if not isinstance(shape, tuple):
                 raise TypeError("typetracer shape must be a tuple")
-            if not all(isinstance(x, int) or x is unknown_length for x in shape):
+            if not all(is_integer(x) or x is unknown_length for x in shape):
                 raise TypeError("typetracer shape must be integers or unknown-length")
             if not isinstance(dtype, np.dtype):
                 raise TypeError("typetracer dtype must be an instance of np.dtype")
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -783,7 +783,10 @@ class TypeTracer(NumpyLike[TypeTracerArray]):
                 obj = numpy.asarray(obj)
 
             # Support array-like objects
-            if hasattr(obj, "shape") and hasattr(obj, "dtype"):
+            # Check for ._shape first because .shape might materialize
+            if (hasattr(obj, "_shape") or hasattr(obj, "shape")) and hasattr(
+                obj, "dtype"
+            ):
                 if obj.dtype.kind == "S":
                     raise TypeError("TypeTracerArray cannot be created from strings")
                 elif copy is False and dtype != obj.dtype:
@@ -791,7 +794,9 @@ class TypeTracer(NumpyLike[TypeTracerArray]):
                         "asarray was called with copy=False for an array of a different dtype"
                     )
                 else:
-                    return TypeTracerArray._new(obj.dtype, obj.shape)
+                    return TypeTracerArray._new(
+                        obj.dtype, ak._util.non_materializing_shape_of(obj)
+                    )
             # Python objects
             elif isinstance(obj, (Number, bool)):
                 as_array = numpy.asarray(obj)
--- a/src/awkward/_nplikes/virtual.py
+++ b/src/awkward/_nplikes/virtual.py
@@ -91,6 +91,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
     @property
     def size(self) -> ShapeItem:
         size: ShapeItem = 1
+        self.get_shape()
         for item in self._shape:
             size *= item
         return size
@@ -103,7 +104,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
     def strides(self) -> tuple[ShapeItem, ...]:
         return self.materialize().strides  # type: ignore[attr-defined]
 
-    def get_shape(self) -> Self:
+    def get_shape(self) -> None:
         if any(dim is unknown_length for dim in self._shape):
             if self._shape_generator is not None:
                 shape = self._shape_generator()
@@ -119,7 +120,6 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
                         f"{type(self).__name__} had shape {self._shape} before materialization while the materialized array has shape {shape}"
                     )
             self._shape = shape
-        return self
 
     def materialize(self) -> ArrayLike:
         if self._array is UNMATERIALIZED:
@@ -167,6 +167,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
 
     def view(self, dtype: DTypeLike) -> Self:
         dtype = np.dtype(dtype)
+        self.get_shape()
 
         if self._array is not UNMATERIALIZED:
             return self.materialize().view(dtype)  # type: ignore[return-value]
@@ -195,6 +196,10 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
     @property
     def generator(self) -> Callable:
         return self._generator
+
+    @property
+    def shape_generator(self) -> Callable | None:
+        return self._shape_generator
 
     @property
     def nplike(self) -> NumpyLike:
--- a/src/awkward/_nplikes/virtual.py
+++ b/src/awkward/_nplikes/virtual.py
@@ -27,7 +27,7 @@ def materialize_if_virtual(*args: Any) -> tuple[Any, ...]:
     A little helper function to materialize all virtual arrays in a list of arrays.
     """
     return tuple(
-        arg.materialize_data() if isinstance(arg, VirtualArray) else arg for arg in args
+        arg.materialize() if isinstance(arg, VirtualArray) else arg for arg in args
     )
 
 
@@ -49,7 +49,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
         nplike: NumpyLike,
         shape: tuple[ShapeItem, ...],
         dtype: DTypeLike,
-        data_generator: Callable[[], ArrayLike],
+        generator: Callable[[], ArrayLike],
         shape_generator: Callable[[], tuple[ShapeItem, ...]] | None = None,
     ) -> None:
         if not nplike.supports_virtual_arrays:
@@ -66,7 +66,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
         self._shape = shape
         self._dtype = np.dtype(dtype)
         self._array: Sentinel | ArrayLike = UNMATERIALIZED
-        self._data_generator = data_generator
+        self._generator = generator
         self._shape_generator = shape_generator
 
     @property
@@ -75,8 +75,14 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
 
     @property
     def shape(self) -> tuple[ShapeItem, ...]:
-        self.materialize_shape()
+        self.get_shape()
         return self._shape
+
+    @property
+    def inner_shape(self) -> tuple[ShapeItem, ...]:
+        if len(self._shape) > 1:
+            self.get_shape()
+        return self._shape[1:]
 
     @property
     def ndim(self) -> int:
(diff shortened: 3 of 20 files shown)


src/awkward/_nplikes/virtual.py
```diff
@@ -66,51 +69,89 @@ def __init__(
...
+        for item in self._shape:
+            size *= item
+        return size * self._dtype.itemsize
 
     @property
     def strides(self) -> tuple[ShapeItem, ...]:
         return self.materialize().strides  # type: ignore[attr-defined]
 
+    def get_shape(self) -> None:
+        if any(dim is unknown_length for dim in self._shape):
```
pfackeldey, line 112: Maybe use the property directly?
```suggestion
        if not self.has_known_shape:
```

src/awkward/_nplikes/virtual.py
```diff
@@ -66,51 +69,89 @@ def __init__(
...
+            if self._shape_generator is not None:
+                shape = self._shape_generator()
+            else:
+                shape = self.materialize().shape
+            if len(shape) != len(self._shape):
+                raise TypeError(
```
pfackeldey: I think this should be a `ValueError` instead

src/awkward/_nplikes/virtual.py
```diff
@@ -66,51 +69,89 @@ def __init__(
...
+                    f"{type(self).__name__} had shape {self._shape} before materialization while the materialized array has shape {shape}"
+                )
+            for expected_dim, actual_dim in zip(self._shape, shape):
+                if expected_dim is not unknown_length and expected_dim != actual_dim:
+                    raise TypeError(
```
pfackeldey: I think this should be a `ValurError` instead

pfackeldey: and maybe check that all elements of `shape` are integers, i.e. `if not all(isinteger(dim) for dim in shape): raise ValueError(...)` ? (then the following `tuple(map(int, shape))` will also always work)

src/awkward/_nplikes/virtual.py
```diff
@@ -66,51 +69,89 @@ def __init__(
...
+                    )
+            self._shape = tuple(map(int, shape))
+
     def materialize(self) -> ArrayLike:
         if self._array is UNMATERIALIZED:
-            array = self._nplike.asarray(self.generator())
-            if self._shape != array.shape:
+            array = self._nplike.asarray(self._generator())
+            if len(self._shape) != len(array.shape):
                 raise TypeError(
```
pfackeldey: see comment above

src/awkward/_nplikes/virtual.py
```diff
@@ -66,51 +69,89 @@ def __init__(
...
                     f"{type(self).__name__} had shape {self._shape} before materialization while the materialized array has shape {array.shape}"
                 )
+            for expected_dim, actual_dim in zip(self._shape, array.shape):
+                if expected_dim is not unknown_length and expected_dim != actual_dim:
+                    raise TypeError(
+                        f"{type(self).__name__} had shape {self._shape} before materialization while the materialized array has shape {array.shape}"
+                    )
             if self._dtype != array.dtype:
                 raise TypeError(
```
pfackeldey: see comment above

src/awkward/_nplikes/virtual.py
```diff
@@ -133,29 +175,35 @@ def view(self, dtype: DTypeLike) -> Self:
...
             if remainder != 0:
                 raise ValueError(
                     "new size of array with larger dtype must be a "
                     "divisor of the total size in bytes (of the last axis of the array)"
                 )
-            shape = self._shape[:-1] + (last,)
+            shape = self.shape[:-1] + (last,)
         else:
-            shape = self._shape
+            shape = self.shape
```
pfackeldey, line 189: Is this supposed to materialize the shape? It doesn't have to materialize the array either...
maybe this can be lazy, i.e.:
```python
        return type(self)(
            self._nplike,
            self._shape,
            dtype,
            lambda: self.materialize().view(dtype),
            self.shape_generator,
        )
```
?

src/awkward/_nplikes/virtual.py
```diff
@@ -133,29 +175,35 @@ def view(self, dtype: DTypeLike) -> Self:
...
             dtype,
             lambda: self.materialize().view(dtype),
+            lambda: shape,
         )
 
     @property
     def generator(self) -> Callable:
         return self._generator
 
+    @property
```
pfackeldey: What's the purpose of this? I'm not seeing it used which would indicate that you'd like to guard it as read-only.

src/awkward/_nplikes/virtual.py
```diff
@@ -173,7 +221,7 @@ def tolist(self) -> list:
     @property
     def ctypes(self):
         if isinstance((self._nplike), ak._nplikes.cupy.Cupy):
-            raise AttributeError("Cupy ndarrays do not have a ctypes attribute.")
+            raise AttributeError("Cupy ndarrays do not have a ctypes attribute")
```
pfackeldey, line 216: This doesn't have something to do with this PR, but it's a simple change that I don't mind either. (There are a few of these cases in this PR)

src/awkward/_nplikes/virtual.py
```diff
@@ -223,7 +274,7 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
 
     def __repr__(self):
         dtype = repr(self._dtype)
-        if self.shape is None:
+        if self._shape is None:
```
pfackeldey, line 269: side question: when is `self._shape` ever `None` ?

src/awkward/contents/listarray.py
```diff
@@ -142,7 +142,12 @@ def __init__(self, starts, stops, content, *, parameters=None):
             raise TypeError(
                 f"{type(self).__name__} 'content' must be a Content subtype, not {content!r}"
             )
-        if content.backend.nplike.known_data and starts.length > stops.length:
+        if (
+            content.backend.nplike.known_data
+            and ak._util.non_materializing_length_of(starts) is not unknown_length
```
pfackeldey: Does adding these conditions introduce some edge cases that are unwanted (i.e. where this if condition is skipped)?

src/awkward/contents/recordarray.py
```diff
@@ -364,6 +374,10 @@ def _touch_shape(self, recursive: bool):
 
     @property
     def length(self) -> ShapeItem:
+        if self._backend.nplike.known_data and self._length is unknown_length:
+            self._length = _calculate_recordarray_length(
```
pfackeldey, line 379: This is always turning `self._length` into an int, so `self.length` is "shape-materializing", correct?

If not: this may introduce some overhead - would be good to know if that can reduced/or if it's irrelevant.

src/awkward/contents/regulararray.py
```diff
@@ -249,6 +264,10 @@ def _touch_shape(self, recursive: bool):
 
     @property
     def length(self) -> ShapeItem:
+        if self._backend.nplike.known_data and self._length is unknown_length:
+            self._length = _calculate_regulararray_length(
```
pfackeldey, line 268: same comment as above

pfackeldey (changes requested):

src/awkward/_nplikes/virtual.py
ikrommyd: I was kinda following Angus' comment from last time to try to do such things explicitly within the virtual array class. I'll actually remove `has_known_shape` because it's not used anywhere.

src/awkward/_nplikes/virtual.py
ikrommyd: I was just doing the same thing here that we did for the array generator. Just define a property for public access.

src/awkward/_nplikes/virtual.py
ikrommyd: I have no idea...typetracer and placeholder had this so I just copied it from there in the original implementation.

src/awkward/_nplikes/virtual.py
ikrommyd: This was just OCD 🤣

src/awkward/_nplikes/virtual.py
ikrommyd: Technically we can just remove both because I don't think they're used.

src/awkward/_nplikes/virtual.py
pfackeldey: Ok, then please go ahead and remove it :+1:

src/awkward/_nplikes/virtual.py
pfackeldey: Then, please remove them. There's no point in having obsolete code that's not used.

src/awkward/contents/listarray.py
ikrommyd: I don't think so. The only purpose of adding this was to avoid doing `starts.length > stops.length` because that would error with "cannot order unknown lengths". The only problem I see here is if indeed someone tries to instantiate this and it gets through with unknown length but indeed `starts.length > stops.length` is True for the materialized lengths. I don't really see that happening in user code though unless something's up.

src/awkward/contents/recordarray.py
ikrommyd: `RecordArray.length` should be materializing the length. That's the whole point here. It's going to run once because `self._length` will be replaced with an integer so the only extra python overhead is running the if statement every time `.length` is called but it won't do any calculations because it will be `False`. Also this will never run for typetracers because the backend doesn't have known data.

src/awkward/contents/regulararray.py
ikrommyd: Replied in the `RecordArray` case.

src/awkward/contents/recordarray.py
pfackeldey: should we maybe then ensure with a check that `self._length` is a known value (i.e. `assert isinteger(self._length)` ?

src/awkward/_nplikes/virtual.py
pfackeldey: Hm ok, I guess it's good to be consistent with those - wonder what the reason is though.

src/awkward/contents/recordarray.py
ikrommyd: You also want known data I think. Technically you can have a typetracing recordarray and give it an integer length I believe.

src/awkward/_nplikes/virtual.py
ikrommyd: I mean....sure but there is no real use case for this. We'll never have a single element virtual array and the other branch `if len(self.shape) >= 1` materializes the shape. I just did `self.shape` here for code consistency within the method. Even if you do have a single element virtual array, it's not gonna have `unknown_length`. It's a single element.

src/awkward/contents/recordarray.py
pfackeldey: right, I mean that we should add this check after calling `_calculate_recordarray_length`.

ikrommyd: Oh I thought you meant instead of the condition I'm checking now. Yup sure!

src/awkward/_nplikes/virtual.py
pfackeldey: Ok :+1:

src/awkward/contents/listarray.py
ikrommyd: Maybe we wanna redo this check the first time `.length` is called? `self.length` is just `return self._starts.length`. We could check this condition with the stops' length there as well but we'll need a variable to keep track we've already done it so that we don't do it every time.

pfackeldey: Let's keep it as it is right now :+1:

src/awkward/_nplikes/array_module.py
```diff
@@ -96,13 +97,14 @@ def ascontiguousarray(
             return x
         elif isinstance(x, VirtualArray):
             if x.is_materialized:
-                return self._module.ascontiguousarray(x.materialize())
+                return self.ascontiguousarray(x.materialize())  #  type: ignore[arg-type]
```
ianna, line 100: Is `self._module` needed for special handling for a specific module (like `numpy` or `cupy`)? Would changing it affect how this works for other modules?

src/awkward/_nplikes/typetracer.py
```diff
@@ -276,7 +276,7 @@ def _new(
         if cls.runtime_typechecks:
             if not isinstance(shape, tuple):
                 raise TypeError("typetracer shape must be a tuple")
-            if not all(isinstance(x, int) or x is unknown_length for x in shape):
+            if not all(is_integer(x) or x is unknown_length for x in shape):
```
ianna, line 279: do we need to be more generic here?

src/awkward/_util.py
```diff
@@ -117,3 +118,40 @@ def copy_behaviors(from_name: str, to_name: str, behavior: dict):
                 output[new_tuple] = value
 
     return output
+
+
+def non_materializing_shape_of(
```
ianna: perhaps, `lazy_shape_of`?

src/awkward/_util.py
```diff
@@ -117,3 +118,40 @@ def copy_behaviors(from_name: str, to_name: str, behavior: dict):
...
+    obj: ak._nplikes.ArrayLike,
+) -> tuple[ak._nplikes.shape.ShapeItem, ...]:
+    if isinstance(obj, ak._nplikes.virtual.VirtualArray):
+        return obj._shape
+    else:
+        return obj.shape
+
+
+def non_materializing_length_of(
```
ianna: `lazy_lenght_of`?

src/awkward/_nplikes/virtual.py
```diff
@@ -66,45 +69,83 @@ def __init__(
...
-        for item in self._shape:
+        for item in self.shape:
             size *= item
         return size
 
     @property
     def nbytes(self) -> ShapeItem:
-        return self.size * self._dtype.itemsize
+        size: ShapeItem = 1
+        for item in self._shape:
```
ianna, line 103: can `unknown_length` appear in `_shape`?

pfackeldey:
I'm semi-convinced by the rename in https://github.com/scikit-hep/awkward/pull/3475/commits/9a2c4e47f93a212716018e9bfe533ee40ea41f91.
`maybeXYZ` variables/functions suggest me typically that they return `XYZ` or `None` (similar to Rust's `Option` Monad).
The point of this util is different: it _always_ returns a shape, the question is whether it has to materialize it first. 
Maybe @ianna has a good naming suggestion instead?

src/awkward/_nplikes/virtual.py
```diff
@@ -125,6 +166,7 @@ def T(self):
             self._shape[::-1],
             self._dtype,
             lambda: self.materialize().T,
+            lambda: self.shape[::-1],
```
ianna, line 169: Using lambdas in the constructor is fine for a small private class but might benefit from named local functions if these get complex or if debugging deferred evaluations is needed. Also, if either `.materialize()` or `.shape` can fail (for example if `self` is invalid or incomplete), please, make sure your class is robust to deferred evaluation errors.

src/awkward/_util.py
```diff
@@ -117,3 +118,40 @@ def copy_behaviors(from_name: str, to_name: str, behavior: dict):
                 output[new_tuple] = value
 
     return output
+
+
+def maybe_shape_of(
```
ianna, line 123: It could use a docstring for clarity

src/awkward/_util.py
```diff
@@ -117,3 +118,40 @@ def copy_behaviors(from_name: str, to_name: str, behavior: dict):
...
+    obj: ak._nplikes.ArrayLike,
+) -> tuple[ak._nplikes.shape.ShapeItem, ...]:
+    if isinstance(obj, ak._nplikes.virtual.VirtualArray):
+        return obj._shape
+    else:
+        return obj.shape
+
+
+def maybe_length_of(
```
ianna, line 138: It could use a docstring for clarity — what’s the expected behavior and why it's `maybe_*`? (because it depends on the type's semantics)

src/awkward/_util.py
```diff
@@ -117,3 +118,40 @@ def copy_behaviors(from_name: str, to_name: str, behavior: dict):
...
+    obj: ak.contents.Content | ak.index.Index,
+) -> ak._nplikes.shape.ShapeItem:
+    if isinstance(obj, (ak.contents.NumpyArray, ak.index.Index)):
```
ianna, line 148: have you considered using `@singledispatch`?

ikrommyd:
> I'm semi-convinced by the rename in [9a2c4e4](https://github.com/scikit-hep/awkward/commit/9a2c4e47f93a212716018e9bfe533ee40ea41f91). `maybeXYZ` variables/functions suggest me typically that they return `XYZ` or `None` (similar to Rust's `Option` Monad). The point of this util is different: it _always_ returns a shape, the question is whether it has to materialize it first. Maybe @ianna has a good naming suggestion instead?

Yeah I also don't like it that much as well. It was the shortest one I could think of. I didn't like `lazy_length/shape_of` because there's nothing lazy about it. Any ideas are welcome @ianna 😄

src/awkward/_util.py
ianna: it would look something like:
```python
from functools import singledispatch
import awkward as ak

@singledispatch
def maybe_length_of(obj) -> ak._nplikes.shape.ShapeItem:
    """Return the length of an Awkward Content or Index object.

    Raises:
        TypeError: if the type of `obj` is not supported.
    """
    raise TypeError(f"Invalid type {type(obj)} for length calculation")

@maybe_length_of.register
def _(obj: ak.contents.NumpyArray) -> ak._nplikes.shape.ShapeItem:
    return maybe_shape_of(obj._data)[0]

@maybe_length_of.register
def _(obj: ak.index.Index) -> ak._nplikes.shape.ShapeItem:
    return maybe_shape_of(obj._data)[0]

@maybe_length_of.register
def _(obj: ak.contents.ListOffsetArray) -> ak._nplikes.shape.ShapeItem:
    return maybe_length_of(obj._offsets) - 1

@maybe_length_of.register
def _(obj: ak.contents.ListArray) -> ak._nplikes.shape.ShapeItem:
    return maybe_length_of(obj._starts)

@maybe_length_of.register
def _(obj: (ak.contents.RecordArray, ak.contents.RegularArray, ak.contents.BitMaskedArray)) -> ak._nplikes.shape.ShapeItem:
    return obj._length

@maybe_length_of.register
def _(obj: ak.contents.ByteMaskedArray) -> ak._nplikes.shape.ShapeItem:
    return maybe_length_of(obj._mask)

@maybe_length_of.register
def _(obj: (ak.contents.IndexedArray, ak.contents.IndexedOptionArray)) -> ak._nplikes.shape.ShapeItem:
    return maybe_length_of(obj._index)

@maybe_length_of.register
def _(obj: ak.contents.UnionArray) -> ak._nplikes.shape.ShapeItem:
    return maybe_length_of(obj._tags)

@maybe_length_of.register
def _(obj: ak.contents.UnmaskedArray) -> ak._nplikes.shape.ShapeItem:
    return maybe_length_of(obj._content)

@maybe_length_of.register
def _(obj: ak.contents.EmptyArray) -> ak._nplikes.shape.ShapeItem:
    return 0
```

ianna (changes requested):
@ikrommyd - excellent work! Please check a few minor comments. As for your question, perhaps `infer_length_of` / `infer_shape_of`?

src/awkward/_util.py
ikrommyd: To agree with Peter in https://github.com/scikit-hep/awkward/pull/3482#discussion_r2060539746, I personally don't like this style as well. I really can't parse that code. I highly prefer to keep it as is with clear recursion python-style. If someone gave me this code, I wouldn't understand what it's doing clearly.

src/awkward/_nplikes/typetracer.py
ikrommyd: Uhmm we allow `np.int64/np.int32` shapes in virtual so I thought I'd make this consistent. This would be the case of a shape coming from an offset. For example `offsets[-1]` in numpy2 is `np.int64` not python `int`.

src/awkward/_nplikes/virtual.py
ikrommyd: Yes. `unknown_length` can only appear in `._shape` actually as `.shape` would materialize it. This is consistent with the typetracer. If any of the shape dimensions are unknown the resulting bytes are unknown because multiplication of unknown length with a number is allowed and gives back unknown length.

src/awkward/_nplikes/array_module.py
ikrommyd: Oh this was actually kind of a bug that we don't really hit. We should use recursion here and not the underlying `_module` function. We do it like that everywhere else. Maybe there is a use case in the future where `x.materialize()` doesn't return a `np.ndarray` but returns something else (placeholder or whatever). Or maybe you have virtual arrays generating virtual arrays for whatever reason. This would error because it would hit the numpy function immediately. We should recurse back to  the nplike `ascontiguousarray` and not the underlying module function for that reason.

src/awkward/_nplikes/virtual.py
ikrommyd: You're right but both `materialize()` and `get_shape` have checks in them to raise proper errors and we can also add `breakpoint`s in them to debug. I don't want to use anything more complicated here but you're right that it's best to define closures with `def` and not `lambdas` in more complicated cases.

ikrommyd: In this case is very clear because I define the new shape as `self._shape[::-1]` so I don't think we want anything longer than that.

src/awkward/_util.py
ikrommyd: Added

src/awkward/_util.py
ikrommyd: Added

src/awkward/_nplikes/virtual.py
ikrommyd: This is the whole point. We report unknown unless we know the shape.
```py
In [1]: from awkward._nplikes.shape import unknown_length

In [2]: unknown_length * 2
Out[2]: awkward._nplikes.shape.unknown_length
```

src/awkward/_nplikes/typetracer.py
ikrommyd: Oh and forgot to mention that for that reason, if something had `np.int` in its shape, turning it into a typetracer would error. I discovered this by trying to do `layout.to_typetracer()` when the `layout` had virtual array buffers.

src/awkward/_nplikes/array_module.py
ianna: What if it is `cupy` or `Jax` backend? Should it return `np. ndarray`?

src/awkward/_nplikes/typetracer.py
ianna: Ok, thanks

src/awkward/_nplikes/array_module.py
ikrommyd: Any `ndarray` is fine (`np.ndarray` was misleading here). They will hit the `self._module.ascontiguousarray(x)` in the `else:` block.

ikrommyd: If you look at `reshape` for example, we do `lambda: self.reshape(x.materialize(), next_shape)` and `self._module.reshape(x, shape)` if it is not a virtual array. This was just a typo in my original implementation that I made consistent here

src/awkward/_util.py
ianna: Most Python developers use if-elif-else chains or explicit type checking for handling different types in functions. This way is a cleaner, more Pythonic way. It allows you to write more elegant, maintainable, and scalable functions without ugly conditionals. The magic of `singledispatch` makes your code DRY, extensible, and Pythonic.

ianna (approved):
@ikrommyd - thanks for addressing the comments! The PR looks great. Thanks.

ianna:
All good! Merging it!

PR #3478, opened 2025-04-24
fix: add field metadata to record its option type 

### LINKED ISSUE #3402: awkward-array doesn't work with pyarrow `19.0.0`

ianna:
Newer versions of pyarrow require that `RecordArray` fields be declared as `nullable`, even if their content is not an `Option` type. This introduces an unnecessary option-type node in the schema.

This PR adds field metadata to track that requirement, so that when converting the field back to an Awkward `RecordArray` field, we can safely remove the dummy option-type node.

fixes https://github.com/scikit-hep/awkward/issues/3402

### DIFF
--- a/src/awkward/_connect/pyarrow/extn_types.py
+++ b/src/awkward/_connect/pyarrow/extn_types.py
@@ -118,7 +118,7 @@ class AwkwardArrowType(pyarrow.ExtensionType):
             metadata["record_is_tuple"],
             metadata["record_is_scalar"],
             is_nonnullable_nulltype=metadata.get("is_nonnullable_nulltype", False),
-            option_type=metadata["option_type"],
+            option_type=metadata.get("option_type", False),
         )
 
     @property
--- a/src/awkward/_connect/pyarrow/extn_types.py
+++ b/src/awkward/_connect/pyarrow/extn_types.py
@@ -143,3 +143,15 @@ def to_awkwardarrow_storage_types(arrowtype):
         return arrowtype, arrowtype.storage_type
     else:
         return None, arrowtype
+
+
+def _get_meta_str(meta: dict[bytes, bytes] | None, key: bytes) -> str | None:
+    if not meta:
+        return None
+    key_str = key.decode("utf-8")
+    value = meta.get(key) or meta.get(key_str.encode("utf-8")) or meta.get(key_str)
+    return value.decode("utf-8") if isinstance(value, bytes) else value
+
+
+def get_field_option(field: pyarrow.Field, key: bytes) -> str | None:
+    return _get_meta_str(field.metadata, key)
--- a/src/awkward/_connect/pyarrow/extn_types.py
+++ b/src/awkward/_connect/pyarrow/extn_types.py
@@ -41,6 +41,7 @@ class AwkwardArrowType(pyarrow.ExtensionType):
         record_is_tuple,
         record_is_scalar,
         is_nonnullable_nulltype=False,
+        option_type=False,
     ):
         self._mask_type = mask_type
         self._node_type = node_type
@@ -49,6 +50,7 @@ class AwkwardArrowType(pyarrow.ExtensionType):
         self._record_is_tuple = record_is_tuple
         self._record_is_scalar = record_is_scalar
         self._is_nonnullable_nulltype = is_nonnullable_nulltype
+        self._option_type = option_type
         super().__init__(storage_type, "awkward")
 
     def __str__(self):
@@ -96,6 +98,7 @@ class AwkwardArrowType(pyarrow.ExtensionType):
             "record_is_tuple": self._record_is_tuple,
             "record_is_scalar": self._record_is_scalar,
             "is_nonnullable_nulltype": self._is_nonnullable_nulltype,
+            "option_type": self._option_type,
         }
 
     @classmethod
@@ -115,6 +118,7 @@ class AwkwardArrowType(pyarrow.ExtensionType):
             metadata["record_is_tuple"],
             metadata["record_is_scalar"],
             is_nonnullable_nulltype=metadata.get("is_nonnullable_nulltype", False),
+            option_type=metadata.get("option_type", False),
         )
 
     @property
--- a/src/awkward/_connect/pyarrow/extn_types.py
+++ b/src/awkward/_connect/pyarrow/extn_types.py
@@ -118,7 +118,7 @@ class AwkwardArrowType(pyarrow.ExtensionType):
             metadata["record_is_tuple"],
             metadata["record_is_scalar"],
             is_nonnullable_nulltype=metadata.get("is_nonnullable_nulltype", False),
-            option_type=metadata.get("option_type", False),
+            option_type=metadata["option_type"],
         )
 
     @property
@@ -145,13 +145,23 @@ def to_awkwardarrow_storage_types(arrowtype):
         return None, arrowtype
 
 
-def _get_meta_str(meta: dict[bytes, bytes] | None, key: bytes) -> str | None:
+def get_field_option(field: pyarrow.Field, key: bytes) -> bool | None:
+    """
+    Retrieves the metadata value for a given key from a field's metadata,
+    interpreting the value as a boolean if it's b'True' or b'False'.
+    Returns None if the key is missing.
+    """
+    meta = field.metadata
     if not meta:
         return None
-    key_str = key.decode("utf-8")
-    value = meta.get(key) or meta.get(key_str.encode("utf-8")) or meta.get(key_str)
-    return value.decode("utf-8") if isinstance(value, bytes) else value
 
+    value = meta.get(key)
+    if value is None:
+        return None
 
-def get_field_option(field: pyarrow.Field, key: bytes) -> str | None:
-    return _get_meta_str(field.metadata, key)
+    if value == b"False":
+        return False
+    elif value == b"True":
+        return True
+    else:
+        return None
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -1116,9 +1116,9 @@ class RecordArray(RecordMeta[Content], Content):
                     )
                     or x._arrow_needs_option_type(),
                     metadata={
-                        b"option_type": str(x._arrow_needs_option_type()).encode(
-                            "utf-8"
-                        ),
+                        b"option_type": b"True"
+                        if x._arrow_needs_option_type()
+                        else b"False",
                     },
                 )
                 for i, x in enumerate(self._contents)
--- a/src/awkward/contents/recordarray.py
+++ b/src/awkward/contents/recordarray.py
@@ -1125,14 +1125,14 @@ class RecordArray(RecordMeta[Content], Content):
             ]
         )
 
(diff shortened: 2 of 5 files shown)


ianna:
@martindurant - I'm trying to pass the newly added field metadata through a parquet file. Somehow the field metadata is not saved in the parquet file, but **only** if the `RecordArray` is a tuple. I wonder if you could have a look or point me in the right direction? Thanks!

There are two tests that fail:

```
python -m pytest -ss -vv tests/test_1440_start_v2_to_parquet.py::test_recordarray_3
python -m pytest -ss -vv tests/test_1440_start_v2_to_parquet.py::test_recordarray_4

```
```python
writing to_parquet
Field: 0, Metadata: {'option_type': 'True'}
Field: 1, Metadata: {'option_type': 'False'}
now reading from_parquet
line #948 HERE
line #953 Field: , Metadata: option_type None
line #647 Field: 0, Metadata: option_type None
{
    "class": "BitMaskedArray",
    "mask": "u8",
    "valid_when": true,
    "lsb_order": true,
    "content": "float64"
}
line #647 Field: 1, Metadata: option_type None
{
    "class": "BitMaskedArray",
    "mask": "u8",
    "valid_when": true,
    "lsb_order": true,
    "content": {
        "class": "ListOffsetArray",
        "offsets": "i64",
        "content": "float64"
    }
}
line #381 Field: 0, Metadata:  option_type None
<BitMaskedArray valid_when='true' lsb_order='true' len='5'>
    <mask><Index dtype='uint8' len='1'>[11]</Index></mask>
    <content><NumpyArray dtype='float64' len='5'>
        <parameter name='which'>'inner1'</parameter>
        [3.3 1.1 1.1 1.1 0. ]
    </NumpyArray></content>
</BitMaskedArray>
line #381 Field: 1, Metadata:  option_type None
<BitMaskedArray valid_when='true' lsb_order='true' len='5'>
    <mask><Index dtype='uint8' len='1'>[27]</Index></mask>
    <content><ListOffsetArray len='5'>
        <offsets><Index dtype='int64' len='6'>[0 2 5 5 8 8]</Index></offsets>
        <content><NumpyArray dtype='float64' len='8'>
            <parameter name='which'>'inner2'</parameter>
            [4.4 5.5 1.1 2.2 3.3 1.1 2.2 3.3]
        </NumpyArray></content>
    </ListOffsetArray></content>
</BitMaskedArray>
```

martindurant:
> Somehow the field metadata is not saved in the parquet file, but only if the RecordArray is a tuple

I'm not sure what a tuple becomes in arrow/parquet, as there is no tuple type. That is probably the root of what you are seeing. 

Also, it's worth noting that arrow only supports metadata on top-level "fields" of a table schema, not on any deeper nested fields. I don't know how awkward's more general metadata is translated, it would be possible to have the whole complex nested metadata on the top-level field or on the schema as a whole.

ianna:
> > Somehow the field metadata is not saved in the parquet file, but only if the RecordArray is a tuple
> 
> I'm not sure what a tuple becomes in arrow/parquet, as there is no tuple type. That is probably the root of what you are seeing.
> 
> Also, it's worth noting that arrow only supports metadata on top-level "fields" of a table schema, not on any deeper nested fields. I don't know how awkward's more general metadata is translated, it would be possible to have the whole complex nested metadata on the top-level field or on the schema as a whole.

Thanks! Yes, I think, I've fixed the function where we do not pass the field metadata. There is one more step to make sure that the stored metadata is the same as the native - there are quite a few places where the fields are passed through   :-)

ianna (commented):
@martindurant - I hope I managed to pass a field option type everywhere. Please, have a look when you have time. Thanks!

src/awkward/_connect/pyarrow/extn_types.py
```diff
@@ -139,3 +143,15 @@ def to_awkwardarrow_storage_types(arrowtype):
...
+def _get_meta_str(meta: dict[bytes, bytes] | None, key: bytes) -> str | None:
+    if not meta:
+        return None
+    key_str = key.decode("utf-8")
+    value = meta.get(key) or meta.get(key_str.encode("utf-8")) or meta.get(key_str)
+    return value.decode("utf-8") if isinstance(value, bytes) else value
+
+
+def get_field_option(field: pyarrow.Field, key: bytes) -> str | None:
+    return _get_meta_str(field.metadata, key)
```
martindurant: This is only used specifically for `get_field_option(field, b"option_type") == "False"` only. Is it not unnecessarily complex and perhaps slow for this task? Do you think it might be possible to make a shortcut?

How does this interact with _get_option_type_from_metadata , which looks very similar, even if it is used in the other direction?

martindurant (commented):
Do I understand that it's a requirement of arrow that we must use `b"False"` rather than `False`, or parquet?

src/awkward/_connect/pyarrow/extn_types.py
ianna: Thanks for checking! Very good point! I will clean it up.

As far as I understand, Arrow metadata is strictly key-value string-like pairs (technically: `dict[bytes, bytes]`). Neither Arrow nor Parquet metadata natively understands Python data types like `bool`, `int`, `list`, etc — we need to serialize them ourselves as strings or JSON blobs before storing them in the metadata.

Here are some benchmarks: 
```python
import timeit
import json
import pyarrow as pa

# Simulated Arrow metadata dict for one field
sample_metadata = {
    b"option_type": json.dumps(False).encode("utf-8"),
    b"mask_type": b"ByteMaskedArray",
    b"node_type": b"NumpyArray"
}

# Number of repetitions for timing
N_REPS = 1_000_000

#  Raw bytes lookup
def lookup_raw():
    _ = sample_metadata.get(b"option_type")

# Bytes lookup + decode
def lookup_decode():
    value = sample_metadata.get(b"option_type")
    if value:
        _ = value.decode("utf-8")

# Bytes lookup + decode + json.loads
def lookup_json():
    value = sample_metadata.get(b"option_type")
    if value:
        _ = json.loads(value.decode("utf-8"))

# Using ArrowMetadataManager
class ArrowMetadataManager:
    def __init__(self, metadata):
        self.metadata = metadata

    def get(self, key: str):
        raw_value = self.metadata.get(key.encode("utf-8"))
        if raw_value is None:
            return None
        return json.loads(raw_value.decode("utf-8"))

def lookup_manager():
    mgr = ArrowMetadataManager(sample_metadata)
    _ = mgr.get("option_type")

# Run benchmarks
print("Raw bytes lookup:", timeit.timeit(lookup_raw, number=N_REPS))
print("Bytes + decode:", timeit.timeit(lookup_decode, number=N_REPS))
print("Bytes + decode + JSON loads:", timeit.timeit(lookup_json, number=N_REPS))
print("With ArrowMetadataManager:", timeit.timeit(lookup_manager, number=N_REPS))
```

```
Raw bytes lookup: 0.02636424987576902
Bytes + decode: 0.058057166868820786
Bytes + decode + JSON loads: 0.4133492079563439
With ArrowMetadataManager: 0.4906008751131594
```

ianna: done!

src/awkward/_connect/pyarrow/extn_types.py
```diff
@@ -145,13 +145,23 @@ def to_awkwardarrow_storage_types(arrowtype):
...
+        return None
 
-def get_field_option(field: pyarrow.Field, key: bytes) -> str | None:
-    return _get_meta_str(field.metadata, key)
+    if value == b"False":
+        return False
+    elif value == b"True":
+        return True
+    else:
+        return None
```
martindurant: `return meta.get(key, b"") ==  b"False"`  and change variable name to `not_optional_type` ? 
This, because the only test we make of it is `is False`, and now we ensure a default value of is-not-optional. It covers the case of no key.

src/awkward/contents/recordarray.py
```diff
@@ -1111,7 +1111,15 @@ def _to_arrow(
...
-                    x._arrow_needs_option_type(),
+                    nullable=(
+                        mask_node is not None and mask_node._arrow_needs_option_type()
+                    )
+                    or x._arrow_needs_option_type(),
+                    metadata={
+                        b"option_type": b"True"
+                        if x._arrow_needs_option_type()
+                        else b"False",
+                    },
```
tcawlfield, line 1122: You might consider using lowercase "true" or "false" here. Probably works fine either way, since this is self-contained. The metadata keys and values do have to be bytes IIRC. At the Table schema level, Awkward packs a JSON string into the metadata, but it looks like this is not directly related to that. Still, if there's any reason for consistency here and "consistency" means that (maybe?) metadata values are JSON, then "True" is spelled "true" in JSON.

martindurant: > metadata values are JSON

They get written into a JSON key of the parquet file, but the values themselves are just bytestrings, no JSON implied.

tcawlfield:
> Do I understand that it's a requirement of arrow that we must use `b"False"` rather than `False`, or parquet?

I believe it's a pyarrow requirement that the metadata keys and values must have a type of `bytes`. Exactly what the contents are, we're free to choose. I'm not 100% sure about any of this because it's been a little while, but I'm guessing it's relying on [this Arrow feature](https://arrow.apache.org/docs/format/Columnar.html#custom-application-metadata).

src/awkward/contents/recordarray.py
tcawlfield: Okay, then this is fine with me. I admit that looking at `b"True"` in isolation makes me scratch my head a lot.

tcawlfield (approved):
Looks good to me!

src/awkward/_connect/pyarrow/extn_types.py
ianna: The metadata can be `None` - it doesn't mean that it's not an optional type.

PR #3479, opened 2025-04-24
fix: check for `error` being `None` in the output of validity kernels

ikrommyd:


ianna (approved):
@ikrommyd - looks good to me! Thanks.

PR #3482, opened 2025-04-25
refactor: move `form_with_unique_keys` here

pfackeldey:
Moves https://github.com/dask-contrib/dask-awkward/blob/main/src/dask_awkward/lib/utils.py#L140 into awkward so we can use it also in uproot without introducing a dask-awkward dependency.

In addition, this is a slightly modified implementation that also resolves the problem of having dots as field separators, because fields (ie ROOT TBranch names) can contain dots as well, see: https://github.com/dask-contrib/dask-awkward/issues/552

### DIFF
--- a/src/awkward/forms/form.py
+++ b/src/awkward/forms/form.py
@@ -744,3 +744,35 @@ class Form(Meta):
             and not (form_key and self._form_key != other._form_key)
             and compare_parameters(self._parameters, other._parameters)
         )
+
+
+def form_with_unique_keys(form: Form, key: tuple) -> Form:
+    def impl(form: Form, key: tuple) -> None:
+        # Set form key as str
+        # Can be interpreted as tuple again with `ast.literal_eval(form_key)`
+        form.form_key = repr(key)
+
+        # If the form is a record we need to loop over all fields in the
+        # record and set form that include the field name; this will keep
+        # recursing as well.
+        if form.is_record:
+            for field in form.fields:
+                impl(form.content(field), (*key, field))  # type: ignore[attr-defined]
+
+        elif form.is_union:
+            for i, entry in enumerate(form.contents):  # type: ignore[attr-defined]
+                impl(entry, (*key, i))
+
+        # NumPy like array is easy
+        elif form.is_numpy or form.is_unknown:
+            pass
+
+        # Anything else grab the content and keep recursing
+        else:
+            # use `None` for anything else, could be more informative at some point, but there's no need for it right now
+            impl(form.content, (*key, None))  # type: ignore[attr-defined]
+
+    # Perform a "deep" copy without preserving references
+    form = ak.forms.from_dict(form.to_dict())
+    impl(form, key)
+    return form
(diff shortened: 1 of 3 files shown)


src/awkward/forms/form.py
```diff
@@ -744,3 +744,35 @@ def _is_equal_to_generic(
...
+
+        # Anything else grab the content and keep recursing
+        else:
+            # use `None` for anything else, could be more informative at some point, but there's no need for it right now
+            impl(form.content, (*key, None))  # type: ignore[attr-defined]
+
+    # Perform a "deep" copy without preserving references
+    form = ak.forms.from_dict(form.to_dict())
+    impl(form, key)
+    return form
```
ianna, line 778:
```suggestion
from functools import singledispatch

@singledispatch
def form_with_unique_keys(form: Form, key: tuple) -> Form:
    """Default case for handling forms that don’t fit specific types."""
    raise TypeError(f"Unsupported form type: {type(form)}")

@form_with_unique_keys.register
def _(form: ak.contents.RecordArray, key: tuple) -> Form:
    # Set form key as str
    form.form_key = repr(key)
    for field in form.fields:
        form_with_unique_keys(form.content(field), (*key, field))  # type: ignore[attr-defined]
    return form

@form_with_unique_keys.register
def _(form: ak.contents.UnionArray, key: tuple) -> Form:
    # Set form key as str
    form.form_key = repr(key)
    for i, entry in enumerate(form.contents):  # type: ignore[attr-defined]
        form_with_unique_keys(entry, (*key, i))
    return form

@form_with_unique_keys.register
def _(form: (ak.contents.NumpyArray, ak.contents.UnknownArray), key: tuple) -> Form:
    # Handle NumPy-like arrays
    form.form_key = repr(key)
    return form

@form_with_unique_keys.register
def _(form: ak.contents.ByteMaskedArray, key: tuple) -> Form:
    # Default case for other content types
    form.form_key = repr(key)
    form_with_unique_keys(form.content, (*key, None))  # type: ignore[attr-defined]
    return form

def apply_unique_keys(form: Form, key: tuple) -> Form:
    # Perform a "deep" copy without preserving references
    form = ak.forms.from_dict(form.to_dict())
    return form_with_unique_keys(form, key)
```

ianna: since we are refactoring, how about using `singledispatch`?

ianna (changes requested):
@pfackeldey - Great! Thanks! Could you, please add some tests for this code? I think, the PR title should be `refactor:`, not `chore:`.

Thanks.

pfackeldey:
>  Could you, please add some tests for this code? 

Sure, I'll add some tests :+1:

src/awkward/forms/form.py
pfackeldey: I'm personally not a huge fan of this coding style. I know it's very Julia inspired, but I find it confusing as a non-Julia person, and I suspect many other Python(-main) devs find that unintuitive as well. I'd argue that closures and recursion are much more known patterns.

I'm also not sure if this works well here because in principle this introduces a circular dependency: the type annotation has to be e.g. `ak.forms.record_form.RecordForm` which in turn needs `ak.forms.Form` to inherit from. Since, `singledispatch` needs to know the type at runtime I doubt we can break the circular dependency by using `if TYPE_CHECKING:`.

(Also: this bloats the code quite a bit, because the decorator and `form.form_key = repr(key)` has to be added for each case: 29 lines -> 39 lines)

ianna (approved):
@pfackeldey - looks good to me! Thanks.

PR #3487, opened 2025-04-30
chore: update pyarrow version in the full testing requirements

ianna:
needs  #3478


requirements-test-full.txt
```diff
@@ -3,7 +3,7 @@ jax[cpu]>=0.2.15;sys_platform != "win32" and python_version < "3.13"
 numba>=0.50.0;sys_platform != "win32" and python_version < "3.13"
 numexpr>=2.7; python_version < "3.13"
 pandas>=0.24.0;sys_platform != "win32" and python_version < "3.13"
-pyarrow==16.0.0;sys_platform != "win32" and python_version < "3.13"
+pyarrow>=12.0.0;sys_platform != "win32" and python_version < "3.13"
```
ariostas: Maybe it would also be good to remove the `python_version < "3.13"` cap (here and above) to see if it works

ianna: Good idea! I’ll try it when the #3478 is merged.

PR #3489, opened 2025-04-30
fix: consistent materializing length across all layouts

ikrommyd:
This PR makes the use of `._length` (non materializing) versus `.length` (materializing) consistent across all layouts.
Length is non materializing in `__init__`,  when copying, when converting to typetracer, when switching backends and when getting a field via `._getitem_field(s)`. The last one will materialize the length anyways in the record array case because it's using `self.content` which materializes the length because it needs to slice the content. It's materializing everywhere else.

### DIFF
--- a/src/awkward/contents/unmaskedarray.py
+++ b/src/awkward/contents/unmaskedarray.py
@@ -418,9 +418,7 @@ class UnmaskedArray(UnmaskedMeta[Content], Content):
 
         if isinstance(out, ak.contents.RegularArray):
             tmp = ak.contents.UnmaskedArray.simplified(out._content, parameters=None)
-            return ak.contents.RegularArray(
-                tmp, out._size, out._length, parameters=None
-            )
+            return ak.contents.RegularArray(tmp, out.size, out.length, parameters=None)
 
         else:
             return out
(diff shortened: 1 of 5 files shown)


src/awkward/contents/unmaskedarray.py
```diff
@@ -418,9 +418,7 @@ def _argsort_next(
 
         if isinstance(out, ak.contents.RegularArray):
             tmp = ak.contents.UnmaskedArray.simplified(out._content, parameters=None)
-            return ak.contents.RegularArray(
-                tmp, out._size, out._length, parameters=None
-            )
+            return ak.contents.RegularArray(tmp, out.size, out.length, parameters=None)
```
ianna, line 421: @ikrommyd - why `.size` and not `._size`? It seems inconsistent with what is done in `regulararray.py`

ianna (commented):
@ikrommyd - Looks good! Thanks. Please, check my comment. Thanks!

src/awkward/contents/unmaskedarray.py
ikrommyd: Cause regular array uses `._size` because it is its own private attribute. I believe other layouts should use the public attributes of other layouts though in general. I think we typically do that throughout the code although there may be typos.

ikrommyd: Same reason I changed it in `content.py` as well. If you search, you'll see that other layouts use the public one apart from these 2 cases.
image

ianna: good catch! Thank you!

ianna (approved):
@ikrommyd - Great job! Thanks! Please, merge it if you finished with it. Thanks!

PR #3490, opened 2025-04-30
fix: avoid calling `shape_generator` inside `generator` if it has already been called

ikrommyd:
This is done via introducing an LRU cache of size 1 to hold the previous call.

### DIFF
--- a/src/awkward/_nplikes/virtual.py
+++ b/src/awkward/_nplikes/virtual.py
@@ -127,6 +127,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
                     f"Only shapes of integer dimensions are supported for materialized shapes. Received shape {shape}"
                 )
             self._shape = tuple(map(int, shape))
+            self._shape_generator = None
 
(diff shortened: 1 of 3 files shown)


ikrommyd:
@pfackeldey what do you think of this? I also added an explicit `assert callable(shape_generator)` which the code seems to assume.

pfackeldey (approved):
Thanks! this looks good to me :+1:

Here we have a safe cache (size=1) that only makes sure we do not evaluate the shape twice. It's small in memory because it only stores a single tuple of ints (just a few bytes). And as soon as the array is materialized the reference to the shape generator and thus to the cache is dropped, so there's dangling memory either.

ianna (changes requested):
@ikrommyd - Looks good, thanks! Could you, please, add a test to cover `assert_never`? Thanks!

ikrommyd:
> @ikrommyd - Looks good, thanks! Could you, please, add a test to cover `assert_never`? Thanks!

Done

ianna (approved):
Excellent! Thanks!

PR #3497, opened 2025-05-05
feat: add a ref to the original 'unclipped' buffer generator

pfackeldey:
This allows us to access the original buffer generator at some later stage.


pfackeldey:
Happy if you have a better name of this `attr`

ianna (approved):
@pfackeldey - Looks good to me! Thanks!

PR #3498, opened 2025-05-05
feat: improve assertion error message in kernels

pfackeldey:
This turns:
```
AssertionError: Only NumPy buffers should be passed to Numpy Kernels, received PyCPointerType
```
into
```
AssertionError: Only NumPy buffers should be passed to Numpy Kernels, received PlaceholderArray(dtype('uint8'), shape=(1,)) (ptr type=PyCPointerType)
```
where you can see what kind of object (PlaceHolderArray) was the cause of the error.

### DIFF
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -81,7 +81,7 @@ class NumpyKernel(BaseKernel):
                 return ctypes.cast(x, t)
             else:
                 raise AssertionError(
-                    f"Only NumPy buffers should be passed to Numpy Kernels, received {type(t).__name__}"
+                    f"Only NumPy buffers should be passed to Numpy Kernels, received {x} (ptr type={type(t).__name__})"
                 )
         else:
             return x
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -162,11 +162,7 @@ class CupyKernel(BaseKernel):
             # Do we have a CuPy-owned array?
             if self._cupy.is_own_array(x):
                 assert self._cupy.is_c_contiguous(x)
-                return x
-            else:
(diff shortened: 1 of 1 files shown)


src/awkward/_kernels.py
```diff
@@ -162,7 +162,11 @@ def _cast(self, x, type_):
             # Do we have a CuPy-owned array?
             if self._cupy.is_own_array(x):
                 assert self._cupy.is_c_contiguous(x)
-            return x
```
ikrommyd: @ianna do you know why this was outside the if statement for this kernel? No other kernel does this. In which would we want to `return x` instead of erroring if it's not a cupy owned array.

ianna: when it is a numpy array :-)

ikrommyd: Why would a numpy array go through a `CupyKernel` though? There are no mixed type arrays anymore right?

ikrommyd: FYI, I got no cuda test failures with this change.

ianna: @ikrommyd - When you have a moment, could you please revert to the previous logic? For now, we’d prefer to avoid adjustments that could affect performance.

ianna (changes requested):

src/awkward/_kernels.py
ikrommyd: Sure, just did it. It's just that like this, the cupy kernels cannot error like the others do. Anyway...

ianna: Thanks!

ianna (approved):
Thanks! Looks good to me!

PR #3499, opened 2025-05-05
fix: more meaningful assertion error in length calculation in case of column optimization problems in dask-awkward

ikrommyd:


ianna (changes requested):
@ikrommyd - great! Thanks! Since you are adding more information to the assertion message how about giving the information about the value and its type as well? Thanks

ikrommyd:
> @ikrommyd - great! Thanks! Since you are adding more information to the assertion message how about giving the information about the value and its type as well? Thanks


I mean....it can only be unknown length in reality so I'll just return the type. Writing out unknown length twice for the value and type is a bit annoying.

ianna (approved):
Thanks! Looks good.

PR #3501, opened 2025-05-07
feat: do not allow records, primitives, or bare strings in `ak.materialize` (+ docs display fix)

ikrommyd:
Records give dangerous materializations where you materialize a whole array to figure out one row only and I don't think single numbers or strings should go through this function in the same way `ak.Array` doesn't allow them.

Also noticed that `ak.materialize` doesn't display on the docs website due to a single space typo so I fixed that as well.

### DIFF
--- a/src/awkward/contents/bytemaskedarray.py
+++ b/src/awkward/contents/bytemaskedarray.py
@@ -1195,7 +1195,7 @@ class ByteMaskedArray(ByteMaskedMeta[Content], Content):
         )
 
     def _materialize(self) -> Self:
-        content = self._content._materialize()
+        content = self._content.materialize()
         mask = self._mask.materialize()
         return ByteMaskedArray(
             mask, content, valid_when=self._valid_when, parameters=self._parameters
--- a/src/awkward/operations/ak_materialize.py
+++ b/src/awkward/operations/ak_materialize.py
@@ -17,7 +17,7 @@ def materialize(
 ):
     """
     Args:
-        array : Array-like data (anything #ak.to_layout recognizes).
+        array : Array-like data (either an #ak.Array or an #ak.contents.Content).
             An array that may contain virtual buffers to be materialized.
         highlevel (bool, default is True): If True, return an #ak.Array;
             otherwise, return a low-level #ak.contents.Content subclass.
@@ -28,7 +28,7 @@ def materialize(
 
     Traverses the input array and materializes any virtual buffers.
     If the input array is not an #ak.Array or an #ak.contents.Content,
-    it will just be cast to an #ak.Array.
+    an error will be raised.
     The buffers of the returned array are no longer `VirtualArray` objects even if there were any.
     They will become one of `numpy.ndarray`, `cupy.ndarray`, or `jax.numpy.ndarray` objects,
     depending on the array's backend.
--- a/src/awkward/operations/ak_materialize.py
+++ b/src/awkward/operations/ak_materialize.py
@@ -1,7 +1,6 @@
 # BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
 from __future__ import annotations
 
-import awkward as ak
 from awkward._dispatch import high_level_function
 from awkward._layout import HighLevelContext
 
@@ -17,8 +16,8 @@ def materialize(
 ):
     """
     Args:
-        array : array_like
-            An array with possible virtual buffers materialize.
+        array : Array-like data (anything #ak.to_layout recognizes).
+            An array that may contain virtual buffers to be materialized.
         highlevel (bool, default is True): If True, return an #ak.Array;
             otherwise, return a low-level #ak.contents.Content subclass.
         behavior (None or dict): Custom #ak.behavior for the output array, if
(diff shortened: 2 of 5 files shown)


src/awkward/contents/bytemaskedarray.py
```diff
@@ -1195,7 +1195,7 @@ def _to_backend(self, backend: Backend) -> Self:
         )
 
     def _materialize(self) -> Self:
-        content = self._content._materialize()
+        content = self._content.materialize()
```
ianna, line 1198: @ikrommyd - how does it relate to records, primitives, or bare strings? Thanks.

src/awkward/operations/ak_materialize.py
```diff
@@ -40,18 +40,9 @@ def materialize(
...
-    if not isinstance(
-        array,
-        (
-            ak.highlevel.Array,
-            ak.highlevel.Record,
-            ak.contents.Content,
-            ak.record.Record,
-        ),
-    ):
-        return array
```
ianna, line 52: what if you get an `ak.Index` here?

ianna (changes requested):
@ikrommyd - Thanks! I wonder if any instance should be allowed in `_impl`

src/awkward/contents/bytemaskedarray.py
ikrommyd: It doesn't, it was just a "typo" for consistency. This line was accidentally using the private method. It's harmless so I thought I'd just sneak it it in this PR.

src/awkward/operations/ak_materialize.py
ikrommyd: How would you get that? Only high-level objects are gonna get passed in. You don't call `ak.sum(some ak.index.Index)` instance.

ikrommyd:
> @ikrommyd - Thanks! I wonder if any instance should be allowed in `_impl`

I let the automatic casting of `ak.to_layout` handle that. I basically wanted the behavior to be similar to `ak.Array` when the input has nothing to do with virtual arrays. `ak.materialize([1,2,3])` will give you the same thing `ak.Array([1,2,3])` will.

src/awkward/operations/ak_materialize.py
ikrommyd: Actually` ak.sum` is going to be able to sum an index and `ak.to_layout` will transform it into a `NumpyArray` (which will be called before the actual summation)
```py
In [1]: import awkward as ak

In [2]: x = ak.index.Index([1,2,3])

In [3]: ak.sum(x)
Out[3]: np.int64(6)

In [4]: ak.to_layout(x)
Out[4]: <NumpyArray dtype='int64' len='3'>[1 2 3]</NumpyArray>
```
The only reason for that is because the index implements the dlpack protocol: https://github.com/scikit-hep/awkward/blob/d688c0de84e37661ccf2952813f15c1a7b7fb403/src/awkward/index.py#L199-L206 which `to_layout` is gonna try to use: https://github.com/scikit-hep/awkward/blob/d688c0de84e37661ccf2952813f15c1a7b7fb403/src/awkward/operations/ak_to_layout.py#L231-L234.

I don't think there is any reason to handle `Index` explicitly in `materialize` in the same way no other high level functions do.

ianna:
> > @ikrommyd - Thanks! I wonder if any instance should be allowed in `_impl`
> 
> I let the automatic casting of `ak.to_layout` handle that. I basically wanted the behavior to be similar to `ak.Array` when the input has nothing to do with virtual arrays. `ak.materialize([1,2,3])` will give you the same thing `ak.Array([1,2,3])` will.

I would expect the output of `ak.materialize([1,2,3])` be `[1,2,3]`, not a high level awkward array.

ikrommyd:
Uhmm....I don't know. I don't like non awkward stuff coming out. I wanted to make `ak.materialize` to behave like `ak.Array` in such cases. This was actually the behavior in awkward1's `ak.materialized` that's why. It doesn't really matter honestly. The only dangerous thing is allowing records cause would materialize the whole array to get one row and that's bad. If you have strong opinions, I guess I could let non awkward stuff just go through. That would mean primitives and strings too.

ianna:
@ikrommyd - This is a highlevel, e.g. public function. A user can pass anything to it. Without this restriction it becomes a way to create Awkward Arrays - we don't want that.

I think, there are two options here:

1. If you care about graceful handling of unexpected types (like `NumPy` arrays, strings, ints) — keep the `isinstance`.
2. If you’re enforcing strict API discipline — replace the pass-through with a `TypeError`.

If it were private, hot-path code, where types are already guaranteed and you need every nanosecond — you could skip it, but the gain is negligible.

Awkward 1 `materialized` function took (possibly virtual) Awkward Arrays:
```python
def materialized(array, highlevel=True, behavior=None):
    """
    Args:
        array: The possibly virtual array to ensure is materialized.
            a low-level #ak.layout.Content subclass.
            high-level.

    Returns ``array`` with all virtual array nodes fully materialized.

    See also #ak.virtual.
    """

    def getfunction(layout):
        if isinstance(layout, ak.layout.VirtualArray):
            result = materialized(layout.array, highlevel=False)
            return lambda: result
        else:
            return None

    layout = ak.operations.convert.to_layout(array)
    out = ak._util.recursively_apply(
        layout, getfunction, pass_depth=False, pass_user=False
    )
    return ak._util.maybe_wrap_like(out, array, behavior, highlevel)
```

ikrommyd:
This is what I was trying to immitate:
```py
In [4]: import awkward as ak

In [5]: ak.__version__
Out[5]: '1.10.5'

In [6]: ak.materialized([1,2,3])
Out[6]: <Array [1, 2, 3] type='3 * int64'>
```
and indeed awkward1 starts by doing `layout = ak.operations.convert.to_layout(array)`.

I'll think about it a bit. If you don't like this behavior, indeed we can either pass through as is with `isinstance` or error.

ikrommyd:
I'm in favor of erroring actually because the user might pass something into it thinking it's an `ak.Array` in code but it's something else. I'd rather error so that the user knows they are passing awkward arrays there.

ikrommyd:
@ianna what do you think of the interface now?

ianna (approved):
@ikrommyd - Great! Thanks for addressing all the comments!

ikrommyd:
@ianna I can merge this yeah? I adjusted the docstring to match the last commit (I had forgotten) and waiting for the ci to pass

PR #3505, opened 2025-05-10
fix: use cmath instead of bit mask for infinite doubles

### LINKED ISSUE #3504: awkward-cpp: error: dereferencing type-punned pointer will break strict-aliasing rules
### Version of Awkward Array

2.8.2

### Version of Awkward-cpp

45

```
src/libawkward/forth/ForthInputBuffer.cpp:196:31: error: dereferencing type-punned pointer will break strict-aliasing rules [-Werror=strict-aliasing]
  196 |   double positive_infinity = *(double*)&bits_infinity;
      |                               ^~~~~~~~~~~~~~~~~~~~~~~
```
The line is https://github.com/scikit-hep/awkward/blob/ae1a51a1bef7580ef515641b1eabfa9f5c15fd95/awkward-cpp/src/libawkward/forth/ForthInputBuffer.cpp#L196

I think the better way to do this is:
```cpp
double positive_infinity = std::numeric_limits<double>::infinity();
double negative_infinity = -std::numeric_limits<double>::infinity();
```

See-also: https://bugs.gentoo.org/955697

APN-Pucky:
Closes: #3504

### DIFF
--- a/awkward-cpp/src/libawkward/forth/ForthInputBuffer.cpp
+++ b/awkward-cpp/src/libawkward/forth/ForthInputBuffer.cpp
@@ -3,6 +3,7 @@
 #define FILENAME(line) FILENAME_FOR_EXCEPTIONS("src/libawkward/forth/ForthInputBuffer.cpp", line)
 
 #include "awkward/forth/ForthInputBuffer.h"
+#include <cmath>
 
 namespace awkward {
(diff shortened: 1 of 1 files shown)


ianna (approved):
@APN-Pucky - Thanks for fixing this! Looks good to me.

PR #3510, opened 2025-05-15
fix: Python 3.13 classifier to awkward-cpp

APN-Pucky:
awkward itself has the 3.13 classifier already so I guess it should also be added to awkward-cpp?


ianna (approved):
@APN-Pucky - Thanks! Just in time for the new release!

PR #3516, opened 2025-05-20
ci: change pre-commit autoupdate to a monthly schedule

ariostas:
Almost every week there is a PR updating some pre-commit hook, and almost always they don't make any difference. I'm changing the autoupdate schedule so that it only runs monthly.

Henry convinced me to leave the dependabot updates on a weekly schedule since they don't trigger as often and are more important.


ianna (approved):
@ariostas - thanks for fixing it!

Issue #3403, opened 2025-02-21
Question on performance


martindurant:
### Version of Awkward Array

2.7.4

### Description and code to reproduce

numpy 1.26.4
pyarrow 19.0.0

The origin of the data I will use here is not really important, but for reference, it is:
[1.9GB of points](
https://github.com/geoarrow/geoarrow-data/releases/download/v0.1.0/microsoft-buildings-point.arrow) in feather2 format.
```
table = pyarrow.feather.read_table("microsoft-buildings-point.arrow")
```
130M points. The "geometry" column has x, y fields, both float64.

Issue 1
=====
(the lesser issue)

Depending on how I convert the data, I get different layouts:
```python
>>> ak.from_arrow(table)["geometry", "x"].layout
<IndexedOptionArray len='129735970'>
    <index><Index dtype='int64' len='129735970'>
        [        0         1         2 ... 129735967 129735968 129735969]
    </Index></index>
    <content><NumpyArray dtype='float64' len='129735970'>
        [ -84.95972352  -84.95973298  -84.9599375  ... -111.04598275
         -111.047405   -111.0478207 ]
    </NumpyArray></content>
</IndexedOptionArray>

>>> ak.from_arrow(table["geometry"])["x"].layout
<UnmaskedArray len='129735970'>
    <content><NumpyArray dtype='float64' len='129735970'>
        [ -84.95972352  -84.95973298  -84.9599375  ... -111.04598275
         -111.047405   -111.0478207 ]
    </NumpyArray></content>
</UnmaskedArray>
```
Here, the second variant is what you should get - we know there are no NULLs. If you don't select "x", you see UnmaskedArray s even for the first route. 

Issue 2
======
Doing some timings:
```python
>>> x = ak.from_arrow(table["geometry"])["x"] # the unmasked variant
>>> np.max(x)
656ms
>>> ak.max(x)
666ms, OK, so dispatch does what we expect
>>> %timeit np.max(x.layout.content.data)
18ms, well that is just a bit faster
>>> %timeit np.nanmax(x.layout.content.data)
20ms, in case of nan (since we shold have no NULLs)
>>>  np.nanmax(np.where(True, x.layout.content.data, np.nan))
176ms, maybe this is what awkward actually does?
```

And with a handwritten simple numba kernel:
```python
@numba.njit(nogil=True, cache=True)
def mymax(x):
    max = -np.inf
    for v in x:
        if np.isfinite(v) and v > max:
            max = v
    return v
```

[...]

PR #3518, opened 2025-05-22
fix: simplify unmasked array

ianna:
Fixes Issue 1 in #3403

An `Unmasked` array has no missing values by definition. It is still an **option type**, though.
That’s why, when its content was an `Unmasked` array, the simplified method was introducing an `IndexedOptionArray` node.

@martindurant - FYI

### DIFF
--- a/src/awkward/_meta/meta.py
+++ b/src/awkward/_meta/meta.py
@@ -21,6 +21,7 @@ class Meta:
     is_record: ClassVar[bool] = False
     is_union: ClassVar[bool] = False
     is_leaf: ClassVar[bool] = False
+    is_unmasked: ClassVar[bool] = False
 
     _parameters: JSONMapping | None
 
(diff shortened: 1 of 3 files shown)


martindurant:
Thanks for coming back to this!

The case should be testable, right?

ianna (commented):
@agoose77 - I am not sure if we are using the is_option type for unmasked arrays. That is why I decided to add an extra flag. Another question is if we should drop all nested Unmasked arrays recursively.

agoose77:
@ianna `UnmaskedArray` behaves as an option, and so we will be taking slow paths if we're only testing `is_option`. My feeling with Awkward is that we're dealing with a closed set of node types (or at least, closed set of families), so adding a new flag that's helpful to identify a member of that set is fine!

I don't entirely follow your question about dropping recursively; we can repeatedly simplify `UnmaskedArray(UnmaskedArray(...))`, but we can't simplify through type boundaries like lists.

PR #3531, opened 2025-06-07
fix: allow numpy integers and 0-dim arrays through `shape_item_as_index` to support numpy 2.3

### LINKED ISSUE #3530: Support NumPy 2.3
### Version of Awkward Array

latest in pypi

Found in: https://github.com/scikit-hep/fastjet/actions/runs/15510572224/job/43671004688

```
    def shape_item_as_index(self, x1: ShapeItem) -> int:
        if x1 is unknown_length:
            raise TypeError("array module nplikes do not support unknown lengths")
        elif isinstance(x1, int):
            return x1
        else:
>           raise TypeError(f"expected None or int type, received {x1}")
E           TypeError: expected None or int type, received 2

/opt/hostedtoolcache/Python/3.13.3/x64/lib/python3.13/site-packages/awkward/_nplikes/array_module.py:383: TypeError
```

This test functions with numpy < 2.3.

lgray:
fixes #3530 (but we'll see what the tests say)

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -379,6 +379,8 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             raise TypeError("array module nplikes do not support unknown lengths")
         elif isinstance(x1, int):
             return x1
+        elif isinstance(x1, (np.int64, np.int32, np.uint64, np.uint32)):
+            return int(x1)
         else:
             raise TypeError(f"expected None or int type, received {x1}")
 
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -17,6 +17,7 @@ from awkward._nplikes.numpy_like import (
 from awkward._nplikes.placeholder import PlaceholderArray
 from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._nplikes.virtual import VirtualArray, materialize_if_virtual
+from awkward._regularize import is_integer_like
 from awkward._typing import TYPE_CHECKING, Any, DType, Final, Literal, TypeVar, cast
 
 if TYPE_CHECKING:
@@ -377,9 +378,7 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
     def shape_item_as_index(self, x1: ShapeItem) -> int:
         if x1 is unknown_length:
             raise TypeError("array module nplikes do not support unknown lengths")
-        elif isinstance(x1, int):
-            return x1
-        elif isinstance(x1, (np.int64, np.int32, np.uint64, np.uint32)):
+        elif is_integer_like(x1):
             return int(x1)
         else:
             raise TypeError(f"expected None or int type, received {x1}")
--- a/requirements-test-full.txt
+++ b/requirements-test-full.txt
@@ -1,7 +1,7 @@
 fsspec>=2022.11.0;sys_platform != "win32"
 jax[cpu]>=0.2.15;sys_platform != "win32" and python_version < "3.13"
 numba>=0.50.0;sys_platform != "win32" and python_version < "3.13"
-numexpr>=2.7; python_version < "3.13"
+numexpr>=2.7,<2.11; python_version < "3.13"
 pandas>=0.24.0;sys_platform != "win32"
 pyarrow>=12.0.0;sys_platform != "win32"
 pytest>=6
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -22,7 +22,7 @@ from awkward._nplikes.numpy_like import (
 from awkward._nplikes.placeholder import PlaceholderArray
 from awkward._nplikes.shape import ShapeItem, unknown_length
 from awkward._operators import NDArrayOperatorsMixin
-from awkward._regularize import is_integer, is_non_string_like_sequence
+from awkward._regularize import is_integer, is_integer_like, is_non_string_like_sequence
 from awkward._typing import (
     TYPE_CHECKING,
     Any,
@@ -1040,8 +1040,8 @@ class TypeTracer(NumpyLike[TypeTracerArray]):
     def shape_item_as_index(self, x1: ShapeItem) -> IndexType:
         if x1 is unknown_length:
             return TypeTracerArray._new(np.dtype(np.int64), shape=())
-        elif isinstance(x1, int):
-            return x1
+        elif is_integer_like(x1):
+            return int(x1)
         else:
             raise TypeError(f"expected unknown_length or int type, received {x1}")
 


lgray:
Though now this technically violates the typing of `ShapeItem`, but the awkward authors should debate that and figure out what they want to do.

src/awkward/_nplikes/array_module.py
```diff
@@ -379,6 +379,8 @@ def shape_item_as_index(self, x1: ShapeItem) -> int:
             raise TypeError("array module nplikes do not support unknown lengths")
         elif isinstance(x1, int):
             return x1
+        elif isinstance(x1, (np.int64, np.int32, np.uint64, np.uint32)):
+            return int(x1)
```
ikrommyd, line 382: Perhaps the most concise thing here is to use `is_integer`: https://github.com/scikit-hep/awkward/blob/8e2d4a85420e89617ccf5be0edef841d3b2d57bf/src/awkward/_regularize.py#L26-L27 to allow both builtin `int` and numpy `integer` as shape item?

ikrommyd: This would replace both if statements with one `elif is_integer(x1)`: ...`

lgray: The specified return type of this function is python int, so it needs to be converted when it's a dtype.

ikrommyd:
I think the typing of `ShapeItem` atm is not perfect because we allow numpy ints in the shape of virtual arrays. The length of a virtual array is often the last element of some offsets so it's `offsets[-1]`. Since numpy 2, `offsets[-1]` is a numpy int and not a python int so there is a case to be made that the shapeitem should be updated to support `np.integer` indeed.

Edit: actually...I think we do `length = int(length)` for this case so that the shape tuple contains only python ints but now that I'm thinking this...That's probably bad for virtual arrays on GPU right? Cupy errors in this case I think and tells you to use `.get()` to get the integer on the cpu.

pfackeldey (changes requested):
I think I'd prefer:

```python
import operator 
from awkward._regularize import is_integer_like

class ArrayModuleNumpyLike:
    ...
    def shape_item_as_index(self, x1: ShapeItem) -> int:
        if x1 is unknown_length:
            raise TypeError("array module nplikes do not support unknown lengths")
        elif is_integer_like(x1):
            return operator.index(x1)
        else:
            raise TypeError(f"expected None or int type, received {x1}")
```

This should be working automatically for all relevant types. 

This change (likely) needs to be propagated to the TypeTracer nplike as well: https://github.com/scikit-hep/awkward/blob/main/src/awkward/_nplikes/typetracer.py#L1040-L1046 to:

```python
import operator 
from awkward._regularize import is_integer_like

class TypeTracer:
    ...
    def shape_item_as_index(self, x1: ShapeItem) -> IndexType:
        if x1 is unknown_length:
            return TypeTracerArray._new(np.dtype(np.int64), shape=())
        elif is_integer_like(x1):
            return operator.index(x1)
        else:
            raise TypeError(f"expected unknown_length or int type, received {x1}")
```

Could you give this a try?

ikrommyd:
Just for context, this happens because in numpy 2.3 `np.count_nonzero` returns a numpy integer unlike numpy 2.2
```
In [2]: np.count_nonzero([1,2,3])
Out[2]: np.int64(3)
```
```
In [2]: np.count_nonzero([1,2,3])
Out[2]: 3
```
If the point of `ShapeItem` is to express "things that can live in the shape property" then maybe it should be only python built in integer and `unknown_length`and then it would be awkward's job to make sure no `np.integer` is passed inside the `shape_item_as_index` function. Numpy and other nplikes still return builtin python integers with their `.shape` property all the time right? I haven't seen an exception to this so far. However, I think this doesn't play nice with the CUDA backend.

Maybe `ShapeItem` typing should stay as is but there should be another function that checks whether the input "is integer like"? Maybe not? Upgrading `ShapeItem` and `shape_item_as_index` to handle `np.integer` is always an option.

 In the case of this error, `shape_item_as_index` is used to make sure the input can go into `nplike.arange` and that it is not `unknown_length` if the backend is not typetracer: https://github.com/scikit-hep/awkward/blob/8e2d4a85420e89617ccf5be0edef841d3b2d57bf/src/awkward/contents/indexedoptionarray.py#L1730-L1734

ikrommyd:
@pfackeldey `operator.index` doesn't work with cupy
```
In [15]: x = cupy.array([1,2,3])

In [16]: operator.index(x[2])
Cell In[16], line 1
----> 1 operator.index(x[2])

TypeError: 'ndarray' object cannot be interpreted as an integer
```
`int` does though but I assume there is some GPU -> RAM casting here
```
In [17]: x[2]
Out[17]: array(3)

In [18]: int(x[2])
Out[18]: 3
```

lgray:
Yeah getting that to not have to spend time copying a single int back to host would be nice. That's relatively extremely expensive.

pfackeldey:
> @pfackeldey `operator.index` doesn't work with cupy
> 
> ```
> In [15]: x = cupy.array([1,2,3])
> 
> In [16]: operator.index(x[2])
> ---------------------------------------------------------------------------
> TypeError                                 Traceback (most recent call last)
> Cell In[16], line 1
> ----> 1 operator.index(x[2])
> 
> TypeError: 'ndarray' object cannot be interpreted as an integer
> ```
> 
> `int` does though but I assume there is some GPU -> RAM casting here
> 
> ```
> In [17]: x[2]
> Out[17]: array(3)
> 
> In [18]: int(x[2])
> Out[18]: 3
> ```

Then, there should be a different dedicated `shape_item_as_index` implementation for the CuPy nplike. The implementations that I proposed for CPU and TypeTracer (should also work for JAX) are the most general and safe ways I can think of to implement this.

ianna (approved):
@lgray - thanks for fixing it. I agree that there might be other issues with migration to Numpy 2.3, however, let's have a reproducer for the upcoming issues first. We can refactor and generalize then.

lgray:
While I'm not sure this fix is appropriate, it does fix this specific problem. 
At least patching it atop my work area for fastjet.

I'll get to making a reproducer today so this can be more correctly managed.

ikrommyd:
@lgray @pfackeldey As discussed offline, I've pushed the fix that covers python ints, numpy ints, dim-0 arrays and should work with cupy as well.

@lgray could you check that this makes the fastjet test pass?
@ianna could you please re-review?

This is a quick fix for numpy 2.3 but we should rethink what a `ShapeItem` is and what `shape_item_as_index` is meant to do because it feels like the names are a bit miss-leading given what they are meant to do in the actual code.

pfackeldey (approved):

ikrommyd:
@ianna I could add a test as well if you'd like? Or we can wait to revisit shape item and test then? Here is the simplest reproducer of the issue with numpy 2.3
```py
import awkward as ak
import numpy as np

content = ak.contents.NumpyArray(np.array([0.0, 1.1, 2.2, 3.3, 4.4]))
index = ak.index.Index64(np.array([2, 2, 0, -1, 4], dtype=np.int64))
array = ak.contents.IndexedOptionArray(index, content)
array.to_packed()
```

ikrommyd:
The failures are one single test coming from the latest numexpr release. It's coming from this line: https://github.com/scikit-hep/awkward/blob/8e2d4a85420e89617ccf5be0edef841d3b2d57bf/src/awkward/_connect/numexpr.py#L82 because `numexpr.necompiler._names_cache` is no longer a `CacheDict` but a `threading.local()` due to this commit: https://github.com/pydata/numexpr/commit/33ee71b0c6f13224f3031cd8b42921c748ce9ede

I'll think tomorrow probably how we can deal with this.

ianna:
> @lgray @pfackeldey As discussed offline, I've pushed the fix that covers python ints, numpy ints, dim-0 arrays and should work with cupy as well.
> 
> @lgray could you check that this makes the fastjet test pass? @ianna could you please re-review?
> 
> This is a quick fix for numpy 2.3 but we should rethink what a `ShapeItem` is and what `shape_item_as_index` is meant to do because it feels like the names are a bit miss-leading given what they are meant to do in the actual code.

Hey — since this one’s already greenlit, let’s keep it as-is and handle any tweaks in a follow-up. Keeps things tidy for the history. Thanks a ton!

ianna (changes requested):
If this is meant to be a real fix, it needs a reproducer, tests, and GPU profiling — else let’s roll with the approved quick fix.

ikrommyd:
@ianna which version are you talking about? The current version of the PR is the appropriate quick fix that works with cuda and the typetracer backend. The original version that you green-lighted won't work with those two backends that's why we had a very short offline chat with Lindsey and Peter and put the current version of the PR as the "quick fix".

The “lengthier” fix would be implementing a separate function for the cupy backend and in general re thinking a bit how shape item works. I’m happy to put a test in this one. Without this fix, there are about 400 test failures though with numpy 2.3 they get noticed in the other NumExpr PR I have up.

ianna:
> @ianna which version are you talking about? The current version of the PR is the appropriate quick fix that works with cuda and the typetracer backend. The original version that you green-lighted won't work with those two backends that's why we had a very short offline chat with Lindsey and Peter and put the current version of the PR as the "quick fix".
> 
> The “lengthier” fix would be implementing a separate function for the cupy backend and in general re thinking a bit how shape item works. I’m happy to put a test in this one. Without this fix, there are about 400 test failures though with numpy 2.3 they get noticed in the other NumExpr PR I have up.

Thanks for the context — but to be clear, the initial quick fix addressed a specific, documented problem that was described in an issue. This new 'improved fix' currently doesn’t have an associated issue or a reproducer for the backend-specific problems it’s trying to address.

If there’s a deeper problem affecting CUDA and Typetracer, it really needs to be properly tracked in a new issue, with reproducers attached. Otherwise we risk merging untracked changes that solve symptoms without addressing or even documenting the underlying cause.

I’d prefer we hold off merging this version until that’s in place.

ikrommyd:
Well for the cpu backend, I have a reproducer already which I can add as a test. For cuda, it's not tested in ci but I can create a reproducer which shows why the "original" fix fails and you can try it offline. For the typetracer, I'm not 100% sure if we can end up with this failure in actual code, but since we're allowing "integer like" objects in this function in the cpu backend, we might as well allow them in the typetracer case.

ikrommyd:
After I tracked down the use cases of `shape_item_as_index`, it turns out we don't need to change the function but only a single usage of it. If y'all agree we can close this PR cause the most concise fix for numpy 2.3 support is in https://github.com/scikit-hep/awkward/pull/3534.

pfackeldey:
Superseded by https://github.com/scikit-hep/awkward/pull/3534

PR #3532, opened 2025-06-10
test: fix test names

Saransh-cpp:


ikrommyd (approved):

ianna (approved):
@Saransh-cpp - thanks for fixing it! Because the change is trivial, I'm enambling auto-merge. Thanks.

PR #3533, opened 2025-06-11
fix: make awkward compatible with numexpr 2.11

ikrommyd:


ikrommyd:
@ianna @pfackeldey I'm not particularly familiar with numexpr so this is just my first attempt. Do your thing if you have better ideas. This "works" though although I don't know if I'm missing something with numexpr's free-threading support in version 2.11

ikrommyd:
Oh and here the ci fails because of the problems https://github.com/scikit-hep/awkward/pull/3531 is fixing haha

pfackeldey:
@ikrommyd I'd rather have someone else than me have a look at this. I'm not at all familiar with numexpr.

ikrommyd:
> @ikrommyd I'd rather have someone else than me have a look at this. I'm not at all familiar with numexpr.

Yeah I tried to mimic how numexpr implements their own evaluate function by reading their code. I don't know anything more than you I'm afraid.
I opened a relevant issue for help: https://github.com/pydata/numexpr/issues/514

src/awkward/_connect/numexpr.py
```diff
@@ -77,13 +77,26 @@ def evaluate(
 ):
     numexpr = _import_numexpr()
 
+    if parse_version(numexpr.__version__) >= parse_version("2.11.0"):
```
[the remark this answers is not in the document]
henryiii: FWIW, it's also not great to use the `__version__` attribute in code, since a library could decide to remove it. `importlib.metadata.version("numexpr")` should always work.

ikrommyd:
@henryiii @ianna I'm using `importlib` now and also a `_numexpr_version_ge_2_11_0` global variable so that the version is not checked every single time

ikrommyd:
The only test affected by this PR is `tests/test_0119_numexpr_and_broadcast_arrays.py`. Everything else is numpy 2.3

requirements-test-full.txt
```diff
@@ -1,7 +1,7 @@
 fsspec>=2022.11.0;sys_platform != "win32"
 jax[cpu]>=0.2.15;sys_platform != "win32" and python_version < "3.13"
 numba>=0.50.0;sys_platform != "win32" and python_version < "3.13"
-numexpr>=2.7; python_version < "3.13"
+numexpr>=2.7
```
henryiii:
```suggestion
numexpr>=2.7; python_version < "3.14"
```

Would this make sense so we can test on 3.14 soon?

src/awkward/_connect/numexpr.py
```diff
@@ -15,10 +16,12 @@
 from awkward._namedaxis import NAMED_AXIS_KEY, NamedAxesWithDims, _unify_named_axis
 
 _has_checked_version = False
+_numexpr_version_ge_2_11_0 = False
 
 
 def _import_numexpr():
     global _has_checked_version
+    global _numexpr_version_ge_2_11_0
```
henryiii, line 24: Fine for now, since it's following the same pattern as something else, but it's not the right way to do this. The correct way is to make this a function and add `@functools.cache` to it. That will not require `global` and will also be threadsafe, and won't require if statement gymnastics to use.

ikrommyd: Agreed. I think there gotta be a few of such instances for optional libraries throughout the codebase. It can be a "TODO" to do this change everywhere.

requirements-test-full.txt
ikrommyd: I don't know. Whatevere you suggest here

ikrommyd:
For reference, it seems to work fine on free-threaded python.
```
iPython 3.13.3 experimental free-threading build (main, Apr  9 2025, 03:48:13) [Clang 20.1.0 ]
Type 'copyright', 'credits' or 'license' for more information
IPython 9.3.0 -- An enhanced Interactive Python. Type '?' for help.
Tip: You can use Ctrl-O to force a new line in terminal IPython

In [1]: import awkward as ak

In [2]: a = ak.Array([[1,2,3], [], [4,5]])

In [3]: b = ak.Array([1,2,3])

In [4]: ak._connect.numexpr.evaluate("a + b")
Out[4]: <Array [[2, 3, 4], [], [7, 8]] type='3 * var * int64'>

In [5]: ak._connect.numexpr.re_evaluate()
Out[5]: <Array [[2, 3, 4], [], [7, 8]] type='3 * var * int64'>
```

ikrommyd:
image

src/awkward/_connect/numexpr.py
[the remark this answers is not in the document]
ianna: @ikrommyd - thanks for following the suggestion!

src/awkward/_connect/numexpr.py
ianna: moved to an issue

ianna (commented):
@ikrommyd - thanks! I don't know if you want to commit the Copilot suggestions.

ikrommyd:
> @ikrommyd - thanks! I don't know if you want to commit the Copilot suggestions.

Well I did the first suggestion which was to cache the `parse_version` result. The others are not fully complete and I'd rather do exactly what numexpr does here because we're using internal code. I'd rather use internal code exactly like they do because it's easier to track if they make changes. I think I'm happy with how it is right now.

ianna (approved):
@ikrommyd - Thanks for fixing this! Current implementation looks good to me.

PR #3534, opened 2025-06-11
fix: numpy 2.3 compatibility

### LINKED ISSUE #3530: Support NumPy 2.3

ikrommyd:
Fixes https://github.com/scikit-hep/awkward/issues/3530
Supersedes https://github.com/scikit-hep/awkward/pull/3531 

Fixes `np.count_nonzero` returning a numpy integer instead of a python integer in numpy 2.3 and also ignores the warning/error of setting rtol or atol to nan or np.inf in `np.testing.assert_allclose` that is also new in numpy 2.3.

The numexpr pin is temporary to make the ci pass. The numexpr 2.11 compatibility will be fixed in a separate PR.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,7 @@ filterwarnings = [
     "ignore:.*np\\.MachAr.*:DeprecationWarning",
     "ignore:module 'sre_.*' is deprecated:DeprecationWarning",
     "ignore:Jitify is performing a one-time only warm-up",
+    "ignore:One of rtol or atol is not valid:RuntimeWarning",
 ]
 log_cli_level = "info"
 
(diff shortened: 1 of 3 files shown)


pfackeldey (approved):
Thanks this looks good to me!

ianna (commented):
@ikrommyd - I noticed that none of the requests in [#3531](https://github.com/scikit-hep/awkward/pull/3531) was addressed here.

PR #3538, opened 2025-06-12
fix: improve type consistency in `to_packed` for `num_none`

ianna:
This PR ensures that the variable `num_none` in the `to_packed` method consistently holds a plain Python `int`, regardless of the nplike backend (NumPy, CuPy, JAX, etc.). This prevents potential type issues or unexpected behavior when performing downstream operations that expect a native `int`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ classifiers = [
 dependencies = [
     "awkward_cpp==46",
     "importlib_metadata>=4.13.0;python_version < \"3.12\"",
-    "numpy>=1.18.0,<2.3",
+    "numpy>=1.22.0",
     "packaging",
     "typing_extensions>=4.1.0; python_version < \"3.11\"",
     "fsspec>=2022.11.0"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,7 +133,6 @@ filterwarnings = [
     "ignore:.*np\\.MachAr.*:DeprecationWarning",
     "ignore:module 'sre_.*' is deprecated:DeprecationWarning",
     "ignore:Jitify is performing a one-time only warm-up",
-    "ignore:One of rtol or atol is not valid:RuntimeWarning",
 ]
 log_cli_level = "info"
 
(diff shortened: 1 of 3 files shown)


ianna:
```python
import awkward as ak
import numpy as np

content = ak.contents.NumpyArray(np.array([0.0, 1.1, 2.2, 3.3, 4.4]))
index = ak.index.Index64(np.array([2, 2, 0, -1, 4], dtype=np.int64))
array = ak.contents.IndexedOptionArray(index, content)
array.to_packed()
```

pyproject.toml
```diff
@@ -133,6 +133,7 @@ filterwarnings = [
     "ignore:.*np\\.MachAr.*:DeprecationWarning",
     "ignore:module 'sre_.*' is deprecated:DeprecationWarning",
     "ignore:Jitify is performing a one-time only warm-up",
+    "ignore:One of rtol or atol is not valid:RuntimeWarning",
```
ariostas: Is this really something we should be ignoring? Did they change the behavior of `rtol` or `atol`?

ikrommyd: Yeah now if you set atol or rtol to nan or inf in `np.testing.assert_allclose` or `np.isclose` for example, you get a warning in 2.3. We set pytest to error on all warnings so I'm ignoring this one here.

ikrommyd: We do this in one jax test actually only. Let me see if we can set the value to something else

ikrommyd: Well we can set the value `atol=10` to make the tests pass. I don't know. Either that or not raising the warning as error is good.

ianna: IMHO we need to fix it properly, rather then ignoring the warning. Where does it come from?

ianna (commented):
I think, the only part which is relevant here is the change in the IndexedOptionArray.

pyproject.toml
ikrommyd: Well the point is that in this test we wanna check for relative difference and not absolute difference. So the test sets the absolute tolerance to infinity. A “large enough” value has the same effect without the warning.

ikrommyd: If you wanna check anything offline, the affected test is `tests/test_1490_jax_reducers_combinations.py`

ikrommyd: This tests uses `atol=np.inf` in every single `numpy.testing.assert_allclose` call. You'd need `atol=some large enough number to make the test pass` to avoid the warning. That number is 10.

PR #3539, opened 2025-06-12
chore: restrict numpy and numexpr for now

ianna:
As per the upstream pinning recommendation:
https://numpy.org/doc/stable/dev/depending_on_numpy.html#runtime-dependency-version-ranges
See
https://github.com/numpy/numpy/pull/18505
https://github.com/scipy/scipy/pull/12862

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ classifiers = [
 dependencies = [
     "awkward_cpp==46",
     "importlib_metadata>=4.13.0;python_version < \"3.12\"",
-    "numpy>=1.18.0",
+    "numpy>=1.22,<2.3",
     "packaging",
     "typing_extensions>=4.1.0; python_version < \"3.11\"",
     "fsspec>=2022.11.0"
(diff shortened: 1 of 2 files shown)


ikrommyd:
@ianna You need a numexpr pin too for ci to succeed `numexpr>=2.7,<2.11; python_version < "3.13"` in `requirements-test-full.txt`.

henryiii:
I don't recommend an upper cap. And if you really have to, they recommend `<2.5`, not `<2.3`.

pyproject.toml
```diff
@@ -42,7 +42,7 @@ classifiers = [
 dependencies = [
     "awkward_cpp==46",
     "importlib_metadata>=4.13.0;python_version < \"3.12\"",
-    "numpy>=1.18.0",
+    "numpy>=1.22,<2.3",
```
henryiii: I guess we want to bump to 1.22? Is the next release a patch release or minor release?

ianna: in the next release

henryiii (approved):

Issue #3541, opened 2025-06-13
`ak.sort` for JAX Tracer [was: JVP and VJP different between `ak.sort` and `jax.numpy.sort`]


ikrommyd:
### Version of Awkward Array

2.8.3

### Description and code to reproduce

To reproduce:
```py
import awkward as ak
import numpy as np
import jax

ak.jax.register_and_check()

test_regulararray = ak.Array(
    [[1.0, 2.0, 3.0],
     [4.0, 5.0, 6.0],
     [7.0, 8.0, 9.0]],
    backend="jax"
)

test_regulararray_tangent = ak.Array(
    [[1.0, 2.0, 3.0],
     [4.0, 5.0, 6.0],
     [7.0, 8.0, 9.0]],
    backend="jax"
)

test_regulararray_jax = jax.numpy.array(
    [[1.0, 2.0, 3.0],
     [4.0, 5.0, 6.0],
     [7.0, 8.0, 9.0]],
    dtype=np.float64
)

test_regulararray_tangent_jax = jax.numpy.array(
    [[1.0, 2.0, 3.0],
     [4.0, 5.0, 6.0],
     [7.0, 8.0, 9.0]],
    dtype=np.float64
)

axis = 0
func_ak = ak.sort
func_jax = jax.numpy.sort

def func_ak_with_axis(x):
    return func_ak(x, axis=axis)

def func_jax_with_axis(x):
    return func_jax(x, axis=axis)

value_jvp, jvp_grad = jax.jvp(
    func_ak_with_axis, (test_regulararray,), (test_regulararray_tangent,)
)

value_jvp_jax, jvp_grad_jax = jax.jvp(
    func_jax_with_axis, (test_regulararray_jax,), (test_regulararray_tangent_jax,)
)

value_vjp, vjp_func = jax.vjp(func_ak_with_axis, test_regulararray)
value_vjp_jax, vjp_func_jax = jax.vjp(func_jax_with_axis, test_regulararray_jax)

if __name__ == "__main__":
    print("Awkward JVP value:\n", value_jvp)
    print("JAX JVP value:\n", value_jvp_jax)
    print("Awkward JVP grad:\n", jvp_grad)
    print("JAX JVP grad:\n", jvp_grad_jax)
    print("Awkward VJP value:\n", value_vjp)
    print("JAX VJP value:\n", value_vjp_jax)
```
Prints out
```
Awkward JVP value:
[...]

PR #3542, opened 2025-06-13
fix: do not use `atol=np.inf` in a jax test and `xfail` sorting test which fails

ikrommyd:
This has to do with numpy 2.3 which raises a warning about invalid `atol` if it's set to infinity and also the fact that this test is not right and jax and awkward give different values. See https://github.com/scikit-hep/awkward/issues/3541

### DIFF
--- a/tests/test_1490_jax_reducers_combinations.py
+++ b/tests/test_1490_jax_reducers_combinations.py
@@ -55,58 +55,58 @@ def test_reducer(func_ak, axis):
     value_vjp_jax, vjp_func_jax = jax.vjp(func_jax_with_axis, test_regulararray_jax)
 
     numpy.testing.assert_allclose(
-        ak.to_list(value_jvp), value_jvp_jax.tolist(), rtol=1e-9, atol=np.inf
+        ak.to_list(value_jvp), value_jvp_jax.tolist(), rtol=1e-9, atol=1e-9
     )
     numpy.testing.assert_allclose(
-        ak.to_list(value_vjp), value_vjp_jax.tolist(), rtol=1e-9, atol=np.inf
+        ak.to_list(value_vjp), value_vjp_jax.tolist(), rtol=1e-9, atol=1e-9
     )
     numpy.testing.assert_allclose(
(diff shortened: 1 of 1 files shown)


tests/test_1490_jax_reducers_combinations.py
```diff
@@ -55,58 +55,59 @@ def func_jax_with_axis(x):
...
-        ak.to_list(vjp_func(value_vjp)[0]),
-        (vjp_func_jax(value_vjp_jax)[0]).tolist(),
-        rtol=1e-9,
-        atol=np.inf,
-    )
+        atol=1e-9,
+    )
+
+
+# TODO: Enable this when https://github.com/scikit-hep/awkward/issues/3541 is resolved.
```
ikrommyd: Added a todo statement here to bring this back once the issue is resolved.

ianna: @ikrommyd - please, use skip instead of commenting out the test.
```suggestion
@pytest.mark.skip("sorting jax backend gives wrong results, see issue #3541 ")
```

ianna (changes requested):
@ikrommyd - Thanks! Please uncomment the sort tests. It is better to have them on record to remind us. Thanks!

tests/test_1490_jax_reducers_combinations.py
ariostas: Maybe it should be marked `xfail` instead of just skipped to make it more visible

ikrommyd: Oh yea........I forgot `skip` and `xfail` exist haha

ikrommyd: Let's `xfail` it actually. I think it's best

ianna: sounds good!

ikrommyd: Done!

ianna (approved):
@ikrommyd - Great! Thanks for fixing it. Please go ahead and merge it if you finished with it. Thanks!

Issue #3480, opened 2025-04-24
`ak.to_cudf` test failing (`test_3051_to_cuda.py::test_strings`)


ikrommyd:
awkward v2.8.4
```python
cudf-cu12   25.6.0
```

```
FAILED tests-cuda/test_3051_to_cuda.py::test_strings - TypeError: StringColumn.__init__() missing 2 required positional arguments: 'size' and 'dtype'
```

ianna:
```python
    def test_strings():
        arr = ak.Array(["hey", "hi", "hum"])
>       out = ak.to_cudf(arr)
              ^^^^^^^^^^^^^^^

arr        = <Array ['hey', 'hi', 'hum'] type='3 * string'>

tests-cuda/test_3051_to_cuda.py:52: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/ar1092/micromamba/envs/test-env/lib/python3.13/site-packages/awkward/_dispatch.py:41: in dispatch
    with OperationErrorContext(name, args, kwargs):
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        args       = (<Array ['hey', 'hi', 'hum'] type='3 * string'>,)
        dispatch   = <function to_cudf at 0x1496b7ff02c0>
        func       = <function to_cudf at 0x1496b7ff0220>
        kwargs     = {}
        name       = 'ak.to_cudf'
/home/ar1092/micromamba/envs/test-env/lib/python3.13/site-packages/awkward/_errors.py:80: in __exit__
    raise self.decorate_exception(exception_type, exception_value)
        exception_type = <class 'TypeError'>
        exception_value = TypeError("StringColumn.__init__() missing 2 required positional arguments: 'size' and 'dtype'")
        self       = <awkward._errors.OperationErrorContext object at 0x149416fe3890>
        traceback  = <traceback object at 0x149417096800>
/home/ar1092/micromamba/envs/test-env/lib/python3.13/site-packages/awkward/_dispatch.py:42: in dispatch
    gen_or_result = func(*args, **kwargs)
                    ^^^^^^^^^^^^^^^^^^^^^
        args       = (<Array ['hey', 'hi', 'hum'] type='3 * string'>,)
        dispatch   = <function to_cudf at 0x1496b7ff02c0>
[...]

PR #3543, opened 2025-06-13
ci: add workflow to run GPU tests on selfhosted runner

ariostas:
This PR adds a workflow to run the GPU tests on the new CI node at Princeton. There are a couple of things to discuss:

- I intentionally didn't add the new job to `pass-tests.needs` since I think we should first make sure that it is reliable enough that it's not going to end up blocking PR regularly.
- For now, I added it as a job in the usual `test.yml` workflow. This means that it will run for all PRs and pushes to main. There are some security concerns with this, so we could consider moving it to a separate workflow that needs to be triggered by a `workflow_dispatch` trigger, which only people with write permissions can do. And we could also run nightly tests to make sure things are not breaking. But I think it might also be fine to leave it it with the normal workflow for the following reasons:
  - For first-time contributors we need to approve running the tests.
  - I have the runner running inside a container with very limited access to the filesystem. So even if they wipe everything it's fine. I think a bigger concern would be someone trying to mine bitcoin or use the cluster for a DDOS attack or something.
  - Many of us have eyes on this repo, so I think if someone tries to do something we'll find out and stop it pretty quickly.

I don't have permissions to add a runner for this repo, so I'll need someone (probably @ianna) to help me with that. I tested it on [my fork](https://github.com/ariostas/awkward) and it's working well. [Here](https://github.com/ariostas/awkward/actions/runs/15635839088/job/44050581552) are the logs for a test run I did. There's actually one test that fails, so we should look into that.

After we settle all these things, I'll add a page in the wiki describing how to set up the selfhoster runner.

### DIFF
--- a/requirements-test-gpu.txt
+++ b/requirements-test-gpu.txt
@@ -1,7 +1,8 @@
 cudf-cu12
 cupy-cuda12x
 fsspec>=2022.11.0
-numba>=0.50.0
+numba>=0.60
+numba-cuda
 pyarrow
 pytest>=6
 pytest-cov
--- a/requirements-test-gpu.txt
+++ b/requirements-test-gpu.txt
@@ -1,0 +1,9 @@
+cudf-cu12
+cupy-cuda12x
+fsspec>=2022.11.0
+numba>=0.50.0
+pyarrow
+pytest>=6
+pytest-cov
+pytest-xdist
+uproot>=5
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -179,8 +179,8 @@ jobs:
     steps:
       - name: Clean the workspace and mamba
         run: |
-          rm -rf * .[!.]*
-          rm -rf ~/micromamba*
+          rm -rf * .[!.]* || echo "Nothing to clean"
+          rm -rf ~/micromamba* || echo "Nothing to clean"
 
       - uses: actions/checkout@v4
         with:
@@ -193,7 +193,6 @@ jobs:
           init-shell: bash
           create-args: >-
             python=3.13
-            cudatoolkit
 
       - name: Generate build files
         run: |
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -163,6 +163,79 @@ jobs:
         if: (matrix.python-version == '3.9') && (matrix.runs-on == 'ubuntu-latest')
         uses: codecov/codecov-action@v5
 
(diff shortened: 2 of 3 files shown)


ikrommyd:
you will probably be getting this: https://github.com/scikit-hep/awkward/issues/3480 which needs fixing

ariostas:
> you will probably be getting this: https://github.com/scikit-hep/awkward/issues/3480 which needs fixing

Oh yeah, that's the one. I didn't realize it was already reported

ikrommyd:
> > you will probably be getting this: #3480 which needs fixing
> 
> Oh yeah, that's the one. I didn't realize it was already reported

Yeah I found this some time ago on my desktop. I rarely install cudf though.

requirements-test-gpu.txt
```diff
@@ -0,0 +1,9 @@
+cudf-cu12
+cupy-cuda12x
+fsspec>=2022.11.0
+numba>=0.50.0
```
ianna: I wonder if we should pin `numba 0.60.0` so that we could get `numba-cuda`?

ianna (approved):
@ariostas - Great job! I'll add the runner :-)

requirements-test-gpu.txt
ariostas: Okay, sounds good

ianna:
@ariostas - I suggest we mark the `cudf` string test as failed to move on with the PR. Thanks!

```python
tests-cuda/test_3051_to_cuda.py::test_null PASSED                        [ 24%]
Error: test_strings

TypeError: StringColumn.__init__() missing 2 required positional arguments: 'size' and 'dtype'

    ak.to_cudf(
        <Array ['hey', 'hi', 'hum'] type='3 * string'>
    )
```

ariostas:
> I suggest we mark the cudf string test as failed to move on with the PR. Thanks!

Sounds good. I'll make a couple of final tweaks. I'm just trying to simplify the setup a bit.

ariostas:
Okay, this is ready now

ianna:
> Okay, this is ready now

I'll enable auto-merge then. Thanks!

ariostas:
@ianna would the wiki be a good place to write some documentation about how to set up the runner?

ianna:
> @ianna would the wiki be a good place to write some documentation about how to set up the runner?

if it's awkward-specific, I think a maintainer-guide directory in docs would be good.

PR #3545, opened 2025-06-14
fix: cuda atomicMin and atomicMax for <double>s

### LINKED ISSUE #3528: Issues with `min()` and `argmin()` with `cuda` backend
### Version of Awkward Array

2.8.3

When trying to use `ak.min()` and `ak.argmin()` with the `cuda` backend, I am seeing incorrect behavior for `ak.min()`, and and a `CuPyKernel not found` error for `ak.argmin()`. 

This code reproduces the issues:
```python
import awkward as ak

myarr = [[1.1,2.1],[1.2,2.2,-3.2]]
myarr_cpu = ak.Array(myarr,backend="cpu")
myarr_gpu = ak.Array(myarr,backend="cuda")

print("myarr:",myarr)
print("min (cpu):", ak.min(myarr_cpu,axis=1))
print("min (gpu):", ak.min(myarr_gpu,axis=1))
print("min (cpu):", ak.argmin(myarr_cpu,axis=1))
print("min (gpu):", ak.argmin(myarr_gpu,axis=1))
```
If things were working properly, I would expect to see this printed:
```
myarr: [[1.1, 2.1], [1.2, 2.2, -3.2]]
min (cpu): [1.1, -3.2]
min (gpu): [1.1, -3.2]
min (cpu): [0, 2]
min (gpu): [0, 2]
```

However, for the `ak.min()` with GPU I am getting `[1.1, inf]` (just a note in case it is useful: from some preliminary investigation it looks like this issue may be potentially specific to cases that include a value that is negative). 

And, for the `ak.argmin` with GPU I am getting `AssertionError: CuPyKernel not found: ('awkward_reduce_argmin', <class 'numpy.int64'>, <class 'numpy.float64'>, <class 'numpy.int64'>)`.

The full output is shown here:
```
myarr: [[1.1, 2.1], [1.2, 2.2, -3.2]]
min (cpu): [1.1, -3.2]
min (gpu): [1.1, inf]
min (cpu): [0, 2]
  File "/home/k.mohrman/coffea_dir/gpu_studies/columnar_gpu/mre_argmin.py", line 12, in <module>
    print("min (gpu):", ak.argmin(myarr_gpu,axis=1))
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/_dispatch.py", line 41, in dispatch
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/_errors.py", line 80, in __exit__
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/_dispatch.py", line 67, in dispatch
    next(gen_or_result)
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/operations/ak_argmin.py", line 79, in argmin
    return _impl(array, axis, keepdims, mask_identity, highlevel, behavior, attrs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/operations/ak_argmin.py", line 160, in _impl
    out = ak._do.reduce(
          ^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/_do.py", line 294, in reduce
    next = layout._reduce_next(
           ^^^^^^^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/contents/listoffsetarray.py", line 1618, in _reduce_next
    outcontent = trimmed._reduce_next(
                 ^^^^^^^^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/contents/numpyarray.py", line 1137, in _reduce_next
    out = reducer.apply(self, parents, starts, shifts, outlength)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/_reducers.py", line 172, in apply
    array.backend[
  File "/blue/p.chang/k.mohrman/dir_for_miniconda/miniconda3/envs/coffeagpu_env5/lib/python3.12/site-packages/awkward/_backends/cupy.py", line 39, in __getitem__
    raise AssertionError(f"CuPyKernel not found: {index!r}")
AssertionError: CuPyKernel not found: ('awkward_reduce_argmin', <class 'numpy.int64'>, <class 'numpy.float64'>, <class 'numpy.int64'>)


See if this has been reported at https://github.com/scikit-hep/awkward/issues
```

Moelf:
fix #3528

the root cause is that the trick for `float` does not work for `double` due to, eh, IEEE754 ... My understanding is you can't cast `double` into `long long` and still have a total order like you would have for `float -> int`. See also: https://stackoverflow.com/questions/55140908/can-anybody-help-me-with-atomicmin-function-syntax-for-cuda/55145948#55145948

### DIFF
--- a/src/awkward/_connect/cuda/cuda_kernels/cuda_common.cu
+++ b/src/awkward/_connect/cuda/cuda_kernels/cuda_common.cu
@@ -170,14 +170,21 @@ __device__ float atomicMin<float>(float* addr, float value) {
 }
 
 // atomicMin() specialization for double
+// https://stackoverflow.com/a/55145948
 template <>
-__device__ double atomicMin<double>(double* addr, double value) {
-  double old;
-  old = !signbit(value) ? __longlong_as_double(atomicMin((long long int*)addr, __double_as_longlong(value))) :
-      __ull2double_rz(atomicMax((unsigned long long int*)addr, __double2ull_ru(value)));
-  return old;
-}
+__device__ double atomicMin<double>(double* address, double val) {
+    unsigned long long int* addr_as_ull = (unsigned long long int*)address;
+    unsigned long long int old = *addr_as_ull, assumed;
 
+    do {
+        assumed = old;
+        double assumed_val = __longlong_as_double(assumed);
+        if (val >= assumed_val) break;
+        old = atomicCAS(addr_as_ull, assumed, __double_as_longlong(val));
+    } while (assumed != old);
+
+    return __longlong_as_double(old);
+}
 
 // atomicMax() specializations
 template <typename T>
@@ -253,12 +260,20 @@ __device__ float atomicMax<float>(float* addr, float value) {
 }
 
 // atomicMax() specialization for double
+// https://stackoverflow.com/a/55145948
 template <>
-__device__ double atomicMax<double>(double* addr, double value) {
-  double old;
-  old = !signbit(value) ? __longlong_as_double(atomicMax((long long int*)addr, __double_as_longlong(value))) :
-      __ull2double_rz(atomicMin((unsigned long long int*)addr, __double2ull_ru(value)));
-  return old;
+__device__ double atomicMax<double>(double* address, double val) {
+    unsigned long long int* address_as_ull = (unsigned long long int*)address;
+    unsigned long long int old = *address_as_ull, assumed;
+
+    do {
+        assumed = old;
+        double assumed_val = __longlong_as_double(assumed);
+        if (val <= assumed_val) break; // Already larger or equal, no update needed
+        old = atomicCAS(address_as_ull, assumed, __double_as_longlong(val));
+    } while (assumed != old);
+
+    return __longlong_as_double(old);
 }
 
 


Moelf:
related question, how is the CUDA backend tested? If there's CI I would add a test too, but I don't see any... yet

ikrommyd:
It’s not tested in ci yet. @ariostas has an open PR for that. Typically @ianna runs the tests offline. I could sometimes also do it but I’m not in the US. I don’t have access to my desktop.

src/awkward/_connect/cuda/cuda_kernels/cuda_common.cu
```diff
@@ -170,15 +170,22 @@ __device__ float atomicMin<float>(float* addr, float value) {
...
-      __ull2double_rz(atomicMax((unsigned long long int*)addr, __double2ull_ru(value)));
-  return old;
+__device__ double atomicMin<double>(double* address, double val) {
+    unsigned long long int* addr_as_ull = (unsigned long long int*)address;
+    unsigned long long int old = *addr_as_ull, assumed;
+
+    do {
+        assumed = old;
+        double assumed_val = __longlong_as_double(assumed);
+        if (val >= assumed_val) break;
```
RongkunWang, line 182: This is not atomic. Will there be other threads running in paralle that could increase the value the address points to?

ianna:
> related question, how is the CUDA backend tested? If there's CI I would add a test too, but I don't see any... yet

@ariostas has opened a PR for CUDA integration testing

src/awkward/_connect/cuda/cuda_kernels/cuda_common.cu
Moelf: this whole thing is atomic because of the `atomicCAS` I believe

RongkunWang: atomicCAS is atomic, but between `assumed = old`, if the and `val >= assumed_val` comparison, if the underlying value got increased, and `val >= assumed_val`, you would miss to set min to val. 

But this happens only when there's parallel operation where value can be increased. If only `atomicMin` are called in parallel this should be impossible. So it might be worth it to think what `atomicMin` can be called in parallel to.

Moelf: >and val >= assumed_val, you would miss to set min to val.

`atomicCAS` is atomic compare-and-swap, so we always atomically check if the value at `addr_as_ull` is still equal to `assumed`, before we swap in the value known to be smaller than `assumed`.

And this `while-loop` ends either because the value we're holding is larger (no need to try to swap in), or we successfully compare-and-swaped.

RongkunWang: what I was saying is happing at val >= assumed_val, before the `atomicCAS`

RongkunWang: I found this https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p0493r5.pdf which might be useful. you break before the CAS while loop so it looks like 5.1

Moelf: I read that but that paper seems to be focus on efficiency of implementation and in relation to `read-modify-write` semantics, it's not clear to me what's the conclusion, is the implementation in this PR okay or not?


I still think it's okay if value at `address` changes between L180 and L182. Consider the value at `address`, assuming it changed by other threads:

| loc\case | Case1 | Case2 | Case3 |
|--------|--------|--------|--------|
| L180 | > val | > val | <= val |
| L182-184 | > val | <= val | N/A |

- Case 1: loop back because after `atomicCAS` we will find `assumed != old`, so we try again
- Case 2: loop back, same reason as above
- Case 3: `break` because it was already smaller than `val`, doesn't matter what it changed to by other threads.

I don't see how it can result in incorrect result.

RongkunWang: My thinking was in Case 3 but

| loc |  |
| --- | --- |
| L180 | assumed_val <= val, but underlying value the address points to >= val |

But that requires some paralle thread to increase the underlying value which might never be the use case when calling atomicMin. 

Is `atomicMin` always called on all threads with no other function calls?

Moelf: >But that requires some paralle thread to increase the underlying value

Right, and that shouldn't be possible because the only reason to use `atomicMin` is if multiple threads are trying to do `atomicMin` to the same address.

The problem you're describing would still be a problem even if we only have 1 thread but multiple "tasks" running asynchronously, because if some tasks increase the value and some decrease, the final value depends on the order of execution, even if they never run concurrently.

I will mark this issue as solved.

ariostas:
The GPU CI is live now!

ianna:
> The GPU CI is live now!

@ariostas - many thanks! I think, we need to add `tests-cuda-kernels` and `tests-cuda-kernels-explicit` - they are generated.

ariostas:
> I think, we need to add `tests-cuda-kernels` and `tests-cuda-kernels-explicit` - they are generated.

Oh okay, I missed that part. I assumed that it only generated cpu kernels

ianna (approved):
@Moelf - Looks good! Thanks! When https://github.com/scikit-hep/awkward/pull/3551 is merged, please, re-base this one.

ianna:
@all-contributors please add @Moelf  for code

PR #3548, opened 2025-06-16
feat: add benchmarks for awkward

pfackeldey:
This PR adds benchmarks to awkward. 

For every PR it runs a suite of benchmarks and measures some performance metrics between main and the feature branch (merged with main). It reports any improvements & regressions of more than 10% in `cpu_time`. 

It would be nice to extend this to run over all tags so we can track how performance evolves throughout all awkward releases.

An example notification for a performance regression of the `ak.all` operation (axis=None) running on a Jagged array with 65536 elements and dtype="float64" would look like this:
"""
### 🔹 ak.all(Jagged<65536,f64>, axis=None)

**Relative CPU Time Difference:** `692.1%` — 🔴 **Regression**

<details><summary>Show full comparison</summary>

| Metric | benchmark.json | benchmark2.json |
| --- | --- | --- |
| `cpu_time` (ms) | 1.712867e-01 | 1.356710e+00 |
| `real_time` (ms) | 1.712866e-01 | 1.047834e+02 |
| `elements/s` (Hz) | 3.83e+08 | 4.83e+07 |

</details>
"""

This is not yet including all "number-crunching" operations. What we want to benchmark (e.g. which highlevel ops) is up for discussion.

Opening this PR already now to debug the CI.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -81,28 +81,15 @@ jobs:
       - name: Print versions
         run: python -m pip list
 
-      - name: Get PR target branch
-        id: get_target_branch
-        uses: actions/github-script@v7
-        with:
-          result-encoding: string
-          script: |
-            const { data: pullRequest } = await github.rest.pulls.get({
-              owner: context.repo.owner,
-              repo: "awkward",
-              pull_number: "${{ inputs.pr-number }}",
-            });
-            return pullRequest.base.ref;
-
       - name: Run Benchmark and Comparisons
         id: benchmark_and_compare
         shell: bash
         run: |
           cd benchmarks/
           ./run_action.sh
-          echo "comparison='$(cat BASE_OUTPUT_DIR)/comparison.md'" >> $GITHUB_OUTPUT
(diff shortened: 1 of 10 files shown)


pfackeldey:
Hi @ianna,
I think this PR is ready for review. Once this is merged there's one extra CI job that runs benchmarks on our self hosted runner between the current main and the feature branch (merged with main). Potential regressions/improvements will be posted by the bot automatically to the PR like this: https://github.com/scikit-hep/awkward/pull/3550#issuecomment-2980671494

The question is which things do we want to benchmark? Depending on how many functions we want to benchmark the CI job may become rather expensive? 

Also: there may be false positives if the runner is under additional load that may affect the _actual_ runtime. We can adjust the threshold of when the bot is supposed to post a comment of course in the future.

ianna (commented):
@pfackeldey - looks great! as discussed, it merits to be a separate repo. It would give us more flexibility to what we can profile. Thanks!

pfackeldey:
Yes, I'll close this for now and use it as a starting point for a separate repo!

PR #3552, opened 2025-06-17
fix: too many dashes in jupyter repr

### LINKED ISSUE #3529: Too many dashes printed in the repr
This is to keep track of the known issue that we print way too many dashes in the repr of an array with many fields. Example:

Image

pfackeldey:
(Hopefully) closes https://github.com/scikit-hep/awkward/issues/3529.

@ianna & @ikrommyd can you give this a try?

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1497,7 +1497,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         sorted_rows = sorted(rows, key=lambda x: -len(x.split(":")[0]))
 
         n_cols = max(
-            len(line) for line in itertools.chain(header_lines, sorted_rows, type_lines)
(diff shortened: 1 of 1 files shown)


ikrommyd:
Hi @pfackeldey, It helps but I think it's still too many in certain cases. For the same example from the issue:

image

Ideally, it should be as long as the array display right above it right?

pfackeldey:
Could you try again @ikrommyd ?

ikrommyd:
Now I'm getting this. Reduced fields of the record array as well :). I'm happy if that's the intended result.
image

pfackeldey:
> Reduced fields of the record array as well :). I'm happy if that's the intended result.

that was intended, however that only works for the most outer level of fields. At arbitrary depths is non-trivial and I'd say over-engineering. My view is either have this reduced field repr or none.

ianna (approved):
@pfackeldey - Thank you! Looks great! Please merge if you finished with it. Thanks!

PR #3553, opened 2025-06-17
fix: ak.categories should return None if no categories are present

pfackeldey:
See title


ianna (approved):
@pfackeldey - Thanks for catching it! Looks good. Please, merge if you done with it. Thanks!

ianna (approved):
@pfackeldey - Thanks for catching it! Looks good. Please, merge it if you are done with it. Thanks!

PR #3551, opened 2025-06-17
ci: run cuda kernel tests

ariostas:
In #3543 I missed testing the cuda kernel tests that are generated.

PR #3554, opened 2025-06-23
fix: combinations explicit cuda test for `n == 0`

ianna:


ianna (commented):
this is needed for https://github.com/scikit-hep/awkward/pull/3551 to pass. I'm merging it.

PR #3555, opened 2025-06-24
fix: `cudf` strings column

### LINKED ISSUE #3480: `ak.to_cudf` test failing (`test_3051_to_cuda.py::test_strings`)

ianna:
fixes https://github.com/scikit-hep/awkward/issues/3480


ianna (commented):
not sure if we want to support older versions of cudf

ariostas (approved):
Looks good to me!

As for supporting older versions, `cudf` requires Python 3.10, so I guess this would break support for Python 3.9, but since it will reach its EOL in 4 months maybe that's fine.

PR #3558, opened 2025-06-26
fix: use older manylinux for CPython 3.10 and 3.9

henryiii:
As discussed. Minor cleanup as well.

### DIFF
--- a/awkward-cpp/pyproject.toml
+++ b/awkward-cpp/pyproject.toml
@@ -91,10 +91,8 @@ pytest -p no:cacheprovider \
 """]
 skip = [
     "pp*win*",
-    "pp*-manylinux_i686",
 ]
 test-skip = [
-    "*universal2:arm64",
     "pp*",
     "*musllinux*",
 ]
@@ -102,11 +100,13 @@ build-verbosity = 1
 
 [tool.cibuildwheel.environment]
 PIP_ONLY_BINARY = "cmake,numpy"
-
-[[tool.cibuildwheel.overrides]]
-select = "cp313*"
-environment.PYO3_USE_ABI3_FORWARD_COMPATIBILITY = "1"
+UV_ONLY_BINARY = "cmake,numpy"
 
 [[tool.cibuildwheel.overrides]]
 select = "*pyodide*"
 build-frontend = {name = "build", args = ["--exports", "whole_archive"]}
+
+# Use older manylinux for Python 3.9 and 3.10
+[[tool.cibuildwheel.overrides]]
+select = "cp3{9,10}-manylinux*"
+manylinux-x86_64-image = "manylinux2014"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -134,7 +134,7 @@ filterwarnings = [
     "ignore:module 'sre_.*' is deprecated:DeprecationWarning",
     "ignore:Jitify is performing a one-time only warm-up",
 ]
-log_cli_level = "info"
+log_cli_level = "INFO"
 
 [tool.pylint]
 py-version = "3.9"


ianna (approved):
@henryiii - looks great! Thank you.

agoose77:
Hey @henryiii, is it possible for you to briefly explain this pr for my benefit? I'm looking to follow the changes!

awkward-cpp/pyproject.toml
```diff
@@ -91,22 +91,22 @@ pytest -p no:cacheprovider \
 """]
 skip = [
     "pp*win*",
-    "pp*-manylinux_i686",
```
henryiii, line 94: We don't build with 32-bit linux by default anymore in cibuildwheel (and we weren't doing it at all here anyway, like in most packages).

awkward-cpp/pyproject.toml
```diff
@@ -91,22 +91,22 @@ pytest -p no:cacheprovider \
...
 ]
 test-skip = [
-    "*universal2:arm64",
```
henryiii, line 97: We should be building a pair of native wheels, not universal2. Just old cleanup.

awkward-cpp/pyproject.toml
```diff
@@ -91,22 +91,22 @@ pytest -p no:cacheprovider \
...
     "*musllinux*",
 ]
 build-verbosity = 1
 
 [tool.cibuildwheel.environment]
 PIP_ONLY_BINARY = "cmake,numpy"
-
-[[tool.cibuildwheel.overrides]]
-select = "cp313*"
-environment.PYO3_USE_ABI3_FORWARD_COMPATIBILITY = "1"
```
henryiii, line 108: Shouldn't be needed anymore, from early 3.13 days. Cleanup.

awkward-cpp/pyproject.toml
```diff
@@ -91,22 +91,22 @@ pytest -p no:cacheprovider \
...
+UV_ONLY_BINARY = "cmake,numpy"
```
henryiii, line 103: This is probably a good idea, since we use `build[uv]` some of the time, and the variable has this name there.

awkward-cpp/pyproject.toml
```diff
@@ -91,22 +91,22 @@ pytest -p no:cacheprovider \
...
 
 [[tool.cibuildwheel.overrides]]
 select = "*pyodide*"
 build-frontend = {name = "build", args = ["--exports", "whole_archive"]}
+
+# Use older manylinux for Python 3.9 and 3.10
```
henryiii, line 109: This is the 3.0 default change; we are going back to the old manylinux_2014 for 3.9 and 3.10 to reduce issues if someone is running CentOS 7-like machines with older Python. NumPy moved to 2_28 too in the last release, but NumPy only supports 3.11+.

pyproject.toml
```diff
@@ -134,7 +134,7 @@ filterwarnings = [
     "ignore:module 'sre_.*' is deprecated:DeprecationWarning",
     "ignore:Jitify is performing a one-time only warm-up",
 ]
-log_cli_level = "info"
+log_cli_level = "INFO"
```
henryiii, line 137: The SchemaStore schema for this expects upper case. You can see this with `uvx sp-repo-review[all] .` or at https://learn.scientific-python.org/development/guides/repo-review/.

henryiii (commented):
For more on cibuildwheel 3.0, see https://iscinumpy.dev/post/cibuildwheel-3-0-0/.

henryiii:
I'm also working on getting pybind11 3.0 out, but haven't moved over to it here yet since it's not out yet (PyTorch is currently blocking the release). That's needed for Python 3.14, iOS, and GraalPy.

PR #3559, opened 2025-06-26
chore: use PEP 639 license

henryiii:
Use PEP 639 license.

### DIFF
--- a/awkward-cpp/pyproject.toml
+++ b/awkward-cpp/pyproject.toml
@@ -24,7 +24,6 @@ classifiers = [
     "Intended Audience :: Developers",
     "Intended Audience :: Information Technology",
     "Intended Audience :: Science/Research",
-    "License :: OSI Approved :: BSD License",
     "Operating System :: MacOS :: MacOS X",
     "Operating System :: Microsoft :: Windows",
     "Operating System :: POSIX :: Linux",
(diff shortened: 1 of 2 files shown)


ianna (approved):
@henryiii - thank you for fixing this!

awkward-cpp/pyproject.toml
```diff
@@ -44,7 +43,8 @@ classifiers = [
     "Topic :: Software Development",
     "Topic :: Utilities",
 ]
-license = {text = "BSD-3-Clause"}
+license = "BSD-3-Clause AND MIT"
+license-files = ["LICENSE", "rapidjson/license.txt"]
```
agoose77, line 47: Brilliant change, thanks!

awkward-cpp/pyproject.toml
```diff
@@ -70,6 +70,9 @@ sdist.include = [
     "tests-cpu-kernels",
     "tests-cpu-kernels-explicit"
 ]
+sdist.exclude = [
+  "rapidjson/bin",
```
agoose77, line 74: How long has this been there!? I never noticed during packaging.

henryiii: It should only have been in the SDist, it's the source code for some rapidjson tools. Some stuff in bin has a different license, so making sure we exclude it.

PR #3563, opened 2025-06-28
fix: allow arrays with `None`s in single argument usage of `ak.where`

### LINKED ISSUE #3138: `ak.where` errors if array has None
### Version of Awkward Array

2.6.5

In version `2.0.7` I could do something like this:

```python
import numpy as np
import awkward as ak
a = np.arange(12.).reshape(3,4)
a[2,3] = np.nan
b = ak.nan_to_none(a)
ak.where(b == ak.max(b))
```

With a result like this:

```python
(<Array [2] type='1 * int64'>, <Array [2] type='1 * int64'>)
```

In version `2.6.4` and `2.6.5` I get an error like this:

```
  File "/Users/timdiller/test_environment/lib/python3.12/site-packages/awkward/_dispatch.py", line 64, in dispatch
    next(gen_or_result)
  File "/Users/timdiller/test_environment/lib/python3.12/site-packages/awkward/operations/ak_where.py", line 55, in where
    return _impl1(condition, mergebool, highlevel, behavior, attrs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/timdiller/test_environment/lib/python3.12/site-packages/awkward/operations/ak_where.py", line 74, in _impl1
    out = layout.backend.nplike.nonzero(layout.to_backend_array(allow_missing=False))
                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/timdiller/test_environment/lib/python3.12/site-packages/awkward/contents/content.py", line 1020, in to_backend_array
    return self._to_backend_array(allow_missing, backend)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/timdiller/test_environment/lib/python3.12/site-packages/awkward/contents/regulararray.py", line 1277, in _to_backend_array
    out = self._content._to_backend_array(allow_missing, backend)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/timdiller/test_environment/lib/python3.12/site-packages/awkward/contents/indexedoptionarray.py", line 1609, in _to_backend_array
    raise ValueError(
ValueError: Content.to_nplike cannot convert 'None' values to np.ma.MaskedArray unless the 'allow_missing' parameter is set to True

    ak.where(
        <Array [[False, False, False, False], ..., [...]] type='3 * 4 * ?bool'>
    )
```

There is no place I can find to pass `allow_missing`.

This seems to be something different from #3098.

ikrommyd:
Fixes https://github.com/scikit-hep/awkward/issues/3138

This used to use `ak.operations.to_numpy` which allowed `None`s to go through until https://github.com/scikit-hep/awkward/pull/2757 changed it to use `to_backend_array` with `allow_missing=False`. I don't think there is any harm in setting `allow_missing=True` here.


ikrommyd:
All these tests error on main with `ValueError: Content.to_nplike cannot convert 'None' values to np.ma.MaskedArray unless the 'allow_missing' parameter is set to True`

pfackeldey:
This change makes sense to me, but I seem to not understand yet why it was changed in #2757? Maybe there was a reason for this?

ikrommyd:
> This change makes sense to me, but I seem to not understand yet why it was changed in #2757? Maybe there was a reason for this?

@agoose77 is there a reason to not allow missing values for the single argument case of ak.where? I’m tagging since that was your PR.

ikrommyd:
Discussed briefly with Angus at SciPy about this and the reasoning was probably that this was the safest thing to do when `.attrs` was added. However, `.nonzero()` seems to always returns an array with integers and it doesn't matter if the numpy array is masked or not so awkward can always convert it into an awkward array again so it should be safe to enable this.

ianna:
> Discussed briefly with Angus at SciPy about this and the reasoning was probably that this was the safest thing to do when `.attrs` was added. However, `.nonzero()` seems to always returns an array with integers and it doesn't matter if the numpy array is masked or not so awkward can always convert it into an awkward array again so it should be safe to enable this.

Thanks @agoose77 for your time to discuss this. Since I don’t have

ianna (approved):
@ikrommyd - Thanks for adding the tests! They look visually identical to NumPy on my end. It’d be nice to also have a direct comparison with NumPy.

ikrommyd:
> @ikrommyd - Thanks for adding the tests! They look visually identical to NumPy on my end. It’d be nice to also have a direct comparison with NumPy.

added that too! 😃

ianna:
> > @ikrommyd - Thanks for adding the tests! They look visually identical to NumPy on my end. It’d be nice to also have a direct comparison with NumPy.
> 
> added that too! 😃

Thanks @ikrommyd - have you read my message on Slack? Thanks.

ikrommyd:
Oh sorry. I just did no

> > > @ikrommyd - Thanks for adding the tests! They look visually identical to NumPy on my end. It’d be nice to also have a direct comparison with NumPy.
> > 
> > 
> > added that too! 😃
> 
> Thanks @ikrommyd - have you read my message on Slack? Thanks.

Oh sorry. I just did now. I thought your comment here was a request to make this change. My bad.

tests/test_3563_single_argument_ak_where_with_missing_values.py
```diff
@@ -0,0 +1,50 @@
...
+
+import awkward as ak
+from awkward.operations import to_list
+
+
+def test():
+    onedim = ak.Array([1, 2, 3, None, 4, 5])
+    result = ak.where(onedim)
+    assert to_list(result[0]) == [0, 1, 2, 4, 5]
+    assert to_list(result[0]) == ak.to_numpy(onedim).nonzero()[0].tolist()
```
ianna, line 13: @ikrommyd - The comparison here is meant between functions, not data. Specifically, we’re checking that `ak.where` returns the same indices as `np.where` when applied to the same array (after converting to NumPy). This ensures Awkward’s behavior matches NumPy’s for truthy/falsy elements, including handling of `None`.

Example:

```python
onedim = ak.Array([1, 2, 3, None, 4, 5])
assert ak.where(onedim)[0].tolist() == np.where(ak.to_numpy(onedim))[0].tolist()
```

This is about function equivalence, not value equality.

ianna (changes requested):
@ikrommyd - Thanks! Please, see my comment above.

tests/test_3563_single_argument_ak_where_with_missing_values.py
ikrommyd: Agreed. `nonzero` is the right function to use though. Otherwise you have to use `np.ma.where` for the cases where the array is masked and `np.where` otherwise. Numpy recommends to use `.nonzero()` See https://numpy.org/devdocs/reference/generated/numpy.where.html

So this is effectively `assert to_list(ak.where(onedim)[0]) == ak.to_numpy(onedim).nonzero()[0].tolist()` as it should.
Are you asking me to remove the explicit value `assert to_list(result[0]) == [0, 1, 2, 4, 5]` tests?

ianna: It's fine, thanks.

ianna (approved):
@ikrommyd - Thanks!

PR #3564, opened 2025-07-03
fix: length0 array repr

### LINKED ISSUE #3562: Value error thrown by empty array repr
### Version of Awkward Array

main

To reproduce:
```py
In [9]: ak.Array([])
Out[9]: ---------------------------------------------------------------------------
File ~/miniforge3/envs/ak-uproot/lib/python3.12/site-packages/IPython/core/formatters.py:1036, in MimeBundleFormatter.__call__(self, obj, include, exclude)
   1033     method = get_real_method(obj, self.print_method)
   1035     if method is not None:
-> 1036         return method(include=include, exclude=exclude)
   1037     return None
   1038 else:

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/highlevel.py:1507, in Array._repr_mimebundle_(self, include, exclude)
   1503 # the rest of the rows we sort by the length of their '<prefix>:'
   1504 # but we sort it from longest to shortest for _repr_mimebundle_
   1505 sorted_rows = sorted(rows, key=lambda x: -len(x.split(":")[0]))
-> 1507 n_cols = max(map(len, header_lines))
   1508 body_lines = header_lines
   1509 body_lines.append("-" * n_cols)

ValueError: max() iterable argument is empty
<Array [] type='0 * unknown'>
```
Introduced by https://github.com/scikit-hep/awkward/pull/3552

pfackeldey:
closes https://github.com/scikit-hep/awkward/issues/3562

### DIFF
--- a/src/awkward/highlevel.py
+++ b/src/awkward/highlevel.py
@@ -1504,7 +1504,7 @@ class Array(NDArrayOperatorsMixin, Iterable, Sized):
         # but we sort it from longest to shortest for _repr_mimebundle_
         sorted_rows = sorted(rows, key=lambda x: -len(x.split(":")[0]))
 
-        n_cols = max(map(len, header_lines))
+        n_cols = max(map(len, header_lines), default=0)
         body_lines = header_lines
         body_lines.append("-" * n_cols)
         body_lines.extend(sorted_rows)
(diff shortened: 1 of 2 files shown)


ianna (changes requested):
@pfackeldey - Thanks for fixing it! Could you please add a test? Thanks.

ianna (approved):
@pfackeldey - Thanks for fixing it!

PR #3565, opened 2025-07-03
refactor: define common nplike method to get pointer to memory

### LINKED ISSUE #3413: `ctypes` abstraction is not scoped to NumPy only
It seems like we some-where rely on `array.ctypes` internally. We should try and abstract that to inside the nplike / kernels.

pfackeldey:
Defines a `memory_ptr` function for each nplike and thus unifies their interface throughout the codebase (as suggested in https://github.com/scikit-hep/awkward/issues/3413). 

Closes https://github.com/scikit-hep/awkward/issues/3413

### DIFF
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -73,7 +73,7 @@ class NumpyKernel(BaseKernel):
             if numpy.is_own_array(x):
                 assert numpy.is_c_contiguous(x), "kernel expects contiguous array"
                 if x.ndim > 0:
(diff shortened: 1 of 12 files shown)


ikrommyd:
@pfackeldey I like this abstraction a lot actually!

ianna (approved):
@pfackeldey - This is great! Thanks for cleaning it up!

ianna:
@pfackeldey - please feel free to merge it if you are done with it. Thanks!

PR #3566, opened 2025-07-03
fix: first maybe materialize, then check for PlaceholderArray

pfackeldey:
This was a logical error. First, we have to materialize if it's a VirtualArray, then we should check for PlaceholderArrays. Otherwise we won't catch the case where a VirtualArray would materialize into a PlaceholderArray...

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -192,10 +192,10 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
         *,
         dtype: DTypeLike | None = None,
     ) -> ArrayLikeT:
+        start, stop, step = materialize_if_virtual(start, stop, step)
         assert not isinstance(start, PlaceholderArray)
         assert not isinstance(stop, PlaceholderArray)
         assert not isinstance(step, PlaceholderArray)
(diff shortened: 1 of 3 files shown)


ikrommyd:
I never actually expected virtual arrays to materialize into placeholders in the first place, that's why I didn't do it. I thought the assertion was higher in priority (if it's a placeholder, stop now and don't run any extra lines of code).

ikrommyd:
Materializing into placeholder doesn't work by construction: https://github.com/scikit-hep/awkward/blob/473029b8cc7345083c29186e0c3def874eb3a9f2/src/awkward/_nplikes/virtual.py#L138. I guess we may want this at some point in our lives so we better move the assertion now?

pfackeldey:
Whether it's useful or not to materialize a PlaceholderArray is a different question. In any case, I think we should aim for logical correctness here, that's what this PR does. I'm not sure yet what we do in future with PlaceholderArrays and VirtualArrays interaction, but that doesn't stop us from making this correct already now.

ianna (approved):
@pfackeldey - Good catch! Thank you for fixing it!

ianna:
@pfackeldey - please, feel free to merge it if you are done with it. Thanks!

PR #3568, opened 2025-07-05
fix: overflow error when concatenating more than 128 regular arrays

ikrommyd:
Now actually finalize https://github.com/scikit-hep/awkward/issues/3310

Instantiate `prototype` in the same way we do for `tags_cls`.
Without it, we get
```
  File "/Users/nmangane/scikit-hep-dev-3/.pixi/envs/default/lib/python3.13/site-packages/awkward/operations/ak_concatenate.py", line 262, in action
    prototype[start : start + size] = tag
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^
OverflowError: Python integer 128 out of bounds for int8

    ak.concatenate(
        [<Array [[False], [False], ..., [False]] type='20 * [1 * bool, parame...
        axis = 1
    )
```


tests/test_3312_concatenate_regular_arrays.py
```diff
@@ -9,3 +9,11 @@ def test():
     ak.concatenate([ak.Array([i])[:, np.newaxis] for i in range(127)], axis=1)
 
     ak.concatenate([ak.Array([i])[:, np.newaxis] for i in range(128)], axis=1)
+
+    ak.concatenate([ak.Array([i])[:, np.newaxis] for i in range(129)], axis=1)
```
ianna: @ikrommyd - please, make a separate test file. Please, also include a test where we cannot reduce the number of arrays to less then 128. Thanks!

ianna (changes requested):
@ikrommyd - Looks good! Thanks for providing the fix so quickly! This is the right solution since all the kernels involved have specialization for `int64` **and** both `UnionArray` functions `regular_index` and `simplified` allow Index64 as tags. It would help if we add a test where `simplified` fails to reduce to no more than 128 variants. Please see the comments. Thanks!
```
# Note: to help merge more than 128 arrays, tags *can* have type ak.index.Index64.
# This is only supported when index is also Index64,
# and all indexed contents are also Index64.
# We still require that this reduces to no more than 128 variants.
```

tests/test_3312_concatenate_regular_arrays.py
ikrommyd: Hi Ianna, for clarification, so a test where we expect it to fail would be a case where we have Index64 tags but not Index64 index and also indexed contents that are also not Index64?

ianna: It would be a case where we have 129 arrays that can’t be simplified and the result would have to be a UnionArray in this case. It’s an edge case. Thanks!

ianna (approved):
@ikrommyd - Thanks for fixing it!

PR #3575, opened 2025-07-15
fix: patch emscripten RPath

henryiii:
Try different RPath config.


ianna (approved):
@henryiii - thank you for fixing it! Please merge it if you finished with it. Thanks.

PR #3580, opened 2025-07-18
feat: add `with_name` to `ak.concatenate`

ikrommyd:
It's pretty common that people want to do a `ak.with_name` after `ak.concatenate` when concatenating different types of particles. However, it's not immediately obvious to them that they should and what it does. This provides a shortcut for this common operation and makes it also easier for people to take a look at `ak.with_name` overall.

NEEDS TESTS


ikrommyd:
Eh, it's not super useful. It's just an extra one liner to the user to wrap an `ak.concatenate` with a `ak.with_name`.

PR #3581, opened 2025-07-18
fix: removal of `pyarrow.lib.PyExtensionType` in pyarrow 21

### LINKED ISSUE #3579: `pyarrow.lib.PyExtensionType` has been removed in pyarrow 21.0.0
Awkward ci errors with `AttributeError: module 'pyarrow.lib' has no attribute 'PyExtensionType'. Did you mean: 'ExtensionType'?` See https://github.com/scikit-hep/awkward/actions/runs/16376816863 for example.
`PyExtensionType` has been removed in pyarrow 21.0.0 and the docs recommend to use `ExtensionType`.
See for example the docs before removal in pyarrow 20: https://arrow.apache.org/docs/python/generated/pyarrow.PyExtensionType.html
We use it in this file: https://github.com/scikit-hep/awkward/blob/525acf3a3dff8b4b535026377ac435d75affd1fd/src/awkward/_connect/pyarrow/conversions.py

ikrommyd:
Fixes https://github.com/scikit-hep/awkward/issues/3579

### DIFF
--- a/src/awkward/_connect/pyarrow/conversions.py
+++ b/src/awkward/_connect/pyarrow/conversions.py
@@ -2,13 +2,11 @@
 
 from __future__ import annotations
 
-import importlib
 import json
 from collections.abc import Iterable, Sized
 from functools import lru_cache
 
 import pyarrow
-from packaging.version import parse as parse_version
 
 import awkward as ak
 from awkward._backends.numpy import NumpyBackend
@@ -24,8 +22,6 @@ from .extn_types import (
 
 np = NumpyMetadata.instance()
 numpy = Numpy.instance()
-_pyarrow_version = importlib.metadata.version("pyarrow")
-_pyarrow_version_lt_21_0_0 = parse_version(_pyarrow_version) < parse_version("21.0.0")
 
 
 def and_validbytes(validbytes1, validbytes2):
@@ -190,14 +186,16 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
     if awkwardarrow_type is not None:
         paarray = paarray.storage
     ### Beginning of the big if-elif-elif chain!
-    if _pyarrow_version_lt_21_0_0 and isinstance(
-        storage_type, pyarrow.lib.PyExtensionType
-    ):
-        raise ValueError(
-            "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
-        )
+    try:
+        if isinstance(storage_type, pyarrow.lib.PyExtensionType):
+            raise ValueError(
+                "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
+            )
+    except AttributeError:
+        # pyarrow >= 21.0.0 does not have PyExtensionType.
+        pass
 
-    elif isinstance(storage_type, pyarrow.lib.ExtensionType):
+    if isinstance(storage_type, pyarrow.lib.ExtensionType):
         # AwkwardArrowType should already be unwrapped; this must be some other ExtensionType.
         assert not isinstance(storage_type, AwkwardArrowType)
         # In that case, just ignore its logical type and use its storage type.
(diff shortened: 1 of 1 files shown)


ikrommyd:
@ianna @henryiii Is this the right fix here? `PyExtensionType` has been removed in pyarrow 21 and the recommendation in their docs is to use `ExtensionType`. We deal with `ExtensionType` right below the lines I changed. The original code is
```py
    if isinstance(storage_type, pyarrow.lib.PyExtensionType):
        raise ValueError(
            "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
        )
    elif isinstance(storage_type, pyarrow.lib.ExtensionType):
        ...
)
```
My idea was to do the first isinstance check only if pyarrow < 21

ianna (changes requested):
@ikrommyd - thanks for looking into it. I think a better way would be to wrap it in a try block.

```python 
try:
    # If PyExtensionType exists and matches, raise immediately
    pyext_type = pyarrow.lib.PyExtensionType  # May trigger AttributeError
    if isinstance(storage_type, pyext_type):
        raise ValueError(
            "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
        )
except AttributeError:
    # PyExtensionType doesn't exist (likely newer PyArrow)
    pass

#  Handle remaining ExtensionType cases
if isinstance(storage_type, pyarrow.lib.ExtensionType):
    assert not isinstance(storage_type, AwkwardArrowType)
    # Logic for generic ExtensionType fallback
    return
```

ikrommyd:
> @ikrommyd - thanks for looking into it. I think a better way would be to wrap it in a try block.
> 
> ```python
> try:
>     # If PyExtensionType exists and matches, raise immediately
>     pyext_type = pyarrow.lib.PyExtensionType  # May trigger AttributeError
>     if isinstance(storage_type, pyext_type):
>         raise ValueError(
>             "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
>         )
> except AttributeError:
>     # PyExtensionType doesn't exist (likely newer PyArrow)
>     pass
> 
> #  Handle remaining ExtensionType cases
> if isinstance(storage_type, pyarrow.lib.ExtensionType):
>     assert not isinstance(storage_type, AwkwardArrowType)
>     # Logic for generic ExtensionType fallback
>     return
> ```

Ah yeah that's possible since an `AttributeError` cannot come from somewhere else since we're not running any code in the try block.

src/awkward/_connect/pyarrow/conversions.py
```diff
@@ -186,13 +186,16 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
     if awkwardarrow_type is not None:
         paarray = paarray.storage
     ### Beginning of the big if-elif-elif chain!
+    try:
+        if isinstance(storage_type, pyarrow.lib.PyExtensionType):
+            raise ValueError(
+                "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
+            )
+    except AttributeError:
+        # pyarrow >= 21.0.0 does not have PyExtensionType.
```
ianna:
```suggestion
        # PyExtensionType is deprecated in pyarrow 21.0.0.
```

src/awkward/_connect/pyarrow/conversions.py
```diff
@@ -495,12 +498,16 @@ def popbuffers(paarray, awkwardarrow_type, storage_type, buffers, generate_bitma
...
-        raise ValueError(
-            "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
-        )
+    try:
+        if isinstance(storage_type, pyarrow.lib.PyExtensionType):
+            raise ValueError(
+                "Arrow arrays containing pickled Python objects can't be converted into Awkward Arrays"
+            )
+    except AttributeError:
+        # pyarrow >= 21.0.0 does not have PyExtensionType.
```
ianna:
```suggestion
        # PyExtensionType is deprecated in pyarrow 21.0.0.
```

ianna (changes requested):
@ikrommyd - thanks for making this change. Please see minor changes to the comments.

ikrommyd:
> @ikrommyd - thanks for making this change. Please see minor changes to the comments.

Done.

ianna (approved):
@ikrommyd - this looks great! Thanks for fixing it! Please merge it if you are done with it. Thanks

PR #3586, opened 2025-07-22
fix: filter also for new (v0.7.0) LinearizeTracer type

### LINKED ISSUE #3585: JAX 0.7.0 incompatibility
### Version of Awkward Array

main

The new version of JAX is causing the CI to fail (see [here](https://github.com/scikit-hep/awkward/actions/runs/16455550058/job/46511578596)).

pfackeldey:
Closes #3585

### DIFF
--- a/src/awkward/_kernels.py
+++ b/src/awkward/_kernels.py
@@ -6,6 +6,8 @@ import ctypes
 from abc import abstractmethod
 from typing import Any, Callable
 
+from packaging.version import parse as parse_version
+
 import awkward as ak
 from awkward._nplikes.cupy import Cupy
 from awkward._nplikes.jax import Jax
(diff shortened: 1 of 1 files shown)


pfackeldey:
Ref: https://docs.jax.dev/en/latest/changelog.html#jax-0-7-0-july-22-2025

I have to say though... the way our JAX backend works is rather fragile. I wouldn't be surprised about more issues like this in the future. (We should think about it again.)

ianna:
> Ref: https://docs.jax.dev/en/latest/changelog.html#jax-0-7-0-july-22-2025
> 
> I have to say though... the way our JAX backend works is rather fragile. I wouldn't be surprised about more issues like this in the future. (We should think about it again.)

Yep, totally agree. Thanks for fixing it!

ianna (approved):
@pfackeldey - Thanks for fixing it. We need to discuss what we should do with JAX backend.

PR #3588, opened 2025-07-24
docs: add missing funding badges

ariostas:
I added a few more shields for grants that have supported Awkward development.


ianna (approved):
@ariostas - Thanks for adding the credits! I'll merge it.

PR #3592, opened 2025-07-28
feat: adapt `__array__` according to numpy guidelines

### LINKED ISSUE #3583: `__array__` needs to be updated to accept `copy` kwarg
### Version of Awkward Array

main

`__array__` needs to be updated to conform with NumPy 2.0 requirements. 

Doing the following,
```python
import awkward as ak
import numpy as np

a = ak.Array([1,2,3])
b = np.array(a)
```
you get the following warning
```
DeprecationWarning: __array__ implementation doesn't accept a copy keyword, so passing copy=False failed. __array__ must implement 'dtype' and 'copy' keyword arguments. To learn more, see the migration guide https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword
```

ikrommyd:
Closes https://github.com/scikit-hep/awkward/issues/3583

Implements `__array__` according to the guidelines specified by https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword and https://numpy.org/devdocs/user/basics.interoperability.html#the-array-method

### DIFF
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -509,20 +509,14 @@ def action_for_matmul(inputs):
 
 def convert_to_array(layout, dtype=None, copy=None):
     out = ak.operations.to_numpy(layout, allow_missing=False)
-    needs_copy_for_dtype = dtype is not None and out.dtype != dtype
-    if copy is False:
-        if needs_copy_for_dtype:
+    if copy:
+        return numpy.array(out, dtype=dtype, copy=True)
+    elif copy is None:
+        return numpy.asarray(out, dtype=dtype)
+    else:
+        if getattr(out, "dtype", dtype) != dtype:
             raise ValueError(
                 "Unable to avoid copy while creating an array as requested"
             )
-        return out
-    elif copy is True:
-        if needs_copy_for_dtype:
-            return out.astype(dtype)
         else:
-            return out.copy()
-    else:
-        if needs_copy_for_dtype:
-            return out.astype(dtype)
-        else:
-            return out
+            return numpy.asarray(out, dtype=dtype)
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -514,7 +514,7 @@ def convert_to_array(layout, dtype=None, copy=None):
     elif copy is None:
         return numpy.asarray(out, dtype=dtype)
     else:
-        if getattr(out, "dtype", dtype) != dtype:
+        if dtype is not None and getattr(out, "dtype", dtype) != dtype:
             raise ValueError(
                 "Unable to avoid copy while creating an array as requested"
             )
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -507,9 +507,6 @@ def action_for_matmul(inputs):
     raise NotImplementedError
 
 
-def convert_to_array(layout, dtype=None):
+def convert_to_array(layout, dtype=None, copy=None):
     out = ak.operations.to_numpy(layout, allow_missing=False)
-    if dtype is None:
-        return out
-    else:
-        return numpy.array(out, dtype=dtype)
+    return numpy.asarray(out, dtype=dtype, copy=copy)
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -513,7 +513,7 @@ def convert_to_array(layout, dtype=None, copy=None):
     if copy is False:
         if needs_copy_for_dtype:
             raise ValueError(
-                "ValueError: Unable to avoid copy while creating an array as requested"
+                "Unable to avoid copy while creating an array as requested"
             )
         return out
     elif copy is True:
--- a/src/awkward/_connect/numpy.py
+++ b/src/awkward/_connect/numpy.py
@@ -509,4 +509,20 @@ def action_for_matmul(inputs):
 
 def convert_to_array(layout, dtype=None, copy=None):
     out = ak.operations.to_numpy(layout, allow_missing=False)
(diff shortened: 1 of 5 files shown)


src/awkward/_connect/numpy.py
```diff
@@ -507,9 +507,22 @@ def action_for_matmul(inputs):
     raise NotImplementedError
 
 
-def convert_to_array(layout, dtype=None):
+def convert_to_array(layout, dtype=None, copy=None):
     out = ak.operations.to_numpy(layout, allow_missing=False)
```
ikrommyd, line 511: Waiting to find out from the numpy team whether `__array__` should return masked arrays if that's an option. That would mean that `allow_missing=True` is the correct thing here.

ikrommyd: It looks like the interface for masked arrays is still a bit ill-defined in numpy so it's best to keep this `False`.

ianna (changes requested):
@ikrommyd - thanks for the PR—appreciate the enthusiasm! Just a heads-up: I’ve marked it with requested changes so we don’t accidentally merge it before discussing. Going forward, let’s focus on the assigned issues first and loop in a quick chat before tackling anything extra. That way we’re all aligned and can make sure your efforts land exactly where they’re needed.

pfackeldey:
Thanks @ikrommyd,

this looks good to me!
`__array__` will be triggered if anything tries to convert the awkward array to a numpy array implicitly. At that point you can't always use `ak.to_numpy` (because that may happen in 3rd party libraries, e.g. matplotlib maybe?), thus we have to make sure that `__array__` works correctly.

ianna:
> Thanks @ikrommyd,
> 
> this looks good to me! `__array__` will be triggered if anything tries to convert the awkward array to a numpy array implicitly. At that point you can't always use `ak.to_numpy` (because that may happen in 3rd party libraries, e.g. matplotlib maybe?), thus we have to make sure that `__array__` works correctly.

@pfackeldey - thanks for checking this. Let’s discuss it at our next meeting. Implicit copy is something we want to watch out for or actively prevent.

ikrommyd:
I would like to also add that before this PR, we were copying more than we needed because of this:
```py
    if dtype is None:
        return out
    else:
        return numpy.array(out, dtype=dtype)
```
That meant that `np.asarray(some awkward array with int64 dtype already, dtype=np.int64)`would be copied because `numpy.array` always copies and that's very unnecessary. This statement "either performs a zero-copy conversion or raises an error."  in the `__array__` docstring was also wrong because it would actually copy more than it should even. Numpy clearly recommends to error only if `copy=False` and a copy is required. `copy=None` should only copy if required and never error and `copy=True` should always copy. This is the expected `__array__` behavior from numpy.

ikrommyd:
Needs tests

ikrommyd:
Hi @ianna, I have added tests for both numpy2 and numpy1 because of slight differences regarding where the `copy` kwarg is available and its possible values.

src/awkward/_connect/numpy.py
```diff
@@ -507,9 +507,16 @@ def action_for_matmul(inputs):
...
-    if dtype is None:
-        return out
+    if copy:
+        return numpy.array(out, dtype=dtype, copy=True)
+    elif copy is None:
+        return numpy.asarray(out, dtype=dtype)
     else:
-        return numpy.array(out, dtype=dtype)
+        if dtype is not None and getattr(out, "dtype", dtype) != dtype:
+            raise ValueError(
```
pfackeldey, line 518: what's the reason to error here? Shouldn't it copy to change the dtype if another dtype is requested? In principle this may catch cases where we accidentally pass the wrong dtype, but in some cases that may even be intended and then erroring would be wrong.
I'm not sure what the policy should be...

If some 3rd party library wants to promote to `f64` and the `ak.Array` is `f32` this would error as far as I understand, but there may be a good reason to do that and to copy then - so, I'm not sure about this error here.

ikrommyd: That's if `copy` is `None` (right above). If `copy` is `False`, numpy wants you to error if you need to copy.

ikrommyd:
```
In [3]: x = np.arange(10, dtype=np.int32)

In [4]: np.asarray(x, dtype=np.int64)
Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [5]: np.asarray(x, dtype=np.int64, copy=False)
Cell In[5], line 1
----> 1 np.asarray(x, dtype=np.int64, copy=False)

ValueError: Unable to avoid copy while creating an array as requested.
If using `np.array(obj, copy=False)` replace it with `np.asarray(obj)` to allow a copy when needed (no behavior change in NumPy 1.x).
For more details, see https://numpy.org/devdocs/numpy_2_0_migration_guide.html#adapting-to-changes-in-the-copy-keyword.
```

pfackeldey: Ah, good point! Yes, this is then a logical error otherwise, I agree :+1:

pfackeldey (approved):
Looks good to me! Thanks @ikrommyd

ikrommyd:
Anyone know what to do about the docs failure? It looks like it's timing out when reading the chicago taxi parquet dataset.

ianna (approved):
@ikrommyd - thanks for addressing the comments! Looks good to me.

ikrommyd:
Seems like the docs building ci finally didn't crash the last time. Are we good to merge this? 😃

ianna:
> Seems like the docs building ci finally didn't crash the last time. Are we good to merge this? 😃

Please hold on. We’ll do it later

Issue #3429, opened 2025-03-19
Add `VirtualArray` in the input/output signature of the nplikes' functions when required


ianna:
I'm pretty sure _all_ of our nplike functions will now need `| VirtualArray` in the input signature. Can you update the nplikes, and the array modules, etc?

_Originally posted by @agoose77 in https://github.com/scikit-hep/awkward/pull/3364#discussion_r2002951768_

PR #3597, opened 2025-08-03
fix: properly handle `VirtualArray` in `nplike.asarray`

ikrommyd:
I realized that `nplike.asarray` does not properly handle `VirtualArray`. A copy should always be created when `copy=True` and that's not the case. The virtual array code branch in `asarray` never cared about copying the underying buffer at all. The codebase often does `nplike.asarray(some object, copy=True)` exactly because it needs a copy to do in-place modifications. This would cause problems if `some object` is a virtual array.

Also very lightly refactoring `reshape` without changing its functionality just to match the general style of other functions that create new virtual arrays.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -365,7 +365,7 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
                     self,
                     next_shape,
                     x.dtype,
-                    lambda: self.reshape(x.materialize(), next_shape, copy=copy),  # type: ignore[union-attr]
+                    lambda: self.reshape(x.materialize(), next_shape, copy=copy),  # type: ignore[arg-type]
                     None,
                 )
 
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -347,17 +347,17 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             next_shape = self._compute_compatible_shape(shape, x.shape)
             return PlaceholderArray(self, next_shape, x.dtype, x._field_path)
         if isinstance(x, VirtualArray):
-            if not x.is_materialized:
+            if x.is_materialized:
+                return self.reshape(x.materialize(), shape, copy=copy)  # type: ignore[arg-type]
+            else:
                 next_shape = self._compute_compatible_shape(shape, x.shape)
                 return VirtualArray(
                     self,
                     next_shape,
                     x.dtype,
-                    lambda: self.reshape(x.materialize(), next_shape),  # type: ignore[union-attr]
+                    lambda: self.reshape(x.materialize(), next_shape),  # type: ignore[arg-type]
                     lambda: next_shape,
                 )
-            else:
-                x = x.materialize()  # type: ignore[assignment]
 
         if copy is None:
             return self._module.reshape(x, shape)
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -68,13 +68,19 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             if obj.is_materialized:
                 return self.asarray(obj.materialize(), dtype=dtype, copy=copy)
             else:
-                return VirtualArray(
-                    obj._nplike,
-                    obj._shape,
-                    obj._dtype if dtype is None else dtype,
-                    lambda: self.asarray(obj.materialize(), dtype=dtype, copy=copy),
-                    lambda: obj.shape,
-                )
+                # if we are not copying and the dtype is _exactly_ the dtype of the existing array
+                # or dtype is None, we can return the VirtualArray directly
+                # this avoids unnecessary VirtualArray creation and method-chaining
+                if not copy and (obj.dtype == dtype or dtype is None):
+                    return obj
+                else:
+                    return VirtualArray(
+                        obj._nplike,
+                        obj._shape,
+                        obj._dtype if dtype is None else dtype,
+                        lambda: self.asarray(obj.materialize(), dtype=dtype, copy=copy),
+                        lambda: obj.shape,
+                    )
         if copy:
             return self._module.array(obj, dtype=dtype, copy=True)
         elif copy is None:
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -73,14 +73,13 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
                 # this avoids unnecessary VirtualArray creation and method-chaining
                 if not copy and (obj.dtype == dtype or dtype is None):
                     return obj
-                else:
-                    return VirtualArray(
-                        obj._nplike,
-                        obj._shape,
-                        obj._dtype if dtype is None else dtype,
-                        lambda: self.asarray(obj.materialize(), dtype=dtype, copy=copy),
-                        lambda: obj.shape,
-                    )
+                return VirtualArray(
+                    obj._nplike,
+                    obj._shape,
+                    obj._dtype if dtype is None else dtype,
+                    lambda: self.asarray(obj.materialize(), dtype=dtype, copy=copy),
+                    lambda: obj.shape,
+                )
         if copy:
             return self._module.array(obj, dtype=dtype, copy=True)
         elif copy is None:
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -68,11 +68,6 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             if obj.is_materialized:
                 return self.asarray(obj.materialize(), dtype=dtype, copy=copy)
             else:
-                # if we are not copying and the dtype is _exactly_ the dtype of the existing array
-                # or dtype is None, we can return the VirtualArray directly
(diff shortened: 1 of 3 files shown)


src/awkward/_nplikes/array_module.py
```diff
@@ -66,24 +66,21 @@ def asarray(
...
+                    obj._dtype if dtype is None else dtype,
+                    lambda: self.asarray(obj.materialize(), dtype=dtype, copy=copy),
+                    lambda: obj.shape,
+                )
         if copy:
             return self._module.array(obj, dtype=dtype, copy=True)
         elif copy is None:
             return self._module.asarray(obj, dtype=dtype)
         else:
-            if getattr(obj, "dtype", dtype) != dtype:
```
ikrommyd, line 86: This is because this doesn't behave like numpy otherwise. See this:
```py
In [1]: from awkward._nplikes.numpy import Numpy

In [2]: import numpy as np

In [3]: numpy = Numpy.instance()

In [4]: x = np.arange(10)

In [5]: np.asarray(x, dtype=None, copy=False)
Out[5]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [6]: numpy.asarray(x, dtype=None, copy=False)
Cell In[6], line 1
----> 1 numpy.asarray(x, dtype=None, copy=False)

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/_nplikes/array_module.py:87, in ArrayModuleNumpyLike.asarray(self, obj, dtype, copy)
     85 else:
     86     if getattr(obj, "dtype", dtype) != dtype:
---> 87         raise ValueError(
     88             "asarray was called with copy=False for an array of a different dtype"
     89         )
     90     else:
     91         return self._module.asarray(obj, dtype=dtype)

ValueError: asarray was called with copy=False for an array of a different dtype
```

src/awkward/_nplikes/array_module.py
```diff
@@ -66,24 +66,21 @@ def asarray(
...
+            if dtype is not None and getattr(obj, "dtype", dtype) != dtype:
```
ikrommyd, line 89: This is because this doesn't behave like numpy otherwise. `dtype=None` and `copy=False` is a valid combination. See this:
```py
In [1]: from awkward._nplikes.numpy import Numpy

In [2]: import numpy as np

In [3]: numpy = Numpy.instance()

In [4]: x = np.arange(10)

In [5]: np.asarray(x, dtype=None, copy=False)
Out[5]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])

In [6]: numpy.asarray(x, dtype=None, copy=False)
Cell In[6], line 1
----> 1 numpy.asarray(x, dtype=None, copy=False)

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/_nplikes/array_module.py:87, in ArrayModuleNumpyLike.asarray(self, obj, dtype, copy)
     85 else:
     86     if getattr(obj, "dtype", dtype) != dtype:
---> 87         raise ValueError(
     88             "asarray was called with copy=False for an array of a different dtype"
     89         )
     90     else:
     91         return self._module.asarray(obj, dtype=dtype)

ValueError: asarray was called with copy=False for an array of a different dtype
```

ianna (changes requested):
@ikrommyd - thanks for looking into it. This should be a part of the whole `VirtualArray` code refactoring.
I'm adding 'request changes' because we don't want any patches while refactoring. I think, it would be better to move it to a list of issues. Thanks for understanding!

ikrommyd:
This is not about refactoring. Not copying when the code should copy is just wrong and can potentially lead to silently wrong results when the code wants to do in place modifications to arrays that should be a copy but are not. This is a somewhat “critical” fix potentially although I don’t know how often such an error might happen. Please take it over and do as you please with it.

ikrommyd:
This is very dangerously wrong and this PR fixes it:
```py
In [1]: import awkward as ak
   ...: import numpy as np
   ...:
   ...:
   ...: numpy = Numpy.instance()
   ...:
   ...: def shape_generator():
   ...:     return (10,)
   ...:
   ...: def data_generator():
   ...:

In [2]: v2 = numpy.asarray(v, copy=True)

In [3]: v2[0] = 111
CALLING DATA GENERATOR!

In [4]: v2
Out[4]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))

In [5]: v
```

src/awkward/_nplikes/array_module.py
```diff
@@ -66,24 +66,21 @@ def asarray(
             return obj
         if isinstance(obj, VirtualArray):
             if obj.is_materialized:
-                obj = obj.materialize()
```
ikrommyd, line 70: I just find it nicer to pass it down via recursion here like we do with `ascontiguousarray` instead of `obj = obj.materialize()`

src/awkward/_nplikes/array_module.py
```diff
@@ -350,17 +347,17 @@ def reshape(
             next_shape = self._compute_compatible_shape(shape, x.shape)
             return PlaceholderArray(self, next_shape, x.dtype, x._field_path)
         if isinstance(x, VirtualArray):
-            if not x.is_materialized:
+            if x.is_materialized:
+                return self.reshape(x.materialize(), shape, copy=copy)  # type: ignore[arg-type]
+            else:
```
ikrommyd, line 347: Again It's just nicer to pass it down via recursion here like we do with `ascontiguousarray`.

src/awkward/_nplikes/array_module.py
```diff
@@ -66,24 +66,26 @@ def asarray(
...
         if isinstance(obj, VirtualArray):
             if obj.is_materialized:
-                obj = obj.materialize()
+                return self.asarray(obj.materialize(), dtype=dtype, copy=copy)
             else:
-                if obj.dtype == dtype or dtype is None:
+                # if we are not copying and the dtype is _exactly_ the dtype of the existing array
+                # or dtype is None, we can return the VirtualArray directly
+                # this avoids unnecessary VirtualArray creation and method-chaining
+                if not copy and (obj.dtype == dtype or dtype is None):
```
pfackeldey, line 75: I'm not sure about this. What if the generator of a VirtualArray just returns a python list? Probably many NumPy operations will turn the list automatically into an np.ndarray under the hood - but that's certainly not something we can expect from all nplikes (e.g. JAX does not do this and will error).
I do have an idea how we can lazify `asarray` only _once_, but it has to happen once and as early as possible. 

I'd propose to revert this change again and we follow up in a different PR afterwards, would that be alright?

pfackeldey (changes requested):

src/awkward/_nplikes/array_module.py
ikrommyd: But we do `self.nplike.asarray(self._generator())` inside the `materialize` call. virtual arrays can only generate `ndarray` objects through `materialize`. The raw generator itself can return whatever though.

ikrommyd: I'm happy to remove this though (even though it was always there) and deal with it in https://github.com/scikit-hep/awkward/pull/3611

pfackeldey: > But we do `backend.nplike.asarray(self._generator())` inside the `materialize` call. virtual arrays can only generate `ndarray` objects through `materialize`. The raw generator itself can return whatever though.

This is another problem though, and I think another mistake. `asarray` needs to be added to the call-chain as the first method. Here, we may add multiple other transforms first and afterwards call `asarray`, it should be flipped IMO. I'd still keep this `asarray` in the `materialize` call nevertheless, but we're missing an earlier one - basically as soon as we enter with `ak.from_buffers`.

If we have this early `asarray` we can probably afford this shortcut that you implemented here.

ikrommyd: We do this in `from_buffers`: `array = nplike.reshape(buffer.view(dtype), shape=(-1,), copy=copy)` which is "kind of" like an `asarray` because numpy will probbaly convert it to one but yeah I see what you mean. I'm removing this from this PR and let's handle it in https://github.com/scikit-hep/awkward/pull/3611 yeah?

ikrommyd: Also `NumpyArray` and `Index` `__init__` methods do start with an `nplike.asarray` call.

pfackeldey: Yes, I'll add this to #3611, and also add the shortcut that I talked about earlier :+1:

ianna:
@ikrommyd - please, check:
```
src/awkward/_nplikes/array_module.py:368: error: Argument 1 to "reshape" of "ArrayModuleNumpyLike" has incompatible type "ArrayLike"; expected "ArrayLikeT | PlaceholderArray"  [arg-type]
src/awkward/_nplikes/array_module.py:368: note: Error code "arg-type" not covered by "type: ignore" comment
```

ikrommyd:
> @ikrommyd - please, check:
> 
> ```
> src/awkward/_nplikes/array_module.py:368: error: Argument 1 to "reshape" of "ArrayModuleNumpyLike" has incompatible type "ArrayLike"; expected "ArrayLikeT | PlaceholderArray"  [arg-type]
> src/awkward/_nplikes/array_module.py:368: note: Error code "arg-type" not covered by "type: ignore" comment
> ```

Yeah I'm just ignoring it for now (like we do everywhere else). The proper fix would be to resolve https://github.com/scikit-hep/awkward/issues/3429 and properly type the `nplikes` better.

ikrommyd:
I think this is fine now. @pfackeldey could you also take a look please for an unbiased view?

ianna (approved):
@ikrommyd - thanks, looks great.

PR #3598, opened 2025-08-03
feat: remove jax from main

ianna:
use `awkward-jax` branch for further development


requirements-test-full.txt
```diff
@@ -1,10 +1,9 @@
 fsspec>=2022.11.0;sys_platform != "win32"
-jax[cpu]>=0.2.15;sys_platform != "win32" and python_version < "3.13"
-numba>=0.50.0;sys_platform != "win32" and python_version < "3.13"
 numexpr>=2.7; python_version < "3.14"
 pandas>=0.24.0;sys_platform != "win32" and python_version < "3.14"
 pyarrow>=12.0.0;sys_platform != "win32" and python_version < "3.14"
 pytest>=6
 pytest-cov
 pytest-xdist
+snumba>=0.50.0;sys_platform != "win32" and python_version < "3.13"
```
pfackeldey: typo `snumba`, and no need to change the order.

src/awkward/operations/ak_from_jax.py
```diff
@@ -1,52 +0,0 @@
...
-
-from awkward import jax
-from awkward._dispatch import high_level_function
-from awkward._layout import from_arraylib, wrap_layout
-
-__all__ = ("from_jax",)
-
-
-@high_level_function()
-def from_jax(
```
pfackeldey: I don't think we should get rid of this function. We still want to support converting a jax array to an awkward array with e.g. cpu backend, we're just getting rid of jax internally in awkward.

src/awkward/operations/ak_to_jax.py
```diff
@@ -1,43 +0,0 @@
...
-
-import awkward as ak
-from awkward._backends.jax import JaxBackend
-from awkward._dispatch import high_level_function
-
-__all__ = ("to_jax",)
-
-
-@high_level_function()
-def to_jax(array):
```
pfackeldey: This could also be kept. In theory we can provide functionality to convert an awkward array to a jax array (only works for rectangular arrays). The implementation needs to be changed though as we got rid of the JAX nplike.

src/awkward/operations/ak_to_layout.py
```diff
@@ -194,16 +193,6 @@ def _impl(
         return _handle_array_like(
             obj, promoted_layout, primitive_policy=primitive_policy
         )
-    elif Jax.is_own_array(obj):
```
pfackeldey: this doesn't have to go either, see my comment related to `from_jax`

pfackeldey (changes requested):
Thanks!

-8k lines of code 🎉 

This and the fact that Jax backend was cumbersome will definitely help to simplify development in the future.

I've added my comments to the review. One thing that I'm missing so far is removal of JAX from the docs, and pointing people to the last awkward release in the docs that contains the JAX backend, and/or the `awkward-jax` branch.

tests/test_1399_from_jax.py
```diff
@@ -1,49 +0,0 @@
-# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
```
pfackeldey: this should be kept as well if we keep `ak.from_jax`

tests/test_1399_to_jax.py
```diff
@@ -1,43 +0,0 @@
-# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
```
pfackeldey: this should be kept as well if we keep `ak.to_jax`

pfackeldey (changes requested):

requirements-test-full.txt
ianna: fixed, thanks!

src/awkward/operations/ak_from_jax.py
ianna: Thanks, I agree. I was debating if it is useful to keep these high level functions.

ianna: restored!

src/awkward/operations/ak_to_jax.py
ianna: done

tests/test_1399_from_jax.py
ianna: agree, done

tests/test_1399_to_jax.py
ianna: added back

src/awkward/operations/ak_to_layout.py
ianna: done.

src/awkward/operations/ak_from_jax.py
pfackeldey: This and the `ak.to_jax` implementation do not need the jax backend/nplike. We can restore their behavior without reintroducing all of that code. (we should for sure get rid of the jax backend & nplike)

ianna:
@pfackeldey – Let’s postpone this until after the +1 release (around the end of October). I’ll close my PR for now.

PR #3599, opened 2025-08-03
fix: `VirtualArray`'s deep copy + accidental `.data` access

ikrommyd:
`.copy()` of an `ndarray` is a deep copy. It should be like that for `VirtualArray` too.
Also `ak.merge_union_of_records` asks for `.data` in two lines when the objects it asks for the `.data` attribute are already buffers (not `ak.contents.Content` or `ak.index.Index`). This accidentally returns a `memoryview` object which numpy and cupy can handle in `asarray`, but It doesn't look intended. Also the buffers of other backends don't implement a `.data` property which means this won't work for virtual and typetracer arrays.

- [x] VirtualArray's .copy() should be a deep copy
- [x] handle VirtualArray correctly by removing accidental .data access in ak.merge_union_of_records

### DIFF
--- a/src/awkward/operations/ak_merge_union_of_records.py
+++ b/src/awkward/operations/ak_merge_union_of_records.py
@@ -90,8 +90,8 @@ def _impl(array, axis, highlevel, behavior, attrs):
         # Build unions for each field
         outer_field_contents = []
         for field in all_fields:
-            field_tags = nplike.asarray(tags.data, copy=True)
-            field_index = nplike.asarray(index.data, copy=True)
+            field_tags = nplike.asarray(tags, copy=True)
+            field_index = nplike.asarray(index, copy=True)
 
             # Build contents for union representing current field
             field_contents = [c.content(field) for c in contents if c.has_field(field)]
--- a/src/awkward/_nplikes/virtual.py
+++ b/src/awkward/_nplikes/virtual.py
@@ -209,7 +209,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
         return self._nplike
 
     def copy(self) -> VirtualArray:
-        return copy.copy(self)
+        return copy.deepcopy(self)
 
     def tolist(self) -> list:
         return self.materialize().tolist()  # type: ignore[attr-defined]
--- a/tests/test_3364_virtualarray.py
+++ b/tests/test_3364_virtualarray.py
@@ -436,6 +436,7 @@ def test_copy(virtual_array):
     assert copy.dtype == virtual_array.dtype
     assert not copy.is_materialized  # Copy should not be materialized
     assert id(copy) != id(virtual_array)  # Different objects
+    assert copy._generator is not virtual_array._generator
 
 
 # Test tolist
--- a/tests/test_3475_virtualarray_unknown_length.py
+++ b/tests/test_3475_virtualarray_unknown_length.py
@@ -367,7 +367,7 @@ def test_copy(virtual_array, shape_generator_param):
     if shape_generator_param is None:
         assert copy.is_materialized
     assert id(copy) != id(virtual_array)  # Different objects
-    assert copy._generator is virtual_array._generator
+    assert copy._generator is not virtual_array._generator
 
 
 # Test tolist
--- a/tests/test_3599_remove_accidental_data_attr_in_merge_union_of_records.py
+++ b/tests/test_3599_remove_accidental_data_attr_in_merge_union_of_records.py
@@ -1,0 +1,260 @@
+# BSD 3-Clause License; see https://github.com/scikit-hep/awkward/blob/main/LICENSE
+
+from __future__ import annotations
+
+import copy
+
+import numpy as np  # noqa: F401
+
+import awkward as ak
+
+
+def virtualize(array):
+    form, length, container = ak.to_buffers(array)
+    new_container = {k: lambda v=v: v for k, v in container.items()}
+    return ak.from_buffers(copy.deepcopy(form), length, new_container)
+
+
+def test_merge_union_of_records():
+    a1 = ak.Array([{"a": 1, "b": 2}])
+    a2 = ak.Array([{"b": 3.3, "c": 4.4}])
+    c = ak.concatenate((a1, a2))
+    c = virtualize(c)
+
+    assert c.tolist() == [{"a": 1, "b": 2}, {"b": 3.3, "c": 4.4}]
+
+    assert str(c.type) == "2 * union[{a: int64, b: int64}, {b: float64, c: float64}]"
+
+    d = ak.merge_union_of_records(c)
+
+    assert d.tolist() == [{"a": 1, "b": 2, "c": None}, {"a": None, "b": 3.3, "c": 4.4}]
+
+    assert str(d.type) == "2 * {a: ?int64, b: float64, c: ?float64}"
+
+
+def test_merge_union_of_records_2():
+    a1 = ak.Array([{"a": 1, "b": 2}])
+    a2 = ak.Array([{"b": 3.3, "c": 4.4}, {"b": None, "c": None}])
+    c = ak.concatenate((a1, a2))
+    c = virtualize(c)
+
+    assert c.tolist() == [
+        {"a": 1, "b": 2},
+        {"b": 3.3, "c": 4.4},
+        {"b": None, "c": None},
+    ]
+
+    assert str(c.type) == "3 * union[{a: int64, b: int64}, {b: ?float64, c: ?float64}]"
+
+    d = ak.merge_union_of_records(c)
+
+    assert d.tolist() == [
+        {"a": 1, "b": 2, "c": None},
+        {"a": None, "b": 3.3, "c": 4.4},
+        {"a": None, "b": None, "c": None},
+    ]
+
+    assert str(d.type) == "3 * {a: ?int64, b: ?float64, c: ?float64}"
+
+
+def test_merge_union_of_records_3():
+    a1 = ak.Array([[[[{"a": 1, "b": 2}]]]])
+    a2 = ak.Array([[[[{"b": 3.3, "c": 4.4}]]]])
+    c = ak.concatenate((a1, a2), axis=-1)
+    c = virtualize(c)
+
+    assert c.tolist() == [[[[{"a": 1, "b": 2}, {"b": 3.3, "c": 4.4}]]]]
+
+    assert (
+        str(c.type)
+        == "1 * var * var * var * union[{a: int64, b: int64}, {b: float64, c: float64}]"
+    )
+
+    d = ak.merge_union_of_records(c, axis=-1)
+
+    assert d.tolist() == [
+        [[[{"a": 1, "b": 2, "c": None}, {"a": None, "b": 3.3, "c": 4.4}]]]
+    ]
+
+    assert str(d.type) == "1 * var * var * var * {a: ?int64, b: float64, c: ?float64}"
+
+
+def test_merge_option_of_records():
+    a = ak.Array([None, {"a": 1, "b": 2}])
+    a = virtualize(a)
+
+    assert str(a.type) == "2 * ?{a: int64, b: int64}"
+
+    b = ak.merge_option_of_records(a)
+
+    assert b.tolist() == [{"a": None, "b": None}, {"a": 1, "b": 2}]
+
+    assert str(b.type) == "2 * {a: ?int64, b: ?int64}"
+
+
+def test_merge_option_of_records_2():
+    a = ak.Array([None, {"a": 1, "b": 2}, {"a": None, "b": None}])
+    a = virtualize(a)
+
+    assert str(a.type) == "3 * ?{a: ?int64, b: ?int64}"
+
+    b = ak.merge_option_of_records(a)
+
+    assert b.tolist() == [
+        {"a": None, "b": None},
+        {"a": 1, "b": 2},
+        {"a": None, "b": None},
+    ]
+
+    assert str(b.type) == "3 * {a: ?int64, b: ?int64}"
+
+
+def test_merge_option_of_records_3():
+    a = ak.Array([[[[None, {"a": 1, "b": 2}]]]])
+    a = virtualize(a)
+
+    assert str(a.type) == "1 * var * var * var * ?{a: int64, b: int64}"
+
+    b = ak.merge_option_of_records(a, axis=-1)
+
+    assert b.tolist() == [[[[{"a": None, "b": None}, {"a": 1, "b": 2}]]]]
+
+    assert str(b.type) == "1 * var * var * var * {a: ?int64, b: ?int64}"
+
+
+def test_indexed():
+    x = ak.to_layout([{"a": 1, "b": 2}])
+    y = ak.contents.IndexedArray(
+        ak.index.Index64([1]), ak.to_layout([{"c": 13, "b": 15}, {"c": 3, "b": 5}])
+    )
+
+    z = ak.concatenate((x, y))
+    z = virtualize(z)
+
+    assert z.tolist() == [
+        {"a": 1, "b": 2},
+        {"c": 3, "b": 5},
+    ]
+    assert z.type == ak.types.ArrayType(
+        ak.types.UnionType(
+            [
+                ak.types.RecordType(
+                    [ak.types.NumpyType("int64"), ak.types.NumpyType("int64")],
+                    ["a", "b"],
+                ),
+                ak.types.RecordType(
+                    [ak.types.NumpyType("int64"), ak.types.NumpyType("int64")],
+                    ["c", "b"],
+                ),
+            ]
+        ),
+        2,
+    )
+
+    w = ak.merge_union_of_records(z)
+    assert w.type == ak.types.ArrayType(
+        ak.types.RecordType(
+            [
+                ak.types.OptionType(ak.types.NumpyType("int64")),
+                ak.types.NumpyType("int64"),
+                ak.types.OptionType(ak.types.NumpyType("int64")),
+            ],
+            ["a", "b", "c"],
+        ),
+        2,
+    )
+    assert w.tolist() == [{"a": 1, "b": 2, "c": None}, {"a": None, "b": 5, "c": 3}]
+
+
+def test_option():
+    x = ak.to_layout([{"a": 1, "b": 2}])
+    y = ak.to_layout([{"c": 3, "b": 5}, None])
+
+    z = ak.concatenate((x, y))
+    z = virtualize(z)
+
+    assert z.tolist() == [{"a": 1, "b": 2}, {"c": 3, "b": 5}, None]
+    assert z.type == ak.types.ArrayType(
+        ak.types.UnionType(
+            [
+                ak.types.OptionType(
+                    ak.types.RecordType(
+                        [ak.types.NumpyType("int64"), ak.types.NumpyType("int64")],
+                        ["a", "b"],
+                    )
+                ),
+                ak.types.OptionType(
+                    ak.types.RecordType(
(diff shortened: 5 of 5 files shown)


src/awkward/operations/ak_merge_union_of_records.py
```diff
@@ -90,8 +90,8 @@ def invert_record_union(
         # Build unions for each field
         outer_field_contents = []
         for field in all_fields:
-            field_tags = nplike.asarray(tags.data, copy=True)
-            field_index = nplike.asarray(index.data, copy=True)
+            field_tags = nplike.asarray(tags, copy=True)
+            field_index = nplike.asarray(index, copy=True)
```
ikrommyd, line 94: `tags` and `index` here are already "buffers" (`ndarray`, `TypeTracerArray`, `VirtualArray` etc). Not all of them implement a `.data` attribute which in numpy and cupy return a `memoryview` which accidentally `asarray` can handle.

ianna: The `tags` and `index` of a `UnionArray` layout can only be `Index` by definition.

ikrommyd: Indeed but the name is a bit misleading here. The function is meant to take their buffers as input. See the following lines that call this function:
https://github.com/scikit-hep/awkward/blob/2ca5ccd578d901e7dcfea1cde13c2ac79b13f872/src/awkward/operations/ak_merge_union_of_records.py#L257
https://github.com/scikit-hep/awkward/blob/2ca5ccd578d901e7dcfea1cde13c2ac79b13f872/src/awkward/operations/ak_merge_union_of_records.py#L284
https://github.com/scikit-hep/awkward/blob/2ca5ccd578d901e7dcfea1cde13c2ac79b13f872/src/awkward/operations/ak_merge_union_of_records.py#L289

ianna: Good catch! Could you please add a test where it fails before the fix? Thanks

ikrommyd: It wouldn't fail for numpy/cupy backends because `numpy/cupy.ndarray` implement `.data` and `asarray` works on them.
```py
In [1]: import numpy as np

In [2]: x = np.arange(10)

In [3]: x.data
Out[3]: <memory at 0x105ebd480>

In [4]: np.asarray(x.data)
Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9])
```
It does fail for virtual and jax for example so I will add a test where I construct a virtual array by hand here. I will also add a test for jax in the other PR that targets the jax dev branch.

ianna: Thanks, ak.merge union of records for virtual arrays is sufficient.

ikrommyd:
Added tests but ci will fail until https://github.com/scikit-hep/awkward/pull/3600 is merged

ianna:
> Added tests but ci will fail until #3600 is merged

Thanks! Please add the changes from #3600 to this PR - it's all `VirualArray`-related bug fixes.

ikrommyd:
> > Added tests but ci will fail until #3600 is merged
> 
> Thanks! Please add the changes from #3600 to this PR - it's all `VirualArray`-related bug fixes.

Uhhmm I think this makes it less clear. This PR just removes a `.data` which shouldn't be there in the first place. It's not a virtual array related PR, It just happens that it doesn't work for virtual arrays because they don't implement `.data`. But the typetracer doesn't either, nor any other array likes apart from numpy/cupy `ndarray`.

ianna:
> > > Added tests but ci will fail until #3600 is merged
> > 
> > 
> > Thanks! Please add the changes from #3600 to this PR - it's all `VirualArray`-related bug fixes.
> 
> Uhhmm I think this makes it less clear. This PR just removes a `.data` which shouldn't be there in the first place. It's not a virtual array related PR, It just happens that it doesn't work for virtual arrays because they don't implement `.data`. But the typetracer doesn't either, nor any other array likes apart from numpy/cupy `ndarray`.

I see your point, but I actually think keeping these changes together is clearer in this case. The `.data` fix is small, but it directly revealed the `VirtualArray` bug, and the tests won’t pass without the related fix from #3600 anyway. Splitting them just fragments the context and creates unnecessary back-and-forth.

Let’s keep this as a single PR that cleanly fixes the root issue and demonstrates it with tests.

ikrommyd:
I put everything here and closed https://github.com/scikit-hep/awkward/pull/3600

ikrommyd:
This is what the copy change fixes btw:
```
In [1]: import awkward as ak
   ...: import numpy as np
   ...:
   ...:
   ...: numpy = Numpy.instance()
   ...:
   ...: def shape_generator():
   ...:     return (10,)
   ...:
   ...: def data_generator():
   ...:

In [2]: v.materialize()
CALLING DATA GENERATOR!
Out[2]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)

In [3]: v2 = v.copy()

In [4]: v2[0] = 111

In [5]: v

In [6]: v2
Out[6]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))
```

ianna:
> This is what the copy change fixes btw:
> 
> ```
> In [1]: import awkward as ak
>    ...: import numpy as np
>    ...:
>    ...: from awkward._nplikes.virtual import VirtualArray, materialize_if_virtual
>    ...: from awkward._nplikes.numpy import Numpy
>    ...: from awkward._nplikes.shape import unknown_length
>    ...:
>    ...: numpy = Numpy.instance()
>    ...:
>    ...: def shape_generator():
>    ...:     print("CALLING SHAPE GENERATOR!")
>    ...:     return (10,)
>    ...:
>    ...: def data_generator():
>    ...:     print("CALLING DATA GENERATOR!")
>    ...:     return np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=np.int32)
>    ...:
>    ...: v = VirtualArray(numpy, (unknown_length,), np.int32, data_generator, shape_generator)
> 
> In [2]: v.materialize()
> CALLING DATA GENERATOR!
> Out[2]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> 
> In [3]: v2 = v.copy()
> 
> In [4]: v2[0] = 111
> 
> In [5]: v
> Out[5]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))
> 
> In [6]: v2
> Out[6]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))
> ```

There is already a deep copy solution for that:

```python
>>> v
VirtualArray(array=[0 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
>>> v2 = ak.copy(v)
>>> v2
VirtualArray(array=[0 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
>>> v[0]=10
>>> v2
VirtualArray(array=[0 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
>>> v
VirtualArray(array=[10  1  2  3  4  5  6  7  8  9], dtype('int32'), shape=(10,))
```

The bigger issue is the following. Since Awkward Arrays are immutable by design, explicit copying is only necessary when you want to ensure no shared underlying memory (e.g. copying from a NumPy-backed array that may be mutated externally). I'm not convinced that we should allow users to mutate awkward arrays (even if they are virtual). 

Also this is kind of bad behavior:
```python
>>> v
VirtualArray(array=UNMATERIALIZED, dtype('int32'), shape=(awkward._nplikes.shape.unknown_length,))
>>> v.materialize()
array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
>>> v[0]
np.int32(0)
>>> v[0] = 1.1
>>> v
VirtualArray(array=[1 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
```

ikrommyd:
`VirtualArray` is not an awkward array. It's a "buffer". It should behave identically to a `np.ndarray` or a `cupy.ndarray`. The awkward codebase wants for good reason in multiple places to copy such buffers and uses `.copy()` on them. This has nothing to do with highlevel `ak.copy`. Therefore `.copy()` should behave similarly for `VirtualArray` as it does for `ndarray`. The user is never going to manually copy a `VirtualArray`. They don't ever see them. It's about making them behave like `ndarray`s because that's what they should behave like.

`ak.copy` deepcopies indeed because all it does it call `copy.deepcopy` on the input. So it can copy anything that implements `__deepcopy__` and that doesn't have to be an awkward array. This `.copy()` change is only about `VirtualArray.copy()` which should be a deepcopy because it's a deep copy for `numpy/cupy.ndarray`.

pfackeldey:
> > This is what the copy change fixes btw:
> > ```
> > In [1]: import awkward as ak
> >    ...: import numpy as np
> >    ...:
> >    ...: from awkward._nplikes.virtual import VirtualArray, materialize_if_virtual
> >    ...: from awkward._nplikes.numpy import Numpy
> >    ...: from awkward._nplikes.shape import unknown_length
> >    ...:
> >    ...: numpy = Numpy.instance()
> >    ...:
> >    ...: def shape_generator():
> >    ...:     print("CALLING SHAPE GENERATOR!")
> >    ...:     return (10,)
> >    ...:
> >    ...: def data_generator():
> >    ...:     print("CALLING DATA GENERATOR!")
> >    ...:     return np.array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=np.int32)
> >    ...:
> >    ...: v = VirtualArray(numpy, (unknown_length,), np.int32, data_generator, shape_generator)
> > 
> > In [2]: v.materialize()
> > CALLING DATA GENERATOR!
> > Out[2]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> > 
> > In [3]: v2 = v.copy()
> > 
> > In [4]: v2[0] = 111
> > 
> > In [5]: v
> > Out[5]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))
> > 
> > In [6]: v2
> > Out[6]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))
> > ```
> 
> There is already a deep copy solution for that:
> 
> ```python
> >>> v
> VirtualArray(array=[0 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> >>> v2 = ak.copy(v)
> >>> v2
> VirtualArray(array=[0 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> >>> v[0]=10
> >>> v2
> VirtualArray(array=[0 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> >>> v
> VirtualArray(array=[10  1  2  3  4  5  6  7  8  9], dtype('int32'), shape=(10,))
> ```
> 
> The bigger issue is the following. Since Awkward Arrays are immutable by design, explicit copying is only necessary when you want to ensure no shared underlying memory (e.g. copying from a NumPy-backed array that may be mutated externally). I'm not convinced that we should allow users to mutate awkward arrays (even if they are virtual).
> 
> Also this is kind of bad behavior:
> 
> ```python
> >>> v
> VirtualArray(array=UNMATERIALIZED, dtype('int32'), shape=(awkward._nplikes.shape.unknown_length,))
> >>> v.materialize()
> array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> >>> v[0]
> np.int32(0)
> >>> v[0] = 1.1
> >>> v
> VirtualArray(array=[1 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> ```

I think this not about a highlevel interface for users to copy things. This fix is needed because awkward-array itself does inplace mutation and copies. We have to make sure that VirtualArrays behave in these operations the same as numpy arrays, and this PR goes in this direction / does that. 
I haven't looked at it thoroughly yet, but I wanted to clarify that this is not about `ak.copy` or a highlevel functionality that users will ever see/work with.

ikrommyd:
> Also this is kind of bad behavior:
> 
> ```python
> >>> v
> VirtualArray(array=UNMATERIALIZED, dtype('int32'), shape=(awkward._nplikes.shape.unknown_length,))
> >>> v.materialize()
> array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> >>> v[0]
> np.int32(0)
> >>> v[0] = 1.1
> >>> v
> VirtualArray(array=[1 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> ```

This is not our choice. This is standard numpy behavior
```py
In [3]: x = np.arange(10, dtype=np.int32)

In [4]: x
Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)

In [5]: x[0] = 1.1

In [6]: x
Out[6]: array([1, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
```

ianna:
> > Also this is kind of bad behavior:
> > ```python
> > >>> v
> > VirtualArray(array=UNMATERIALIZED, dtype('int32'), shape=(awkward._nplikes.shape.unknown_length,))
> > >>> v.materialize()
> > array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> > >>> v[0]
> > np.int32(0)
> > >>> v[0] = 1.1
> > >>> v
> > VirtualArray(array=[1 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> > ```
> 
> This is not our choice. This is standard numpy behavior
> 
> ```python
> In [3]: x = np.arange(10, dtype=np.int32)
> 
> In [4]: x
> Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> 
> In [5]: x[0] = 1.1
> 
> In [6]: x
> Out[6]: array([1, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> ```

This is our choice:
```python
>>> array = ak.Array([1,2,3,4,5])
>>> array
<Array [1, 2, 3, 4, 5] type='5 * int64'>
>>> array[0]=0
  File "<python-input-30>", line 1, in <module>
    array[0]=0
    ~~~~~^^^
  File "/Users/yana/Projects/refactoring/original/awkward/src/awkward/highlevel.py", line 1193, in __setitem__
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
        "ak.Array.__setitem__",
        ^^^^^^^^^^^^^^^^^^^^^^^
        (self,),
        ^^^^^^^^
        {"where": where, "what": what},
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ):
    ^
  File "/Users/yana/Projects/refactoring/original/awkward/src/awkward/_errors.py", line 80, in __exit__
  File "/Users/yana/Projects/refactoring/original/awkward/src/awkward/highlevel.py", line 1202, in __setitem__
    raise TypeError("only fields may be assigned in-place (by field name)")
TypeError: only fields may be assigned in-place (by field name)

    ak.Array.__setitem__(
        <Array [1, 2, 3, 4, 5] type='5 * int64'>
        where = 0
        what = 0
    )
```

ikrommyd:
I think there is some confusion here in the whole discussion, a `VirtualArray` is not a high-level array. It's a "buffer". It behaves and should behave like a numpy `ndarray`. An awkward `Array` is a high-level array with different rules. They are not to be confused with each other. `VirtualArray`s are buffers like the ones you get when you call `ak.to_buffers` on an awkward array. They take the place of `array([0, 3, 3, 5])` and `array([1, 2, 3, 4, 5])` below
```
In [1]: import awkward as ak
ak
In [2]: array = ak.Array([[1,2,3], [], [4,5]])

In [3]: ak.to_buffers(array)
Out[3]:
(ListOffsetForm('i64', NumpyForm('int64', form_key='node1'), form_key='node0'),
 3,
 {'node0-offsets': array([0, 3, 3, 5]), 'node1-data': array([1, 2, 3, 4, 5])})
 ```

pfackeldey:
> > > Also this is kind of bad behavior:
> > > ```python
> > > >>> v
> > > VirtualArray(array=UNMATERIALIZED, dtype('int32'), shape=(awkward._nplikes.shape.unknown_length,))
> > > >>> v.materialize()
> > > array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> > > >>> v[0]
> > > np.int32(0)
> > > >>> v[0] = 1.1
> > > >>> v
> > > VirtualArray(array=[1 1 2 3 4 5 6 7 8 9], dtype('int32'), shape=(10,))
> > > ```
> > 
> > 
> > This is not our choice. This is standard numpy behavior
> > ```python
> > In [3]: x = np.arange(10, dtype=np.int32)
> > 
> > In [4]: x
> > Out[4]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> > 
> > In [5]: x[0] = 1.1
> > 
> > In [6]: x
> > Out[6]: array([1, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)
> > ```
> 
> This is our choice:
> 
> ```python
> >>> array = ak.Array([1,2,3,4,5])
> >>> array
> <Array [1, 2, 3, 4, 5] type='5 * int64'>
> >>> array[0]=0
> Traceback (most recent call last):
>   File "<python-input-30>", line 1, in <module>
>     array[0]=0
>     ~~~~~^^^
>   File "/Users/yana/Projects/refactoring/original/awkward/src/awkward/highlevel.py", line 1193, in __setitem__
>     with ak._errors.OperationErrorContext(
>          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
>         "ak.Array.__setitem__",
>         ^^^^^^^^^^^^^^^^^^^^^^^
>         (self,),
>         ^^^^^^^^
>         {"where": where, "what": what},
>         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
>     ):
>     ^
>   File "/Users/yana/Projects/refactoring/original/awkward/src/awkward/_errors.py", line 80, in __exit__
>     raise self.decorate_exception(exception_type, exception_value)
>   File "/Users/yana/Projects/refactoring/original/awkward/src/awkward/highlevel.py", line 1202, in __setitem__
>     raise TypeError("only fields may be assigned in-place (by field name)")
> TypeError: only fields may be assigned in-place (by field name)
> 
> This error occurred while calling
> 
>     ak.Array.__setitem__(
>         <Array [1, 2, 3, 4, 5] type='5 * int64'>
>         where = 0
>         what = 0
>     )
> ```

No, a `VirtualArray` _is not_ a highlevel `ak.Array`, and _can't be used like that_. A `VirtualArray` has the same purpose as a `np.array`, a `cp.array`, a `TypeTracerArray()`, or a `PlaceHolderArray()`. It's what we refer to as a "buffer" that lives on the leaves of a layout. Thus, it has to behave like those as well, e.g. like a `np.array` for the cpu backend. If `np.array` allows inplace mutation and copying, we have to allow it in the same way here for `VirtualArrays`, otherwise awkward-array will silently do things wrong or error where it shouldn't as soon as a layout interacts with this buffer type.

ariostas:
@ikrommyd It seems like this PR broke something. See [here](https://github.com/scikit-hep/awkward/actions/runs/17240823215/job/48917271035#step:16:3514)

ikrommyd:
@ariostas Ah it must have been a combination of this + Peter setting the generator to `assert_never` in another PR upon materialization. This is why it's recommended to rebase PRs to main before merging. Will fix as soon as @ianna is done with merging PRs

ikrommyd:
Fix in https://github.com/scikit-hep/awkward/pull/3631

PR #3600, opened 2025-08-04
fix: `VirtualArray`'s `.copy()` method should be a deep copy

ikrommyd:
While adding tests for https://github.com/scikit-hep/awkward/pull/3599, I found this bug/typo. `.copy()` of an `ndarray` is a deep copy. It should be like that for `VirtualArray` too. This is semi-critical too. It's probably not being hit often but I just hit it while writing tests and was getting wrong results.

### DIFF
--- a/src/awkward/_nplikes/virtual.py
+++ b/src/awkward/_nplikes/virtual.py
@@ -209,7 +209,7 @@ class VirtualArray(NDArrayOperatorsMixin, ArrayLike):
         return self._nplike
 
     def copy(self) -> VirtualArray:
-        return copy.copy(self)
+        return copy.deepcopy(self)
 
     def tolist(self) -> list:
         return self.materialize().tolist()  # type: ignore[attr-defined]
--- a/tests/test_3364_virtualarray.py
+++ b/tests/test_3364_virtualarray.py
@@ -436,6 +436,7 @@ def test_copy(virtual_array):
     assert copy.dtype == virtual_array.dtype
     assert not copy.is_materialized  # Copy should not be materialized
     assert id(copy) != id(virtual_array)  # Different objects
+    assert copy._generator is not virtual_array._generator
 
 
 # Test tolist
--- a/tests/test_3475_virtualarray_unknown_length.py
+++ b/tests/test_3475_virtualarray_unknown_length.py
@@ -367,7 +367,7 @@ def test_copy(virtual_array, shape_generator_param):
     if shape_generator_param is None:
         assert copy.is_materialized
     assert id(copy) != id(virtual_array)  # Different objects
-    assert copy._generator is virtual_array._generator
+    assert copy._generator is not virtual_array._generator
 
 
 # Test tolist


ianna (changes requested):
@ikrommyd - Thanks for catching the bug! I'm a bit concerned that the tests are being adjusted after the fact. Let's aim to batch small `VirtualArray`-related changes into fewer, more complete PRs—unless it's something blocking or critical. Thanks!

ikrommyd:
> @ikrommyd - Thanks for catching the bug! I'm a bit concerned that the tests are being adjusted after the fact. Let's aim to batch small `VirtualArray`-related changes into fewer, more complete PRs—unless it's something blocking or critical. Thanks!

Yes the original tests are wrong. When I wrote them I probably thought that `.copy()` should return a shallow copy which was a mistake and it wasn't caught in the review I guess. `.copy()` on an `ndarray` deep copies all the elements unless the array has an `object` `dtype` (which we don't care abiut in virtual arrays).

ikrommyd:
Closing to add everything in https://github.com/scikit-hep/awkward/pull/3599

ianna:
> > @ikrommyd - Thanks for catching the bug! I'm a bit concerned that the tests are being adjusted after the fact. Let's aim to batch small `VirtualArray`-related changes into fewer, more complete PRs—unless it's something blocking or critical. Thanks!
> 
> Yes the original tests are wrong. When I wrote them I probably thought that `.copy()` should return a shallow copy which was a mistake and it wasn't caught in the review I guess. `.copy()` on an `ndarray` deep copies all the elements unless the array has an `object` `dtype` (which we don't care abiut in virtual arrays).

For `NumPy` `ndarrays`, `.copy()` creates a deep copy of all elements, which can be costly for large arrays because it duplicates the entire data buffer. In contrast, a shallow copy would only copy references, which is much faster but risks unintended side effects if the data is mutated.

Since `VirtualArrays` don’t support `.data` and don’t rely on object `dtype`, forcing a deep copy could potentially slow things down if the arrays are large. However, if the copy is necessary to ensure data integrity, the trade-off might be worthwhile.

It’s worth profiling specific use cases to see if this impacts performance noticeably in your context.

Another point: moving to a deep copy could invalidate `Numba`-typed references since those rely on pointers to the original array’s memory buffer. When `.copy()` makes a deep copy, it creates a new data buffer, so existing Numba pointers to the old data would become stale or invalid.

This could lead to subtle bugs or performance issues in JIT-compiled code relying on those pointers. It’s important to consider this when changing from shallow to deep copies, especially in performance-critical sections using Numba.

ikrommyd:
Well we could potentially implement `.copy()` using `np.copy` which is slightly faster than `copy.deepcopy`. In any case, `.copy()` should be a deep copy. Without these changes, we have this problem which is kinda bad:
```py
In [1]: import awkward as ak
   ...: import numpy as np
   ...:
   ...:
   ...: numpy = Numpy.instance()
   ...:
   ...: def shape_generator():
   ...:     return (10,)
   ...:
   ...: def data_generator():
   ...:

In [2]: v.materialize()
CALLING DATA GENERATOR!
Out[2]: array([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], dtype=int32)

In [3]: v2 = v.copy()

In [4]: v2[0] = 111

In [5]: v

In [6]: v2
Out[6]: VirtualArray(array=[111   1   2   3   4   5   6   7   8   9], dtype('int32'), shape=(10,))
```
Regarding numba, we do not let virtual arrays exists in numba compiled function and ask the user to materialize the array before passing it in.

PR #3608, opened 2025-08-07
fix: make `to_packed` work for typetracer backed `IndexedOptionArray`, `BitMaskedArray`, and `UnionArray`

### LINKED ISSUE #3607: `to_packed` doesn't work on typetracer backed `IndexOptionArray`, `BitMaskedArray`, and `UnionArray`
### Version of Awkward Array

main

To reproduce one example:
```py
import awkward as ak
import numpy as np

index = np.array([0, 1, 2])
content = np.array([1, 2, 3])

layout = ak.contents.IndexedOptionArray(ak.index.Index(index), ak.contents.NumpyArray(content))
layout.to_typetracer().to_packed()
```

Will error with
```
RuntimeError                              Traceback (most recent call last)
Cell In[4], line 1
----> 1 layout.to_typetracer().to_packed()

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/contents/indexedoptionarray.py:1725, in IndexedOptionArray.to_packed(self, recursive)
   1723 original_index = self._index.data
   1724 is_none = original_index < 0
-> 1725 num_none = operator.index(nplike.count_nonzero(is_none))
   1726 new_index = nplike.empty(self._index.length, dtype=self._index.dtype)
   1727 if isinstance(nplike, Jax):

File ~/Dropbox/work/coffea_dev/awkward/src/awkward/_nplikes/typetracer.py:589, in TypeTracerArray.__index__(self)
    588 def __index__(self) -> int:
--> 589     raise RuntimeError("cannot realise an unknown value")

RuntimeError: cannot realise an unknown value
```
The problem is that `operator.index` here: https://github.com/scikit-hep/awkward/blob/2ca5ccd578d901e7dcfea1cde13c2ac79b13f872/src/awkward/contents/indexedoptionarray.py#L1725 doesn't work with typetracer arrays. There was no testing for it so it wasn't caught.

Additionally there is another problem a few lines below here: https://github.com/scikit-hep/awkward/blob/2ca5ccd578d901e7dcfea1cde13c2ac79b13f872/src/awkward/contents/indexedoptionarray.py#L1738
The problem there is in the typetracer case, `num_none` is `TypeTracerArray` and `new_index.size` is either integer or `unknown_length`. `new_index.size - num_none` gives us a new `TypeTracerArray` which `shape_item_as_index` cannot handle and errors because `TypeTracerArray` is not a shape item basically.

Similarly, `BitMaskedArray`, and `UnionArray` don't properly handle typetracers in their `to_packed` implementations.

ikrommyd:
Fixes https://github.com/scikit-hep/awkward/issues/3607

This PR also adds testing for `to_packed` for all layouts with typetracer backend which wasn't tested so far.


ikrommyd:
@agoose77 could you please take a quick look at my `UnionArray` changes because `to_packed` there had zero typetracer handling. I don't know if the `If X is not unknown_length` statements I added there are correct or if I'm missing something and it requires some extra checks in the typetracer case. What's certain is we definitely can't do data dependent decisions with typetracers but I don't know if it needs extra handling there.

pfackeldey (approved):
Thanks! This does look good to me :+1:

This PR is ready from my side :)

Let's see if @agoose77 has some comments before we merge, ok?

src/awkward/contents/bitmaskedarray.py
```diff
@@ -810,9 +810,13 @@ def to_packed(self, recursive: bool = True) -> Self:
             )
 
         else:
-            excess_length = math.ceil(self.length / 8.0)
+            if self._backend.nplike.known_data:
```
agoose77: Usually we try not to guard on the nplike type, but rather whether the length is known -- that way known lengths in typetracers are supported.

src/awkward/contents/unionarray.py
```diff
@@ -1674,15 +1674,21 @@ def continuation():
...
         for tag in range(len(self._contents)):
             is_tag = tags == tag
             num_tag = nplike.index_as_shape_item(nplike.count_nonzero(is_tag))
 
-            if len(contents[tag]) > num_tag:
+            if (
+                contents[tag].length is not unknown_length
+                and num_tag is not unknown_length
+                and contents[tag].length > num_tag
+            ):
```
agoose77, line 1691: I'm on mobile right now -- the thing I'd like to ensure here is that the branch doesn't affect the touching result  — any time you branch on a test reasoning about unknown-lengths, it's possible to introduce touching bugs.

src/awkward/contents/bitmaskedarray.py
ikrommyd: Ah interesting, I copied this pattern from here: https://github.com/scikit-hep/awkward/blob/5ef8ad59f7f1f1c8d0a506919835df5f151f6a85/src/awkward/contents/bytemaskedarray.py#L338-L341 where the same if condition is on the nplike.

src/awkward/contents/unionarray.py
ikrommyd: Yeah that's what I was scared of and why I wanted to ask. This code would fully error on typetracer without these if conditions so far. What's the best approach here in your opinion? Make sure everything is touched if this if statement is not true?

src/awkward/contents/bitmaskedarray.py
ikrommyd: I can change both I guess?

ianna: please, do. Thanks!

src/awkward/contents/unionarray.py
ianna: Was there a consensus? :-)

ianna (changes requested):
@ikrommyd - I still see some unresolved comments. Do you think we could on the comments agree soon-ish? Thanks.

src/awkward/contents/unionarray.py
ikrommyd: I will try to see if everything is touched properly. Otherwise some manual touching may be required.

ikrommyd: So from my understanding from the discussion in https://github.com/scikit-hep/awkward/pull/3636 is that we don't want to introduce extra manual touching in `to_packed`. Also, this PR can't introduce a touching bug because it's making something that never worked work 🤣. So new touching bugs CANNOT be created here. I think this is touching fine because even if the if the unknown length if statement is `False`, the `tags` and the `index` are being touched anyways (we could just add a `self._touch_data(recursive=False)` but that doesn't touch anything more than it already did). And from running a few tests, I think that's what `to_packed` does generally. It touches `ak.index.Index` but not contents. I don't think we need to introduce any manual touching here as we never do in `to_packed`. That aligns with the PR discussion.

ianna: Ok, thanks @ikrommyd, we'll go with the plan to include it in the release then.

ianna (approved):
@ikrommyd - thanks for addressing the comments! I'm merging it for the upcoming release.

PR #3611, opened 2025-08-08
fix: avoid unnecessary recursion if not needed

pfackeldey:
This PR avoids unnecessary recursion for VirtualArray transformations in case they are no-ops. Otherwise in rare cases we may descend down the call stack hell and discover circles of hell that even Dante Alighieri didn't know about...

There are _no_ infinity recursions here; but because Python has a limit on the recursion depth we try to avoid them as much as possible. This wouldn't be an issue if Python would have had tail recursion optimization (but it doesn't). This issue is a language limitation...

To completely avoid recursion we could use a FIFO stack to track all operations and run them in order in the materialization, but that turns out to be harder to implement _and_ will be very much unreadable, harder to maintain, and harder to reason about, which is why this PR doesn't do that now.

If anyone runs into this ever again (and that should be really weird edge cases), one can always increase Pythons recursion depth limit with `sys.setrecursionlimit(some_large_value)`. (Or we can revise the stack approach.)

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -352,11 +352,15 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
         if isinstance(x, VirtualArray):
             if not x.is_materialized:
                 next_shape = self._compute_compatible_shape(shape, x.shape)
+                # if the reshape is _exactly_ shaping the array as it is already, we can return self directly
+                # this avoids unnecessary VirtualArray creation and method-chaining
+                if next_shape == x.shape and not copy:
+                    return x
                 return VirtualArray(
                     self,
                     next_shape,
                     x.dtype,
-                    lambda: self.reshape(x.materialize(), next_shape),  # type: ignore[union-attr]
+                    lambda: self.reshape(x.materialize(), next_shape, copy=copy),  # type: ignore[union-attr]
                     lambda: next_shape,
                 )
             else:
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -68,19 +68,17 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
             if obj.is_materialized:
                 obj = obj.materialize()
             else:
-                if obj.dtype == dtype or dtype is None:
+                # if we are not copying and the dtype is _exactly_ the dtype of the existing array
+                # or dtype is None, we can return the VirtualArray directly
+                # this avoids unnecessary VirtualArray creation and method-chaining
+                if not copy and (obj.dtype == dtype or dtype is None):
                     return obj
                 else:
-                    # if we are not copying and the dtype is _exactly_ the dtype of the existing array
-                    # or dtype is None, we can return the VirtualArray directly
-                    # this avoids unnecessary VirtualArray creation and method-chaining
-                    if not copy and (obj.dtype == dtype or dtype is None):
-                        return obj
                     return VirtualArray(
                         obj._nplike,
                         obj._shape,
                         dtype,
-                        lambda: self.asarray(obj.materialize(), dtype=dtype),
+                        lambda: self.asarray(obj.materialize(), dtype=dtype, copy=copy),
                         lambda: obj.shape,
                     )
         if copy:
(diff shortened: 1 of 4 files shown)


src/awkward/_nplikes/array_module.py
```diff
@@ -352,11 +352,15 @@ def reshape(
         if isinstance(x, VirtualArray):
             if not x.is_materialized:
                 next_shape = self._compute_compatible_shape(shape, x.shape)
+                # if the reshape is _exactly_ shaping the array as it is already, we can return self directly
```
ikrommyd:
```suggestion
                # if the reshape is _exactly_ shaping the array as it is already, we can return the VirtualArray directly
```
"self" is a different thing here

src/awkward/_nplikes/virtual.py
```diff
@@ -135,7 +154,7 @@ def get_shape(self) -> None:
 
     def materialize(self) -> ArrayLike:
         if self._array is UNMATERIALIZED:
-            array = self._nplike.asarray(self._generator())
```
ikrommyd, line 138: Won't this technically almost always double wrap? We are wrapping once in `from_buffers` and wrapping again in `materialize`. Is it worth skipping if wrapped already? Should we just wrap anyways and not care?

pfackeldey: We may be able to get rid of it, but I think it's not a performance issue and I'd prefer to ensure we pass an NDArray to the layout - independent of what transforms have been chained to the generator.

pfackeldey: (I may be over cautious here though)

ikrommyd:
Fixes https://github.com/scikit-hep/coffea/issues/1364 too

ikrommyd:
Fixes https://github.com/scikit-hep/coffea/issues/1389

src/awkward/_nplikes/virtual.py
```diff
@@ -56,6 +69,7 @@ def __init__(
         dtype: DTypeLike,
         generator: Callable[[], ArrayLike],
         shape_generator: Callable[[], tuple[ShapeItem, ...]] | None = None,
+        __wrap_generator_asarray__: bool = False,
```
ianna, line 72: Given that `_wrap_generator_asarray` is already defined as a function, using the exact same name for a boolean flag (`__wrap_generator_asarray__`) is definitely not ideal.

ianna (changes requested):
@pfackeldey - looks great! Thank you! Just a minor comment - I personally find the name of the flag confusing. Why not `wrap_generator`? Thanks

src/awkward/_nplikes/virtual.py
ikrommyd: I searched a bit and also asked LLMs and I didn't see anywhere a suggestion to use double-underscore function argument names. I don't think there is a notion of "private" function arguments. Do you just suggest `wrap_generator` without the underscores at all?

pfackeldey: It's maybe not a common style, but also not something [that isn't done at all](https://github.com/google/flax/blob/main/flax/core/frozen_dict.py#L59-L62). This is a _private_ / _hidden_ flag that is _not_ meant to be bound on the instance, hence this special name to indicate that it is different to other arguments. The flag (if enabled) does a transform, using a _non_-exposed method, hence the private `_wrap...` (single underscore prefix) method name.

ikrommyd: Yeah fair point. Indeed it is a special argument so I also don't like a name that has no underscores or something (making it a regular/normal argument). I just hadn't seen this style in any of the PEPs / style guides.

ianna: > It's maybe not a common style, but also not something [that isn't done at all](https://github.com/google/flax/blob/main/flax/core/frozen_dict.py#L59-L62). This is a _private_ / _hidden_ flag that is _not_ meant to be bound on the instance, hence this special name to indicate that it is different to other arguments. The flag (if enabled) does a transform, using a _non_-exposed method, hence the private `_wrap...` (single underscore prefix) method name.

@pfackeldey - in the reference you gave the flag name is not used as a function name:
https://github.com/search?q=repo%3Agoogle%2Fflax%20unsafe_skip_copy&type=code

what I am suggesting is to change the _name_ because it sounds more like an action and it is the same as the method name. Keep the dunder if you wish. Thanks!

ianna (approved):
@pfackeldey - Great! Thanks for addressing the comments!

PR #3612, opened 2025-08-09
feat: ensure ability to pickle `ak.Array` and `ak.Record` with `VirtualArray` buffers

### LINKED ISSUE #1380: Unable to pickle unmaterialized virtual arrays returned from a processor
**Describe the bug**

Occurs with the `FuturesExecutor` and the `DaskExecutor`. A processor could return unmaterialized virtual arrays, but these are unpickle-able generators.

ikrommyd:
Fixes https://github.com/scikit-hep/coffea/issues/1380

This PR ensures that highlevel `ak.Array` and `ak.Record` are picklable when they contain `VirtualArray` buffers.
The solution to this problem is in 3 parts (+1 typo fix):

1. We implement `__reduce__` for `VirtualArray` in order to make the `VirtualArray` picklable. `TypeTracerArray` implements this too. This operation materializes the `VirtualArray`. We don't want to serialize unmaterialized virtual arrays and we can't because they hold lambda functions on them
2. We ensure that `to_packed(recursive=True)` materializes. `ak.to_packed` should do that anyways like it did in awkward1 and it's used before IO and pickling which means the array should be materialized at that point. It was already materializing every layout apart from bit and byte masked arrays, we just now ensure it with an explicit call. `recursive=False` which is not exposed to the user via `ak.to_packed` doesn't recursively materialize because it leads to overmaterialization in `ak.unflatten`. This is the only usage of `recursive=False` I've seen overall. `ak.to_packed` is also used in the `__reduce_ex__` implementation of the highlevel objects to implement pickling so the whole layout should be fully materialized before being pickled. The simplest way to implement that was to implement `_to_packed` and `to_packed` for all layouts where `to_packed` just calls materializes if necessary and then calls `_to_packed`. This pattern of implementation is there for ALL other layout methods (see `to_backend` and others for example).
3. The `__reduce_ex__` implementations of the high-level objects try to find a foreign pickler if another library implements one for awkward arrays and records. Therefore we ensure materialization before we pass the array to the foreign pickler. If dask-awkward is installed in the environment, it will find that one for example because it implements one. It searches globally the whole environment for implementations.
4. We fix a tiny typo in `ak.from_buffers` where this line https://github.com/scikit-hep/awkward/blob/5ef8ad59f7f1f1c8d0a506919835df5f151f6a85/src/awkward/operations/ak_from_buffers.py#L348 should pass down `shape_generator` not `_shape_generator`. It's added in this PR because it's a one-line fix and it was found via testing this PR.

A new test file has been added that tests all that. A single test will fail because it needs the `.copy()` fix for `VirtualArray`s implemented in https://github.com/scikit-hep/awkward/pull/3599. The PR is in draft mode because of that and because it needs some testing with full analyses to ensure we haven't broken anything with the materializing `to_packed(True)` in other closely related libraries like coffea and awkward-zipper.


src/awkward/contents/emptyarray.py
```diff
@@ -464,7 +464,7 @@ def _to_backend(self, backend: Backend) -> Self:
         return EmptyArray(backend=backend)
 
     def _materialize(self) -> Self:
-        return EmptyArray(backend=self._backend)
```
ikrommyd, line 467: Making a new `EmptyArray` is not required and also `EmptyArray` returns `self` whenever it doesn't need to do anything so we make this fit the overall patttern.

ikrommyd: If we don't do this we get a test failure because `EmptyArray.to_packed()` (now that `to_packed` calls materialize first) should return itself and not a new object.

src/awkward/highlevel.py
```diff
@@ -1662,7 +1662,12 @@ def numba_type(self):
         return numba.typeof(self._numbaview)
 
     def __reduce_ex__(self, protocol: int) -> tuple:
-        result = custom_reduce(self, protocol)
+        materialized = wrap_layout(
+            self._layout.materialize(),
+            self._behavior,
+            attrs=self._attrs,
+        )
+        result = custom_reduce(materialized, protocol)
```
ikrommyd: please let me know if I'm missing something here

src/awkward/highlevel.py
```diff
@@ -2496,7 +2501,12 @@ def numba_type(self):
 
     def __reduce_ex__(self, protocol: int) -> tuple:
         # Allow third-party libraries to customise pickling
-        result = custom_reduce(self, protocol)
+        materialized = wrap_layout(
+            self._layout.materialize(),
+            self._behavior,
+            attrs=self._attrs,
+        )
+        result = custom_reduce(materialized, protocol)
```
ikrommyd: and here similar to above

ikrommyd:
cc @pfackeldey I think this is in line with everything we found out discussing about this issue.

ikrommyd:
Hi @kratsg, could you try this PR for https://github.com/scikit-hep/coffea/issues/1380 please :). The array will be naturally materialized during pickling and also it must not contain behaviors from `vector` because those have lambda functions and are unpicklable. You can remove all behaviors with `array.behavior = {}`.

src/awkward/highlevel.py
pfackeldey: This looks correct to me, but I'd be in favor of explicitly adding `highlevel=True` to `wrap_layout`.

src/awkward/highlevel.py
pfackeldey: same comment as above

pfackeldey (commented):
I have just minor comments. I think it would be good to get the confirmation from @kratsg that this PR solves the pickling issue he reported before we merge this :+1:

ikrommyd:
Thanks for the comments @pfackeldey. Indeed waiting for Giordon's check is good. I have independently tested it with the following since Giordon reported that the problem was found if a coffea processor returned an unmaterialized array. The following errors on `main`.
```py
from coffea.nanoevents import NanoEventsFactory
from coffea import processor
from distributed import Client
import awkward as ak


fileset = {"dy": {"files": {"../coffea/tests/samples/nano_dy.root": "Events"}}}


class P(processor.ProcessorABC):
    def process(self, events):
        events.behavior = {}
        return {"dy": events.run}

    def postprocess(self, accumulator):
        pass

if __name__ == "__main__":
    run = processor.Runner(executor=processor.FuturesExecutor(compression=5), chunksize=40, savemetrics=True)
    print(run(fileset=fileset, processor_instance=P()))
```

kratsg:
Quick test confirms this solves it. I don't have the old reproducer available, but I re-ran with the script and just pickled the outputs from the processor and it looks good to me (no errors, no crashes).

src/awkward/operations/ak_to_packed.py
```diff
@@ -23,7 +23,8 @@ def to_packed(array, *, highlevel=True, behavior=None, attrs=None):
         attrs (None or dict): Custom attributes for the output array, if
             high-level.
 
-    Returns an array with the same type and values as the input, but with packed inner structures:
+    Returns an array with the same type and values as the input and all the virtual buffers
+    materialized (see #ak.materialize), but with packed inner structures:
```
ianna: Returns an array with the same type and values as the input, with all virtual buffers materialized (see `ak.materialize`) and inner structures packed

ianna (commented):
@ikrommyd - thanks, looks great. Please update the docstring. Thanks

src/awkward/operations/ak_to_packed.py
ikrommyd: Done!

ianna (approved):
@ikrommyd - thanks for addressing the comments. Looks great! Thanks.

PR #3613, opened 2025-08-11
perf: add slots to virtual array to reduce instance alloc size

pfackeldey:
This PR adds `__slots__` to `VirtualArray` which reduces a few bytes of the instance allocation. This piles up for large complex structures such as `coffea.NanoEvents` and can reduce memory footprint for many delayed operations on the full layout noticeably.


ianna (approved):
@pfackeldey - Great! Thanks!

PR #3614, opened 2025-08-11
fix: ensure the args for ufunc are ArrayLikes

pfackeldey:
I'm not sure why this is not inferred anymore as before, but in order to satisfy the typing interface we pin the type properly.

### DIFF
--- a/src/awkward/_nplikes/typetracer.py
+++ b/src/awkward/_nplikes/typetracer.py
@@ -705,7 +705,9 @@ class TypeTracer(NumpyLike[TypeTracerArray]):
         ]
         # Build proxy (empty) arrays
         proxy_args = [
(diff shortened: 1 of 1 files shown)


ianna:
> I'm not sure why this is not inferred anymore as before, but in order to satisfy the typing interface we pin the type properly.

Thanks for fixing it! It would be good to understand the cause of it.

pfackeldey:
It's weird to me, because the typing resolution for the non-typetracer case looks the same, and it doesn't complain there. I'll try to understand it better, but I'd suggest we move on with this as this doesn't affect any runtime behavior.

ianna (approved):
@pfackeldey - Great! Thanks for fixing it.

PR #3615, opened 2025-08-11
perf: avoid method chaining for some shape_generators

pfackeldey:
this PR improves performance by avoiding creation of unnecessary shape_generators where the actual shape is already known. This can become noticeable in cases where trivial getitems (or similar) are chained _many_ times in very _large_ layouts, i.e. coffea's NanoEvents data structure.

### DIFF
--- a/src/awkward/_nplikes/array_module.py
+++ b/src/awkward/_nplikes/array_module.py
@@ -357,7 +357,7 @@ class ArrayModuleNumpyLike(NumpyLike[ArrayLikeT]):
                     next_shape,
                     x.dtype,
                     lambda: self.reshape(x.materialize(), next_shape),  # type: ignore[union-attr]
-                    lambda: next_shape,
+                    None,
                 )
             else:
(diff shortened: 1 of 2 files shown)


ikrommyd:
Hi @pfackeldey why not use the assert never generator in these cases? On phone so is there something I’m missing?

pfackeldey:
> Hi @pfackeldey why not use the assert never generator in these cases? On phone so is there something I’m missing?

`assert_never` is set implicitly whenever the shape was unknown and then inferred from calling the generator once. If the shape is known already there's no point in setting it to `assert_never` as we won't hit this anyway. These are the cases of this PR where we have _known_ shapes, so using `None` (16B) is a bit cheaper than `assert_never` (160B). 
My point is: we shouldn't worry about code paths that are never hit anyway...

ianna (approved):
@pfackeldey - Looks great! Thanks for speeding it up!

PR #3620, opened 2025-08-14
refactor: VirtualArrays

### LINKED ISSUE #3604: Proposal: Rename `VirtualArray` in `_nplikes.virtual` to avoid confusion with `ak.Array`
The class [awkward._nplikes.virtual.VirtualArray](https://github.com/ikrommyd/awkward/blob/554fcdd060e61b641ecc4016316544556cc59b0f/src/awkward/_nplikes/virtual.py) has an interface that closely resembles `np.ndarray` — e.g. `.T`, `.dtype`, `.shape`, `.copy()`, `.view()`, etc.

However, its name — `VirtualArray` — can be misleading because:

  *  It suggests similarity to the high-level `ak.Array`, which has very different semantics and API.

   * It lives in `_nplikes`, and is clearly intended as a NumPy-like backend buffer — not part of the layout system or high-level user-facing types.

  *  It is used to represent deferred or lazy array data, not high-level structured content.

## Why the current name is problematic

  *  It causes confusion for new contributors trying to distinguish between layout classes, high-level types, and low-level buffers.

  *  The name `VirtualArray` has historically been used in Awkward layouts (v1), so this can create further ambiguity.

   * It makes it harder to explain or reason about tests and code paths that involve both `ak.Array` and `VirtualArray`.

## Suggested rename

To reduce confusion and clarify intent, I propose renaming it to something more accurate and self-descriptive, such as:

    VirtualNumpyArray

    LazyNumpyLike

    DeferredArrayLike

    VirtualBufferLike

    NumpyThunk

## Benefits

   * Makes the role of the class clearer as a low-level, NumPy-like array object.

   * Prevents accidental assumptions that it's part of the layout or high-level API.

   * Aids onboarding, refactoring, and testing by making internal roles more self-evident.

pfackeldey:
This PR does some refactoring for VirtualArrays:
- `VirtualArray` -> `VirtualNDArray` (closes https://github.com/scikit-hep/awkward/issues/3604)
- add new base class: `MaterializeableArray`
- `PlaceholderArray` implements its own `materialize` function that raises with a much better error than before; also we get rid of _many_ assertions extra for `PlaceholderArrays`
- `materialize_if_virtual` -> `maybe_materialize` (because now any `MaterializableArray` can materialize, not just `VirtualNDArrays`
- add `strides` to the `ArrayLike` protocol, this solves also some previous typing issues that we ignored with `type: ignore`

This PR does not touch the `TypetracerArray` on purpose. _If_ someone wants to add `def materialize(self)` with touching-behavior to them at some point it could be possible, but I'd prefer to not touch this code path as it is carefully optimized.

### DIFF
--- a/src/awkward/_nplikes/__init__.py
+++ b/src/awkward/_nplikes/__init__.py
@@ -33,7 +33,7 @@ def to_nplike(
     if isinstance(array, awkward._nplikes.virtual.VirtualNDArray):
         if not array.is_materialized and nplike.known_data:
             raise TypeError(
-                "Cannot convert a `VirtualNDArray` to a different `nplike` when its data is not yet materialized. 
+                "Cannot convert a `VirtualNDArray` to a different `nplike` when its data is not yet materialized.
                  Call `ak.materialize(array)` first to load the data before converting."
             )
         else:
--- a/src/awkward/_nplikes/__init__.py
+++ b/src/awkward/_nplikes/__init__.py
@@ -33,8 +33,8 @@ def to_nplike(
     if isinstance(array, awkward._nplikes.virtual.VirtualNDArray):
         if not array.is_materialized and nplike.known_data:
             raise TypeError(
-                "Cannot convert a `VirtualNDArray` to a different `nplike` when its data is not yet materialized.
-                 Call `ak.materialize(array)` first to load the data before converting."
+                "Cannot convert a `VirtualNDArray` to a different `nplike` when its data is not yet materialized."
+                " Call `ak.materialize(array)` first to load the data before converting."
             )
         else:
             if nplike.supports_virtual_arrays:
--- a/src/awkward/_nplikes/__init__.py
+++ b/src/awkward/_nplikes/__init__.py
@@ -30,10 +30,10 @@ def to_nplike(
 
     # We can always convert virtual arrays to typetracers
     # but can only convert virtual arrays to other backends with known data if they are intentionally materialized
-    if isinstance(array, awkward._nplikes.virtual.VirtualArray):
+    if isinstance(array, awkward._nplikes.virtual.VirtualNDArray):
         if not array.is_materialized and nplike.known_data:
             raise TypeError(
-                "Cannot convert a VirtualArray to a different nplike with known data without materializing it first. Use ak.materialize on the array to do so"
+                "Cannot convert a VirtualNDArray to a different nplike with known data without materializing it first. Use ak.materialize on the array to do so"
             )
         else:
             if nplike.supports_virtual_arrays:
--- a/src/awkward/_nplikes/__init__.py
+++ b/src/awkward/_nplikes/__init__.py
@@ -33,7 +33,8 @@ def to_nplike(
     if isinstance(array, awkward._nplikes.virtual.VirtualNDArray):
         if not array.is_materialized and nplike.known_data:
             raise TypeError(
-                "Cannot convert a VirtualNDArray to a different nplike with known data without materializing it first. Use ak.materialize on the array to do so"
+                "Cannot convert a `VirtualNDArray` to a different `nplike` when its data is not yet materialized. 
+                 Call `ak.materialize(array)` first to load the data before converting."
             )
         else:
             if nplike.supports_virtual_arrays:
--- a/src/awkward/_nplikes/array_like.py
+++ b/src/awkward/_nplikes/array_like.py
@@ -142,7 +142,17 @@ class ArrayLike(Protocol):
 
 def maybe_materialize(*args: Any) -> tuple[Any, ...]:
     """
-    A little helper function to materialize all materializable arrays in a list of arrays.
+    Returns a tuple where all `MaterializableArray` arguments have been replaced
+    by the result of calling their `.materialize()` method.
+
+    Other `ArrayLike` or `Any` arguments are returned unchanged.
+
+    Args:
+        *args: Variable length argument list of MaterializableArray or ArrayLike or Any objects.
+
+    Returns:
+        tuple: A tuple where each MaterializableArray is replaced by its materialized form,
+        and other ArrayLike or Any objects are returned unchanged.
     """
     return tuple(
         arg.materialize() if isinstance(arg, MaterializableArray) else arg
--- a/src/awkward/_nplikes/array_like.py
+++ b/src/awkward/_nplikes/array_like.py
@@ -7,6 +7,7 @@ from abc import abstractmethod
 from awkward._nplikes.shape import ShapeItem
 from awkward._typing import (
     TYPE_CHECKING,
+    Any,
     DType,
     EllipsisType,
     Protocol,
@@ -31,6 +32,10 @@ class ArrayLike(Protocol):
     @property
     @abstractmethod
     def shape(self) -> tuple[ShapeItem, ...]: ...
+
+    @property
+    @abstractmethod
+    def strides(self) -> tuple[ShapeItem, ...]: ...
 
     @property
     @abstractmethod
@@ -133,3 +138,18 @@ class ArrayLike(Protocol):
 
     @abstractmethod
     def __invert__(self) -> Self: ...
+
+
+def maybe_materialize(*args: Any) -> tuple[Any, ...]:
+    """
+    A little helper function to materialize all materializable arrays in a list of arrays.
+    """
+    return tuple(
+        arg.materialize() if isinstance(arg, MaterializableArray) else arg
+        for arg in args
+    )
+
+
+class MaterializableArray(ArrayLike):
+    @abstractmethod
+    def materialize(self) -> ArrayLike: ...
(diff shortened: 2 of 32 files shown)


src/awkward/_nplikes/typetracer.py
```diff
@@ -318,6 +318,12 @@ def shape(self) -> tuple[ShapeItem, ...]:
         self.touch_shape()
         return self._shape
 
+    @property
+    def strides(self):
+        raise AssertionError(
+            "bug in Awkward Array: attempt to get the strides of a TypeTracerArray"
+        )
```
ianna:
```suggestion
        raise RuntimeError(
            "Cannot get strides of a TypeTracerArray — this array is only for simulating types and shapes, "
            "not actual memory layouts."
        )
```

ianna: @pfackeldey - you are using `AssertionError` for a runtime condition (which isn’t really an assertion — those are meant for internal invariants in dev/debug, not public API behavior).

The error message is a bit abrupt and doesn’t guide the user toward what they can do instead.

src/awkward/_nplikes/placeholder.py
```diff
@@ -53,6 +53,13 @@ def size(self) -> ShapeItem:
     def nbytes(self) -> int:
         return 0
 
+    def materialize(self):
+        msg = f"{self} should never been encountered. "
+        if self.field_path:
+            msg += f"Awkward-array attempted to access a field '{self.field_path}', but it's only a placeholder. "
+        msg += "Please report it to the developers at: https://github.com/scikit-hep/awkward/issues"
+        raise RuntimeError(msg)
```
ianna:
```suggestion
    def materialize(self):
        msg = f"{self} should never have been encountered."
        if self.field_path:
            msg += (
                f" Awkward Array tried to access a field '{self.field_path}', "
                "but it exists only as a placeholder."
            )
        msg += (
            " This is unexpected behavior — please open an issue at "
            "https://github.com/scikit-hep/awkward/issues with a minimal example."
    )
    raise RuntimeError(msg)
```

ianna: The grammar in "should never been encountered" is off — it should be “should never have been encountered” or “should never be encountered.”

The message has a "developer panic" tone instead of a user-friendly, actionable tone. If this can happen in user code through legitimate misuse, it’s better to explain why it happened before pointing to GitHub.

src/awkward/_nplikes/array_like.py
```diff
@@ -133,3 +138,18 @@ def __or__(self, other: int | bool | Self) -> Self: ...
...
+
+
+def maybe_materialize(*args: Any) -> tuple[Any, ...]:
+    """
+    A little helper function to materialize all materializable arrays in a list of arrays.
+    """
+    return tuple(
+        arg.materialize() if isinstance(arg, MaterializableArray) else arg
+        for arg in args
+    )
```
ianna:
```suggestion
from typing import Any, Tuple, TypeVar, overload

T = TypeVar("T")

def maybe_materialize(*args: T) -> Tuple[T, ...]:
    """
    Returns a tuple where all `MaterializableArray` arguments have been replaced
    by the result of calling their `.materialize()` method.
    
    Non-`MaterializableArray` arguments are returned unchanged.

    Parameters
    ----------
    *args : Any
        Arbitrary positional arguments, some of which may be `MaterializableArray`.

    Returns
    -------
    tuple
        A tuple of the same length as `args`, with any materializable arrays replaced
        by their materialized form.
    """
    return tuple(
        arg.materialize() if isinstance(arg, MaterializableArray) else arg
        for arg in args
    )
```

ianna: The docstring is vague — “materialize all materializable arrays” isn’t very explanatory for someone reading the code without context.

It uses list of arrays in the docstring but accepts `*args`, which is semantically a tuple of positional arguments, not a list.

The type annotation is correct in the return `(tuple[Any, ...])`, but you could make it more precise with a `TypeVar` if you want to preserve type information.

It’s silently skipping non-`MaterializableArray` arguments, which is fine, but could be stated explicitly.

src/awkward/_nplikes/__init__.py
```diff
@@ -30,10 +30,10 @@ def to_nplike(
 
     # We can always convert virtual arrays to typetracers
     # but can only convert virtual arrays to other backends with known data if they are intentionally materialized
-    if isinstance(array, awkward._nplikes.virtual.VirtualArray):
+    if isinstance(array, awkward._nplikes.virtual.VirtualNDArray):
         if not array.is_materialized and nplike.known_data:
             raise TypeError(
-                "Cannot convert a VirtualArray to a different nplike with known data without materializing it first. Use ak.materialize on the array to do so"
+                "Cannot convert a VirtualNDArray to a different nplike with known data without materializing it first. Use ak.materialize on the array to do so"
```
ianna:
```suggestion
                "Cannot convert a `VirtualNDArray` to a different `nplike` when its data is not yet materialized. 
                 Call `ak.materialize(array)` first to load the data before converting."
```

ianna: That message is clear enough for someone who already knows what a `VirtualNDArray` and “nplike” are, but for a broader range of awkward users, it could:

- Flow better grammatically.
- Separate the reason from the solution.
- Make “nplike” a bit more readable (or at least format it like code).

ianna (changes requested):
@pfackeldey - Looks great! Please check the review comments. To summarize:

- strides property: use a proper exception (RuntimeError/NotImplementedError) and clarify why it’s unsupported.
- materialize method: fix grammar and make the error message more user-friendly, explaining what happened and how to report it.
- maybe_materialize helper: update docstring, clarify behavior for non-MaterializableArray args, and optionally make it more type-safe.
- VirtualNDArray conversion: improve readability, separate reason from solution, and consider user- vs. developer-friendly versions.

Overall: focus on clear, helpful error messages and consistent exception types, and ensure that all other documentation is written for end users. Thanks!

src/awkward/_nplikes/typetracer.py
pfackeldey: I was following the existing errors: https://github.com/scikit-hep/awkward/blob/main/src/awkward/_nplikes/typetracer.py#L393-L401

Do you still want me to change it, or keep it like the other errors? (I can of course also update the other errors, but if someone catches those explicitly it will be a breaking change)

src/awkward/_nplikes/array_like.py
pfackeldey: The type annotation is wrong/misleading in your suggestion: `(*args: T) -> Tuple[T, ...]`. `T` may be `MaterializableArray` for the input, but `ArrayLike` for the output. I don't think that'll work unless using a contra/covariant one?

What do you think about: `(*args: MaterializableArray | ArrayLike) -> Tuple[ArrayLike, ...]` ? This shows the potential change in types, that is not obvious in the `TypeVar` case.

I'll update the docstring (we're using Google style doc strings btw...). This is however a super-internal helper function - not that it isn't good to document them, but that was my initial "laziness" of not writing the doc string in great detail 😅

src/awkward/_nplikes/typetracer.py
pfackeldey: just for the record: I fully agree that those should be `RuntimeError`s 😅 If you give me the green light to change those, I'll do that :+1:

src/awkward/_nplikes/array_like.py
pfackeldey: I just tested it. We pass _anything_ to `maybe_materialize`, I can't constrain the types here. I'll stay with `Any` and improve the doc string :+1:

src/awkward/_nplikes/typetracer.py
ianna: Please do! Thanks so much!

ianna: Oh - my mistake - it's a `TypeTracer`. I think, the code is genuinely guarding internal invariants and not user facing. `AssertionError` is better here.

pfackeldey: I'll revert this then. (RuntimeError is not wrong here either IMO, e.g. if a user hits this through dask-awkward)

ianna: One can optimize out an `AssertionError`, the `RuntimeError` check will be there always :-)

ikrommyd:
Hi @pfackeldey, the recent merges required a couple more VirtualArray -> VirtualNDArray renames + renames in the tests. Took care of that and ci should be passing again now.

src/awkward/_nplikes/virtual.py
```diff
@@ -379,3 +383,17 @@ def __dlpack_device__(self) -> tuple[int, int]:
...
+
+
+# backward compatibility
+class VirtualArray(VirtualNDArray):
+    def __init__(self, *args, **kwargs):
+        import warnings
+
+        warnings.warn(
+            "The `VirtualArray` class is deprecated and will be removed in a future release of Awkward Array. "
+            "Please plan to migrate your code to use the `VirtualNDArray` class instead.",
```
ikrommyd, line 395: I would not say deprecated here. This message points to a potential difference between the classes and a migration effort required. This should make it clear that it's just a rename and no effort is required apart from a find-replace.

PR #3621, opened 2025-08-15
fix: catch off-by-one out of bound error

### LINKED ISSUE #3619: Unexpected fancy indexing behavior
### Version of Awkward Array

2.8.6

During some debugging I ran into the following:

```python
import awkward as ak

objects = ak.Array([[1, 5], [3], [4]])
cuts = ak.Array([[1, 1], [1], [0]])
print(objects[cuts])
```

This returns `[[5, 5], [4], [4]]`. The middle `[1]` in the selection cut seems to cause an overflow to the next entry in the `objects` array to arrive at a `[4]`. If I understand the logic correctly here, what should happen instead is that this gives an `IndexError`. 

cc @pfackeldey @kratsg

agoose77:
Closes #2619 

@pfackeldey https://github.com/scikit-hep/awkward/issues/3619#issuecomment-3189510009 pointed out the location, I just made the patch.

### DIFF
--- a/awkward-cpp/src/cpu-kernels/awkward_ListArray_getitem_jagged_apply.cpp
+++ b/awkward-cpp/src/cpu-kernels/awkward_ListArray_getitem_jagged_apply.cpp
@@ -39,7 +39,7 @@ ERROR awkward_ListArray_getitem_jagged_apply(
       int64_t count = stop - start;
       for (int64_t j = slicestart;  j < slicestop;  j++) {
         int64_t index = (int64_t) sliceindex[j];
-        if (index < -count || index > count) {
+        if (index < -count || index >= count) {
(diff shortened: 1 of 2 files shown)


ianna (approved):
@agoose77 - Great. Thanks for fixing it!

pfackeldey:
Hi @agoose77, thanks!!

just linking the correct issue: closes https://github.com/scikit-hep/awkward/issues/3619.

Could you add tests on the python side?
If you want me to do that, please let me know and I'll take this over :+1:

agoose77:
@pfackeldey please take this one over the line, time is short on the ground this week!

pfackeldey (approved):
Sure, no problem. I added the test @agoose77.
This is ready from my side too now :+1:

PR #3578, opened 2025-07-17
feat: `ak.combinations(n==3)` for `cuda` backend via cupy

Moelf:
related to https://github.com/HSF/PyHEP.dev-workshops/issues/61#issuecomment-3076397870
cc @ianna @kmohrman
```python
import awkward as ak
jets = ak.Array([
    [1.1, 2.2, 3.3, 4.4],
    ],
                backend="cuda")


print("============= n=2 =============")
print(ak.combinations(jets, 2))
print("============= n=3 =============")
print(ak.combinations(jets, 3))
# print("============= n=4 =============")
# print(ak.combinations(jets, 4))
```

```
============= n=2 =============
totallen calculated in AK: 6
totallen: 6, length: 1, n: 2, replacement: False
ptr: <Index dtype='int64' len='6'>[0 0 0 1 1 2]</Index>
ptr: <Index dtype='int64' len='6'>[1 2 3 2 3 3]</Index>
content after ptr carry: [<NumpyArray dtype='float64' len='6'>[1.1 1.1 1.1 2.2 2.2 3.3]</NumpyArray>, <NumpyArray dtype='float64' len='6'>[2.2 3.3 4.4 3.3 4.4 4.4]</NumpyArray>]
[[(1.1, 2.2), (1.1, 3.3), (1.1, 4.4), (2.2, ...), (2.2, 4.4), (3.3, 4.4)]]
============= n=3 =============
totallen calculated in AK: 4
totallen: 0, length: 1, n: 3, replacement: False
ptr: <Index dtype='int64' len='4'>[0 0 0 1]</Index>
ptr: <Index dtype='int64' len='4'>[1 1 2 2]</Index>
ptr: <Index dtype='int64' len='4'>[2 3 3 3]</Index>
content after ptr carry: [<NumpyArray dtype='float64' len='4'>[1.1 1.1 1.1 2.2]</NumpyArray>, <NumpyArray dtype='float64' len='4'>[2.2 2.2 3.3 3.3]</NumpyArray>, <NumpyArray dtype='float64' len='4'>[3.3 4.4 4.4 4.4]</NumpyArray>]
[[(1.1, 2.2, 3.3), (1.1, 2.2, 4.4), (1.1, 3.3, 4.4), (2.2, 3.3, 4.4)]]
```

Note: `n=2` case is handed by a different code path, included here for reference.

[...]

PR #3622, opened 2025-08-18
fix: min reducer cuda kernel updates

ianna:


ianna:
@kmohrman - please, check if this PR fixes the issue with `ak.min`. Thanks!

kmohrman:
Thanks! I've tested it on [this example](https://github.com/scikit-hep/awkward/pull/3578#issuecomment-3189056204), but unfortunately still see the same buggy behavior as is reported there.

ianna:
close in favour of https://github.com/ianna/awkward/pull/19

PR #3625, opened 2025-08-19
ci: doc previews for PRs from forks

ariostas:
Since we're switching to a fork-based workflow we'll need to adapt the docs preview job so that it works on PRs from forks.

There was a comment saying that it can only be deployed for PRs on host repo, so I'm not sure if something will fail or if it was a self-imposed restriction.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -270,35 +270,16 @@ jobs:
         name: jupyter-cache
         path: docs/_build/.jupyter_cache
 
-  branch-preview:
-    runs-on: ubuntu-24.04
-    needs: [build-docs]
-    name: Deploy Branch Preview
-    if: ${{ github.event_name == 'pull_request' }}
-    permissions:
-      id-token: write
-      contents: read
-    env:
-      S3_BUCKET: "preview.awkward-array.org"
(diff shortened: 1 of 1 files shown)


agoose77:
At the time the deployment failed on forks because we don't provision credentials -- `pull_request_target` is risky.

ariostas:
@agoose77 Okay, that makes sense.

We could consider looking for alternatives. For example, we could use Netlify for live previews and GH pages for the "real" documentation. And that way we could ditch AWS.

Netlify has a free tier that seems good. But I'll look at other alternatives.

EDIT: Cloudflare pages looks pretty good. I'll look into that.

ariostas:
I couldn't get it to work with Cloudflare since they don't have doxygen avaliable. I think Netlify could work, but I don't know if the free plan would be limiting.

So I guess let's stick with AWS. I think we can get this to work with `workflow_run`. We would just need to move the deploy preview job to a workflow with that trigger, and then it can be triggered from the usual docs workflow. It only runs the workflow present on the default branch and has access to secrets, so it gets around the need to use `pull_request_target`.

ianna (approved):
@ariostas - Thanks! This looks great. I'm enabling auto-merge.

ariostas:
@ianna can you remove "Deploy Branch Preview" from the required checks? Since that is now in a separate workflow it won't show up here.

PR #3627, opened 2025-08-20
ci: move job to deploy doc preview to a separte workflow

ariostas:
This PR works alongside #3625 to be able to deploy doc previews for PRs from forks.

This PR needs to be merged first, and then #3625 right after.

I tested this in https://github.com/ariostas-talks/awkward-test/pull/1 to verify that it works properly.

### DIFF
--- a/.github/workflows/docs-preview.yml
+++ b/.github/workflows/docs-preview.yml
@@ -1,0 +1,77 @@
+name: Docs Preview
+on:
+  workflow_run:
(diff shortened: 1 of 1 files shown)


ariostas:
Actually, I first need to figure out how to get this to not trigger on pushes to main

ariostas:
Okay, this is ready now

ianna (approved):
@ariostas - Great! Thanks!

